PR #68, opened 2020-03-05
chore: add mdx to storybook and edit storybook config

ti10le:


package.json
```diff
@@ -12,9 +12,7 @@
     "node": ">=10.13.0"
   },
   "devDependencies": {
-    "@storybook/addon-actions": "5.3.13",
     "@storybook/addon-docs": "5.3.12",
```
viktoria-schwarz: Should we update @storybook/addon-docs and @storybook/react to the latest version already (5.3.14 both)?

PR #75, opened 2020-03-06
chore(setup): add type checking

### LINKED ISSUE #70: Typecheck job for CI
In addition to the test + coverage jobs, we should also check if typings are correct before a PR can be merged.

Basically this command should be executed:

```
tsc --noEmit
```

viktoria-schwarz:
closes #70

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,6 +62,28 @@ jobs:
           paths:
             - ./coverage
 
+  type_check:
+    docker:
+      - image: circleci/node:12
+
+    steps:
+      - checkout
+      - restore_cache:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -62,6 +62,28 @@ jobs:
...
+          command: yarn install
+      
+      - save_cache:
+          key: deps1-{{ checksum ".circleci/config.yml" }}-{{ checksum "yarn.lock" }}
+          paths:
+            - node_modules
+
+      - run:
+          name: Check typings
+          command: yarn tsc --noEmit
```
sebald: can we have this as a script in the `package.json` that we also can execute it locally! 🙂

viktoria-schwarz: ups, vergessen:)

PR #95, opened 2020-03-12
chore: box documentation

### LINKED ISSUE #84: Documentation for <Box> component
# What's in the Box!!!

Initial documentation for the `<Box>` component. The following things should be included:

- [x] difference between theme/css prop/style props/base styles
- [x] application order of theme/css prop/style props/base styles
- [x] `as` property

ti10le:
closes #84

### DIFF
--- a/packages/system/src/Box/Box.stories.mdx
+++ b/packages/system/src/Box/Box.stories.mdx
@@ -1,0 +1,107 @@
+import { Meta, Story, Preview } from '@storybook/addon-docs/blocks';
+import { Box } from './Box';
+
+<Meta title="Components/Box Component" />
+
+# Box-Component
+
+## Description
+
+`Box` is the most abstract component on top of which all other Marigold components are build.
+The `Box` allows us to apply styling via a dedicated prop (`css`) while respecting the rules and constraints of our design system.
+Instead of exposing the underlying tools that will create CSS, the `Box` component wraps them and exposes a more React-like API to style components.
+
+### Style components
+
+The style of components is defined by three different layers that are structured hierarchically. A layer can overwrite the attributes of the previous layer.
+
+**Default styles, theme styles and style props**
+
+- **default styles**
+  - add default styling to a component, even without a theme
+  - should be used to achieve minimal level of consistency between browsers
+
+  .. waiting for an example
+
+- **theme styles**
+  - style values from a specific theme can be easily used with the box
+  - the theme can overwrite the default styles
+  - in this way you can stay inside the boundaries of the design system.
+
+  .. waiting for an example
+
+- **style props**
+  - style props can be used to add and modify component styles beyond the boundaries of the design system
+
+  .. waiting for an example
+  
+> Staying inside the boundaries of the design system and its contraints should be the norm. But limiting styling only to allowed values defined by the system can be very restrictive up to a point where the design system becomes an obstacle. This is why there is a an escape hatch for when it is absolutely necessary to apply styling from a third party or style it in a way that is non-compliant with the design system.
+
+## Properties
+
+| Property       | Type                         | Default |
+| :------------- | :--------------------------- | :------ |
+| `as`           | `HTMLTag`, `ComponentType`   | `'div'` | 
+| `css`          | [`css props`](#cssProps)     |         |
+| `themeSection` | `string`                     |         |
+| `variant`      | `string`                     |         |
+| `SpacingProps` | [`space props`](#spaceProps) |         |
+
+### as-Property
+
+The HTML element used for the root node. Either a string to use a DOM element or a component. By default the `Box` component will render a `<div/>`. And while this might be fine most of the time, sometimes you would rather render a `<button/>` or an `<input/>`. To allow this, the `Box` component has a special prop called `as`, which accepts an HTML tag as input.
+
+### css-Properties <a id="cssProps"></a>
+
+The `css` props let you style elements inline, using values from your theme. The css prop allows all CSS properties and the below shorthands.
+
+| Category       | Prop                                                                                                                                                                                                                  |
+| ------------   | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


packages/system/src/Box/stories.mdx
```diff
@@ -0,0 +1,100 @@
...
+The `Box` allows us to apply styling via a dedicated prop (`css`) while respecting the rules and constraints of our design system.
+Instead of exposing the underlying tools that will create CSS, the `Box` component wraps them and exposes a more React-like API for applying styling to a component.
+
+<br />
+
+### Style components
+
+The style of components is defined by base, theme and style props. These three layers are structured hierarchically. A layer can overwrite the attributes of the previous layer.
+
+**base < theme < style props**
```
sebald: I am not sure if this headline is very descriptive 😄

packages/system/src/Box/stories.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+<br />
+
+## Description
+
+`Box` is a mutli-purpose component and the most abstract component on top of which all other Marigold UI components are built.
+The `Box` allows us to apply styling via a dedicated prop (`css`) while respecting the rules and constraints of our design system.
+Instead of exposing the underlying tools that will create CSS, the `Box` component wraps them and exposes a more React-like API for applying styling to a component.
+
+<br />
```
sebald: You are using a lot of <br>s. This shouldn't be necessary

packages/system/src/Box/stories.mdx
sebald: What about something like:

Default styles, theme styles and style props?

packages/system/src/Box/stories.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+<Preview>
+  <Story name="button-box">
+    <Box as="button" css={{ border: '1px solid black' }}>I am a simple button!</Box>
+  </Story>
+</Preview>
+
+<Preview>
+  <Story name="h1-box">
+    <Box as="h1" css={{ color: 'blue', fontSize: '32px' }}>I am a big blue headline!</Box>
```
sebald: `fontSize` should be a scale from the theme!

packages/system/src/Box/stories.mdx
```diff
@@ -0,0 +1,100 @@
...
+<br />
+
+### Style components
+
+The style of components is defined by base, theme and style props. These three layers are structured hierarchically. A layer can overwrite the attributes of the previous layer.
+
+**base < theme < style props**
+
+- **base** <br />
+  The base style is the minimum of styles each component has. With the base style the component is displayed normalized in every browser.
```
sebald: Is not really about the "minimum of styles", it's about giving some basic styling to a component. It should be used to have elements looks the same in every browser, but it could do more than that. E.g. in a `<Text>` component this can be used to the `fontfamily` is alsways used from the `theme.fontFamily.body` value.

packages/system/src/Box/stories.mdx
```diff
@@ -0,0 +1,100 @@
...
+- **theme** <br />
+  Style values from a specific theme can be easily used with the box. The theme overwrites the base styles. In this way you can stay inside the boundaries of the design system.
+- **style props** <br />
+  Custom styles can be used to add and modify component styles beyond the boundaries of the design system.
+
+> Staying inside the boundaries of the design system and its contraints should be the norm. But limiting styling only to allowed values defined by the system can be very restrictive up to a point where the design system becomes an obstacle. This is why there is a an escape hatch for when it is absolutely necessary to apply styling from a third party or style it in a way that is non-compliant with the design system.
```
sebald: What is that doing here? 😄

packages/system/src/Box/stories.mdx
sebald: Overall, I think this sections need example right next to the description. This would help to understand it better.

packages/system/src/Box/stories.mdx
```diff
@@ -0,0 +1,100 @@
...
+| :------------- | :------------------------------------------------------------------------ | :------ |
+| `as`           | `HTMLTag`, `ComponentType`                                                | `'div'` | 
+| `css`          | [Theme-UI css props](https://theme-ui.com/sx-prop#theme-aware-properties) |         |
+| `themeSection` | `string`                                                                  |         |
+| `variant`      | `string`                                                                  |         |
+| `SpacingProps` | available [space props](#spacePropsId)                                    |         |
+
+### as-Property
+
+The HTML element used for the root node. Either a string to use a DOM element or a component. By default the `Box` component will render a `<div/>`. And while this might be fine most of the times, sometimes you would rather render a `<button/>` or an `<input/>`. To allow this, the `Box` component has a special prop called `as`, which accepts all HTML tag names as input.
```
sebald:
```suggestion
The HTML element used for the root node. Either a string to use a DOM element. By default the `Box` component will render a `<div/>`. And while this might be fine most of the time, sometimes you would rather render a `<button/>` or an `<input/>`. To allow this, the `Box` component has a special prop called `as`, which accepts an HTML tag as input.
```

packages/system/src/Box/stories.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+### css-Properties
+
+The `css` properties from [Theme-UI](https://theme-ui.com/sx-prop#theme-aware-properties) are usable with the Box.
```
sebald: What do you mean by "`css` properties"?

packages/system/src/Box/stories.mdx
ti10le: Dachte so allgemeine Info das dadurch ein Rahmen vorgegeben wird aber auch spezielle style prop einzelner Komponenten möglich sind

packages/system/src/Box/stories.mdx
ti10le: A lot of <br /> or a lot of the letter s ?

packages/system/src/Box/Box.stories.mdx
```diff
@@ -0,0 +1,107 @@
...
+| `variant`      | `string`                     |         |
+| `SpacingProps` | [`space props`](#spaceProps) |         |
+
+### as-Property
+
+The HTML element used for the root node. Either a string to use a DOM element or a component. By default the `Box` component will render a `<div/>`. And while this might be fine most of the time, sometimes you would rather render a `<button/>` or an `<input/>`. To allow this, the `Box` component has a special prop called `as`, which accepts an HTML tag as input.
+
+### css-Properties <a id="cssProps"></a>
+
+The `css` props let you style elements inline, using values from your theme. The following props are usable with the Box.
```
sebald: The css prop allows all CSS properties PLUS the below shorthands

PR #97, opened 2020-03-12
docs: how to write components

### LINKED ISSUE #85: Guide: How to write components
We should establish some minimal guidelines before going into writing actual components.

Here are some ideas:

*API*
- We should expose the spacing `style props`
- We should expose the `as` prop
- We should expose the `variant` prop

*Misc*
- We should have a11y in mind and support some level of [ARIA Standard](https://www.w3.org/TR/wai-aria-practices/examples/)

viktoria-schwarz:
closes #85

### DIFF
--- a/packages/system/src/writeComponent.stories.mdx
+++ b/packages/system/src/writeComponent.stories.mdx
@@ -1,0 +1,114 @@
+import { Meta, Story, Preview } from '@storybook/addon-docs/blocks';
+
+<Meta title="Guides/How to Write and Use Components" />
+
+# How to Write and Use Components
+
+Small guidance for creating components using the `<Box>` primitive.
+
+## API
+
+### Use the Box
+
+Use the `<Box>` primitive as a basic structure to apply styling to the element you wish to build.
+With the `as` prop you define the rendered HTML element like in the example where a `<button>` element is rendered:
+
+```tsx
+import React from 'react';
+import { Box } from '@marigold/system';
+
+<Box as="button" themeSection="buttons" variant="outlined" />;
+```
+
+### Add theme
+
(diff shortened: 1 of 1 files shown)


packages/system/src/stories.mdx
```diff
@@ -0,0 +1,49 @@
+import { Meta, Story, Preview } from '@storybook/addon-docs/blocks';
+
+<Meta title="MDX/How to Write Components" />
```
sebald: Should we put this in a "guides" folder?

packages/system/src/stories.mdx
```diff
@@ -0,0 +1,49 @@
+import { Meta, Story, Preview } from '@storybook/addon-docs/blocks';
```
sebald: Merge `master`. You have to rename the story then, see #98 .

packages/system/src/stories.mdx
viktoria-schwarz: Makes sense if we have more guides at some point :+1:

packages/system/src/write-component.stories.mdx
```diff
@@ -0,0 +1,114 @@
...
+// Use the component: don't forget the imports!
+render(
+  <MarigoldProvider theme={theme}>
+    <Button /> // Default Button component instance
+    <Button mx={3} /> // Button component instance with custom margin that overrides
+    theme and given styles
+  </MarigoldProvider>
+);
+```
+
```
ti10le: It would be nice if we can see the button at the end :) Maybe withe the <Preview> and <Story> tags from storybook.

ti10le: *`<Preview>` and `<Story> tags

viktoria-schwarz: then I'll add it to the TestComponents so I can import it

PR #106, opened 2020-03-15
feat(system): Add `system` helper

sebald:
Add a `system` helper that supports us to write component for our design system and adhere to a certain API (see #97).


packages/system/src/system.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+      return (
+        <Comp {...props}>
+          {typeof children === 'function' ? children(56) : children}
+        </Comp>
+      );
+    }
+  );
+
+  /**
+   * Prop `extra` is requierd and button attributes allowed.
```
ti10le: * required

packages/system/src/system.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+   */
+  expect(() => {
+    render(<Component extra="prop" type="button" />);
+  }).not.toThrow();
+
+  /**
+   * Yay, the `href` attribute is allwoed because we're rendering
```
ti10le: *allowed

PR #136, opened 2020-03-23
feat: package for sharable tsconfig

### LINKED ISSUE #134: Move tsconfig to a dedicated config package
Follow up of #100

sebald:
Closes #134


sebald:
Depending on if this is merged first we can adjust #133 to use `@marigold/tsconfig` in all packages.

Issue #164, opened 2020-03-29
Kill @emotion/core typings


sebald:
Because they polute the global namespace with the `css` prop, where non is and it conflicts with our own css prop :(

See https://github.com/emotion-js/emotion/issues/1257 for more information.

PR #165, opened 2020-03-29
feat: introduce @marigold/storybook-config

### LINKED ISSUE #100: RFC: @marigold/config
The dependencies in our root `package.json` grow very fast and it not easy anymore which dependencies are doing what and where they belong to.

To counteract this we could move configurations to a dedicated package (something like [this](https://github.com/sebald/unscripted/tree/master/config)). This would not remove any dot files in the root directory, but it would bundle dependencies with their corresponding tool. (e.g. all the webpack loaders that are required by storybook).

sebald:
- Move storybook configuration into its own package.
- Update `@storybook/preset-typescript` (this version is really good, yay!)
- Add support for TS path mapping (via webpack plugin)
- Remove emotion typings fix (see #164)

Closes #100


sebald:
@ti10le Auto-reload works again with this one! 😎

PR #176, opened 2020-03-30
feat: implement <Label/> component

### LINKED ISSUE #139: Component - Label
- [x] create label component
- [x] create story file
- [x] create a test
- [x] create index file

Thank's to: #97 

https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=428%3A267

ti10le:
- add Label component
- edit Text component because of Label
- made better Text component test (100% coverage now)
- made better stories file for Text comp.

closes #139


packages/components/src/Text/Text.tsx
```diff
@@ -2,12 +2,14 @@ import React from 'react';
...
 
 // Define custom properties for your component
-type TextProps = {};
+type TextProps = {
+  themeSection?: string;
+};
 
 export const Text = system<TextProps, 'span'>(
-  ({ as = 'span', variant = '', children, ...props }) => {
+  ({ as = 'span', themeSection = 'text', variant, children, ...props }) => {
```
sebald: (1) you actually can just drop `variant` here, because you pass it via `...props` anyway.
(2) not sure if allowing `themeSection` to be changed here. This way `<Label>` just becomes `<Box>` with the `as` default changed. Maybe we shouldn't reuse Text everywhere?

packages/components/src/Text/Text.tsx
```diff
@@ -2,12 +2,14 @@ import React from 'react';
 import { Box, system } from '@marigold/system';
 
 // Define custom properties for your component
```
sebald: unncessary comment? :D

packages/components/src/Label/Label.tsx
```diff
@@ -0,0 +1,24 @@
...
+};
+
+export const Label = system<LabelProps, 'label'>(
+  ({ variant = 'label', htmlFor, children, ...props }) => {
+    return (
+      <Text
+        as="label"
+        themeSection="form"
+        variant={variant}
+        htmlFor={htmlFor}
```
sebald: This already is part of `label` no need to add it here.

packages/components/src/Text/Text.tsx
ti10le: hm but your words: "And also a label IS a text component" (in issue discussion) - that's why I made it like this 😊 make sense but true it's nearly the box

ti10le: Maybe we add later more TextProps. Then It's more different to the Box.

sebald: I know, but the implementation does not make sense. The quote still stands. Question is how we implement it. I guess we'll figure it out later (e.g. we could introduce a useText hook or create a typography helper for the theme).

We definitely should not break our agreement about what props will be exposed.

packages/components/src/Label/Label.tsx
ti10le: Ok. I wanted to define the prop as a duty attribute. That users get a fault by non-using htmlFor. What do you mean @sebald ?

packages/components/src/Text/Text.tsx
ti10le: Ok then I use the Box for the Label component.

packages/components/src/Label/Label.tsx
sebald: Yeah sorry. That's a good idea!

PR #205, opened 2020-04-21
feat: add <Profiler> and log function

### LINKED ISSUE #162: Use new <Profiler> component to Profile <Box>?
The `<Box>` component will be used a loooooooot. We should make sure it doens't becomes the bottleneck.

Can we use the <Profiler>? More info: https://reactjs.org/docs/profiler.html

viktoria-schwarz:
closes #162

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -1,14 +1,16 @@
-import React from 'react';
-import { Box, system } from '@marigold/system';
+import React, { Profiler } from 'react';
+import { Box, onRenderCallback, system } from '@marigold/system';
 
 type ButtonProps = {};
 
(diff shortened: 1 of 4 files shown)


sebald:
The profiler is not meant to be added to the component code. rather have a separate file/place to output some numbers, so that if we change the <Box/> we can make sure that we didn't add a performance regression 🙂

viktoria-schwarz:
Does that mean that we output the numbers for the plain Box as a "test" to see if  it still performs well? As e.g. in a test file?

sebald:
@viktoria-schwarz Yes, exactly. #162 wasn't really ready to work on :D Maybe we should add a *ready*  label.

Here is an example of what I had in mind: https://github.com/styled-system/styled-system/tree/master/benchmarks

PR #206, opened 2020-04-22
chore: add svg-component (first part to implement Icons)

ti10le:
- create icons package
- create svg component
- example icon Feedback, in first step without changing size or fill


ti10le:
@sebald @viktoria-schwarz - Can we merge despite the missing coverage? I would like to write the test in one of the next PRs.

packages/icons/src/Feedback.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Svg } from './Svg';
```
sebald, line 2: Should <SVG/> be part of components package? So that others can use it from there without having to import all the icons they don't need? And also, svg is a very genral component, it can be used for more than just icons, I guess.

packages/icons/src/Feedback.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export const Feedback = () => (
+  <Svg>
+    <path
+      fill-rule="evenodd"
+      clip-rule="evenodd"
+      d="M80 12.5H20C15.875 12.5 12.5375 15.875 12.5375 20L12.5 87.5L27.5 72.5H80C84.125 72.5 87.5 69.125 87.5 65V20C87.5 15.875 84.125 12.5 80 12.5Z"
+      fill="black"
```
sebald, line 10: Can not have a fun otherwise it will not be able to change to another color.

Same for the other attributes. Are they needed?

packages/icons/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@marigold/icons",
+  "description": "",
+  "version": "0.0.0",
+  "license": "UNLICENSED",
```
sebald, line 5: MIT

packages/icons/package.json
```diff
@@ -0,0 +1,7 @@
...
+  "main": "src/index.ts"
+}
```
sebald, line 7: Peer deps should be react

packages/icons/src/Svg.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const Svg = system<SvgProps, 'svg'>(
+  ({ size = 24, fill = 'currentcolor', children, ...props }) => {
+    return (
+      <Box
+        as="svg"
+        xmlns="http://www.w3.org/2000/svg"
+        width={size}
+        height={size}
+        viewBox="0 0 100 100"
```
sebald, line 17: Why 100?

packages/icons/src/Svg.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { Box, system } from '@marigold/system';
+
+type SvgProps = {
+  size?: number;
+  fill?: string;
```
sebald, line 6: This is part of the svg props anyway.

packages/icons/src/Svg.tsx
ti10le: At "0 0 24 24" three quarters of the icon are cut off. At "0 0 100 100" it is displayed completely. I don't understand this either, because "0 0 24 24" should work

packages/icons/src/Svg.tsx
ti10le: I think it's just the fill attribute. Width and height would be also possible.

packages/icons/src/Feedback.tsx
ti10le: Yes true. Misconception.

packages/icons/src/Svg.tsx
sebald: Are all Icons of this size? This has to be the same all the time then. If so, do they have to be that detailed? CC @Toshibot

packages/icons/src/Svg.tsx
sebald: I meant that this is not necessary to add here, since you're extending the `SVGAttributes` anyway.

ti10le: I think ist necessary for the `size` prop.
"Property 'size' does not exist on type 'SystemProps<SvgProps, "svg">'." - if I delete line 5.

sebald: Yes, I am only talking about the fill property 🙂

ti10le: Oh sure, sorry :D

packages/icons/src/Svg.tsx
Toshibot: Yes icons were all at 100x100 in the Figma file; this was to keep each of them at an easily editable size. I've shrunk them all down to 24x24 just to simplify things development-wise.

ti10le: Its working now, thank you!

packages/components/src/Svg/Svg.stories.mdx
```diff
@@ -0,0 +1,48 @@
...
+The `fill` prop is the second attribute and per default `currentcolor`.
+Specific css style prop can also be added.
+
+## Properties
+
+| Property | Type        | Default          |
+| :------- | :---------- | :--------------- |
+| `size`   | `number`    | `'24'`           |
+| `fill`   | `string`    | `'currentcolor'` |
+| `css`    | `css props` |                  |
```
sebald, line 22: Should not be exposed

sebald: PS: Is also not exposed

PR #208, opened 2020-04-28
feat: improve and cleanup <Button>

viktoria-schwarz:
- rearranged theme order: added hover and disabled as pseudo classes to the variants `primary`, `secondary` and `ghost`
- added `disabled` as a property
- updated stories, properly filled in the property table

### DIFF
--- a/packages/components/src/Button/Button.stories.mdx
+++ b/packages/components/src/Button/Button.stories.mdx
@@ -13,10 +13,9 @@ Custom CSS styles can be added with the CSS prop.
 
 ## Properties
 
-| Property              | Type                                                                                                 | Default         |
-| :-------------------- | :--------------------------------------------------------------------------------------------------- | :-------------- |
-| `variant` (optional)  | `primary.small`, `primary.large`, `secondary.small`, `secondary.large`, `ghost.small`, `ghost.large` | `primary.large` |
-| `disabled` (optional) | boolean                                                                                              | `false`         |
+| Property             | Type                                                                                                 | Default         |
+| :------------------- | :--------------------------------------------------------------------------------------------------- | :-------------- |
+| `variant` (optional) | `primary.small`, `primary.large`, `secondary.small`, `secondary.large`, `ghost.small`, `ghost.large` | `primary.large` |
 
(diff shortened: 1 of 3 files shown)


packages/components/src/Button/Button.stories.mdx
```diff
@@ -3,22 +3,20 @@ import { Button } from './Button';
...
+The style variant and size of the button can be added with the variant prop and is composed like `variant.size`.
+Custom CSS styles can be added with the CSS prop.
 
 ## Properties
 
-| Property       | Type        | Default    |
-| :------------- | :---------- | :--------- |
-| `themeSection` | `string`    | `'button'` |
-| `variant`      | `string`    | `''`       |
-| `css`          | `css props` |            |
```
ti10le, line 21: Why you've deleted the css prop?

sebald: they should not be exposed by our components ;)

packages/components/src/Button/Button.tsx
```diff
@@ -1,12 +1,20 @@
 import React from 'react';
 import { Box, system } from '@marigold/system';
 
-type ButtonProps = {};
+type ButtonProps = {
+  disabled?: boolean;
```
sebald, line 5: isn't that part of the `ButtonAttributes` anyway? Do we have to add this ourselves 🤔

packages/components/src/Button/Button.stories.mdx
viktoria-schwarz: because it's a "default" property for all components and specified in the Box; I thought it's redundant as we don't list all Button Attributes either, but we can leave it if you want to :)

packages/components/src/Button/Button.tsx
viktoria-schwarz: true, we don't need it

packages/components/src/Button/Button.stories.mdx
ti10le: It's ok for me. I think we will see how good the users understand the docs :)

PR #209, opened 2020-04-28
feat: update text component

viktoria-schwarz:
- removed headings and variant
- adjusted tests, description, property tables and examples
- added correct margin styles for `p` element

### DIFF
--- a/packages/components/src/Text/Text.tsx
+++ b/packages/components/src/Text/Text.tsx
@@ -4,9 +4,9 @@ import { Box, system } from '@marigold/system';
 type TextProps = {};
 
 export const Text = system<TextProps, 'span'>(
-  ({ as = 'span', variant, children, ...props }) => {
+  ({ as = 'span', children, ...props }) => {
     return (
-      <Box as={as} themeSection="text" variant={variant} {...props}>
+      <Box {...props} as={as} themeSection="text" variant={as}>
         {children}
       </Box>
     );
--- a/packages/components/src/Text/Text.tsx
+++ b/packages/components/src/Text/Text.tsx
@@ -4,9 +4,9 @@ import { Box, system } from '@marigold/system';
 type TextProps = {};
 
 export const Text = system<TextProps, 'span'>(
-  ({ as = 'span', children, ...props }) => {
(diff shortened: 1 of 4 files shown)


packages/components/src/Text/Text.tsx
```diff
@@ -4,9 +4,9 @@ import { Box, system } from '@marigold/system';
 type TextProps = {};
 
 export const Text = system<TextProps, 'span'>(
-  ({ as = 'span', variant, children, ...props }) => {
+  ({ as = 'span', children, ...props }) => {
     return (
-      <Box as={as} themeSection="text" variant={variant} {...props}>
+      <Box {...props} as={as} themeSection="text" variant={as}>
```
sebald, line 9: Hm, why using the `as` as variant? What about if people want to use more semantic variants like `display`, `caption` etc.

viktoria-schwarz: Hm, I thought `span` and `p` would be enough to begin with. How does a `caption` tag differ from a `span`?

sebald: captain is not a HTML element. What I meant is, that only allowing to use variants that are HTML elements, we limit the expressiveness of the Text component.

Does that make more sense? Basically, most of the time you'll end up having something more semantic, like text for body, text for subtitle, etc. Take Material Design for example: https://material.io/design/typography/the-type-system.html#type-scale

viktoria-schwarz: I see the point that text is applicable in all the different contexts and styles, but then my changes don't make much sense. Should we discuss how to structure it then and how to properly use the variant here?

sebald: Sure. I think just let the user choose any variant (as a string) is fine here, like with every component we built 🙃

viktoria-schwarz: Okay, but then how do we determine the Text element itself? Should we always use a `span` ?

sebald: Can we also not use the `as` prop for the `variant` as discussed? :)

Issue #192, opened 2020-04-07
Implement Icons


ti10le:
## Add Icons to marigold.

- as svg files

Icons:
https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=428%3A269

- [x] add icons as svg component to marigold
- [x] write tests
- [x] stories file
- [x] seperate folder for the icon categories

ti10le:
- I prefer an extra icon package with nothing but the marigold icons
- then I would create an icon component (+ tests and stories file)

PR #214, opened 2020-05-04
chore: add icons package with social icons

ti10le:
#192


ti10le:
We can merge this after this PR #216

packages/icons/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@marigold/icons",
+  "description": "",
+  "version": "0.0.0",
+  "license": "MIT",
+  "main": "src/index.ts",
+  "devDependencies": {
+    "react": "16.13.1"
```
sebald, line 8: peerDependency (see component package)

packages/icons/package.json
```diff
@@ -0,0 +1,10 @@
...
+  }
```
sebald, line 9: @marigold/components is missing as a dependency!

PR #250, opened 2020-05-12
refa: standardize tests

viktoria-schwarz:
* use the same test patterns in each component
* make tests more reader-friendly


packages/components/src/Label/Label.test.tsx
```diff
@@ -1,50 +1,76 @@
...
 
 const theme = {
   form: {
     label: {
       fontFamily: 'Inter Regular',
-      fontSize: 14,
-      lineHeight: 24,
     },
-    labels: {
+    primary: {
```
sebald, line 11: Because this is a form, this does not make sense :D

packages/components/src/Text/Text.tsx
```diff
@@ -4,7 +4,7 @@ import { Box, system } from '@marigold/system';
 type TextProps = {};
 
 export const Text = system<TextProps, 'span'>(
-  ({ as = 'span', variant, children, ...props }) => {
+  ({ as = 'span', variant = 'body', children, ...props }) => {
```
sebald, line 7: Why? Plus -> this really should be an extra PR.

packages/components/src/Svg/Svg.test.tsx
```diff
@@ -1,50 +1,47 @@
...
 
-  expect(element2.getAttribute('fill')).toEqual('orange');
+  expect(svg.getAttribute('width')).toEqual('24');
 });
 
-test('support size prop', () => {
-  const t1 = render(
-    <Svg title="test3" size={32}>
-      <path d="M9.9 20.113V13.8415H14.1V20.113H19.35V11.751H22.5L12 2.34375L1.5 11.751H4.65V20.113H9.9Z" />
+test('supports size prop', () => {
```
ti10le, line 27: Support fill prop is missing.

packages/components/src/Text/Text.tsx
viktoria-schwarz: I thought to set a default variant. But I'll just treat the Text component in the other PR completely

sebald:
@viktoria-schwarz tried to resolve the conflict, but I rather have you do it before I break something 😇

Issue #249, opened 2020-05-12
Let icons support size and fill prop


ti10le:
--> continuation from 192 

the icons must be able to pass the two props to the SVG component.

ti10le:
Hey @sebald @viktoria-schwarz 
How should we test the icons? I think it's too much to write the same test for every Icon. An example with one icon should be the best way. What do you think? (Actually jest ignores the icons folder)

sebald:
Yeah, I guess not having tests for icons is fine.

PR #261, opened 2020-05-18
feat: add a test for all icons

ti10le:
#249


packages/icons/src/Icon.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { system } from '@marigold/system';
+import { Svg } from '@marigold/components';
+
+type IconProps = {
+  size?: number;
+};
+
+export const Icon = system<IconProps, 'svg'>(
```
viktoria-schwarz, line 9: change name to ExampleIcon so it's not confused with a real component

ti10le:
Rename to: ExampleIcon.tsx and ExampleIcon.test.tsx

packages/icons/src/ExampleIcon.test.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { ExampleIcon } from './ExampleIcon';
```
sebald, line 3: Why not just use one of the existing icons?

ti10le: Because the exclude from the icons folders by jest. How to easily choose an icon without change the whole folder structure? 
You have a better idea? I am not sure about this and then chose this way.

sebald: Wer only exclude for coverage

ti10le: We are also exclude for jest in config/jest/index.js
What should I do? 
- Make exclude a little bit more complicated 
- or a seperate Example Icon like in this PR? 
@sebald @viktoria-schwarz

sebald: Can you give me the line where it is ignored? 🤔

ti10le: Sorry, my fault.

PR #263, opened 2020-05-18
chore: add component template script

### LINKED ISSUE #247: Script to create component template
In order to enforce consistency and help with some gotchas, we should have a script that generates component a boilerplate.

It should

- generate a component folder inside ‘packages/components/<name>’
- Barrel file (index.ts) as well as component, test and story file
- each generate file should have the base content
  - Test: all common Tests (e.g. internals can not be overridden)
  - first story
  - component file: import Box, empty react component and props

viktoria-schwarz:
* new command `yarn create:component <name>` for creating a new component
* creates new folder with component name and tsx file, index file, test file and stories file
* excluded templates from linting, typecheck, coverage and test
* updated Readme

closes #247


scripts/template/Component.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+const theme = {};
+
+test('supports default variant and themeSection', () => {});
+
+test('accepts other variant than default', () => {});
+
+test('renders correct HTML element', () => {});
+
+test('supports custom prop', () => {});
```
ti10le, line 14: Do we call it "other variant" or "custom prop" overall or we have to change here to the name of the variant/prop in de actual component? @sebald @viktoria-schwarz

scripts/template/Component.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { Box, system } from '@marigold/system';
+
+type Props = {};
```
ti10le, line 4: Maybe here COMPProps?

scripts/template/Component.test.tsx
viktoria-schwarz: this test should test if e.g. in Icon it supports the fill or size prop, so you need to insert the prop name here if there is one :)

scripts/template/Component.tsx
viktoria-schwarz: need to insert an extra sed for it but we can do it :+1:

scripts/template/Component.test.tsx
ti10le: Yes, I think also it's better to add the prop name 👍

scripts/template/Component.tsx
ti10le: I think it's better that we have separate names like in the other comps :)

PR #267, opened 2020-05-19
add simple test for variant in theme

### LINKED ISSUE #223: Add a test in the Box if the given variant key exists in theme
The `<Box>` should test if a given `variant` key exists in the theme and (in development mode) outputs an error message if not.

viktoria-schwarz:
refers to and closes #223 
@sebald I know the underscore lodash is not the best, but idk how to do it in a better way.
This is just to get an idea of how it could work


viktoria-schwarz:
I know the tests all pass now but the error is still logged in the console. Don't know how to chance the tests to not do it

Issue #146, opened 2020-03-25
Component - Button w Icon


Toshibot:
Add the ability to display icons in the standard Button Component as outlined [here](https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=428%3A0).

Sizes:
- [ ] Large
- [ ] Small

Colors:
- [ ] White (Primary & Secondary)
- [ ] Black (Ghost)

ti10le:
Hey @sebald @viktoria-schwarz do you have ideas how to implement the icon button? 

In my opinion there are two parts but I can't stuck them together. 
- On one side we have the theme where we can add the icon params (x/y padding etc.) like 'primary.large.icon'.
- On the other hand we have the icon. How to add the <Icon /> to the button component? 

We could do something like this:
``` 
<Button variant="primary.small.icon">
   <Ticket /> Primary
 </Button> 
```

But then the user have to look about the variant with icon param and have to add the icon themselve. Is there an easier way? And I dont know if this would work 😄

ti10le:
![image](https://user-images.githubusercontent.com/59875255/82307694-003da800-99c1-11ea-880a-8951bece7065.png)

At the moment this is the reason if we type this: 
```
<Button variant="primary.small">
  <Ticket /> Primary
</Button>
```

viktoria-schwarz:
From a user perspective I'd find it easiest if we can use it like
```
<Button icon="Facebook" variant="primary.small">
	Buttontext
</Button>
```
and the icon automatically appears in the size mentioned in the button variant. Depending on the button variant, the icon is rendered in the respective color like Toshi mentioned :)

ti10le:
First I try to write a test :) But this is a good approach.

viktoria-schwarz:
Another idea (and maybe easier to implement) would be like this:
```
<Button variant="primary.small">
	<Facebook />
	<Text> blabla </Text>
</Button>
```

and adjust the Button and its children for spacings and alignment like

```
<Box as="button" themeSection="button" variant={variant} {...props}>
    <span style={{ display: 'inline-flex', alignItems: 'center' }}>
        {children}
    </span>
</Box>
```
with respective additions in the theme (add 'inline-flex' to `button.root` and a marginRight to the Svg).

ti10le:
This is a good method! But this two aspects are important, too. (Maybe later)
[...]

PR #284, opened 2020-05-26
feat: align icon in button

viktoria-schwarz:
#146


ti10le:
For me is this version good. But we shouldn't forget to implement the two points I have wrote in the issue.

viktoria-schwarz:
@ti10le For me this is a good first version as well and I merge it, if you want you can write a new issue for enhancing the icon button we have now :)

Issue #173, opened 2020-03-30
Component - Radios


Toshibot:
[Radios](https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=467%3A1)

- [x] create component
- [x] add CSS normalization
- [x] add theme styles
- [x] create story file
- [x] create test file
- [x] create index file

ti10le:
With the last commmit to to PR #296 the radio button is nearly completed.
Missing:
- if checked attribute ist given you cant uncheck the radio
- mouse hover pointer is only for the radio, not for the label. It should be for both

ti10le:
@sebald @viktoria-schwarz 
How we should handle this two points?

sebald:
1) That's a statement not a problem? :-D 
2) So adjust the label? I am not sure what's better for a11y here. I guess also do a pointer over the label?

ti10le:
@sebald 
1. Ok, maybe later if we need this.  
2. Yes we would have to change the label. I think basically it is better if the label is clickable.

sebald:
1. maybe it's because you have state inside the `<Radio>` as mentioned in a comment, this is not correct. Maybe lets talk about this on Monday? CC @viktoria-schwarz

sebald:
BTW, the radio does not work for me (can select multiple in the demo) is this known?

ti10le:
Oh yes, that changed. Worked at first. I'll correct it.

PR #296, opened 2020-06-02
feat: radio button component

ti10le:
#173


packages/components/src/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export const RadioButton = system<RadioButtonProps, 'input'>(
+  ({ variant = 'radio', ...props }) => {
+    return (
+      <Box
+        as="input"
+        type="radio"
+        themeSection="form"
+        variant={variant}
+        {...props}
```
sebald, line 14: Props need to come before `themeSection` etc. otherwise it can be overridden!

packages/components/src/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Box, system } from '@marigold/system';
+
+type RadioButtonProps = {};
+
+export const RadioButton = system<RadioButtonProps, 'input'>(
```
sebald, line 6: Hm, not a button?

ti10le:
The other possibility would be that they look like our icons everywhere? Is that possible? What's standard from other systems?

sebald:
IDK if there is a standard. But it should look the same on every OS I guess because otherwise styling might become complicated. 

Maybe we can get inspiration from material-ui? Theme-ui uses SVGs that are not replaceable. I thin the later approach should be fine for now.

ti10le:
Yes we orient ourselves on https://github.com/system-ui/theme-ui/blob/master/packages/components/src/Radio.js

ti10le:
@viktoria-schwarz @sebald 
I pushed a new commit. Can you review and help me with the test? Coverage is not on 100%.

packages/components/src/Radio/Radio.tsx
```diff
@@ -0,0 +1,49 @@
...
+          css={{
+            position: 'absolute',
+            opacity: 0,
+            zIndex: -1,
+            width: 1,
+            height: 1,
+            overflow: 'hidden',
+          }}
+          onChange={() => toggleChecked()}
+          onClick={() => toggleChecked()}
```
viktoria-schwarz, line 26: coverage is not on 100% because there are no tests for click events (as we talked about)

packages/components/src/Radio/Radio.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Circle0, Circle1 } from '@marigold/icons';
+
+type RadioProps = {};
+
+export const Radio = system<RadioProps, 'input'>(
+  ({ variant = 'radio', checked, ...props }) => {
+    const [isCheckedOn, setChecked] = useState(false);
+    const toggleChecked = () => setChecked(!isCheckedOn);
+    return (
+      <Box>
```
viktoria-schwarz, line 12: add css: display: inline-block

packages/components/src/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const radio = screen.getByTitle(/radio/);
+
+  expect(radio.getAttribute('type')).toEqual('radio');
+});
+
+test('supports checked prop', () => {
+  render(<Radio title="radio" checked />);
+  const radio = screen.getByTitle(/radio/);
+
+  expect(radio.getAttribute('checked')).toBeDefined();
```
viktoria-schwarz, line 36: is "null" and therefore true

packages/components/src/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+});
+
+test('supports HTMl radio default props', () => {
+  render(<Radio title="radio" id="RadioTest" disabled />);
+  const radio = screen.getByTitle(/radio/);
+
+  expect(radio.getAttribute('id')).toEqual('RadioTest');
+  expect(radio.getAttribute('disabled')).toBeDefined();
```
viktoria-schwarz, line 44: same as above, toBeDefined() returns true for all attributes because they are null

packages/components/src/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+test('variant styles cannot be overridden with CSS prop', () => {
+  render(
+    <MarigoldProvider theme={theme}>
+      <Radio title="radio" css={{ background: 'blue' }} />
+    </MarigoldProvider>
+  );
+  const radio = screen.getByTitle(/radio/);
+
+  expect(radio).not.toHaveStyle('background-color: blue');
```
viktoria-schwarz, line 55: radio doesn't have any defined background-color (just white as default).
The test should show that you can't override given variant styles (in this case only color) and not any new ones :)

packages/components/src/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+});
+
+test('renders <input> element', () => {
+  render(<Radio title="radio" />);
+  const radio = screen.getByTitle(/radio/);
+
+  expect(radio instanceof HTMLInputElement).toBeTruthy();
+});
```
viktoria-schwarz, line 63: Maybe add a Test to show that the correct SVG (the Circle0 or Circle1) is rendered

Issue #144, opened 2020-03-24
Document Design Principles


sebald:
**By "design principles" I mean on a technological level, not design design.**

--- 

When we really start to do some documentation, I guess we also need to tell people about the philosophy and our thinking behind the components and their API. For example, why did we chose the "Box-principle" or why do we use layout components (#118)?

This issue should be used to collect these things. I'll update the description based on our discussion below.

---

## Structure

Let's plan a structure of the docs first:

**INTRODUCTION**

- [x] Getting Started (#1744)
- [x] About (#1761)
- [ ] Design Principles
  - strict vs flexible API
  - style props + variants + theming
  - composition
  - a11y
- FAQ
  - [x] What are Design Systems (#1752) 
  - [ ] Why is component `X` missing?
  - [ ] Why React?
  - [ ] Why emotion (or CSS-in-JS)?

**CONCEPTS**

- [ ] Primitives
  - Box
- [ ] Styling
  - Design Tokens
  - CSS-in-JS
  - Responsive Values
- [x] Layout
  - isolated layouts (margin considered harmful)
- [ ] Theming

ti10le:
We will do this when we have the Marigold docs gatsby page.
Maybe we get content from UX/UI team

viktoria-schwarz:
For the record: https://visly.app/blogposts/layout-isolated-components

ti10le:
Todos:
- [ ] Why CSS-in-JS?
- [X] The Box as a primitive
- [ ] Design Tokens + Scales / 8pt Grid (maybe infos from [Confluence Marigold Layout](https://confluence.rsvx.it/display/DESIGN/Layout)
- [ ] Responsive Style Values (@theme-ui/css)
- [X] Theming (design token interpolation + variants)
- [ ] Layout components (margin considered harmful)

ti10le:
> * Design Tokens + Scales / 8pt Grid (maybe infos from [Confluence Marigold Layout](https://confluence.rsvx.it/display/DESIGN/Layout))

Should we use the same doc like in confluence? With all links an images? 
@sebald @viktoria-schwarz @andres-dediego
I would say yes.
Have made a branch to start with this guide: origin/guide-design-tokens

viktoria-schwarz:
I don't think we have implemented all the functions yet, nor the 8 pt grid or am I wrong?

sebald:
That's not a component but rather a design principle :)

[...]

PR #298, opened 2020-06-04
feat(docs): First part of "concepts & principles"

sebald:
- Design tokens
- Harmony and Rhythm

Ref #144

### DIFF
--- a/packages/system/src/concepts-principles.mdx
+++ b/packages/system/src/concepts-principles.mdx
@@ -51,9 +51,9 @@ $base: 16px;
 
 /**
  * The formular is always "$base * $ratio ** $level", the level is a integer.
- * The resulting value is usuallay rounded to avoid poor sub-pixel rendering.
+ * The resulting value is usually rounded to avoid poor sub-pixel rendering.
  *
- * Note that in the below example the number is not euqal to the $level in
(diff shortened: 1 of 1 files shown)


packages/system/src/concepts-principles.mdx
```diff
@@ -0,0 +1,84 @@
...
+$ratio: 1.5;
+
+/* browser default */
+$base: 16px;
+
+/**
+ * The formular is always "$base * $ratio ** $level", the level is a integer.
+ * The resulting value is usuallay rounded to avoid poor sub-pixel rendering.
+ *
+ * Note that in the below example the number is not euqal to the $level in
```
ti10le, line 56: *equal

packages/system/src/concepts-principles.mdx
```diff
@@ -0,0 +1,84 @@
...
+```scss
+/* perfect fifth */
+$ratio: 1.5;
+
+/* browser default */
+$base: 16px;
+
+/**
+ * The formular is always "$base * $ratio ** $level", the level is a integer.
+ * The resulting value is usuallay rounded to avoid poor sub-pixel rendering.
```
ti10le, line 54: *usually

PR #441, opened 2020-08-20
feat: add field component

### LINKED ISSUE #440: Component - Field
- [ ] create component
- [ ] add CSS normalization
- [ ] add theme styles
- [ ] create story file
- [ ] create test file
- [ ] create index file

[Figma Design](https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=471%3A0)

#160

ti10le:
closes #440 
#562

### DIFF
--- a/packages/components/src/Field/Field.tsx
+++ b/packages/components/src/Field/Field.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import { Input, Label, ValidationMessage } from '@marigold/components';
+import { useStyles, system } from '@marigold/system';
+import { Exclamation, Required } from '@marigold/icons';
+import b2bTheme from '@marigold/theme-b2b';
+
+type FieldProps = {
+  variant?: string;
+  htmlFor: string;
+  label: string;
+  error?: string;
+};
+
+export const Field = system<FieldProps, 'input'>(
+  ({
+    variant = 'field',
+    type = 'text',
+    htmlFor,
+    label,
+    error,
+    className = '',
+    ...props
+  }) => {
+    const classNames = useStyles({
+      variant: `form.${variant}`,
+    }).concat(' ', className);
+
+    return (
+      <div>
+        <Label htmlFor={htmlFor} className={classNames} {...props}>
+          {label}
+          {error ? <Required size={16} fill={b2bTheme.colors.red60} /> : ''}
+        </Label>
+        <Input type={type} for={htmlFor} />
+        {error ? (
+          <ValidationMessage>
+            <Exclamation size={16} />
+            {error}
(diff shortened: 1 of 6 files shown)


ti10le:
Hey @sebald @viktoria-schwarz, 
in this PR I created a field component like in this prototype: https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=471%3A0

My problem is: in Field.tsx are a lot of special b2b style and I don't know which is the best solution. 
- We can add more to the theme and we need more parameters to get the Validation icons to the component. I think maybe it would be a confusing comp. 

I wanted to ask you before I put in too much unnecessary work. What do you mean?

packages/components/src/Field/Field.tsx
```diff
@@ -0,0 +1,87 @@
...
+  label?: string;
+  error?: string;
+};
+
+export const Field = system<FieldProps, 'input'>(
+  ({
+    variant = 'field',
+    type = 'text',
+    label,
+    htmlFor = 'id',
```
sebald, line 17: This should be generated randomly. Otherwise it will not work properly.

packages/components/src/Field/Field.tsx
```diff
@@ -0,0 +1,87 @@
...
+    ...props
+  }) => {
+    return label ? (
+      <div>
+        <Label htmlFor={htmlFor} variant="fieldLabel">
+          {label}
+          {error ? (
+            <Required
+              size={16}
+              fill="#dd4142"
```
sebald, line 28: Can we get this from the theme?

packages/components/src/Field/Field.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <Box
+          as="input"
+          type={type}
+          themeSection="form"
+          variant={variant}
+          {...props}
+          css={
+            error
+              ? {
+                  border: '1px solid #dd4142',
```
sebald, line 46: From theme? Theme-ui support borders, we could create a general "borders.error" or something :)

sebald (changes requested):

packages/components/src/Field/Field.tsx
ti10le: ok sure. I had not a good idea for this.

ti10le: @sebald can you help me with this? I don't know how to use this additionally to the normal variant.

viktoria-schwarz (approved):

Issue #562, opened 2020-11-03
Use classnames in all existing components


ti10le:
This Issue is for tracking the state of rewriting all components. We want to use the new classnames in all components. --> #361 

The person who works on a component writes his name behind the component in the list. After finishing the checkbox should be checked.

## Components with new classnames

- **Typography**
	- [x] Text 
	- [x] Heading
	- [x] Link
- **Form**
	- [x] Button
	- [x] Label
	- [x] Input
	- [x] Textarea
	- [x] Select
	- [x] Checkbox
	- [x] Radio
	- [x] Slider
	- [x] Field
- **Content**
	- [x] Message
	- [x] Alert
	- [x] Image
	- [x] Badge
	- [x] Validation Message
- **Layout**
	- [x] Container
	- [x] Hidden
	- [x] Divider
- **Iconography**
	- [x] SVG
	- [x] Icon

PR #589, opened 2020-11-17
Rewrite Label and Input

ti10le:
#562


packages/components/src/Input/Input.tsx
```diff
@@ -1,21 +1,20 @@
...
-        type={type}
-        themeSection="form"
-        variant={variant}
-        {...props}
-        css={{
-          border: 0,
-        }}
-      />
+      <input type={type} className={classNames} {...props}>
+        {children}
```
viktoria-schwarz, line 16: remove children

viktoria-schwarz (approved):

PR #608, opened 2020-11-25
use new useStyles hook in Text component

### LINKED ISSUE #595: Improve API of `createStyles` for common use cases
Currently setting up the hook is a two step process (1) variant base and normalization (2) actual hook. This is bad for end users. A regular hook would be better. Also, it would be nice if it could handle an additional `className` and append that to the resulting `classNames` string.

Can we remove the setup step and still have a good API for the hook?

**Usage**

```js
const classNames = useStyles({ className, variants, ...styles })
```

**API**

- **className (string):** append to resulting classNames
  Why? We need to be able to easily add "external" CSS classes if necessary.
- **variants (string|string[]):** list of variants from theme to apply
  Why? In some cases (buttons) we need to support multiple variants. Instead of specifying the base variant we can just `base.${variant}` ... maybe?
- **...styles:** custom styles that are interpolated through theme-ui  

---

We could also add a "variantBase" or similar to the parameters of `useStyles`.

ti10le:
closes #595


packages/components/src/Text/Text.stories.mdx
```diff
@@ -28,7 +29,7 @@ import { Text } from '@marigold/components';
 
 <Preview>
   <Story name="body">
-    <Text>
+    <Text className={useStyles({ bg: '#f3f3f3' })}>
```
sebald, line 32: Geht das? O.o

packages/components/src/Text/Text.tsx
```diff
@@ -7,17 +7,19 @@ type TextProps = {
...
     textColor = 'inherit',
+    className = '',
     children,
     ...props
   }) => {
-    const classNames = useStyles({ variant, color: textColor });
+    const classNames = useStyles({
+      variant: `text.${variant}`,
+      color: textColor,
+    }).concat(' ', className);
```
sebald, line 22: Das sollten wir vielleicht im Nachhinein in useStyles packen

sebald (commented):
Sieht gut aus. Was failed?

ti10le:
@sebald das failed: https://github.com/marigold-ui/marigold/pull/608/checks?check_run_id=1452314094

sebald:
That's on me. You can not use hooks inside loops. Guess we have to make UseClassname accept multiple inputs.

viktoria-schwarz (approved):

PR #721, opened 2021-01-25
Desgin update: message

### LINKED ISSUE #716: Design update - Message
Update color bar and title section, see [figma](https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=582%3A119)

viktoria-schwarz:
closes #716

### DIFF
--- a/packages/components/src/Message/Message.stories.mdx
+++ b/packages/components/src/Message/Message.stories.mdx
@@ -13,7 +13,7 @@ Simple component to display a message with a title and icon based on a `div` ele
 
 | Property             | Type     | Default |
 | :------------------- | :------- | :------ |
-| `title`              | `string` |         |
+| `messageTitle`       | `string` |         |
 | `variant` (optional) | `string` | `info`  |
 
 ## Import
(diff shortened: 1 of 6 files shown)


packages/components/src/Message/Message.stories.mdx
```diff
@@ -25,6 +26,18 @@ import { Message } from '@marigold/components';
 
 <Preview>
   <Story name="Message">
-    <Message>Hello, I am a simple message.</Message>
+    <Message variant="warning" title="Danger Zone!">
```
ti10le, line 29: Info: if i remove title prop here, storybook shows just the icon and no error. I think its special storybook because in the test I get an error if I remove title prop.

ti10le (approved):

packages/components/src/Message/Message.stories.mdx
viktoria-schwarz: You're right I shouldn't overwrite title. I rename it to messageTitle

ti10le (approved):

PR #735, opened 2021-02-01
feat(system): Expose type aliases from system

sebald:


sebald:
I need this to make the `<Box>`🙈  ... but it also helps with "leaking types".

ti10le (approved):

PR #771, opened 2021-02-09
add Stack component

### LINKED ISSUE #770: Component - Stack
- [x] create component
- [x] create story file
- [x] create test file
- [x] create index file

ti10le:
closes #770

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "pretty-quick": "3.1.0",
     "react": "17.0.1",
     "react-dom": "17.0.1",
-    "react-flatten-children": "1.1.2",
+    "react-keyed-flatten-children": "1.3.0",
     "shelljs": "0.8.4",
     "ts-node": "9.1.1",
     "tsdx": "0.14.1",
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "pretty-quick": "3.1.0",
     "react": "17.0.1",
     "react-dom": "17.0.1",
+    "react-flatten-children": "1.1.2",
     "shelljs": "0.8.4",
     "ts-node": "9.1.1",
     "tsdx": "0.14.1",
--- a/packages/components/src/Stack/Stack.test.tsx
+++ b/packages/components/src/Stack/Stack.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
 import { Stack, Text } from '@marigold/components';
+import flattenChildren from 'react-flatten-children';
 
 test('supports default space prop', () => {
   render(
@@ -58,4 +59,17 @@ test('accepts custom styles prop className', () => {
   const stack = screen.getByTitle(/stack/);
 
   expect(stack.className).toMatch('custom-class-name');
+});
+
+test('flatten children works', () => {
+  const children = [<>test1</>, <>test2</>];
+
+  flattenChildren(children).map(child => {
+    render(<div>{child}</div>);
+  });
+
+  const stack = screen.getByText(/test1/);
+  const stack2 = screen.getByText(/test2/);
+  expect(stack).toBeDefined();
+  expect(stack2).toBeDefined();
 });
--- a/packages/components/src/Stack/Stack.test.tsx
+++ b/packages/components/src/Stack/Stack.test.tsx
@@ -1,0 +1,94 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { Stack, Text } from '@marigold/components';
+
+test('supports default space prop', () => {
+  render(
+    <Stack title="stack">
+      <Text>stack</Text>
+    </Stack>
+  );
+  const stack = screen.getByTitle(/stack/);
+
+  expect(stack).toHaveStyle(`padding: 0px`);
+});
+
+test('supports custom space prop', () => {
+  render(
+    <Stack space={4} title="stack">
+      <Text>stack</Text>
+    </Stack>
+  );
+  const stack = screen.getByTitle(/stack/);
+
+  expect(stack).toHaveStyle(`padding: 32px`);
+});
+
+test('supports default align prop', () => {
+  render(
+    <Stack title="stack">
+      <Text title="stackContent">stack</Text>
+    </Stack>
+  );
+  const stack = screen.getByText(/stack/);
+
+  stack.hasAttribute(`align-items: flex-start`);
+});
+
+test('supports center align prop', () => {
+  render(
+    <Stack align="center" title="stack">
+      <Text title="stackContent">stack</Text>
+    </Stack>
+  );
+  const stack = screen.getByText(/stack/);
+
+  stack.hasAttribute(`align-items: center`);
+});
+
+test('supports right align prop', () => {
+  render(
+    <Stack align="right" title="stack">
+      <Text title="stackContent">stack</Text>
+    </Stack>
+  );
+  const stack = screen.getByText(/stack/);
+
+  stack.hasAttribute(`align-items: flex-end`);
+});
+
+test('supports two children', () => {
+  render(
+    <Stack align="right" title="stack">
+      <Text title="stackContent">stack</Text>
+      <Text title="stackContent2">secondStack</Text>
+    </Stack>
+  );
+  const stack = screen.getByText(/stack/);
+  const secondStack = screen.getByText(/secondStack/);
+
+  expect(stack).toBeDefined;
+  expect(secondStack).toBeDefined;
+});
+
+test('renders correct HTML element', () => {
+  render(
+    <Stack title="stack">
+      <Text>sdf</Text>
+    </Stack>
+  );
+  const stack = screen.getByTitle(/stack/);
+
+  expect(stack instanceof HTMLDivElement).toBeTruthy();
+});
+
+test('accepts custom styles prop className', () => {
+  render(
+    <Stack className="custom-class-name" title="stack">
+      <Text>text</Text>
+    </Stack>
+  );
+  const stack = screen.getByTitle(/stack/);
+
+  expect(stack.className).toMatch('custom-class-name');
+});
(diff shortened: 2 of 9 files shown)


packages/components/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,51 @@
...
+        p={space}
+        display={display}
+        flexDirection={flexDirection}
+        className={classNames}
+        {...props}
+      >
+        {stackItems.map((child, index) => {
+          return (
+            <Box
+              key={index.toString()}
```
ti10le, line 37: This is because of a react console error that every child needs an unique key. Is this ok or do you have a better idea? @viktoria-schwarz @sebald

package.json
```diff
@@ -34,6 +34,7 @@
     "pretty-quick": "3.1.0",
     "react": "17.0.1",
     "react-dom": "17.0.1",
+    "react-flatten-children": "1.1.2",
```
ti10le, line 37: use this to get every child of the Stack

packages/components/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import { useStyles, system } from '@marigold/system';
+import { Box } from '../Box';
+import flattenChildren from 'react-flatten-children';
+
+type StackProps = {
+  space?: 0 | 4 | 8 | 16 | 24 | 32 | 40 | 48 | 56 | 64 | 88;
```
sebald, line 7: Why doesn't it use `spaces` from the theme 🤔

packages/components/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,51 @@
...
+  align?: 'left' | 'right' | 'center';
+};
+
+export const Stack = system<StackProps, 'div'>(
+  ({ space = 0, align = 'left', className, children, ...props }) => {
+    const classNames = useStyles({}, className);
```
sebald, line 13: Does this do anything?

packages/components/src/Stack/Stack.tsx
sebald: https://reactjs.org/docs/fragments.html#short-syntax

packages/components/src/Stack/Stack.test.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { Stack, Text } from '@marigold/components';
+
```
sebald, line 4: Maybe add a test with Fragments to see that `react-flatten-children` works :)

packages/components/src/Stack/Stack.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+test('supports default space prop', () => {
+  render(
+    <Stack title="stack">
+      <Text>stack</Text>
+    </Stack>
+  );
+  const stack = screen.getByTitle(/stack/);
+
+  expect(stack).toHaveStyle(`padding: 0px`);
```
sebald, line 13: Shouldn't that be `padding-top`?

sebald (changes requested):

packages/components/src/Stack/Stack.tsx
ti10le: I thought we dont want a dependency from theme

packages/components/src/Stack/Stack.tsx
ti10le: No, but if I give custom classNames

packages/components/src/Stack/Stack.test.tsx
ti10le: Could be both. In the first box I rendered its padding and in the box from the map() its pt. I think this is enough?

packages/components/src/Stack/Stack.tsx
sebald: No variants, but will still have to make sure that everything adheres the spaces of the whole app.

packages/components/src/Stack/Stack.tsx
sebald: You mean the base css props? The box will have them anyway.

packages/components/src/Stack/Stack.test.tsx
sebald: But you are testing the inner components. Shouldn't they only have top padding

packages/components/src/Stack/Stack.tsx
ti10le: No I mean custom user styling when i use <Stack classNames="my-custom-style">

packages/components/src/Stack/Stack.test.tsx
ti10le: No in this case i am testing the outer component.
"getByTitle"

packages/components/src/Stack/Stack.tsx
sebald: When you use? :D

packages/components/src/Stack/Stack.tsx
ti10le: hm ok but if i import space from theme i have also a dependency to the b2b theme? Os there another way?

packages/components/src/Stack/Stack.tsx
ti10le: e.g. like this: `<Stack space={24} className={useStyles({ bg: '#f3f3f3' })}>`

packages/components/src/Stack/Stack.tsx
sebald: That's not what I meant. The space printers should read from theme.space. there is no need to depend on a certain theme.

packages/components/src/Stack/Stack.tsx
sebald: But in this case you get the className anyway, so you can just put the className on the inner div.

ti10le: ooooh so right

packages/components/src/Stack/Stack.tsx
ti10le: Updated this PR :)

packages/components/src/Stack/Stack.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <Stack className="custom-class-name" title="stack">
+      <Text>text</Text>
+    </Stack>
+  );
+  const stack = screen.getByTitle(/stack/);
+
+  expect(stack.className).toMatch('custom-class-name');
+});
+
+test('flatten children works', () => {
```
sebald, line 64: You should test that with the <Stack> component :D

packages/components/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { system } from '@marigold/system';
+import { Box } from '../Box';
+import flattenChildren from 'react-flatten-children';
```
sebald, line 4: what about the other package that has more downloads? :)

packages/components/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+type StackProps = {
+  space?: Theme['space'];
+  align?: 'left' | 'right' | 'center';
+};
+
+export const Stack = system<StackProps, 'div'>(
+  ({ space = 0, align = 'left', className, children, ...props }) => {
+    let stackItems = flattenChildren(children);
+    let spaceValue: number = +space;
```
sebald, line 15: Already a Number. Don't copy/paste code blindly :P

packages/components/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,52 @@
...
+      flexDirection = 'row';
+      alignItems = 'flex-start';
+    }
+
+    return (
+      <Box
+        p={spaceValue}
+        display={display}
+        flexDirection={flexDirection}
+        className={className}
```
sebald, line 31: No need to get that in line 13. just spread it with the rest of the props!

packages/components/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,52 @@
...
+        {...props}
+      >
+        {stackItems.map((child, index) => {
+          return (
+            <React.Fragment key={index}>
```
sebald, line 36: Use the shorthand `<>`

sebald (changes requested):

packages/components/src/Stack/Stack.tsx
ti10le: I would like to use React.Children.map then we need no extra dependency, ok?

sebald: That will not cover the case of someone uses Fragments

packages/components/src/Stack/Stack.tsx
ti10le: Hm.. I tried it

packages/components/src/Stack/Stack.tsx
ti10le: OK

packages/components/src/Stack/Stack.tsx
ti10le: But then I get an error from p and pt.
Bildschirmfoto 2021-02-17 um 08 50 05

packages/components/src/Stack/Stack.tsx
ti10le: But then this test tests if flattenChildren working? 
test('supports two children' ...)

packages/components/src/Stack/Stack.tsx
sebald: Yeah, because spaces can also be an object

packages/components/src/Stack/Stack.tsx
sebald: Hm. What's the error?

packages/components/src/Stack/Stack.tsx
sebald: It needs more nesting, not multiple Childs

ti10le: Please have one more look at my changes :) :D

sebald (approved):

Issue #753, opened 2021-02-04
Remove `system` helper


sebald:
I started with #607 and was wondering if I should use the `system` helper or not. The problem is that the system helper includes the `variant` prop. Something the `<Box>` not really cares about since we have the `useStyles` hook for that.

Also, the component we already build use the `system` helper, but declare the `variant` in their own props. Even this is not necessary because it's part of the `system` helper.

https://github.com/marigold-ui/marigold/blob/a9cea1e0151a60bdc8598125b4be1bd8cb6ded01/packages/components/src/Badge/Badge.tsx#L4-L8

**So the question is, should we deprecate the `system` helper?**

Reasons:

- Both the `as` and `variant` prop should be opt in rather than always there.
- We do not need the `ref` in every component (this was the reason for the `system` helper in the first place`
- The `as` prop is only important on certain component (e.g. `<Box>` and `<Text>`)

_Originally posted by @sebald in https://github.com/marigold-ui/marigold/discussions/656_

viktoria-schwarz:
Does that mean we will change all our components to something like that?

```type AlertProps = {
  variant?: string;
  className: string;
  children: ReactNode;
};

export const Alert = ({
  variant = 'info',
  className,
  children,
  ...props
}: AlertProps) => {
  const classNames = useStyles({...})

...
```

sebald:
Yeah, I guess so. Most components will be a regular React component. Which is a good thing 🙂

PR #775, opened 2021-02-10
refa(components): Remove system helper from <Badge>

sebald:
Ref #753


sebald:
@viktoria-schwarz @ti10le I also remove two props. I think these should not be there, right? 🤔 That's what variants are for.

ti10le (approved):

ti10le (approved):

PR #776, opened 2021-02-10
refa(components): Remove system helper from <Button>

sebald:
Ref #753


ti10le (approved):

ti10le:
@sebald custom className doesn't work with this fix. See: Storybook docs page bg color override from theme

PR #777, opened 2021-02-10
feat(types): Add helper type to easy add HTML props

sebald:
Also ... smal rearrange stuff and delete stuff we don't use anyway.


sebald:
Well ... I noticed that by "just" removing the `system` helper, we'll not allowing the regular HTML props on our elements anymore. Guess we don't want that? :thinking:

sebald:
I am honestly not sure what to do here. `<Button>` is the first component where I have missed it. I wonder what is better. Allowing HTML props for a certain element (this would also include `style` which we **NEVER** want or just adding the necessary props. In case of the `<Button>` -> `type` and `onClick`.

Thoughts? @viktoria-schwarz @ti10le

viktoria-schwarz:
I agree that we do not want to have the `style` prop, but is there no way to realize this? Like we previously did in the Box, allowing all base HTML props BUT `style`.

sebald:
@viktoria-schwarz updated PR :) Omitting the `style` should work 🤞

viktoria-schwarz (approved):

PR #785, opened 2021-02-12
feat(components): Let <Box> support variant prop

### LINKED ISSUE #784: Let <Box> support variant
Starting to remove the `system` helper I noticed that the only reason to use `useStyles` is often that the `<Box>` doesn't support variants. I guess variants are such a important and crucial aspect of our components, that `<Box>` should support it.

_Originally posted by @sebald in https://github.com/marigold-ui/marigold/discussions/779_

sebald:
Closes #784

### DIFF
--- a/packages/components/src/Box/Box.tsx
+++ b/packages/components/src/Box/Box.tsx
@@ -4,6 +4,7 @@ import { ComponentWithAs } from '@marigold/types';
 
 export type BoxProps = {
   className?: string;
+  variant?: string | string[];
 
   display?: ResponsiveStyleValue<string>;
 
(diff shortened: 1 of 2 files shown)


packages/components/src/Box/Box.tsx
```diff
@@ -4,6 +4,7 @@ import { ComponentWithAs } from '@marigold/types';
 
 export type BoxProps = {
   className?: string;
+  variant?: string | string[];
```
viktoria-schwarz, line 7: I like the idea of an array of variants!

viktoria-schwarz (approved):

PR #811, opened 2021-02-22
feat: add Columns component

### LINKED ISSUE #727: Component - Columns
- [x] create component
- [x] create story file
- [x] create test file
- [x] create index file

ti10le:
closes #727


packages/components/src/Columns/Columns.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <Box
+        width={`calc(100% + ${space}px)`}
+        m={-space / 2}
+        display="flex"
+        flexWrap="wrap"
+        alignItems={alignItems}
+        justifyContent={justify}
+        {...props}
+      >
+        {Children.map(
```
sebald, line 52: Use flatten children?

packages/components/src/Columns/Columns.tsx
```diff
@@ -0,0 +1,69 @@
...
+export const Columns: React.FC<ColumnsProps> = ({
+  space = 0,
+  horizontalAlign = 'left',
+  verticalAlign = 'top',
+  className,
+  children,
+  ...props
+}) => {
+  let columnItems = flattenChildren(children);
+  let childClassNames = useStyles({ p: `${space / 2}px` });
```
sebald, line 23: Shouldn't this read from theme?

packages/components/src/Columns/Columns.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  // horizontal Alignment
+  let justify = 'flex-start';
```
sebald, line 26: We use const not let

packages/components/src/Columns/Columns.tsx
```diff
@@ -0,0 +1,69 @@
...
+        justifyContent={justify}
+        {...props}
+      >
+        {Children.map(
+          (columnItems as unknown) as React.ReactElement,
+          (child: React.ReactElement) => {
+            return React.cloneElement(
+              child,
+              {
+                className: childClassNames,
```
sebald, line 58: This does removes any className from the child, correct?

sebald (changes requested):

packages/components/src/Columns/Columns.tsx
ti10le: Used in in line 22. In Stack component and in Braid DS is it also used like this.

packages/components/src/Columns/Columns.tsx
ti10le: No I think not. It's the padding space we have to pass to the children (Column). User should only pass space to Columns and not to Column, too.

packages/components/src/Columns/Columns.tsx
ti10le: Ok but then the assignee doesn't work. Should I use state vars?

packages/components/src/Columns/Columns.tsx
ti10le: It works. Se example in stories file: width

packages/components/src/Columns/Columns.tsx
ti10le: or just var?

sebald: Hu? What's the error?

packages/components/src/Columns/Columns.tsx
sebald: Ah, I see `space` is passed from the parent? Does the parent read it from the theme?

packages/components/src/Columns/Columns.tsx
sebald: 🤔 what works?

packages/components/src/Columns/Columns.tsx
ti10le: Our discussion again :D It should read from theme but not at the moment

ti10le: Need a solution here

packages/components/src/Columns/Columns.tsx
ti10le: Error in the two lines below: Cannot assign to 'justify' because it is a constant.
In the past we used var in such cases

packages/components/src/Columns/Columns.tsx
ti10le: It takes custom classnames from user and my childClassnames. It works, both styles are applied.

packages/components/src/Columns/Columns.tsx
sebald: oh yeah sorry, I didn't caught that. I was confused by line 22-23 because you used `let` there too.

packages/components/src/Columns/Columns.tsx
sebald: you can pass in a function, e.g.

```
useStyles({ p: theme => theme.spaces[space]/2 });
```

sebald: not sure about the function signature, could also be an option. autocomplete should work here I guess 🤔

packages/components/src/Columns/Columns.tsx
```diff
@@ -0,0 +1,69 @@
...
+  if (verticalAlign === 'bottom') {
+    alignItems = 'flex-end';
+  } else if (verticalAlign === 'center') {
+    alignItems = 'center';
+  }
+
+  return (
+    <Box p={space} display="flex" className={className}>
+      <Box
+        width={`calc(100% + ${space}px)`}
```
sebald, line 44: Can you help me understand what this does? Why do we need this? This does not work, right? since `space` is not necessarily equal to the pixel, right? The other box might have a different pixel value, depending on the theme.

sebald (changes requested):

packages/components/src/Columns/Columns.tsx
ti10le: It works. This calculation is to set padding between different columns and include them to 100%. That you can do something like in the first example. Three Columns width: 4/12 + padding and you get 100%. At the moment the number you give to space is px. If we take the space values from the theme then this would be changed a little bit. It's just a proposal from me.

sebald: Yeah, sorry. you' re right. That so confusing :D but you're right!

sebald (approved):

PR #815, opened 2021-02-22
refa: Textarea design and component change

### LINKED ISSUE #720: Design & component update - Textarea
Have variant with required: label and red star, see [figma](https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=528%3A211)

ti10le:
closes #720


ti10le:
I will add docs und test changes after completing the component itself.
Can somebody review @sebald @viktoria-schwarz

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -1,22 +1,58 @@
...
+import { Exclamation, Required } from '@marigold/icons';
+import { ComponentProps } from '@marigold/types';
 
-export type TextareaProps = {
+import { ValidationMessage } from '../ValidationMessage';
+import { Label } from '../Label';
+import { Box } from '../Box';
+
+type TextareaProps = {
+  // how to check if theres a label and then htmlFor is required
```
viktoria-schwarz, line 11: I would just make both optionial and mention it in the docs

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -1,22 +1,58 @@
...
+export const Textarea: React.FC<TextareaProps> = ({
+  variant = 'textarea',
+  htmlFor = 'textarea',
+  label,
+  error,
+  required = false,
+  children,
+  ...props
+}) => {
+  let errorColor = '#dd4142';
```
viktoria-schwarz, line 28: why not taking it from the theme?

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -1,22 +1,58 @@
...
+          {label}
+          {(error || required) && (
+            <Required size={16} className={errorClassName} />
+          )}
+        </Label>
+      )}
+      <Box
+        as="textarea"
+        display="block"
+        border={error && `1px solid ${errorColor}`}
```
viktoria-schwarz, line 44: use Outline to avoid the small "jump" of 1px on click (or add transparent 1px border if not clicked)

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -1,22 +1,58 @@
...
+  error,
+  required = false,
+  children,
+  ...props
+}) => {
+  let errorColor = '#dd4142';
+  const errorClassName = useStyles({ color: errorColor });
+
+  return (
+    <Box display="block">
```
viktoria-schwarz, line 32: why `display="block"`?

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -1,22 +1,58 @@
...
+import { Label } from '../Label';
+import { Box } from '../Box';
+
+type TextareaProps = {
+  // how to check if theres a label and then htmlFor is required
   variant?: string;
-};
+  label?: string;
+  htmlFor?: string;
+  error?: string;
```
viktoria-schwarz, line 15: I would expect error to be boolean, maybe rename it to errorMessage

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -1,22 +1,58 @@
...
+  children,
+  ...props
+}) => {
+  let errorColor = '#dd4142';
+  const errorClassName = useStyles({ color: errorColor });
+
+  return (
+    <Box display="block">
+      {label && (
+        <Label {...props} htmlFor={htmlFor}>
```
viktoria-schwarz, line 34: I think having `{...props}` used twice (in Label and in the textarea) doesn't make sense for the user

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -1,22 +1,58 @@
...
+        </Label>
+      )}
+      <Box
+        as="textarea"
+        display="block"
+        border={error && `1px solid ${errorColor}`}
+        variant={`form.${variant}`}
+        {...props}
+      >
         {children}
```
viktoria-schwarz, line 48: Do we actually need children here?

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -1,22 +1,58 @@
...
+            <Required size={16} className={errorClassName} />
+          )}
+        </Label>
+      )}
+      <Box
+        as="textarea"
+        display="block"
+        border={error && `1px solid ${errorColor}`}
+        variant={`form.${variant}`}
+        {...props}
```
viktoria-schwarz, line 46: I would move the props to line 42, for consistency

packages/components/src/Textarea/Textarea.tsx
ti10le: still dont know how :D

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -1,22 +1,58 @@
...
+          {(error || required) && (
+            <Required size={16} className={errorClassName} />
+          )}
+        </Label>
+      )}
+      <Box
+        as="textarea"
+        display="block"
+        border={error && `1px solid ${errorColor}`}
+        variant={`form.${variant}`}
```
viktoria-schwarz, line 45: If you use the Box, I think we should take all the styles like padding etc. away from the theme and have them in the Box. Theme styles are then only for colors

packages/components/src/Textarea/Textarea.tsx
ti10le: Made it like in field Component. We should use both in the same way but yes maybe errorMessage is better

viktoria-schwarz: Then I would consider changing Field aswell so it's clear :)

packages/components/src/Textarea/Textarea.tsx
viktoria-schwarz: Actually you don't really need to expose the color in the component, isn't it enough to move it to the theme as it would be better if it changes with a different theme?

packages/components/src/Textarea/Textarea.tsx
ti10le: That Label and textarea are not in the same line. Otherwise Label is not above the area

packages/components/src/Textarea/Textarea.tsx
ti10le: no think not :)

packages/components/src/Textarea/Textarea.tsx
ti10le: But then wie have padding etc. hard coded in the component?

packages/components/src/Textarea/Textarea.stories.mdx
```diff
@@ -32,3 +39,27 @@ import { Textarea } from '@marigold/components';
     <Textarea placeholder="Placeholder disabled" disabled />
   </Story>
 </Canvas>
+
+### labeled & required
+
+<Canvas>
+  <Story name="Textarea label">
+    <Textarea label="Labeled" htmlfor="label" />
```
viktoria-schwarz, line 47: du hast hier überall aus Versehen `htmlfor` statt `htmlFor` geschrieben :)

viktoria-schwarz (changes requested):

packages/components/src/Textarea/Textarea.tsx
viktoria-schwarz: When I test it, it works  without (at least in the outer Box)

packages/components/src/Textarea/Textarea.tsx
ti10le: @viktoria-schwarz what do you mean here?

packages/components/src/Textarea/Textarea.tsx
ti10le: ok yes the outter box doesnt need it

viktoria-schwarz (approved):

PR #838, opened 2021-03-04
chore: init marigold docs with gatsby

viktoria-schwarz:
first merge #766 
https://marigold-ui.github.io/marigold/


docs/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "gatsby-plugin-sharp": "^3.0.0",
+    "gatsby-plugin-sitemap": "^3.0.0",
+    "gatsby-source-filesystem": "^3.0.0",
+    "gatsby-transformer-sharp": "^3.0.0",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1"
+  },
+  "devDependencies": {
+    "gh-pages": "^3.1.0",
+    "react-helmet": "6.1.0"
```
sebald, line 37: Guess this should be a dependency?

package.json
```diff
@@ -6,6 +6,7 @@
   "license": "MIT",
   "workspaces": [
     "config/*",
+    "docs/*",
```
sebald, line 9: Since the package is right inside "docs" we have to remove the glob part (/*)

sebald (changes requested):
We later switch to a theme-docs, right? 🙂

sebald (approved):

sebald (approved):

sebald (approved):

PR #861, opened 2021-03-08
refa(components): Remove system helper from <Text>

sebald:


ti10le:
@sebald you should also edit stories.mdx if you change varaint and as prop, right?

viktoria-schwarz:
@ti10le I guess if we move to our new docs page we will rewrite the docs anyways, so the "old" storybook stories don't really matter now

viktoria-schwarz (approved):

Issue #946, opened 2021-04-19
docs: Documentation for theme structure


viktoria-schwarz:
## Description

We want to document our theme structure so that it is more clear what values we demand for and why.

## Context

A global and complete documentation that explains all our design and structural decisions.

## Consequences

Our theme is well-documented.

viktoria-schwarz:
For our new Gatsby docs site, document

1) the general theme structure:
* how it is structured (and why)
* what values need to be there for a new theme
* how we intend to use the values
* explain semantic colors

2) each theme individually:
* when to use this theme
* what makes it a unique theme

sebald:
What about "how to use a theme?"! 😄

viktoria-schwarz:
Well yes that would be a good start :P
Though I expect that we will have a "Get Started" where all this is explained aswell

viktoria-schwarz:
Considering the [discussion](https://github.com/marigold-ui/marigold/discussions/998#discussion-3348533) about the details of the docs outline, this issue covers the **Theming** chapter.

Issue #925, opened 2021-04-12
Create infrastructure and template for component pages


ti10le:
We need a structure to create all component markdown pages. Before we add all mdx files from storybook to gatsby we have to add one functional and good looking example.

**Structure**:
- Headline
- Component Description
	- Typical component description (written by UX/UI)
	- Component Description what users/developers need to know about the Usage (written by Marigold developers)
- Props table
- Import
- Usage Examples

**Todos** for mdx in gatsby:
- [x] enable code highlighting with \`...`
- [x] enable code block highlighting with \```tsx or ```jsx
- [x] enable own table styling in gatsby markdown

viktoria-schwarz:
In the headline you mention to create a template, is that intentional? Or should we rather just start with one component and the outline you gave and then adjust this to each individual component. I don't think working with a real template saves us time....

Also I don't fully understand your point 
> enable own table styling in gatsby markdown

Can you explain what you mean by that?

ti10le:
> Or should we rather just start with one component and the outline you gave and then adjust this to each individual component. 

This is what I mean

> Can you explain what you mean by that?

It means that actually you got a non-bordered table with no padding, etc. if you create a table in gatsby. You need to add a plugin or s.e. to show a nice table view

ti10le:
Template could be a few things. It's not 100% clear. 
What I mean: Create something like a frame/base structure to show how to add our existing stories pages to Gatsby. Not sure if its just adding a few plugins and copy/paste some parts or a Gatsby template for the whole page. This is what we have to find out 🤷🏽‍♂️ 
Difficult to explain in a good way.. maybe talk about it on monday

viktoria-schwarz:
We conclude we want all component pages in our documentation with the same structure (the one given in this issue).
We need to look for suitable plugins to implement the table, code blocks and code highlighting.

sebald:
What we should do in this cycle for this umbrella issue:

- [ ] Dynamic navigation based on content (#1056 is some groundwork for that)
- [ ] Unify the gatsby configs for SSR and browser. We "wraps" need to be in both! Also, we should not have all the MDX components setup there.

ti10le:
actually we have:

text highlighting
heading highlighting
table highlighting
inline code highlighting with: `
code block highlighting: ```tsx
live code block highlighting: ```tsx live
only code highlighting without preview: ```tsx code

ti10le:
Two open points in #1111 and #1112

PR #1015, opened 2021-05-10
docs: Theming

viktoria-schwarz:
#946


docs/src/pages/designTokens/theme.mdx
```diff
@@ -0,0 +1,148 @@
...
+};
+```
+
+### Access Theme Variables
+
+Theme variables are styles that are stored in key value pairs and defined as described above.
+If using a different theme, the value for a certain variable can change because a different style is stored in the new theme.
+Reuse theme variables within the theme while defining variants of a component. Like this, you create a unique theme that is recognized in all the components:
+
+```tsx
```
ti10le, line 67: Should be ```tsx code-olny
like we discussed it 🙂

ti10le (dismissed):

ti10le (approved):

docs/src/components/Navigation.tsx
```diff
@@ -4,72 +4,65 @@ import { Text } from '@marigold/components';
...
-              <Text>Logos</Text>
+              <Link to="/foundation/theming">Theming</Link>
             </li>
+            <li>Layout</li>
+            <li>Box Primitive</li>
             <li>
-              <Text>Guidelines</Text>
+              <Link to="/foundation/icons">Iconography</Link>
             </li>
+            {/* <li>
```
ti10le, line 28: We should make the colors depending on a theme, right?
I think it's better to delete instead of comment until we have it in the right way. Do you agree?

ti10le, line 28: We should make the colors depending on a theme, right?
I think it's better to delete instead of comment until we have it in the right way. Do you agree?

ti10le (approved):

docs/src/components/Layout.tsx
```diff
@@ -20,10 +20,10 @@ export const Layout: React.FC = ({ children }) => {
       </Helmet>
       <ThemeProvider theme={b2bTheme}>
         <Columns space={4}>
-          <Column width={2}>
+          <Column width={4}>
```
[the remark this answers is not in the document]
sebald: Also not related to the related issue :D

sebald (dismissed):

docs/src/components/Layout.tsx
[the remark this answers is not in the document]
ti10le: What do you mean?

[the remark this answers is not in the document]
sebald: Well, changing the Layout of all pages when adding documentation for theming is weird :)

docs/src/pages/foundation/theming.mdx
```diff
@@ -0,0 +1,213 @@
...
+- `lineHeights`
+
+### Variants
+
+Marigold components extract styles from the theme induced by a ThemeProvider. You can define theme styles in scopes of variants for the components
+you intend to use. By creating variant objects on the same level with the component prefix on the top level, you can access its values by using the
+variant name in the component. If you want the same base styles applied to all the variants of one component, use a root object which you define outside the
+theme. See an example:
+
+```tsx code-only
```
ti10le, line 56: API changed to: ```tsx code
see #925 for usage of codeBlock 🙂

ti10le (dismissed):

docs/src/components/Layout.tsx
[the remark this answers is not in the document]
viktoria-schwarz: I actually forgot to put it back before pushing (I found 4/8 to be a lot better on wide AND small screens, but I don't mind)

docs/src/components/Navigation.tsx
viktoria-schwarz: Yes I didn't want to delete it so that we don't forget that there already exists a nice page about colors, but yes it should not appear in the menu like this

ti10le (approved):

PR #1056, opened 2021-05-22
feat: move content pages out of source

sebald:
- move content (mdx files) out the source folder
- automatically apply default layout to content and other pages (removes the need to import <Layout/> manually)

Ref #925


ti10le (approved):

ti10le:
So much better 😍🙂

Issue #1115, opened 2021-06-15
Demistify GH Actions


sebald:
## Description

Currently we have the following actions.

- Coverage / Coverage (pull_request)
- Coverage / Coverage (push)
- Validate / Lint, typecheck & test (push)

In every actions the test suit runs and coverage reports are created. This is unnecessary overhead. Plus, The "Validate" actions includes three different types of checks. They should be separated so that we have an easier time identifying what's working and what not.

We have to clean up the tasks and have one action for each type of check: 

- Is the software working? → test+coverage+coverall
- Does the code adhere to our formatting rules? → lint+prettier
- Are there any type issues? → typecheck

## Context

The actions will run longer and longer with the progression of the project. Uncool!

## Consequences

Hopefully we'll have clearer and more informative GH Actions and they finish faster.

viktoria-schwarz:
The new actions could look like this:

- Typecheck --> `yarn typecheck` [on push]
- Format --> `yarn lint` and `yarn format` [on push]
- Test --> `yarn test` and `yarn coverage` [on PR and push]

Also, we can use cache for the yarn deps to save time.

sebald:
Yarn format doesn't check, but actually runs prettier. Sorry this even work? 😳

Yarn coverage also runs the tests if I am not mistaken.

PR #1129, opened 2021-06-21
chore(gh-actions): add new, sorted actions with caching

viktoria-schwarz:
#1115


.github/workflows/format.yml
```diff
@@ -0,0 +1,35 @@
...
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install Dependencies
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn
+
+      - name: Lint
+        run: yarn lint
+
+      - name: Format
```
sebald, line 34: This is a manual tasks and it is also run by husky as a commit hook ;)

.github/workflows/format.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: actions/checkout@master
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js 12.x
+        uses: actions/setup-node@master
+        with:
+          node-version: 12.x
+          
+      - uses: actions/cache@v2
```
sebald, line 22: 🔥

.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Install Dependencies
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn
+
+      - name: Test
+        run: yarn test --ci
+        env:
+          COVERALLS_REPO_TOKEN: '${{ secrets.COVERALLS_REPO_TOKEN }}'
+
+      - name: Run coverage (with lcov file as output)
```
sebald, line 36: maybe we should rename this script? Because it is not only collecting coverage but also runs the tests. I think we do not need the previous step, right? 🤔

sebald (changes requested):
Had some comments, but using the cache is sweeeeeet! How much faster is it!? :-O

.github/workflows/format.yml
viktoria-schwarz: Why was it then that when I ran `yarn format` it changed more than 20 files that haven't been touched in quite a while? :thinking:

sebald (approved):

PR #1132, opened 2021-06-22
feat(components): Make `<Text>` and `<Link>` polymorphic

sebald:
- `<Text>`
  - the `as` prop supports arbitrary inputs
  - supporst ref
  - supports style props (text-align, color, cursor, user-select)

- `<Link>`
  - the `as` prop supports arbitrary inputs
  - does not support `ref`!
  - improved accessibility (react-aria)


packages/components/src/Text/Text.tsx
```diff
@@ -1,35 +1,49 @@
...
+export const Text: PolymorphicComponentWithRef<TextOwnProps, 'span'> =
+  forwardRef(
+    (
+      {
+        as = 'span',
+        variant = 'body',
+        children,
+        className,
+        align,
+        color,
```
ti10le, line 29: You have to change the Text example in docs from textColor to color :)

ti10le (approved):

ti10le (approved):

Issue #1034, opened 2021-05-17
docs: create switch theme functionality to show components in their theme


ti10le:
## Description

We need a functionality to show all components with all themes. That means that you can switch between our existing theme eg b2b, unicorn, ...

## Context

Better User Experience at using Marigold Docs. Users could switch between all themes. Makes easier usage for them.

## Use Cases 

I (developer) need a component from marigold. I can show the comp in my theme and not in an other theme.

## Consequences

Better user experience for all users of marigold.

viktoria-schwarz:
Does it mean that for each page of the documentation site we can set a default theme and have the option on the site to switch to any other theme?

ti10le:
It would be the best I think. Then we have as default our Marigold docs theme and you can show all parts of Marigold (which depends on a theme) in the theme you need. 
What do you mean @sebald ?

sebald:
I would not use the docs theme for the component examples. Rather use the B2B or b2c (when there is one) as default and make a global switcher.

ti10le:
I did not mean it that way. I would not take the docs theme for the components, too.

PR #1133, opened 2021-06-22
feat(docs): switch themes functionality

ti10le:
#1034 

- initial change to marigold-theme if we have one
- first version, please review and I will make it better :)


docs/src/components/ThemeSelect.tsx
```diff
@@ -0,0 +1,43 @@
...
+import b2bTheme from '@marigold/theme-b2b';
+
+export const themes = {
+  b2bTheme,
+  unicornTheme,
+};
+
+const themeSelectStyles = {
+  position: 'absolute',
+  right: b2bTheme.space.none,
```
sebald, line 14: That's not how this works :-O Use the scaling values! :P

docs/src/components/ThemeSelect.tsx
```diff
@@ -0,0 +1,43 @@
...
+  cursor: 'pointer',
+  fontSize: b2bTheme.fontSizes.xxsmall,
+  fontFamily: b2bTheme.fonts.body,
+};
+
+export const ThemeSelect: React.FC = () => {
+  const { current, themes, setTheme } = useThemeSwitch();
+
+  return (
+    <select
```
sebald, line 30: Why are you not using Marigold's `<Select>` component?

docs/src/components/ThemeSelect.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+import { useThemeSwitch } from './ThemeSwitch';
+import { useStyles } from '@marigold/system';
+import unicornTheme from '@marigold/theme-unicorn';
+import b2bTheme from '@marigold/theme-b2b';
+
+export const themes = {
```
sebald, line 7: You are using the same var name that is returned from the hook.

docs/src/components/CodeBlock.tsx
```diff
@@ -71,7 +72,9 @@ export const CodeBlock: React.FC<CodeBlockProps> = ({
                     <LivePreview />
                   </LiveProvider>
                 </div>
-                <ShowHideButton hide={hide} onHideChange={setHide} />
+                <ThemeProvider theme={b2bTheme}>
```
sebald, line 75: Shouldn't the switch theme stuff be here?

sebald (changes requested):

docs/src/components/CodeBlock.tsx
ti10le: Then you have the problem that you have to change the theme for every component example separately. Copy/show button should have the initial theme

docs/src/components/ThemeSelect.tsx
ti10le: Our Select doesn't like it should. We have to rewrite some things. Is this e separate issue our should I do this here?

docs/src/components/ThemeSelect.tsx
ti10le: I am using normal css values until we have a marigold-theme.

docs/src/components/ThemeSelect.tsx
sebald: That's why we are dogfooding! I've done the same thing #1132 ... I wanted to use our `<Link>` component but it didn't work. That's our chance to fix this!

docs/src/components/CodeBlock.tsx
sebald: Not quite sure I follow. You can add the switch globally and then pass it to the `<ThemeProvider>` inside the `<Codeblock!>` :)

docs/src/components/ThemeSwitch.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const useThemeSwitch = () => useContext(Context);
+
+// Component
+// ---------------
+export type MarigoldThemeSwitchProps = {
+  themes: { [name: string]: BaseTheme };
+  initial?: string;
+};
+
+export const MarigoldThemeSwitch: React.FC<MarigoldThemeSwitchProps> = ({
```
sebald, line 29: Separate the switch from the `<ThemeProvider>`.

docs/src/components/ThemeSelect.tsx
ti10le: New issue/PR to rewrite the Select component

sebald:
Great work!

sebald (approved):

PR #1134, opened 2021-06-22
feat(docs): add landing page and link in the nav

### LINKED ISSUE #1096: Have a starting page in docs
## Description

On "/" there is no file so far (i.e. on entering the documentation site), and in order to see `start.mdx` you need to redirect to `/start`. It would be nice to have a proper index page.

## Context

404 page when you enter the docs page

## Consequences

Create index page or redirect to start.mdx, no matter what but we have a proper landing page!

viktoria-schwarz:
close #1096

### DIFF
--- a/docs/src/pages/index.tsx
+++ b/docs/src/pages/index.tsx
@@ -1,0 +1,13 @@
+import React from 'react';
+import { Text, Heading } from '@marigold/components';
+import { ThemeProvider } from '@marigold/system';
+import b2bTheme from '@marigold/theme-b2b';
+
+export default function LandingPage() {
+  return (
+    <ThemeProvider theme={b2bTheme}>
+      <Heading variant="h1">Welcome to Marigold</Heading>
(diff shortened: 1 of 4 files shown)


docs/src/pages/index.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
```
sebald, line 1: You can actually just create an `index.mdx` in the content folder and it will work! :) No need for "start" or something.

sebald: or ... rename `start.mdx` to `index.mdx` for that matter 🙈

viktoria-schwarz: huh? An `index.mdx` or `start.mdx` is exactly what doesn't work with empty routes, and I thought if we stick to a simple landing page we can just create an `index.tsx` in `src/pages`?

sebald (approved):

PR #1137, opened 2021-06-22
feat(docs): Inital theme for docs

sebald:
- removed hardcoded dependency to b2b theme
- removed all usage of `styles`


sebald:
@ti10le @viktoria-schwarz not perfect, but I couldn't stand the hardwired usage of the b2b theme 😝

docs/src/theme/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    background: colors.gray[80],
+    primary: colors.brand.red,
+    secondary: colors.brand.orange,
+    disabled: colors.gray[70],
+    error: colors.red,
+    warning: colors.yellow,
+    info: colors.blue,
+    success: colors.green,
+  },
+  radii: {
```
ti10le, line 29: do you really mean radii?

docs/src/wrapper.tsx
```diff
@@ -16,7 +16,17 @@ export const WrapPageElement: GatsbyBrowser['wrapPageElement'] = ({
 export const WrapRootElement: GatsbyBrowser['wrapRootElement'] = ({
   element,
 }) => (
-  <ThemeProvider theme={b2bTheme}>
```
ti10le, line 19: After this removal our components doesn't have the b2b-theme style like before. Do you did this in regard of my switch-theme functionality?

docs/src/wrapper.tsx
```diff
@@ -16,7 +16,17 @@ export const WrapPageElement: GatsbyBrowser['wrapPageElement'] = ({
...
+  <ThemeProvider theme={theme}>
     <MDXProvider components={mdxComponents}>{element}</MDXProvider>;
   </ThemeProvider>
 );
+
+/**
+ * Enforce reloading to update styles.
```
ti10le, line 25: Can you tell me why we need this?

ti10le (approved):

ti10le (changes requested):
change from approve to request changes

docs/src/theme/index.ts
sebald: Yes, see https://theme-ui.com/theme-spec

viktoria-schwarz: plural of radius right?

docs/src/wrapper.tsx
sebald: No, I just switched to the docs theme :)

sebald: The theme from our docs app shouldn't leak into the examples in the first place.

docs/src/wrapper.tsx
sebald: Webpack HMR does not recognize changes in the styles (e.g. when you update a theme and this causes changes in the CSS styles).

This makes sure the page automatically refreshes, so we don't have to do it manually.

docs/src/theme/index.ts
sebald: I guess so? I just read the spec 😂

docs/src/wrapper.tsx
ti10le: Thanks :)

ti10le (dismissed):

viktoria-schwarz:
@sebald changeset is missing :)

sebald:
@viktoria-schwarz done! :P

ti10le (approved):

PR #1147, opened 2021-06-29
feat(components): improve <Select> with react-aria

### LINKED ISSUE #1146: Improve <Select> and use react-aria
## Description

Our Select component doesn't work like it should. I would use it in #1034 / #1133. This is a good time to make the Select component polymorphic and improve the accessibility (react-aria) like in #1132.

## Context

We want to have a Select component which works correctly and is complete accessible.

## Consequences

Easier and better usage of Select component. Good accesibility!

ti10le:
closes #1146


packages/components/src/Select/Select.tsx
```diff
@@ -1,30 +1,185 @@
...
+import { Box, BoxOwnProps } from '../Box';
+import { Text } from '../Text';
+
+export type SelectOwnProps = {
+  variant?: string;
+  label?: string;
+  disabled?: boolean;
+} & BoxOwnProps &
+  SelectProps<any>;
+export type SelectPolymorphicProps = PolymorphicProps<SelectOwnProps, 'select'>;
```
sebald, line 25: Why is this component polymorphic?

packages/components/src/Select/Select.tsx
```diff
@@ -1,30 +1,185 @@
...
+        <Text>{label}</Text>
       </Box>
-      <ArrowDown className={iconStyles} />
+      <HiddenSelect
+        state={state}
+        triggerRef={ref}
+        label={label}
+        name={props.name}
+      />
+      <button
```
sebald, line 59: Why not Marigold's button?

packages/components/src/Select/Select.tsx
```diff
@@ -1,30 +1,185 @@
...
   );
+}) as PolymorphicComponent<SelectOwnProps, 'select'>;
+
+// ListBox
+type ListBoxPopupProps = {
+  state: SelectState<any>;
+  otherProps?: React.HTMLAttributes<HTMLElement>;
+};
+
+const ListBoxPopup = ({ state, ...otherProps }: ListBoxPopupProps) => {
```
sebald, line 92: Maybe put that in a separate file for better overview? :)

packages/components/src/Select/Select.tsx
ti10le: First I have to learn more about Polymorphic 😬

packages/components/src/Select/Select.tsx
ti10le: Yes good Idea. But where? It is a good place between the other "normal" components?

sebald: If it's only usable die the select, or it in the select folder and don't expose it

packages/components/src/Select/ListBox.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { useListBox } from 'react-aria';
+import type { AriaListBoxOptions } from '@react-aria/listbox';
+import type { ListState } from 'react-stately';
+
+import { Box } from '../Box';
+import { Option } from './Option';
+import { useStyles } from '@marigold/system';
+
+interface ListBoxProps extends AriaListBoxOptions<unknown> {
+  listBoxRef?: React.RefObject<HTMLUListElement>;
```
sebald, line 11: Why not just `ref`?

packages/components/src/Select/ListBox.tsx
```diff
@@ -0,0 +1,44 @@
...
+  state: ListState<unknown>;
+}
+
+export const ListBox = (props: ListBoxProps) => {
+  const ref = useRef<HTMLUListElement>(null);
+  const { listBoxRef = ref, state } = props;
```
sebald, line 17: Can you explain what is happening here? Why do you override it?

packages/components/src/Select/ListBox.tsx
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  return (
+    <Box
+      as="ul"
+      {...listBoxProps}
+      className={selectBoxClassName}
+      ref={listBoxRef}
+    >
+      {[...state.collection].map((item: any) => (
```
sebald, line 39: That is super weird. Why do we have to do that?

sebald: If you're using any, please at least write a comment why or what the item is.

packages/components/src/Select/Option.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useRef } from 'react';
```
sebald, line 1: Again. More specific imports!

packages/components/src/Select/Option.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const ref = useRef<HTMLLIElement>(null);
+  const { optionProps, isSelected, isFocused } = useOption(
+    {
+      key: item.key,
+    },
+    state,
+    ref
+  );
+  const selectedClassName = useStyles({
+    css: {
```
sebald, line 24: Don't style stuff. Only layouting. Especially do *not* hardcode colors or spacing!

packages/components/src/Select/Option.tsx
```diff
@@ -0,0 +1,47 @@
...
+      cursor: 'pointer',
+      color: isSelected ? '#ffffff' : '#4b4b4b',
+      bg: isSelected ? '#3ab3d5' : isFocused ? '#c1f0fc' : 'white',
+    },
+  });
+
+  return (
+    <Box
+      as="li"
+      {...mergeProps(optionProps)}
```
sebald, line 40: You don't merge props. You only pass one argument.

packages/components/src/Select/Popover.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Box } from '../Box';
+
+interface PopoverProps {
+  popoverRef?: React.RefObject<HTMLDivElement>;
+  children: React.ReactNode;
+  isOpen?: boolean;
+  onClose: () => void;
+}
+
+export function Popover(props: PopoverProps) {
```
sebald, line 13: `function` -> `const`

packages/components/src/Select/Popover.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import { useOverlay, DismissButton, FocusScope } from 'react-aria';
+
+import { Box } from '../Box';
+
+interface PopoverProps {
+  popoverRef?: React.RefObject<HTMLDivElement>;
+  children: React.ReactNode;
```
sebald, line 8: If you use a `React.FC` for the component you do not need to add this here.

packages/components/src/Select/Select.test.tsx
```diff
@@ -1,67 +1,195 @@
...
       fontFamily: 'Inter',
     },
-    other: {
-      fontFamily: 'Oswald',
-    },
   },
 };
 
-test('supports default variant and themeSection', () => {
+test('supports default variant', () => {
```
sebald, line 16: It doesn't have a default variant anymore, right?

packages/components/src/Select/Select.test.tsx
```diff
@@ -1,67 +1,195 @@
...
+      <Select label="MyLabel">
+        <Item>1</Item>
+      </Select>
+    </ThemeProvider>
+  );
+  const selectLabel = screen.getAllByText(/MyLabel/);
+  expect(selectLabel[0]).toHaveAttribute('for');
+});
+
+test('supports disabled prop', () => {
```
sebald, line 57: Maybe test if it does not respond to any interaction? So it really is disabled?

packages/components/src/Select/Select.test.tsx
```diff
@@ -1,67 +1,195 @@
...
       </Select>
     </ThemeProvider>
   );
-  const select = screen.getByTitle(/select/);
+  const select = screen.getByTestId('selectId');
+  expect(select).toHaveAttribute('disabled');
+});
 
-  expect(select).toHaveStyle(`font-family: Oswald`);
+test('supports placeholder prop', () => {
```
sebald, line 69: What does the test do? The prop doesn't matter -> for the user: "You can set a placeholder"

packages/components/src/Select/Select.test.tsx
```diff
@@ -1,67 +1,195 @@
...
   );
-  const select = screen.getByTitle(/select/);
+  const button = screen.getByTestId('selectId');
+  expect(button).toHaveAttribute('aria-expanded', 'false');
 
-  expect(select instanceof HTMLSelectElement).toBeTruthy();
+  fireEvent.click(button);
+
+  // more than one item found because of the HiddenSelect component
+  const items = screen.getAllByText(/Red/);
```
sebald, line 95: Select by aria role? Does this work?

packages/components/src/Select/Select.test.tsx
```diff
@@ -1,67 +1,195 @@
...
+  expect(button).toHaveAttribute('aria-expanded', 'true');
+  const items = screen.getAllByText(/Red/);
+  expect(items[1]).toBeVisible();
+
+  fireEvent.click(items[1]);
+  expect(button).toHaveTextContent('Red');
+  expect(items[1]).not.toBeVisible();
+});
+
+test('component has hidden Items from HiddenSelect component and displayed Items', () => {
```
sebald, line 148: What do you test here?

packages/components/src/Select/Select.tsx
```diff
@@ -1,30 +1,106 @@
...
+  const state = useSelectState(props);
+  const ref = useRef(null);
+  const { labelProps, triggerProps, valueProps, menuProps } = useSelect(
+    props,
+    state,
+    ref
+  );
+  const { buttonProps } = useButton(triggerProps, ref);
+  const { focusProps } = useFocusRing();
+  const selectColor = '#4B4B4B';
```
sebald, line 43: Hardcoded -> use theme!

packages/components/src/Select/Select.tsx
```diff
@@ -1,30 +1,106 @@
...
+  const labelClassName = useStyles({
```
sebald, line 44: Maybe you can make use of `currentColor` property?

packages/components/src/Select/Select.tsx
```diff
@@ -1,30 +1,106 @@
...
+    css: { color: disabled ? 'disabled' : selectColor },
+  });
+  const iconClassName = useStyles({
+    css: { color: disabled ? 'disabled' : selectColor, pl: '5px' },
```
sebald, line 48: Hardcoded padding

packages/components/src/Select/Select.tsx
```diff
@@ -1,30 +1,106 @@
...
+  });
+  const buttonTextClassName = useStyles({
     css: {
-      alignSelf: 'center',
-      pointerEvents: 'none',
-      ml: '-28px',
+      fontFamily: 'body',
+      fontSize: 'xsmall',
+      fontWeight: 400,
+      lineHeight: '32px',
```
sebald, line 55: hardcoded value

sebald (approved):
More storybook examples would be nice (e.g. label and no label, disabled, ...)

sebald (changes requested):
wrong button

packages/components/package.json
```diff
@@ -8,8 +8,11 @@
   "license": "MIT",
   "dependencies": {
     "@marigold/system": "workspace:*",
-    "@react-aria/link": "^3.1.3",
-    "react-keyed-flatten-children": "1.3.0"
+    "@react-types/shared": "^3.7.1",
```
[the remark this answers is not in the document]
ti10le: in Option.tsx

packages/components/src/Select/ListBox.tsx
ti10le: I can make it better. Little bit to much copy/paste :)

packages/components/src/Select/Select.test.tsx
ti10le: I thought we test this that nobody can remove the placeholder prop from our component. That custom placeholder works.

packages/components/src/Select/Select.test.tsx
ti10le: That HiddenSelect works and the displayed items visible. This test is not absolutely necessary

ti10le:
> wrong button

do this if the aria button is merged

docs/package.json
```diff
@@ -12,6 +12,7 @@
     "@marigold/types": "workspace:*",
     "@mdx-js/mdx": "1.6.22",
     "@mdx-js/react": "1.6.22",
+    "@react-stately/collections": "^3.3.2",
```
sebald, line 15: Remove it?! :)

packages/components/src/Select/Option.tsx
```diff
@@ -0,0 +1,34 @@
...
+    state,
+    ref
+  );
+
+  return (
+    <Box
+      as="li"
+      {...optionProps}
+      ref={ref}
+      p="2px 5px"
```
sebald, line 28: :-0

packages/components/src/Select/Option.tsx
```diff
@@ -0,0 +1,34 @@
...
+      variant={isSelected ? 'select.option.selected' : 'select.option.default'}
```
sebald, line 29: Should we make use of theme-ui's `__default` here?

ti10le: Niiiice, it works. Will try it on other parts of the Select, too.

sebald:
Bug 🥲 
image

packages/components/src/Select/Select.tsx
```diff
@@ -1,30 +1,94 @@
...
+            {props.label}
+          </Label>
+        </Box>
+      )}
+      <HiddenSelect
+        state={state}
+        triggerRef={ref}
+        label={props.label}
+        name={props.name}
+        isDisabled={true}
```
sebald, line 62: This should not be disabled all the time, right?

sebald (changes requested):
The issue remains the same. The select (or the popup to be specific) is shifting the layout, which it should not do. I changed the example to show what I mean:

image

This is what a user would expect:

image

PS: The select still jumps because of the focus ring or something else?

packages/components/src/Select/Select.tsx
```diff
@@ -1,30 +1,94 @@
...
+import { Popover } from './Popover';
 
-export const Select: React.FC<SelectProps> = ({
-  variant = 'default',
-  children,
+export type SelectProps = {
+  placeholder?: string;
+  disabled?: boolean;
+} & ComponentProps<'select'> &
+  BoxOwnProps &
```
sebald, line 22: Why? 🤔

ti10le:
> The issue remains the same. The select (or the popup to be specific) is shifting the layout, which it should not do. I changed the example to show what I mean:
> 
> image
> 
> This is what a user would expect:
> 
> image
> 
> PS: The select still jumps because of the focus ring or something else?

But in the second picture that the CodeBlock grow by the listbox, right?

sebald:
Don't quite understand what you mean 😐

ti10le:
> Don't quite understand what you mean 😐

Basically, the listbox is in the foreground above everything. (Actually like on your second picture) But if the select is opened in the codeBlock then the block should increase, right?

sebald:
No, why should it? 🤔 The listbox doesn't change the underlying layout.

ti10le:
> No, why should it? 🤔 The listbox doesn't change the underlying layout.

Ok, this was my opinion too but then I understood something wrong in our last chat. Now everything clear :)

docs/src/components/ThemeSelect.tsx
```diff
@@ -1,8 +1,8 @@
 import * as React from 'react';
 import { useThemeSwitch } from './ThemeSwitch';
-import { Box, SelectProps } from '@marigold/components';
+import { Box } from '@marigold/components';
 
-export const ThemeSelect: React.FC<SelectProps> = () => {
+export const ThemeSelect: React.FC = () => {
```
sebald, line 5: You can remove the React.FC since the component doesn't accept any children

packages/components/src/Select/Select.tsx
```diff
@@ -1,30 +1,125 @@
...
+            {props.label}
+          </Label>
+        </Box>
+      )}
+      <HiddenSelect
+        state={state}
+        triggerRef={triggerRef}
+        label={props.label}
+        name={props.name}
+        isDisabled={true}
```
sebald, line 86: Why is it disabled? shouldn't that be based on the disabled state of the whole component?

packages/system/src/MarigoldProvider.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
sebald, line 1: There is no test! :P

packages/components/src/Select/Select.tsx
ti10le: Yes I thought give the disabled prop ist better, too. But I got this error if theres the disabled prop and I don't know how to solve it.
image

sebald: Okay, but how are we or is react-aria using the regular select element? Why is it there if we have a complete custom select?

ti10le: I'm just trying to figure this out 🙈

sebald (approved):

PR #1149, opened 2021-07-01
feat(docs): Add dynamic navigation (based on mdx content)

sebald:
- Use `gatsby-image` for Logo
- Combine Gatsby's and Marigold's link component
- Create a dynamic nav based on files inside `docs/content`
- Add initial styling for nav
- Move theme switch provider into root wrapper


docs/src/components/Navigation/useNavigation.ts
```diff
@@ -0,0 +1,119 @@
+import { graphql, useStaticQuery } from 'gatsby';
+
```
ti10le, line 2: Very helpful comments for me :)

docs/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
ti10le, line 1: Components with a big letter like MenuItem and ValidationMessage are broken.
You make /menuItem in the nav link but the original is /menu-item

ti10le (changes requested):

docs/src/components/Navigation/Navigation.tsx
sebald: fixed ... freaking gatsby

docs/src/components/Navigation/useNavigation.ts
sebald: thx ^^

ti10le:
@sebald renaming to menuI-iem.mdx is not good, right?

sebald:
> @sebald renaming to menuI-iem.mdx is not good, right?

Well, it will work, but the url will look funny 😬

ti10le (approved):

PR #1163, opened 2021-07-09
feat(system): Expose helper to transpile style object

### LINKED ISSUE #1162: Option to get the "raw" CSS Values
## Description

Currently there is no way to get the transpiled CSS object (by theme-ui). The system helper all return classnames.

## Context

In order to fix <Stack> and <Columns> we need access to the padding value, since we have to calculate the whitespace.

## Consequences

A more open system lib.

sebald:
Closes #1162

### DIFF
--- a/packages/components/src/Stack/Stack.tsx
+++ b/packages/components/src/Stack/Stack.tsx
@@ -10,21 +10,32 @@ export type StackProps = {
   title?: string; // For Testing
 };
 
+const ALIGN_STYLES = {
+  left: {
+    display: 'block',
+    flexDirection: 'row',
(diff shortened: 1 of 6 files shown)


packages/components/src/Stack/Stack.tsx
```diff
@@ -10,21 +10,32 @@ export type StackProps = {
   title?: string; // For Testing
 };
```
sebald, line 12: Ups ... this shouldn't be here. But fine I guess?

ti10le: Ok for me

ti10le (approved):

PR #1188, opened 2021-07-21
Make <Button> handle icons the right way

### LINKED ISSUE #297: Design & Component update - Button with icon
We already have button with icon component. Issue to this: see #146 

[Button Figma styles](https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=428%3A0)

**Problem** - two things are missing right now:
- we have to find a solution for the small button default size (at the moment 24px icon size). Default icon size is `24px`. Small button icon should have `20px`.
- If a button has an icon, padding-left should be: `button.small: 12px` instead of 16 and `button.large: 24px` instead of 32.

ti10le:
closes #297 

Note: For Select component we need a button with an icon at the end. Maybe we can do something like material-ui. They have a `startIcon` and `endIcon` prop.


packages/components/src/Button/Button.tsx
```diff
@@ -31,18 +33,28 @@ export const Button: PolymorphicComponentWithRef<BoxOwnProps, 'button'> =
...
           {...props}
           as={as}
-          variant={[`button.${variant}`, `button.${size}`]}
+          variant={[`button.${variant}`, buttonSize]}
           className={className}
           ref={ref}
         >
           <Box as="span" display="inline-flex" alignItems="center">
-            {children}
+            {icon ? (
```
sebald, line 48: Why do we want to our default button to handle icons?

packages/components/src/Button/Button.tsx
```diff
@@ -31,18 +33,28 @@ export const Button: PolymorphicComponentWithRef<BoxOwnProps, 'button'> =
...
+              <>
+                {icon}
+                <Box as="span" pl={size === 'large' ? 'xsmall' : 'xxsmall'}>
```
sebald, line 51: This is not good, since we don't know what values the theme uses for `xmsall` and so on. Could look weird.

sebald (approved):
Not sure about this PR 🤔

sebald (changes requested):
I don't understand why this is necessary. Do we want to specifically handle icons? Aren't icons just another inline element?

ti10le:
I thought because of the figma design: https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=428%3A0

Our button can't handle exactly this. But yes, maybe we dont have to show this like exactly in the figma design.

sebald:
Well I guess we should sometimes question our designers. I really don't see why we need to support non even paddings 😳

ti10le:
I will close this PR. Designers should check the icon buttons design.

PR #1190, opened 2021-07-21
fix(system): List reset should not remove the enumeration

sebald:


packages/system/src/reset.ts
```diff
@@ -18,7 +18,7 @@ const block = css({
 });
 
 const list = css({
-  listStyle: 'none',
+  // empty
```
ti10le, line 21: would you not rather leave it without comment?

ti10le (approved):

packages/system/src/reset.ts
sebald: 🤷‍♂️

Issue #926, opened 2021-04-12
Storybook: New Structure with more focus on development


viktoria-schwarz:
## Description
Until now we used Storybook as a canvas for our development and as a documentation page for the components.

As we move the documentation to our Gatsby site, we want to exclusively focus on the development part and our components in Storybook.

We want to see all variants at once (per component), the source code, functionalities and prop table.

## Problems
* Storybook Tabs as we have them now are not convenient for a good overview of all variants and combinations of a component
* We need new plugins that allow us to see the source code (usage), the functionalities of a component, the prop table with little code-wise effort
* focus on the components in storybook and don't mix it with documentation!

## Todos

- [x] instead of "Canvas" and "Docs" as tabs, use a Default in storybook and the Variants as subpages of each component, so that on click of a component name in the sidebar navigation menu, the component shows up with *all* its variants
- [x] we can use [Knobs addon](https://storybook.js.org/addons/@storybook/addon-knobs/) to edit/test props and values dynamically
- [x] maybe use Doc [blocks addon](https://storybook.js.org/docs/react/writing-docs/doc-blocks) to automatically create props table
- [x] delete "How to write components" section

## Consequences
We have a clean storybook with all our components that helps us testing and seeing our component library with different themes. Documentation is excluded from this.

viktoria-schwarz:
- Knobs addon is deprecated --> use [Controls addon](https://storybook.js.org/addons/@storybook/addon-controls)
- ArgsTable can be generated but values need to be specified in the Component.stories.mdx file 
- one or two Tabs??

viktoria-schwarz:
- added "Welcome" story with very short explanation and links
- added useStyles hook story (for completeness as it is important for developers)
- I would stick with 2 tabs, Canvas and Docs: Canvas shows a single component with addon tabs (Accessibility, Controls, Actions --> testing) and Docs shows the (self-generated) Args Table where you can play around with the input values

PR #1218, opened 2021-08-16
feat(docs): add controls and essentials

viktoria-schwarz:
#926


packages/components/src/Field/Field.stories.mdx
```diff
@@ -1,57 +1,64 @@
...
+    },
+    label: {
+      control: {
+        type: 'text',
+      },
+      type: { required: true },
+    },
+    error: {
+      control: {
+        type: 'boolean',
```
ti10le, line 31: Prop error is type string at the moment and changing some props here has no effect.

packages/components/src/Select/Select.stories.mdx
```diff
@@ -1,51 +1,69 @@
...
+      options: [true, false],
+      table: {
+        defaultValue: {
+          summary: false,
+        },
+      },
+    },
+    error: {
+      control: {
+        type: 'boolean',
```
ti10le, line 42: string because of the error message

ti10le (changes requested):

ti10le:
When you click show code in storybook you got <object object> if you used our <Text> component. Storybook doesnt have the import correctly.

viktoria-schwarz:
> When you click show code in storybook you got if you used our component. Storybook doesnt have the import correctly.

I think it happens each time a Polymorphic Component is used (storybook doesn't know early enough what to render, so it just outputs object object), but I would leave it for now and note it down for later

ti10le (approved):

PR #1227, opened 2021-08-20
bugfix: jumping border at Select button

### LINKED ISSUE #1225: Select component "jumps" if it is focused
## Description

Select component "jumps" if you have the blue focused border

## How to reproduce

1. Go to Select component doc
2. Click on Select button
3. See error

## Expected behavior

Shouldn't jump. Use outline instead of border.

ti10le:
closes #1225


themes/theme-b2b/src/index.ts
```diff
@@ -126,6 +127,7 @@ const button = {
       borderRight: '1px solid',
       borderLeft: '1px solid',
       borderBottom: 'none',
+      marginBottom: '1px',
```
sebald, line 130: This doesn't uses the theme 😐

ti10le: hm this should be just an hack to fix the problem that the Select jumps 1px if its open. Do you have a better solution? I know thats not part of our theme space values..

sebald: Is it a hack or a fix? 🤪

Do you know what causes the jumping?

ti10le: yes because of the border usage. But i cant use outline because in outline you cant set different borders at left right bottom or top. You only can set the same for all four.

sebald: Buuuut, you can make the border transparent instead of adding margins that also shift the layout! :)

ti10le: I hope this is a good solution 😬

sebald (approved):

PR #1228, opened 2021-08-20
docs: Use Select component in ThemeSelect

### LINKED ISSUE #1223: Use our own Select component in marigold docs ThemeSwitch
## Description

Use our own Select component in marigold docs ThemeSwitch instead of <Box as="select">

## Context

We want to use in marigold docs our own written components to test them all in production.

## Consequences

We get a product where we use all our written components.

ti10le:
closes #1223


packages/components/src/Select/Select.tsx
```diff
@@ -24,6 +24,7 @@ export type SelectProps = {
   disabled?: boolean;
   required?: boolean;
   error?: string;
+  width?: string;
```
viktoria-schwarz, line 27: I would like to restrict the type here more, like responsive style values?

docs/content/components/select.mdx
```diff
@@ -20,6 +20,7 @@ Dropdown for selecting an item among different options.
 | `disabledKeys` (optional)       | [Iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols)`<Key>` |         |
 | `required` (optional)           | `boolean`                                                                                                |         |
 | `error` (optional)              | `string`                                                                                                 |         |
+| `width` (optional)              | `string`                                                                                                 |         |
```
viktoria-schwarz, line 23: can you add the new prop to storybook aswell?

packages/components/src/Select/Select.tsx
ti10le: Yes I thought about this. Like this at menuWidth: https://react-spectrum.adobe.com/react-spectrum/Picker.html#props
but we have not yet, right? We could add it to the BaseTheme?

viktoria-schwarz: I think we have ResponsiveStyleValue which is used in other components as well :)

ti10le: oh sure, I did not think of that just now 🙈

viktoria-schwarz (approved):

Issue #1197, opened 2021-07-29
Add Root styles


sebald:
## Description

We currently don't have an option to add root styles (e.g. html, body). This is convenient for building layouts to set background and spacing for apps.

## Context

If you want to emulate this with the current state you have to wrap your app in an addition `div`. This shouldn't be necessary :-/

## Use Cases 

- Set `body` background color
- Set `body` margin/padding
- Set global font family

## Consequences

Option to "globally" set CSS.

viktoria-schwarz:
Where do you imagine these root styles to be? Should they be set once for all applications or somehow flexible and manageable each time you use a component?

sebald:
So here is the question: do the rest styles solve the mentioned use cases?

I imagine there global or root styles be set via the theme. So that we (for example) can set the font once in our docs or give the whole page a background color.

viktoria-schwarz:
Ahhh, you mean setting these values within the theme. Okay that makes sense :+1:

viktoria-schwarz:
So to make that work we could add a Component `<RootStyles />` with some rootStyles (given in the theme) and place it next to the children in the Provider?

sebald:
Exactly

sebald:
BTW hast checked theme-ui. They got a reference for you ☺️

https://github.com/system-ui/theme-ui/blob/develop/packages/theme-provider/src/index.tsx

In the file there is also a solution for the overlay issue (since root styles should only be applied once too).

viktoria-schwarz:
Yes, I've seen how they do it, but they use their jsx to create the styles, that's what I meant that I don't know how to inject the root object from the theme.

sebald:
> Yes, I've seen how they do it, but they use their jsx to create the styles, that's what I meant that I don't know how to inject the root object from the theme.

@viktoria-schwarz trick question: is that different from what we do? And if so, how? 

Also, how far are you in the react course 😬

viktoria-schwarz:
Not as far as I would like to be :)

And no it's not different, I could make it work (I think yesterday when I tried it I just used wrong imports), though now my tests still fail. I'm trying to make it work!

viktoria-schwarz:
closing via #1230

ti10le:
PR was reverted. Issue still in work.

sebald:
> PR was reverted. Issue still in work.

Is there a commit to reproduce the issue? And what's not working correctly?

ti10le:
waiting for #1317

PR #1230, opened 2021-08-23
feat: add GlobalStyles via theme

viktoria-schwarz:
#1197


viktoria-schwarz:
@sebald with a wrapping `<Box>` I could make it work. Or should I try to use emotion's `Global`? I tried to use it but couldn't access the theme (theme.styles.root) in a proper way... what do you think?

sebald:
I would use `<Global>` since it does some extra stuff and optimization.

But I works not expose anything of that and keep it private to the `<MarigoldProvider>`.

And one more thing: since we don't do the `styles` thing, this is rather a theme-ui thing, use put "root" on the root theme object? 😄

viktoria-schwarz:
In the `<body>` the styles work but not in the `<html>` when injected via the root in the theme. Not sure if I'm missing something :thinking:

packages/components/src/Provider/MarigoldProvider.tsx
```diff
@@ -1,6 +1,32 @@
...
-import { ThemeProvider, ThemeProviderProps } from '@marigold/system';
+import {
+  ThemeProvider,
+  ThemeProviderProps,
+} from '@marigold/system';
+import { Global, jsx } from '@emotion/react';
+import { css, Theme } from '@theme-ui/css';
+
+const GlobalStyles = () =>
+  jsx(Global, {
```
sebald, line 11: You don't need the helper here. Just use JSX! -> `<Global>`

packages/components/src/Provider/MarigoldProvider.tsx
```diff
@@ -1,6 +1,32 @@
...
+    styles: emotionTheme => {
+      const theme = emotionTheme as Theme;
+      const { useRootStyles } = theme.config || theme;
```
sebald, line 14: Do we need this?

packages/components/src/Provider/MarigoldProvider.tsx
```diff
@@ -1,6 +1,32 @@
...
+
+      if (useRootStyles === false || (theme.styles && !theme.styles.root)) {
```
sebald, line 16: Ehem 🤪

sebald:
@viktoria-schwarz I just took a quick look, but I think you copy/pasted something out shouldn't have 🤪

ti10le (approved):

PR #1231, opened 2021-08-24
bugfix: add OverlayProvider just once with the MarigoldProvider

### LINKED ISSUE #1195: OverlayProvider ist added multiple times with the MarigoldProvider
## Description

The `<MarigoldProvider>` always adds react-aria's OverlayProvider. Especially if you using cascading theming. There should only be a single instance of the provider at any time.

Otherwise this can lead to inaccessible overlays.

ti10le:
closes #1195


packages/components/src/Provider/MarigoldProvider.tsx
```diff
@@ -1,15 +1,33 @@
...
 
 // a merge of the ThemeProvider and the react-aria OverlayProvider
 export const MarigoldProvider: React.FC<ThemeProviderProps> = ({
   theme,
   children,
 }) => {
+  const outerTheme = useTheme();
+  const outerThemeJson = JSON.stringify(outerTheme, null, 2);
+  const defaultThemeValueJson = JSON.stringify(defaultThemeValue, null, 2);
+  const isTopLevel = outerThemeJson === defaultThemeValueJson;
```
sebald, line 26: Sorry, but that's not the proper way 🙈 Why can you check with the returned value from the theme hook?

sebald (changes requested):

packages/components/src/Provider/MarigoldProvider.tsx
```diff
@@ -17,12 +40,13 @@ export const MarigoldProvider: React.FC<ThemeProviderProps> = ({
...
+  const isTopLevel = outerTheme === defaultThemeValue
+
   return (
     <ThemeProvider theme={theme}>
-      <OverlayProvider>
-        <GlobalStyles />
-        {children}
-      </OverlayProvider>
+      {isTopLevel && <GlobalStyles />}
+      <OverlayProvider>{children}</OverlayProvider>
```
sebald, line 49: This provider should also be added only once!

sebald (changes requested):

packages/components/src/Provider/MarigoldProvider.tsx
ti10le: Hey @sebald I think now is the check correct but I am not sure. 
And we don't know how to check in jest if there are not more than one OverlayProvider. Do you have a tip for this?

sebald:
@ti10le It's fine not to have a sheet test for this. Just test if cascading marigold providers works :)

ti10le:
> @ti10le It's fine not to have a sheet test for this. Just test if cascading marigold providers works :)

Yes we already have this test, also we have the test: OverlayProvider is present
I think then its ready for final review :)

sebald:
There is still commented out code.

ti10le:
> There is still commented out code.

But this ist from the GlobalStyles PR not from this 🙈 We can also wait until the GlobalStyles is ready 🤷🏽‍♂️

sebald:
> > There is still commented out code.
> 
> But this ist from the GlobalStyles PR not from this 🙈 We can also wait until the GlobalStyles is ready 🤷🏽‍♂️

Maaaaaybe that works be fixed first then? 🤪

ti10le:
will be fixed in #1395

PR #1232, opened 2021-08-25
Field component redesign

### LINKED ISSUE #715: Design update - Field
Add variant with icon and differentiate between left and right position of it.
See [figma](https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=454%3A0)

ti10le:
#715


docs/content/components/field.mdx
```diff
@@ -7,17 +7,17 @@ title: Field
 ## Description
 
 With the Field component you can add a HTML `<label>` and `<input>` element to your form.
-The variant can be added with the variant prop. The default variant is `default`.
 
 ## Properties
 
-| Property             | Type                                                                                       | Default   |
-| :------------------- | :----------------------------------------------------------------------------------------- | :-------- |
-| `variant` (optional) | `string`                                                                                   | `default` |
```
viktoria-schwarz, line 16: Isn't it a bit odd that we don't have a variant here anymore?

ti10le: Yes it is. But how would you handle this? It's just an addition of Label + Input and they have their fitting variants. I can't think of a suitable situation for a field variant.

PR #1269, opened 2021-09-10
feature: section to select component

### LINKED ISSUE #1268: feature: Select component allows Sections around the items
## Description

New feature to the Select component. You can add different Sections to the listbox. Under the Sections are Items.

## Context

Users can add sections to select component. This would be necessary if marigold is in usage.

## Consequences

The result is a more complete component than before.

ti10le:
closes #1268

### DIFF
--- a/packages/components/src/Select/ListBoxSection.tsx
+++ b/packages/components/src/Select/ListBoxSection.tsx
@@ -1,0 +1,41 @@
+import React from 'react';
+import { useListBoxSection } from '@react-aria/listbox';
+import type { ListState } from '@react-stately/list';
+import { useStyles } from '@marigold/system';
+import type { Node } from '@react-types/shared';
+
+import { Box } from '../Box';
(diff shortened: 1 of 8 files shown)


packages/components/src/Select/ListBoxSection.tsx
```diff
@@ -0,0 +1,41 @@
...
+import type { ListState } from '@react-stately/list';
+import { useStyles } from '@marigold/system';
+import type { Node } from '@react-types/shared';
+
+import { Box } from '../Box';
+import { Option } from './Option';
+
+interface SectionProps {
+  section: Node<unknown>;
+  state: ListState<unknown>;
```
viktoria-schwarz, line 12: I guess the `<unknown>` type causes the problem in Storybook that we can't import and use `Item` as intended.

viktoria-schwarz (approved):

PR #1316, opened 2021-09-14
feat: Add and expose SSRProvider to support SSR

ti10le:


.changeset/early-berries-speak.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": patch
+"@marigold/components": patch
+---
+
+remove all warnings in gatsby build
```
sebald, line 6: This is not what this PR really does. It's not helpful.

packages/components/src/Provider/MarigoldProvider.tsx
```diff
@@ -19,11 +21,13 @@ export const MarigoldProvider: React.FC<ThemeProviderProps> = ({
   children,
 }) => {
   return (
-    <ThemeProvider theme={theme}>
-      <OverlayProvider>
-        {/* <GlobalStyles /> */}
-        {children}
-      </OverlayProvider>
-    </ThemeProvider>
+    <SSRProvider>
```
sebald, line 24: This shouldn't be added here all the time, because we don't know if people use SSR by default!

sebald (approved):

PR #1350, opened 2021-09-30
docs: foundation about css-in-js

ti10le:
#114


viktoria-schwarz:
Maybe move this subject to `Foundation` section? I don't really see a guide here (there are no steps to follow, just questions being answered). I am a bit unsure whether the article really covers a question (wouldn't the questions be "What is CSS-in-JS and why do we use it?" ?) or rather just talks about how we use it, because there's also SSR and className property handled and this is already covered in the foundation parts. Maybe I am just lacking some more structure :blush: 

And maybe use "CSS-in-JS" consistently :see_no_evil:

viktoria-schwarz (approved):

Issue #1244, opened 2021-08-31
Design & Component update - Dialog


ti10le:
## Description

Update our Dialog component that it's fully functional. Implement react-aria, overlay, popover like in the Select component.
https://react-spectrum.adobe.com/react-aria/useDialog.html

## Context

Have a full functional Dialog component wich passes react-aria rules.

## Consequences

easier usage of Dialog component, ready to use in production.
passes react-aria rules

PR #1365, opened 2021-10-05
feature: redesign dialog component

ti10le:
#1244

### DIFF
--- a/packages/components/src/Dialog/Dialog.tsx
+++ b/packages/components/src/Dialog/Dialog.tsx
@@ -1,31 +1,28 @@
-import React from 'react';
+import React, { LegacyRef, RefObject } from 'react';
 import { ComponentProps } from '@marigold/types';
-import {
-  OverlayTriggerState,
-  useOverlayTriggerState,
-} from '@react-stately/overlays';
+import { useOverlayTriggerState } from '@react-stately/overlays';
 import { OverlayContainer } from '@react-aria/overlays';
 import { useButton } from '@react-aria/button';
-import { LegacyRef, RefObject } from 'react-helmet/node_modules/@types/react';
 import { Close } from '@marigold/icons';
 
 import { Box } from '../Box';
 import { Button } from '../Button';
-import { Text } from '../Text';
+import { Heading } from '../Heading';
 
 import { ModalDialog } from './ModalDialog';
 
 export type DialogProps = {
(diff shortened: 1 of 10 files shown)


packages/components/src/Dialog/ModalDialog.tsx
```diff
@@ -0,0 +1,69 @@
...
+      display: 'flex',
+      alignItems: 'center',
+      justifyContent: 'center',
+    },
+  });
+  const modalBodyClassname = useStyles({
+    css: {
+      background: 'white',
+      minWidth: '510px',
+      minHeight: '240px',
```
viktoria-schwarz, line 50: why not move these styles as a variant to the theme? it seems wrong to have so many fix values hardcoded...

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -1,32 +1,84 @@
...
+            )}
+            {children}
+          </Box>
           <Box variant="dialog.onClose">
-            <Button variant="close" size="xsmall" onClick={onClose}>
+            <Button
+              variant="close"
+              size="xsmall"
+              {...closeButtonProps}
+              ref={closeButtonRef as LegacyRef<HTMLButtonElement>}
```
viktoria-schwarz, line 57: I don't really understand what `LegacyRef` is or does, can you explain that?

packages/components/src/Dialog/ModalDialog.tsx
viktoria-schwarz: I think we are more flexible when we have these styles in the theme :)

packages/components/src/Dialog/Dialog.tsx
ti10le: Oh I think I had needed that.... but is probably not necessary

viktoria-schwarz:
> Yes it's crazy.. can't see anything if I type yarn coverage but coveralls says -0.2 thinkingsee_no_evil

You see it if you click on the coveralls.io link (on the badge), it's in line 24 of Dialog.tsx :)

ti10le:
> > Yes it's crazy.. can't see anything if I type yarn coverage but coveralls says -0.2 thinkingsee_no_evil
> 
> You see it if you click on the coveralls.io link (on the badge), it's in line 24 of Dialog.tsx :)

Now we have 100% :)

Issue #1395, opened 2021-10-13
Re-introducing @emotion/react [Summary Issue]


sebald:
## Description

Last year we switched from `@emotion/react`(back then it was `@emotion/core`) to the framework agnostic `@emotion/css`. One of the main reasons for this was that we wanted to use `className` and hooks for styling instead of using the `css` prop.

Using the `css` prop came with some issue, like leaking the `css` prop to every single component, which was not our intention. Just to name one.

## Context

The reason we want to switch back is that we ran into multiple issues implementing global styles and the order in which styles are applied (see 
#1197 #1317 and #1372). Especially the later one will could always lead to problems in the future. 

Thus, the decision to use `@emtoion/react` again. The hope is that we have better control over style ordering (emotion currently only has a "prepend" option for the cache, there is no way to insert at a specific order, see also https://github.com/emotion-js/emotion/issues/2037) and can remove custom implementations like the `@marigold/server`.

## Consequences

We have to reimplement all the features of the current styling solution. Meaning:

- [x] A Component that acts as base for styling with the following features:
	- [x] renders an arbitrary HTML element (via `as` prop -> polymorphic component)
	- [x] passes down all props to the rendered element (expect styling props)
	- [x] adds normalization based on given element
	- [x] accepts a style objects (via `css` prop) that will generated styles
	- [x] interpolates style values based on theme (with the help of theme-ui)
	- [x] accepts variants (via `variants` prop)
	- [x] respects the weighting of styles (normalization < variants < css prop)
	- [x] normalization can only be applied once and must not override variants/css prop styles (see #1372 for an example)
- [x] Provider to pass `theme` and other configurations
	- [x] accepts a theme based on current specification
- [x] SSR render styles
- [x] Allow to add global styles (html, body via `theme.root`)
	- [x] apply global styles
- [x] Change Marigold Documentation files (replace useClassname/useStyles hooks)
	- [x] rewrite Guide useStyles
	- [x] rewrite Box Primitive
	- [x] rewrite Theme & Variants

sebald:
@ti10le close? :)

PR #1440, opened 2021-10-29
feature: use Element in Box + necessary fix changes

ti10le:
#1395


sebald (approved):
Please don't mix so much different stuff :P

ti10le:
> Please don't mix so much different stuff :P

Yes.. but these two other files had errors after the change in the Box

PR #1441, opened 2021-10-29
refa(comp): replace useStyles usages in docs package

ti10le:
#1395

### DIFF
--- a/docs/src/components/CodeBlock.tsx
+++ b/docs/src/components/CodeBlock.tsx
@@ -147,7 +147,7 @@ export const CodeBlock: React.FC<CodeBlockProps> = ({
               <LiveProvider scope={{ ...Components, ...Icons }}>
                 <Box
                   as="pre"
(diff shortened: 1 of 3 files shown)


docs/src/components/CodeBlock.tsx
```diff
@@ -131,21 +145,20 @@ export const CodeBlock: React.FC<CodeBlockProps> = ({
...
             <>
               <LiveProvider scope={{ ...Components, ...Icons }}>
-                <pre
-                  className={className + codeBoxStyles}
-                  style={{
-                    ...style,
-                  }}
+                <Box
+                  as="pre"
+                  css={{ ...style, ...codeBoxStyles }}
```
viktoria-schwarz, line 150: in line 92 you add the styles the other way around, is this on purpose here?

ti10le: Oh no it wasn't on purpose. I will correct it.

viktoria-schwarz (approved):

PR #1448, opened 2021-11-11
feat(comp): replace useStyles in Badge, Menu, Message, ValMes

ti10le:
#1395

### DIFF
--- a/packages/components/src/Menu/Menu.tsx
+++ b/packages/components/src/Menu/Menu.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { useStyles } from '@marigold/system';
 import { ComponentProps } from '@marigold/types';
 
 import { Button } from '../Button';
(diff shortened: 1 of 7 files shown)


packages/components/src/Menu/Menu.tsx
```diff
@@ -22,22 +21,24 @@ export const Menu: React.FC<MenuProps> = ({
...
       </Button>
-      {show ? <div className={itemStyles}>{children}</div> : null}
+      {show ? (
+        <Box
+          display="block"
+          position="absolute"
+          minWidth="120px"
+          borderRadius="2px"
+          css={{
+            textAlign: 'left',
```
viktoria-schwarz, line 36: I don't think that's needed, is it? It actually excludes an optional `text-align: right` or `center`.

packages/components/src/Message/Message.tsx
```diff
@@ -34,18 +33,11 @@ export const Message: React.FC<MessageProps> = ({
     >
       <Box display="flex" alignItems="center" variant="messages.title">
```
viktoria-schwarz, line 34: why do we need `flex` and `align-items: center` here?

ti10le: We dont need it 👍🏼

viktoria-schwarz (approved):

PR #1449, opened 2021-11-12
refa(comp): replace useStyles in ListBox, Textarea, Field.test

ti10le:
#1395

### DIFF
--- a/packages/components/src/Select/ListBoxSection.tsx
+++ b/packages/components/src/Select/ListBoxSection.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { useListBoxSection } from '@react-aria/listbox';
 import type { ListState } from '@react-stately/list';
-import { useStyles } from '@marigold/system';
 import type { Node } from '@react-types/shared';
 
 import { Box } from '../Box';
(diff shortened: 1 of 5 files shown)


packages/components/src/Select/ListBoxSection.tsx
```diff
@@ -17,15 +16,16 @@ export const ListBoxSection = ({ section, state }: SectionProps) => {
...
-    },
-  });
 
   return (
-    <Box as="li" {...itemProps} className={listClassName}>
+    <Box
+      as="li"
+      {...itemProps}
+      css={{
+        listStyle: 'none',
```
viktoria-schwarz, line 25: Doesn't t make more sense to have the `list-style` defined on the `ul` or `ol` element?

ti10le: Do you mean here https://github.com/marigold-ui/marigold/blob/main/packages/system/src/normalize.ts#:~:text=const%20list%20%3D%20%7B,%7D%20as%20const%3B ?

viktoria-schwarz: No but in the `<ListBox>` where the `<ListBoxSection>` is used.

PR #1467, opened 2021-11-22
feat(docs): add ComponentPreview + ActionType "tsx demo" to CodeBlock

### LINKED ISSUE #1379: Improve docs CodeBlock actions to display demo components
## Description

We need a possibility to display our components as a demo like in the current [marigold prototype](https://www.figma.com/proto/TIRxkIDpdqb52J52PldLxl/Guidelines?page-id=0%3A1&node-id=12%3A847&viewport=241%2C48%2C0.12&scaling=min-zoom&starting-point-node-id=12%3A847&show-proto-sidebar=1)
[Demo Design with styles](https://www.figma.com/file/TIRxkIDpdqb52J52PldLxl/Guidelines?node-id=12%3A847)
We have to extend the current ActionTypes in CodeBlock.tsx. ```tsx ...

There should be one option to display them:
- one component alone or next to each other like in the button example
- without any bordered box around them
- without showCode button
- without copy button

## Context

With this implementation we can step further to our current main target: **improve the marigold docs page**

## Use Cases

You have to start with a 'demo' at creating a components page in marigold docs. You can implement this with this new improvement.

## Consequences

We have a reusable implementation to display the marigold components

ti10le:
closes #1379 

Create a component named ComponentPreview for easier preview (also some work for DoAndDont)
add a new tsx CodeBlock ActionType named "Demo" - easier demonstration than import components in every mdx file
example in button.mdx


docs/src/components/CodeBlock.tsx
```diff
@@ -19,11 +19,13 @@ import * as Icons from '@marigold/icons';
 import { CopyButton } from './CopyButton';
 import { ShowHideButton } from './ShowHideButton';
 import { useThemeSwitch } from './ThemeSwitch';
+import { ComponentPreview } from './ComponentPreview';
 
 enum ActionType {
   Preview = 'preview',
   Live = 'live',
   Code = 'code',
+  Demo = 'demo',
```
viktoria-schwarz, line 28: to be honest I don't understand anymore when to use which ActionType.. can you add a one-line comment to each of these?

ti10le: Do you have any ideas for better naming?

viktoria-schwarz: No because I don't know the differences :see_no_evil:

viktoria-schwarz: what is the difference between `preview` and `live`?

ti10le: Ok bad description. Live ist the live coding example where you can type in the code box.

viktoria-schwarz: I just don't see why we need these 4 different kinds, because it's not bad to always edit the code, or to always be able to copy the code :)

ti10le: Ok yes, we can also always edit or copy the code

ti10le: I will make some changes and push something new stuff :)

ti10le:
fyi @sebald : @viktoria-schwarz and I decided to make every code example live editable with the difference that you can collapse or expand the code by default.
We decided to use following ActionType names as well:
 ```
CollapseCode = 'collapseCode',
  ExpandCode = 'expandCode',
  OnlyCode = 'onlyCode'
```

PR #1469, opened 2021-11-23
docs: add all marigold components to mdx

sebald:


ti10le:
we decided to take another way --> #1470

PR #1515, opened 2021-12-03
feat: add normalization for body and html & fix emotion leak

sebald:
- Remove emotion import form `components` so that it doesn't leak outside system package


sebald:
Note: This will only work after #1514 is merged

ti10le (approved):

PR #1546, opened 2021-12-07
feat(docs): copy SVG icon to clipboard

### LINKED ISSUE #1520: Search and copy Marigold Icons as <svg> in documentation
## Description

We want have a possibility to copy the marigold icons as svg directly on the docs page. Integration on our existing icons page.
https://www.figma.com/proto/TIRxkIDpdqb52J52PldLxl/Guidelines?page-id=701%3A8113&node-id=710%3A11078&viewport=247%2C48%2C1.2&scaling=min-zoom&starting-point-node-id=701%3A8330
https://www.figma.com/file/TIRxkIDpdqb52J52PldLxl/Guidelines?node-id=701%3A8113

## Context

Some teams that do not use react need a way to extract icons as <svg> element from the documentation.

## Use Cases

Core member copy icons from marigold and insert the svg in their code.

## Consequences

More people are using marigold and take something from our documentation. Higher number of visitors and users.

ti10le:
closes #1520


docs/src/components/CopySVGIcon.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { ReactElement, useState } from 'react';
+import { renderToString } from 'react-dom/server';
+import { Snackbar, Tooltip } from '@mui/material';
```
sebald, line 3: No lazy imports, see https://mui.com/guides/minimizing-bundle-size/

sebald (changes requested):
Also see slack

PR #1556, opened 2021-12-13
feat(comp): remove span from button + add some styles incl. space prop to button

ti10le:
prework for #1520

### DIFF
--- a/.changeset/happy-peaches-arrive.md
+++ b/.changeset/happy-peaches-arrive.md
@@ -1,0 +1,6 @@
+---
+'docs': patch
+'@marigold/components': patch
+---
+
+feat(comp): remove span from button and add some styles to the button element, add space prop to button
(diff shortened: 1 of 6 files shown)


packages/components/src/Inline/Inline.tsx
```diff
@@ -29,10 +29,15 @@ export const Inline: React.FC<InlineProps> = ({
...
   >
     {Children.map(
-      flattenChildren(children) as unknown as React.ReactElement,
-      (child: React.ReactElement) => (
-        <Box>{React.cloneElement(child, {}, child.props.children)}</Box>
-      )
+      flattenChildren(children) as unknown as ReactElement,
+      (child: ReactElement) =>
+        React.isValidElement(child) ? (
+          <Box display="inline-flex" alignItems={ALIGNMENT[align]}>
```
sebald, line 35: Do we really ned the extra element, the css can be applied to the root <Box>?

sebald (changes requested):

sebald (approved):

PR #1563, opened 2021-12-14
feat: Introduce a `Theme` with all available scales

sebald:


sebald:
hardned the type with makes TS scream at us because we are bad people :D

ti10le (approved):

PR #1582, opened 2021-12-17
fix(comp): inline-flex to Inline Box

ti10le:
wrong since the last Inline comp change.

from this:
image

to this: 
image

same for ActionGroup

### DIFF
--- a/packages/components/src/Inline/Inline.tsx
+++ b/packages/components/src/Inline/Inline.tsx
@@ -31,7 +31,9 @@ export const Inline: React.FC<InlineProps> = ({
     {Children.map(
       flattenChildren(children) as unknown as React.ReactElement,
       (child: React.ReactElement) => (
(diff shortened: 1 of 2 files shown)


packages/components/src/Inline/Inline.tsx
```diff
@@ -31,7 +31,9 @@ export const Inline: React.FC<InlineProps> = ({
     {Children.map(
       flattenChildren(children) as unknown as React.ReactElement,
       (child: React.ReactElement) => (
-        <Box>{React.cloneElement(child, {}, child.props.children)}</Box>
+        <Box display="inline-flex">
```
sebald, line 34: Do we want that? Why do we want the container to be inline?

sebald:
@ti10le I don't know the markup, but why does this fix this? Shouldn't be the check icon and its label inside<inline>?

Issue #1486, opened 2021-11-29
🎄 Presents for end of year cycle [Summary Issue]


sebald:
We want to start fresh next year so this is a lose bucket list for the rest of the year:

- [x] Remove Storybook Docs -> use regular TS(X) (https://storybook.js.org/docs/react/api/csf)
- [x] Re-fit the ordering of how styles are applied: #1494
- [x] Merge Box and Element: #1550
- [x] Remove all `useStyles` and delete related packages: #1395 
- [x] add GlobalStyles: #1395 
- [x] Icons site and search (with copy-pasta for the SVG code): #1520 
- [x] Refactor layout components
	- [x] Column/Columns: #1035 
	- [x] Stack: #1510 
	- [x] Inline: #1511 
- [x] Cleanup Themes and Theme structure 
	- [x] #996
	- [x] check all component variants (if they are correct, better naming, etc.) --> last of this is handled in #1570, #1571, #1638

ti10le:
What do you mean exactly with icons site and search? .. copy-pasta sounds tasty ☺️

sebald:
@ti10le the request from @julianthiel about a searchable page for icons where you can copy the svg code (for usage outside marigold)

ti10le:
@sebald can we collect the points within:  Cleanup Themes and Theme structure
Whats all included in this point? How can theme-ui help us with this issue?

ti10le:
Last missing is to rewrite the component stories to .tsx
Then we can check if #1226 and #1255 is also done with this work.

PR #1620, opened 2022-01-05
feat(comp): extend Box to handle __default variants

ti10le:
#1486


packages/system/src/Box.tsx
```diff
@@ -69,8 +69,19 @@ export type BoxProps = PolymorphicPropsWithRef<BoxOwnProps, 'div'>;
 const isNotEmpty = (val: any) =>
   !(val && Object.keys(val).length === 0 && val.constructor === Object);
 
-const ensureArray = <T extends any>(val?: T | T[]) =>
-  Array.isArray(val) ? val : [val];
+/**
+ * Ensure that variant is an array and remove dot at the end of the string
+ * to set __default in theme without giving a default variant to the component
+ */
+const ensureCorrectVariant = <T extends string>(variant?: T | T[]) => {
```
sebald, line 76: Can we make this function more simple to read please? :D

sebald (changes requested):

packages/system/src/Box.tsx
ti10le: Better? Or how can I make it easier?

packages/system/src/Box.tsx
```diff
@@ -72,6 +72,19 @@ const isNotEmpty = (val: any) =>
 const ensureArray = <T extends any>(val?: T | T[]) =>
   Array.isArray(val) ? val : [val];
 
+/**
+ * Ensure that variant is an array and remove dot at the end of the string
+ * to set __default in theme without giving a default variant to the component
+ */
+const ensureCorrectVariant = (variant?: any | any[]) => {
```
sebald, line 79: can only be string or string array, right?

packages/system/src/Box.tsx
```diff
@@ -72,6 +72,19 @@ const isNotEmpty = (val: any) =>
...
+  if (variant) {
+    return ensureArray(variant).map(v => {
+      return v.replace(/\.$/, '');
```
sebald, line 82: unnecessary return, use shorthand? :)

packages/system/src/Box.tsx
```diff
@@ -72,6 +72,19 @@ const isNotEmpty = (val: any) =>
...
+    });
+  }
+  return [variant];
```
sebald, line 85: This will always return [undefined]. Why do you need this here :)

packages/system/src/Box.tsx
ti10le: or undefined

sebald (approved):

PR #1634, opened 2022-01-12
feat(comp): Radio component rewrite

### LINKED ISSUE #1570: <Radio> has no prefix for its variant and hardcoded variants
## Description

Radio is missing a prefix for its ' variant'. We should add a prefix and also add it to the `Theme`.
Radio shouldn't have hardcoded variants.

## Context

We don't want to force the Theme to have specific variants in a component.
Radio should have a variant.

## Consequences

Easier to write and access a theme. Easier to use the Radio component.

ti10le:
closes #1570


packages/components/src/Radio/Radio.tsx
```diff
@@ -101,28 +63,26 @@ export const Radio: React.FC<RadioProps> = ({
...
-      </>
-    );
-  }
-
-  return <RadioInput {...props} />;
-};
+}) => (
+  <>
+    <Box
+      as={Label}
```
sebald, line 69: Why no `<Label>` ?

packages/components/src/Radio/Radio.tsx
```diff
@@ -16,78 +17,39 @@ export interface RadioThemeExtension<Value> {
...
   return (
-    <Box
-      as={RadioUnchecked}
-      variant={`radio.${variant}`}
-      disabled={disabled}
-      error={error}
-    />
+    <Box pr="xsmall">
+      <VisuallyHidden>
+        <Box
```
sebald, line 30: No need to use the `<Box>` here we don't style it anyway

packages/components/src/Radio/RadioIcon.tsx
```diff
@@ -0,0 +1,51 @@
...
+    >
+      <Box
+        variant={conditional(`radio.${variant}`, conditionalStates)}
+        as="circle"
+        cx="8"
+        cy="16"
+        r="7.5"
+      />
+      {checked && (
+        <Box __baseCSS={{ fill: 'gray00' }} as="circle" cx="8" cy="16" r="3" />
```
sebald, line 47: No need to use `<Box>` you can use a regular cicle here.

.changeset/sharp-shrimps-bake.md
```diff
@@ -0,0 +1,6 @@
+---
```
sebald, line 1: One changeset is enough? :D

packages/components/src/Radio/Radio.test.tsx
```diff
@@ -98,22 +103,29 @@ test('supports error and errorMessage prop', () => {
...
+        disabled
+      />
     </ThemeProvider>
   );
 
   const radio = screen.getByTitle(/radio/);
-  expect(radio).toBeDefined();
+  expect(radio).toHaveAttribute('checked');
   expect(radio).toHaveAttribute('disabled');
 });
```
sebald, line 131: Add an interaction test!? :)

sebald (changes requested):

sebald (changes requested):
see comments

packages/components/src/Radio/Radio.tsx
ti10le: because of the __baseCSS prop

packages/components/src/Radio/RadioIcon.tsx
ti10le: but how to use the fill prop? Is it better to use `fill='white'` ?

packages/components/src/Radio/Radio.tsx
sebald: I don't see where the difference between a hover effect and the color is. Setting the color like you did will override anything. Why not using `css` prop and do this with the cursor too? :)

packages/components/src/Radio/RadioIcon.tsx
sebald: `gray80` isn't part of any theme too and might me just undefined :-/

ti10le: Means this "yes" to my question?

packages/components/src/Radio/Radio.tsx
ti10le: Yes, I can use cursor and color in the css prop. Then we wouldn't need the color prop on the Label.
At the end we have also Box as label.
```
 <Box
      as={Label}
      htmlFor={props.id}
      required={required}
      variant={`label.${labelVariant}`}
      css={
        props.disabled
          ? { color: 'disabled', ':hover': { cursor: 'not-allowed' } }
          : { color: 'text', ':hover': { cursor: 'pointer' } }
      }
    >
```

sebald: Ah right. No CSS prop on the label. My bad, sorry.

Makes me think of we should expose the css prop on more components actually 😄 but this is another issue

sebald (approved):

PR #1647, opened 2022-01-17
feat(system): conditional helper to allow state in variants

### LINKED ISSUE #1640: Introduce the ability to have stateful styles in the theme for forms
### API "statesful" form components

```js
const theme = {
  '<some input>': {
    __default: {},
    ':focus': {},
    ':hover': {},
    ':disabled': {},
  },
  radio: {
    __default: {},
    ':checked': {}, // only when using type=[checkbox|radio]
    ':focus': {},
    ':hover': {},
    ':disabled': {},
  },
};
```

### From declarative style object to styles

```js
<Input variant="radio" />
```

1. Get styles from `theme` -> `theme.radio`
2. Always apply `__default`
3. Apply state styles (`:disabled`, ...) based on props

```js
<Input variant="radio" disabled />
```

### Programatically add variants based on states

```tsx
import { conditional } from '@marigold/system`;

const Input = ({ variant, disabled, checked, ...props }) => {
  return <Box variant={conditional({ variant, disabled, checked })} />;
};

type State = {
  checked?: boolean;
  focus?: boolean;
  hover?: boolean;
  disabled?: boolean;
};

const toVariantState = (variant, state) => `${variant}.:${state}`;

const conditional = ({ variant, disabled, ...state }) => {
  const conditionalVariants = Object.entries(state)
    .filter(([_, val]) => Boolean(v))
    .map(([key, val]) => toVariantState(variant, key));
  // TODO: Make sure disabled is last conditional variant
  return [variant, ...conditionalVariants, ...(disabled && toVariantState(variant, 'disabled'))];
};
```

Example:

```js
conditional({ variant: 'input', disabled: true });
// -> ['input', 'input.:disabled']
```

### Alternative: Instead if JS use CSS and transform theme

**Downside**: Based on HTML structure :( -> Don't do it

```js
const theme = {
  '<some variant>': {
    __default: {},
    ':focus': {},
    ':hover': {},
    ':disabled': {},
  },
};
```

transform into:

```js
const theme = {
  '<some variant>': {
    __default: {
      // all default styles

      // transformed "sate" styles
      '<some variant> + input[type="radio"]:focus ~ svg': {}
      '<some variant> + input[type="radio"]:focus ~ span': {}

      // etc
    },
  }
};
```

ti10le:
closes #1640

### DIFF
--- a/packages/system/src/conditional.test.tsx
+++ b/packages/system/src/conditional.test.tsx
@@ -9,9 +9,6 @@ test('returns the variant and the conditional states', () => {
       "input.:disabled",
     ]
   `);
-  // expect(['input', 'input.:checked', 'input.:disabled']).toEqual(
-  //   expect.arrayContaining(variant)
-  // );
 });
 
 test('returns only truthy states', () => {
@@ -27,10 +24,10 @@ test('returns only truthy states', () => {
 test('sets disabled at the last position in the returned array', () => {
   const variant = conditional('input', { disabled: true, checked: true });
   expect(variant).toMatchInlineSnapshot(`
-  [
-    "input",
-    "input.:checked",
-    "input.:disabled",
-  ]
-`);
+      [
+        "input",
+        "input.:checked",
+        "input.:disabled",
+      ]
+  `);
 });
--- a/packages/system/src/conditional.test.tsx
+++ b/packages/system/src/conditional.test.tsx
@@ -1,0 +1,31 @@
+import { conditional } from './conditional';
+
+test('returns an array of strings', () => {
+  const variant = conditional('input', { checked: true, disabled: true });
+  expect(variant).toEqual(expect.arrayContaining([expect.any(String)]));
+});
+
+test('returns the variant and the conditional states', () => {
+  const variant = conditional('input', { checked: true, disabled: true });
+  expect(['input', 'input.:checked', 'input.:disabled']).toEqual(
+    expect.arrayContaining(variant)
+  );
+});
+
+test('returns only truthy states', () => {
+  const variant = conditional('input', { checked: true, disabled: false });
+  expect(['input', 'input.:checked']).toEqual(expect.arrayContaining(variant));
+});
+
+test('removes dot from the end of the variant string', () => {
+  // This happens if you have variant='' in your component and define __default in your theme
+  const variant = conditional('input.', { checked: true, disabled: false });
+  expect(['input', 'input.:checked']).toEqual(expect.arrayContaining(variant));
+});
+
+test('sets disabled at the last position in the returned array', () => {
+  const variant = conditional('input', { disabled: true, checked: true });
+  expect(['input', 'input.:checked', 'input.:disabled']).toEqual(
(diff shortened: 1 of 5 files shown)


packages/system/src/conditional.test.tsx
```diff
@@ -0,0 +1,31 @@
+import { conditional } from './conditional';
+
+test('returns an array of strings', () => {
```
sebald, line 3: I think this test is more confusing than helping. remove it, the next test already tests this too

packages/system/src/conditional.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const variant = conditional('input', { checked: true, disabled: true });
+  expect(variant).toEqual(expect.arrayContaining([expect.any(String)]));
+});
+
+test('returns the variant and the conditional states', () => {
+  const variant = conditional('input', { checked: true, disabled: true });
+  expect(['input', 'input.:checked', 'input.:disabled']).toEqual(
```
sebald, line 10: use an inline snapshot :D

packages/system/src/conditional.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+    expect.arrayContaining(variant)
+  );
+});
+
+test('returns only truthy states', () => {
+  const variant = conditional('input', { checked: true, disabled: false });
+  expect(['input', 'input.:checked']).toEqual(expect.arrayContaining(variant));
+});
+
+test('removes dot from the end of the variant string', () => {
```
sebald, line 20: Isn't that part what box does?

packages/system/src/conditional.tsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+const toVariantState = (variant: string, state: string) => {
+  return `${variant}.:${state}`;
+};
+
+/**
+ * function to create an array of variant and their conditional props
+ */
+export const conditional = (variant: string, state: State) => {
```
sebald, line 16: Can you destruct `disabled` here to simplify the function? 
`(variant, { disabled = false , ... states})

packages/system/src/conditional.test.tsx
ti10le: why are you laughing, it works  😀

packages/system/src/conditional.test.tsx
ti10le: Yes, but I need this function directly in conditional() this is why I cannot use it with the Box like others.
One option is to export the ensureCorrectVariant() and use it. But then I have to make [0] because this returns an array.

packages/system/src/conditional.test.tsx
ti10le: Done.

packages/system/src/conditional.test.tsx
ti10le: I can make it like this: ` const correctVariant = ensureCorrectVariant(variant)[0];`

sebald (approved):

Issue #1654, opened 2022-01-20
Component update - Field text & Textara - implement react-aria functionality


ti10le:
## Description

Make our Textfields full accessible. This has impact on the following components: **Field, Textarea**
See react-aria [useTextfield](https://react-spectrum.adobe.com/react-aria/useTextField.html) example.

## Context

We want to have components that meet the accessibilty restrictions.

## Consequences

Accessible Text components

PR #1708, opened 2022-01-27
feat(comp): rewrite Field component with react-aria

ti10le:
#1654


.changeset/tricky-clouds-attend.md
```diff
@@ -0,0 +1,6 @@
+---
+"@marigold/docs": patch
```
sebald, line 2: minor!

sebald (approved):

PR #1727, opened 2022-02-02
feat(components): <Text> change font size via style props

sebald:


sebald:
@ti10le we didn't talk about this but doing #1726 this would have been a nice featur :)

packages/components/src/Text/Text.tsx
```diff
@@ -50,7 +52,14 @@ export const Text: PolymorphicComponentWithRef<TextOwnProps, 'span'> =
         {...props}
         as={as}
         variant={`text.${variant}`}
-        css={{ textAlign: align, color, cursor, outline, userSelect }}
+        css={{
```
ti10le, line 55: @sebald should these props be able to be overwritten by the variant or not? 🤔

ti10le (changes requested):

ti10le:
Please add size prop to docs table :) Maybe to storybook, too.

ti10le (approved):

packages/components/src/Text/Text.tsx
sebald: I think not ... I hope not :D (since its `undefined`)

Issue #1691, opened 2022-01-25
Component - Tooltip


ti10le:
## Description

Tooltip component to display some information by hovering over an info icon.
- A11y from aria: [useTooltipTrigger with example](https://reactspectrum.blob.core.windows.net/reactspectrum/d77b35e970e5549f66b47a83f07423f5c93b7297/docs/react-aria/useTooltipTrigger.html)
- [Figma Design](https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=2461%3A2252)
- [Braid DS Tooltip](https://seek-oss.github.io/braid-design-system/components/TooltipRenderer)
- [Carbon DS Tooltip](https://react.carbondesignsystem.com/?path=/story/components-tooltip--default-bottom)
- [mui](https://mui.com/components/tooltips/)

## Context

We need this component to make the Design System more complete and to support some use cases with a tooltip.

## Use Cases

In the new Checkout to show information with hover over an info icon.

## Consequences

Tooltip can be used from marigold and does not have to be built by yourself.

ti10le:
I am on a first version of the Tooltip. I would differentiate in steps because of complexity.
done and coming with the first PR:
- simple Tooltip without arrow in different placements (coming in next steps)
- react aria implementation
- tests
- doc file with Do and Don't
- storybook file

ti10le:
First version of Tooltip is already done. If it needs more functionality we can do it in an extra issue.

PR #1807, opened 2022-02-24
feat(comp): add Tooltip component v1

ti10le:
#1691


packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+// Component
+// ---------------
+export const Tooltip: React.FC<TooltipProps> = ({
+  variant = '',
+  children,
+  ...props
+}) => {
+  const { state, ...tooltipProviderProps } = useContext(TooltipContext);
+  props = mergeProps(props, tooltipProviderProps);
```
sebald, line 31: Don't mutate `props`.

sebald (changes requested):
see comments

sebald (approved):

PR #1809, opened 2022-02-25
feat(create-theme): Introduce helpers to create themes

### LINKED ISSUE #1704: Use open-props as basis for themes
## Description

Prevent repetition! Use [Open Props](https://open-props.style/) for themes. This way we can also create a helper to create themes that is type safe!

We need to convert the naming convention of open props (number based) into ours (sizes based).

## Context

We currently have a lot of repetition in our themes, especially when it comes to the design tokens / scales. [Open Props](https://open-props.style/) is a good baseline

## Consequences

We have a good baseline with semantic meaning. This could be irrelevant in the future though when the design language matures.

sebald:
Introduce helpers to create themes based on some sensible deault tokens.

Closes #1704


packages/system/src/theme.ts
```diff
@@ -28,30 +28,43 @@ export type ScaleValue<T> = T | T[] | NestedScaleDict<T> | undefined;
...
+  none: ScaleValue<T>;
+}
 
 /**
  * Predefined {@link Scale} scale which uses size values.
  */
-export type SizeScale<T> = {
+export interface SizeScale<T> {
+  regular?: ScaleValue<T>;
+
```
ti10le, line 48: Is this line intentionally?

packages/system/src/theme.ts
```diff
@@ -28,30 +28,43 @@ export type ScaleValue<T> = T | T[] | NestedScaleDict<T> | undefined;
...
+export interface Scale<T> {
   [key: string]: ScaleValue<T>;
-};
+}
+
+/**
+ * A {@link Scale} that also includes a required `none` value, which is
+ * usually used to define the blank value (e.g `0`).
+ */
+export interface ZeroScale<T> extends Scale<T> {
```
ti10le, line 39: Why not putting the two zeroScales to the other SizeScales?

packages/system/src/theme.ts
```diff
@@ -107,13 +120,13 @@ export interface Theme {
   /**
    * Used to define a `letter-spacing` scale.
    */
-  letterSpacings?: ZeroScale<CSS.Property.LetterSpacing<string | 0 | number>>;
+  letterSpacings?: Scale<CSS.Property.LetterSpacing<string | 0 | number>>;
```
ti10le, line 123: is it not ZeroScale to ZeroSizeScale?

themes/create-theme/src/create-theme.test.ts
```diff
@@ -0,0 +1,295 @@
...
+    {
+      "large": "clamp(1.5rem, 6vw, 2.5rem)",
+      "regular": "clamp(1rem, 4vw, 1.5rem)",
+      "small": "clamp(.75rem, 2vw, 1rem)",
+      "xlarge": "clamp(2rem, 9vw, 3.5rem)",
+    }
+  `);
+});
+
+test('uses fixed values for dimensaions by default', () => {
```
ti10le, line 172: *dimensions

ti10le (changes requested):

ti10le:
Sooo good, so much structure 😍

ti10le:
I still have one question. We specify with open props certain things like colors and typographies and I saw you also allowed to add custom. Now it would be great if we discuss this with our designers and only take the values defined here to remain consistent.
This would be a major overhaul of all themes (especially b2b) but it would be much better if our themes would also use the values defined here. Otherwise we use that but have a thousand custom things back in it....

How do you see it @sebald?

packages/system/src/theme.ts
sebald: yes, we didn't have enough values and didn't hat the "base" value.

sebald: Doing this PR I actually find our scale cumbersome ... open props originally used numbers which might be the better approach

ti10le: I mean the blank line :) its also in other files, a blank line after "regular" or "none"

packages/system/src/theme.ts
sebald: Not sure what you mean 🤔

packages/system/src/theme.ts
sebald: There is no `lineheight: 0` so know :)

packages/system/src/theme.ts
ti10le: Why we cant combine Scale and ZeroScale together in one interface?
Or SizeScale with zeroSizeScale. E.g: add none to SizeScale without having an extra ZeroSizeScale

sebald:
@ti10le We don't necessary need to use it for the B2B theme yet. OpenProps should be the basis for the B2C theme and our docs theme. But yes, it would be nice to have this in the future.

Also, since Toshi didn't use arbitrary values for scales most of the values are identical with ours in the B2B theme.

This is the first iteration and was already a very large PR, it definitely need improvement. For example, in the future we should also include sensible defaults for the colors we already have in the B2B theme (like error, warning, info, muted, ...).

packages/system/src/theme.ts
sebald: ZeroScale and Scale: because we have properties that don't have a "zero", same for SizeScale and ZeroSizeScale. it depends how strickt we want to be with our `Theme` definition.

packages/system/src/theme.ts
sebald: I did it to make the value stand out :D

ti10le: ok for me 😀

ti10le (approved):

PR #1819, opened 2022-03-02
feat(comp): add new Component Switch

### LINKED ISSUE #1653: Component - Switch
## Description

Create a toggle switch component to choose between two states.
- [Switch Design](https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=2458%3A2252)
- [react-aria useSwitch](https://react-spectrum.adobe.com/react-aria/useSwitch.html)

## Context

Have one more component we need in future

## Use Cases 

First usage could be in the new rx Checkout or in marigold docs to switch between dark and light theme

## Consequences

Easier to use a Switch functionality in an application which uses marigold

ti10le:
closes #1653


packages/components/src/Switch/Switch.test.tsx
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
```
sebald, line 1: Add a test that shows the controlled state of the component.

packages/components/src/Switch/Switch.tsx
```diff
@@ -0,0 +1,97 @@
...
+  disabled?: boolean;
+} & AriaSwitchProps &
+  ToggleProps &
+  ComponentProps<'input'>;
+
+// Component
+// ---------------
+export const Switch: React.FC<SwitchProps> = ({
+  variant = '',
+  labelVariant = 'above',
```
sebald, line 37: Can we use a child like in `<Checkbox>`?

packages/components/src/Switch/Switch.tsx
```diff
@@ -0,0 +1,97 @@
...
+        <input
+          {...inputProps}
+          {...focusProps}
+          disabled={disabled}
+          ref={ref as LegacyRef<HTMLInputElement>}
+        />
+      </VisuallyHidden>
+      <Box
+        as="svg"
+        __baseCSS={{
```
sebald, line 63: add `userSelect: 'none'` so text doesn't selected on accident

sebald: image

sebald (changes requested):

packages/components/src/Switch/Switch.tsx
ti10le: Yes I do it with props.children but I can also change to children

packages/components/src/Switch/Switch.tsx
sebald: No in the baseCSS prop :)

ti10le: but this has no effect

ti10le: I've tried it like this but no effect: https://css-tricks.com/almanac/properties/u/user-select/

ti10le:
@sebald its ready now :)

sebald (approved):

PR #1822, opened 2022-03-02
feat(comp/docs): Alert update variant + docs

ti10le:
context: https://www.figma.com/file/TIRxkIDpdqb52J52PldLxl/Guidelines?node-id=60%3A552


sebald (approved):

sebald:
@ti10le maybe we should rename this component since `info` is not really an alert :D

PR #1857, opened 2022-03-11
feat(comp): remove flattenChildren from Inline

ti10le:
afterwork from this #1851


sebald:
@ti10le there is no reason for a changeset :P

sebald (approved):

Issue #1659, opened 2022-01-20
Release npm packages with CI Pipeline


ti10le:
## Description

Currently we release only every few weeks in irregular intervals. We want an automated solution so that our npm packages are always up to date. Best would be an integration in the CI pipeline.

e.g.
- github actions
- [GitHub](https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages)

In this case check also if the released changesets will be deleted.

## Context

npm packages have another version than marigold docs and the DS repo

## Consequences

Continuous releases, up to date npm packages

sebald:
https://github.com/changesets/action

sebald:
Blocked by changesets/changesets#714

sebald:
Unblocked by #1850

sebald:
We can start with this example: https://github.com/changesets/action#with-publishing

```
name: Release

on:
  push:
    branches:
      - master

concurrency: ${{ github.workflow }}-${{ github.ref }}

jobs:
  release:
    name: Release
    runs-on: ubuntu-latest
    steps:
      - name: Checkout Repo
        uses: actions/checkout@v2
        with:
          # This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits
          fetch-depth: 0

      - name: Setup Node.js 12.x
        uses: actions/setup-node@v2
        with:
          node-version: 12.x

      - name: Install Dependencies
        run: yarn

      - name: Create Release Pull Request or Publish to npm
        id: changesets
        uses: changesets/action@v1
        with:
          # This expects you to have a script called release which does a build for your packages and calls changeset publish
          publish: yarn release
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}

      - name: Send a Slack notification if a publish happens
        if: steps.changesets.outputs.published == 'true'
        # You can do something when a publish happens.
        run: my-slack-bot send-notification --message "A new version of ${GITHUB_REPOSITORY} was published!"
        ```

sebald:
Also for slack notification: https://github.com/marketplace/actions/slack-notify

sebald:
Add GH releases:

https://github.com/MylesBorins/node-osc/blob/main/.github/workflows/create-release.yml

PR #1859, opened 2022-03-11
chore: slack notify

sarahgm:
#1659


sarahgm:
todo: get the version number from package.json

PR #1865, opened 2022-03-14
Fix getting started

snragund:


sebald:
@all-contributors please add @snragund for doc

Issue #1860, opened 2022-03-11
All the layouts!


sebald:
## Description

We currently have `<Stack>`, `<Inline>` and `<Columns>` to make layout easier. But we need MOAR!

**We should review these**

- [x] Inline: see https://every-layout.dev/layouts/cluster/ #1851 
- [x] Stack: see https://every-layout.dev/layouts/stack/ #1867 
- [x] Columns  #1925 

---

**We should do these**

- [x] Container: does nothing currently, should constraint content width #1868 --> #1926
- [x] Aspect: For images and other media, element that has a certain aspect ratio #1881 
- [x] Center: see https://every-layout.dev/layouts/center/ #1870
- [x] Breakout: see https://www.joshwcomeau.com/css/full-bleed/ #1930 
- [x] Grid: with sensible defaults, not `display: grid` #1716 #1882 
- [x] Aside: see https://every-layout.dev/layouts/sidebar/ #1874 

## Context

There are still a few common cases missing that we don't have ready to use layout components for.

## Consequences

Users don't have to fallback to the `<Box>` so much.

ti10le:
- I think we need a separation in docs after creating all layout components @sebald.
Components and Layouts e.g.

- And I think we have to write Do's and Don'ts to show e.g. when to use Center and Container

ti10le:
Currently we do not know where and in which cases we can use center. Therefore, we leave it out for now and add it when needed.

sebald:
Implementing `<Center>` would make sense if we want layouts with certain `maxHeight` and `maxWidth` that are not based on content (`ch` unit).

PR #1867, opened 2022-03-14
feat(comp): improve Stack component with gap

ti10le:
#1860


sebald:
@ti10le no API change, no changeset :P

PR #1868, opened 2022-03-15
feat(comp): renew Container component

ti10le:
#1860


docs/content/components/container.mdx
```diff
@@ -2,30 +2,139 @@
...
 ---
 
+import { Box, Container, Text } from '@marigold/components';
+
 # Container
 
 ## Description
 
-Primitive layout component to embed other elements.
+Containers are used to constrain a content's width to the current breakpoint, while keeping it fluid. You can use `header`or `content`as content type and set the `maxWidth` to `small`, `medium` or `large`.
```
sebald, line 11: We don't respect breakpoints here, I think 🤔

Maybe: width is based on used font and is optimized for readability

sebald (approved):

PR #1869, opened 2022-03-15
feat: add size scale to themes

sebald:


sebald:
@ti10le no perfect but okay for the moment. 

the reason the sizes are small is because I didn't had a good solution when writing the design tokens. See:

https://github.com/marigold-ui/marigold/blob/fb735b2d2d6f979370c52d7977cfdc258ec4965e/packages/create-theme/src/create-theme.ts#L74-L75

We should think about a better solution for this (e.g. have more sizes in the tokens and then map them in a meaningful way when creating the theme).

ti10le:
> @ti10le no perfect but okay for the moment.
> 
> the reason the sizes are small is because I didn't had a good solution when writing the design tokens. See:
> 
> https://github.com/marigold-ui/marigold/blob/fb735b2d2d6f979370c52d7977cfdc258ec4965e/packages/create-theme/src/create-theme.ts#L74-L75
> 
> We should think about a better solution for this (e.g. have more sizes in the tokens and then map them in a meaningful way when creating the theme).

Ok then I think we should find a solution to display more than this values soon.

ti10le (approved):

PR #1870, opened 2022-03-16
feat(comp): Center component

ti10le:
#1860

### DIFF
--- a/docs/content/components/center.mdx
+++ b/docs/content/components/center.mdx
@@ -1,0 +1,93 @@
+---
+title: Center
+---
+
+# Center
+
+Center is a layout component that centers its children horizontally.
+You can use it to center a single child or a list of children.
+
+<Center maxWidth="min(350px, 50vw)">
+  <Box border="1px solid #ced4da" bg="#e9ecef" height="100px" width="100%" />
+</Center>
+<br />
+
+```tsx onlyCode
+import { Center } from '@marigold/components';
+```
+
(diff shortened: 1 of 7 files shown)


sebald:
@ti10le add gap? :)

docs/content/components/center.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+## Properties (API)
+
+| Property         | Type                           | Default  |
+| :--------------- | :----------------------------- | :------- |
+| `maxHeight`      | `string`                       |          |
+| `maxWidth`       | `string`                       |          |
+| `textAlign`      | `left, center, right`          | `center` |
+| `space`          | `ResponsiveStyleValue<string>` | `none`   |
+| `superCenntered` | `boolean`                      | `false`  |
```
sebald, line 27: Shouldn't the layout always be centered? "Max height" didn't make sense otherwise 🤔

ti10le:
Currently we do not know where and in which cases we can use center. Therefore, we leave it out for now and add it when needed.

ti10le:
@sebald I've made some rewrite. It's not for super centering with an position absolute.
orientation: [every-layout](https://every-layout.dev/layouts/center/) & [chakra](https://chakra-ui.com/docs/components/layout/center)

sebald:
Will take a look at it tomorrow :)

sebald (approved):

PR #1886, opened 2022-03-22
fix release script for pnpm

ti10le:


scripts/release.mjs
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env zx
 
 // Set available globals for eslint
-/* global $, cd, question, chalk */
+/* global $, question, chalk */
```
ti10le, line 4: there was an warning that cd is unused

Issue #1832, opened 2022-03-04
Script to generate component doc template


sebald:
## Description

Create a script that generated a template to document a component.

Should 

- allow to set a name
- has a documentation sections and info what to put there

## Context

Our component documentation is not unified, sometimes sections are missing or the order is wrong.

## Consequences

Unify documentation layout.

sebald:
We could use https://www.npmjs.com/package/plop to generate template

PR #1887, opened 2022-03-22
docs: template for component docs

sebald:
Ref #1832

### DIFF
--- a/docs/templates/component.mdx
+++ b/docs/templates/component.mdx
@@ -10,18 +10,6 @@ figma: <link to figma> (optional)
 import {} from '@marigold/';
 ```
 
-## Examples
-
-<!-- show some cases and scenarios in which the component is used -->
-
-### Use case
-
-<!-- describe the use case -->
-
-```tsx
-<div>Hello there!</div>
-```
-
 ## Props
 
 <!-- list all props with values and description -->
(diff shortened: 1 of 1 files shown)


sebald:
We should create a `<PropsTable>` and wrapper (that includes the grid/stack stuff` for the Dos and Don'ts

docs/templates/component.mdx
```diff
@@ -0,0 +1,37 @@
...
+figma: <link to figma> (optional)
+---
+
+<!-- longer description about capabilities and usage of the component -->
+
+```tsx onlyCode
+import {} from '@marigold/';
+```
+
+## Examples
```
ti10le, line 13: Examples is like showing different props in action or dont show every prop separate in an example?

docs/templates/component.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+<!-- show some cases and scenarios in which the component is used -->
+
+### Use case
```
ti10le, line 17: Use case should be one or two cases or what do you think about it?

docs/templates/component.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+<!-- describe the use case -->
+
+```tsx
+<div>Hello there!</div>
+```
+
+## Props
```
ti10le, line 25: I would move this under the import

ti10le (dismissed):

docs/templates/component.mdx
sebald: yes, see description in comment

docs/templates/component.mdx
sebald: This is only meant as a placeholder, it's "examples" -> "primary button" for example and the latter is the "use case"

ti10le (approved):

Issue #1889, opened 2022-03-23
Form Revolution


sebald:
## Description

We have to go over the forms and unify their API!

**New**

- [x] `<CheckboxGroup>`  #2016
- [x] `<RadioGroup>` #2029

---

**Refactor**

- [x] Form components should have a common props
- [x] Label should have only one variant
- [x] For Checkbox und Radios don't use our Label component for `<label>` use Box ( `<Box as="label" variant="checkbox.value"/>`)
- [x] Label #1914
- [x] HelpText (support error and description) #1920
- [x] TextField #1963
- [x] NumberField #2056
- [x] TextArea #1964 
- [x] Checkbox #2008
- [x] Radio #2029
- [x] Select #2048
- [x] Slider #2034 
- [x] Switch #2049
- [x] FieldBase supports `width` from theme via prop  #2070

**Bugs**
- [x] #2011

**Delete**
- [x] ValidationMessage #2052

**API**

- [x] ~~Make everything available under the "Field.<...>" API!~~ Not doing this since `Field.Select.Option` is super long and weird to type. This doesn't make the usage of Marigold better.

## Consequences

Easier to use forms in Marigold 🤞

sebald:
@ti10le we could also change the import API like the following:

```jsx
import { Field } from '@marigold/components';

<Field.Text/>
<Field.TextArea/>
<Field.Checkbox/>
<Field.Select/>
```

I could life with that 😅

sebald:
What's the common API of Field Elements?

- name
- label
- description
- errorMessage

- disabled
- readonly
- required

- defaultValue
- state = valid | invalid | warning

Plus the regular props that the form HTML element has

ti10le:
This would be also good for me. Then we don't include the same things from field in every field component 😊 It would be clear that all of them a Field component.

sebald:
Yes, exactly 😄 question is how we document this and structure it inside of `@marigold/components`.

Do we create a Field/Form folder and put everything there?

ti10le:
I would create a folder for every component like now but dont show the Field component in docs or storybook. Only together with the other part. Not all inside Field i think 🤔

sebald:
Might be confusing for imports

ti10le:
Why? Not publishing the Field folder

sebald:
> Why? Not publishing the Field folder

@ti10le not sure what do you mean.

[...]

PR #1914, opened 2022-03-25
feat: Refactor <Label>

sebald:
- no variants anymore
- no disabled (not possible anyway)
- improve a11y

Ref #1889

### DIFF
--- a/packages/components/src/Field-Next/Label.tsx
+++ b/packages/components/src/Field-Next/Label.tsx
@@ -1,0 +1,51 @@
+// TODO: But this back into the root, when we realease all the new fields!
+import React from 'react';
+import { Required } from '@marigold/icons';
+import { ComponentProps } from '@marigold/types';
+
+import { Box } from '../Box';
+
+// Theme Extension
+// ---------------
+export interface LabelThemeExtension<Value> {
+  label?: {
+    [key: string]: Value;
+  };
+}
+
+// Props
+// ---------------
+export interface LabelProps extends ComponentProps<'label'> {
+  as?: 'label' | 'span';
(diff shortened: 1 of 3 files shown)


sebald:
No changeset since it is not publicly available yet.

packages/components/src/Field-Next/Label.tsx
```diff
@@ -0,0 +1,51 @@
+// TODO: But this back into the root, when we realease all the new fields!
```
ti10le, line 1: *Put 
*release
😀

packages/components/src/Field-Next/Label.tsx
```diff
@@ -0,0 +1,51 @@
...
+      as={as}
+      variant="label"
+      __baseCSS={{ display: 'flex', alignItems: 'center', gap: 4 }}
+    >
+      {children}
+      {/*
+       * aria-required is set on the field and will already be announced,
+       * so we don't need to add it here.
+       */}
+      {required && <Required role="img" size={16} />}
```
ti10le, line 48: Just for m information: why role img?

ti10le (approved):

ti10le:
@sebald questions hard ignored 😂

sebald:
@ti10le don't mark it as approved 😝

packages/components/src/Field-Next/Label.tsx
sebald: https://lmgtfy.app/?q=role+img 🤪

ti10le:
That was intentional, because it can be gemerged, just thought I still get an answer 😀

PR #1919, opened 2022-03-27
feat: SVG supports size token from theme

sebald:


ti10le (approved):

packages/system/src/SVG.tsx
```diff
@@ -21,10 +23,8 @@ export const SVG: React.FC<SVGProps> = ({
   return jsx(
     'svg',
     {
-      width: size,
```
ti10le, line 24: @sebald this is the reason why the icons e.g. in checkbox have w and h 24. The default overrides.

ti10le: if width and height is outside the css prop it works

ti10le: https://github.com/marigold-ui/marigold/pull/1935

PR #1925, opened 2022-03-28
feat(comp): Rewrite Columns and remove Column

ti10le:
#1860


.changeset/tidy-eggs-try.md
```diff
@@ -0,0 +1,6 @@
+---
+"@marigold/docs": major
```
sebald, line 2: If we already do this, the next release will be a major. Ist das okay?

ti10le: Oh yes sure. I thought just about that is a breaking change.

sebald: Until we make a 1.0 it's fine to do a minor. We did this until now 🤠

packages/components/src/Columns/Columns.tsx
```diff
@@ -1,69 +1,55 @@
...
-              <Box css={{ p: `${spaceValue / 2}px` }}>
-                {child.props.children}
-              </Box>
-            );
-          }
-        )}
-      </Box>
+    <Box
+      display="flex"
+      __baseCSS={{
```
sebald, line 41: put everything into `css` or `__baseCSS`. makes it hard to read when it is spread over 3 options

sebald (changes requested):
great work, just some minor fixes needed :)

sebald (approved):

PR #1929, opened 2022-03-30
feat: Introduce redone <Field>

sebald:
- Field now is only a "layout" helper to construct different types of fields.


packages/components/src/Field-Next/HelpText.tsx
```diff
@@ -40,10 +40,16 @@ export const HelpText = ({
   const iconSize = get('helpText.icon.size');
   const isErrorMessage = errorMessage && error;
 
+  let variant = disabled
```
ti10le, line 43: why not const?

packages/icons/src/info/Exclamation.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
-import { SVG } from '@marigold/system';
+import { SVG, SVGProps } from '@marigold/system';
 
-export const Exclamation = ({ className = '', ...props }) => (
-  <SVG className={className} {...props}>
+export const Exclamation = (props: SVGProps) => (
```
ti10le, line 4: Should we do this in all Icons? Not in this PR but good point.

packages/components/src/Field-Next/HelpText.tsx
```diff
@@ -19,6 +17,7 @@ export interface HelpTextThemeExtension<Value> {
 // Props
 // ---------------
 export interface HelpTextProps {
+  disabled?: boolean;
```
ti10le, line 20: New prop should be also at Helptext.stories file

packages/components/src/Field-Next/Field.test.tsx
```diff
@@ -0,0 +1,85 @@
+import React, { ReactNode } from 'react';
+import { AriaTextFieldOptions, useTextField } from '@react-aria/textfield';
+import { render, screen, within } from '@testing-library/react';
+
+import { Field } from './Field';
+
+// Setup
+// ---------------
+interface MockedTestFieldProps extends AriaTextFieldOptions<'input'> {
+  disabled?: boolean;
```
ti10le, line 10: Whats your plan about this prop? Does the user have to give the disabled prop to the Field + to the Textinput e.g.? Or give the disabled prop only to Field and from there to his children?

ti10le (changes requested):

packages/icons/src/info/Exclamation.tsx
sebald: Guess so :)

packages/components/src/Field-Next/Field.test.tsx
sebald: It's only to style it differently. Can be passed by whoever uses <Field>

ti10le: Yes but maybe its confusing API to give disabled to Field and to Textinput, or not?

sebald: ah, sorry, I thought that comment was about `<HelpText>` this will be passed down to the `<input>`

ti10le (approved):

PR #1937, opened 2022-03-31
refa: Redo `<Field>` now called `<TextField>`

sebald:
- correctly passes down `type`
- supports help text and error message
- supports validation state
- correct prop types
- supports current styling


sebald:
Will replace the old Field with this in an extra PR after this gets merged :)

packages/components/src/Field-Next/TextField.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+        'time',
+        'time',
+        'url',
+        'week',
+      ],
+      defaultValue: 'text',
+      table: {
+        defaultValue: {
+          summary: 'text',
+        },
```
ti10le, line 77: I would add defaultValue prop also to storybook

ti10le (changes requested):

packages/components/src/Field-Next/Label.tsx
```diff
@@ -36,15 +36,15 @@ export const Label: React.FC<LabelProps> = ({
     <Box
       {...props}
       as={as}
-      variant="label"
+      variant="label.above"
```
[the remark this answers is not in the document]
sebald: Yes, because it's not "above" but a label (unlike what we use in checkbox. We use the element there but only because we need the functionality) and as sad in the PR description: it's temporary so that they new one looks like the old one

packages/components/src/Field-Next/TextField.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { Box } from '@marigold/system';
+import { ComponentProps } from '@marigold/types';
+
+import { Field, FieldProps } from './Field';
+
+export interface TextFieldProps
+  extends Omit<
+      ComponentProps<'input'>,
+      'value' | 'defaultValue' | 'onChange' | 'onFocus' | 'onBlur'
```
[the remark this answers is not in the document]
sebald: See comment :)

packages/components/src/Field-Next/TextField.stories.tsx
sebald: Hm, why? Does storybook reload when pros change?

ti10le: no just to test the prop in storybook

sebald: then a `defaultValue` shouldn't  work since it's only applied on first render

ti10le: ok

ti10le (approved):

Issue #1888, opened 2022-03-23
📄 Marigold Docs 2.0


ti10le:
## Description

We wanna do some work to improve our documentation. This is a list with our todos to get a better doc.

### Done

- [x] Update components docs to adhere the template structure (#1887) #1939 #1966 #1967 #1968
- [x] improve DoAndDont component with our new Layout components and make it more responsive. #1936
- [x] #2234
- [x] rewrite our colors in docs theme. Maybe use the picture below as a base.
- [x] rewrite the Navigation that its visually better und separate components (Form Components?, Layout, ..) Maybe two or three categories (think also about this #1805)
- [x] ~Callout component for Infos/Warning/Hints/...~ #2232  used our `<Message>`component instead
- [x] #2198 implement #2245 instead
- [x] #2211 
- [x] style mdx content #2219 
- [x] Show correct version of docs #2227 
- [x] #2229
- [x] #2250
- [x] Create new Marigold Logo that is easier to use in responsive designs
- [x] #2254
- [x] #2257
- [x] ~`<Select>` does not work in nextjs~ (doesn't work in mdx for reasons, elsewhere it's fine 🤷‍♂️)
- [x] #2293
- [x] Deploy NextJS App

### Components Check

Check component pages and show demos

- [x] Aside #2296 
- [x] Aspect #2296 
- [x] Badge
- [x] Box #2308 
- [x] Breakout #2309
- [x] Button
- [x] Card
- [x] Center
- [x] Checkbox
- [x] Columns
- [x] Container
- [x] Content
- [x] Dialog
- [x] Divider
- [x] Footer
- [x] Header
- [x] Headline
- [x] Icon
- [x] Image
- [x] Inline
- [x] Input
- [x] Label
- [x] Link
- [x] List
- [x] Menu
- [x] Message
- [x] NumberField
- [x] Provider
- [x] Radio
- [x] Select
- [x] Slider
- [x] Split
- [x] Stack
- [x] Switch
- [x] Table
- [x] Text
- [x] TextArea
- [x] TextField
- [x] Tiles
- [x] Tooltip
- [x] VisuallyHidden

## Context & Consequences

Have a better Marigold documentation to imagine our users and avoid a high asking rate.

sebald:
Migrate to next? 🥸

ti10le:
Do we have the time? Mh it would be cool..

sebald:
[...]

PR #1938, opened 2022-04-01
feat(docs): docs table component for mdx pages

ti10le:
#1888


sebald:
I am not sure we should change the mdx table or we should make a dedicated props table. How do we make this responsive? 🤔

ti10le:
> I am not sure we should change the mdx table or we should make a dedicated props table. How do we make this responsive? 🤔

I think responsiveness could be a problem in both options? I think the best way is to build a responsive and accessible Marigold Table and use them as PropsTable.
You mean to create a fix PropsTable and use it as a component instead of writing with mdx?

Issue #1923, opened 2022-03-28
Themes 1.0


sebald:
## Description

We need to refactor our themes!

**Requirements**

- Support single and multiple variants of components
- Support components that consist of more than one part (see https://developer.mozilla.org/en-US/docs/Web/CSS/::part)
- Support base styles for components without repeating
- Support state (because we can not use things like `:disabled`)
- Support sizes in a common way (currently only available to button)

**Tasks**

- [x] Add theme (variant) debugging #1684 
- [x] "Translator" that helps `theme-ui` select stuff from the theme #1972 
- [x] Parse states to pesudo-classes #1979

**Theme Spec**

```js
const theme = {
  ...tokens,
  root: {},
  components: {
    Button: {
      base: {
        // styles that are applied every time
      },
      size: {
        small: {},
        medium: {},
        large: {},
      },
      variant: {
        primary: {},
        secondary: {},
      },
    },
    Checkbox: {
      base: {
        // Component consists of multiple parts
        container: {},
        icon: {},
        label: {},
      },
      size: {
        small: {
          container: {},
          icon: {},
          label: {},
        },
        medium: {
          container: {},
          icon: {},
          label: {},
        },
        large: {
          container: {},
          icon: {},
          label: {},
        },
      }
    },
  },
};
```

## Context

Our themes became very large and hard to maintain, we need more structure!

## Consequences

More to write, but maintainable 🥳

sebald:
@ti10le we should gather "semantic colors" (error, info, warning, ...) and other value we want to support in `createTheme`. Do you want to make a list?

ti10le:
> @ti10le we should gather "semantic colors" (error, info, warning, ...) and other value we want to support in `createTheme`. Do you want to make a list?

Yes I can! But what other values do you mean? Do you mean some values we already have (space, sizes) or more values like from open-props: gradient, animation or s.e.?

sebald:
@ti10le fonts for example, not sure if there are any others. but the colors are the most important ones, i guess.

ti10le:
```tsx
colors: {
	text: ...,
    background: ...,
    primary: ...,
    secondary: ...,
    disabled: ...,
    error: ...,
    warning: ...,
    info: ...,
    success: ...,
	white: // ? like in docs theme,
	black: // ? like in docs theme,
}
fonts: {
	body: ...,
	heading: ...,
}
```

for the future maybe something like this:
```tsx
// brand colors
[...]

PR #1972, opened 2022-04-07
feat: introduce `useComponentStyles` hook

sebald:
Ref #1923


packages/system/src/useComponentStyles.ts
```diff
@@ -0,0 +1,131 @@
...
+  options: any = {}
+) {
+  const { theme } = useTheme();
+  const componentStyles = get(theme, `components.${componentName}`);
+
+  // Store styles in ref to prevent re-computation
+  const stylesRef = useRef({});
+
+  if (componentStyles) {
+    const base = componentStyles.base || {};
```
ti10le, line 108: Coverage alarm line 108: I think because theres no test which tests if theres no base {}

ti10le (approved):

PR #1979, opened 2022-04-11
feat: transform state to corresponding css selector and data-attr

sebald:


packages/system/src/hooks/useStateProps.ts
```diff
@@ -0,0 +1,45 @@
...
+  [key in StateAttrsKey]?: '';
+};
+
+/**
+ * Given a map of states (e.g. `{ hover: true, focus: false }`) returns an
+ * object that can be used to set state props on a component
+ * (e.g. `[data-hover]` and `[data-focus]`).
+ */
+export const useStateProps = (
+  states: { [key in ComponentState]?: boolean } = {}
```
ti10le, line 27: coverage has pointed out that there is no test when {} is specified without states. So the default. Then I asked myself, should I just add the test or check in the hook if states are specified and if not show an error message? Or why should you use the hook without specifying states? What do you mean?

packages/system/src/components/Box/utils.test.ts
```diff
@@ -0,0 +1,121 @@
...
+          },
+          "color": "#1864ab",
+        },
+        "padding": 8,
+      }
+    `);
+  });
+});
+
+// -> add to box and test it there
```
ti10le, line 121: remove this comment?

ti10le (changes requested):

packages/system/src/components/Box/utils.test.ts
sebald: Yes 😅

packages/system/src/hooks/useStateProps.ts
ti10le: But then fix coverage with a test for this, right? 😊

ti10le: hm.. still saying line 27 is not tested. Whats the problem?

ti10le (approved):

Issue #1338, opened 2021-09-28
Component - Table


ti10le:
## Description

Actually we have no component to create a table.

[Figma Component Table Design](https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMarigold?node-id=2387%3A1928)
[Figma Docs Table Design](https://www.figma.com/file/TIRxkIDpdqb52J52PldLxl/Guidelines?node-id=452%3A4856)
In the figma design we can see that there are two kinds of Headers and Cells. In the first version we implement the Component without the sortable options.

Try to implement react-aria function from [useTable](https://react-spectrum.adobe.com/react-aria/useTable.html)

We create a few components to deliver all use cases for table usage:
- [x] `<Table>`
- [x] `<TableHead>`
- [x] `<TableBody>`
- [x] `<TableRow>`
- [x] `<TableCell>`

Component todos:
- [x] new Table folder in `@packages/components`
- [x] write tests for all new components
- [x] create storybook file
- [x] create docs mdx file
- [x] add variant styles to all three themes

## Context

With this new component we can create tables with marigold.

## Consequences

Easier to create a table with marigold. If you need a table you can use this component instead of plain html element or "Box as".

First practical usage could be in our marigold/docs components at mdx.tsx instead of `<Box as="table">`

> In seperate Issue after closing this issue.

PR #1981, opened 2022-04-11
feat(comp): introduce Table component v1

ti10le:
#1338


sebald:
Can we just use the other PR? It isn't that much more code 😀

PR #2016, opened 2022-04-20
refa: <Checkbox> and <CheckboxGroup>

sebald:


sebald:
Heads up: I have no clue why but in our docs the focus will be shown ... this doesn't happen in Storybook though.

sarahgm (approved):

PR #2220, opened 2022-07-22
Next docu code editor

benediktgrether:


sebald:
@benediktgrether can you merge main into the branch. I think this is forked from when the feature branch was broken. For example the "storbook-static" folder should not be there 😄

PR #2227, opened 2022-07-26
docs: display version correctly

sarahgm:
display the current version number from package.json in docs.


sebald:
@sarahgm lockfile :D

PR #2301, opened 2022-08-11
docs: load nav from env

sebald:


sebald:
Closed because we decided to keep `next-mdx-remote` and wait for the layout RFC to be implemented

PR #2334, opened 2022-08-19
feat: add breakpoints to all themes

### LINKED ISSUE #2331: Cannot get theme-breakpoint values in user-code
## Description

Getting the breakpoint values used by the current theme in user-code is not possible because it is `undefined`.

## How to reproduce

```
const { theme } = useTheme();
theme.breakpoints; // -> undefined
```

## Expected behavior

Breakpoints array should be non-null, even if theme does not specify breakpoints the fallback value should be returned.

sebald:
closes #2331

### DIFF
--- a/.changeset/chilly-needles-move.md
+++ b/.changeset/chilly-needles-move.md
@@ -1,0 +1,5 @@
+---
+"@marigold/components": patch
+---
+
+fix: justify content in `<Button>`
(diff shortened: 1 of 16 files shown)


michael-p:
Could we then also mark `breakpoints` as required in the `Theme` type here: https://github.com/marigold-ui/marigold/blob/main/packages/system/src/types/theme.ts#L139 ? If yes then also remove duplicated breakpoints here: https://github.com/marigold-ui/marigold/blob/main/packages/system/src/hooks/useResponsiveValue.ts#L24

sarahgm (approved):

PR #2397, opened 2022-09-06
feat: toc intersection observer

benediktgrether:


sebald:
@benediktgrether awesome job 🥳 I have one suggestion to make it even more awesome: if the user scrolled all the way down to the buttom, can we make sure the last section is always highlighted? (e.g. "if scroll === bottom of page -> select last item)

Screen Shot 2022-09-06 at 13 27 09

benediktgrether:
@sebald can we have a look tomorrow on this task? Because i found two issues.

sebald (changes requested):
I found a little thing 🙈 there is no section highlighted initially :D

docs/src/components/Toc.tsx
```diff
@@ -2,25 +2,90 @@ import { useEffect, useRef, useState } from 'react';
...
+      ) as any;
+      const elementsLenght = elements.length;
+      const activeElement =
+        elements[elementsLenght - 2]?.dataset.intersectionActive;
+      if (
+        window.innerHeight + window.scrollY >= document.body.scrollHeight &&
+        activeElement === 'true' &&
+        isBottom.current === false
+      ) {
+        elements[elementsLenght - 2].style.fontWeight = '400';
```
sebald, line 74: Also: can we add a class or `[data-attribute]` and do the styling inside Toc!? :)

sebald (approved):

PR #2398, opened 2022-09-06
fix: add title to button in navigation for a11y

benediktgrether:


docs/src/components/Navigation/NavigationLayout.tsx
```diff
@@ -24,6 +24,7 @@ export const NavigationLayout = ({ navigation }: NavigationLayoutProps) => {
             <Button
               variant="navigationSmall"
               onPress={() => setShowNavigation(!showNavigation)}
+              title="navigation"
```
sebald, line 27: should this be an `aria-label` 🤔

benediktgrether: oh you are right. I will add aria-label
image

sebald (approved):

PR #2418, opened 2022-09-09
chore: made container more responsive

sarahgm:


packages/components/src/Container/Container.tsx
```diff
@@ -9,26 +9,29 @@ export interface ContainerProps extends ComponentProps<'div'> {
...
-  alignItems?: 'left' | 'right' | 'center';
+  alignItems?: 'left' | 'right' | 'center' | 'none';
 }
 
+// for the case that tables were used we needed to set the align-items to unset(it's now default behavior)
 const ALIGN_ITEMS = {
   left: 'start',
   center: 'center',
   right: 'end',
+  none: 'unset',
```
sebald, line 20: Use `undefined` instead, so the CSS prop is not applied? :)

packages/components/src/Container/Container.tsx
```diff
@@ -9,26 +9,29 @@ export interface ContainerProps extends ComponentProps<'div'> {
   contentType?: 'content' | 'header';
   size?: keyof typeof tokenSize.content | keyof typeof tokenSize.header;
   align?: 'left' | 'right' | 'center';
-  alignItems?: 'left' | 'right' | 'center';
+  alignItems?: 'left' | 'right' | 'center' | 'none';
 }
 
+// for the case that tables were used we needed to set the align-items to unset(it's now default behavior)
```
sebald, line 15: Instead of tables -> elements with overflow

sebald (changes requested):

PR #2463, opened 2022-09-15
refa: Improve behavior of `<Stack>` and remove option to render it as a list (use `<List>` instead)

sebald:


sebald:
@lauraNawrotzki @michael-p this should make your life much easier 🙃🤞

sarahgm (approved):

PR #2465, opened 2022-09-16
docs: refactor columns

benediktgrether:


docs/src/pages/components/columns.mdx
```diff
@@ -3,15 +3,27 @@ title: Columns
 group: Layout
 github: https://github.com/marigold-ui/marigold/blob/main/packages/components/src/Columns/Columns.tsx
 edit: https://github.com/marigold-ui/marigold/blob/main/docs/src/pages/components/columns.mdx
+caption: Component with sized columns in one row.
 ---
 
-Columns component with sized columns in one row. Add an array of numbers to set the size of the children. The columns array length and the count of children must be the same. You can add space between the columns und set a collapseAt prop to collapse the columns at a certain width.
+The `<Columns>` is a responsive layout component, with sized columns in one row.
+Add an array of numbers to set the size of the children. The columns array length and the count of children must be the same. You can add space between the columns und set a collapseAt prop to collapse the columns at a certain width.
```
sarahgm, line 10: maybe "with the `columns` property you can set with an array of numbers the size of the children" ? collapseAt also in the `` so it is clear that this is also code?

docs/src/pages/components/columns.mdx
```diff
@@ -41,28 +53,24 @@ Columns component with sized columns in one row. Add an array of numbers to set
 
 ### Columns and a space between them
 
+The example shows how the `space` prop is to use. It defines the gap between the contents.
```
sarahgm, line 56: I would write it: The example shows how to use the `space` prop.

sarahgm (approved):

PR #2487, opened 2022-09-19
docs: headline docs

sarahgm:


docs/src/pages/components/headline.mdx
```diff
@@ -5,19 +5,27 @@ figma: https://www.figma.com/file/DFKyTGHAoDxOsUBPszLLxP/%F0%9F%8F%B5%EF%B8%8FMa
 github: https://github.com/marigold-ui/marigold/blob/main/packages/components/src/Headline/Headline.tsx
 edit: https://github.com/marigold-ui/marigold/blob/main/docs/src/pages/components/headline.mdx
 switchTheme: true
+caption: Component for headlines.
 ---
 
+The main task of the `<Headline>` component is tot is to serve as a title of a content section.
+It should be used when you want to structure different parts of informations. That could be the page title, text headings or the title of different content parts.
+
+The component has a `level` property which behaves similar to the HTML Headings. There you can select a level from 1-6. Alternative you can set this with the `size` prop.
```
sebald, line 14: This is actually a feature for SEO. Rendering an `h5` that looks like an H2 for example :)

sebald (changes requested):
We should explain why level and size are there: to separate rendered Header level from how it looks to the user

docs/src/pages/components/headline.mdx
sarahgm: I changed it. Can you have a look?

sebald (approved):

PR #2507, opened 2022-09-22
docs: menu docs

sarahgm:


docs/src/pages/components/menu.mdx
```diff
@@ -47,12 +62,16 @@ The `Trigger`has also a prop for `disabled`
...
 
+In this example you can see a simple `<Menu>` with some items to select.
+
 ```tsx preview file=components/Menu/menu-selected.demo.tsx
 
 ```
 
 ### Disabled Menu
 
+In this example the `<Menu.Trigger>` has set its prop `disabled`. So you can't interact with the `<Menu>` anymore.
```
benediktgrether, line 73: If the Menu is disabled then we should change the cursor to not-allowed? `cursor: not-allowed;`

benediktgrether (changes requested):

docs/src/pages/components/menu.mdx
sarahgm: hm i don't know. I also have the question in my mind: why would i disable a complete menu? :D

benediktgrether (approved):

PR #2568, opened 2022-10-17
feat: improve headline with align and color property

sarahgm:


sarahgm:
why can't I add a changeset 🤔

sebald (changes requested):
Tests? 😬

sebald:
> why can't I add a changeset 🤔

😵‍💫 You can do it via CLI too "pnpm changeset add" I think

sebald (approved):

PR #2576, opened 2022-10-20
feat: core label on the left side

sarahgm:


.changeset/metal-insects-decide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@marigold/docs": patch
+"@marigold/components": patch
```
sebald, line 3: Minor! Since this is a new feature ;)

sebald (changes requested):
The implementation looks great. Can we maybe change the API? What about extending the label prop? 

label = ReactNode | { content: ReactNode, side: ... }

.changeset/metal-insects-decide.md
sarahgm: done

sarahgm:
> 

How would it work with that? I don't understand

sebald:
> How would it work with that? I don't understand @sebald

@sarahgm 

1. Extend the `FieldBaseProps` interface, instead of `label?: ReactNode` you have something like `label?: ReactNode | { content: ReactNode, side: 'left' | 'right' | 'top' }`
2. Inside the component check the label prop. If it is a ReactNode, just it as label, otherwhise use `label.content` as label (you could also switch that and check if `label.content` exists)
3. Then you can set them like you already did :) 

Why might this be better? Because the props are always together:

```jsx
    <TextField
      label="Promo Code"
      description="You'll find your promo code on the back of your ticket."
      errorMessage="Your promo code is not valid! Please review your input."
      value={value}
      onChange={setValue}
      error={error}
      inputMode="numeric"
      pattern="[0-9]*"
      side="left"
    />
```

VS

```jsx
    <TextField
      label={ 
          content: "Promo Code", 
          side: "left" 
      }
      description="You'll find your promo code on the back of your ticket."
      errorMessage="Your promo code is not valid! Please review your input."
      value={value}
      onChange={setValue}
      error={error}
      inputMode="numeric"
      pattern="[0-9]*"
    />
 ```

**Comment by sarahgm** [comment]
> 

@sebald  can we have a look together monday? i have tried something

**Comment by sarahgm** [comment]
if I type it as string it works for FieldBase but not for the others, with ReactNode its a typescript problem but I don't know how to solve it.

File: docs/src/pages/components/number-field.mdx [hunk 2]
```diff
@@ -137,6 +137,12 @@ import { NumberField } from '@marigold/components';
       description: 'Sets the amount of steps the input should take.',
       default: '1',
     },
+    {
+      property: 'side',
```
sebald, line 141: This prop does not exist anymore right? 😃

sebald (changes requested):
I think there are some old props in there 😃

sarahgm:
> I think there are some old props in there 😃

sorry, not anymore 😬

sebald (approved):

Issue #2537, opened 2022-09-28
Introduce a layout component to constrain height/width


sebald:
## Description

We got layout components for all major use cases and can construct whole pages with them. One piece that is used from time to time in our applications and is currently has to be done in user land is limiting the size of a certain area in the layout. For example a table should not be more than 500px high or a certain area can take between 200px and 500px.

This would allow for a very fine grained control of sizing if needed.

## Proposal

Introduce a layout component that allows to specify min and max heights/widths -> clamping the size. E.g.

```jsx
<Clamp width={["50vw", "300px"]} >....</Clamp>
<Clamp height={["25vw", "1000px"]} >....</Clamp>
<Clamp width={["50vw", "300px"]} height={["25vw", "1000px"]>....</Clamp>
```

In this case the element will be 300px at most, and less if the viewport is smaller than 600px, but at least 50% of the viewport (because of the `50vw`). This is basically an API for [min](https://developer.mozilla.org/en-US/docs/Web/CSS/min).

Also, since most of our layout components let the user specify gab and alignment this should maybe too? We have to check if this would make sense, since you can put a `<Stack>` or a `<Inline>` as children for this component and have the same effect.

**Props**

- `width`: [string, string]
- `height`: [string, string]

Maybe:
- `space`: responsive string
- `alignX`: left, center, right
- `alignY`: top, center, bottom

Implementing this with as a grid would also allow our `Breakout` component to work with `Clamp`. `Clamp` is also very similar to `Container`we have to make sure there is a clear distinction between the two.

## Previous Work

- https://lyts.christiankaindl.com/components/clamp

sebald:
difference between `Container` and `Clamp`, the first is for content and has a `contentType` prop to suggest that. The later is for arbitrary elements.

sarahgm:
@sebald  so is this still relevant?

sebald:
Maaaybe

PR #2584, opened 2022-10-21
docs: Add some more complex dialog examples

sebald:
Co-authored-by: sarahgm <sarahgm@users.noreply.github.com>

### DIFF
--- a/packages/components/src/Dialog/Dialog.stories.tsx
+++ b/packages/components/src/Dialog/Dialog.stories.tsx
@@ -8,6 +8,12 @@ import { Text } from '../Text';
 import { Dialog } from './Dialog';
 import { TextField } from '../TextField';
 import { Stack } from '../Stack';
+import { Body } from '../Body';
+import { Container } from '../Container';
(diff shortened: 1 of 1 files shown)


packages/components/src/Dialog/Dialog.stories.tsx
```diff
@@ -66,16 +72,89 @@ export const Form: ComponentStory<typeof Dialog.Trigger> = args => {
...
+        </Body>
+        <Footer>
+          <Button variant="primary">Ok</Button>
+        </Footer>
+      </Container>
+    </Dialog>
+  </Dialog.Trigger>
+);
+
+export const StickyFooter: ComponentStory<typeof Dialog.Trigger> = args => (
```
sarahgm, line 126: maybe we can add this in a demo too?

sebald: in the docs you mean? I would like to wait for #2537 because this should make the Box obsolete

sarahgm (approved):

PR #2585, opened 2022-10-21
fix: remove duplicated key in `Card` styles

sebald:


themes/theme-core/src/components/Card.style.ts
```diff
@@ -13,8 +13,6 @@ export const Card: Theme['components']['Card'] = {
     small: {
       minHeight: '100px',
     },
-  },
-  size: {
     medium: {
       height: 100,
```
sarahgm, line 17: remove the complete size medium

themes/theme-core/src/components/Card.style.ts
```diff
@@ -13,8 +13,6 @@ export const Card: Theme['components']['Card'] = {
...
     },
```
sarahgm, line 18:
```suggestion

```

sarahgm (approved):

PR #2683, opened 2022-12-13
feat: add <Tray> for responsiveness in <Menu> and <Select>

sarahgm:


packages/components/src/theme.ts
```diff
@@ -35,6 +35,7 @@ import type { SwitchThemeExtension } from './Switch';
 import type { TableThemeExtension } from './Table';
 import type { TextThemeExtension } from './Text';
 import type { TextAreaThemeExtension } from './TextArea';
+import type { TrayThemeExtension } from './Overlay';
```
sebald, line 38: Could add this to the overlay above :)

packages/components/src/Menu/MenuTrigger.tsx
```diff
@@ -5,6 +5,8 @@ import { MenuContext, MenuContextProps } from './Context';
 import { PressResponder } from '@react-aria/interactions';
 import { Popover } from '../Overlay/Popover';
 import { useObjectRef } from '@react-aria/utils';
+import { useResponsiveValue } from '@marigold/system';
+import { Tray } from '../Overlay';
```
sebald, line 9: import with Popover? :)

themes/theme-b2b/src/components/Tray.style.ts
```diff
@@ -0,0 +1,9 @@
+import { Theme } from '@marigold/components';
+
+export const Tray: Theme['components']['Tray'] = {
+  base: {
```
sebald, line 4: If we have to add this to every theme, maybe we should add this as `__baseCSS` to the `Tray`?

sarahgm: yes, so should we keep the possibility to style or remove the themeextension possibility?

sebald (approved):

PR #2734, opened 2023-01-30
add new semantic colors

benediktgrether:


themes/theme-b2b/src/design-token.ts
```diff
@@ -0,0 +1,75 @@
...
+    background: themeColors.primary,
+    text: themeColors.background,
+    hover: colors.orange40,
+  },
+};
+
+export const secondaryColors = {
+  secondary: {
+    background: themeColors.secondary,
+    text: themeColors.background,
```
sarahgm, line 33: why background? shouldn't that be text

sarahgm:
I don't understand why it's sometimes `themeColors.something` and sometimes `colors.something`

sebald:
Superseeded by the move to tailwind

PR #2758, opened 2023-02-08
feat: add default color to body

### LINKED ISSUE #2757: Add default text color in b2b theme
## Description

during implementing the SBP we noticed that the default global Text has no color. We should add it in the `index.ts` in the root/body object.

## Proposal

`color: text`

sebald:
closes #2757

### DIFF
--- a/.changeset/plenty-penguins-drum.md
+++ b/.changeset/plenty-penguins-drum.md
@@ -1,0 +1,5 @@
+---
+"@marigold/theme-b2b": minor
+---
+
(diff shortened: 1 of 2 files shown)


sebald:
@sarahgm I think this changed the whole snapshots 😄

sarahgm (approved):

sarahgm:
> @sarahgm I think this changed the whole snapshots 😄

@sebald  yes but I think Chromatic is still broken :(

PR #2759, opened 2023-02-09
feat: update Menu.Trigger with open and onOpenChange property

sarahgm:
- add demo in documentation

### DIFF
--- a/packages/components/src/Menu/MenuTrigger.tsx
+++ b/packages/components/src/Menu/MenuTrigger.tsx
@@ -10,15 +10,25 @@ import { Popover, Tray } from '../Overlay';
 export interface MenuTriggerProps {
   children: [trigger: ReactNode, menu: ReactNode];
   disabled?: boolean;
+  open?: boolean;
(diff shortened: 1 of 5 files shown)


packages/components/src/Menu/MenuTrigger.tsx
```diff
@@ -10,15 +10,25 @@ import { Popover, Tray } from '../Overlay';
 export interface MenuTriggerProps {
   children: [trigger: ReactNode, menu: ReactNode];
   disabled?: boolean;
+  open?: boolean;
+  onOpenChange?: (isOpen: boolean) => void;
 }
 
-export const MenuTrigger = ({ disabled, children }: MenuTriggerProps) => {
```
sebald, line 15: Tests are missing

sebald (changes requested):

packages/components/src/Menu/MenuTrigger.tsx
sarahgm: added two tests

sebald (approved):

PR #2764, opened 2023-02-13
refa: fix polymorph types + remove style props from `<Box>`

### LINKED ISSUE #2747: Deprecated Style Props
## Description

Our `<Box>` component supports so called "style props", which is a shorthand for using the `css` prop. This means the following is equivalent:

```jsx
<Box color="red">...</Box>
```

```jsx
<Box css={{ color: 'red' }}>...</Box>
```

This is a bit confusing and already made some application code harder to read and comprehend since both options were used by the developer.

## Proposal

We already talked about this in the past, we should remove the style props, since they are just an additional API with not enough benefits.

> __Warning__
> **This is a breaking change!**

### LINKED ISSUE #2751: Polymorphism doesn't work on `<Box>` if no `as` is specified
## Description

Looks like our `PolimorphicComponent` type broke / is not 100% accurate.

## How to reproduce

```jsx
// Does not throw a type error
const A = () => <Box foo="bar">Test</Box>

//Throw a type error
const B  = () => <Box as="div" foo="bar">Test</Box>
```

## Expected behavior

Both of the example should behave the same and throw a type error.

sebald:
closes #2751
closes #2747


docs/src/demos/components/Breakout/text-breakout.demo.tsx
```diff
@@ -2,7 +2,7 @@ import { Container, Breakout, Text } from '@marigold/components';
 
 export const TextBreakout = () => (
```
sarahgm, line 3: @sebald I can't find where this is used :S

sarahgm:
@sebald  in Tiles.mdx the preview is not showing

packages/tokens/src/shadow.stories.tsx
```diff
@@ -22,7 +22,12 @@ export const Shadows = () => {
 
   const shadow = Token.shadow;
```
sarahgm, line 23: @sebald  where do I find this in storybook?

sarahgm (approved):

PR #2873, opened 2023-03-20
Adding Tabs component

OsamaAbdellateef:


packages/components/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,52 @@
...
+                </Box>
+                <TabPanel key={state.selectedItem?.key} state={state} />
+            </Box >
+        </TabContext.Provider>
+    )
+}
+
+Tabs.Item = Item
+
+interface Tabs {
```
sarahgm, line 48: I think you don't need this interface

OsamaAbdellateef: You are right, i just did that to add a new prop on Item component, but then i found out that there's no need for this prop so i deleted it and forgot to delete the interface .

sebald:
@OsamaAbdellateef did you rebase the branch? It has a lot of changes that are not related to your work 🤔

OsamaAbdellateef:
@sebald  I don't think so

sebald (changes requested):
There are changes that are already on merges, also the changeset markdown files shouldn't be there since they were in the latest, release. I have a hard time to revert/fix these things, maybe it is easier to copy/paste our changes and additions from the tabs to a new clean branch?

OsamaAbdellateef:
> There are changes that are already on merges, also the changeset markdown files shouldn't be there since they were in the latest, release. I have a hard time to revert/fix these things, maybe it is easier to copy/paste our changes and additions from the tabs to a new clean branch?
I have just created a new branch `components/new-tabs` of main and added Tabs stuff and i'll create a new PR for this .

PR #2912, opened 2023-04-03
Move Tabs on a new branch .

OsamaAbdellateef:
I just moved the `Tabs` on a new **branch** , nothing changed .


packages/components/src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const disabled = tabProps['aria-disabled'];
+  const { hoverProps, isHovered } = useHover({
+    isDisabled: disabled as boolean,
+  });
+  const { focusProps } = useFocus({});
+  const stateProps = useStateProps({ active: isSelected, hover: isHovered });
+  const { styles } = useTabContext();
+  return (
+    <Box
+      className={disabled ? 'disabled' : ''}
```
sebald, line 29: we don't use `className`s :) we have the `stateProps` for this

packages/components/src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,43 @@
...
+export interface TabProps {
+  item: Node<object>;
+  state: TabListState<object>;
+}
+
+export const Tab = ({ item, state }: TabProps) => {
+  const { key, rendered } = item;
+  const ref = useRef(null);
+  const { tabProps, isSelected } = useTab({ key }, state, ref);
+  const disabled = tabProps['aria-disabled'];
```
sebald, line 20: Rather use the tab context du pass down the disabled props from the `Tabs` :)

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Tab } from './Tab';
+import { TabPanel } from './TabPanel';
+import { TabContext } from './Context';
+import { Item } from '@react-stately/collections';
+
+export interface TabsThemeExtension
+  extends ThemeExtensionsWithParts<'Tabs', ['tabs', 'tab', 'tabPanel']> {}
+export interface TabsProps
+  extends Omit<AriaTabListProps<object>, 'orientation'> {
+  gap?: number;
```
sebald, line 21: This doesn't need to be expoed here, a `variant` in the `theme` can override this.

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+import { Tab } from './Tab';
+import { TabPanel } from './TabPanel';
+import { TabContext } from './Context';
+import { Item } from '@react-stately/collections';
+
+export interface TabsThemeExtension
+  extends ThemeExtensionsWithParts<'Tabs', ['tabs', 'tab', 'tabPanel']> {}
+export interface TabsProps
+  extends Omit<AriaTabListProps<object>, 'orientation'> {
```
sebald, line 20: `react-aria` uses a lot of `isDisabled` props (basically they all start with "is"), we use the HTML names. So instead of `isDisabled` the component should accept `disabled`. And ofc it then needs to be cast inside the component :D

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const ref = useRef(null);
+  const { tabListProps } = useTabList(props, state, ref);
+  const styles = useComponentStyles(
+    'Tabs',
+    { size },
+    { parts: ['tabs', 'tab', 'tabPanel'] }
+  );
+  return (
+    <TabContext.Provider value={{ styles }}>
+      <Box>
```
sebald, line 35: If you don't add any props, rather use a regular `div`

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,52 @@
...
+          css={styles.tabs}
+          __baseCSS={{ display: 'flex', gap: `${gap}rem` }}
+          {...tabListProps}
+          ref={ref}
+        >
+          {[...state.collection].map(item => {
+            return <Tab key={item.key} item={item} state={state} />;
+          })}
+        </Box>
+        <TabPanel key={state.selectedItem?.key} state={state} />
```
sebald, line 46: Can you help me? What does this do?

packages/components/src/Tabs/Tab.tsx
OsamaAbdellateef: that is right, 
in the styling i am using `aria-disable`, i just forgot to delete the className

packages/components/src/Tabs/Tabs.tsx
OsamaAbdellateef: yes this will be better and more convenient, 
now it accepts disabled

packages/components/src/Tabs/Tabs.tsx
OsamaAbdellateef: As you know when we click on `Tab` we will get the content "`TabPanel`" which is connected to this `Tab` we clicked.
Every `Tab` has a **key**, so when you click on `Tab` , the **key** is set into `state.selectedItem` and we use this to show the active `TabPanel`

packages/components/src/Tabs/Tab.tsx
OsamaAbdellateef: Ok i'll try this

sebald:
@OsamaAbdellateef there are some conflicts 🙈 can you resolve these?

packages/components/src/Tabs/Tab.tsx
sebald: ping :)

OsamaAbdellateef:
> @OsamaAbdellateef there are some conflicts see_no_evil can you resolve these?

Conflicts are resolved :white_check_mark:

sebald (approved):

sebald (changes requested):
Just noticed that there is no documentation! :D

sebald (approved):

PR #2929, opened 2023-04-11
feat: Introduce <DateField> Component

OsamaAbdellateef:
``


docs/src/pages/components/date-field.mdx
```diff
@@ -0,0 +1,131 @@
+---
+title: DateField
+group: Forms
+edit: https://github.com/marigold-ui/marigold/blob/main/docs/src/pages/components/container.mdx
+caption: Component used to add date value .
+---
+
+The `<DateField>` allow users to enter and edit date values using a keyboard. Each part of a date value is displayed in an individually editable segment.
```
sarahgm, line 8: maybe a bit more description

packages/components/src/DateField/DateField.tsx
```diff
@@ -0,0 +1,179 @@
...
+        <Box
+          __baseCSS={{
+            display: 'flex',
+            alignItems: 'center',
+            flexBasis: '100%',
+            gap: '11px',
+            paddingX: '7px',
+          }}
+        >
+          {iconLeft && (
```
sarahgm, line 132: we have `icon` and `action` for leading and trailing icons. You can have a look on the Autocomplete or Input. Maybe this works too

sarahgm:
also maybe an example with calendar icon and date format as in the specification https://www.figma.com/file/HkoXCRkoUbCxRbg1CuaTDx/Design-System-Workflow?node-id=525-20351&t=SPSTTxxV8jzBAGfU-0

docs/src/pages/components/date-field.mdx
```diff
@@ -0,0 +1,131 @@
+---
+title: DateField
+group: Forms
+edit: https://github.com/marigold-ui/marigold/blob/main/docs/src/pages/components/container.mdx
```
sarahgm, line 4: the link is right now set to container.mdx ist should be to this file: .....date-field.mdx

sarahgm: also github link is missing (github link is where the component lies. edit link is where the mdx file lie.

packages/components/src/DateField/DateField.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  const ref = useRef(null);
+  const { fieldProps, labelProps, descriptionProps } = useDateField(
+    props,
+    state,
+    ref
+  );
+  const styles = useComponentStyles(
+    'DateField',
+    {},
```
sarahgm, line 87: here you should add variant and size. even if it's not used yet, but it's consistent.

sarahgm:
on story with icons calendar icon is a bit too close to the right side, maybe align this like the leading icon so it's the same

OsamaAbdellateef:
> 

Fixed

PR #2939, opened 2023-04-14
feat: add new package for creating tailwind presets

sarahgm:


sarahgm:
if i run the tests local my test all pass

jest.config.js
```diff
@@ -12,9 +12,11 @@ module.exports = createConfig({
     '!packages/components/src/Accordion/useAccordionItem.ts',
     '!packages/components/src/Accordion/Accordion.tsx',
   ],
+  modulePathIgnorePatterns: ['<rootDir>/dist/'],
   moduleNameMapper: {
     '^@marigold/theme-(.+)$': '<rootDir>/themes/theme-$1/src',
     '^@marigold/(.+)-config$': '<rootDir>/config/$1',
     '^@marigold/(.+)$': '<rootDir>/packages/$1/src',
+    '^@marigold/(.*)$': '<rootDir>/dist/myscope/$1',
```
sebald, line 20: Can we add a comment here why we do that, especially since "dist" is also on the ignore pattern 🫣

packages/theme-preset/src/create-preset.ts
```diff
@@ -0,0 +1,27 @@
...
+import { sync as findUpSync } from 'find-up';
+import type { Config, OptionalConfig } from 'tailwindcss/types/config';
+
+import { defaultTheme } from '@marigold/system';
+
+export const createPreset = (name: string, config: Partial<OptionalConfig>) => {
+  const parent = path.resolve(__dirname, '..');
+  const root = path.dirname(findUpSync('package.json', { cwd: parent }) || '.');
+
+  console.log(__dirname, root);
```
sebald, line 12: Remove log?

packages/theme-preset/src/create-preset.test.tsx
```diff
@@ -0,0 +1,38 @@
+import { defaultTheme } from '@marigold/system';
+import { createPreset } from './create-preset';
+
+test('create preset and merge it with defaultTheme', () => {
```
sebald, line 4: Maybe another tests that shows that `preset.content` finds the closest marigold path?

sebald (changes requested):

jest.config.js
sarahgm: But .... i still don't know why 😂🙈🙈🙈

packages/theme-preset/src/create-preset.test.tsx
sarahgm: Tried to but i can have a closer look

sebald (approved):

PR #2980, opened 2023-05-22
Tailwind/switch

OsamaAbdellateef:


packages/components/src/Switch/Switch.tsx
```diff
@@ -72,86 +61,52 @@ export const Switch = forwardRef<HTMLInputElement, SwitchProps>(
...
-            position: 'absolute',
-            width: '100%',
-            height: '100%',
-            top: 0,
-            left: 0,
-            zIndex: 1,
-            opacity: 0.0001,
-            cursor: inputProps.disabled ? 'not-allowed' : 'pointer',
-          }}
+          className="absolute left-0 top-0 z-[1] h-full w-full opacity-[0.0001] disabled:cursor-not-allowed"
```
sarahgm, line 82: shouldn't be here for disabled the group used? @sebald @OsamaAbdellateef ? 🤔

packages/components/src/Switch/Switch.test.tsx
```diff
@@ -1,193 +1,157 @@
...
-          container: {
-            p: 'small',
-          },
-          track: {
-            bg: 'green',
-          },
-          thumb: {
-            bg: 'hotpink',
+      container: cva(),
+      track: cva(
```
sebald, line 14: you don't need real styling here, it's just for testing! :)

sebald: but leave it, its fine! ;)

packages/components/src/Switch/Switch.tsx
sebald: makes sense, yes the input should be not visible anyway.

packages/components/src/Switch/__snapshots__/Switch.test.tsx.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
sebald, line 1: use inline snapshots please :)

themes/theme-core/src/components/Switch.styles.ts
```diff
@@ -0,0 +1,34 @@
...
+        size: {
+          large: 'w-[96] h-[48] rounded-[40]',
+        },
+      },
+    }
+  ),
+  thumb: cva(
+    ['bg-root-body', 'h-5 w-5'],
+
+    // 'shadow-[inset_1px_1px_4px] shadow-switch-thumb-shadow disabled:bg-switch-thumb-disabled',
```
sebald, line 24: remove?

themes/theme-core/src/components/Switch.styles.ts
```diff
@@ -0,0 +1,34 @@
+import { ThemeComponent } from '@marigold/system';
+import { cva } from 'class-variance-authority';
+
+export const Switch: ThemeComponent<'Switch'> = {
+  container: cva(),
+  track: cva(
+    [
+      'border-input-border border bg-white shadow-[inset_0px_0px_1px',
+      'checked:bg-switch-track-primary checked:shadow-switch-track-checked',
+      'disabled:opacity-[0.5] disabled:shadow-switch-track-shadow disabled:bg-switch-track-shadow focus:outline-none focus:outline-offset[3]',
```
sarahgm: should here also be used group?

OsamaAbdellateef: yes sure 
used now

themes/theme-core/src/components/Switch.styles.ts
OsamaAbdellateef: removed

sarahgm (approved):

PR #2996, opened 2023-05-30
Refactoring Tabs

OsamaAbdellateef:


OsamaAbdellateef (commented):
Reviewed

packages/components/src/Tabs/Tab.tsx
```diff
@@ -19,25 +19,22 @@ export const Tab = ({ item, state }: TabProps) => {
   const { tabProps, isSelected } = useTab({ key }, state, ref);
   const disabled = tabProps['aria-disabled'];
   const { hoverProps, isHovered } = useHover({
-    isDisabled: disabled as boolean,
+    isDisabled: (disabled as boolean) || isSelected,
```
sarahgm, line 22: @OsamaAbdellateef why also `isSelected` ?

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -1,60 +1,57 @@
...
-import { TabContext } from './Context';
 import { Item } from '@react-stately/collections';
+import { TabContext } from './Context';
 
-export interface TabsThemeExtension
-  extends ThemeExtensionsWithParts<'Tabs', ['tabs', 'tab', 'tabPanel']> {}
 interface TabsProps
   extends Omit<AriaTabListProps<object>, 'orientation' | 'isDisabled'> {
-  gap?: number;
+  gap?: string;
```
sarahgm, line 15: you can use for this also `gapSpace` which is in `style-props.tsx`

packages/components/src/Tabs/Tab.tsx
OsamaAbdellateef: i wanna disable the hover if the tab is `selected` 
i guess it shouldn't have `hover` effect when it is already `selected`

packages/components/src/Tabs/Tabs.tsx
OsamaAbdellateef: Yeah it is better, now it is added

sarahgm (approved):

PR #2998, opened 2023-05-31
tailwind: datefield

sarahgm:


packages/components/src/DateField/DateSegment.tsx
```diff
@@ -32,57 +34,38 @@ export const DateSegment = ({
...
+      </span>
+      <span
+        className={cn(
           type === 'literal'
-            ? `literal  ${!isPrevPlaceholder && 'activeLiteral'}`
-            : ''
-        }
+            ? `literal ${
+                !isPrevPlaceholder &&
+                'activeLiteral text-datefield-segment group-disabled/field:text-disabled-text'
```
OsamaAbdellateef, line 65: I don't think that we  need  `activeLiteral` anymore

themes/theme-core/src/components/DateField.styles.ts
```diff
@@ -0,0 +1,14 @@
...
+import { cva } from 'class-variance-authority';
+
+export const DateField: ThemeComponent<'DateField'> = {
+  segment: cva([
+    'text-datefield-segmentText focus-visible:bg-datefield-segment focus-visible:text-secondary-50 ',
+  ]),
+
+  field: cva([
+    'border border-solid border-border-color bg-secondary-50',
+    'group-disabled/field:cursor-not-allowed group-disabled/field:bg-disabled-bg group-disabled/field:text-disabled-text',
```
OsamaAbdellateef, line 11: `group-disabled/field:cursor-not-allowed` 
i guess this should be on the div itself not the theme file  because it will be in all themes

OsamaAbdellateef (approved):

PR #3009, opened 2023-06-02
Refactoring date picker

OsamaAbdellateef:


packages/system/src/types/theme.ts
```diff
@@ -101,6 +101,7 @@ export type Theme = {
       'calendar' | 'calendarCell' | 'calendarControllers',
       ComponentStyleFunction<string, string>
     >;
+    DatePicker?: Record<'actionButton', ComponentStyleFunction<string, string>>;
```
sarahgm, line 104: why not use without parts? :)

sarahgm:
One thing I noticed that if you try to open the Calendar in smaller view it is first thing, not responsive and the calendar is not opening

OsamaAbdellateef:
> One thing I noticed that if you try to open the Calendar in smaller view it is first thing, not responsive and the calendar is not opening

Yeah this is true , but i tried this on the  documentation and it is working :"

sarahgm:
> > One thing I noticed that if you try to open the Calendar in smaller view it is first thing, not responsive and the calendar is not opening
> 
> Yeah this is true , but i tried this on the documentation and it is working :"

What do you mean?

sarahgm (approved):

PR #3010, opened 2023-06-02
Fixing tests in tailwind

sarahgm:


packages/components/src/Card/Card.test.tsx
```diff
@@ -61,7 +58,9 @@ test('accepts a variant and size', () => {
     </ThemeProvider>
   );
   const card = screen.getByTestId('card');
-  expect(card).toHaveClass(`bg-yellow-300 p-4`);
```
sebald, line 64: if you want to only check if certain classes are set:

```
expect(card.className).toContain('...');

sebald (approved):

PR #3020, opened 2023-06-06
Adjust datefield

sebald:


sarahgm:
@sebald  typecheck and test :o

sarahgm:
@sebald  what about these lint warnings? Should this be resolved?

sebald:
> @sebald  what about these lint warnings? Should this be resolved?

Maybe we should change then to be an error?

sebald:
> @sebald yes do you know how? I can merge this than, right?

Yes.

For the lint stuff: needs to be changed in our config

sarahgm (approved):

PR #3036, opened 2023-06-12
Adding b2b `theme` to link component 

OsamaAbdellateef:


themes/theme-b2b/src/tokens.ts
```diff
@@ -32,7 +32,7 @@ export const colors = {
     primary: {
       DEFAULT: brand.secondary[700],
     },
-
+    blue: '#3ab3d5',
```
sarahgm, line 35: hm.. should we have a blue color? I'm wondering if we could have a better name for it. In core we used dark primary....

sarahgm: maybe `link` ? or we make it a lighter gray and name it `primary-light`?

OsamaAbdellateef: hmmm , for me i think `link` is better since this color isn't one of `brand` colors 
what do you think ?!

sarahgm: yes. I would name it link

OsamaAbdellateef: now switched to `link`

sarahgm (approved):

PR #3045, opened 2023-06-12
Tailwind: add radio b2b style

sarahgm:


OsamaAbdellateef:
> _ No description provided. _
@sarahgm 
the background for `disabled` state is a little bit different than the one  in the `main` 
is it ok ?!

sarahgm:
> > _ No description provided. _
> > @sarahgm
> > the background for `disabled` state is a little bit different than the one  in the `main`
> > is it ok ?!

we should have one background color so...

PR #3054, opened 2023-06-14
Tailwind: `Slider` b2b styles

OsamaAbdellateef:


themes/theme-b2b/src/components/Slider.styles.ts
```diff
@@ -3,17 +3,16 @@ import { cva } from 'class-variance-authority';
 
 export const Slider: ThemeComponent<'Slider'> = {
   track: cva([
-    'absolute top-4 h-2 w-full',
     'rounded-lg border-none border-transparent',
-    'bg-bg-surface-raised text-transparent',
+    'bg-[#e3e3e3] text-transparent',
```
OsamaAbdellateef, line 7: @sarahgm  
hint: this color is used only here

sarahgm: ok which token would you use than? Because there should be tokens used :)

sarahgm: thats why we used bg-surface-raised to have a token

sarahgm:
@OsamaAbdellateef conflicts :)

sarahgm (approved):

PR #3061, opened 2023-06-15
Tailwind: `DatePicker` b2b styles

OsamaAbdellateef:


themes/theme-b2b/src/components/DatePicker.styles.ts
```diff
@@ -0,0 +1,15 @@
+import { ThemeComponent } from '@marigold/system';
+import { cva } from 'class-variance-authority';
+
+export const DatePicker: ThemeComponent<'DatePicker'> = {
+  container: cva(),
+  button: cva([
+    ' h-full w-6',
+    'p-0',
```
sarahgm, line 8: I would leave the padding, I think it looks better, what do you think?

OsamaAbdellateef: yeah sure and we don't need it now

sarahgm (approved):

PR #3087, opened 2023-06-30
[DST-63]: Adding`useSmallScreen` hook

OsamaAbdellateef:


sebald:
@OsamaAbdellateef can you also use it in components previously used `useResponsive` hook? :)

OsamaAbdellateef:
> > @OsamaAbdellateef can you also use it in components previously used `useResponsive` hook? :)
> 
> @sebald   Done 100

sebald:
There is an issue with the docs build, guess we have to check if window exists 🙈

OsamaAbdellateef:
> There is an issue with the docs build, guess we have to check if window exists see_no_evil

@sebald  Done :ballot_box_with_check:

sebald (approved):

PR #3131, opened 2023-07-13
chore[DST-42]: added gray and white and black to b2b theme

sarahgm:


themes/theme-b2b/src/tokens.ts
```diff
@@ -27,6 +27,22 @@ const brand = {
     '950': '#030712',
   },
 
+  white: '#ffff',
```
sebald, line 30: Is this different from the tw one? Should we explicitly add these 🤔

sarahgm: what do you mean?

sarahgm: @sebald

sebald: this line basically does nothing, because `bg-white` is already there in tailwind :) So my question is if we add this explicitly.

sebald (approved):

PR #3150, opened 2023-07-18
Adding: Code hightligher & `MarigoldProvider`

OsamaAbdellateef:


docs/lib/registry.ts
```diff
@@ -0,0 +1,10 @@
+import dynamic from 'next/dynamic';
```
sarahgm, line 1: we don't need this file anymore

sarahgm:
I made a comment here: https://marigold-docs-git-code-hightligher-marigold.vercel.app/components/button

docs/contentlayer.config.ts
```diff
@@ -88,6 +94,66 @@ export default makeSource({
...
+              if (match) {
+                node.__event__ = match ? match[1] : null;
+                codeEl.data.meta = codeEl.data.meta.replace(regex, '');
+              }
+            }
+            node.__rawString__ = codeEl.children?.[0].value;
+          }
+        });
+      },
+      [rehypeAutolinkHeadings, { behavior: 'wrap' }],
```
sarahgm, line 119: overrides the current settings in main!

docs/contentlayer.config.ts
```diff
@@ -88,6 +94,66 @@ export default makeSource({
...
+      [
+        rehypePrettyCode,
+        {
+          getHighlighter: async () => {
```
sarahgm, line 123: Why did you chose a custom theme and not a prepacked one?

docs/contentlayer.config.ts
```diff
@@ -88,6 +94,66 @@ export default makeSource({
...
+            const theme = await loadTheme(
+              // loading theme file
+              path.join(process.cwd(), './light.json')
```
sarahgm, line 126: this could fail if `cwd` is not `/docs`, this is from the documentation

```js
const options = {
  theme: JSON.parse(
    readFileSync(new URL('./themes/theme.json', import.meta.url))
  ),
};
```

docs/contentlayer.config.ts
```diff
@@ -88,6 +94,66 @@ export default makeSource({
...
+            );
+            return await getHighlighter({ theme });
+          },
+          onVisitLine(node: any) {
+            if (node.children.length === 0) {
+              node.children = [{ type: 'text', value: ' ' }];
```
sarahgm, line 132: What does this do? :D

docs/registry/index.tsx
```diff
@@ -3,7 +3,22 @@
 // Do not edit this file directly.
 import dynamic from 'next/dynamic';
 
-export const registry = {
+interface Registery {
```
sarahgm, line 6: this is not coherent with the `main` branch. please use the script to update the file :)

docs/ui/ComponentDemo.tsx
```diff
@@ -1,28 +1,41 @@
...
-export const ComponentDemo = ({
-  name,
-  source,
-  children,
-}: ComponentDemoProps) => {
+export const ComponentDemo = ({ name, children }: ComponentDemoProps) => {
   const Demo = registry[name].demo;
-
+  const { current, themes } = useThemeSwitch();
+  const [showCode, setShowCode] = useState(true);
```
sarahgm, line 14: Why not use the Marigold Tabs :D

docs/ui/CopyButton.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { Button } from '@marigold/components';
+
+interface CopyProps {
+  codeString: string;
+}
+
+export const CopyButton = ({ codeString }: CopyProps) => {
+  const [isCopied, setIsCopied] = React.useState(false);
```
sarahgm, line 9: we should use a hook for this :)

docs/ui/ComponentDemo.tsx
OsamaAbdellateef: I agree this is the First option came to my mind 
but i wanted it to look like shadcn "more elegant"

docs/registry/index.tsx
OsamaAbdellateef: oh i got it

docs/ui/CopyButton.tsx
OsamaAbdellateef: because we need to change the copy icon based on if it is copy or not

docs/contentlayer.config.ts
OsamaAbdellateef: @sarahgm 
Here we check if the length of children is  0 then this means that the node has no children and then we add 
`node.children = [{ type: 'text', value: ' ' }];` to ensure that the the node has a child element t, even if it doesn't contain any content , it can be useful in cases you need to manipulate or render the HTML like syntax tree

docs/contentlayer.config.ts
OsamaAbdellateef: I don't understand what do you mean here

docs/ui/ComponentDemo.tsx
sarahgm: but you can style them like shadcn :)

docs/contentlayer.config.ts
sarahgm: some from rehypePrettycode

sebald: We don't want to just copy 😄 it's fine to use a prebuild one

docs/contentlayer.config.ts
sebald: Hm. But what issue she's this fix?

docs/ui/CopyButton.tsx
sebald: We could use something like this too :)

https://www.npmjs.com/package/react-use-clipboard

docs/contentlayer.config.ts
OsamaAbdellateef: in some scenarios the code may rely on the presence of at least one child element within the `node` object. If the code assumes the existence of a child element but the `children` array is empty, it could lead to errors or unexpected behavior. By adding a default child element, the code can ensure that it always has something to work with 
IDK if this is applicable here 
i am trying to figure why did they do that 
but that's what i think

docs/contentlayer.config.ts
OsamaAbdellateef: you know what 
The prebuild  one looks better xD

sebald (approved):

PR #3215, opened 2023-07-24
docs: add Hook docs pages

sarahgm:


sebald:
Hm. Do we need another page type? Can't this be the same like a regular page?

sarahgm:
> Hm. Do we need another page type? Can't this be the same like a regular page?

@sebald  So as you see I would say so :D :D

OsamaAbdellateef (approved):

PR #3226, opened 2023-07-25
chore[DST-106]: remove transitions

sarahgm:


sarahgm:
also tried to use tailwind classes for transition but also not very good :/

sebald (approved):

PR #3229, opened 2023-07-26
chore[DST-60]: change depricated API

sarahgm:


packages/components/src/Table/Table.tsx
```diff
@@ -90,17 +90,21 @@ export const Table: Table = ({
...
-              {[...row.childNodes].map(cell =>
-                cell.props?.isSelectionCell ? (
-                  <TableCheckboxCell key={cell.key} cell={cell} />
-                ) : (
-                  <TableCell key={cell.key} cell={cell} />
-                )
-              )}
-            </TableRow>
-          ))}
+          {/* {...bodyElements.map(row as  => ( */}
```
sebald, line 93: Remove? 😀

packages/components/src/Table/TableRow.tsx
```diff
@@ -22,13 +22,10 @@ export const TableRow = ({ children, row }: TableRowProps) => {
   const ref = useRef(null);
   const { interactive, state, ...ctx } = useTableContext();
 
-  const { variant, size } = row.props;
```
sebald, line 25: Is this no longer possible?

sebald (changes requested):

packages/components/src/Table/Table.tsx
sarahgm: ups

packages/components/src/Table/TableRow.tsx
sarahgm: to be honest, I don't actually see the use case

sarahgm: should I add this question comment also again? ;)

sebald: the idea was to have a highlighted row, maybe. but we can also readd it later if necessary :)

sebald (approved):

PR #3232, opened 2023-07-27
[DST-65]: Adding Space for Popover based on theme

OsamaAbdellateef:


OsamaAbdellateef:
@sebald 
There are two solutions for me 
- first is to use `useTheme` hook and based on the current theme we set the space 
- second is to create styles files for popover in `core/b2b-theme` which will be a good base to build on it

sebald:
@OsamaAbdellateef yeah I think the second one makes the most sense, right?

OsamaAbdellateef:
> @OsamaAbdellateef yeah I think the second one makes the most sense, right?

yeah right

packages/components/src/Overlay/Popover.test.tsx
```diff
@@ -13,6 +13,13 @@ import { setup } from '../test.utils';
 const theme: Theme = {
   name: 'test',
   components: {
+    Popover: cva(['mt-0.5'], {
```
sebald, line 16: Hm, I think we always need to consider the position, right? The placement has to be passed to the styles inside the component.

sebald (changes requested):

packages/components/src/Overlay/Popover.test.tsx
OsamaAbdellateef: @sebald  
then should i create a component and create test for it  ?

OsamaAbdellateef: @sebald 
I don't understand

sarahgm: I also don't understand @sebald

sarahgm: I thought this issue is about to make a space prop where you can put in any value you want 🤔

sebald: Yes, but adding matching to the top will not work when the popover is shown above the anchor (e g. when there is not enough space before the anchor)

sebald (dismissed):

sebald (approved):

PR #3244, opened 2023-08-01
[DST-92]: fix styles for `Input` of type `file` & `color`

OsamaAbdellateef:


OsamaAbdellateef:
> how can I test?

@sarahgm 
added some stories for them so you can test them and i will delete them later

OsamaAbdellateef:
> how can I test?

I

> how can I test?

@sarahgm 
I have added more types, you can change them from storybook options for input `type `

packages/components/src/Input/Input.stories.tsx
```diff
@@ -26,6 +26,8 @@ const meta = {
         'time',
         'url',
         'week',
+        'file',
```
sebald, line 29: Can you also make a separate story to show them! :)

sebald (changes requested):

packages/components/src/Input/Input.stories.tsx
OsamaAbdellateef: @sebald 
already did that in this commit but i deleted them  :D 
https://github.com/marigold-ui/marigold/pull/3244/commits/e1dc6ab8c0fa08e27e01496695c16f1fdf14527a
i will get them back :100:

sarahgm (approved):

PR #3250, opened 2023-08-04
[DST-111]:  enhance styling tabs 

OsamaAbdellateef:


sebald:
@OsamaAbdellateef Tests are missing! :)

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -38,21 +33,42 @@ export const Tabs = ({
...
+          ref={ref}
+        >
+          {[...state.collection].map(item => {
+            return <Tab key={item.key} item={item} state={state} />;
+          })}
+        </div>
+        <TabPanel
+          key={state.selectedItem?.key}
+          state={state}
+          className={state.selectedItem?.props?.className}
```
sebald, line 52: Can we do this via an aria selector instead?

OsamaAbdellateef: @sebald 
can't use it 
i moved it to `TabPanel` too , since we don't need to pass it down to it

packages/components/src/Tabs/TabPanel.tsx
```diff
@@ -10,9 +10,10 @@ export interface TabPanelProps extends AriaTabPanelProps {
 export const TabPanel = ({ state, ...props }: TabPanelProps) => {
   const ref = useRef(null);
   const { tabPanelProps } = useTabPanel(props, state, ref);
+  const selectedItemProps = state.selectedItem?.props;
   return (
-    <div {...tabPanelProps} ref={ref}>
-      {state.selectedItem?.props.children}
+    <div className={selectedItemProps?.className} ref={ref} {...tabPanelProps}>
```
sebald, line 15: Can this be styled via theme?

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -38,21 +33,38 @@ export const Tabs = ({
...
-      <div
-        className={cn('flex', gapSpace[space], classNames.tabs)}
-        {...tabListProps}
-        ref={ref}
-      >
-        {[...state.collection].map(item => {
-          return <Tab key={item.key} item={item} state={state} />;
-        })}
+      {/* tabs container */}
+      <div className={className}>
```
sebald, line 39: this should be possible to be styled via theme? :) Can you extend the component style thing?

sebald (changes requested):

packages/components/src/Tabs/Tabs.test.tsx
```diff
@@ -110,28 +128,34 @@ test('cursor indicates interactivity', () => {
...
+      aria-label="tabs container"
+      className="flex gap-2"
+      disabledKeys={['2']}
+    >
+      <Tabs.Item className="bg-green-300" key="1" title={'tab1'}>
         tab-1 content
       </Tabs.Item>
-      <Tabs.Item title="tab2" key="2">
+      <Tabs.Item key="2" title={'tab2'}>
         tab-2 content
```
sarahgm, line 149: can I also set a variant on the `<Tab.Panel>` ?

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -9,19 +9,14 @@ import { Tab } from './Tab';
...
-  extends Omit<AriaTabListProps<object>, 'orientation' | 'isDisabled'>,
-    GapSpaceProp {
-  size?: 'small' | 'medium' | 'large';
-  disabled?: boolean;
-  variant?: string;
-}
-export const Tabs = ({
+export const Tabs: Tabs = ({
   space = 2,
   size = 'medium',
```
sarahgm, line 16: for what is this used? @OsamaAbdellateef

packages/components/src/Tabs/Tabs.test.tsx
OsamaAbdellateef: no we can't , why should we pass variant for `<Tabs.Item>` ?

OsamaAbdellateef: > no we can't , why should we pass variant for `<Tabs.Item>` ?

@sarahgm

packages/components/src/Tabs/Tabs.tsx
OsamaAbdellateef: This is for padding and i think it is useless 
`      variants: {
        size: {
          small: 'px-1 pb-1',
          medium: 'px-2 pb-2 text-lg',
          large: 'px-4 pb-4 text-2xl',
        },
      },`

packages/components/src/Tabs/Tabs.test.tsx
sarahgm: don't know maybe we need this in future, or how can I set styles on the panel?

OsamaAbdellateef: you can style panel like this : 
`
<Tabs.Item className='bg-red-300'>
   tab panel content to be styled 
</Tabs.Item>
`
so the `className` here is for tab panel not for the tab button 
@sarahgm

sarahgm: but how would a user than style it?

OsamaAbdellateef: you mean the tab button ?!  
if that what u mean then idk 
because className is for tab panel but what if you want to style the tab button ?! then you should make this className reserved for tab  button instead of tab panel 
what i mean here you only have one className not two then you either chose to style for tab panel or button ?

sarahgm: Ehhhh Now I'm also confused :D :D @sebald you as author of the issue can you explain to me? :D

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -38,21 +33,38 @@ export const Tabs = ({
...
     </TabContext.Provider>
   );
 };
 
 Tabs.Item = Item;
+
+interface TabsProps
+  extends Omit<AriaTabListProps<object>, 'orientation' | 'isDisabled'>,
+    GapSpaceProp {
+  size?: 'small' | 'medium' | 'large';
```
sebald, line 60: string :)

packages/components/src/Tabs/Context.ts
```diff
@@ -1,6 +1,11 @@
 import { createContext, useContext } from 'react';
 
 export const TabContext = createContext<{
-  classNames: { tabs: string; tab: string };
+  classNames: {
```
sarahgm, line 4: do we than need this still? @OsamaAbdellateef

packages/components/src/Tabs/TabPanel.tsx
```diff
@@ -10,9 +12,12 @@ export interface TabPanelProps extends AriaTabPanelProps {
 export const TabPanel = ({ state, ...props }: TabPanelProps) => {
   const ref = useRef(null);
   const { tabPanelProps } = useTabPanel(props, state, ref);
+  const selectedItemProps = state.selectedItem?.props;
+  const { classNames } = useTabContext();
```
sarahgm, line 16: I think this we also don't need than, right @OsamaAbdellateef

OsamaAbdellateef: @sarahgm 
I think we need it, these `classNames` for styles via theme

packages/components/src/Tabs/Context.ts
OsamaAbdellateef: @sarahgm 
`
    component: 'Tabs',
    size,
    variant,
  });
`
`    <TabContext.Provider value={{ classNames }}>
`
instead of using `useClassNames` multiple times

packages/components/src/Tabs/Tabs.test.tsx
sebald: We want the variant to be used by the container, tab, tablist and tabpanel :) Does this help?

sarahgm (approved):

packages/components/src/Tabs/Context.ts
OsamaAbdellateef: @sarahgm 
is it clear ?

sebald (approved):

sebald (approved):

PR #3340, opened 2023-09-12
docs[DST-189]: write utils documentation

sarahgm:


docs/content/hooks-and-utils/utils/utils.mdx
```diff
@@ -0,0 +1,55 @@
+---
```
sebald, line 1: Move to own pages?

docs/content/hooks-and-utils/utils/utils.mdx
```diff
@@ -0,0 +1,55 @@
...
+  )}
+/>
+```
+
+In this code you can see that `p-3` will be overwritten by `p-4`, so keep in mind the order of your class names.
+
+## cva function
+
+To write our styles for a component we use [cva](https://cva.style/docs). This library comes with a great utility function which we use to write styles for our components.
+If you want to style your own component you can use this function which we updated with some additional style checks.
```
sebald, line 41: add note that it will resolve conflicting tailwind classnames

sebald (changes requested):

docs/content/hooks-and-utils/utils/utils.mdx
sarahgm: done

docs/content/hooks-and-utils/utils/utils.mdx
sarahgm: done

sarahgm:
@sebald also added badges

sebald (approved):

PR #3365, opened 2023-09-15
fix: typescript error message in fields (props)

sarahgm:


sarahgm:
not sure about this :S

sebald (approved):

PR #3373, opened 2023-09-27
[DST-40]: delete hover state

OsamaAbdellateef:


.changeset/silver-eagles-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@marigold/theme-b2b': major
```
sarahgm, line 2: I would say minor

sarahgm (changes requested):

PR #3395, opened 2023-10-06
[DST-213]: Adding `tailwindcss-react-aria-components` plugin

OsamaAbdellateef:
We need to check all the component styles 😢 

- [ ] Accordion
- [ ] Autocomplete
- [ ] Badge
- [ ] Button
- [ ] Calendar
- [ ] Card
- [ ] Checkbox
- [ ] Chevron
- [ ] ComboBox
- [ ] DateField
- [ ] DatePicker
- [ ] Dialog
- [ ] Divider
- [ ] FieldBase
- [ ] Header
- [ ] Headline
- [ ] HelpText
- [ ] Input
- [ ] Label
- [ ] Link
- [ ] List
- [ ] ListBox
- [ ] Menu
- [ ] Message
- [ ] NumberField
- [ ] Overlay
- [ ] Provider
- [ ] Radio
- [ ] SearchField
- [ ] Select
- [ ] Slider
- [ ] Switch
- [ ] Table
- [ ] Tabs
- [ ] TagGroup
- [ ] Text
- [ ] TextArea
- [ ] TextField
- [ ] Tooltip
- [ ] XLoader


package.json
```diff
@@ -30,6 +30,7 @@
     "prettier": "3.0.3",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "tailwindcss-react-aria-components": "1.0.0-beta.1",
```
sebald, line 33: Should not be added to the root 😄

themes/theme-b2b/tailwind.config.ts
```diff
@@ -3,6 +3,7 @@ import type { Config } from 'tailwindcss';
 import { preset } from './src/preset';
 
 export default {
+  plugins: [require('tailwindcss-react-aria-components')],
```
sebald, line 6: If we add it to our `createPreset` helper every theme will get it! :)

OsamaAbdellateef: now they are seperated

themes/theme-core/tailwind.config.ts
```diff
@@ -5,6 +5,7 @@ import { preset } from './src/preset';
 // Figma File: https://www.figma.com/file/RiWJBV4Z8L8ycVvUuMYXbm/%F0%9F%93%93-CR---Components-2022?node-id=1452-1785&t=YaLGVHzniD5mOtbJ-0
 
 export default {
+  plugins: [require('tailwindcss-react-aria-components')],
```
sebald, line 8: Can't we add it to the preset? :)

sebald (changes requested):

themes/theme-core/tailwind.config.ts
sebald: Here:

https://github.com/marigold-ui/marigold/blob/main/packages/theme-preset/src/create-preset.ts#L17

sebald:
@OsamaAbdellateef @sarahgm prefixed the plugin so we can merge it, can you check again if it breaking any CSS please :)

packages/theme-preset/src/create-preset.ts
```diff
@@ -15,6 +17,7 @@ export const createPreset = (name: string, config: Partial<OptionalConfig>) =>
         },
       },
       plugins: [
+        racSelectorPlugin({ prefix: 'rac' }),
```
sarahgm, line 20: how do we should use this than? @sebald

sarahgm (approved):

PR #3408, opened 2023-10-12
RAC: `<Headline>` as React Aria Component

sarahgm:


packages/components/src/Headline/Headline.tsx
```diff
@@ -9,31 +10,23 @@ import {
...
-// ---------------
-export const Headline = ({
-  children,
+const _Headline = ({
   variant,
   size,
+  children,
   align = 'left',
   color,
-  level = '1',
```
sebald, line 35: Hm. should we allow both so this is not a breaking change?

sebald (changes requested):

packages/components/src/Headline/Headline.tsx
sarahgm: But than I can't use the prop from RAC oder?👀 @sebald

sebald: You can cast it with `Number(level)`. If we only allow 1-6 as string plug 1-6 as number this should be safe :)

sarahgm: done please have a loook :)

sebald (approved):

sarahgm:
Why did you changed it? :( @sebald

sebald:
@sarahgm if we are only using string, there is no good autocomplete and the cast to number might not work :)

sarahgm:
I wrote a test for it... also i ask earlier that I want to use the level from react aria components. But yes your right. Just feels a bit not so good @sebald

PR #3450, opened 2023-10-25
refa: use non-rac checkbox in table

sebald:


sebald:
@OsamaAbdellateef @sarahgm this makes it work again (AFAIK) but the checkboxes in the row are not styled anymore. why? :-O

sarahgm:
> @OsamaAbdellateef @sarahgm this makes it work again (AFAIK) but the checkboxes in the row are not styled anymore. why? :-O

maybe because the classnames slipped to another element?

sarahgm:
> > > @OsamaAbdellateef @sarahgm this makes it work again (AFAIK) but the checkboxes in the row are not styled anymore. why? :-O
> > 
> > 
> > maybe because the classnames slipped to another element?
> 
> But there is no className involved 🤔 or is there?

I think about className.checkbox :)

sebald:
superseeded by #3451

PR #3471, opened 2023-11-09
 adding fixed height scrollable table example

OsamaAbdellateef:


aromko:
There is some weird behaviour in Storybook. When I click on a header column, e.g. completed , the third row will be checked. It also works for title and id 😅

sebald (approved):

PR #3476, opened 2023-11-13
Fix scrollable table story

OsamaAbdellateef:


packages/components/src/Table/Table.stories.tsx
```diff
@@ -575,7 +579,7 @@ export const ScrollableTable: Story = {
                 </Table.Header>
                 <Table.Body>
                   {todos.map(todo => (
-                    <Table.Row key={todo.id}>
+                    <Table.Row key={todo.title}>
```
aromko, line 582: For me it doesn't feels right to set title as key. For this example it works but I think in a real world application it can lead to using the key more than once, because you can't expect a title to be unique.

I also take a look and it seems the key prop from Table.Column and the key prop from Table.Row affects each other. So for column and row exists same keys and this leads to this weird behavior?! I'm not sure why. Maybe @sebald knows why?

So a cleaner fix for me would be to use a composite key, e.g.  key={`${todo.id}-${todo.title`} or some generated key  like key={Math.random()} or key={crypto.randomUUID()}

aromko (approved):

PR #3477, opened 2023-11-13
RAC: migrate input

aromko:


packages/components/src/Input/_Input.test.tsx
```diff
@@ -0,0 +1,85 @@
+/* eslint-disable testing-library/no-node-access */
+import { screen } from '@testing-library/react';
```
sarahgm, line 2: why extra file for tests? @aromko

aromko: @sarahgm because codecov is moaning when I add a new file and no tests are available.

sarahgm: Ah as I understand we used for stories and test the new component anyway 🤔 or how do you mean?

sarahgm (approved):

PR #3478, opened 2023-11-13
introducing RAC Modal & Dialog components

OsamaAbdellateef:


sebald:
Is it possible to separate these PRs :D

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -60,79 +35,58 @@ const CloseButton = ({ className }: CloseButtonProps) => {
...
   ...props
 }: DialogProps) => {
-  const ref = useRef(null);
-  const { close } = useDialogContext();
-  const { dialogProps, titleProps } = useDialog(props, ref);
-
   const classNames = useClassNames({ component: 'Dialog', variant, size });
-
+  const DialogWrapper = ({ children }: { children: ReactNode }) => {
+    if (isNonModal) return <>{children}</>;
```
OsamaAbdellateef, line 58: I don't know if this is right or not 
i am trying to do it like before , so Dialog can behave as a modal or as a normal dialog

packages/components/src/Overlay/Modal.tsx
```diff
@@ -1,69 +1,35 @@
...
-        <Underlay {...underlayProps} variant="modal" />
-        <div
-          className="pointer-none fixed inset-0 z-50 flex items-center justify-center"
-          ref={modalRef}
-          {...mergeProps(props, overlayProps, modalProps)}
-        >
-          <div style={{ pointerEvents: 'auto' }}>{children}</div>
-        </div>
-      </FocusScope>
+      <Underlay variant="modal">
```
OsamaAbdellateef, line 26: I added Underlay to Modal , 
should i separate it or this is better ?
@sebald

packages/components/src/Dialog/Dialog.tsx
sebald: Maybe we don't need this anymore? what does the wrapper do?

packages/components/src/Overlay/Modal.tsx
sebald: is there no "underlay" component from RAC?

packages/components/src/Dialog/Dialog.tsx
OsamaAbdellateef: in the beginning i was confused about the Dialog and Modal 
what is normal is that dialog is just that box inside the modal (in most of libraries and based on my researches)
but what i saw in our old `Dialog` that it is a combination of(`modal` + `dialog`) and it can behave like the `modal` 
because there's `Modal` inside `DialogTrigger` component 
so i won't put Modal over the Dialog component where should it be ?
or should i do like what react-aria did 
```
<DialogTrigger>
  <Button>Sign up…</Button>
  <Modal>
    <Dialog>
      {({ close }) => (
        <form>
          <Heading slot="title">Sign up</Heading>
          <TextField autoFocus>
            <Label>First Name</Label>
            <Input />
          </TextField>
          <TextField>
            <Label>Last Name</Label>
            <Input />
          </TextField>
          <Button onPress={close}>
            Submit
          </Button>
        </form>
      )}
    </Dialog>
  </Modal>
</DialogTrigger>
```

packages/components/src/Overlay/Modal.tsx
OsamaAbdellateef: I don't think so , 
there's `ModalOverlay` which is being used now

PR #3480, opened 2023-11-14
RAC: add `FieldError` in `HelpText`

sarahgm:


packages/components/src/FieldBase/FieldBase.test.tsx
```diff
@@ -101,28 +103,27 @@ test('render Field with label and helptext', () => {
   expect(error).not.toBeInTheDocument();
 });
 
-test('render Field with label and errorMessage', () => {
+test('render Field with label and errorMessage', async () => {
```
aromko, line 106: Why async needed? I would it expect when `await` is somewhere used.

sarahgm: true :D

aromko (approved):

PR #3484, opened 2023-11-16
docs: design token spec page

sebald:


sebald:
Took my way longer than expected ... but while writing I noticed that we should also define the naming convention :D 

So I tried here: https://marigold-docs-git-tokens-prep-marigold.vercel.app/__draft__/design-token-spec#naming-convention

Please give some feedback! @sarahgm @KadaStrophe @aromko @OsamaAbdellateef

sarahgm:
@sebald conflicts

sarahgm (approved):

PR #3487, opened 2023-11-17
RAC: `<Tooltip>` to react aria components

sarahgm:


packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -1,45 +1,39 @@
...
   variant?: string;
   size?: string;
+  open?: RAC.TooltipProps['isOpen'];
 }
 
-// Component
-// ---------------
-export const Tooltip = ({ children, variant, size }: TooltipProps) => {
-  const { arrowProps, state, overlayRef, placement, ...rest } =
-    useTooltipContext();
```
aromko, line 23: Because we don't use useTooltipContext anymore the file Tooltip/Context.ts can be removed. Then codecov should also be happy.

packages/components/src/Tooltip/Tooltip.stories.tsx
```diff
@@ -27,56 +20,62 @@ const meta = {
...
     containerPadding: {
       control: {
         type: 'number',
       },
-      description: 'The padding around the tooltip',
+      description:
+        'The placement padding that should be applied between the element and its surrounding container.',
     },
-    shouldFlip: {
+    open: {
```
aromko, line 43: The open and isOpen functions do not seem to work properly. If I understand the documentation correctly, it must also be with controlled state?!

https://react-spectrum.adobe.com/react-aria/Tooltip.html#controlled-open-state

aromko (approved):

PR #3499, opened 2023-11-21
RAC: `DateField` to react-aria-components

sarahgm:


themes/theme-b2b/src/components/DateField.styles.ts
```diff
@@ -1,25 +1,24 @@
...
     inputBackground,
-    inputDisabled('group-disabled/field'),
-    inputFocus('group-focus/field'),
     inputSpacing,
     'h-[34px]',
-    'group-hover/field:border-border-hover',
-    'group-readonly/field:group-focus/field:border-border-light group-readonly/field:group-focus/field:outline-none',
-    ' group-readonly/field:bg-bg-transparent group-readonly/field:text-text-disabled',
+    'data-[hovered]:border-border-hover',
+    // inputDisabled('data-[disabled]') color tokens are not working with in this component
```
sarahgm, line 12: maybe you have a better idea here? @OsamaAbdellateef @sebald @aromko ?

sebald: can you explain the issue more? :)

sebald: is it because of the group?

sarahgm: hard to write better to show you, but basically it's when i use the disabled and focus styles directly without the inputDisabled() function it works. Otherwise it also works but not the colors just the cursor.

sarahgm: so in core theme everything works fine because there it also was like this before

sebald (approved):

PR #3506, opened 2023-11-22
chore: new core token structure

sarahgm:


themes/theme-core/src/tokens.ts
aromko, line 1: Do we need the secondary on brand? They are not listed in the new tokens spec.

sebald (approved):

PR #3512, opened 2023-11-23
fix([DST-243]): Fix `<Listbox>` overflowing the `<Popover>`

sebald:


sarahgm:
@sebald Was it intended that it's not working for core theme?

sebald:
> @sebald Was it intended that it's not working for core theme?

I guess not. What isn't working?

sarahgm (approved):

PR #3513, opened 2023-11-23
RAC: migrate `<TagGroup>` component

aromko:


packages/components/src/TagGroup/TagGroup.test.tsx
```diff
@@ -163,7 +163,8 @@ test('should remember last focused element', async () => {
   expect(tags[0]).toHaveFocus();
 });
 
-test('renders error message', () => {
+//test currently not work because there is no FieldErrorContext in TagGroup component
+/*test('renders error message', () => {
```
sarahgm, line 167: why do you need a FieldErrorContext?

aromko: To render the errorMessage like in our other components. I looked at the github repo from react-aria and currently they haven't the FieldErrorContext in this component. So our logic doesn't work to render the errorMessage.

sebald: I am also wondering *when* is a tag group invalid? Required? Only 3 tags? 🤔

aromko: Do we need required anymore when we remove the error state?

sebald: no I guess we remove it too

sarahgm (approved):

PR #3517, opened 2023-11-24
chore[DST-220]: seperate data-theme attribute and classnames on provider

sarahgm:


packages/components/src/Provider/MarigoldProvider.tsx
```diff
@@ -4,27 +4,36 @@ import {
...
   const isTopLevel = outerTheme === defaultTheme;
+  const root = theme.root?.();
 
   return (
     <ThemeProvider theme={theme}>
-      {isTopLevel ? <OverlayProvider>{children}</OverlayProvider> : children}
+      {isTopLevel ? (
+        <OverlayProvider>{children}</OverlayProvider>
+      ) : (
+        <div className={cn(root ? root : '', className)}>{children}</div>
```
sebald, line 35: This will not get applied in regular setups, right? Because they only have only have ony provider.

sarahgm:
can't remember, too many weeks between we wait because of unstable_portalcontainer?

sarahgm:
Bildschirmfoto 2024-01-09 um 17 08 21


why? :(

sebald (approved):

PR #3535, opened 2023-12-04
RAC: `<Select>` component

OsamaAbdellateef:


packages/components/src/Overlay/Popover.tsx
```diff
@@ -20,7 +20,10 @@ export interface PopoverProps
 // Component
 // ---------------
 const _Popover = forwardRef<HTMLDivElement, PopoverProps>(
-  ({ keyboardDismissDisabled, placement, open, children, ...rest }, ref) => {
+  (
+    { keyboardDismissDisabled, placement, open, className, children, ...rest },
```
aromko, line 24: As far as I see you don't use the **className** prop somewhere or? So you can undo it and remove it from props again in L.14?

packages/components/src/Select/Select.stories.tsx
```diff
@@ -87,13 +88,13 @@ export const Basic: StoryObj<typeof Select> = {
...
-          <Select.Option key="Harry Potter">Harry Potter</Select.Option>
-          <Select.Option key="Lord of the Rings">
+          <Select.Option id="Harry Potter">Harry Potter</Select.Option>
+          <Select.Option id="Lord of the Rings">
             Lord of the Rings
           </Select.Option>
-          <Select.Option key="Star Wars">Star Wars</Select.Option>
-          <Select.Option key="Star Trek">Star Trek</Select.Option>
-          <Select.Option key="Firefly">Firefly</Select.Option>
+          <Select.Option id="Star Wars">Star Wars</Select.Option>
```
aromko, line 95: Same issue like in the menu @sebald mentioned: "the items now have "id" but the props are still called "key" should we rename them?"

aromko (changes requested):

packages/components/src/Select/Select.stories.tsx
aromko: Ok I think it is already decieded, so never mind. https://github.com/marigold-ui/marigold/pull/3546/files

.changeset/loud-sheep-lay.md
```diff
@@ -0,0 +1,6 @@
+---
+'@marigold/components': minor
+'@marigold/docs': minor
+---
+
+Migrate Select component to RAC
```
aromko, line 6: This breaking change info should also be added here: https://github.com/marigold-ui/marigold/pull/3546/files#diff-7617ce6947d6789d501bb7df466ddca07d7d0aa5264a4c3f43f9cdbbed05f22b

packages/components/src/Select/Select.stories.tsx
```diff
@@ -179,13 +180,15 @@ export const LotsOfOptions: StoryObj<typeof Select> = {
 export const Sections: StoryObj<typeof Select> = {
   render: args => (
     <Select {...args}>
-      <Select.Section title="Fantasy">
-        <Select.Option key="harry-potter">Harry Potter</Select.Option>
-        <Select.Option key="lord-of-the-rings">Lord of the Rings</Select.Option>
+      <Select.Section>
+        <Header>Fantasy</Header>
```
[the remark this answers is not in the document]
sarahgm: yes thank you

sarahgm (approved):

PR #3536, opened 2023-12-04
RAC: add "Tray" behavior inside Popover 

sarahgm:


sebald (approved):

sarahgm:
> @sarahgm merged, but I am not sure if the tray semantically should be a modal instead 🤔

but the Model is not used in Menu? Or how do you think? @sebald

sebald:
No, I mean the tray is not a popover but a dialog since it blocks interaction with the rest of the page and is not placed relative to an element :)

sarahgm:
I'm confused again. The popover is the only overlay element I thought inside the menu? @sebald

sebald:
> I'm confused again. The popover is the only overlay element I thought inside the menu? @sebald 

if you are displaying the overpay next to the button yes

PR #3544, opened 2023-12-08
feat[DST-245]: new table variant borderedTable for b2b theme

sarahgm:
added a bordered Table because:
- looked better
- in core already exists and now we have the same variant in both themes


themes/theme-b2b/src/components/Table.styles.ts
```diff
@@ -14,6 +14,7 @@ export const Table: ThemeComponent<'Table'> = {
         variant: {
           compact: 'py-2',
```
sebald, line 15: should we move `compact` and `expanded` to `size` then? Otherwehise it is not possible to make a compact bordered table?

themes/theme-b2b/src/components/Table.styles.ts
```diff
@@ -14,6 +14,7 @@ export const Table: ThemeComponent<'Table'> = {
...
           expanded: 'py-6',
+          borderedTable: 'border-x-[1px] border-t-[1px]',
```
sebald, line 17: - I thought we wanted to show all borders?
- Shouldn't it be kebab-cased? Maybe we can also call it `grid` or something.

sarahgm: I call it like in core to make it consistent :) We have there all borders? or what you mean= :)

.changeset/empty-jokes-tickle.md
```diff
@@ -0,0 +1,6 @@
+---
+"@marigold/components": patch
+"@marigold/theme-b2b": patch
```
sebald, line 3:
```suggestion
"@marigold/theme-b2b": major
```

.changeset/empty-jokes-tickle.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+feat: new table variant borderedTable for b2b theme
```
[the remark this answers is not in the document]
sarahgm: What do you mean? @sebald

themes/theme-b2b/src/components/Table.styles.ts
sarahgm: @sebald ?

sebald: I thought we also show the surrounding borders. I think we should stick to at least kebab case for naming, because this would end up in CSS :)

sebald (approved):

PR #3550, opened 2023-12-12
RAC: Autocomplete 

sarahgm:


.changeset/tiny-worms-drum.md
```diff
@@ -0,0 +1,6 @@
+---
+"@marigold/docs": patch
```
sebald, line 2: This one is breaking too :D Guess we should include this in the next release

sarahgm: oh sorry yes

PR #3551, opened 2023-12-13
[DST-130]: Adding column alignment feature

OsamaAbdellateef:


docs/content/components/collection/table/table.mdx
```diff
@@ -76,6 +76,12 @@ The `core-Theme` has also two variants called `tableLines` and `tableBorder`.
 
 <ComponentDemo file="./table-compact.demo.tsx" />
 
+### Aligned Columns
+
+In this example you can align any Table Column by using `align` property over `<Table.Column align='right'>` .
```
sarahgm, line 81: can you also add the new prop in the `<PropsTable/>`?

sarahgm: and a badge with `updated` ? ::)

OsamaAbdellateef: The badge is already there  ? 🙂
hmm but i thought that the props table contains only props which is related to `<Table>` only,  not `<Table.Column>`

sebald:
Just a proposal: Maybe we put the different components props documentation in tabs?

sarahgm:
> Just a proposal: Maybe we put the different components props documentation in tabs?

different ticket I would say

sebald (approved):

PR #3554, opened 2023-12-13
RAC: DatePicker as react aria component

sarahgm:


sarahgm:
If you want to know why codecov misses lines: we wrote this in the HelpText file:
`don't tested error messages because it can't be shown (react-aria-components)` 

@sebald @aromko @OsamaAbdellateef

packages/components/src/DateField/DateField.tsx
```diff
@@ -1,152 +1,71 @@
...
 
-import { AriaDateFieldProps, useDateField } from '@react-aria/datepicker';
-import { useFocusRing } from '@react-aria/focus';
-import { useLocale } from '@react-aria/i18n';
-import { useHover } from '@react-aria/interactions';
-import { mergeProps } from '@react-aria/utils';
+import { WidthProp } from '@marigold/system';
 
-import { useDateFieldState } from '@react-stately/datepicker';
+import { FieldBaseProps } from '../FieldBase';
```
aromko, line 7: Here are the wrong Props used or? Should be from _Fieldbase

Can we get rid of the old Fieldbase?

packages/components/src/DateField/DateInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+  variant?: string;
+  size?: string;
+  action?: ReactElement;
+}
+
+const _DateInput = ({ variant, size, action, ...props }: DateInputProps) => {
+  const classNames = useClassNames({ component: 'DateField', variant, size });
+
+  return (
+    <Group className={cn('flex flex-1 items-center', classNames.field)}>
```
aromko, line 21: same style in l.21,22. So we could move it into the style file?

packages/components/src/DateField/DateInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <DateInput className={cn('flex flex-1 items-center')} {...props}>
+        {segment => (
+          <DateSegment
+            className={cn(
+              'group/segment',
```
aromko, line 26: classnames.segment is only used here. So we can put the additional also in segment?

packages/components/src/DateField/DateSegment.tsx
```diff
@@ -1,74 +1,47 @@
...
+            aria-hidden="true"
+            className={cn(
+              isPlaceholder ? 'visible block' : 'invisible hidden',
+              'pointer-events-none w-full text-center'
+            )}
+          >
+            {isPlaceholder && placeholder?.toUpperCase()}
+          </span>
+          <span>
+            {isPlaceholder
```
aromko, line 33: Pretty nasty :-D 
Maybe one nested level less:

`<span>
  {isPlaceholder
    ? ''
    : (segment.type === 'month' || segment.type === 'day') && Number(segment.text) < 10
      ? '0' + segment.text
      : text}
</span>`

packages/components/src/DateField/DateSegment.tsx
```diff
@@ -1,74 +1,47 @@
...
-      )}
+    <DateSegment
+      {...props}
+      segment={segment}
       style={{
-        ...segmentProps.style,
-        minWidth: maxValue != null ? String(maxValue).length + 'ch' : undefined,
+        minWidth:
+          segment.maxValue != null
+            ? String(segment.maxValue).length + 'ch'
```
aromko, line 17: Here I would like a template string:

style={{
  minWidth: segment.maxValue != null ? `${String(segment.maxValue).length}ch` : undefined,
}}

themes/theme-b2b/src/components/DateField.styles.ts
```diff
@@ -10,9 +10,10 @@ export const DateField: ThemeComponent<'DateField'> = {
     'h-[34px]',
     'data-[hovered]:border-border-hover',
     // inputDisabled('data-[disabled]') color tokens are not working with in this component
-    'data-[disabled]:bg-bg-disabled data-[disabled]:border-border-disabled data-[disabled]:text-text-disabled data-[disabled]::cursor-not-allowed',
+    'data-[disabled]:bg-bg-disabled data-[disabled]:border-border-disabled data-[disabled]:text-text-disabled data-[disabled]:cursor-not-allowed',
     // inputFocus('data-[focus-within]') is not working with this component
```
aromko, line 14: I don't understand the comment here. Does that mean the style doesn't work? If so, why do we have it in there?

packages/components/src/DatePicker/DatePicker.stories.tsx
```diff
@@ -81,8 +82,10 @@ export const Controlled: Story = {
           <DatePicker
             label="Date Picker"
             value={value}
-            onChange={setValue}
+            defaultValue={value}
```
aromko, line 85: Do I need defaultValue here? It is a controlled state so I don't need it or?

sarahgm: we need it because to set the default... otherwise its not shown :(

themes/theme-b2b/src/components/DateField.styles.ts
sarahgm: Oh that comments are there, because for reasons the `input` styles where not working. You know ?

packages/components/src/DateField/DateField.tsx
sebald: After this one maybe :)

packages/components/src/DateField/DateInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+              'text-center leading-none outline-0',
+              '[data-type="literal]"' && 'p-[0.75px]',
+              classNames.segment
+            )}
+            segment={segment}
+          />
+        )}
+      </DateInput>
+      {action ? (
```
sebald, line 35: Should we make this not part of the date input? It is only needed by the picker anyway. Yes, it should be removed from the DateField 👀

sarahgm: the only child of a `DateInput` which is allowed is a `segment`, so I move it as it is to the `DateInput`, can you be more speciic? thank youu 👀  Also I leave the icon on the `DateField` because I don't want to change the design. 😬

packages/components/src/DateField/DateField.tsx
sarahgm: ups... should be noticed by the DateField migration :D resolved

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -1,116 +1,90 @@
...
   required,
-  readonly,
-  open,
+  readOnly,
   error,
-  shouldCloseOnSelect,
   variant,
   size,
-  width,
+  value,
```
aromko, line 41: According to my other comment from packages/components/src/DatePicker/DatePicker.stories.tsx. The value prop is destructured here and nowhere used anymore. So if we remove that the default value should work with only the **value** prop in a controlled state. Alternatively we can pass it to Fieldbase in l.61 but I think it isn't needed.

sarahgm: oh thank you :D resolved

aromko (approved):
From my side ok now 👍

PR #3555, opened 2023-12-14
chore: remove old Popover, HelpText and FieldBase

sarahgm:


sebald (approved):
This was very confusing because of the renaming the diffs where schown as if you changed a lot of stuff :D

PR #3557, opened 2023-12-14
[DSTSUP-5]: hide section when collapsed

aromko:


packages/components/src/Accordion/AccordionItem.tsx
```diff
@@ -93,14 +93,16 @@ export const AccordionItem = ({
...
-        >
-          {item.props.children}
-        </div>
-      ) : null}
+      <div
+        {...mergeProps(regionProps, focusProps, stateProps)}
+        className={
+          expanded || defaultExpanded
+            ? classNames.item
+            : cn(classNames.item, classNames.hideItem)
```
sarahgm, line 101: for me it looks a bit strange to have an `item` and `hideItem` I would put the `hidden ` in the file 🤔

sebald: There should be aria/data attributes for that (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded)

aromko: @sarahgm I thought we try to avoid to write directly styles in the component files?But I agree to not use hideItem. Maybe we can merge it.

@sebald Is there a way to say "When the button has aria-expanded=false hide all siblings/items?".

sarahgm (approved):

PR #3571, opened 2023-12-19
fix[DSTSUP-10]: add min/max props to textfield

sarahgm:


packages/components/src/TextField/TextField.tsx
```diff
@@ -30,6 +30,8 @@ export interface TextFieldProps
   required?: RAC.TextFieldProps['isRequired'];
   error?: RAC.TextFieldProps['isInvalid'];
   readOnly?: RAC.TextFieldProps['isReadOnly'];
+  min?: HTMLInputElement['min'];
```
aromko, line 33: We could use here the minLength and maxLength attributes from [Textfield](https://react-spectrum.adobe.com/react-aria/TextField.html#textfield-1) or?

sebald: min/max reference the actual value of the field

> minLength/maxLength reference the length of the input (characters) :)

aromko (approved):

PR #3592, opened 2023-12-21
feat: Enable built-in browser validation

sebald:


packages/components/src/Radio/RadioGroup.test.tsx
```diff
@@ -139,9 +139,7 @@ test('support horizontal orientation', () => {
     </ThemeProvider>
   );
 
-  // eslint-disable-next-line testing-library/no-node-access
-  const group = screen.getByLabelText('With Label').lastChild;
-
+  const group = screen.getByTestId('group');
```
aromko, line 142: @sebald: I wondering why this has to be changed and why it worked before? In my opinion this test was never right because the selected element has no data-orientation attribute...

sarahgm:
is this ready to review? :)

sarahgm (approved):

PR #3596, opened 2024-01-02
[DST-259]: Adding Helpers for percentage, date and currency 

OsamaAbdellateef:
Creating helpers for  currency ,  percentage  & date with an option to have  them in equal digit width


packages/system/src/components/Formatters/Date.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { DateFormatterOptions, useDateFormatter } from '@react-aria/i18n';
+
+interface DateFormatterProps extends DateFormatterOptions {
+  value: Date;
+  fullDigitWidth?: boolean;
+}
+
+export const DateFormatter = ({
+  value,
+  fullDigitWidth,
```
sebald, line 12: Not sure if devs will understand that name. I think "full" is kind of misleading?

packages/system/src/components/Formatters/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './Currency';
```
sebald, line 1: Can we also have a `Number` formatter :D (Maybe we need to name it `Numeric` because "Number" is part of the language)

packages/components/src/Table/TableCell.tsx
```diff
@@ -40,14 +40,12 @@ export const TableCell = ({ cell, align }: TableCellProps) => {
 
   const { focusProps, isFocusVisible } = useFocusRing();
   const stateProps = useStateProps({ disabled, focusVisible: isFocusVisible });
-
   return (
     <td
       ref={ref}
-      className={classNames?.cell}
+      className={cn(align ? `text-${align}` : '', classNames?.cell)}
```
sebald, line 46: Tailwind needs to see the full CSS class here `text-right` etc. otherwhise this might break in the future :)

sebald (approved):

sebald:
@OsamaAbdellateef needs docs! :)

packages/system/src/components/Formatters/index.ts
OsamaAbdellateef: yes sure 🙂

packages/system/src/components/Formatters/Date.tsx
OsamaAbdellateef: hmm, ok i have some suggestions for this,
for example : wideDigits, tabularDigits, expandedDigits, numericWidth
@sebald

docs/content/hooks-and-utils/date/date.mdx
```diff
@@ -0,0 +1,24 @@
+---
+title: DateFormatter
```
sebald, line 2: Not about "formatter", maybe "DateValue"? Or "DateFormat"?

OsamaAbdellateef:
i am thinking about why don't we collect `Percentage`, `Currency` and `Numeric` in on single component and we can call it `Formatter`, because from what you see , they the same a little bit, the only diff is `style`, 
or should we keep them as it is because it is more expressive and makes it clear and reusable 
@sebald  @sarahgm

docs/content/hooks-and-utils/date/date.mdx
sarahgm: I would say DateFormat is a good name

sebald:
> i am thinking about why don't we collect `Percentage`, `Currency` and `Numeric` in on single component and we can call it `Formatter`, because from what you see , they the same a little bit, the only diff is `style`, 
> or should we keep them as it is because it is more expressive and makes it clear and reusable 
> @sebald  @sarahgm 

I really like the idea!

sarahgm:
> > i am thinking about why don't we collect `Percentage`, `Currency` and `Numeric` in on single component and we can call it `Formatter`, because from what you see , they the same a little bit, the only diff is `style`,
> > or should we keep them as it is because it is more expressive and makes it clear and reusable
> > @sebald  @sarahgm
> 
> I really like the idea! Then we can also make one page for all the formatting

I agree :)

sebald:
@OsamaAbdellateef maybe the component then goes under "content"?

sebald:
> > > i am thinking about why don't we collect `Percentage`, `Currency` and `Numeric` in on single component and we can call it `Formatter`, because from what you see , they the same a little bit, the only diff is `style`,
> > > or should we keep them as it is because it is more expressive and makes it clear and reusable
> > > @sebald  @sarahgm
> > 
> > 
> > I really like the idea! Then we can also make one page for all the formatting
> 
> yesss ofc ✔️

Reviewed the PR. I thought we move it to a single component like we did with `Table` (`Table.Row`, etc). But this is also fine :) 

CC @sarahgm

sarahgm:
@OsamaAbdellateef conflicts

OsamaAbdellateef:
> @OsamaAbdellateef conflicts

resolved :heavy_check_mark:

sebald (approved):

PR #3621, opened 2024-01-11
[DST-266]:feat: use new b2b tokens in code

sarahgm:


.changeset/tiny-actors-applaud.md
```diff
@@ -0,0 +1,7 @@
+---
+"@marigold/components": minor
+"@marigold/theme-preset": minor
+"@marigold/theme-b2b": major
```
sebald, line 4: CC @lauraNawrotzki just to notify you adhead of time. If you are using any of the semantic tokens. With this PR they are now unified and all themes use the same tokens. And they will not change in the near future 🤞

sebald:
@KadaStrophe and Alex (who we need to setup with a Github Account today :D) please review this too!

themes/theme-b2b/src/components/Underlay.styles.ts
```diff
@@ -3,7 +3,7 @@ import { ThemeComponent, cva } from '@marigold/system';
 export const Underlay: ThemeComponent<'Underlay'> = cva('', {
   variants: {
     variant: {
-      modal: ' bg-bg-surface-underlay/40 backdrop-blur-sm',
+      modal: ' bg-bg-surface-sunken/40 backdrop-blur-sm',
```
tirado-rx, line 6: The underlay color in the story ("Overlay)" doesn't correspond to this value here.

themes/theme-b2b/src/components/TextArea.styles.ts
```diff
@@ -18,6 +18,6 @@ export const TextArea: ThemeComponent<'TextArea'> = cva([
   inputDisabled(),
   inputError('group-error/field'),
   inputSpacing,
-  'text-text-body font-body leading-8',
-  'read-only:focus:border-border-light read-only:focus:shadow-none',
+  'text-text-base font-body leading-8',
+  'read-only:focus:border-border-base read-only:focus:shadow-none',
```
tirado-rx, line 22: This one I don't really understand. When focused, the border of the text area turns blue, but this color here is gray[400]. Maybe I'm misunderstanding something?

sarahgm: Hmm... It should work when read-only is true 🤔 thats why read-only is set before focus

sarahgm: @tirado-rx I fixed it, should be work now :)

KadaStrophe (approved):

themes/theme-b2b/src/components/Switch.styles.ts
```diff
@@ -4,9 +4,9 @@ export const Switch: ThemeComponent<'Switch'> = {
   container: cva(''),
   track: cva(
     [
-      'bg-bg-surface-underlay shadow-shadow-light shadow-[0_0_0_1px]',
```
tirado-rx, line 7: Captura de pantalla 2024-01-16 a la(s) 20 05 01

This piece of code has seemingly been correctly updated, however, there is still a manual shadow used later in the same line of code. Is that intentional?

themes/theme-b2b/src/components/Radio.styles.ts
```diff
@@ -2,13 +2,13 @@ import { ThemeComponent, cva } from '@marigold/system';
 
 export const Radio: ThemeComponent<'Radio'> = {
   container: cva(),
-  label: cva('disabled:text-text-disabled'),
```
tirado-rx, line 5: "text-text-disabled" still exists in the span component—intentional?
Captura de pantalla 2024-01-16 a la(s) 20 18 09

tirado-rx: Oops, this is actually part of the Select component. I think the question still applies though, right?

tirado-rx: The above screenshot is actually from the Select component, but the same old class name still exists in this component as well. 
Captura de pantalla 2024-01-16 a la(s) 20 22 12

themes/theme-b2b/src/components/Popover.styles.ts
```diff
@@ -1,12 +1,15 @@
 import { ThemeComponent, cva } from '@marigold/system';
 
-export const Popover: ThemeComponent<'Popover'> = cva(['mt-0.5'], {
```
tirado-rx, line 3: I don't see a Popover component in Storybook—where do I find this?

themes/theme-b2b/src/components/NumberField.styles.ts
```diff
@@ -2,18 +2,18 @@ import { ThemeComponent, cva } from '@marigold/system';
 
 export const NumberField: ThemeComponent<'NumberField'> = {
   group: cva([
-    'border-border-light rounded-sm border border-solid',
-    'group-hover/field:border-border-hover',
-    'group-readonly/field:group-focus/field:border-border-light group-readonly/field:group-focus/field:shadow-none',
-    'group-disabled/field:bg-bg-disabled group-disabled/field:text-text-disabled group-disabled/field:cursor-not-allowed ',
```
tirado-rx, line 8: "text-text-disabled" is still being used in the label element. Intentional?
Captura de pantalla 2024-01-16 a la(s) 20 30 21

themes/theme-b2b/src/components/NumberField.styles.ts
```diff
@@ -2,18 +2,18 @@ import { ThemeComponent, cva } from '@marigold/system';
...
     'outline-2 -outline-offset-1 ',
-    'group-error/field:outline-outline-error group-error/field:outline',
+    'group-error/field:border-border-error ',
   ]),
   stepper: cva([
-    'text-text-body w-7',
-    'border-border-light border-solid first-of-type:border-r',
-    'border-border-light border-solid last-of-type:border-l ',
+    'text-text-base w-7',
+    'border-border-base border-solid first-of-type:border-r',
```
tirado-rx, line 15: I'm unable to find this line of code anywhere in the component.

themes/theme-b2b/src/components/Switch.styles.ts
sarahgm: I have removed the shadow

themes/theme-b2b/src/components/Radio.styles.ts
sarahgm: Oh your right this is the old token :) I fixed it, good spotted

themes/theme-b2b/src/components/Popover.styles.ts
sarahgm: You can see this for example in the `DatePicker` when you open the `Calendar`, or in the `Select` component. here is an example: 
Bildschirmfoto 2024-01-17 um 09 03 41

themes/theme-b2b/src/components/NumberField.styles.ts
sarahgm: So some of our components are divided into parts, and these parts can be styled sperat. So in this case the Numberfield contains multiple parts which can be styled. This line of code is from the `stepper` part. It's a bit difficult to spot it. The stepper are the plus and minus buttons.

themes/theme-b2b/src/components/Popover.styles.ts
tirado-rx: Thanks Sarah, I found it. In both the Date Picker and Select components, these lines of code have an additional class name at the end: "w-[--trigger-width]". Does that matter?
Captura de pantalla 2024-01-17 a la(s) 11 35 30

themes/theme-b2b/src/components/NumberField.styles.ts
tirado-rx: The change looks right, nice work!

tirado-rx: You just changed "text-text-disabled" to "text-text-base-disabled", right? Just to make sure I'm checking the right thing.

themes/theme-b2b/src/components/NumberField.styles.ts
tirado-rx: Alles claro

tirado-rx (approved):

PR #3642, opened 2024-01-16
[DSTSUP-14]: add example on how to use `onSelectionChange` within `<ComboBox>`

sarahgm:


sebald:
Because we anticipated that confusion, we actually adjusted this with the `<Select>` already. Maybe we should instead do the same with the combobox too?

See:

https://github.com/marigold-ui/marigold/blob/main/packages/components/src/Select/Select.tsx#L77

OsamaAbdellateef (approved):

PR #3666, opened 2024-01-23
fix: fix font in docs

sarahgm:


sarahgm:
@OsamaAbdellateef @sebald please checkout and test in our localhost aswell :) I'm not 100% sure why this happend 🤔

sarahgm:
don't know what this was but even now its working for me

PR #3668, opened 2024-01-23
[DSTSUP-30]: Fix-select-width

OsamaAbdellateef:


packages/components/src/Select/Select.tsx
```diff
@@ -56,7 +56,6 @@ export interface SelectComponent
 const _Select = forwardRef<any, SelectProps<object>>(
   (
     {
-      width,
```
sarahgm, line 59: why did we remove it? Shouldn't it be allowed?

sarahgm: @OsamaAbdellateef  :D

sarahgm (approved):

PR #3669, opened 2024-01-24
docs: fix demos in select docs page

sarahgm:


OsamaAbdellateef:
do we need a changeset  ? :) @sarahgm

OsamaAbdellateef (approved):

PR #3673, opened 2024-01-29
docs: wrap core demos with a default fieldgroup

sebald:


docs/ui/ComponentDemo.tsx
```diff
@@ -35,6 +35,13 @@ export const ComponentDemo = ({ name, children }: ComponentDemoProps) => {
   const Demo = registry[name].demo;
   const { current, themes } = useThemeSwitch();
 
+  const Wrapper = ({ children }: { children: ReactNode }) =>
+    current === 'core' ? (
+      <FieldGroup labelWidth="100px">{children}</FieldGroup>
```
sarahgm, line 40: in some demos are currently `<FieldGroups>` around, can you remove them than? :)

sebald: Should we? I thought this can be the default, but maybe, if you need more than 100px you add them in the demo too :)

sarahgm: ok I understand wasn't sure if its possible to have a FieldGroup in a FieldGroup

sebald: should be since it is just a context :)

sarahgm (approved):

PR #3683, opened 2024-01-30
fix: `placeholder` prop was missing on `TextArea`

sebald:


packages/components/src/TextArea/TextArea.tsx
```diff
@@ -32,6 +32,7 @@ export interface TextAreaProps
   readOnly?: RAC.TextFieldProps['isReadOnly'];
   value?: string;
   defaultValue?: string;
+  placeholder?: string;
```
sarahgm, line 35: @sebald  can you also add it in the docs page?

sarahgm: and in our storybook as prop? :)

PR #3721, opened 2024-02-15
feat: improve `<Tag>` responsive behavior and core theme styles

sarahgm:


themes/theme-b2b/src/components/Tag.styles.ts
```diff
@@ -1,7 +1,7 @@
 import { ThemeComponent, cva } from '@marigold/system';
 
 export const Tag: ThemeComponent<'Tag'> = {
-  tag: cva('border-border-base flex items-center rounded-md border p-1'),
+  tag: cva('border-border-base flex items-center gap-1 rounded-md border p-1'),
   closeButton: cva('right-1 size-4 cursor-pointer p-0 outline-none'),
-  listItems: cva('flex items-center gap-1'),
+  listItems: cva('flex flex-wrap items-center gap-1'),
```
sebald, line 6: Move it to the component instead of the themes?

themes/theme-core/src/components/Tag.styles.ts
```diff
@@ -2,10 +2,10 @@ import { ThemeComponent, cva } from '@marigold/system';
 
 export const Tag: ThemeComponent<'Tag'> = {
   tag: cva([
-    'border-border-brand bg-bg-surface flex items-center gap-2 rounded border',
+    'border-border-base bg-bg-base flex items-center gap-1 rounded border',
```
sebald, line 5: I think they should be more distinguishable from the background.

sebald: What token is bg-base? Should this even exist? 😮

sarahgm: I extra make it more in direction of b2b theme :) and bg-base is the  opposite of bg-inverted. Our docs says: 

base: Default values for a specific category of tokens, serving as the foundational colors

sarahgm: Otherwise we also have to adjust the b2b theme

tirado-rx (commented):
I think it's too easy to mistake our tags for a button. If you look at the image below, you'll see that other "North Star" systems make use of a different text style, container shape, or container size to differentiate the two. Our tags and buttons, by contrast, are much harder to differentiate, as we use the same text style and almost the same component shape/size.

I recommend that we add more differentiation before releasing. What do the others think? @sebald @KadaStrophe @OsamaAbdellateef 

Captura de pantalla 2024-02-16 a la(s) 8 56 07

tirado-rx (approved):
Decision w/ @KadaStrophe: We'll approve the changes that @sarahgm has made and write a separate issue ticket to better differentiate buttons and tabs. This is to balance our two new core values: 1.) Aiming to make long-term sustainable decisions, and 2.) Sticking to our established priority list. Thanks Sarah!

PR #3774, opened 2024-02-28
feat: remove normalization from themes when using CSS files

sebald:
(this causes a lot of trouble when using them because it is **NOT** prefixed!)


sebald:
I defined this as a `minor` because it is not relevant for our normal users, the only things affect by this are storybook and our docs (but they will get the normalization anyway)

OsamaAbdellateef (approved):

PR #3775, opened 2024-02-28
docs: change background back to white

sebald:
(as preparation for the cmdk feature)


docs/theme/tokens.ts
```diff
@@ -38,7 +38,7 @@ export const colors = {
   // Background
   // ---------------
   bg: {
-    body: tw.slate[50],
+    body: tw.white,
```
sarahgm, line 41: why?

PR #3777, opened 2024-02-28
[DST-271]:feat: Column allow fit width of item

sarahgm:


docs/content/components/layout/columns/columns.mdx
```diff
@@ -23,9 +24,9 @@ import { Columns } from '@marigold/components';
     {
       property: 'columns (required)',
       description:
-        'An array of numbers to set the size of the children. The columns array length and the count of children must be the same.',
-      type: 'number[]',
-      default: '',
+        'An array of numbers to set the size of the children. The columns array length and the count of children must be the same. Write "fit" for the column you want to have it fitting the contens width ',
```
aromko, line 27: typo content and missing dot at the end

docs/content/components/layout/columns/columns.mdx
```diff
@@ -23,9 +24,9 @@ import { Columns } from '@marigold/components';
...
+      type: 'number[] | "fit"',
```
aromko, line 28: I think this type definition is not entirely correct. It should be **type: '(number | 'fit')[]** or can I write columns='fit?

docs/content/components/layout/columns/fit-width-columns.demo.tsx
```diff
@@ -0,0 +1,10 @@
+import { Columns, Switch, TextField } from '@marigold/components';
+
+export default () => (
+  <Columns space={2} columns={[8, 'fit']} collapseAt="25em">
+    <TextField label="Setting 1" />
+    <div className="align-center flex">
```
aromko, line 6: The alignment doesn't seems to work.

docs/content/components/layout/columns/columns.mdx
sarahgm: your right :)

docs/content/components/layout/columns/fit-width-columns.demo.tsx
sarahgm: it is aligned, it only looks weird because of the label from the field... :/ Do you have an idea to solve this?

aromko (approved):

PR #3778, opened 2024-02-28
Fixing table header

OsamaAbdellateef:


OsamaAbdellateef:
@sebald 
Sorry i missed zIndex, so i make sure that the checkbox in table header is clickable

sebald (approved):

PR #3780, opened 2024-02-29
docs: removed some old props

aromko:


OsamaAbdellateef:
needs changeset ? @aromko

PR #3787, opened 2024-03-01
fix: Switch remove Selected border core theme

sarahgm:


themes/theme-core/src/components/Switch.styles.ts
```diff
@@ -8,9 +8,5 @@ export const Switch: ThemeComponent<'Switch'> = {
     'group-selected/switch:bg-bg-selected-input group-selected/switch:border-border-selected',
     'group-disabled/switch:bg-bg-base-disabled group-disabled/switch:border-border-base',
   ]),
-  thumb: cva([
-    'size-[20px]',
-    'bg-bg-base',
-    'group-selected/switch:border-border-selected',
-  ]),
+  thumb: cva(['size-[20px]', 'bg-bg-base']),
```
aromko, line 11: We can replace size-[20px] with size-5 :-)

aromko (approved):

PR #3800, opened 2024-03-05
Fixing divider component

OsamaAbdellateef:


aromko (commented):
Is it intended that the divider now looks darker and if bold selected it is much more bold than before?

OsamaAbdellateef:
> Is it intended that the divider now looks darker and if bold selected it is much more bold than before?

 The old way wasn't correct because there was border, so it looked wrong, after removing the border it looks correct(didn't change bg though)  @aromko

OsamaAbdellateef:
It wasn't intended to look like the old way :smile:  @aromko

aromko (approved):

PR #3812, opened 2024-03-08
[DSTSUP-59]:fix if no label no gap style

sarahgm:


packages/components/src/FieldBase/FieldBase.tsx
```diff
@@ -56,7 +56,13 @@ const _FieldBase = <T extends ElementType>(
   return (
     <Component
       ref={ref}
-      className={cn('group/field', twWidth[width], classNames, className)}
+      className={cn(
+        'group/field',
+        twWidth[width],
+        classNames,
+        className,
+        !label && `gap-x-0`
```
sebald, line 64: This is tragic specific to the core. Not sure if this will cause problems down the road

sarahgm: Why is it other than the other label if condition?

aromko (approved):

PR #3824, opened 2024-03-13
[DST-363]: Adding Guideline-Page of how to use `Multiselection`

OsamaAbdellateef:


OsamaAbdellateef:
now test is fixed  @sebald

sebald (dismissed):

packages/components/src/TagGroup/Tag.tsx
```diff
@@ -5,6 +5,9 @@ import { cn, useClassNames } from '@marigold/system';
 
 import { TagGroup } from './TagGroup';
 
+{
```
sebald, line 8: I think this does not belong here? :D

packages/components/src/index.ts
```diff
@@ -33,6 +33,7 @@ export * from './Link';
 export * from './List';
 export * from './Menu';
 export * from './Message';
+export * from './Multiselect';
```
sebald, line 36: Only a recipe, do not export as component.

sebald (changes requested):
@OsamaAbdellateef is this now the only PR to merge? 

E.g. this includes the recipe and guide?

OsamaAbdellateef:
> @OsamaAbdellateef is this now the only PR to merge?
> 
> E.g. this includes the recipe and guide?

Yeah it includes both @sebald

packages/components/src/index.ts
sebald: please remove the line :) we don't want to expose this yet!

sebald, line 36:
```suggestion
```

sebald (dismissed):

sebald (approved):

PR #3848, opened 2024-04-04
[DSTSUP-68] Change color token of the divider in b2b theme

sarahgm:


sarahgm:
I didn't change the files codecov does not cover :/ why

PR #3867, opened 2024-04-08
[DST-384]chore(docs): replace color orange with slate in demos

sarahgm:


KadaStrophe (approved):

docs/content/components/hooks-and-utils/extendTheme/extendTheme.mdx
```diff
@@ -21,7 +21,7 @@ const theme = extendTheme(someStyles, currentTheme);
```
aromko, line 21: Think we can change the badge to _updated_ or?

sarahgm: I don't thing we have to, only the demo color chaged not the functonality🤔

aromko (approved):

PR #3875, opened 2024-04-17
fix(DST-426, DST-426): remove dublicate dialog

sebald:
Restructure everthing so we don't accidentally open two dialogs.

Also refactored the `SiteHeader` to make this possible


docs/app/_components/SiteMenu.tsx
```diff
@@ -1,73 +1,92 @@
...
+  return (
+    <kbd className="hidden h-5 w-10 items-center justify-center rounded-md text-xs lg:inline-flex lg:border lg:border-gray-300 lg:bg-gray-200">
+      <span className="text-xs">{isMacOS ? '⌘' : 'Ctrl+'}</span>K
+    </kbd>
+  );
+};
+
+// Component
+// ---------------
+export const SiteMenu = () => {
```
aromko, line 52: Why it is now called SiteMenu? I was really irritated.

aromko (approved):

PR #3881, opened 2024-04-23
[DST-405]:chore: adjust Command styles and menu

sarahgm:


docs/theme/components/Menu.styles.ts
```diff
@@ -1,10 +1,33 @@
...
+          command:
+            '[&_[cmdk-list-sizer]]:divide-secondary-100 size-full px-4 sm:w-[500px] [&_[cmdk-input-wrapper]_svg]:size-5 [&_[cmdk-input]]:h-12 [&_[cmdk-item]]:py-2 [&_[cmdk-item]_svg]:size-5 [&_[cmdk-list-sizer]]:divide-y',
+        },
+      },
+    }
+  ),
+  item: cva('focus:bg-bg-hover cursor-pointer rounded p-2 outline-none', {
+    variants: {
+      variant: {
+        command: ['aria-selected:bg-bg-hover py-1.5'],
```
aromko, line 21: Some styles are missing now or? For example when I hover over an item the font-weight doesn't change. Or is it intended?

sarahgm: it was intented to align the Menu and Command styles.

sebald:
@sarahgm can we have the no padding back for the dialog? Looks a bit weird, like that 🙈

docs/theme/components/Menu.styles.ts
```diff
@@ -1,10 +1,33 @@
 import { ThemeComponent, cva } from '@marigold/system';
 
 export const Menu: ThemeComponent<'Menu'> = {
```
sebald, line 3: Before on mobile, it took the whole width. It's a bit small now

sarahgm: As I read this, it was always this applied: `size-full w-[500px] ` which doesn't make it look good on mobile: 
Bildschirmfoto 2024-04-26 um 08 22 07

sebald (approved):

PR #3884, opened 2024-04-25
docs: improve page layout

sebald:


packages/components/src/Text/Text.tsx
```diff
@@ -52,7 +52,9 @@ export const Text = ({
     variant,
     size,
   });
-
+  if (color) {
+    console.log(theme, color, getColor(theme, color, color /* fallback */));
+  }
```
sebald, line 57:
```suggestion

```

sebald:
I think we slightly need to optimize the mobile view (less px)

aromko (approved):

PR #3889, opened 2024-04-29
refa[DSTSUP-73]: Improve elevation (more distinguisable)

sebald:


sebald:
> [!NOTE]
> Waiting for feedback from our designers.

Also: I would move the input styles that are also shared to the `mixins` file afer this gets merged.

themes/theme-b2b/src/mixins.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Reusable set of styles (Tailwind classes).
+ */
+
+export const ELEVALTION_RING = 'ring-1 ring-gray-600/5';
```
sarahgm, line 5: Should this be also documented?  @sebald

sarahgm:
I think than needs to be also updated the elevation documentation

sebald:
@KadaStrophe can you send us an updated image (elevation page)

themes/theme-b2b/src/mixins.ts
sebald: becaaaaause?

packages/components/src/global.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+            SURFACE_BASE,
+            'bg-bg-surface-sunken shadow-surface-sunken'
+          )}
+        >
+          SUNKEN
+        </div>
+        <div
+          className={cn(
+            SURFACE_BASE,
+            'bg-bg-surface-raised shadow-surface-raised ring-1 ring-gray-600/5'
```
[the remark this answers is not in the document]
sebald: No, because this would make components depend on the b2b theme, I guess I delete this story because I have no cue where to put it.

themes/theme-b2b/src/mixins.ts
sarahgm: Is it a token?

packages/components/src/global.stories.tsx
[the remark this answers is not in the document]
sarahgm: Do you than want to delete it than now? @sebald

sarahgm (approved):

PR #3918, opened 2024-05-23
docs: update to `contenlayer2` + `rehype-pretty-code` galore

sebald:


aromko (commented):
I found some imports like for the old contentlayer/ .  Everything seems to work but can we than remove them?

docs/tsconfig.json
docs/app/_components/SiteMenu.tsx
docs/app/[...slug]/page.tsx
docs/app/_components/Navigation.tsx

sebald:
> I found some imports like for the old contentlayer/ . Everything seems to work but can we than remove them?
> 
> docs/tsconfig.json docs/app/_components/SiteMenu.tsx docs/app/[...slug]/page.tsx docs/app/_components/Navigation.tsx

these are the generated files! :) they have nothing to do with the lib itself, we adjust the path in the `tsconfig` file.

PR #3920, opened 2024-05-24
[DST-407]: enhance command menu

sarahgm:
- added keywords to search (e.g. search: help -> support bord and slack displayed, search: copy -> available copy objects displayed
- fixed search 
- added links, tokens and theme switch option
- copy icons & tokens
- added subpages when pressing a chosen page


sebald:
I think there is no need adding feature all at once. maybe we can already merge this?

docs/lib/commandlist.ts
```diff
@@ -0,0 +1,61 @@
+export const links = [
```
sebald, line 1: put this into the site config file?

docs/lib/utils.tsx
```diff
@@ -0,0 +1,17 @@
+interface NestedStringObject {
+  [key: string]: NestedStringObject | string;
+}
+
+export const iterateTokens = (colors: NestedStringObject, prefix = '') => {
```
sebald, line 5: Can we have some docs here. This is only for the command menu I imagine?

docs/app/_components/SiteMenu.tsx
```diff
@@ -66,33 +181,119 @@ export const SiteMenu = () => {
...
+                          )
+                        )}
+                      </>
+                    )}
+                  </>
+                ))}
+              </CommandGroup>
+            ))}
+            {/* update themes command */}
+            {themeswitch.map(({ name, items }) => (
```
sebald, line 270: Maybe pull these out into their own components? this would maek this part more readable.

sebald:
I think the Cmd+D shortcut should only be shown on the active item, this get's very noisy otherwhise and you also can judst open the active one, not all at once :)

image

docs/lib/utils.tsx
sarahgm: No I moved it out from the design-tokens.tsx because I used it also for the command :)

docs/app/_components/SiteMenu.tsx
sarahgm: You mean every different item should have its own component?

docs/lib/commandlist.ts
```diff
@@ -0,0 +1,61 @@
...
+        name: 'Support Center',
+        href: 'https://reservix.atlassian.net/servicedesk/customer/portal/77',
+        keywords: ['help'],
+      },
+      {
+        name: 'Storybook',
+        href: 'https://marigold-latest.vercel.app/?path=/story/components-accordion--basic',
+      },
+      {
+        name: 'Stackblitz',
```
sebald, line 29: Rename to "Marigold Playground"?

docs/lib/commandlist.ts
```diff
@@ -0,0 +1,61 @@
...
+        href: 'https://stackblitz.com/edit/github-v1evcv?file=src%2FApp.tsx',
+        keywords: ['playground', 'starter'],
+      },
+      {
+        name: 'Reference App',
```
sebald, line 34: too early to put here :D

sarahgm:
> I think the Cmd+D shortcut should only be shown on the active item, this get's very noisy otherwhise and you also can judst open the active one, not all at once :)
> 
> image


hm but how would you know that its possible? maybe you just search, and press enter that you wouldn't even know about it? Or am I thinking too complicated

docs/lib/commandlist.ts
```diff
@@ -0,0 +1,61 @@
...
+        href: 'https://marigold-rapp.vercel.app/',
+        keywords: ['help', 'example', 'code'],
+      },
+      {
+        name: 'Github',
```
sebald, line 39: Rename to "Code Repository"?

docs/app/_components/SiteMenu.tsx
```diff
@@ -101,16 +302,72 @@ export const SiteMenu = () => {
...
+                  >
+                    {elements.iconElement}
+                    {elements.icon}
+                  </CopyItem>
+                ))}
+              </CommandGroup>
+            )}
           </Command.List>
+          <div className="flex h-10 items-center justify-end gap-4 border-t px-2">
+            <div>
```
sebald, line 364: needs a `flex items-center` so these to are vertically aligned.

docs/app/_components/SiteMenu.tsx
```diff
@@ -101,16 +302,72 @@ export const SiteMenu = () => {
...
+                    copyValue={elements.svg}
+                  >
+                    {elements.iconElement}
+                    {elements.icon}
+                  </CopyItem>
+                ))}
+              </CommandGroup>
+            )}
           </Command.List>
+          <div className="flex h-10 items-center justify-end gap-4 border-t px-2">
```
sebald, line 363: would maybe look better if is `text-xs`?

docs/lib/commandlist.ts
sarahgm: but do you really think when the people talk about it they mean Marigold Playground? I also added keyword for playground.

docs/app/_components/SiteMenu.tsx
```diff
@@ -101,16 +302,72 @@ export const SiteMenu = () => {
...
+            <div>
+              Action <span className="opacity-50">↵</span>
+            </div>
+            <div>
+              More Details <Hotkey letter="D" />
```
sebald, line 368: Maybe this would look better if the hotkey would look more like a anctually key in this scenario?

sarahgm:
> I think the "theme" and "external link" icons are to prominent in regard to the actual items. They also look like a button to me.
> 
> image

for the themes I thought it makes more clear what the action is, because its already used in our theme switch menu :) Do you have a suggestion how to improve the externalLink icon and the change theme icon?

docs/lib/commandlist.ts
sarahgm: done

docs/app/_components/SiteMenu.tsx
sarahgm: what do you mean?

sarahgm: the hotkey is currently only for command key (hard coded) and than a letter you want (currently K and D)

docs/app/_components/SiteMenu.tsx
sarahgm: eh.... what? :D Its already there?

docs/lib/utils.tsx
sebald: Ah just saw it. But what does it do 👀

docs/app/_components/SiteMenu.tsx
sebald: Yes something like that e.g. `<ThemeComppands commands={themeswitch}/>`

docs/lib/utils.tsx
sarahgm: used to iterate through the colors and combine it in the right way e.g. 'bg-surface-sunken' I add it as comment

docs/app/_components/SiteMenu.tsx
sarahgm: I moved it into a new file, can you have a look again?

docs/app/_components/SiteMenu.tsx
sebald: in line 363 yes, but not in 364 :P

docs/app/_components/SiteMenu.tsx
sebald: I meant the styling here. in the navigation I think its fine, but in the menu, maybe it should be more prominent. 

e.g. make it look like an actual keyboard key: https://codepen.io/dudleystorey/pen/AgEpXj

docs/app/_components/SiteMenu.tsx
sarahgm: but why? the two divs are just as container for the text inside of it, it is in the middle

tirado-rx (commented):
Wording recommendations

-"Action"–>"Go"
-"More details"–>"Show sublevels" or "Show subcategories"

tirado-rx (commented):
After thinking about it some more: I would recommend changing "Show sublevels" to "Expand".

sebald (approved):

PR #3923, opened 2024-05-28
[DST-387]: Generate property tables in our documentation from code

OsamaAbdellateef:
Making the props dynamic based on the the types inside the component


docs/scripts/build-props-table.mjs
```diff
@@ -0,0 +1,64 @@
...
+  // include asserted types for components eg. Autocomplete & Select
+  customComponentTypes: ['AutocompleteComponent', 'SelectComponent'],
+});
+
+// Resolve __dirname for ESM
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+const componentsDir = path.resolve(__dirname, '../../packages/components/src');
+const outputFilePath = path.resolve(__dirname, '../.table-props/index.json');
```
sebald, line 23: I find table props somewhat confusing. Rather call it "component props"?, maybe we can put it next to the registry file? 

The file should be added to gitignore

docs/scripts/build-props-table.mjs
```diff
@@ -0,0 +1,64 @@
...
+
+fs.ensureDirSync(path.dirname(outputFilePath));
+
+const generatePropsTables = () => {
+  // Getting all directories
+  const componentDirs = fs
```
sebald, line 29: Use globby or minmatch here so you don't have to do it by hand?

docs/scripts/build-props-table.mjs
```diff
@@ -0,0 +1,64 @@
+import fs from 'fs-extra';
```
sebald, line 1: This should be zx script, so you have access to all relevant libs (fs-extra, globby, etc)

sebald (changes requested):

docs/scripts/build-props-table.mjs
OsamaAbdellateef: :heavy_check_mark: 
didn't get "we put it next to reg file "

docs/scripts/build-props-table.mjs
OsamaAbdellateef: zx doesn't include fs-extra :( @sebald

docs/package.json
```diff
@@ -24,10 +25,12 @@
     "contentlayer2": "0.4.6",
     "eslint": "8.57.0",
     "eslint-config-next": "14.1.4",
+    "fs-extra": "11.2.0",
```
sebald, line 28: Remove? :)

docs/scripts/build-component-props.mjs
```diff
@@ -0,0 +1,57 @@
+import fs from 'fs-extra';
```
sebald, line 1: This is also part of zx

docs/scripts/build-component-props.mjs
```diff
@@ -0,0 +1,57 @@
...
+import path from 'path';
+import { dirname } from 'path';
```
sebald, line 3: Imported twice. Also when importing node modules, we should use node:path for security :)

sebald (changes requested):

sebald:
Sorry, some small adjustments 🙈

docs/scripts/build-props-table.mjs
sebald: https://google.github.io/zx/api#fs :P

OsamaAbdellateef: aaaaaaaaaaaah mb ✅

sebald (approved):

PR #3936, opened 2024-06-06
Props Table:  Add description to  types of all components

OsamaAbdellateef:


packages/components/src/Table/Table.tsx
```diff
@@ -36,8 +36,23 @@ export interface TableProps
...
+
+  /**
+   * Stretch table to fill the container.
+   * @default false
+   */
   stretch?: boolean;
+
+  /**
+   * Make the column sticky to the top of the table.
+   * @default none
```
sarahgm, line 48: i think this should be default false right?

packages/components/src/Tiles/Tiles.tsx
```diff
@@ -3,25 +3,43 @@ import { ReactNode } from 'react';
...
+   */
   equalHeight?: boolean;
 }
 
 export const Tiles = ({
   space = 0,
   stretch = false,
   equalHeight = false,
-  tilesWidth,
+  tilesWidth = '250px',
```
sarahgm, line 34: why is there a default value?

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -10,9 +10,15 @@ import { TooltipTrigger } from './TooltipTrigger';
 type RemovedProps = 'className' | 'isOpen' | 'style';
 
 export interface TooltipProps extends Omit<RAC.TooltipProps, RemovedProps> {
+  /**
+   * The children of the coomponent.
```
sarahgm, line 14: typo by component

packages/components/src/Tiles/Tiles.tsx
OsamaAbdellateef: oh, i think i was testing & forgot it :D @sarahgm

OsamaAbdellateef: but i noticed it exists our docs table with default value `250px` idk why @sarahgm

packages/components/src/Table/Table.tsx
OsamaAbdellateef: yeah it shouldn't be none, but i think it should be true

sarahgm (approved):

PR #3939, opened 2024-06-13
feat[DST-481]: rename <Message> in <SectionMessage>

sarahgm:


sebald (approved):

sebald (approved):

docs/content/components/collection/table/table.mdx
```diff
@@ -123,10 +123,10 @@ We have formatting helper for dates and values you can easily ensure consistent
 
 Example for using controlled state with input fields inside a table.
 
-<Message messageTitle="Avoid mutation" variant="warning">
+<SectionMessage messageTitle="Avoid mutation" variant="warning">
```
sebald, line 126: props -> component

sebald (changes requested):
I think you also need to change the Messages now to use the compound component? :D

sebald (approved):

PR #3967, opened 2024-06-27
refa: Update TypeScript and adjust `<NumericFormat>` props

sebald:
BREAKING CHANGE: The `styleFormat` prop is now `style`. Not sure why we did this in the first place, but correcting this was the straight forward way to fix the typings.

The benefits is no that we (or TS) catches typos!


docs/content/components/formatters/numericformat/currency-format.demo.tsx
```diff
@@ -29,11 +29,7 @@ export default () => {
           <Radio value="EUR">EUR</Radio>
           <Radio value="JPY">JPY</Radio>
         </Radio.Group>
-        <NumericFormat
-          styleFormat="currency"
-          value={233}
-          currency={currentFormat}
-        />
+        <NumericFormat style="currency" value={233} currency={currentFormat} />
```
sarahgm, line 32: here is a warning that style prop must be an object

sarahgm: in every demo

sarahgm: and test

sarahgm (approved):

PR #3968, opened 2024-06-27
docs([DST-485]): Create reusable component for Do's and Don'ts

OsamaAbdellateef:


docs/ui/Do.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Image } from '@/ui/Image';
+
+interface Props {
+  img: string;
+  description: string;
+}
+
+export const Do = ({ description, img }: Props) => {
+  return (
+    <div className="border-border flex flex-col border">
```
sebald, line 12: use a card as a rapper? this would fit niceley with all the other stuff we have :)

OsamaAbdellateef:
Also what do you think to have two components Do and Dont 
I could make them into one component but in terms of code it didn't look great and a lot of conditions @sebald  @sarahgm

docs/ui/Do.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <Image width={700} height={900} src={img} alt="best practice" />
```
sebald, line 13: Maybe this can be a `children` so we can use something different here if we want to

docs/ui/Dont.tsx
```diff
@@ -0,0 +1,30 @@
+import { Text } from '@marigold/components';
```
sebald, line 1: I think we can put them in one file, makes it easier to find them and we need to adjust both anyway when we need to change something.

sebald:
> Also what do you think to have two components Do and Dont 
> I could make them into one component but in terms of code it didn't look great and a lot of conditions @sebald  @sarahgm 

Yes that would be awesome

docs/ui/AdviceCard.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { ReactNode } from 'react';
+
+import { Card, Headline, Text } from '@marigold/components';
+
+import { cn } from '../../packages/system/dist';
+
+interface Props {
+  description: string;
+  type: 'DO' | "DON'T";
```
sarahgm, line 9: I think it's a bad idea to write it uppercase and with a ' because its more error-prone, I would think

sarahgm: and you can also change type to variant?

sarahgm:
how do we create actual do and don'ts? I mean the content of it

docs/ui/Do.tsx
OsamaAbdellateef: Now there is discussion about the Card to use it or not 
https://marigold-docs-git-do-card-marigold.vercel.app/components/content/card?vercelThreadId=kuiD6
@sebald

.changeset/thin-suits-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@marigold/docs': patch
```
sebald, line 2:
```suggestion
'@marigold/docs': minor
```

sebald (approved):

PR #3979, opened 2024-07-01
Fix: fix some props and add command to build step

sarahgm:


docs/scripts/build-component-props.mjs
```diff
@@ -13,6 +13,7 @@ const parser = reactDocgenTypescript.withCustomConfig('./tsconfig.json', {
     'SelectComponent',
     'ComboBoxComponent',
     'RadioComponent',
+    'SelectListComponent',
```
sebald, line 16: What does this do?

sebald:
Can we maybe move props JSON to the registry and remove it from the repo?

docs/scripts/build-component-props.mjs
sarahgm: this is because this are compound components with the collection item thing i think and thats why they have to be in this, I can explain it tomorrow to you :) HArd to explain for me

sarahgm:
> Can we maybe move props JSON to the registry and remove it from the repo?

what do you mean?

sebald (approved):

sebald (approved):

sebald (approved):

PR #3985, opened 2024-07-05
fix: adjust the triggerwidth of the autocomplete

sarahgm:


packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -43,13 +45,17 @@ const AutocompleteInput = ({
...
+  // needed to get the triggerwidth on the right button
+  const classNames = useClassNames({ component: 'ComboBox' });
 
   return (
     <SearchInput
       ref={ref}
-      className={{
-        action: state?.inputValue === '' ? 'hidden' : undefined,
-      }}
+      // @ts-ignore
```
sebald, line 54: why the ts-ignore?

sebald (changes requested):

packages/components/src/Autocomplete/Autocomplete.tsx
sarahgm: because input expects an object but we have just a string. This was my clever idea :D Do you have another?

sebald: you have to use an objecht here, otherwhise the `hidden` class will not be added anywhere. you can also use the `cn` function on the `action` or `input`. I think the classes from the `Combobox` should be on the input anyway? 🤔

sebald (approved):

PR #4023, opened 2024-07-18
docs: refa `DonAndDonts` to work without an image

sebald:


docs/content/__internal__/index.mdx
```diff
@@ -0,0 +1,64 @@
...
+        src="https://upload.wikimedia.org/wikipedia/commons/1/18/Dog_Breeds.jpg"
+        alt="Don't"
+      />
+    </Do.Figure>
+    <Do.Description>
+      Use one primary call to action to help people to proceed
+    </Do.Description>
+  </Do>
+  <Dont>
+    <Do.Figure>
```
sarahgm, line 25: Shouldn't there be used `Dont.Figure` and `Dont.Description`?

sarahgm (changes requested):

docs/content/__internal__/index.mdx
sarahgm: is there a difference for them?

sarahgm: Can't we just use `Figure` and `Description`?

sebald: we could but then hey must be reusable
and yes it should be the dont figure!

sarahgm: I don't understand, you did it accedently wrong and still do.description works for dont component :D

sarahgm (approved):

PR #4028, opened 2024-07-21
docs: Introduce an appearance demo

sebald:
# Description

We are currently trying to improve our component documentation. One thing we noticed is that we are always including the appearance props and then the first demos are basically applying some of the appearance props.

See DST-499

Note: This is just the first iteration. There are additional things we could/should tackle.

- What to do if there is no variant and/or size? Basically we need to tell the user this by e.g. disabling the controls and adding a help text

- Do we want to include the table in the component somehow?

- Do we want to merge stuff from the `ComponentDemo` with this component?

# What should be tested?

Go to the button documentation page.

-

# Reviewers:

@sarahgm


docs/ui/AppearanceTable.tsx
```diff
@@ -1,5 +1,6 @@
 'use client';
 
+import { getAppearance } from '@/lib/utils';
```
sarahgm, line 3: in this file are some lint warnings

themes/theme-docs/src/components/Field.styles.ts
```diff
@@ -0,0 +1,21 @@
+import { ThemeComponent, cva } from '@marigold/system';
+
+export const Field: ThemeComponent<'Field'> = cva('grid gap-y-0.5', {
+  variants: {
+    variant: {
+      default: 'gap-y-0.5',
```
sarahgm, line 6: this is already in default styles, why not use this?

themes/theme-docs/src/components/Select.styles.ts
```diff
@@ -0,0 +1,57 @@
+import { ThemeComponent, cva } from '@marigold/system';
+
+import { inputBackground, inputBox, inputSpacing } from './Input.styles';
+
+export const Select: ThemeComponent<'Select'> = {
+  icon: cva('text-secondary-400', {
+    variants: {
+      variant: {
+        default: '',
```
sarahgm, line 9: why is it empty?

sarahgm (changes requested):

sarahgm (approved):

PR #4036, opened 2024-07-23
docs[DST-451]:Guidelines for Table component

sarahgm:
# Description

To prevent misusing from Marigolds Table component, I reworked the current Table page on the documentation

# What should be tested?

You should have a look on the content of the table page, is it clear? is it understandable?

Through the context of misusing - do you think they understand what our table is for?

# Reviewers:

Examples:
@marigold-ui/developer
@marigold-ui/designer & @KadaStrophe


KadaStrophe (dismissed):
See vercel comments

sebald (approved):

PR #4039, opened 2024-07-24
chore[DST-487]: align core styles to marigold

sarahgm:
# Description

- Added `<Tabs>` variant sunken from core extended theme
- Added group className on the button styles from `<Accordion>`, to allow changing classNames based on the expanded state
- Mastermark variant for tables NOT added
- Dialog legacy variant NOT added

# What should be tested?

Have a look on the extended theme from core (link in issue). Is it ok? Should there be more in Marigold? Should there be less?

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer & @KadaStrophe


sarahgm:
> @sarahgm are they some reasons why not also adding mastermark and dialog custom themes?

Dialog because it is a legacy version @sebald  added it only because it looks bad with the design. And we should not use `bg-bg-surface` as background normally for overlays.

Mastermark because I think we need to find a better solution here, since they use it only as tables variant, which doesn't mean they use it everytime with tables, also with divs, columns etc,
TLDR:  for me this topic seems more general not only table specific

aromko (approved):

sebald:
can we merge this?

aromko:
> can we merge this?

From my side is fine but I don't know who @sarahgm would like to handle the info to alex. So she can decide when to merge.

PR #4041, opened 2024-07-25
docs: improve prop and appearance table styles

sebald:
# Description

Moved the tables out of the card to improve visual hierachy.

# What should be tested?

Open a component page. Like it!?

no.

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


docs/ui/PropsTable.tsx
sarahgm, line 1: there is a Lint warning @sebald

sarahgm (changes requested):

sarahgm:
@sebald conflics

sarahgm (approved):

PR #4042, opened 2024-07-25
Table-empty-state

sebald:
# Description

Added the possibility to render an empty state in a table.

# What should be tested?

checkout storybook and tests

no

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer

### DIFF
--- a/packages/components/src/Checkbox/Checkbox.tsx
+++ b/packages/components/src/Checkbox/Checkbox.tsx
@@ -140,7 +140,7 @@ const _Checkbox = forwardRef<HTMLLabelElement, CheckboxProps>(
 
     const { labelWidth } = useFieldGroupContext();
     const group = useCheckboxGroupContext();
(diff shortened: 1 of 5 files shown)


docs/ui/AppearanceTable.tsx
```diff
@@ -19,58 +19,48 @@ export const AppearanceTable = ({ component }: AppearanceTableProps) => {
 
   const appearances = getAppearance(component, themes[current]);
 
+  if (appearances?.variant?.length === 0 && appearances?.size?.length === 0) {
```
aromko, line 22: Can't we use now the emptyState prop instead of this?

sebald: yes, this is what I plan to do in a separate PR! :) 

(this is actually why I wanted this feature ^^)

aromko (approved):

PR #4051, opened 2024-07-29
updating eslint

OsamaAbdellateef:
# Description

[ ] This change requires a UI-Kit update - inform Alex Tirado

# What should be tested?

# Reviewers:


config/eslint/index.js
```diff
@@ -11,19 +17,33 @@ module.exports = {
     },
   },
   rules: {
-    // Disabled cause it doesn't currently work as of 01/02/2021
     'no-redeclare': 'off',
```
sebald, line 20: Why deleting the comments? Can we enable these rules again?

config/eslint/index.js
```diff
@@ -11,19 +17,33 @@ module.exports = {
...
   },
   parserOptions: {
     babelOptions: {
       parserOpts: {
-        // Allow imports like `import pkg from './package.json' assert { type: 'json' };`
         plugins: ['importAssertions'],
       },
     },
   },
+  ignorePatterns: ['docs/.contentlayer/generated/ContentPage/_index.mjs'],
```
sebald, line 32: This should not be part of the shareable config, right? It's very specific to our use case

sebald (changes requested):

sebald:
@OsamaAbdellateef  conflicts! :)

config/eslint/index.js
OsamaAbdellateef: I am so sorry, did notice or notified by this @sebald

sarahgm: @OsamaAbdellateef did you already resolve this?

aromko: @OsamaAbdellateef Is the change pushed? Didn't see it 😅

aromko (dismissed):

config/eslint/index.js
aromko: @OsamaAbdellateef Thanks for moving it into the global eslint. Now we merge it and should have an eye on it if all rules are applied.

PR #4065, opened 2024-08-01
feat: add `secondary` variant to links and the introduce a secondary core button

sebald:
# Description

Added `secondary` variants to the `<Link>` component so it can be effectively used as a secondary "button". Also had to add a `secondary` button to the core ... because there is none.

# What should be tested?

Does a link and button with `variant=secondary` look identical :D 

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


themes/theme-b2b/src/components/Link.styles.ts
```diff
@@ -6,6 +6,10 @@ export const Link: ThemeComponent<'Link'> = cva(
     variants: {
       variant: {
         menuItemLink: 'text-text-base hover:no-underline',
+        secondary: [
```
sarahgm, line 9: can you add this to storybook please?

sarahgm (changes requested):

themes/theme-b2b/src/components/Link.styles.ts
sebald: done

aromko, line 9: There is a test in packages/components/src/Link/Link.test.tsx which could be adjusted. And I think there is a variant used which isn't actual any more?

sebald: If you are referring to the "second" the theme is just a fixture for testing. We don't have any real themes in the tests.

sarahgm (approved):

PR #4067, opened 2024-08-01
docs: make props table more awesome

sarahgm:
# Description

TODO:

- [x] fix some long props (e.g. `width`)
- [ ] sometimes it renders nothing, what is the default then?

- [] This change requires a UI-Kit update - inform Alex Tirado (@tirado-rx)

# What should be tested?

# Reviewers:

Examples:
@marigold-ui/developer
@marigold-ui/designer


docs/scripts/build-component-props.mjs
```diff
@@ -44,13 +199,18 @@ const files = await globby([
   `!${systemDir}/**/*.ts`,
 ]);
 
+//TODO: remove
```
sebald, line 202: Ehm do we need to remove this? :D

sebald (changes requested):

sebald (approved):

PR #4069, opened 2024-08-01
docs: add disabled states for appearancedemo

sarahgm:
# Description

added disabeld state and message for appearancedemo component, you can see it in table page e.g

# Reviewers:
@marigold-ui/developer
@marigold-ui/designer


docs/ui/AppearanceDemo.tsx
```diff
@@ -56,6 +69,16 @@ export const AppearanceDemo = ({
         visual style and dimensions of the component, available values are based
         on the active theme.
       </p>
+      {isVariantOrSizeMissing ? (
```
sebald, line 72: I think this message should be part of the demo card. not be outside of it. Any yes, then it can not be a `SectionMessage` :D I think a hint text, that looks something like a help text is more appropiate here.

Screenshot 2024-08-02 at 08 49 28

docs/ui/AppearanceDemo.tsx
```diff
@@ -48,6 +49,18 @@ export const AppearanceDemo = ({
       children
     );
 
+  let messageTitle = '';
```
sebald, line 52: I think this is a bit messy with all the `else if`. Also, `isVariantOrSizeMissing` can be inferred by just checking the `messageTitle` (why is it a title btw 😜).

Alternatively: Is it enough to just put a more generic message here? Like "Appearance options are not available with the current theme."

themes/theme-docs/src/components/Input.styles.ts
```diff
@@ -23,6 +23,7 @@ export const inputSpacing = {
 // States
 // ---------------
 export const inputFocus = '';
-export const inputDisabled = '';
+export const inputDisabled =
+  'disabled:cursor-not-allowed disabled:text-text-disabled';
```
sebald, line 27: needs also a light grey background so it is easier distinguishable from the an enabled input.

sebald (changes requested):

themes/theme-docs/src/components/Input.styles.ts
sarahgm: please have a loook again

sebald: nice

sebald (approved):

PR #4075, opened 2024-08-04
docs: revise `Link` page

sebald:


docs/content/components/content/link/link-appearance.demo.tsx
```diff
@@ -0,0 +1,9 @@
+import { Link } from '@marigold/components';
+
+export default () => (
```
sarahgm, line 3: here are the props missing

sarahgm (approved):

PR #4088, opened 2024-08-06
docs: persist theme when using cmdk

sebald:
# Description

Our cmdk did not keep the `?theme=` search param which caused some dumb issues with vercel comments :D 

This makes sure it is always there!


sarahgm:
Is it intended that its not persist when using cmdk and also a subsection?

sebald (commented):
> Is it intended that its not persist when using cmdk and navigating to a subsection?  

@sarahgm nope! rixed

sarahgm:
> > Is it intended that its not persist when using cmdk and navigating to a subsection?
> 
> @sarahgm nope! rixed

you broke the url -> /introduction/design-tokens?theme=corewhat-makes-marigold-differenttypography it doesn't jump to the section

sebald:
@sarahgm third time should be the charm! :D

docs/app/_components/CommandItems.tsx
```diff
@@ -203,11 +203,16 @@ export const PagesItem = ({
   setPages,
 }: PagesItemProps) => {
   const router = useRouter();
-  const goto = (slug: string) => {
-    router.push(`/${slug}`);
+  const params = useSearchParams();
+
+  const goto = ({ slug, hash }: { slug: string; hash?: string }) => {
```
aromko, line 208: `hash` irritates me. I thought first it's a hashed value. I would call it subpage or anchor. Would be more clear to me.

sarahgm (dismissed):

docs/app/_components/CommandItems.tsx
sebald: sadly that is its name :D see https://developer.mozilla.org/en-US/docs/Web/API/URL/hash

aromko (approved):

PR #4092, opened 2024-08-07
docs: cards in teaser list take full width and the cards will be of equal height

sebald:
see title :D

### DIFF
--- a/themes/theme-docs/src/components/Card.styles.ts
+++ b/themes/theme-docs/src/components/Card.styles.ts
@@ -2,7 +2,7 @@ import { ThemeComponent, cva } from '@marigold/system';
 
 export const Card: ThemeComponent<'Card'> = cva(
   [
-    'bg-bg-surface border-secondary-300 relative overflow-hidden rounded-xl border shadow',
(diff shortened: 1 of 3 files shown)


themes/theme-docs/src/components/Card.styles.ts
```diff
@@ -2,7 +2,7 @@ import { ThemeComponent, cva } from '@marigold/system';
 
 export const Card: ThemeComponent<'Card'> = cva(
   [
-    'bg-bg-surface border-secondary-300 relative overflow-hidden rounded-xl border shadow',
+    'bg-bg-surface border-secondary-300 relative w-full overflow-hidden rounded-xl border shadow',
```
sarahgm, line 5: should they really take full width? Looks a bit weird when only one item is there

sarahgm (changes requested):

themes/theme-docs/src/components/Card.styles.ts
sebald: fixed

sarahgm (approved):

PR #4093, opened 2024-08-07
chore: make <Rectangle> more placeholder like and remove <Squircle>

sarahgm:
# Description

Through working on  revising`<Inline>` examples, I want to have a component for all the layout stuff, I though t Rectangle would be a good choice, I updated the styles a bit. Also I noticed that `<Squircle>` doesn't support borders and is difficult to handle, so I decide to delete it and use rectangles instead. Let me know what you think.

# What should be tested?

Have a look on Layout page e.g. there you can see the components.

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


docs/ui/Rectangle.tsx
```diff
@@ -2,13 +2,18 @@ import type { ReactNode } from 'react';
...
-export const Rectangle = ({ children, height }: RectangleProps) => (
+export const Rectangle = ({
+  children,
+  height,
+  width = '100%',
+}: RectangleProps) => (
   <div
-    className="w-full rounded-sm bg-gradient-to-r from-[hsl(29,_37%,_70%)] to-[hsl(29,_37%,_40%)] shadow"
-    style={{ height: height }}
+    className="rounded-sm border-2 border-dashed border-[hsl(29,_37%,_40%)] bg-[hsl(29,_37%,_70%)]"
```
sebald, line 15: I think if we want it to look more like a placeholder, it should not be that far I'm when I did the rectangles, I made them brown because of the docs theme at this point. I think now it should be just gray?

Maybe be can use transparent with the "neutral" tailwind color. So it looks good in all themes?

sarahgm: We can, I don't want to make a big difference because I think there was an intension :D

sarahgm: have a look again

sebald (approved):

PR #4151, opened 2024-09-02
chore: add internal page to command menu

sarahgm:
# Description

add internal page to command menu

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer

### DIFF
--- a/docs/app/_components/CommandItems.tsx
+++ b/docs/app/_components/CommandItems.tsx
@@ -1,4 +1,4 @@
-import { links, themeswitch } from '@/lib/config';
+import { internal, links, themeswitch } from '@/lib/config';
 import { iterateTokens } from '@/lib/utils';
 import { Icons, cn } from '@/ui';
 import { Command, CommandGroup, useCommandState } from 'cmdk';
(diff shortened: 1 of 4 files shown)


docs/app/_components/CommandItems.tsx
```diff
@@ -255,3 +255,33 @@ export const PagesItem = ({
     </CommandGroup>
   );
 };
+
+export const InternalPage = ({ classNames, setOpen }: ChangeOpenItemProps) => {
+  const router = useRouter();
+  const params = useSearchParams();
+
+  const goto = ({ slug, hash = '' }: { slug: string; hash?: string }) => {
```
aromko, line 263: Can't we reuse the goto function from PagesItem? Only `setPages([]);` is missing. Don't we need this?

sarahgm: we don't need setPages because its for the subitems

sarahgm: but I will have a look again and see if I can reuse it :)

sarahgm: can you have a look again please? :) @aromko

aromko (approved):

PR #4154, opened 2024-09-04
chore: add coveragereport in command

sarahgm:
# Description

see uncovered lines

# Reviewers:

@marigold-ui/developer

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "registry": "pnpm --filter @marigold/docs registry",
     "build-component-props": "pnpm --filter @marigold/docs build-component-props",
     "test": "jest --config jest.config.js",
-    "test:coverage": "jest --config jest.config.js --coverage --coverageReporters html",
+    "test:coverage": "jest --config jest.config.js --coverage --coverageReporters text",
     "test:ci": "jest",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -53,7 +53,7 @@
     "registry": "pnpm --filter @marigold/docs registry",
     "build-component-props": "pnpm --filter @marigold/docs build-component-props",
     "test": "jest --config jest.config.js",
-    "test:coverage": "jest --config jest.config.js --coverage --coverageReporters html",
+    "test:coverage": "jest --config jest.config.js --coverage --coverageReporters text",
```
aromko, line 56: Why this change is necessary?

sarahgm: because otherwise you can't see it? Or were is it? :D

sarahgm: I just did this to have the report back, see uncovered lines, etc

aromko: But now you just see it in the console where lines are uncovered but don't see the details right?

And with the flag `html` a directory is generated called coverage. The index.html you can open in the browser and have an interactive website with filter/search possibilities etc. Do you know that? So for me personally the html version is more handy. What do you think?

sarahgm: I didn't know that xD

PR #4160, opened 2024-09-10
[DST-573]: Dialog Sizes

OsamaAbdellateef:
# Description
In our current implementation the size of a dialog is based on its contents (a.k.a https://developer.mozilla.org/en-US/docs/Glossary/Intrinsic_Size) . While this is a good default, there should also be the possibility to have different sizes for dialogs, for example when the content of a difalog itself does not have an intrinsic size.

You can check this for more info : https://reservix.atlassian.net/browse/DST-378?atlOrigin=eyJpIjoiOTUyMTE3Mzg1NTQ1NDRmYmFmYmNkOGQ5ZjFkODhkYjAiLCJwIjoiaiJ9

[] This change requires a UI-Kit update - Alex Tirado (@tirado-rx)

# What to test ?
-Test sizes in b2b them 

# Reviewers:
@marigold-ui/developer


themes/theme-b2b/src/components/Dialog.styles.ts
```diff
@@ -3,8 +3,19 @@ import { ELEVALTION_RING } from '../mixins';
...
-    ELEVALTION_RING,
-  ]),
+  container: cva(
+    [
+      'font-body bg-bg-surface-overlay shadow-surface-overlay rounded-sm px-8 pb-8 pt-4',
+      ELEVALTION_RING,
+    ],
+    {
+      variants: {
+        size: {
```
sarahgm, line 13: why has only b2b theme different sizes?

sarahgm (changes requested):

sarahgm:
did you use `--visual-viewport-height` like written in the ticket? Just wondering :)

OsamaAbdellateef:
> did you use `--visual-viewport-height` like written in the ticket? Just wondering :)

Ah i didn't use it because when i inspected that specific element i noticed that it  inherited `--visual-viewport`  from its parent @sarahgm 
But you can re-check to make sure that it is working

sarahgm:
Ah I also noticed, this change require a UI kit update right? @OsamaAbdellateef

sarahgm (approved):

PR #4164, opened 2024-09-17
[DST-462]Refactor <Dialog> component use Grid Areas

OsamaAbdellateef:
> [!WARNING]  
> This PR contains a breaking change! See below. This should be merged with #4187

# Description
Our current API for the <Dialog> component is rather cumbersome and has some hard limitation when it comes to composability. We should improve this and align its API with the API of our other components (Components have their children “attached” e.g. Table.Cell)

- [ ] This change requires a UI-Kit update -  Alex Tirado (@tirado-rx)

# What should be tested?

-The title and close button are vertically aligned 
-The dialog is more composable 
-The order of dialog sub-components always are the same 

# Reviewers:

@marigold-ui/developer


packages/components/src/Dialog/Dialog.tsx
```diff
@@ -81,17 +116,22 @@ const _Dialog = ({
...
-      </>
+      {closeButton && <CloseButton className={classNames.closeButton} />}
+      {children}
     </Dialog>
   );
 };
 
 _Dialog.Trigger = DialogTrigger;
-_Dialog.Headline = DialogHeadline;
+_Dialog.Title = DialogTitle;
```
sarahgm, line 132: would you also write some tests for the new components?

packages/components/src/Dialog/Dialog.stories.tsx
```diff
@@ -72,20 +67,23 @@ export const Form: Story = {
...
                 <TextField label="Username" />
                 <TextField label="Password" type="password" />
-                <Inline space={4}>
-                  <Button variant="ghost" onPress={close}>
-                    Cancel
-                  </Button>
-                  <Button variant="primary">Login</Button>
-                </Inline>
-              </Stack>
+              </Dialog.Content>
```
sarahgm, line 83: from this example I'm not sure how  dialog actions should be used.

sarahgm: and how its different from other parts of the dialog

sarahgm (changes requested):

.changeset/early-sloths-work.md
```diff
@@ -0,0 +1,7 @@
+---
+'@marigold/components': major
+'@marigold/theme-core': major
+'@marigold/theme-b2b': major
+---
+
+Refactoring dialog, by improving its API so it is more composable
```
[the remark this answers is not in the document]
sarahgm: **BREAKING CHANGE**: 

like this and than whats new and what changed

sarahgm:
also documentation is missing

packages/components/src/Dialog/Dialog.stories.tsx
OsamaAbdellateef: There are two reasons behind this : 
1-  Layout , to make sure that dilaog actions take the same place or area 
2-  Semantic clarity : it defines the purpose of the section as containing dialog actions .
@sarahgm

sarahgm: And where should this actions be always placed? maybe with a better story it would be easier for me to understand

sarahgm: With docs I mean how to use the different dialog components, and that they exists. It will be a breaking change so it should be clear what it is and what needs to change.

sarahgm:
Ah I also noticed, this change require a UI kit update right? @OsamaAbdellateef

OsamaAbdellateef:
> Ah I also noticed, this change require a UI kit update right? @OsamaAbdellateef

Yes you are right @sarahgm  👏🏻

.changeset/early-sloths-work.md
```diff
@@ -0,0 +1,13 @@
+---
+'@marigold/components': major
+'@marigold/theme-core': major
+'@marigold/theme-b2b': major
+---
+
+**Breaking changes**
+
+- `Dialog.Content`, `Dialog.Title`, and `Dialog.Footer` have been introduced for better organization and flexibility.
```
sarahgm, line 9: add angle brackets around the components :D `<Dialog.Content>`

.changeset/early-sloths-work.md
```diff
@@ -0,0 +1,13 @@
+---
+'@marigold/components': major
+'@marigold/theme-core': major
+'@marigold/theme-b2b': major
+---
+
+**Breaking changes**
+
+- `<Dialog.Content>`, `<Dialog.Title>`, and `<Dialog.Footer>` have been introduced for better organization and flexibility.
```
sarahgm, line 9: also that Dialog.Headline is now Dialog.Title?

sarahgm (approved):

PR #4187, opened 2024-10-01
refa(listbox): Allow sections in Combobox + Autocomplete, adjust Section API

sebald:
> [!WARNING]  
> This PR contains a breaking change! See below. This should be merged with #4164 

# Description

- Added the possibility to use sections with `Combobox` and `Autocomplete`
- Refactored the `Section` (from `Listbox`) to fit our API, no need for the extra `Header` anymore. Instead you can do `<Select.Section header="My header">`, same for the other components
- Renamed `Item` to `Option` in `Combobox` and `Autocomplete` to align with `Select`
- Updated the docs for `Select`, `Combobox` and `Autocomplete`
- Updated Storybook for `Select`, `Combobox` and `Autocomplete` with section stories

- [x] Requires a change in the UI kit @tirado-rx 

# What should be tested?

Using sections with `Select`, `Combobox` and `Autocomplete`

nope.

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer

Closes [DSTSUP-89]

[DSTSUP-89]: https://reservix.atlassian.net/browse/DSTSUP-89?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/Multiselect/Multiselect.tsx
+++ b/packages/components/src/Multiselect/Multiselect.tsx
@@ -111,9 +111,9 @@ export const Multiselect = ({
         {...props}
       >
         {unselected.map((item: MultiSelectItemProps) => (
-          <ComboBox.Item key={item.id} id={item.id}>
+          <ComboBox.Option key={item.id} id={item.id}>
             {item.children}
-          </ComboBox.Item>
+          </ComboBox.Option>
         ))}
       </ComboBox>
     </div>
--- a/.changeset/bright-masks-love.md
+++ b/.changeset/bright-masks-love.md
@@ -5,14 +5,15 @@
 "@marigold/theme-b2b": patch
 ---
 
-refa(listbox): Allow sections in `Combobox` and `Autocomplete`, adjust Section API in `Select`, `Combobox` and `Autocomplete`.
+refa(listbox): Allow sections in `<Combobox>` and `<Autocomplete>`, adjust Section API in `<Select>`, `<Combobox>` and `<Autocomplete>`.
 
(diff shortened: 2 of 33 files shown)


packages/components/src/Multiselect/Multiselect.tsx
```diff
@@ -111,9 +111,9 @@ export const Multiselect = ({
         {...props}
       >
         {unselected.map((item: MultiSelectItemProps) => (
-          <ComboBox.Item key={item.id} id={item.id}>
+          <ComboBox.Option key={item.id} id={item.id}>
```
aromko, line 114: I wonder if you should also change MultiSelect.Item to Multiselect.Option?

And I can't use Sections now in MultiSelect or?

And what about using also Section as selectable Item? (asking because it's used in the customer-search - doesn't mean I like this)

sarahgm (approved):

packages/components/src/Multiselect/Multiselect.tsx
sarahgm: @sebald

sebald: Since multi select is not a real component yet, I would not touch it.

Selectable Section: we talked about it. I think this would require some research, if we want to allow this pattern. Also it's an enhancement, we can add it independent of a breaking change

aromko (approved):

PR #4191, opened 2024-10-02
chore(deps): update react-aria to 1.4

sebald:
# Description

Updat to the latest `react-aria` releases.

# What should be tested?

Is everything still workin!? :-O

Especially test everything that has an overlay.

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


packages/components/src/Button/Button.test.tsx
```diff
@@ -41,15 +41,9 @@ test('supports base styling classes', () => {
...
-  expect(button).toMatchInlineSnapshot(`
-<button
-  class="items-center justify-center gap-[0.5ch] align-center flex disabled:bg-gray-600"
-  data-rac=""
-  type="button"
->
-  button
-</button>
-`);
+  expect(button.className).toMatchInlineSnapshot(
```
aromko, line 44: Why not the whole button anymore and just the css classes are tested?

sarahgm (approved):

PR #4198, opened 2024-10-06
feat(components): Mark layout coponents as regions for accessibility.

sebald:
# Description

Added the ability to mark layout components as regions (ARIA role). This improves accessibility by allowing assistive technologies to identify significant sections of the page, making navigation easier for users with disabilities.

# What should be tested?

Can you add these props to elements?

-

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


.changeset/fuzzy-terms-relax.md
```diff
@@ -0,0 +1,8 @@
+---
+"@marigold/components": minor
+"@marigold/types": minor
+---
+
+feat(components): Mark layout coponents as regions for accessibility.
```
sarahgm, line 6: typo components

packages/components/src/Inset/Inset.tsx
```diff
@@ -1,16 +1,19 @@
...
+import type {
+  PaddingSpaceProp,
+  PaddingSpacePropX,
+  PaddingSpacePropY,
+} from '@marigold/system';
+import type { AriaRegionProps } from '@marigold/types';
 
 export type InsetProps =
-  | {
+  | (AriaRegionProps & {
```
sarahgm, line 16: why is there a and ?

sebald: So that every of the following types also has the AriaRegionProps :)

.changeset/fuzzy-terms-relax.md
sarahgm: @sebald

sebald: where is the typo?

sarahgm (approved):

PR #4204, opened 2024-10-08
feat(components): add close button on `<SectionMessage>`

sarahgm:
## Description

- Revised the `<SectionMessage>` page according to our new component pages.
- While revising adding a close button to align the component with our feedback messages pattern. 

## What should be tested?
- if the close button closes
- if the new docs are understandable

## Reviewers
@marigold-ui/developer 
@marigold-ui/designer 

closes [DST-582](https://reservix.atlassian.net/jira/software/projects/DST/boards/134?selectedIssue=DST-582)

[DST-582]: https://reservix.atlassian.net/browse/DST-582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sebald:
I would not have 2 things in the headline, just having the `feat` here. Adding docs should be obvious and it is enought to have it in the PR description.

docs/content/components/content/section-message/section-message.mdx
```diff
@@ -1,56 +1,162 @@
...
 
+## Usage
+
+Section messages are ideal for displaying important feedback related to a specific section of the page.
+
+Unlike modal dialogs, which interrupt the user's workflow, section messages remain visible without blocking interaction with other parts of the interface.
+
+<Do>
+  <Do.Description>
+    <ul>
```
sebald, line 37: Did you not use list here because it does not have any list indicators (or whatever the dot before each item is called?)

sarahgm: what do you mean? We always just list if there are more points for a do?

sebald (approved):

PR #4216, opened 2024-10-15
docs:  ref `<Image>` in docs ui components

OsamaAbdellateef:


OsamaAbdellateef:
I don't think that this needs a changeset right ?! @sarahgm

themes/theme-docs/src/components/Dialog.styles.ts
```diff
@@ -1,7 +1,7 @@
 import { ThemeComponent, cva } from '@marigold/system';
 
 export const Dialog: ThemeComponent<'Dialog'> = {
-  closeButton: cva('absolute right-2 top-2', {
+  closeButton: cva('', {
```
sarahgm, line 4: I think you have to add this again :)

sarahgm (changes requested):

themes/theme-docs/src/components/Dialog.styles.ts
OsamaAbdellateef: I added it directly on CloseButton component @sarahgm

sarahgm: Bildschirmfoto 2024-10-21 um 13 33 24
but this doesn't look right to me? :O

sarahgm (changes requested):

themes/theme-docs/src/components/Dialog.styles.ts
sarahgm: I think we said the button should align with the headline

sarahgm (approved):

PR #4229, opened 2024-10-21
refa: remove footer from `<Dialog>` + allow styles

sebald:
# Description

# What should be tested?

# Reviewers:

Examples:
@marigold-ui/developer
@marigold-ui/designer


packages/components/src/Dialog/Dialog.tsx
```diff
@@ -105,21 +64,21 @@ const _Dialog = ({
...
 }: DialogProps) => {
   const classNames = useClassNames({ component: 'Dialog', variant, size });
-  let state = useContext(OverlayTriggerStateContext);
+  const state = useContext(OverlayTriggerStateContext);
 
-  let children = props.children;
+  const children =
+    typeof props.children === 'function'
+      ? props.children({
+          close: state?.close || (() => {}),
```
sarahgm, line 72: this is not covered by jest

PR #4230, opened 2024-10-22
bugfix([DSTSUP-92]): fix broken variants in table stories

aromko:
# Description

The Controls are not working in the Nested Columns-Variant of the Table in Storybook. 
`args` was missing in certain stories.

# What should be tested?
Have a look at all table stories. Variants should work now.

# Reviewers:

@marigold-ui/developer
@KadaStrophe


KadaStrophe (approved):
Thank you!

PR #4233, opened 2024-10-23
fix([DSTSUP-94]): Adjust date unavailable property and add styles for it on date picker

sarahgm:
# Description

- this was caused by a support ticket, changed isDateUnavailable to dateUnavailable, matching our code guidelines.
- added disabled styles for b2b and core theme
- added a test
- added a story

# What should be tested?

everything working?

# Reviewers:
@marigold-ui/developer
@marigold-ui/designer

### DIFF
--- a/.changeset/beige-snakes-develop.md
+++ b/.changeset/beige-snakes-develop.md
@@ -1,0 +1,10 @@
+---
+"@marigold/components": patch
+"@marigold/theme-b2b": patch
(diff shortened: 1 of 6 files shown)


sebald:
@sarahgm I think we need to flag this as a breaking change :D

sarahgm:
> @sarahgm I think we need to flag this as a breaking change :D

please have a look again on the changeset :)

PR #4245, opened 2024-10-28
docs: fix release date in changelogs

sarahgm:
# Description

Fixed the dates, with using github api, instead of simple-git. Because docs branch cannot read it

# Reviewers:

@marigold-ui/developer


.gitignore
```diff
@@ -25,6 +25,7 @@ out
 .component-props
 
 ## env
+docs/.env
```
sebald, line 28: Why is this necessary. the local env should be enough, no?

docs/scripts/build-changelog.mjs
```diff
@@ -1,8 +1,9 @@
+import { Octokit } from '@octokit/core';
 import path from 'path';
-import { simpleGit } from 'simple-git';
 import { fs, globby } from 'zx';
 
-const git = simpleGit();
+require('dotenv').config();
```
sebald, line 5: Maybe change to also use `.env.local` so we don't have multiple env files?

sebald (changes requested):

docs/scripts/build-changelog.mjs
sebald: (Or does this not work in vercel then?)

.gitignore
sarahgm: no didn't work for me

docs/scripts/build-changelog.mjs
sarahgm: For me it doesn't work with the local env file

sebald: But why? You can specify a different .env file for the lib.

sarahgm: my token is than undefined

docs/scripts/build-changelog.mjs
```diff
@@ -12,27 +13,68 @@ let changelogPath = await globby([
 
 console.log('📑 Generating changelogs...');
 
-const getReleaseInformation = async file => {
-  const log = await git.log({ file: path.resolve(file) });
+const getReleaseData = async () => {
```
aromko, line 16: I asked Copilot to simplify this code snippet from l.16-77 and thats the result

```
const getReleaseData = async () => {
  const octokit = new Octokit({ auth: GITHUB_TOKEN });
  const response = await octokit.request('GET /repos/marigold-ui/marigold/releases', {
    headers: { 'X-GitHub-Api-Version': '2022-11-28' },
  });

  return response.data.reduce((releaseDates, release) => {
    const [, version] = release.name.match(/@([^@]+)$/) || [];
    if (!version) return releaseDates;

    if (!releaseDates.has(version)) {
      const releaseDate = new Date(release.published_at);
      const daysDifference = (new Date() - releaseDate) / (24 * 60 * 60 * 1000);
      releaseDates.set(version, {
        releaseDate,
        badge: daysDifference < 30 ? 'new' : undefined,
      });
    }
    return releaseDates;
  }, new Map());
};

const getReleaseInformation = (sourceText, releaseDates) => {
  const versionRegex = /## (\d{1,2}\.\d{1,2}\.\d{1,2})/gm;
  const versions = sourceText.match(versionRegex)?.map(v => v.replace('## ', '')) || [];

  if (!versions.length) {
    console.log('No versions found in the source text.');
    return [];
  }

  return versions.map(version => ({
    version,
    releaseDate: releaseDates.get(version)?.releaseDate || null,
    badge: releaseDates.get(version)?.badge || undefined,
  }));
```

Changes made:

- Combined the Map initialization and population in one step within reduce.
- Removed unnecessary intermediate variables.
- Simplified the calculation of daysDifference.
- Used optional chaining for safer access in getReleaseInformation.

sebald:
Can we remove the changelog files from git in this PR too?

docs/scripts/build-changelog.mjs
sebald: can you ask him to not use `.reduce`? :D I always find this makes the code harder to comprehend

sebald: I gave this chatGPT 4o:


```js
const getReleaseData = async () => {
  const octokit = new Octokit({ auth: GITHUB_TOKEN });
  const response = await octokit.request('GET /repos/marigold-ui/marigold/releases', {
    headers: {
      'X-GitHub-Api-Version': '2022-11-28',
    },
  });

  const releases = response.data;
  const releaseDates = new Map();

  releases.forEach((release) => {
    const versionMatch = release.name.match(/@([^@]+)$/);
    const version = versionMatch ? versionMatch[1] : null;
    if (!version) return;

    if (!releaseDates.has(version)) {
      const releaseDate = new Date(release.published_at);
      const daysSinceRelease = calculateDaysSince(releaseDate);
      const badge = daysSinceRelease < 30 ? 'new' : undefined;

      releaseDates.set(version, {
        releaseDate,
        badge,
      });
    }
  });

  return releaseDates;
};

const calculateDaysSince = (date) => {
  const today = new Date();
  const timeDifference = today - date;
  const millisecondsInADay = 24 * 60 * 60 * 1000;
  return Math.round(timeDifference / millisecondsInADay);
};
```

### Changes Made:
1. **Extracted Regular Expression Match**: Stored the regex match result in `versionMatch` to simplify `version` extraction and improve readability.
2. **Renamed Variables for Clarity**:
   - Changed `aDayInMs` to `millisecondsInADay` for better understanding.
3. **Extracted Day Calculation Logic**: Moved the calculation of the number of days since the release into a separate function, `calculateDaysSince`, for clarity and reuse.
4. **Simplified Badge Logic**: Directly assigned `badge` during the calculation.

docs/scripts/build-changelog.mjs
sarahgm: how can I test it in vercel? I added a env variable in vercel for it. And the env file from me is created in docs/.env

sarahgm: https://marigold-docs-3meutb371-marigold.vercel.app/releases/packages/components/release?theme=core

docs/scripts/build-changelog.mjs
sarahgm: This is exactly the same except you moved. the calculatoin of date difference out of it right?

sarahgm: I changed it can you have a look again?

docs/scripts/build-changelog.mjs
sebald: https://www.npmjs.com/package/dotenv#path

```
require('dotenv').config({ path: ['.env.local', '.env'] })
```

sarahgm: I just have one env file.....

sarahgm: By default, config will look for a file called .env in the current working directory.

sebald (approved):

PR #4268, opened 2024-10-29
feat(dialog): Introduce a dedicated button to cancel a dialog

sebald:
# Description

Make it more convenient to have a button that closes the `<Dialog>`. With this, there is less need to use the child function to access the `close` method. Instead you can now use `<Button slot="close">` to render a close button.

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


docs/content/components/overlay/dialog/dialog-appearance.demo.tsx
```diff
@@ -11,7 +11,7 @@ export default (props: DialogProps) => (
         details pre-filled, except for the date.
       </Dialog.Content>
       <Dialog.Actions>
-        <Button variant="secondary">Cancel</Button>
+        <Dialog.Cancel>Cancel</Dialog.Cancel>
```
sarahgm, line 14: Cancel should than only be used for cancel actions? 

I think its strange to write a Dialog.Cancel and than also a normal Button inside the actions?

sarahgm: Also there must not be canceled written down

sarahgm (changes requested):

packages/components/src/intl.ts
```diff
@@ -0,0 +1,13 @@
+const intl = {
```
sarahgm, line 1: do we need something like this? Should this decision be documented somewhere

sarahgm (changes requested):

packages/components/src/Dialog/Dialog.test.tsx
```diff
@@ -519,3 +519,54 @@ test('close state has a listener', async () => {
...
+    expect(cancel).toBeInTheDocument();
+  });
+
+  await user.click(cancel);
+  await waitFor(() => {
+    expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
+  });
+});
+
+test('render custom cancel button', async () => {
```
sarahgm, line 551: Is it allowed to have a custom cancel button still? Like before? (anyways the test text must be adjusted)

sarahgm (dismissed):

docs/content/components/overlay/dialog/dialog-appearance.demo.tsx
sebald: I added the `Cancel` to it does not accidentally get translated ;)

sebald: > I think its strange to write a Dialog.Cancel and than also a normal Button inside the actions?

@sarahgm suggestions?

packages/components/src/intl.ts
sebald: Usually we use the translations from `react-aria`. Would a comment in the file be enough?

packages/components/src/Dialog/Dialog.test.tsx
sebald: What should be adjusted?

sebald: What should be adjusted?

sarahgm: the test text ('render custom cancel button) because its not custom

docs/content/components/overlay/dialog/dialog-appearance.demo.tsx
sarahgm: I'm not sure if I would call it Dialog.Cancel, I think its also ok to have a dedicated cancel button for everything

packages/components/src/Dialog/Dialog.test.tsx
sebald: It's a custom label. Very nitpicky 😜

docs/content/components/overlay/dialog/dialog-appearance.demo.tsx
sebald: But for everything it not possible, this is a "dedicated" button, because you no longer have to use the callback function so you can access the `close` function. Can also name it close instead? But I wasn't sure if this is confusing with the `closeButton` prop on the `<Dialog>`

sarahgm: Just my opinion :P I mean this:
```tsx
<Dialog.Actions>
	<Button variant="close"/>
	<Button variant="primary"/>
</Dialog.Actions>
```
But its ok

sebald: Yeah, but having the `Button` depend on the `OverlayContext` seems really bad.

sebald: Any thoughts @aromko @OsamaAbdellateef

aromko: For me it's fine. Also the fallback with button.cancel.

sebald: So merge it? :P

aromko (dismissed):

sebald:
> @sebald Merging is blocked 😬

Hu. Why? Review is missing. Right?

aromko:
> > @sebald Merging is blocked 😬
> 
> Hu. Why? Review is missing. Right?

No. I think cause of requested changes of @sarahgm .

sebald:
@aromko yes, that is why I asked in the thread. you need to resolve it too :P

aromko (approved):

docs/content/components/overlay/dialog/dialog-from-menu.demo.tsx
```diff
@@ -32,10 +32,10 @@ export default () => {
                 be undone.
               </Dialog.Content>
               <Dialog.Actions>
-                <Button size="small" variant="secondary" onPress={close}>
+                <Button variant="secondary" slot="close">
                   Cancel
                 </Button>
-                <Button size="small" variant="primary" onPress={close}>
+                <Button variant="primary" onPress={close}>
```
aromko, line 38: The `onPress={close}` isn't here necessary or? So maybe as example and alert like `onPress={() => alert('deleted!'}`

aromko (dismissed):

.changeset/few-points-vanish.md
```diff
@@ -0,0 +1,8 @@
+---
+"@marigold/docs": minor
+"@marigold/components": minor
+---
+
+feat(dialog): Introduce a dedicated button to close a dialog
+
+Make it more convenient to have a button that closes the `<Dialog>`. With this, there is less need to use the child function to access the `close` method. Instead you can now use `<Button slote="close">` to render a close button.
```
sarahgm, line 8: slot not slote :)

packages/components/src/Dialog/DialogActions.tsx
```diff
@@ -12,9 +23,26 @@ export interface DialogActions {
...
   return (
     <div className={cn('[grid-area:actions]', classNames.actions)}>
-      {children}
+      <Provider
+        values={[
+          [
+            ButtonContext,
+            {
+              slots: {
+                [DEFAULT_SLOT]: {},
```
sarahgm, line 37: what is a default_slot?

packages/components/src/Select/intl.ts
```diff
@@ -1,8 +0,0 @@
-export const messages = {
```
sarahgm, line 1: why is this deleted?

sarahgm (changes requested):

packages/components/src/Dialog/DialogActions.tsx
sebald: Needed so you can but a button on there without a slot :) 

https://react-spectrum.adobe.com/react-aria/advanced.html#default-slot

packages/components/src/Select/intl.ts
sebald: we did not use this for a long time (was used when we used hooks)

docs/content/components/overlay/dialog/dialog-from-menu.demo.tsx
sebald: but the modal will not close then 🤔

aromko: ok true. In Storybook and also in the documentation is no example of doing an action with the primary button and then close the dialog 🤔 At least we should add one or?

sebald:
Hm this is approved. Merge? @sarahgm @aromko

docs/content/components/overlay/dialog/dialog-from-menu.demo.tsx
sebald: I thought this is clear 😄

PR #4269, opened 2024-10-29
docs: make `code` in `<SectionMessage>` look nicer

sebald:
# Description

Adjusted the text size of `code` blocks inside a `SectionMessage` so they are not bigger than the regular text in side the messages.

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer

### DIFF
--- a/.changeset/sweet-squids-design.md
+++ b/.changeset/sweet-squids-design.md
@@ -1,0 +1,6 @@
+---
+"@marigold/docs": patch
+"@marigold/theme-docs": patch
+---
+
+docs: make `code` in `<SectionMessage>` look nicer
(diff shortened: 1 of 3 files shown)


.changeset/sweet-squids-design.md
```diff
@@ -0,0 +1,6 @@
+---
+"@marigold/docs": patch
+"@marigold/theme-docs": patch
+---
+
+docs: make `code` in `<SectionMessage>` look nicer
```
aromko, line 6: These are the patch notes our users see now in the whats new page ( I know before they also could read it) -  so I would maybe write more whats exactly looks nicer?

aromko (approved):

aromko (approved):

PR #4291, opened 2024-11-08
refa[DST-607]: revise `<Badge>` page

aromko:
# Description

Revising badge component according to our component page guidelines.

- [x] This change requires a UI-Kit update - inform Alex Tirado (@tirado-rx), Benji (@benjirsvx)

# What should be tested?

I changed some color cause of the poor contrast. It's better now but still not the recommended contrast. You can see it in storybook choosing the accessiblity tab.

I'm open for propsals of usage guidelines. No content atm for Related or Alternatives.

# Reviewers:
@marigold-ui/developer
@marigold-ui/designer


packages/components/src/Badge/Badge.stories.tsx
```diff
@@ -34,3 +40,14 @@ export default meta;
 type Story = StoryObj<typeof meta>;
 
 export const Basic: Story = { render: args => <Badge {...args} /> };
+
+export const Icon: Story = {
+  parameters: {
```
sebald, line 45: :-O that's so cool!

CC @sarahgm @OsamaAbdellateef

sarahgm: good to know!

sebald (approved):

PR #4302, opened 2024-11-15
feat(DST-596): add pending indicator in `<Button>` component

sarahgm:
# Description

Get rid of the Xloader in the loading pattern for inline loading and use another alternative.
I used the react-aria loading indicator and add this to the button pending state. If the button is in pending then the loading indicator will always shown up. 

I added own story and own test file for the component.

I updated the documentation according to the changes which are made in button page and in loading pattern.
 
# What should be tested?

everything as expected? 

# Reviewers:
@marigold-ui/developer
@marigold-ui/designer


.changeset/wise-spoons-bow.md
```diff
@@ -0,0 +1,10 @@
+---
+"@marigold/docs": patch
+"@marigold/components": patch
+"@marigold/theme-b2b": patch
+"@marigold/theme-core": patch
+---
+
+feat(DST-596): add pending indicator in `<Button>` component
```
aromko, line 8: Here it is called _pending indicator_ and in the description _loading indicator_ and _pending state_. I would standardise it everywhere. Also in the code I would use either pending or loading. I think loading is the term we use.

aromko (changes requested):

docs/content/components/form/button/button.mdx
```diff
@@ -187,13 +188,21 @@ In general, it's best to avoid using disabled buttons. However, if they are nece
...
+    and they should never be used on non-interactive elements. For more
+    information see the [tooltip page](/components/overlay/tooltip).
   </SectionMessage.Content>
 </SectionMessage>
 
+### Pending states
+
+The `<Button>` component includes a pending property that can be used to display loading behavior. This visually indicates the progress of system operations, such as downloading, uploading, or processing tasks. The progress circle always shows an indeterminate progress to signify that the duration of the operation is unknown.
+
+This functionality is especially useful when performing actions like submitting a form or saving changes in your application that require time to complete a request. While the button is in the pending state, it becomes disabled, preventing further interaction until the operation is completed.
```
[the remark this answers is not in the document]
sarahgm: Yes the example just show one with label and one without label - the waiting time is reduced on this one for better distinguish when its ok to not use a label

.changeset/wise-spoons-bow.md
sarahgm: what do you mean?  In code the button prop is called "isPending" i replaced this to "pending" and wrote pending everywhere

docs/content/patterns/loading-states/inline-indicator.demo.tsx
```diff
@@ -6,7 +6,6 @@ import {
   Headline,
```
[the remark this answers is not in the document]
sarahgm: not sure what you mean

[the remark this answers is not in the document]
aromko: This sentence isn't right anymore or? We don't use XLoader anymore in the inline example.

.changeset/wise-spoons-bow.md
aromko: Exactly that. For my understanding we call it normally "loading" everywhere but you called prop and everything pending. Now we have a mix of these words. I would call loading everywhere cause of consistency reasons.

docs/content/components/form/button/button.mdx
```diff
@@ -187,13 +188,21 @@ In general, it's best to avoid using disabled buttons. However, if they are nece
...
+
+The button's label should clearly describe the action being performed, providing context for the current state of the process, like also mentioned in [Icons and labels](#icons-and-labels). Besides this labels are not always necessary, if the action does not take long or the provided action is clear enough you should not use a label.
+
+<ComponentDemo file="./button-pending.demo.tsx" />
```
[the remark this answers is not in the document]
aromko: ok

.changeset/wise-spoons-bow.md
sarahgm: yes because the original prop from button was called "isPending" :D Of course I can rename it. @tirado-rx  what do you think?

docs/content/patterns/loading-states/inline-indicator.demo.tsx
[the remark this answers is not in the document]
sarahgm: ah found

docs/content/components/form/button/button.mdx
```diff
@@ -187,13 +188,21 @@ In general, it's best to avoid using disabled buttons. However, if they are nece
...
   </SectionMessage.Content>
 </SectionMessage>
 
+### Pending states
+
+The `<Button>` component includes a pending property that can be used to display loading behavior. This visually indicates the progress of system operations, such as downloading, uploading, or processing tasks. The progress circle always shows an indeterminate progress to signify that the duration of the operation is unknown.
+
+This functionality is especially useful when performing actions like submitting a form or saving changes in your application that require time to complete a request. While the button is in the pending state, it becomes disabled, preventing further interaction until the operation is completed.
+
+The button's label should clearly describe the action being performed, providing context for the current state of the process, like also mentioned in [Icons and labels](#icons-and-labels). Besides this labels are not always necessary, if the action does not take long or the provided action is clear enough you should not use a label.
```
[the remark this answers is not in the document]
tirado-rx: The issue is that loading-indicator labels that are only on-screen for a brief 1-2 seconds introduce new cognitive costs. They're:

1. difficult to read (due to the use of the disabled-state colors),
2. disappear very quickly, potentially before the user has had the chance to read and understand them, and
3. change the label text, potentially requiring new considerations of the button's width.

That's why I'd prefer to encourage label-less usage of button loading states. The benefit needs to be large enough to offset the new costs.

_However_, labels (both any written `label` in the button element, whether visible or invisible, as well as the `aria-label`) still have to be readable for screen readers. That's why I wonder if [this solution](https://www.bekk.christmas/post/2023/24/accessible-loading-button) has something to offer here.

Ignoring that article for now, maybe this paragraph could work: 

> Avoid using labels to show during the loading state, as these appear only briefly and can be difficult to read before disappearing. Only use them if testing shows that users cannot understand what's happening without them.

.changeset/wise-spoons-bow.md
tirado-rx: Yes, I agree that we should change it to "loading" everywhere as much as possible. I thought @sebald mentioned this issue today in the daily, but maybe I'm mixing things up.

sarahgm: this is the same conversation like in vercel right? xD I think we can close this and discuss there? If not please reopen

docs/content/components/form/button/button.mdx
```diff
@@ -187,13 +188,23 @@ In general, it's best to avoid using disabled buttons. However, if they are nece
...
-    For more information see the [tooltip page](/components/overlay/tooltip).
-
+    and they should never be used on non-interactive elements. For more
+    information see the [tooltip page](/components/overlay/tooltip).
   </SectionMessage.Content>
 </SectionMessage>
 
+### Loading states
+
+The `<Button>` component includes a loading property that can be used to display loading behavior. This visually indicates the progress of system operations, such as downloading, uploading, or processing tasks. The progress circle always shows an indeterminate progress to signify that the duration of the operation is unknown.
```
aromko, line 198: `loading`property

packages/components/src/Button/Button.tsx
```diff
@@ -51,10 +66,12 @@ const _Button = forwardRef<HTMLButtonElement, ButtonProps>(
         className={cn(
           'inline-flex items-center justify-center gap-[0.5ch]',
           classNames,
-          fullWidth ? 'w-full' : undefined
+          fullWidth ? 'w-full' : undefined,
+          loading && '!cursor-progress'
```
aromko, line 70: this line could be tested

packages/components/src/Button/Button.tsx
```diff
@@ -51,10 +66,12 @@ const _Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
         )}
-        isDisabled={disabled}
+        isDisabled={disabled || loading ? true : false}
```
aromko, line 72: this line could be tested

.changeset/wise-spoons-bow.md
```diff
@@ -0,0 +1,14 @@
...
```
aromko, line 8: If I scan for pending they seems a lot of places that should be renamed

aromko (approved):

packages/components/src/ProgressCycle/ProgressCycle.tsx
```diff
@@ -0,0 +1,55 @@
...
+  ...props
+}: ProgressCycleProps) => {
+  let strokeWidth = 3;
+  if (size <= '24') {
+    strokeWidth = 2;
+  } else if (size >= '32') {
+    strokeWidth = 4;
+  }
+
+  let radius = `calc(50% - ${strokeWidth / 2}px)`;
```
sebald, line 24:
```suggestion
  const radius = `calc(50% - ${strokeWidth / 2}px)`;
```

packages/components/src/Button/Button.tsx
```diff
@@ -51,11 +66,24 @@ const _Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
+        isPending={loading}
         isDisabled={disabled}
       >
-        {children}
+        {loading ? (
+          <>
+            <span className="absolute">
+              <ProgressCycle />
+            </span>
+            <span className="invisible flex items-center gap-2 opacity-0">
```
sebald, line 80: either `invisible` or `opactiy`?

Also: hardcoding the gap etc here might get us in trouble if these things change in the theme (e.g. default is `gap-[0.5ch]`).

This might work (also centers the loading spinner :P):
https://play.tailwindcss.com/09rhHtzot3

sebald (changes requested):

themes/theme-core/src/components/Button.styles.ts
```diff
@@ -3,7 +3,7 @@ import { ThemeComponent, cva } from '@marigold/system';
 export const Button: ThemeComponent<'Button'> = cva(
   [
     'border-border-base bg-bg-inverted text-text-base ease-ease-out h-component cursor-pointer rounded-sm border px-4 py-0 text-sm leading-[22px] transition-all duration-200 disabled:cursor-none',
-    'disabled:border-border-base-disabled disabled:bg-bg-inverted-disabled disabled:text-text-base-disabled disabled:cursor-not-allowed',
```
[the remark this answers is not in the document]
aromko: yeah true

sebald (approved):

PR #4305, opened 2024-11-15
docs: move resources pages to introduction (and hide develoepr guidelines)

sebald:
# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


sarahgm:
@sebald can you add this to the toc? Currently when you type internal inside of it it shows the internal page, but no subpages

sebald:
@sarahgm what is "this"? :D

sarahgm:
> @sarahgm what is "this"? :D

Can you add the page "Code Guidelines" or how it is called to the command menu, because currently is not, only internal

sarahgm:
> > > @sarahgm what is "this"? :D
> > 
> > 
> > Can you add the page "Code Guidelines" or how it is called to the command menu, because currently is not, only internal @sebald
> 
> But then it will show up there, right? I think this shouldn't be a public page

I mean like we have with internal page.... if you type in internal in the command it will appear - i think it was also your idea :/

sarahgm (approved):

PR #4318, opened 2024-11-26
feat(DST-608): Make XLoader accessible and docs update

sebald:
# Description

- Refactored the `<XLoader>` component to be more accessible
- Updated the `<XLoader>` page
- Added an "venue" API to build demos and examples

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


docs/app/api/venues/route.ts
```diff
@@ -0,0 +1,36 @@
...
+import { z } from 'zod';
+import type { NextRequest } from 'next/server';
+import { NextResponse } from 'next/server';
+
+const schema = z.object({
+  delay: z.string().regex(/^\d+$/, 'delay must be a number').transform(Number),
+  q: z.string().default(''),
+});
+
+// GET
```
aromko, line 12: I think this comment is useless. The method has the same name.

docs/app/api/venues/route.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  if (!params.success) {
+    return NextResponse.json(params.error.flatten().fieldErrors, {
+      status: 400,
+    });
+  }
+
+  const { q, delay } = params.data;
+
+  // Poor mans matching
```
aromko, line 27: I would also remove this comment.

docs/content/components/content/xloader/xloader-inline.demo.tsx
```diff
@@ -0,0 +1,107 @@
...
+          List of Venues
+        </Text>
+        <Split />
+        <Button
+          variant="primary"
+          onPress={() =>
+            queryClient.invalidateQueries({ queryKey: ['venues'] })
+          }
+        >
+          Reload
```
aromko, line 101: From your explanations in the check-in meeting. Maybe we should just call it Simulate page load or Simulate initial load? But maybe also not helpful.

docs/content/components/content/xloader/xloader.mdx
```diff
@@ -1,31 +1,66 @@
 ---
 title: XLoader
-caption: Should be used for loading state.
+caption: Used to indicate the progress of the task.
 ---
 
-The `<XLoader>` component, is a simple `<SVG>` that should be used for loading states.
-It should be displayed when the user is waiting for an upcoming event.
+The `<XLoader>` component visually indicates that a task is currently in progress. It can be used as a placeholder for content that is loading or to block interaction with the screen while waiting for a process to complete, such as data fetching or form submission. This ensures users are aware that something is happening in the background, preventing confusion and enhancing the user experience.
```
aromko, line 6: > It can be used as a placeholder for content that is loading or to block interaction with the screen while waiting for a process to complete, such as data fetching or form submission. This ensures users are aware that something is happening in the background, preventing confusion and enhancing the user experience.

I think thats also written in the usage part.

aromko (commented):
To the topic with changing a prop name which leads to a breaking change. We could use deprecated for this. So between to releases or more (it depends maybe) we announce "xy is now deprecated and will be removed in the next release". So could now also introduce the new prop and set the old one to deprecated. I think @sarahgm  has had a similiar issue in another ticket. We could also talk about it in the mob.

docs/content/components/content/xloader/xloader-inline.demo.tsx
sebald: "trigger reload"?

docs/content/components/content/xloader/xloader.mdx
sebald: updated. not sure what else to write here honestly 🙈

aromko (approved):

sarahgm:
> To the topic with changing a prop name which leads to a breaking change. We could use deprecated for this. So between to releases or more (it depends maybe) we announce "xy is now deprecated and will be removed in the next release". So could now also introduce the new prop and set the old one to deprecated. I think @sarahgm has had a similiar issue in another ticket. We could also talk about it in the mob.

is this still open? Its approved?

sarahgm:
> @sarahgm I didn't change the prop values for now. Was just an idea of @aromko :)
> 
> the "problem" here is that we can not mark values as ´@depreacted` I think, just props. But we could add at least the props already?

not sure i understand it correctly but there is no new props?

sebald:
@sarahgm no. the issue is not he prop name but that "if you want a fullscreen loader you do `<XLoader mode="fullsize"/>` which might be confusing (fullsize vs fullscreen). same for the section loading that is called "inline".

PR #4320, opened 2024-11-27
feat[DST-603]: implement pagination component

aromko:
# Description

# What should be tested?

Acceptance criteria:

- [ ] A "next" and "previous" button are always visible.
- [ ] The first and last pages are always visible.
- [ ] The number of page containers increases with the total number of pages up to a max of 7.
- [ ] "Page containers" are the buttons representating pages, as well as the ellipses (“...”) that represent hidden pages.
- [ ] Ellipses are only shown if they hide 2 or more pages. If only one page would be hidden, the page is shown instead.
- [ ] One page on either side of the current page must always be visible, provided these exist

FigmaLink: https://www.figma.com/design/q4qLJc7zb2yr66sLKnqj1f/%F0%9F%8F%B5%EF%B8%8F-IN-PROGRESS%3A-Marigold-(Both-Themes)?node-id=5922-18690&node-type=canvas&t=xrZbfEimFjwybb1W-0

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


packages/components/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,148 @@
...
+                }
+              />
+            )
+          )
+        ) : (
+          <PageButton key={1} page={1} isDisabled />
+        )}
+      </div>
+
+      <PaginationButton
```
sarahgm, line 132: Maybe I would call it NavigationButton or something like this, I was at first a little bit confused about `PageButton` and `PaginationButton`.

packages/components/src/Pagination/PageButton.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { useEffect, useRef } from 'react';
+import { useButton } from '@react-aria/button';
+import { useClassNames } from '@marigold/system';
+
+interface PageButtonProps {
+  page: number;
+  isSelected?: boolean;
```
sarahgm, line 7: I know they are internal but we should follow here also our guidelines on prop naming conventions `selected` instead of `isSelected`

packages/components/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,148 @@
+import { useEffect, useState } from 'react';
+import { ChevronLeft, ChevronRight } from '@marigold/icons';
```
sarahgm, line 2: we don't want dependencies to our icons page, if i remember correctly.

packages/components/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,148 @@
...
+   */
+  controlLabels?: [string, string];
+}
+
+const _Pagination = ({
+  defaultPage = 1,
+  page,
+  totalItems,
+  pageSize,
+  onChange = () => {},
```
sarahgm, line 49: does it need to have the default?

packages/components/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,148 @@
...
+                isSelected={pageNumber === currentPage}
+                onPress={() => handlePageChange(pageNumber)}
+                registerRef={ref =>
+                  registerRef(NavigationTypes.Page, pageNumber, ref)
+                }
+              />
+            )
+          )
+        ) : (
+          <PageButton key={1} page={1} isDisabled />
```
sarahgm, line 128: is this when there is only one page?

sarahgm (dismissed):

packages/components/src/Pagination/PageButton.tsx
sarahgm: same for other

aromko: If this really a guideline I wouldn't agree to this, because using `isSelected `is more descriptive in my opinion than just `selected`. I will change it now for consistency reasons.

packages/components/src/Pagination/Pagination.tsx
aromko: jup

sarahgm (dismissed):

tirado-rx (dismissed):

sarahgm (approved):

PR #4326, opened 2024-12-02
fix(<Text>): prevent `elementType` prop from being passed down into the DOM

sebald:


packages/components/src/Text/Text.tsx
```diff
@@ -69,12 +69,17 @@ const _Text = ({
     size,
   });
 
+  /**
+   * Use `<Text>` when a `slot` is used. Make sure `elementType`
+   * prop is only used in combindation the `<Text>`.
```
aromko, line 74: typo

sebald: fixed

sarahgm (approved):

PR #4330, opened 2024-12-05
refa: improve `<Container>` and `<Breakout>`  component

sebald:
# Description

Make the `<Container>` and `<Breakout>` component more usable for real world scenarios.

- `<Container>` supporst `space` style prop
- simplify `<Container>` usage by removing unnecessary props
- simplify ´<Breakout>` and make it composable with other layout components

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


packages/components/src/Text/Text.tsx
```diff
@@ -82,6 +82,7 @@ const _Text = ({
       {...elementType}
       className={cn(
         'text-[--color] outline-[--outline]',
+        'max-w-[--maxTextWidth]', // possibly set by a <Container>
```
sarahgm, line 85: so is this only used by `<Container>` ?

packages/components/src/Breakout/Breakout.tsx
```diff
@@ -1,46 +1,10 @@
...
-  /**
-   * Set the height of the breakout element.
-   */
-  height?: string;
-
-  /**
-   * Horizontal alignment of the items inside the breakout element.
-   * @default left
-   */
-  alignX?: keyof typeof alignment.horizontal.alignmentX;
```
sarahgm, line 17: why is this useless? With this props you can position the Breakout elements or not?

sarahgm (changes requested):

packages/components/src/Breakout/Breakout.tsx
sebald: Because you can just use another layout component to do that. Do you have an actual use case where this is useful?

packages/components/src/Text/Text.tsx
sebald: yes

packages/components/src/Breakout/Breakout.tsx
sarahgm: Actually I don't know who ever used our breakout :D

sarahgm (dismissed):

PR #4331, opened 2024-12-06
feat([DSTSUP-98]): introduce controlled dismissable `<SectionMessage>`

sarahgm:
# Description

- add property `onDismiss` and `dismissable` to control the `<SectionMessage>` 
- add tests
- add story
- updated the docs in dismiss section 

# What should be tested?

every thing like it should be

# Reviewers:
@marigold-ui/developer


.changeset/gorgeous-books-warn.md
```diff
@@ -0,0 +1,8 @@
+---
+"@marigold/docs": patch
+"@marigold/components": patch
+---
+
+feat([DSTSUP-98]): introduce controlled dismissable `<SectionMessage>`
+
+Added possibility to  controll the dismissable state in `<SectionMessage>`, you can now use `onDismiss` and `dismissable` to control the closing of a section message.
```
aromko, line 8: typo control

docs/content/components/content/section-message/section-message.mdx
```diff
@@ -86,6 +87,8 @@ To provide this, you have to use the `closeButton` property on the `<SectionMess
 
 <ComponentDemo file="./section-message-dismissable.demo.tsx" />
 
+You can also controll the dismissable message by using the `dismissable` and `onDismiss` property.
```
aromko, line 90: typo

packages/components/src/SectionMessage/SectionMessage.tsx
```diff
@@ -74,6 +75,14 @@ export interface SectionMessageProps {
    * Adds a close button, makes the section message dismissable.
    */
   closeButton?: boolean;
+  /**
+   * Handler that is called when you need to controll the dismissable message.
```
aromko, line 79: typo

packages/components/src/SectionMessage/SectionMessage.tsx
```diff
@@ -83,6 +92,8 @@ export const SectionMessage = ({
   size,
   children,
   closeButton,
+  dismissable,
```
aromko, line 95: I think I'm not really happy with `dismissable` and `onDismiss`. I compared it to Dialog and there we use `dismissable` to avoid closing the dialog when clicking outside. Here it is another usage for this prop. To keep our props or API consistent and maybe also the user wouldn't expect this I would prefer to rename them. So my proposal is `dismiss` = `open` and `onDismiss` = `onOpenChange`. What do you think? @marigold-ui/developer

packages/components/src/SectionMessage/SectionMessage.tsx
```diff
@@ -92,13 +103,18 @@ export const SectionMessage = ({
   });
   const Icon = icons[variant];
 
-  const [isVisible, setIsVisible] = useState(true);
+  const isUncontrolled = dismissable === undefined;
+  const [internalVisible, setInternalVisible] = useState(true);
+  const isCurrentlyVisible = isUncontrolled ? internalVisible : dismissable;
```
aromko, line 108:
```suggestion
  const [internalVisible, setInternalVisible] = useState(true);
  const isCurrentlyVisible = dismissable ?? internalVisible;
```
Can a liitle bit simplified.

packages/components/src/SectionMessage/SectionMessage.tsx
```diff
@@ -92,13 +103,18 @@ export const SectionMessage = ({
...
 
-  const handleClose = () => {
-    setIsVisible(false);
+  const handleDismissable = () => {
+    onDismiss?.();
+    if (isUncontrolled) {
+      setInternalVisible(false);
+    }
```
aromko, line 114:
```suggestion
 const handleDismissable = () => {
    onDismiss?.();
    if (dismissable === undefined) {
      setInternalVisible(false);
    }
```

aromko (changes requested):

packages/components/src/SectionMessage/SectionMessage.tsx
sarahgm: I think you can not open and close a SectionMessage, this was also why i was not sure how it is meant in the SUP

packages/components/src/SectionMessage/SectionMessage.tsx
sarahgm: thank you

packages/components/src/SectionMessage/SectionMessage.tsx
aromko: With `open` I agree. But than maybe `close`. At least the user can click a close button and it affects the state.

That's whats ChatGPT says:

<strong>Key Differences</strong></h3>
Feature | dismiss | open / close
-- | -- | --
Type | Action (callback function) | State or event prop (boolean/handler)
Responsibility | Notifies the parent to dismiss | Controls visibility directly
Parent's Role | Handles dismissal logic | Manages component state
Child's Role | Calls the callback to dismiss | Reacts to state changes
Flexibility | More decoupled and reusable | Clearer control of visibility

sarahgm: Hm I at first had it like this, if its ok for @sebald  i can change it back to `close` and `onClose` https://reservix.atlassian.net/browse/DSTSUP-98?focusedCommentId=453484

packages/components/src/SectionMessage/SectionMessage.test.tsx
```diff
@@ -159,3 +160,31 @@ test('allow to close message with button in message', async () => {
...
+
+test('support controlled dismiss message', async () => {
+  const Controlled = () => {
+    const [close, setClose] = useState(false);
+    return (
+      <ThemeProvider theme={theme}>
+        <SectionMessage
+          data-testid="messages"
+          closeButton
+          dismissable={!close}
```
aromko, line 172: At least it should be consistent 😅

aromko (dismissed):
I won't block it anymore. I think I gave my thoughts into it and it's fine then.

packages/components/src/SectionMessage/SectionMessage.test.tsx
sarahgm: haha :D true

packages/components/src/SectionMessage/SectionMessage.tsx
sebald: yeah, that sounds fine with me. Maybe a personal preferrence, but you open/close a dialog, but you dismiss a message :)

aromko (dismissed):

packages/components/src/SectionMessage/SectionMessage.tsx
```diff
@@ -116,10 +131,10 @@ export const SectionMessage = ({
           {Icon && <Icon />}
         </div>
         {closeButton && (
-          <button
+          <Button
```
sebald, line 134: `useButton` hook might be more appropiate here, would not require to override all button styles then.

packages/components/src/SectionMessage/SectionMessage.tsx
```diff
@@ -74,6 +75,14 @@ export interface SectionMessageProps {
    * Adds a close button, makes the section message dismissable.
    */
   closeButton?: boolean;
+  /**
+   * Handler that is called when you need to control the dismissable message to close.
+   */
+  onClose?: () => void;
```
sebald, line 81: should we do a `onCloseChange` here so you can have both (open/close) like in the dialog?

sarahgm: you mean rename it?

sarahgm: open would not make sense to me, how can you open a section message?

sebald (approved):

PR #4336, opened 2024-12-10
bugfix([DST-627]): replace useState import from storybook

aromko:
# Reviewers:
@marigold-ui/developer

### DIFF
--- a/.changeset/new-paws-carry.md
+++ b/.changeset/new-paws-carry.md
@@ -1,0 +1,16 @@
+---
+"@marigold/components": patch
+---
+
+bugfix([DST-627]): replace useState import from storybook
+
+Some controlled stories in Storybook were incorrect because the useState import from react was used, resulting in an error.
+
+The following components were affected:
+
+- Calendar
+- DateField
+- DatePicker
+- Dialog
+- HelpText
+- SearchField
--- a/.changeset/new-paws-carry.md
+++ b/.changeset/new-paws-carry.md
@@ -6,7 +6,7 @@ bugfix([DST-627]): replace useState import from storybook
 
 Some controlled stories in Storybook were incorrect because the useState import from react was used, resulting in an error.
 
-The following components were affected:
+The stories of the following component were affected:
 
 - Calendar
 - DateField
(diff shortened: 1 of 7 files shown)


.changeset/new-paws-carry.md
```diff
@@ -0,0 +1,16 @@
+---
+"@marigold/components": patch
+---
+
+bugfix([DST-627]): replace useState import from storybook
+
+Some controlled stories in Storybook were incorrect because the useState import from react was used, resulting in an error.
+
+The following components were affected:
```
aromko, line 9:
```suggestion
The stories of the following component were affected:
```

sebald:
@aromko I think we had this issues a few times. Is it maybe also related to this: https://stackoverflow.com/questions/71993857/using-usestate-along-with-global-decorators-throws-error-storybook-preview-hooks/77859321#77859321

aromko:
> @aromko I think we had this issues a few times. Is it maybe also related to this: https://stackoverflow.com/questions/71993857/using-usestate-along-with-global-decorators-throws-error-storybook-preview-hooks/77859321#77859321

Yeah. And I looked now through all stories. So it should be fixed now.

sebald:
@aromko but using their preview API seems not the correct way to solve this. Even in their official docs, they don't use this API:

https://storybook.js.org/docs/8.5/writing-stories#working-with-react-hooks

aromko:
> @aromko I think we had this issues a few times. Is it maybe also related to this: https://stackoverflow.com/questions/71993857/using-usestate-along-with-global-decorators-throws-error-storybook-preview-hooks/77859321#77859321

To put it to the end doesn't work. And i go though some issues on their github repo. It seems to came up multiple times (latest https://github.com/storybookjs/storybook/issues/29189). Another solution was to use Story() instead of <Story /> and we do this already. So @sebald have another solution in mind or let it for now?

sebald:
@aromko so changing the order didn't fix it?

aromko:
> @aromko so changing the order didn't fix it?

If I haven't done something wrong - it doesn't helped

sebald (approved):

PR #4338, opened 2024-12-13
docs([DST-624]): Revise `<Calendar>` page and change some properties

sarahgm:
# Description

Revised the page according to our new guidelines and change `isDateUnavailable` to `dateUnavailable`. Remove the props:  `visibleDuration` and `pageBehavior`. Added outline focus styles for keyboard navigation

# Reviewers:
@marigold-ui/developer
@marigold-ui/designer


sebald:
@sarahgm should mark this as `major` because it has breacking changes, right? 👀

sebald:
@sarahgm still a BR, would rather stay consistent with versioning. Plus, we have outstanding  breaking change anyway

sarahgm:
> @sarahgm still a BR, would rather stay consistent with versioning. Plus, we have outstanding breaking change anyway

but for what reason? You also said to me that the renaming of `isPending` to `loading` in button would not be as a breaking change because nobody is using it anyway ;) so why this is than? @sebald

sebald (approved):

PR #4374, opened 2025-01-09
fix([DST-638]): update package.jsons to resolve warning

sarahgm:
# Description

while building the themes, there appears always: `The condition "types" here will never be used as it comes after both "import" and "require" [package.js]`

Fixed through removing the types entry in `exports` should be covered through: 
`"types": "./dist/index.d.ts",`

# Reviewers:
@marigold-ui/developer


sebald:
Did you check that the nested imports stil work with typescript? (see https://www.kravchyk.com/typescript-npm-package-json-exports/)

sebald (approved):

PR #4395, opened 2025-01-21
fix(DST-644): Make `<XLoader>` use underlay styles from theme in docs

sebald:


.changeset/quick-jars-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@marigold/components": patch
+---
+
+fix(DST-644): Make `<XLoader>` use underlay styles from theme in docs
```
aromko, line 5: Isn't it more about using now the <Underlay> and it's styles. And styles are dependent from the theme I use or?

sebald: Can't quite follow 🙈

aromko (dismissed):

aromko:
@sebald argh conflicts

aromko (approved):

PR #4412, opened 2025-02-06
feat[DST-641]: migrate core tailwind normalization to marigold repo with tailwind v4

aromko:
# Description

I've added a css file and copy the normalization from the core and add a command to copy the file into the dist.

Core needs just to import this file than 🤞 @import './../node_modules/@marigold/theme-core/dist/normalization.css';

# Reviewers:

@marigold-ui/developer


aromko:
Moved normalization to the system itself.

PR #4415, opened 2025-02-07
DST-641: add own normalization file v3

aromko:
# Description

- [] This change requires a UI-Kit update - inform Alex Tirado (@tirado-rx)

# What should be tested?

# Reviewers:

Examples:
@marigold-ui/developer
@marigold-ui/designer


sebald:
done directly in the ticketting system

PR #4426, opened 2025-02-13
style([DST-642]): Setup new blank theme for RUI 

OsamaAbdellateef:
# Description

Setup new blank theme for RUI in Marigold Repo

@marigold-ui/developer


sarahgm:
can you also add here the opportunity to change theme in storybook?

themes/theme-rui/src/root.ts
```diff
@@ -0,0 +1,3 @@
+import { Theme, cva } from '@marigold/system';
+
+export const root: Theme['root'] = cva('');
```
aromko, line 3: In the core repo is css/global.css. I think we can use these styles as root styles here or remove the root thing here and it in the styles.css?

OsamaAbdellateef:
> can you also add here the opportunity to change theme for rui in storybook?

Done :heavy_check_mark:  @sarahgm

aromko (approved):

PR #4443, opened 2025-02-17
feat[DST-676]: add button styles to rui theme

aromko:
# Description

Button styles added in 4 variants according to OriginUI:

- primary
- secondary (default when no variant specified)
- destructive
- ghost

Also added states:

- disabled
- hover
- pending (for loading)
- focus-visible (with outline when tabbed into)

- [] This change requires a UI-Kit update - inform Alex Tirado (@tirado-rx , @benjirsvx )

# What should be tested?

Look into Storybook "Button".

The basic show all variants add once. Toggle disable to true.

Also look add loading story.

In general color should be checked.

For hover I used our accent color in secondary and ghost variant like OriginUI - I think this should be changed but idk which color to use.

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer

### DIFF
--- a/themes/theme-rui/src/components/Button.styles.ts
+++ b/themes/theme-rui/src/components/Button.styles.ts
@@ -1,18 +1,29 @@
-import { cva } from '@marigold/system';
 import type { ThemeComponent } from '@marigold/system';
+import { cva } from '@marigold/system';
 
 export const Button: ThemeComponent<'Button'> = cva(
   [
-    'inline-flex items-center justify-center whitespace-nowrap rounded-lg text-sm font-medium transition-colors outline-offset-2 focus-visible:outline-2 focus-visible:outline-ring/70 disabled:pointer-events-none disabled:opacity-50',
+    'inline-flex items-center justify-center whitespace-nowrap rounded-lg text-sm font-medium transition-colors outline-hidden',
+    '[&_svg]:pointer-events-none [&_svg]:shrink-0',
+    'focus-visible:outline-offset-2 focus-visible:outline focus-visible:outline-ring/70',
+    'h-button px-4 py-2',
+    'disabled:pointer-events-none disabled:bg-disabled disabled:text-disabled-foreground disabled:border-none disabled:cursor-not-allowed',
+    'pending:text-disabled-foreground pending:bg-disabled pending:cursor-not-allowed pending:border-none',
   ],
   {
     variants: {
       variant: {
         primary:
-          'bg-brand text-brand-foreground shadow-sm shadow-black/5 hover:bg-primary/90',
-
-        destructive: 'bg-destructive text-destructive-foreground',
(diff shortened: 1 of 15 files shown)


themes/theme-rui/src/components/Button.styles.ts
```diff
@@ -1,18 +1,29 @@
...
         primary:
-          'bg-brand text-brand-foreground shadow-sm shadow-black/5 hover:bg-primary/90',
-
-        destructive: 'bg-destructive text-destructive-foreground',
+          'bg-brand text-brand-foreground shadow-sm shadow-black/5 hover:bg-brand/90',
+        secondary:
+          'border border-border bg-background shadow-sm shadow-black/5 hover:bg-accent hover:text-foreground',
+        destructive:
+          'bg-destructive text-destructive-foreground hover:bg-destructive/90',
+        ghost: 'hover:bg-accent hover:text-foreground',
```
sebald, line 22: think we wanted to call this "tertiary". But ghost is much easier to write :D we can leave it like that and ask/communicate this via our design doc (see Confluence)

themes/theme-rui/src/components/Button.styles.ts
```diff
@@ -1,18 +1,29 @@
-import { cva } from '@marigold/system';
 import type { ThemeComponent } from '@marigold/system';
+import { cva } from '@marigold/system';
 
 export const Button: ThemeComponent<'Button'> = cva(
```
sebald, line 4: Why did you use accent for the hovers? 🤔

sebald (approved):

themes/theme-rui/src/theme.css
```diff
@@ -43,6 +43,10 @@
   --color-brand: var(--color-stone-950);
   --color-brand-foreground: var(--color-stone-50);
 
+  /* stone (use for secondary) */
```
sarahgm, line 46: what is secondary? Only for buttons?

themes/theme-rui/src/components/Button.styles.ts
aromko: I just took it from OriginUI. They called it the same way but idk which color it should have. That was one of my questions in the description above.

themes/theme-rui/src/theme.css
aromko: For now yes or how we call other colors. I'm really confused about our naming now anyways. So I'm open for other names.

themes/theme-rui/src/components/Button.styles.ts
aromko: I added to the docs.

themes/theme-rui/src/theme.css
sarahgm: I think its fine for now just curios if they use it also for other components

PR #4448, opened 2025-02-20
fix([DST-695]):refactor pull request template

sarahgm:
# Description

Add benji as informer, remove duplicated template

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


sebald:
What about using @marigold-ui/designer instead?

aromko (approved):

PR #4457, opened 2025-02-24
fix[DST-699]: fix console error

aromko:
# Description

See https://reservix.atlassian.net/browse/DST-699

# Reviewers:

@marigold-ui/developer

### DIFF
--- a/packages/components/src/Dialog/Dialog.test.tsx
+++ b/packages/components/src/Dialog/Dialog.test.tsx
@@ -130,7 +130,7 @@ test('optionally renders a close button', () => {
   expect(dialog).not.toBeVisible();
 });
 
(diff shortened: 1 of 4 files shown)


packages/components/src/Dialog/Dialog.test.tsx
```diff
@@ -277,28 +277,6 @@ test('child function is passed an id for the dialog title (a11y)', () => {
   );
 });
 
-// test('warns if no element to attach the title can be found', () => {
```
sarahgm, line 280: why is this commented?

aromko: idk. i just deleted it because it was commented.

sarahgm (approved):

PR #4463, opened 2025-03-03
feat(overlay): allow passing HTMLElement

imdhemy:
# Description  

## Context:
We use a custom HTML element with a shadow root to render our application. This encapsulates our element from the document's DOM, preventing it from accessing main document elements. 

Currently, the `OverlayContainerProvider` expects an element ID and retrieves it using `document.getElementById(portalContainer)`. However, this approach fails in our use case, as the query always returns `null` due to shadow DOM encapsulation.  Additionally, there is no `body` in our case, which further limits our ability to append overlays to the standard document structure.  

Screenshot 2025-03-03 at 11 06 22 PM

## This PR:
- Adds support for passing an `HTMLElement` directly to the `OverlayContainerProvider`.  
- Preserves the existing behavior of accepting an element ID.  
- Enhances and fixes unit tests for `OverlayContainerProvider`.

# What should be tested?

- Render portal container by element ID.
- Fallback the portal container to the `body`.
- Render portal container by HTMLElement.

# Reviewers:

Examples:
@marigold-ui/developer
@marigold-ui/designer

### DIFF
--- a/packages/components/src/Provider/OverlayContainerProvider.tsx
+++ b/packages/components/src/Provider/OverlayContainerProvider.tsx
@@ -1,21 +1,24 @@
 import { createContext, useContext } from 'react';
 import { useIsSSR } from '@react-aria/ssr';
 
-// needs an id for the portal container, it will be overgiven as string,
-// if there is no the body is used
-const OverlayContainerContext = createContext<string | undefined>(undefined);
+/**
+ * Context to provide the container element for overlay components.
+ * This is used by overlay components to append themselves to the correct container.
+ * If no container is provided, it falls back to `document.body`.
+ */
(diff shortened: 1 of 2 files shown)


packages/components/src/Provider/OverlayContainerProvider.tsx
```diff
@@ -1,21 +1,24 @@
...
 export const usePortalContainer = () => {
   const portalContainer = useContext(OverlayContainerContext);
-  const isSSR = useIsSSR();
 
-  const portal = isSSR
-    ? null
-    : portalContainer
-      ? document.getElementById(portalContainer)
-      : document.body;
+  if (useIsSSR()) return null;
```
imdhemy, line 18: I kept this logic, but why isn't this available for SSR?

PR #4469, opened 2025-03-13
feat(rui): Next batch of changes!

sebald:


themes/theme-rui/src/theme.css
```diff
@@ -44,8 +45,8 @@
   --color-brand-foreground: var(--color-stone-50);
 
   /* used for hovering (e.g. calendar, button) */
-  --color-hover: var(--color-stone-300);
-  --color-hover-foreground: var(--color-stone-950);
+  --color-hover: var(--color-stone-100);
+  --color-hover-foreground: var(--color-stone-900);
 
   /* stone (use for secondary) */
   --color-secondary: var(--color-stone-50);
```
sebald, line 52: Why do we have this? And what is secondary? 👀

sebald: CC @marigold-ui/developer

sarahgm: hover for example buttons -> need a foreground also, secondary also used in button i think @sebald

sarahgm (approved):

PR #4500, opened 2025-04-01
refa(DST-747): Enhance Overlay Components

OsamaAbdellateef:
# Description

The current overlay components (Modal, Dialog) were designed with modularity in mind, using separate components like Underlay and explicit prop drilling for state management. While this approach provided flexibility, evolving requirements have highlighted opportunities to streamline the implementation for better maintainability and scalability. For example, integrating accessibility features like lang/dir attributes and ensuring consistent state propagation across nested components (e.g., DialogTrigger, Modal, Dialog) now requires additional coordination.

# What should be tested?

## Dialog should have the same behaviour and styles 

@marigold-ui/developer


packages/components/package.json
```diff
@@ -37,6 +37,7 @@
     "@react-aria/table": "3.17.1",
     "@react-aria/utils": "3.28.1",
     "@react-aria/visually-hidden": "3.8.21",
+    "@react-spectrum/utils": "3.12.3",
```
sebald, line 40: Do we need this? 👀

packages/components/src/Dialogv2/Modal.tsx
```diff
@@ -0,0 +1,57 @@
...
+  size?: string;
+}
+
+export const _Modal = forwardRef(
+  (props: ModalProps, ref: DOMRef<HTMLDivElement>) => {
+    let domRef = useDOMRef(ref);
+    let { locale, direction } = useLocale();
+
+    // TODO: should we pass through lang and dir props in RAC?
+    let modalRef = useCallback(
```
sebald, line 22: what is the use case to set the lang guage on this?

packages/components/src/Dialogv2/Dialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+  size?: string;
+  /**
+   * Show the close button.
+   */
+  closeButton?: boolean;
+  /**
+   * If `true`, the dialog will be non-modal, meaning it will not block interaction with the background content.
+   * @default false
+   */
+  isNonModal?: boolean;
```
sebald, line 31: I don't think this is used anywhere in the component itself?

packages/components/src/Dialogv2/Modal.tsx
OsamaAbdellateef: it is used for other langs like Arabic so it is rtl, but i think we don't need it ! @sebald

packages/components/src/Dialogv2/DialogTrigger.tsx
```diff
@@ -0,0 +1,29 @@
...
+  extends RAC.DialogTriggerProps,
+    Omit<ModalProps, 'children'> {}
+
+export const DialogContext = createContext<ModalProps>({});
+
+const _DialogTrigger = (props: DialogTriggerProps): ReactNode => {
+  return (
+    <DialogContext.Provider value={props}>
+      <DialogTrigger {...props}>
+        <PressResponder isPressed={false}>{props.children}</PressResponder>
```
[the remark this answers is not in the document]
OsamaAbdellateef: It's used to set isPressed , normally when we press the button and dialog is open the button is still visually pressed
so PressResponder handles this  @sebald

packages/components/package.json
OsamaAbdellateef: Yeah, i don't think so , deleted the useDomRef @sebald

packages/components/src/Dialogv2/Dialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+ * Depending on the kind of Dialog, further interactions may be blocked until the Dialog is acknowledged.
+ */
+const _Dialog = forwardRef((props: DialogProps, ref: DOMRef) => {
+  let domRef = useDOMRef(ref);
+  const classNames = useClassNames({
+    component: 'Dialog',
+    variant: props.variant,
+    size: props.size,
+  });
+  const { isDismissable, isKeyboardDismissDisabled, isOpen } =
```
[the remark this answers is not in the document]
OsamaAbdellateef: You mean for variant and sizes ? @sebald

OsamaAbdellateef:
> @OsamaAbdellateef the changes are still under "Dialog2" :D

:heavy_check_mark:

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -35,56 +66,49 @@ const CloseButton = ({ className }: CloseButtonProps) => {
...
-  const state = useContext(OverlayTriggerStateContext);
+    const state = useContext(OverlayTriggerStateContext);
 
-  const children =
-    typeof props.children === 'function'
-      ? props.children({
-          close: state?.close ?? (() => {}),
-        })
-      : props.children;
+    const children =
```
sebald, line 81: Do we need this part? can't we use the `children` from RAC?

sebald (approved):

PR #4518, opened 2025-04-08
docs([DST-722]): Add RUI to the docs

sebald:


.changeset/late-pumpkins-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+"@marigold/docs": minor
+---
+
+docs([DST-722]): Add RUI to the docs
```
sarahgm, line 5: Would call it: Allow to switch to RUI theme in docs. Add RUI to the docs sound more that its officially inside the docs also with updated docs

sarahgm:
Can you also add it to the command menu? @sebald

.changeset/late-pumpkins-swim.md
aromko: Should we put rui as default already?

sebald: I would do it when we have adjusted everything else and after other themes are deprecated

sebald:
> Can you also add it to the command menu? @sebald

@sarahgm I thought it automatically adds it because of the theme switch. Where do I have to add it?

aromko (approved):

PR #4535, opened 2025-04-14
infra([DST-771]): Add safelist to theme-docs

sarahgm:
# Description

# What should be tested?

test if the docs still look the same as before

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer


themes/theme-docs/src/styles.css
```diff
@@ -1,10 +1,10 @@
 @import 'tailwindcss';
-@import './theme.css' theme(inline);
+@import './theme.css';
 
 /* from tailwind.config.ts */
 @source './**/*.*.ts';
-@source './safelist.txt';
-
+@source inline("{bg,text,border,shadow}-{amber,blue,lime,neutral,red,secondary,primary,code}-{50,{100..900..100},950}");
+@source inline("{bg,text,border,shadow}-{text-primary,text-primary-muted,text-disabled,text-info,text-warning,bg-body,bg-inverted,bg-hover,bg-muted,bg-disabled,bg-underlay,bg-surface,bg-surface-raised,bg-surface-lowered,bg-surface-overlay,bg-success,bg-warning,bg-error,bg-info,bg-border,bg-border-primary,bg-border-success,bg-border-error}");
```
aromko, line 7: Interpreted from the comments in safelist.txt I would say bg-primary-50/25, bg-secondary-400/10, text-text-primary/60 is missing.

themes/theme-core/src/styles.css
```diff
@@ -1,5 +1,5 @@
 @import 'tailwindcss';
-@import './theme.css' theme(inline);
```
aromko, line 2: We don't need this because this is default?

sarahgm: what do we not need? `@import './theme.css';`?

themes/theme-docs/src/styles.css
sarahgm: but should it not be already working? Its only used for the reference app right? Can the reference app not create this?

aromko: No these 3 are used in docs. All other props has a comment when used in reference app. I think this works because we just copied the safelist. E.g. Rectangles looks different now.

themes/theme-core/src/styles.css
sarahgm: Bildschirmfoto 2025-04-22 um 14 12 40
this one this not working when having `theme(inline)` around it thats why i removed it. Or why was it needed? 🙈

sebald: the `theme(inline)` only inlines values instead of using css vars. so the outcome should be the same. only difference is in the css that there might be css vars missing because they are inlinded. 

Removing the inline from the theme *should* not have any impact on the docs.

aromko (approved):

PR #4540, opened 2025-04-15
ref(DST-760): Extend useClassNames to Automatically Pick Up Styles from Context

OsamaAbdellateef:
# Description

Extend useClassNames to Automatically Pick Up Styles from Context e.g `size` & `variant`

# Reviewers:

@marigold-ui/developer


.changeset/solid-walls-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@marigold/system': patch
+---
+
+Extend useClassNames to Automatically Pick Up Styles from Context
```
aromko, line 5: lower case

Extend useClassNames to automatically pick up styles from context.

packages/system/src/hooks/useClassNames.tsx
```diff
@@ -18,14 +26,22 @@ export type ComponentClassNames<C extends ComponentNames> =
...
   className,
   variant,
   size,
+  context: ComponentContext,
 }: UseClassNamesProps<C>): ComponentClassNames<C> => {
   const theme = useTheme();
 
+  const ctx = useContext(ComponentContext ?? FallbackContext);
+
+  const currentSize = size ?? ctx.size;
```
aromko, line 42: Are these lines covered by tests? If not we should add new tests for this.

packages/system/src/hooks/useClassNames.tsx
```diff
@@ -18,14 +26,22 @@ export type ComponentClassNames<C extends ComponentNames> =
         [slot in keyof ThemeComponent<C>]: string;
       };
 
+const FallbackContext = createContext({}) as Context<ComponentContextProps>;
```
sebald, line 29: should this be `null`? as this is the default of not having any context.

OsamaAbdellateef: didn't get it 
use null instead of {} ? @sebald

aromko (dismissed):

packages/system/src/hooks/useClassNames.tsx
OsamaAbdellateef: i  get run time error because of   const currentSize = size ?? ctx.size;
if no size then it will go to ctx.size (null.size) @sebald

sebald:
Not sure if this should be done in this PR, but can we also use it? Would be a good way to verify the API.

packages/system/src/hooks/useClassNames.tsx
sebald: yes but you should be able to catch this by doing `ctx?.size`

OsamaAbdellateef:
> Not sure if this should be done in this PR, but can we also use it? Would be a good way to verify the API.

Use what ? :thinking: @sebald

aromko:
> > Not sure if this should be done in this PR, but can we also use it? Would be a good way to verify the API.
> 
> Use what ? 🤔 @sebald

I also don't get it 🤷‍♂️

aromko (approved):

sebald:
@OsamaAbdellateef @aromko use the new `useClassName` API in a component so that we now it is working like we want it to. E.g. the `Table`, `Dialog` or `Drawer` :P

OsamaAbdellateef:
> @OsamaAbdellateef @aromko use the new `useClassName` API in a component so that we now it is working like we want it to. E.g. the `Table`, `Dialog` or `Drawer` :P
done :heavy_check_mark:  @sebald

aromko (approved):

PR #4541, opened 2025-04-16
refa([DST-720]): Rename `option` to `item` in styles for `<ListBox>`

sarahgm:
# Description

Rename `option` in styles file for `<ListBox>` in `item`

# Reviewers:

@marigold-ui/developer

### DIFF
--- a/.changeset/afraid-steaks-hang.md
+++ b/.changeset/afraid-steaks-hang.md
@@ -1,0 +1,12 @@
+---
+"@marigold/components": patch
+"@marigold/system": patch
+"@marigold/theme-b2b": patch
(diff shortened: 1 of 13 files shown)


.changeset/afraid-steaks-hang.md
```diff
@@ -0,0 +1,12 @@
+---
+"@marigold/components": patch
+"@marigold/system": patch
```
sebald, line 3: Patch or BR? 😜 I think at least the themes should be a BR. For components, we only use the listbox internally so it is fine I guess

sebald (approved):

PR #4543, opened 2025-04-17
[DST-790]: migrate to vitest

aromko:
# Description

- setup vitest
- migrated nearly all tests to vitest
- migrated build job to use vitest
- added coverage report generated by vitest

Open Todos:

- [x] remove jest from codebase
- [x] fix broken tests labeled with `.skip` (DatePicker.test.tsx, SelectList.test.tsx)
- [x] fix extendTheme.test.tsx - remove exclude from vitest.config

# What should be tested?

Run all tests locally.

```
pnpm test

pnpm test:coverage
```

# Reviewers:

@marigold-ui/developer


package.json
```diff
@@ -4,17 +4,18 @@
   "repository": "github:marigold-ui/marigold",
   "private": true,
   "license": "MIT",
+  "type": "module",
   "engines": {
     "node": "20.x"
   },
   "packageManager": "pnpm@10.8.0",
   "dependencies": {
+    "-": "0.0.1",
```
sebald, line 13: What is this? :D

package.json
```diff
@@ -72,6 +73,11 @@
   },
   "devDependencies": {
```
sebald, line 74: can we move the whole deps into the `dependencies` prop? :D

package.json
aromko: idk. I think I remove this

package.json
aromko: We could but why? We don't need them in production or?

sebald (approved):

PR #4544, opened 2025-04-17
refa(DST-736): Convert class mixins to Tailwind utilities

sebald:
Mixins -> Tailwind Utilities


themes/theme-rui/src/utils.css
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/*
+ * Applies a consistent visual style for keyboard focus states across components.
+ * This utility sets a colored focus ring based on the given theme color,
+ * while removing the default outline.
+ *
+ * Use this when the component has its own border and you need to keep it.
+ */
+@utility util-focus-borderless-* {
```
aromko, line 35: When I red this name before reading the comment I thought it would remove the border. I would prefer names which says what they are doing. So maybe this should be util-focus and above util-focus-border?

sebald: hm, but then people might think `util-focus` is the default, which it isn't. That is why I chose this naming. "A variant of the default focus, but without a border" :D

aromko: Leave it but for me it's not clear. Now I know it but I will forget this when I read this next time and wonder why until I have a look on this utility 😅

sebald: maybe we find a better naming next time :D

PR #4549, opened 2025-04-22
refa([DST-754]): Add `<FocusScope>` to `<Pagination>`

sarahgm:
# Description

- added Focusscope to pagination
- fix keyboard navigation when pressing enter on keyboard

# What should be tested?

# Reviewers:
@marigold-ui/developer


packages/components/src/Pagination/PageButton.tsx
```diff
@@ -8,6 +8,7 @@ interface PageButtonProps {
   isDisabled?: boolean;
   onClick?: () => void;
   registerRef?: (ref: HTMLButtonElement | null) => void;
+  onKeyDown?: (event: React.KeyboardEvent) => void;
```
aromko, line 11: React already imported at the top.

packages/components/src/Pagination/Pagination.stories.tsx
```diff
@@ -35,25 +36,81 @@ const meta = {
...
+    await step('Click on the previous button', async () => {
+      const prevButton = canvas.getByLabelText('Page previous');
+      await userEvent.click(prevButton);
+      expect(canvas.getByLabelText('Page 2')).toHaveAttribute(
+        'data-selected',
+        'true'
+      );
+    });
+
+    await step('use arrow right navigation', async () => {
```
aromko, line 87: I think this is copied from what I initially did and I don't like the structure. I would prefer the pattern Arrange, act, assign (s. https://thinkster.io/tutorials/blogs/the-aaa-structure-of-tests)

Means basically create the initially state, act on this means do some changes / interactions, assign or test is everything as expect(ed).

We can approach this putting a blank line between the different steps. I think we have no guidelines for this so we also can talk about in the mob.

packages/components/src/Pagination/Pagination.stories.tsx
```diff
@@ -35,25 +36,81 @@ const meta = {
...
+      pageButton.focus();
+      expect(pageButton).toHaveFocus();
+      const nextPageButton = pageButton.nextElementSibling as HTMLElement;
+
+      userEvent.keyboard(`ArrowRight`);
+
+      nextPageButton.focus();
+      expect(nextPageButton).toHaveFocus();
+      expect(nextPageButton).toHaveTextContent('3');
+    });
```
aromko, line 101: blank line between steps

packages/components/src/Pagination/Pagination.tsx
```diff
@@ -42,21 +39,30 @@ export interface PaginationProps {
   controlLabels?: [string, string];
 }
 
-const _Pagination = ({
-  defaultPage = 1,
-  page,
-  totalItems,
+interface InnerPaginationProps {
```
aromko, line 42: The props are very similiar, so you could use something like extend or union multiple types.

packages/components/src/DatePicker/DatePicker.test.tsx
```diff
@@ -456,5 +456,6 @@ test('DatePicker supports data unavailable property', async () => {
   expect(popover).toBeVisible();
   const date = screen.getAllByRole('gridcell');
 
-  expect(date[1].firstChild).toHaveAttribute('data-unavailable', 'true');
+  // select a date in the month that is not the first
+  expect(date[10].firstChild).toHaveAttribute('data-unavailable', 'true');
```
aromko, line 460: You get a date or a day? I think last or? Maybe it would be helpful to also check which value is in there?

packages/components/src/DatePicker/DatePicker.test.tsx
```diff
@@ -456,5 +456,6 @@ test('DatePicker supports data unavailable property', async () => {
   expect(popover).toBeVisible();
   const date = screen.getAllByRole('gridcell');
 
-  expect(date[1].firstChild).toHaveAttribute('data-unavailable', 'true');
+  // select a date in the month that is not the first
```
aromko, line 459: Is this comment necessary?

.changeset/smart-buckets-double.md
```diff
@@ -0,0 +1,5 @@
+---
+"@marigold/components": patch
+---
+
+refa([DST-754]): Add `<FocusScope>` to `<Pagination>`
```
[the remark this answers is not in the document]
aromko: Or do we need a changelog here anyways? It doesn't change anything for the users or?

packages/components/src/DatePicker/DatePicker.test.tsx
sarahgm: if its clear i can remove it

aromko: I couldn't read out from this comment that you didi it because of a broken test.

aromko (approved):

PR #4559, opened 2025-04-24
refa([DST-762]): Unify Label API Across Form Fields

sarahgm:
# Description

Added label prop to Switch and Checkbox,
deprecated the children prop as in `Slider` already exists

# What should be tested?

everything working as expected

# Reviewers:
@marigold-ui/developer
@marigold-ui/designer

### DIFF
--- a/.changeset/tiny-feet-enjoy.md
+++ b/.changeset/tiny-feet-enjoy.md
@@ -1,0 +1,6 @@
+---
+"@marigold/docs": patch
+"@marigold/components": patch
+---
+
+refa([DST-762]): support `label` prop for `<CheckBox>` and `<Switch>`. Deprecate `children` property, use `label` instead. With the next major version `children` will be removed.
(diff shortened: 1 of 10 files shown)


packages/components/src/Switch/Switch.tsx
```diff
@@ -20,7 +20,15 @@ type RemovedProps =
 export interface SwitchProps extends Omit<RAC.SwitchProps, RemovedProps> {
   variant?: string;
   size?: string;
+  /**
+   * The children of the component.
+   * @deprecated Will be removed in the next major version. Use `label` prop instead.
+   */
   children?: ReactNode;
+  /**
+   * The label of the component.
```
aromko, line 29: I looked through other components how we describe the label prop. Was mainly this style: Set the label of the "component name".

.changeset/tiny-feet-enjoy.md
```diff
@@ -0,0 +1,6 @@
+---
+"@marigold/docs": patch
+"@marigold/components": patch
```
sebald, line 3: Needs to be flagges as a breaking change!

sarahgm: like thsi?

aromko (approved):

PR #4560, opened 2025-04-24
refa(DST-738): Adjust `<ListBox>` and `<SelectList>` styles to new  standards

OsamaAbdellateef:
# Description

Revise the styling for the <ListBox> and <SelectList> components to align with Origin-UI. Instead of using a background color to indicate the selected state, these components will now display a checkmark. 

# Reviewers:

Examples:
@marigold-ui/developer
@marigold-ui/designer


packages/components/src/ListBox/ListBoxItem.tsx
```diff
@@ -1,14 +1,40 @@
...
+    <ListBoxItem
+      {...props}
+      className={classNames.item}
+      // textValue needed because ListBoxItem in this case has multiple children
+      textValue={props.textValue ?? (String(props.children) as string)}
+    >
+      {({ isSelected }) => {
+        return (
+          <>
+            {isSelected && <CheckMark />}
```
sebald, line 31: Show/hide via CSS instead? Maybe we can also use the context instead of the function to check if an item is selected?

sebald (approved):

PR #4568, opened 2025-04-28
[DST-763]: migrate to flat config

aromko:
# Description

Migrate the eslint config to the new flat config. 

1. Tried to put more general stuff into the index.js because this could used in other projects. If overrides were necessary I put them into the eslint.config in the root directory
2. There are a lot of file changes because our eslint wasn't working since v9.x. I think it makes no sense to ask me why I did a specific change. If you want to know why undo the change and run `pnpm lint`

# What should be tested?

`pnpm lint` still works locally

# Reviewers:

@marigold-ui/developer


docs/ui/ThemeSwitch.tsx
```diff
@@ -12,11 +12,14 @@ import {
 } from 'react';
 import { useRouter, useSearchParams } from 'next/navigation';
 
+// Context
```
sarahgm, line 15: this is doubled 🤔

docs/ui/PropsTable.tsx
```diff
@@ -43,48 +43,52 @@ export const PropsTable = ({ component }: PropsTableProps) => {
...
-              </code>
-              <div
-                dangerouslySetInnerHTML={{ __html: prop.type.value }}
-                className="*:m-0 *:bg-transparent! *:p-0 *:text-xs"
-              />
-            </Inline>
+      {
+        // eslint-disable-next-line react/prop-types
+        props.length ? (
+          // eslint-disable-next-line react/prop-types
```
sarahgm, line 49: same here better for whole file disable

packages/components/src/Checkbox/CheckboxGroup.test.tsx
```diff
@@ -98,8 +98,11 @@ test('passes down "disabled" to checkboxes', () => {
     </Checkbox.Group>
   );
 
+  // eslint-disable-next-line testing-library/no-node-access
```
sarahgm, line 101: same here I think disable it for the whole file is better, so you don't need to write it that oftern

docs/app/_components/landingpage/SupportGrid.tsx
```diff
@@ -40,6 +42,7 @@ const ContributeIcon = ({ className = 'size-7' }) => (
   </svg>
 );
 
+// eslint-disable-next-line react/prop-types
```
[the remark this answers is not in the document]
aromko: In general I wouldn't recommend to to this for the whole file. If we have linting errors it is normally sign that something isn't good. Maybe it's depending on the error but if we disabled the whole file then we run the risk of missing real bugs.

I did on the specific line so we can and should fix it in the future. In some places I just don't know the solution without putting a lot of effort into it. But in general we shouldn't use any comments IMHO.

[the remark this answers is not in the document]
aromko: Btw I found the fix for this.

docs/ui/PropsTable.tsx
aromko: Like written in the other comment. I wouldn't recommend doing it for the whole file.

packages/components/src/Checkbox/CheckboxGroup.test.tsx
aromko: Like written in the other comment. I wouldn't recommend doing it for the whole file. 

I think this should also be fixable. We just need to use something like getByText or getByTestId. So for me it's more clear when I refa these tests I know where I should have a closer look.

docs/app/_components/landingpage/SupportGrid.tsx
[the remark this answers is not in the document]
sarahgm: Ah nice! Yes I think it depends on the error

sarahgm (approved):

PR #4584, opened 2025-05-05
feat([DST-804]): Allow `Tag` to be used in forms and overhaul docs

sebald:
# Description

- added `name` attribute to `Tag` (and hidden controls)
- added component tests (storybook)
- refactored tests of TagGroup to use setup from stories
- overhauled the documentation to be more thorough

# Reviewers:

@marigold-ui/developer


packages/components/src/TagGroup/TagGroup.tsx
```diff
@@ -2,27 +2,34 @@ import type RAC from 'react-aria-components';
...
+   * The name of the field, used when submitting form data.
+   */
+  name?: string;
+
   /**
-   * Displays a remove button on each tag.
-   * @default false
+   * Provides content to display when there are no items in the tag list.
    */
-  allowsRemoving?: boolean;
```
sarahgm, line 25: isn't this than also a breaking change if you remove this prop?

packages/components/src/TagGroup/TagGroup.tsx
```diff
@@ -43,12 +51,15 @@ const _TagGroup = ({
...
         {children}
       </TagList>
+      {name ? <TagGroupHiddenInput name={name} /> : null}
     </FieldBase>
   );
 };
 
 export { _TagGroup as TagGroup };
+
+// TODO: Why is the helptext showing?
```
sarahgm, line 65: because its a form component, and they have a help text? Not sure what you mean here 🙈

aromko (approved):

packages/components/src/TagGroup/TagGroup.tsx
sebald: ah sorry, this was from the story. There is a bottom/top margin because of the help text. Will go an check.

packages/components/src/TagGroup/TagGroup.tsx
sebald: well... it actually did nothing :D not sure when. But there is no "allowsRemoving" on RACs `TagGroup` component

sarahgm: okay but maybe its used by someone?

sebald: okay, I found *one* usage of the prop in our apps (but it doesn't do anything since months...). Do you want me to tag this as a breaking change? I would categorize it as a bug 🙈

sarahgm (approved):

PR #4603, opened 2025-05-14
chore([DST-853]): Refa styles for `<Menu>` button

sarahgm:
# Description

Menu does not match Origin UI standards, that because inside our Menu component we use a variant menu which does not exists on RUI Button. 
My proposal is to take the menu button styles out of the button and add it to the menu

# What should be tested?

everything working as expected and styles are the same as origin ui 

# Reviewers:

@marigold-ui/developer
@marigold-ui/designer

### DIFF
--- a/packages/components/src/Menu/Menu.test.tsx
+++ b/packages/components/src/Menu/Menu.test.tsx
@@ -1,13 +1,11 @@
 import { composeStories } from '@storybook/react';
 import { fireEvent, render, screen } from '@testing-library/react';
-import userEvent from '@testing-library/user-event';
 import { SVGProps } from 'react';
 import { vi } from 'vitest';
 import * as stories from './Menu.stories';
 
 // Setup
 // ---------------
-const user = userEvent.setup();
 
 const { Basic, BasicActionMenu, MenuSection } = composeStories(stories);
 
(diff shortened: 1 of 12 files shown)


packages/components/src/Menu/Menu.test.tsx
```diff
@@ -74,244 +24,76 @@ const mockMatchMedia = (matches: string[]) =>
...
-  const button = screen.getByText('Choose');
+  render(<Basic data-testid="menu" variant="one" />);
+  const button = screen.getByText('Hogwarts Houses');
   fireEvent.click(button);
 
   const menu = screen.getByRole('menu');
-  const item = screen.getByText('Burger');
+  const item = screen.getByText('Gryffindor');
 
   expect(menu.className).toMatchInlineSnapshot(
```
sebald, line 49: not sure if we should keep these tests, they will always break if we change something in the theme (e.g. when we make RUI the default)

I wonder if we can delete these tests alltogether (meaning the "is the variant/size applied" tests)

@marigold-ui/developer

sarahgm: For me its ok to delete it, I can also do it in a seperate PR

sebald: yeah lets talk about it in mob tomorrow?

PR #4605, opened 2025-05-15
[DST-854]: Refactor form component tests

aromko:
# Description

Refactor some component tests.

- Autocomplete

# Reviewers:

@marigold-ui/developer


sarahgm:
@aromko there are some conflicts and an open vercel comment

aromko:
> @aromko there are some conflicts and an open vercel comment

@sarahgm You're right. But it is now more time consuming to fix because there are new tests in the autocomplete which are not migrated. I'm not sure if I want/can spent the time for this right now.

sebald:
But leaving it like this feels kind of bad too.

aromko:
> But leaving it like this feels kind of bad too.

Yeah I feel the same. That's why it is still open.

PR #4607, opened 2025-05-16
refa([DST-715]): Refactor code from Calendar Component

OsamaAbdellateef:
# Description

Fix resizing of listboxes for month and year when calendar is open .

# What should be tested?
Calendar looks visually correct . 

# Reviewers:

Examples:
@marigold-ui/developer
@marigold-ui/designer

### DIFF
--- a/packages/components/src/Calendar/YearListBox.tsx
+++ b/packages/components/src/Calendar/YearListBox.tsx
@@ -62,7 +62,7 @@ const YearListBox = ({ setSelectedDropdown }: YearDropdownProps) => {
           <li className="flex justify-center" key={index}>
             <div
               ref={isSelected ? activeButtonRef : (null as any)}
-              style={{ height: '100%', width: '100%' }}
+              className="flex size-full justify-center"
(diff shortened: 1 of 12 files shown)


packages/components/src/Calendar/YearListBox.tsx
```diff
@@ -54,25 +56,27 @@ const YearListBox = ({ setSelectedDropdown }: YearDropdownProps) => {
...
+              ref={isSelected ? activeButtonRef : (null as any)}
               style={{ height: '100%', width: '100%' }}
             >
               <Button
                 slot="previous"
-                disabled={state.isDisabled}
-                variant={isActive ? 'secondary' : 'text'}
-                size="small"
+                className={cn(
+                  classNames.dateSelectorOption,
```
sarahgm, line 70: I think a better name would be `calendarListboxButton` or something like this? I think of date if i read `date` but thats just a part of it? What do you think

sarahgm (approved):

PR #4618, opened 2025-05-22
feat[DST-856]: Add TimeField Component

pauleitenbichler:
# Description

# What should be tested?

# Reviewers:

@aromko 
@marigold-ui/developer 
@marigold-ui/designer


packages/components/src/TimeField/TimeField.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+      description="Pick a time"
+      error
+      errorMessage="Invalid time"
+    />
+  );
+  expect(screen.queryByText('Pick a time')).not.toBeInTheDocument();
+  expect(screen.getByText('Invalid time')).toBeInTheDocument();
+});
+
+test('renders TimeField with default value and reacts to focus', async () => {
```
aromko, line 76: You are testing a lot here which is basically fine. But I would recommend to split these up. We also decided to keep tests basically in this form https://thinkster.io/tutorials/blogs/the-aaa-structure-of-tests

packages/components/src/TimeField/TimeField.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+    `"text-lime-300 p-1 inline-flex w-[var(--labelWidth)]"`
+  );
+
+  expect(screen.getByText('Description')).toBeInTheDocument();
+  expect(screen.getByTestId('time-field').className).toMatchInlineSnapshot(
+    `"group/field w-full"`
+  );
+});
+
+test('renders readOnly, disabled, required states', () => {
```
aromko, line 131: Also here split up these tests.

packages/components/src/TimeField/TimeField.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+  expect(onFocusSpy).toHaveBeenCalled();
+  expect(onFocusChangeSpy).toHaveBeenCalled();
+  expect(onBlurSpy).not.toHaveBeenCalled();
+
+  await userEvent.tab();
+  expect(segments[1]).toHaveFocus();
+});
+
+test('supports variant and size', () => {
```
aromko, line 107: This test is quite useless because it does not reflect the real component. We have no size and variant for this component as far as I see. You could write a more general test to check if all styles or at least the important ones are applied. E.g. segment styles are special for this component.

themes/theme-b2b/src/components/Input.styles.ts
```diff
@@ -25,6 +25,7 @@ export const inputHover = {
 export const Input: ThemeComponent<'Input'> = {
   icon: cva('left-1'),
   action: cva(''),
+  segment: cva(''),
```
sarahgm, line 28: Hmm I'm not sure but maybe you could also reuse the `DateField` styles, because there is already the segment defined, or we can reuse this style to style the `DateField` instead of own date field styles. What do you think? @pauleitenbichler @aromko

packages/components/src/TimeField/TimeField.test.tsx
```diff
@@ -0,0 +1,184 @@
+import { parseTime } from '@internationalized/date';
```
aromko, line 1: In general this tests follow our old structure - you couldn't know. Since some weeks we are doing some migrations for our tests. You can look for this here and apply these changes. Don't hesitate to ask if something unclear. 
https://reservix.atlassian.net/wiki/spaces/DST/pages/3626336415/Refactoring+writing+tests+in+storybook

themes/theme-b2b/src/components/Input.styles.ts
pauleitenbichler: I changed that. It's working better that way I think

packages/components/src/TimeField/TimeField.stories.tsx
```diff
@@ -0,0 +1,135 @@
...
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Basic: Story = {
+  render: args => <TimeField defaultValue={parseTime('13:45:30')} {...args} />,
+};
+
+export const FocusEvents: Story = {
+  args: {
+    label: 'Time',
```
aromko, line 121: You don't need the label here it's the same like in the default args.

packages/components/src/TimeField/TimeField.test.tsx
```diff
@@ -0,0 +1,78 @@
+import { composeStories } from '@storybook/react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
```
aromko, line 3: Can be removed if it's not used anymore.

PR #4620, opened 2025-05-23
feat([DST-863]): Improve `Accordion.Header` when using additional content in the header.

sebald:


packages/components/src/Accordion/Accordion.stories.tsx
```diff
@@ -246,29 +246,43 @@ export const CoreExample: Story = {
...
-      </Accordion.Item>
-    </Accordion>
+    <Stack space={8}>
+      <Accordion {...args}>
+        <Accordion.Item id="1">
+          <Columns columns={[1, 'fit']} space={4}>
+            <Accordion.Header>
+              Accordion with button on the right
+            </Accordion.Header>
+            <Button onPress={() => alert('Do NOT click! Come on!')}>
```
aromko, line 256: Is it intended to have the button after the chevron or outside the header? It looks really weird to me.

sebald: from the code perspective or the rendered UI?

aromko: rendered UI

sebald: I just updated the Story :) Was here before. I think it was just to show that it is possible.

aromko (approved):

PR #4627, opened 2025-05-28
feat([DST-866]): Support align properties for `<Grid>`

sarahgm:
# Description

added properties for `alignX` and `alignY`

# Reviewers:
@marigold-ui/developer
@marigold-ui/designer

### DIFF
--- a/packages/components/src/Grid/Grid.test.tsx
+++ b/packages/components/src/Grid/Grid.test.tsx
@@ -108,3 +108,22 @@ test('converts numbers to fractions in a grid', () => {
     'grid-template-rows': '1fr 2fr 1fr',
   });
 });
+
+test('allows to set alignment', () => {
+  render(
+    <Grid
+      data-testid="layout"
+      areas={['header']}
+      columns={['auto']}
+      rows={['auto']}
+      alignX="center"
+      alignY="center"
+    >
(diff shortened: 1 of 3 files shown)


packages/components/src/Grid/Grid.test.tsx
```diff
@@ -108,3 +108,22 @@ test('converts numbers to fractions in a grid', () => {
...
+      columns={['auto']}
+      rows={['auto']}
+      alignX="center"
+      alignY="center"
+    >
+      <Grid.Area name="header" />
+    </Grid>
+  );
+
+  const layout = screen.getByTestId('layout');
```
aromko, line 126: blank line missing AAA

packages/components/src/Grid/Grid.test.tsx
```diff
@@ -108,3 +108,22 @@ test('converts numbers to fractions in a grid', () => {
...
 });
+
+test('allows to set alignment', () => {
+  render(
+    <Grid
+      data-testid="layout"
+      areas={['header']}
+      columns={['auto']}
+      rows={['auto']}
+      alignX="center"
```
aromko: I'm wondering why you use only center - also in the story? Do we only need this? If not you could add tests for every value e.g. with parametrized test.

PR #4669, opened 2025-07-01
[DSTSUP-161]: adjust multiselect styles

aromko:
# Description

The Multiselect styles matches now our default input styles especially in height.

# What should be tested?

check the styles or change a story and put a Multiselect near to a Inputfield (Inline)

# Reviewers:

@marigold-ui/developer

Bildschirmfoto 2025-07-01 um 17 54 26


themes/theme-rui/src/components/Multiselect.styles.ts
```diff
@@ -1,7 +1,7 @@
 import { ThemeComponent, cva } from '@marigold/system';
 
 export const MultiSelect: ThemeComponent<'MultiSelect'> = {
-  field: cva('grid gap-y-0.5'),
+  field: cva('w-fit space-y-2'),
```
sarahgm: Shouldn't be the `width` 100% per default and than customizable with `width` property, like all input fields?

themes/theme-rui/src/components/Multiselect.styles.ts
```diff
@@ -1,7 +1,7 @@
 import { ThemeComponent, cva } from '@marigold/system';
 
 export const MultiSelect: ThemeComponent<'MultiSelect'> = {
-  field: cva('grid gap-y-0.5'),
+  field: cva('w-full space-y-2'),
```
sarahgm: the w-full must be removed otherwise the property will not work, or its is not working correctly in storybook 🤔

aromko: I think it should work now.

sarahgm (approved):

PR #4672, opened 2025-07-02
feat([DSTSUP-158]): List accessibility for `<Stack>` component

sarahgm:
Added possibility to render a list in stack
updated docs for List to new component guidelines

# Reviewers:
@marigold-ui/developer
@marigold-ui/designer


packages/components/src/Stack/Stack.tsx
```diff
@@ -36,10 +42,16 @@ export const Stack = ({
   stretch = false,
   alignX,
   alignY,
+  asList = false,
   ...props
 }: StackProps) => {
+  const Component = asList ? 'ul' : 'div';
+  const StackChildren = asList
```
sebald: lowercase, since it is not a component but a variable that holds a list of components

.changeset/perfect-bikes-hang.md
```diff
@@ -0,0 +1,8 @@
+---
+"@marigold/docs": patch
+"@marigold/components": patch
```
sebald:
```suggestion
"@marigold/components": minor 
```

New feature!

sebald (changes requested):

PR #4694, opened 2025-07-09
refa([DST-904]): Deprecate className Prop from <Button> Component

OsamaAbdellateef:
### ⚠️⚠️⚠️ Breaking Changes ⚠️⚠️⚠️
Now , Button component doesn't accept prop className, and as a result : 
- Created a new variant for docs theme `<Button variant="icon" >`
- Created SelectListAction to handle button position when using a button or another component for actions 

# What should be tested?
Button component doesn't accept className anymore and it can accept size, variant as a replacement .

# Reviewers:

@marigold-ui/developer

### DIFF
--- a/docs/content/components/form/button/button.mdx
+++ b/docs/content/components/form/button/button.mdx
@@ -247,7 +247,7 @@ To visually imitate a button element that navigates (e.g., a "Cancel" action tha
 
 ### Press Event
 
-The `<Button>` component supports user interactions via mouse, keyboard, and touch, handled through the `onPress` prop. This prop is similar to the standard `onClick` event but is normalized for all interaction methods. Additionally, `onPressStart`, `onPressEnd`, and `onPressChange` events are fired during user interaction.
+The `<Button>` component supports user interactions via mouse, keyboard, and touch, handled through the `onPress ` prop. This prop is similar to the standard `onClick` event but is normalized for all interaction methods. Additionally, `onPressStart`, `onPressEnd`, and `onPressChange` events are fired during user interaction.
 
 Each handler receives a `PressEvent`, providing details about the target and the type of event that triggered the interaction, like the pointer (mouse, keyboard or touch) that was used.
 


docs/ui/CopyButton.tsx
```diff
@@ -17,32 +17,30 @@ export const CopyButton = ({ codeString }: CopyProps) => {
...
-        </svg>
-      ) : (
-        <svg
-          className="size-5 fill-white"
-          viewBox="0 0 16 16"
-          fill="text.regular"
-        >
-          <path
-            d="M8.8421 0H1.26316C0.568421 0 0 0.568421 0 1.26316V10.1053H1.26316V1.26316H8.8421V0ZM10.7368 2.52632H3.78947C3.09474 2.52632 2.52632 3.09474 2.52632 3.78947V12.6316C2.52632 13.3263 3.09474 13.8947 3.78947 13.8947H10.7368C11.4316 13.8947 12 13.3263 12 12.6316V3.78947C12 3.09474 11.4316 2.52632 10.7368 2.52632ZM10.7368 12.6316H3.78947V3.78947H10.7368V12.6316Z"
+    <div className="[&>button]:translate-y-0.5 [&>button]:border-none [&>button]:p-0 [&>button]:outline-0">
```
sebald: New `variant` in the docs theme? :)

docs/ui/FullsizeViewDemo.tsx
```diff
@@ -9,20 +9,22 @@ export interface FullsizeViewProps {
...
-      <svg width="20px" height="20px" viewBox="0 0 24 24">
-        <path
-          className="fill-white"
-          fillRule="evenodd"
-          d="M15 3.75a.75.75 0 0 1 .75-.75h4.5a.75.75 0 0 1 .75.75v4.5a.75.75 0 0 1-1.5 0V5.56l-3.97 3.97a.75.75 0 1 1-1.06-1.06l3.97-3.97h-2.69a.75.75 0 0 1-.75-.75Zm-12 0A.75.75 0 0 1 3.75 3h4.5a.75.75 0 0 1 0 1.5H5.56l3.97 3.97a.75.75 0 0 1-1.06 1.06L4.5 5.56v2.69a.75.75 0 0 1-1.5 0v-4.5Zm11.47 11.78a.75.75 0 1 1 1.06-1.06l3.97 3.97v-2.69a.75.75 0 0 1 1.5 0v4.5a.75.75 0 0 1-.75.75h-4.5a.75.75 0 0 1 0-1.5h2.69l-3.97-3.97Zm-4.94-1.06a.75.75 0 0 1 0 1.06L5.56 19.5h2.69a.75.75 0 0 1 0 1.5h-4.5a.75.75 0 0 1-.75-.75v-4.5a.75.75 0 0 1 1.5 0v2.69l3.97-3.97a.75.75 0 0 1 1.06 0Z"
-          clipRule="evenodd"
-        />
-      </svg>
-    </Button>
+    <div className="[&>button]:hidden [&>button]:border-none [&>button]:p-0 [&>button]:outline-0 [&>button]:md:block">
```
sebald: also variant? :)

packages/components/src/SelectList/SelectList.stories.tsx
```diff
@@ -82,15 +82,16 @@ export const Action: Story = {
...
-          <Button
-            aria-label="Info"
-            onPress={() => alert(`Info for ${item.name}...`)}
-            className="ml-auto"
-          >
-            <svg width="30px" height="30px" viewBox="0 0 24 24">
-              <path d="M12 2.85938C6.95437 2.85938 2.85938 6.95437 2.85938 12C2.85938 17.0456 6.95437 21.1406 12 21.1406C17.0456 21.1406 21.1406 17.0456 21.1406 12C21.1406 6.95437 17.0456 2.85938 12 2.85938ZM12.7875 15.9374H11.2125V11.2124H12.7875V15.9374ZM12.7875 9.6375H11.2125V8.0625H12.7875V9.6375Z"></path>
-            </svg>
-          </Button>
+          <div className="ml-auto">
```
sebald: does this mean our users always need to wrap this in a ml-auto? 🙈

sebald (changes requested):

docs/ui/CopyButton.tsx
OsamaAbdellateef: I added a new variant as `icon` but the issue now it is conflicting with size variant since the size has `px-3 py-2` by default , 

`  variants: {
    variant: {
      ghost: 'text-secondary-700 hover:text-secondary-900 p-0',
      sunken:
        'text-secondary-600 hover:bg-secondary-400/20 bg-secondary-400/10 h-8 justify-start rounded-lg',
      inverted: 'bg-secondary-100',
      icon: 'translate-y-0.5 border-none outline-none p-0',
    },
    size: {
      small: 'px-3 py-1.5 text-sm',
      default: 'px-3 py-2',
    },
  },`
  Also we need to move padding from variant icon since it is related to size ? 
  @sebald

packages/components/src/SelectList/SelectList.stories.tsx
OsamaAbdellateef: What about creating something like prop called `action` on SelectListItem or create <SelectList.Action> @sebald

docs/ui/CopyButton.tsx
sebald: Hm. Maybe we use this to fix it and make it the same as we have a good icon because in the docs?

packages/components/src/SelectList/SelectList.stories.tsx
sebald: And use grid areas to place it? Think it is a good idea. 

An alternative would be to have a "> button" class on the component I guess?

OsamaAbdellateef: yeah grid would good idea, also i used  grid order to order the action to be always last  @sebald

sebald:
@OsamaAbdellateef typecheck is failing :) will approve after it's fixed

OsamaAbdellateef:
> @OsamaAbdellateef typecheck is failing :) will approve after it's fixed

:heavy_check_mark:  @sebald

PR #4714, opened 2025-07-19
fix(DST-942): Fix `<Drawer>` on mobile and refactor `<OverlayContainerProvider>`

sebald:
Making the `<Dialog>` responsive cause the `<Drawer>` to be broken on small screens. Fixed it by not reusing the `<Dialog>` from Marigold but instead use it directly from RAC. This should also prevent regressions in the future.

Also, refacoted how the `<OverlayContainerProvider>` works. Previously there were `<PortalProvider>` (from RAC) around a lot of overlay components. This is unnecessary and can be set once per App. The `<OverlayContainerProvider>` now contains the `<PortalProvider>` and manages where overlays will get rendered.

### DIFF
--- a/packages/components/src/Toast/ToastProvider.tsx
+++ b/packages/components/src/Toast/ToastProvider.tsx
@@ -4,9 +4,7 @@ import {
 } from 'react-aria-components';
 import type RAC from 'react-aria-components';
 import { flushSync } from 'react-dom';
-import { UNSAFE_PortalProvider } from '@react-aria/overlays';
 import { useClassNames } from '@marigold/system';
-import { usePortalContainer } from '../Provider';
 import { Toast } from './Toast';
 import { ToastContentProps } from './Toast';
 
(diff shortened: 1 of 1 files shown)


packages/components/src/Provider/OverlayContainerProvider.test.tsx
```diff
@@ -1,13 +1,9 @@
 /**
```
aromko, line 1: I think here can more tests be added to coverage all cases, e.g. no container given etc.

sebald: I tried that but for reasons this doesn't work in JSDOM 💩

packages/components/src/Provider/OverlayContainerProvider.tsx
```diff
@@ -1,21 +1,40 @@
...
 
-// needs an id for the portal container, it will be overgiven as string,
-// if there is no the body is used
-const OverlayContainerContext = createContext<string | undefined>(undefined);
-
-export const OverlayContainerProvider = OverlayContainerContext.Provider;
-
-export const usePortalContainer = () => {
-  const portalContainer = useContext(OverlayContainerContext);
-  const isSSR = useIsSSR();
```
sarahgm, line 12: why is this not needed anymore?

sebald: because the `PortalContainer` already does this :)

sarahgm (dismissed):

PR #4717, opened 2025-07-21
[DST-887]: put certain pagination styles into the styles file

aromko:
# Description

During creating pagination styles in the core I have noticed that not all styles are generated with the core script.

# Reviewers:

Examples:
@marigold-ui/developer


packages/components/src/Pagination/Pagination.tsx
```diff
@@ -104,10 +106,10 @@ const InnerPagination = ({
         position="left"
         onKeyDown={handleKeyDown(() => handlePageChange(currentPage - 1))}
       >
-        <ChevronLeft className={classNames.icon} />
+        <ChevronLeft className={icon} />
```
sebald, line 109: why themove the `classNames` prefix. until now we used it everywhere.

packages/system/src/types/theme.ts
```diff
@@ -112,7 +112,12 @@ export type Theme = {
       ComponentStyleFunction<string, string>
     >;
     Pagination?: Record<
-      'navigationButton' | 'pageButton' | 'icon',
+      | 'navigationContainer'
+      | 'navigationButton'
```
sebald: What is the difference in stylew from the page and navigation button?

packages/components/src/Pagination/Pagination.tsx
```diff
@@ -151,10 +153,11 @@ const _Pagination = ({
   const [currentPage, setCurrentPage] = useState(page ?? defaultPage);
   const totalPages = Math.ceil(totalItems / pageSize);
   const pageRange = usePageRange({ currentPage, totalPages });
+  const { navigationContainer } = useClassNames({ component: 'Pagination' });
 
   return (
     <nav
-      className="flex items-center justify-center space-x-2"
+      className={navigationContainer}
```
sebald: we usually call this just `container` in other components

packages/system/src/types/theme.ts
sebald: maybe we can merge them before we make this "public"?

aromko: I splitted it up. Now you can see it better

packages/components/src/Pagination/Pagination.tsx
aromko: I know and  it's annoying. Because IMHO it is useless to read it all the time 😅 I gives me no helpful info. And when classname is before I know what it is about - so it's kind of doubled

packages/components/src/Pagination/Pagination.tsx
aromko: Yes, I had thought of that too, but thought it was clearer because we have 2 containers. But no problem to use for the outer thing just container.

packages/components/src/Pagination/Pagination.tsx
sebald: Yes, but especially "icon" is a loaded term. Could also be a JSX component :D

packages/system/src/types/theme.ts
sebald: What can I see better?

packages/components/src/Pagination/PageButton.tsx
```diff
@@ -25,7 +25,7 @@ export const PageButton = (props: PageButtonProps) => {
       {...buttonProps}
       aria-label={`Page ${page}`}
       aria-current={selected ? 'page' : undefined}
-      className={classNames.pageButton}
+      className={cn(button, pageButton)}
```
sarahgm: Why are there two styles applied now?

packages/system/src/types/theme.ts
sarahgm: I think I understand, button is for all buttons and than additionally either pagebutton or navigationbutton styles on top? I wonder whats the differennce between pagebuttoncontainer and container. only one class i think? maybe these can be merged in one?

sebald: yeah same :D thought they should look exactly the same

packages/components/src/Pagination/Pagination.tsx
aromko: when className is in front of it can be a JSX component?

packages/components/src/Pagination/PageButton.tsx
aromko: You can see it in the diff of the Pagination.styles. Before they were a lot of same styles in pageButton and navigationButton. So I put these into the button slot. But I think we need no slot I do it in a separate const.

packages/system/src/types/theme.ts
aromko: pageButton and navigationButton looks different. I changed it to use one container class seems also to work.

packages/components/src/Pagination/PageButton.tsx
sebald: Do we need two different styles here? I cant see a difference 👀

aromko: I changed it

packages/components/src/Pagination/PageButton.tsx
```diff
@@ -25,7 +25,7 @@ export const PageButton = (props: PageButtonProps) => {
       {...buttonProps}
       aria-label={`Page ${page}`}
       aria-current={selected ? 'page' : undefined}
-      className={cn(button, pageButton)}
+      className={cn(pageButton)}
```
sarahgm: you don't need `cn` then here

sarahgm (approved):

PR #4722, opened 2025-07-24
refa([DST-969]): Remove obsolete "text" variant from `<Button>`

sebald:


aromko (approved):
In Input.stories the text variant is still used.

PR #4732, opened 2025-07-30
fix([DSTSUP]): Fix alignment of buttons and inputs in `Inline` component

OsamaAbdellateef:
# Description

# What should be tested?

# Reviewers:

Examples:
@marigold-ui/developer
@marigold-ui/designer

### DIFF
--- a/docs/content/components/layout/inline/inline.mdx
+++ b/docs/content/components/layout/inline/inline.mdx
@@ -1,6 +1,7 @@
 ---
 title: Inline
 caption: Display children horizontally in a row.
+badge: updated
 ---
 
 The `<Inline>` component is a responsive layout component based on [CSS Flexbox](https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox). It aligns its content horizontally in a row and automatically wraps to a new line when there isn't enough space on the screen.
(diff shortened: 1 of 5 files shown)


sebald:
Can you show this tomorrow in mob. I don't get the bug we had 🙈

docs/content/components/layout/inline/inline.mdx
```diff
@@ -29,7 +30,7 @@ For horizontally alignment you can use the `alignX` prop.
 
 ### Form Input Alignment
 
-When aligning form inputs with buttons or elements, use `alignY="center"` to mantain vertically alignment with the input field rectangle. This automatically adjusts when descriptions or error messages appear.
+When aligning form inputs with buttons , use `alignToInputBaseline={true}` to mantain vertically alignment with the input field rectangle. This automatically adjusts when descriptions or error messages appear.
```
aromko: I think in the last mob we said not to use a new prop for this. Instead extend the alignY="input" or something like this.

Or am I wrong @sebald @sarahgm ?

OsamaAbdellateef: yeah sorry , it is already done as we said but i forgot to update the docs 
now should be updated  @aromko

sarahgm: is it now with a prop or not? 🤔

aromko (approved):

PR #4733, opened 2025-07-30
refa([DST-978]): Add item and section as child component of `ActionMenu`

sebald:


.changeset/metal-plants-ring.md
```diff
@@ -0,0 +1,6 @@
+---
+"@marigold/docs": patch
+"@marigold/components": patch
+---
+
+refa([DST-978]): Add item and section as child component of `ActionMenu`
```
sarahgm, line 6: Wouldn't this be a breaking Change?

sebald: Why? :-O I just added stuff, They way it worked before still works. I think?

aromko (approved):

.changeset/metal-plants-ring.md
sarahgm: You think or you know?  I thought because of ActionMenu.Item should be used instead

sebald: I know :P

PR #4737, opened 2025-07-31
[DST-986]: adjusted the vrt workflow

aromko:
# Description

The VRT workflow should be triggered again for every PR with the following restrictions:

- not for branches created by renovate
- PR isn't a draft
- PR hasn't the label run-vrt

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/.github/workflows/visual-regression-tests.yml
+++ b/.github/workflows/visual-regression-tests.yml
@@ -1,13 +1,13 @@
 name: Visual-Regression-Tests
 on:
-  push:
-    branches:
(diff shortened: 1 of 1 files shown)


aromko:
@sarahgm , @OsamaAbdellateef , @sebald can be reviewed. The Visual-Regression-Test failed the first time because one Autocomplete was broken. I just reruned them directly on chromatic and it's now green. Sadly it doesn't reflect this state here without doing a commit again.

PR #4766, opened 2025-08-11
fix([DSTSUP-176]): Calendar - Hide, disable, limit the MonthListBox and YearListBox depening on minValue and maxValue

sinan-rsvx:
# Description
Partially refactored the Calendar code to improve structure and maintainability. Now, the month and year dropdown buttons are automatically disabled when only one option is available, and the dropdowns themselves correctly disable non-selectable months or years based on min/max values. Additionally, the dropdown can be closed again even when minValue and maxValue are set to the same year or month.

# Reviewers:

@marigold-ui/developer


packages/components/src/Calendar/Calendar.tsx
```diff
@@ -55,19 +59,26 @@ export interface CalendarProps
...
   readOnly,
   size,
   variant,
   width = 'fit',
   dateUnavailable,
+  minValue: _minValue,
+  maxValue: _maxValue,
   ...rest
 }: CalendarProps) => {
+  const minValue = _minValue ?? undefined;
```
sebald: Why is this necessary? Shouldn't it be `undefined` by default? 🤔

packages/components/src/Calendar/ListBox.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { cn } from '@marigold/system';
+
+interface ListBoxProps<T> {
+  dataTestid: string;
+  items: T[];
+  isDisabled: (item: T, index: number) => boolean;
+  isSelected: (item: T, index: number) => boolean;
+  onSelect: (item: T, index: number) => void;
+  format: (item: T, index: number) => ReactNode;
+  buttonClassName?: string;
```
sebald: instead of passing this down, use `useCalendarContext`

sebald: some for the other classes?

packages/components/src/Calendar/ListBox.tsx
```diff
@@ -0,0 +1,60 @@
...
+  ulClassName?: string;
+  activeButtonRef?: React.RefObject<HTMLButtonElement>;
+}
+
+export function ListBox<T>({
```
sebald, line 18: Guess this is not accessible like this. What about using the `react-aria` listbox?

packages/components/src/Calendar/Calendar.stories.tsx
```diff
@@ -52,9 +52,6 @@ const meta = {
       },
     },
   },
-  args: {
-    defaultValue: new CalendarDate(2025, 8, 7),
```
aromko: I think this needs to be undone, because it was intended to have fixed value here. Otherwise our VRT always shows changes. You can override this argument for each story separately if needed.

packages/components/src/Calendar/Calendar.stories.tsx
```diff
@@ -152,3 +149,21 @@ export const ReadOnly: Story = {
...
+
+export const OnlyOneMonthAndYear: Story = {
+  ...Basic,
+  tags: ['component-test'],
+  args: {
+    minValue: new CalendarDate(2020, 5, 5),
+    maxValue: new CalendarDate(2020, 5, 20),
+    onChange: fn(),
+  },
+  play: async ({ canvasElement }) => {
```
aromko, line 187: This could also be moved in the normal .test.tsx. It isn't really a behavior test. If you want a behavior test you can add something like trying to click on month and year and nothing should open and so on.

packages/components/src/Calendar/calendarListBoxSelectableCheck.ts
```diff
@@ -0,0 +1,21 @@
+import { DateValue } from 'react-aria-components';
```
aromko, line 1: You can write tests for this class.

packages/components/src/Calendar/YearListBox.tsx
```diff
@@ -1,90 +1,49 @@
...
-              </Button>
-            </div>
-          </li>
-        );
-      })}
-    </ul>
+    <ListBox
+      dataTestid="yearOptions"
+      items={years}
+      isDisabled={year => year.year < minYear || year.year > maxYear}
```
aromko: Please destructure year.

packages/components/src/Calendar/MonthListBox.tsx
```diff
@@ -1,51 +1,46 @@
...
   const state = useContext(CalendarStateContext)!;
   const months = useFormattedMonths(state.timeZone, state.focusedDate);
+  const currentYear = state.focusedDate.year;
 
-  const onPress = (index: Key) => {
-    const date = state.focusedDate.set({ month: Number(index) + 1 });
-    state.setFocusedDate(date);
-    setSelectedDropdown(undefined);
-  };
+  const minMonth =
```
aromko, line 21: These are also not covered by tests. Run `pnpm test:coverage` to generate coverage report. Than open coverage/index.html in the browser and chek Calendar component.

packages/components/src/Calendar/MonthListBox.tsx
```diff
@@ -1,51 +1,46 @@
...
-              {month.substring(0, 3)}
-            </Button>
-          </li>
-        );
-      })}
-    </ul>
+    <ListBox
+      dataTestid="monthOptions"
+      items={months}
+      isDisabled={(_, monthIndex) =>
```
aromko, line 30: A more general question. Using disabled also means the month and year box are not accessible right? So someone who is blind doesn't know which month or year is selected here. I think the styles are fine but we should rethink the disabled state.

packages/components/src/Calendar/Calendar.stories.tsx
sinan-rsvx: @sarahgm removed it here (#4758) three weeks ago. Should I add it back again?

sarahgm: @sinan-rsvx I added it because otherwise Chromatic has every day new changes in the Calendar component :) So yes please add back

packages/components/src/Calendar/MonthListBox.tsx
sinan-rsvx: I now use React Aria list boxes, so the list boxes should be accessible.

aromko (approved):

PR #4772, opened 2025-08-12
refa([DST-1007]): Rename "secondary" to "default" variant and fix appearance demo

sebald:


sarahgm:
Can you remove the variant secondary from the `Menu.stories.tsx` ?

PR #4774, opened 2025-08-13
feat([DST-1008]): Introduce a "remove all" function for `<Tag.Group>`

sebald:
Introduce a new feature that allows users to remove all tags at once. Update documentation and examples to reflect this new functionality.

### DIFF
--- a/packages/components/src/TagGroup/TagGroup.stories.tsx
+++ b/packages/components/src/TagGroup/TagGroup.stories.tsx
@@ -1,7 +1,7 @@
 import type { Meta, StoryObj } from '@storybook/react';
 import { I18nProvider } from 'react-aria-components';
 import { useState } from 'storybook/preview-api';
-import { expect, fn, within } from 'storybook/test';
+import { expect, fn, waitFor, within } from 'storybook/test';
 import { Key } from '@react-types/shared';
 import { Tag } from '.';
 import { Button } from '../Button';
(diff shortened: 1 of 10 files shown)


packages/components/src/TagGroup/TagGroup.stories.tsx
```diff
@@ -86,3 +84,59 @@ export const RemovableTags: Story = {
...
+          >
+            {(item: { id: number; name: string }) => <Tag>{item.name}</Tag>}
+          </Tag.Group>
+          <Button onPress={() => setItems(defaultItems)}>Reset</Button>
+        </Stack>
+      </I18nProvider>
+    );
+  },
+  play: async ({ canvas, userEvent }) => {
+    expect(canvas.getByText('News')).toBeInTheDocument();
```
aromko: This is not AAA.

packages/components/src/TagGroup/TagGroup.stories.tsx
```diff
@@ -86,3 +84,59 @@ export const RemovableTags: Story = {
...
+    expect(canvas.getByText('Travel')).toBeInTheDocument();
+    expect(canvas.getByText('Gaming')).toBeInTheDocument();
+    expect(canvas.getByText('Shopping')).toBeInTheDocument();
+
+    const removeAll = canvas.getByText('Remove all');
+    await userEvent.click(removeAll);
+
+    expect(canvas.queryByText('News')).toBeInTheDocument();
```
aromko: Why they should be in the document after removing them? They should be removed and the empty text should be there or?

packages/components/src/TagGroup/TagGroup.tsx
```diff
@@ -41,19 +48,28 @@ const _TagGroup = ({
...
-      </TagList>
+      <div className={classNames.container}>
+        <TagList
+          items={items}
+          className={classNames.listItems}
+          renderEmptyState={emptyState}
+        >
+          {children}
+        </TagList>
+        {rest.onRemove && removeAll ? (
```
aromko, line 66: Why no destructuring?

sebald: because the onRemove also needs to get passed to the TagGroup (in the FieldBase)

aromko (approved):

PR #4776, opened 2025-08-14
refa([DST-1010]): Separate `aria-label` and `name` for slider's thumbs

sebald:
**Breaking Change:** The aria-label (`thumbLabels`) and name attributes for slider thumbs are now separate. Update implementations to set both explicitly if needed.

### DIFF
--- a/packages/components/src/Slider/Slider.stories.tsx
+++ b/packages/components/src/Slider/Slider.stories.tsx
@@ -80,11 +80,11 @@ export const Basic: Story = {
 };
 
 export const ValueFormatting: Story = {
+  args: { label: 'Price' },
   render: args => (
     <I18nProvider locale="en-US">
       <Stack space={4}>
         <Slider
-          label="Price"
           {...args}
           formatOptions={{ style: 'currency', currency: 'EUR' }}
         />
@@ -123,17 +123,13 @@ export const MultipleThumbs: Story = {
 };
 
 export const Controlled: Story = {
+  args: { label: 'Cookies to buy' },
   render: args => {
     const [value, setValue] = useState<number | number[]>(25);
 
     return (
       <Stack space={2}>
-        <Slider
-          value={value}
-          onChange={setValue}
-          {...args}
-          label="Cookies to buy"
-        />
+        <Slider value={value} onChange={setValue} {...args} />
         <p>Current value: {value}</p>
       </Stack>
     );
@@ -141,17 +137,13 @@ export const Controlled: Story = {
 };
 
 export const MultiThumbsControlled: Story = {
+  args: { label: 'Tickets for sale' },
   render: args => {
     const [value, setValue] = useState<number | number[]>([25, 75]);
 
     return (
       <Stack space={2}>
-        <Slider
-          value={value}
-          onChange={setValue}
-          {...args}
-          label="Tickets for sale"
-        />
+        <Slider value={value} onChange={setValue} {...args} />
         <p>
           Current value: {typeof value !== 'number' ? value?.join(' - ') : null}
         </p>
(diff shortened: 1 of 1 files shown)


.changeset/smooth-ads-peel.md
```diff
@@ -0,0 +1,8 @@
+---
+"@marigold/docs": major
+"@marigold/components": major
+---
+
+refa([DST-1010]): Separate `aria-label` and `name` for slider's thumbs
+
+**Breaking Change:** The aria-label (`thumbLabels`) and name attributes for slider thumbs are now separate. Update implementations to set both explicitly if needed.
```
aromko: `name` can also be set in code

docs/content/components/form/slider/slider-appearance.demo.tsx
```diff
@@ -1,5 +1,6 @@
+import type { SliderProps } from '@marigold/components';
 import { Slider } from '@marigold/components';
 
-export default () => {
-  return <Slider defaultValue={50} label="Volume (%)" />;
-};
+export default (props: SliderProps<any>) => (
+  <Slider defaultValue={50} label="Volume" width="2/3" {...props} />
```
aromko, line 5: Can we agree to put props first in general? I see it very often and sometimes the attributes before has no effect because they are overwritten by props :-/

packages/components/src/Slider/Slider.stories.tsx
```diff
@@ -66,62 +67,75 @@ const meta = {
     },
   },
   args: {
-    label: 'Label',
```
aromko, line 69: Why removing these default values?

packages/components/src/Slider/Slider.stories.tsx
```diff
@@ -66,62 +67,75 @@ const meta = {
...
-        step={0.01}
-        maxValue={1}
-        label="Percent"
-      />
-    </Stack>
+    <I18nProvider locale="en-US">
+      <Stack space={4}>
+        <Slider
+          label="Price"
+          {...args}
```
aromko, line 88: Also here args should be first especially when having default values.

docs/content/components/form/slider/slider-appearance.demo.tsx
sebald: it does not make any difference here. Because it will only get passed the variant and size 🤔

packages/components/src/Slider/Slider.stories.tsx
sebald: so it doesn't override some stories that already have a label (e.g. price) so the demo makes more sense

packages/components/src/Slider/Slider.stories.tsx
sebald: but then the laben is never applied

packages/components/src/Slider/Slider.stories.tsx
aromko: it shouldn't affect them, when {...args} comes always first. Than you can override everything you want. But the default values are useful to not apply attribute all the time.

packages/components/src/Slider/Slider.stories.tsx
aromko: hä no. everything after {..args} override the args. Or am I wrong?

packages/components/src/Slider/Slider.stories.tsx
sebald: Yes, but since there was a label by default, the label of this story never applied

packages/components/src/Slider/Slider.stories.tsx
sebald: Yes. In this case. There is no default label. So it can be applied if you set it in storybook

aromko: Yeah because you deleted the default values above 🤣

sebald: yeah I think I got now what you mean. I will add the labels als defaults per story

aromko: hm still not 100% but we let it now 😄

PR #4779, opened 2025-08-15
fix: correct broken classnames in `<Pagination>` and `<Slider>`

sebald:


.changeset/tasty-laws-study.md
```diff
@@ -0,0 +1,5 @@
+---
```
aromko, line 1: @sebald I think VRT should run here

PR #4788, opened 2025-08-21
bugfix([DST-935]):🐛Controlled Dialog doesn't close when dismissable is set 

OsamaAbdellateef:
# Description
Dismissing dialog isn't working when it's controlled 

# Test Instructions:
 commandK menu in our docs should close when clicking outside 

# Reviewers:

@marigold-ui/developer

### DIFF
--- a/.changeset/blue-pumas-ask.md
+++ b/.changeset/blue-pumas-ask.md
@@ -3,4 +3,4 @@
 '@marigold/docs': patch
 ---
 
-bugfix([DST-935]): Controlled `<Dialog>` doesn't close when dismissable is set to `true`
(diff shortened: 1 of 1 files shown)


.changeset/blue-pumas-ask.md
```diff
@@ -0,0 +1,6 @@
+---
+'@marigold/components': patch
+'@marigold/docs': patch
+---
+
+bugfix([DST-935]): Controlled `<Dialog>` doesn't close when dismissable is set to `true`
```
aromko: @OsamaAbdellateef I would write it in the way how it is now. e.g. Controlled `<Dialog>`close now correctly when `dismissable` is set to `true`

aromko (approved):

PR #4794, opened 2025-08-25
feat([DST-1003]):  Update `variant` property to use defined types 

OsamaAbdellateef:
# Description

# Test Instructions:

-Make sure that every component variants and sizes are defined in the components types 

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -14,9 +14,6 @@ export interface ButtonProps extends Omit<RAC.ButtonProps, RemovedProps> {
     | 'destructive'
     | 'ghost'
     | 'link'
-    | 'sunken'
-    | 'inverted'
-    | 'icon'
     | (string & {});
   size?: 'default' | 'small' | 'large' | 'icon' | (string & {});
 
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -8,8 +8,17 @@ import { ProgressCycle } from '../ProgressCycle';
 type RemovedProps = 'isDisabled' | 'isPending' | 'className' | 'style';
 
 export interface ButtonProps extends Omit<RAC.ButtonProps, RemovedProps> {
-  variant?: string;
-  size?: string;
+  variant?:
+    | 'primary'
+    | 'secondary'
+    | 'destructive'
+    | 'ghost'
+    | 'link'
+    | 'sunken'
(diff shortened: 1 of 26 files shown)


packages/components/src/Button/Button.tsx
```diff
@@ -8,8 +8,17 @@ import { ProgressCycle } from '../ProgressCycle';
...
 export interface ButtonProps extends Omit<RAC.ButtonProps, RemovedProps> {
-  variant?: string;
-  size?: string;
+  variant?:
+    | 'primary'
+    | 'secondary'
+    | 'destructive'
+    | 'ghost'
+    | 'link'
+    | 'sunken'
```
aromko: From where sunken, icon is coming? Some variants I can't found.

packages/components/src/Link/Link.tsx
```diff
@@ -6,8 +6,16 @@ import { useClassNames } from '@marigold/system';
 type RemovedProps = 'className' | 'isDisabled' | 'slot';
 
 export interface LinkProps extends Omit<RAC.LinkProps, RemovedProps> {
-  variant?: string;
-  size?: string;
+  variant?:
+    | 'default'
+    | 'secondary'
+    | 'primary'
```
aromko: I think I know now whats wrong. For some components you took the variants from the docs or? Also I think we won't them as proposals in our intellisense because in our stories and so on we shouldn't use the variants from the docs.

Our users also get the docs variants or?

OsamaAbdellateef:
> @OsamaAbdellateef see my comments. I think we should only add variant and sizes from the rui theme not from the docs or?

yes you are right i missed some files, but now i reviweed them again hopefully  they are complete now

aromko (changes requested):
These I found which are missing:

- ContextualHelp
- SectionMessage
- Toast
- XLoader

packages/components/src/Table/Table.tsx
```diff
@@ -193,7 +193,7 @@ Table.Header = Header;
 Table.Row = Row;
 
 export interface RowProps extends ReactAriaRowProps<any> {
-  variant?: string;
+  variant?: 'default' | 'grid' | 'admin' | 'master';
```
sarahgm: there is `muted` missing

sarahgm: is this not also missing: `| (string & {});` ?

PR #4797, opened 2025-09-01
feat([DST-981]) ✨Drawer size and placement enhancement

OsamaAbdellateef:
# Description

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/themes/theme-rui/src/components/Drawer.styles.ts
+++ b/themes/theme-rui/src/components/Drawer.styles.ts
@@ -3,15 +3,15 @@ import { ThemeComponent, cva } from '@marigold/system';
 export const Drawer: ThemeComponent<'Drawer'> = {
   overlay: cva([
     'group/overlay',
-    'data-[placement=top]:entering:animate-slide-in-top data-[placement=top]:exiting:animate-slide-out-top data-[placement=top]:top-0 data-[placement=top]:left-0',
-    'data-[placement=bottom]:entering:animate-slide-in-bottom data-[placement=bottom]:exiting:animate-slide-out-bottom data-[placement=bottom]:bottom-0 data-[placement=bottom]:left-0',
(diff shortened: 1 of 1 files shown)


packages/components/src/Drawer/Drawer.tsx
```diff
@@ -17,8 +17,7 @@ import { DrawerTrigger } from './DrawerTrigger';
 // ---------------
 export interface DrawerProps
   extends Omit<DialogProps, 'className' | 'style' | 'isOpen' | 'role'> {
-  variant?: string;
```
sarahgm, line 20: if you remove this we and other cannot use variant anymore. Think we should keep variant

packages/components/src/Drawer/Drawer.tsx
```diff
@@ -53,16 +59,20 @@ export interface DrawerProps
...
   closeButton,
   role = 'complementary',
+  placement = 'right',
   ...props
 }: DrawerProps) => {
   const ref = useRef<HTMLElement | null>(null);
-  const classNames = useClassNames({ component: 'Drawer', variant, size });
+  const classNames = useClassNames({
+    component: 'Drawer',
+    variant: placement,
```
sarahgm: I think placement should be no variant, what if you have a variant and want to use another position🤔

themes/theme-rui/src/components/Drawer.styles.ts
```diff
@@ -1,26 +1,50 @@
...
+          'entering:animate-slide-in-right exiting:animate-slide-out-right top-0 right-0',
+        left: 'entering:animate-slide-in-left exiting:animate-slide-out-left top-0 left-0',
+        top: '[&>:first-child]:h-full entering:animate-slide-in-top exiting:animate-slide-out-top top-0 left-0',
+        bottom:
+          '[&>:first-child]:h-full entering:animate-slide-in-bottom exiting:animate-slide-out-bottom bottom-0 left-0',
       },
-      defaultVariants: {
-        size: 'default',
+      size: {
+        default: 'w-80',
```
sarahgm: shouldn't default also have a meaningful name? default is between medium and small. or default should be medium?

PR #4830, opened 2025-09-14
feat([DST-1039]): Allow `<Text>` to not wrap lines

sebald:


sebald:
The style prop for `xs` text was 13px for reasons, should we flag this as a BR? I didn't find any usage and would rather see this is a legacy bug 🤫

sarahgm (approved):

PR #4843, opened 2025-09-18
feat([DST-1047]): Improve `<List>` styles and add `small` variant

sebald:
# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/themes/theme-rui/src/components/List.styles.ts
+++ b/themes/theme-rui/src/components/List.styles.ts
@@ -1,7 +1,30 @@
 import { ThemeComponent, cva } from '@marigold/system';
 
 export const List: ThemeComponent<'List'> = {
-  ul: cva('ml-6 list-outside list-disc'),
-  ol: cva('ml-6 list-outside list-decimal'),
-  item: cva('not-first:pt-3'),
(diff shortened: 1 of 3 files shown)


themes/theme-rui/src/components/List.styles.ts
```diff
@@ -1,7 +1,30 @@
 import { ThemeComponent, cva } from '@marigold/system';
 
 export const List: ThemeComponent<'List'> = {
-  ul: cva('ml-6 list-outside list-disc'),
-  ol: cva('ml-6 list-outside list-decimal'),
-  item: cva('not-first:pt-3'),
+  ul: cva('ml-6 list-outside list-disc space-y-1 marker:text-foreground/50', {
```
sarahgm, line 4: why `foreground/50`? Thought we do no percent values from colors, should be `muted`?

sebald: Was too light 🫣

sarahgm: do we need a new token then

sebald: IDK I would wait? makes things more complex?

sarahgm (approved):

PR #4844, opened 2025-09-19
fix([DST-1049]): Add 'text-sm' class to Link component styles

sarahgm:
# Description

add font size to link component

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/themes/theme-rui/src/components/Link.styles.ts
+++ b/themes/theme-rui/src/components/Link.styles.ts
@@ -1,7 +1,7 @@
 import { ThemeComponent, cva } from '@marigold/system';
 
 export const Link: ThemeComponent<'Link'> = cva(
-  'aria-[disabled]:cursor-not-allowed',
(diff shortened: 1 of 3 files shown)


themes/theme-rui/src/components/Link.styles.ts
```diff
@@ -1,7 +1,7 @@
 import { ThemeComponent, cva } from '@marigold/system';
 
 export const Link: ThemeComponent<'Link'> = cva(
-  'aria-[disabled]:cursor-not-allowed',
+  'text-sm aria-[disabled]:cursor-not-allowed',
```
sebald: shouldn't this be rather a size variant? otherwise ALL links are small :D

sarahgm: ahhh true i saw this also somewhere else

PR #4845, opened 2025-09-19
feat([DST-1050]): Card master and adminmark

sarahgm:
# Description

added because usage in forms, 

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/docs/content/components/content/card/card.mdx
+++ b/docs/content/components/content/card/card.mdx
@@ -1,6 +1,7 @@
 ---
 title: Card
 caption: Component to support the structuring of information.
+badge: updated
 ---
 
 A `<Card>` component is a versatile UI element used to display content in a concise, visually organized format.


themes/theme-rui/src/components/Card.styles.ts
```diff
@@ -1,17 +1,20 @@
...
-    variants: {
-      size: {
-        default: '',
-        full: 'w-full',
-      },
+export const Card: ThemeComponent<'Card'> = cva(['p-4 rounded-md'], {
+  variants: {
+    variant: {
+      default: 'util-surface-raised',
+      master: 'bg-access-master',
```
sebald: I think changing the background color will lead to a lot of issues, not only accessibility-wise.

Maybe we can play around with border/shadow colors?

https://ibelick.com/blog/create-gradient-shadows-tailwind-css

sarahgm: what kind of issues? I mean background is also for tables like this...

I can try to find a better solution...

sebald: Helptext for example. Yeah I know this is also on the tables. But I expect to be content of a Card to be much more complex than with a table.

sarahgm: Bildschirmfoto 2025-09-22 um 11 12 39
Bildschirmfoto 2025-09-22 um 11 07 16


for me this is also not soo good because i just see more lines ... I'm not sure about this

sarahgm: this is with hue set to 30% the helptext now suits with the bg color. 

Bildschirmfoto 2025-09-22 um 11 37 53

aromko: For me personally the light variants are very hard to see. Try it also on different monitors.

themes/theme-rui/src/components/Card.styles.ts
```diff
@@ -1,17 +1,21 @@
...
-      size: {
-        default: '',
-        full: 'w-full',
-      },
+export const Card: ThemeComponent<'Card'> = cva(['p-4 rounded-md'], {
+  variants: {
+    variant: {
+      default: 'util-surface-raised',
+      master: 'rounded-md bg-access-master/30 p-4 border border-access-master',
+
```
aromko: remove blank line

sebald (approved):

PR #4847, opened 2025-09-21
feat([DST-1051]): Introduce `ConfirmationDialog`

sebald:
# Description

- Create a specialzied confirmation dialog based on `<Dialog>`
- Fixed dialog props not appearing in docs' prop table
- Refactored some weird stuff that was happening in the Dialog/Modal/Underlay part

# Test Instructions:

- Made some new Stories
- Also check if I broke the dialog please

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


packages/components/src/Dialog/Dialog.stories.tsx
```diff
@@ -124,23 +125,34 @@ export const OpenFromMenu: Story = {
...
+        <Dialog open={open} onOpenChange={setDialogOpen} closeButton>
+          {({ close }) => (
+            <>
+              <Dialog.Title>Confirm delete</Dialog.Title>
+              <Dialog.Content>
+                <Text>Do you really wanna delete this?</Text>
+              </Dialog.Content>
+              <Dialog.Actions>
+                <Button slot="close">Cancel</Button>
+                <Button variant="primary" onPress={close}>
```
sarahgm: to follow the guidelines, should this also be destructive in my opinion

sarahgm: and also a confirmation

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -9,16 +9,73 @@ import type RAC from 'react-aria-components';
...
+        : props.children;
+
+    return (
+      <Dialog
+        {...props}
+        ref={ref}
+        className={cn(
+          'relative mx-auto outline-hidden',
+          "grid [grid-template-areas:'title'_'content'_'actions']",
+          className.container
```
sarahgm: why not get this from the `useClassNames` hook?

themes/theme-rui/src/components/Dialog.styles.ts
```diff
@@ -2,17 +2,25 @@ import { ThemeComponent, cva } from '@marigold/system';
...
+  container: cva(
+    [
+      'flex flex-col gap-0 rounded-xl p-6 overflow-y-auto',
+      'util-surface-overlay',
+    ],
+    {
+      variants: {
+        variant: {},
+        // Does not do anything, just to make the size appear in the appearance demo (Modal is setting the size)
+        size: {
```
sarahgm, line 14: default is missing

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -29,82 +86,37 @@ interface DialogComponent
...
-   * If `true`, the dialog will be non-modal, meaning it will not block interaction with the background content.
-   * @default false
-   */
-  isNonModal?: boolean;
-}
-
 const _Dialog = forwardRef(
   (
-    { variant, size, ...props }: DialogProps,
+    { variant, size, open, onOpenChange, children, ...props }: DialogProps,
```
sarahgm: open has no description in the props tabe

packages/components/src/Dialog/DialogTrigger.tsx
```diff
@@ -1,40 +1,46 @@
...
-  open?: boolean;
-  dismissable?: boolean;
-  keyboardDismissable?: boolean;
-  size?: string;
-}
 
 export interface DialogTriggerProps
   extends Omit<RAC.DialogTriggerProps, 'isOpen'>,
-    Omit<ModalProps, RemovedProps> {}
+    Omit<DialogContextProps, RemovedProps> {
```
sarahgm, line 18: some props like `dismissable`, or `keyboarddismissable` have no description in the props table

packages/components/src/Dialog/ConfirmationDialog.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+      >
+        Are you sure you want to do this?
+      </ConfirmationDialog>
+    </ConfirmationDialog.Trigger>
+  );
+
+  expect(screen.getByRole('button', { name: 'Abort' })).toBeInTheDocument();
+});
+
+it('calls onPrimaryAction on primary button press', async () => {
```
sarahgm: does this not make sense as Component Test in storybook?

packages/components/src/Dialog/ConfirmationDialog.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+      </ConfirmationDialog>
+    </ConfirmationDialog.Trigger>
+  );
+
+  await user.click(screen.getByRole('button', { name: 'Confirm' }));
+
+  expect(onPrimaryAction).toHaveBeenCalled();
+});
+
+it('closes on primary button press', async () => {
```
sarahgm, line 144: same for this - also a good component test i think everything with interaction or not?

packages/components/src/Dialog/Dialog.tsx
sebald: moved to inner dialog. good catch!

packages/components/src/Dialog/DialogTrigger.tsx
sebald: fixed

packages/components/src/Dialog/ConfirmationDialog.test.tsx
sebald: tried it. problem is that the dialog is rendered outside of the storybook root, which is hard to reach within a play. Adding our OverlayProvider fixes this. but then 20+ other tests break, so I didn't want to go through all of them too

themes/theme-rui/src/components/Dialog.styles.ts
sebald: there is no default in the modal :-| (was confused about this too) should I add it?

PR #4849, opened 2025-09-22
fix([DST-1053]): Add missing bg in Accordion

sarahgm:
# Description

In accordion when setting a background there is no background for the accordion itself

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/themes/theme-rui/src/components/Accordion.styles.ts
+++ b/themes/theme-rui/src/components/Accordion.styles.ts
@@ -12,7 +12,7 @@ export const Accordion: ThemeComponent<'Accordion'> = {
       variant: 'default',
     },
   }),
-  item: cva('bg-white border-b last:border-b-0 border-border ', {
+  item: cva('bg-background border-b last:border-b-0 border-border ', {
     variants: {
       variant: {
         default: '',


.changeset/hot-wolves-return.md
```diff
@@ -0,0 +1,5 @@
+---
+"@marigold/theme-rui": patch
+---
+
+fix([DST-1053]): Add missing bg in Accordion
```
aromko: That's informative. Why not writing what you did? e.g. "Added white background color to Accordion"

themes/theme-rui/src/components/Accordion.styles.ts
```diff
@@ -12,7 +12,7 @@ export const Accordion: ThemeComponent<'Accordion'> = {
       variant: 'default',
     },
   }),
-  item: cva('border-b last:border-b-0 border-border ', {
+  item: cva('bg-white border-b last:border-b-0 border-border ', {
```
sebald: Instead use `bg-background`?

sarahgm: I thought background is only for the background itself not components?

PR #4851, opened 2025-09-24
feat([DST-1056]): Add a helper  to conveniently parse form data 

sebald:
# Description

Moved the helper from the Reservix System in our repo since it might prevent people from using forms all the time.

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/packages/components/src/utils/form.utils.ts
+++ b/packages/components/src/utils/form.utils.ts
@@ -1,0 +1,31 @@
+import type { FormEvent } from 'react';
+
+/**
+ * Parses the data from a submitted HTML form and returns an object mapping form field names to their values.
+ *
+ * If a field appears multiple times (e.g., checkboxes with the same name), its value will be an array of values.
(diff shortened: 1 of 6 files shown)


packages/components/src/utils/form.utils.ts
```diff
@@ -0,0 +1,31 @@
...
+>(
+  e: FormEvent<HTMLFormElement>
+): T => {
+  const data = new FormData(e.currentTarget);
+  const result: Record<string, FormDataEntryValue | FormDataEntryValue[]> = {};
+
+  for (const [key, value] of data.entries()) {
+    if (result[key]) {
+      result[key] = Array.isArray(result[key])
+        ? [...(result[key] as FormDataEntryValue[]), value]
```
aromko, line 23: From the coverage report you didn't test this line. I think this should be tested because it's an essential logic.

sebald: fixed :) good catch!

aromko (approved):

PR #4853, opened 2025-09-25
feat([DST-1025]): Paste Date values to DatePicker

OsamaAbdellateef:
# Description

# Test Instructions:
Copy  and paste valid and invalid dates 

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


packages/components/src/DateField/DateInput.tsx
```diff
@@ -1,22 +1,122 @@
...
+    /^(\d{4})-(\d{1,2})-(\d{1,2})$/,
+    // European format: DD.MM.YYYY or DD/MM/YYYY
+    /^(\d{1,2})[./](\d{1,2})[./](\d{4})$/,
+    // US format: MM/DD/YYYY or MM-DD-YYYY
+    /^(\d{1,2})[/-](\d{1,2})[/-](\d{4})$/,
+  ];
+
+  //  ISO format (YYYY-MM-DD)
+  const isoMatch = trimmed.match(formats[0]);
+  console.log('isoMatch', isoMatch);
```
sebald: Remove the logs :)

packages/components/src/DateField/DateInput.tsx
```diff
@@ -1,22 +1,122 @@
...
+  }
+
+  return day <= daysInMonth[month - 1];
+};
+
 export interface DateInputProps extends Omit<RAC.DateInputProps, RemovedProps> {
   variant?: string;
   size?: string;
   action?: ReactElement<any>;
+  onDatePaste?: (date: CalendarDate) => void;
```
sebald: Do we need this here? Not sure if you need this in addition to `onPase`

packages/components/src/DateField/DateInput.tsx
```diff
@@ -1,22 +1,122 @@
...
+      const clipboardData = event.clipboardData.getData('text');
+
+      const parsedDate = parseDateFromString(clipboardData);
+      if (parsedDate) {
+        event.preventDefault();
+        onDatePaste(parsedDate);
+        ctx?.setValue(parsedDate);
+      }
+    } catch (error) {
+      console.warn('Failed to parse pasted date:', error);
```
sebald, line 104: Just an idea: When this throws (maybe the parsing can throw too?) briefly show an error that pasting didn't work and display an `error` with and `errorMessage` telling the user the input could not be pasted an what formats are supported.

.changeset/empty-chicken-allow.md
```diff
@@ -0,0 +1,6 @@
+---
+'@marigold/components': patch
+'@marigold/docs': patch
+---
+
+feat([DST-1025]): Adding possibilty to paste date values in `<DatePicker>` component.
```
aromko, line 6: What about `<input type="date" />`  (it's scope of the ticket)?
And should it also work for DateField? So basically for all date components we have?

packages/components/src/DatePicker/DatePicker.test.tsx
```diff
@@ -460,3 +460,68 @@ test('DatePicker supports data unavailable property', async () => {
 
   expect(date[10].firstChild).toHaveAttribute('data-unavailable', 'true');
 });
+
+test('should handle pasting a valid date string', async () => {
```
aromko: I think here a parametrized test would be suitable to test different date formats.

packages/components/src/DateField/DateInput.tsx
```diff
@@ -1,22 +1,122 @@
...
+  DateInput,
+  DatePickerStateContext,
+  Group,
+} from 'react-aria-components';
 import { useClassNames } from '@marigold/system';
 import { DateSegment } from './DateSegment';
 
 type RemovedProps = 'style' | 'className' | 'children';
 
+const parseDateFromString = (dateString: string): CalendarDate | null => {
```
aromko: There's a lot of repetitive code. Maybe you can restructure it and make it less repetitive.

packages/components/src/DateField/DateInput.tsx
```diff
@@ -1,22 +1,122 @@
...
+  const trimmed = dateString.trim();
+  //  different date formats
+  const formats = [
```
aromko: These could be constants for each format, e.g const ISO_REGEX = /^(\d{4})-(\d{1,2})-(\d{1,2})$/;

packages/components/src/DateField/DateInput.tsx
```diff
@@ -1,22 +1,122 @@
...
+    /^(\d{1,2})[./](\d{1,2})[./](\d{4})$/,
+    // US format: MM/DD/YYYY or MM-DD-YYYY
+    /^(\d{1,2})[/-](\d{1,2})[/-](\d{4})$/,
+  ];
+
+  //  ISO format (YYYY-MM-DD)
+  const isoMatch = trimmed.match(formats[0]);
+  console.log('isoMatch', isoMatch);
+  if (isoMatch) {
+    const year = parseInt(isoMatch[1], 10);
```
aromko: These lines are repetitive logic. I would create a function for these.

.changeset/empty-chicken-allow.md
sarahgm: Osama and I also had a discussion about `<input type="date" />`I would say it needs to be done in a separat ticket as its usefull for the core or what do you think?

aromko: ok for me. just asked because it was written down in the ticket.

packages/components/src/DateField/DateInput.tsx
OsamaAbdellateef: we include the onChange into onDatePaste together and pass it down, 
So when we paste the value it should trigger onChange @sebald

packages/components/src/DateField/DateInput.tsx
OsamaAbdellateef: I added a new logic and it would be really useful in this new way if it is still an array , if it is not also let me know @aromko

OsamaAbdellateef:
> Should we share this fucntionality between the `DatePicker` and `DateField`?
:heavy_check_mark:  @sebald

packages/components/src/DateField/DateInput.tsx
```diff
@@ -1,22 +1,124 @@
...
+const _DateInput = ({
+  variant,
+  size,
+  action,
+  onDatePaste,
+  ...props
+}: DateInputProps) => {
   const classNames = useClassNames({ component: 'DateField', variant, size });
 
+  const dateFieldContext = useContext(DateFieldStateContext);
```
sebald: Move these to the corresponding components? The `parsed` date is passed to them anyway. So they can take care of the `setValue`.

OsamaAbdellateef: if we do it this way in date picker or field we won't be able to access the context because it needs to be wrapped with either DateFieldStateContext or DatePickerStateContext

  ` const datePickerContext = useContext(DatePickerStateContext);
    
    const handleDatePaste = (date: CalendarDate) => {
      datePickerContext?.setValue(date);
      
      if (onChange) {
        onChange(date);
      }
    };`

OsamaAbdellateef: Also would you recommend to pass down Context and then it is consumed inside the DateInput 
`const ctx = useContext(PassedContext)`

sebald: We could pass down an update function so the context is still scoped

sebald (approved):

PR #4854, opened 2025-09-25
feat([DST-1061]): Add more alignment options to `<Inline>`, `<Stack>` and `<Grid>`

sebald:
# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


docs/content/components/layout/inline/inline-horizontally.demo.tsx
```diff
@@ -27,5 +27,29 @@ export default () => (
       </Rectangle>
       <Rectangle height="60px" width="60px" />
     </Inline>
+    <Divider />
+    <Inline space={3} alignX="between">
+      <Rectangle height="60px" width="60px" />
+      <Rectangle height="60px" width="60px">
+        <div className="text-text-primary-muted m-auto">Between</div>
```
sarahgm, line 34: doesn't fit in the rectangle needs to be wider

packages/components/src/Inline/Inline.stories.tsx
```diff
@@ -39,7 +40,7 @@ const meta = {
         type: { summary: 'text' },
         defaultValue: { summary: 'undefined' },
       },
-      options: ['top', 'center', 'bottom'],
+      options: [...Object.keys(alignment.vertical.alignmentY), 'input'].flat(),
```
sarahgm: here is also `evenly`, `around` etc applied but this has no effect. There is no `align-items:evenly`

sarahgm: you need to use this: horizontal.alignmentY

packages/components/src/Stack/Stack.stories.tsx
```diff
@@ -18,14 +19,14 @@ const meta = {
       control: {
         type: 'select',
       },
-      options: ['none', 'left', 'center', 'right'],
+      options: Object.keys(alignment.horizontal.alignmentX),
```
sarahgm: here you need to use this `alignment.vertical.alignmentX`

packages/components/src/Stack/Stack.stories.tsx
```diff
@@ -18,14 +19,14 @@ const meta = {
...
       description: 'Horizontal Alignment',
     },
     alignY: {
       control: {
         type: 'select',
       },
-      options: ['none', 'top', 'center', 'bottom'],
+      options: Object.keys(alignment.vertical.alignmentY),
```
sarahgm, line 29: this has no effect, no option is working, did you use the correct tokens?

sarahgm (approved):

PR #4856, opened 2025-09-26
feat([DST-1069]): Add `noWrap` prop to `<Inline>`

sebald:
# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


packages/components/src/Inline/Inline.stories.tsx
```diff
@@ -43,6 +43,9 @@ const meta = {
       description: 'Set the Vertical Alignment',
     },
   },
+  args: {
+    noWrap: undefined,
```
sarahgm, line 47: why do you need undefined here? We can set default to false

sebald: So it looks better in the story :D

sarahgm: In the story it are two clicks then, you wrote in the docs that the default is anyway wrap so I find this unnecessary

PR #4857, opened 2025-09-26
docs: Release Notes 15.3.0.

sebald:


aromko:
@marigold-ui/developer fyi I adjusted the badges to the release notes. So if you add a new component or something else make sure that the site has the correct badge.

PR #4858, opened 2025-09-27
feat([DST-1074]): Introduce a `useConfirmation` hook to conveniently open confirmation dialogs

sebald:
# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/packages/components/src/Dialog/useConfirmation.test.tsx
+++ b/packages/components/src/Dialog/useConfirmation.test.tsx
@@ -1,0 +1,73 @@
+import { composeStories } from '@storybook/react';
+import { render, renderHook, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import * as stories from './ConfirmationDialog.stories';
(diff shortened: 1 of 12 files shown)


packages/components/src/Dialog/useConfirmation.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+test('throw if ConfirmationProvider is missing', () => {
+  expect(() =>
+    renderHook(() => useConfirmation())
+  ).toThrowErrorMatchingInlineSnapshot(
+    `[Error: \`useConfirmation\` must be used within a \`ConfirmationProvider\`]`
+  );
+});
+
+test('open confirmation dialog', async () => {
```
aromko, line 18: Why not as behavior test in the story?

sebald: because of this :D https://github.com/marigold-ui/marigold/pull/4864

If fixed this in this PR but there were so many errors then. Needs to be done separately.

PR #4863, opened 2025-09-29
feat([DST-1077]): Add white space control to `<Text>`

sebald:
# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/packages/components/src/Text/Text.test.tsx
+++ b/packages/components/src/Text/Text.test.tsx
@@ -128,24 +128,15 @@ test('supports wrap prop', () => {
   expect(pretty).toHaveClass('text-pretty');
 });
 
-test('supports whiteSpace prop', () => {
-  render(
-    <>
-      <Basic whiteSpace="normal">normal</Basic>
-      <Basic whiteSpace="pre">pre</Basic>
-      <Basic whiteSpace="preLine">preLine</Basic>
-      <Basic whiteSpace="preWrap">preWrap</Basic>
-    </>
-  );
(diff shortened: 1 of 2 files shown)


packages/components/src/Text/Text.test.tsx
```diff
@@ -127,3 +127,25 @@ test('supports wrap prop', () => {
   expect(balance).toHaveClass('text-balance');
   expect(pretty).toHaveClass('text-pretty');
 });
+
+test('supports whiteSpace prop', () => {
+  render(
+    <>
+      <Basic whiteSpace="normal">normal</Basic>
```
aromko: nowrap is missing? And maybe as parametrized test so you can avoid to have the multiple elements with pre on the page.

packages/system/src/style-props.tsx
```diff
@@ -254,6 +254,14 @@ export const textWrap = {
   pretty: 'text-pretty',
 };
 
+export const whiteSpace = {
+  normal: 'whitespace-normal',
+  nowrap: 'whitespace-nowrap',
+  pre: 'whitespace-pre',
+  preLine: 'whitespace-pre-line',
+  preWrap: 'whitespace-pre-wrap',
```
aromko, line 262: whitespace-break-spaces is missing? You used it in the test above.

packages/components/src/Text/Text.test.tsx
sebald: Fixed!

PR #4864, opened 2025-09-29
test: make it possible to write interaction tests for overlay components

sebald:
# Description

Test helpers only look for elements in side the storybook root, but our overlays are appenended to the body. Since stories are rendered in an iframe the `screen` helpers will not find the elements -> use the `OverlayProvider` to render the overlays in side the storybook iframe.

# Test Instructions:

Checkout the play test of the `<Dialog>`

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/packages/components/src/Dialog/Dialog.stories.tsx
+++ b/packages/components/src/Dialog/Dialog.stories.tsx
@@ -1,5 +1,6 @@
 import type { StoryObj } from '@storybook/react';
 import { useState } from 'storybook/preview-api';
+import { expect, waitFor } from 'storybook/test';
 import { Button } from '../Button';
 import { Menu } from '../Menu';
 import { Text } from '../Text';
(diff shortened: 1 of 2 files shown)


sebald:
@aromko ah draft. Need to fix all the things that broke :D

packages/components/src/Dialog/Dialog.stories.tsx
```diff
@@ -71,6 +71,17 @@ export const Basic: Story = {
       </Dialog>
     </Dialog.Trigger>
   ),
+  play: async ({ canvas, userEvent }) => {
+    // Open the dialog
```
aromko: I think the comments are not helpful here.

aromko (commented):
Maybe also run VRT jsut to be sure nothing changed.

packages/components/src/Dialog/Dialog.stories.tsx
sebald: removed :)

PR #4865, opened 2025-09-29
feat(DST-810): responsive-dialog-height

marekgieldatsh:
# Description

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


packages/components/src/Dialog/Dialog.stories.tsx
```diff
@@ -28,55 +34,52 @@ const meta = {
...
-          </Button>
-          <Button variant="primary" slot="close">
-            Enable
-          </Button>
-        </Dialog.Actions>
-      </Dialog>
-    </Dialog.Trigger>
-  ),
+  render: args => {
+    const { size, ...triggerArgs } = args;
```
aromko: Why this extra line?

packages/components/src/Dialog/Dialog.stories.tsx
```diff
@@ -28,55 +34,52 @@ const meta = {
       },
       description: 'Size of the dialog',
       options: ['default', 'xsmall', 'small', 'medium'],
```
aromko, line 37: Do we need default anymore?

packages/components/src/Dialog/Dialog.test.tsx
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable testing-library/no-node-access */
```
aromko, line 1: @marekgieldatsh If you are motivated you could do the tests refactoring for this component. Some months ago we decided to refactor the tests according to https://reservix.atlassian.net/wiki/spaces/DST/pages/3626336415/Refactoring+writing+tests+in+storybook
 If you have question just contact me.

packages/components/src/Dialog/Dialog.stories.tsx
marekgieldatsh: It separates the size prop for the Dialog component from the other props that go to Dialog.Trigger

packages/components/src/Dialog/Dialog.stories.tsx
marekgieldatsh: No, I'll remove it

packages/components/src/Overlay/Underlay.tsx
```diff
@@ -37,7 +37,7 @@ export const Underlay = ({
     <ModalOverlay
       className={({ isEntering, isExiting }) =>
         cn(
-          'fixed inset-0 z-40 grid place-items-center overflow-y-auto',
+          'fixed inset-0 z-40 flex flex-col items-center justify-center overflow-hidden p-4',
```
sarahgm, line 40: if we remove `grid` the `grid-areas` doesn't have any effect anymore or am I wrong?

packages/components/src/Dialog/Dialog.stories.tsx
aromko: yeah ok my bad. I mean why you put it into an extra line? Before it was `({ size, ...args })` which does the same.

packages/components/src/Dialog/Dialog.test.tsx
```diff
@@ -1,36 +1,24 @@
...
 import { MockInstance, vi } from 'vitest';
 import { Theme, cva } from '@marigold/system';
 import { Button } from '../Button';
 import { setup } from '../test.utils';
 import { Dialog } from './Dialog';
+import * as stories from './Dialog.stories';
+
+const { Basic, VeryLongContent } = composeStories(stories);
 
 const theme: Theme = {
```
aromko, line 13: The theme object you shouldn't need anymore because using the stories the styles are already applied.

packages/components/src/Dialog/Dialog.test.tsx
```diff
@@ -42,8 +30,7 @@ const theme: Theme = {
   },
 };
 
-const user = userEvent.setup();
-const { render } = setup({ theme });
+const { render: renderWithTheme } = setup({ theme });
```
aromko: Also this shouldn't be needed.

packages/components/src/Dialog/Dialog.test.tsx
```diff
@@ -56,194 +43,50 @@ afterEach(() => {
...
-  expect(headline).toBeInTheDocument();
-
-  const dialog = screen.getByText('Content');
-  expect(dialog).toBeInTheDocument();
 });
 
 test('supports children as function', () => {
   const spy = vi.fn().mockReturnValue(<div>I am a spy!</div>);
-  render(
+  renderWithTheme(
```
aromko: For example here you can add a story and use this story here then.

packages/components/src/Dialog/Dialog.test.tsx
```diff
@@ -56,194 +43,50 @@ afterEach(() => {
...
 
-  expect(spy).toHaveBeenCalled();
+  expect(spy).not.toHaveBeenCalled();
 });
 
-test('dialog can be opened by button', () => {
-  render(
+test('supports title for accessibility reasons', async () => {
+  const user = userEvent.setup();
+  renderWithTheme(
```
aromko: If we have interaction tests like this one we decided to move these to the stories s.https://storybook.js.org/docs/writing-tests/interaction-testing

packages/components/src/Dialog/Dialog.test.tsx
```diff
@@ -56,194 +43,50 @@ afterEach(() => {
...
-    </Dialog.Trigger>
-  );
-  const button = screen.getByText('Open');
-  fireEvent.click(button);
-
-  const dialog = screen.getByText('Content');
-  expect(dialog).toBeVisible();
-});
-
-test('optionally renders a close button', () => {
```
aromko, line 111: I think you removed too much. It's good to test if a closeButton appears when expected. Couldn't find this sort of test now.

themes/theme-rui/src/components/Modal.styles.ts
```diff
@@ -2,9 +2,10 @@ import { ThemeComponent, cva } from '@marigold/system';
 
 export const Modal: ThemeComponent<'Modal'> = cva(
   [
-    'sm:max-h-[min(640px,80vh)]',
     '[--dialog-spacing-x:1rem]', // Minimal padding to the horizontal edges
-    'w-[min(calc(100%_-_var(--dialog-spacing-x)),calc(var(--dialog-width)_-_var(--dialog-spacing-x)))]',
+    'max-h-[calc(100dvh-2rem)] h-[calc(100dvh-2rem)] w-full my-4', // Mobile: Full height and width with margin
```
sebald: `react-aria` calculates a `--visual-viewport-height` css custom prop. But not sure if this will end up having the same effect as `dvh`.

EDIT: just saw that we have inset etc on the underlay. I think this is where the `--visual-viewport-height` should be set (see: https://react-spectrum.adobe.com/react-aria/Modal.html#modaloverlay). This might also cause some issues

packages/components/src/Dialog/Dialog.stories.tsx
```diff
@@ -27,56 +33,55 @@ const meta = {
...
+        </Dialog>
+      </Dialog.Trigger>
+    );
+  },
 };
 
 export const Form: Story = {
-  render: ({ size, ...args }) => {
+  render: args => {
+    const { size, ...triggerArgs } = args;
```
sebald: Please destructure diretctly in the function :)

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -51,8 +51,8 @@ const InnerDialog = forwardRef(
         {...props}
         ref={ref}
         className={cn(
-          'relative mx-auto outline-hidden',
-          "grid [grid-template-areas:'title'_'content'_'actions']",
+          'relative mx-auto max-h-full min-h-0 w-full max-w-lg outline-hidden',
```
sebald: Setting the width here, causes the `mx-auto` to overlap the `Underlay` which in result creates an areay where you click the margin and not the underlay, so the dialog is not closing as expected (because it looks like you are clicking on the underlay).

Hope this image explains it:
Image

Orange area = margin created my `mx-auto` and `w-*`.

In the current implementation the dialog takes the width it is given by the modal.

packages/components/src/Dialog/Dialog.stories.tsx
marekgieldatsh: Fixed

packages/components/src/Dialog/Dialog.test.tsx
marekgieldatsh: I hope I will in the future. Currently, I use an interaction test for the scroll behavior of the Dialog component (Long content). @aromko  Please check if it is correct

packages/components/src/Dialog/Dialog.test.tsx
marekgieldatsh: Done

packages/components/src/Dialog/Dialog.test.tsx
marekgieldatsh: Not valid anymore, there is no refactor in this commit

packages/components/src/Dialog/Dialog.stories.tsx
marekgieldatsh: done

packages/components/src/Dialog/Dialog.tsx
marekgieldatsh: It should work now

packages/components/src/Dialog/Dialog.test.tsx
aromko: @marekgieldatsh the scrolling story looks good to me 👍

packages/components/src/Dialog/Dialog.stories.tsx
```diff
@@ -159,3 +159,175 @@ export const Confirmation: StoryObj<ConfirmationDialogProps> = {
...
+        </Dialog>
+      </Dialog.Trigger>
+    );
+  },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+
+    // Open the dialog
+    await userEvent.click(canvas.getByText('Open Dialog with Long Content'));
+    await new Promise(resolve => setTimeout(resolve, 100));
```
sebald: Does `waitFor` work here too?

marekgieldatsh: Yes, it does

themes/theme-rui/src/components/Modal.styles.ts
marekgieldatsh: The part of the code does not exist anymore

packages/components/src/Dialog/Dialog.stories.tsx
```diff
@@ -24,27 +35,20 @@ const meta = {
       },
       description: 'Size of the dialog',
       options: ['default', 'xsmall', 'small', 'medium'],
```
sebald, line 37:
```suggestion
      options: ['xsmall', 'small', 'medium'],
```

sebald: `default` makes it so width style is applied.

PR #4866, opened 2025-09-30
feat([DSTSUP-185]): add sticky header to accordion

aromko:
# Description

Added stickyHeader and iconPosition prop.

# Test Instructions:

Test the accordion in storybook

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/.changeset/stupid-snakes-give.md
+++ b/.changeset/stupid-snakes-give.md
@@ -1,0 +1,6 @@
+---
+"@marigold/docs": patch
+"@marigold/components": patch
+---
+
+feat([DSTSUP-185]): added `stickyHeader`and `iconPosition`props to `Accordion`


packages/components/src/Accordion/Accordion.tsx
```diff
@@ -13,19 +13,23 @@ export interface AccordionProps
   disabled?: RAC.DisclosureGroupProps['isDisabled'];
   variant?: 'default' | 'card' | (string & {});
   size?: string;
+  isStickyHeader?: boolean;
```
sebald: no `is`, table prop is called `stickyHeader`

packages/components/src/Accordion/AccordionHeader.tsx
```diff
@@ -9,21 +9,31 @@ export interface AccordionHeaderProps {
...
-    <Heading>
-      <Button slot="trigger" className={classNames.header}>
-        <div className="flex-1">{children}</div>
-        <ChevronDown
-          className={cn(classNames.icon, isExpanded && 'rotate-180')}
-        />
-      </Button>
-    </Heading>
+    <div
+      className={cn(
```
sebald: does not need the `cn` function, I think. If you return `undefined` the `className` prop will not appear on the rendered HTML element

PR #4881, opened 2025-10-10
feat([DST-1091]): Add multiselection mode to `<Select>`

sebald:
# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/.changeset/purple-chairs-yell.md
+++ b/.changeset/purple-chairs-yell.md
@@ -1,0 +1,8 @@
+---
+"@marigold/docs": minor
+"@marigold/components": minor
+"@marigold/types": minor
+"@marigold/theme-rui": minor
+---
+
+feat([DST-1091]): Add multiselection mode to `<Select>`


packages/components/src/Select/Select.tsx
```diff
@@ -83,73 +73,59 @@ export interface SelectProps<T extends object>
...
+    isRequired: required,
+    ...rest,
+  };
+  const classNames = useClassNames({ component: 'Select', variant, size });
 
-_Select.Option = ListBox.Item;
-_Select.Section = ListBox.Section;
+  return (
+    <FieldBase
+      as={ReactAraiSelect}
```
aromko: typo

packages/components/src/Select/Select.stories.tsx
```diff
@@ -75,18 +77,19 @@ const meta = {
...
+    width: 64,
   },
 } satisfies Meta<typeof Select>;
 
 export default meta;
 
 export const Basic: StoryObj<typeof Select> = {
   render: args => {
-    const [selected, setSelected] = useState<string | number>('');
+    const [selected, setSelected] = useState<any>('');
```
aromko, line 92: "Any" is always not good.

packages/components/src/Select/Select.tsx
sebald: fixed

packages/components/src/Select/Select.stories.tsx
sebald: No, but in this case necessary, because `args` is breaking the types because the Storybook types merge the types for both `selectionMode`s.

PR #4885, opened 2025-10-14
feat([DST-725]): Switch to Lucide icons

grzegorz-sajdok:
# Description
Refactoring the icons system in components package to use Lucide icons in the icons folder

# Test Instructions:
A storybook example comparing the old icons to new icons is provided under the Icons section.

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


packages/components/src/icons/Accessible.tsx
```diff
@@ -1,13 +0,0 @@
-import { forwardRef } from 'react';
-import { SVG, SVGProps as SVGPropsType } from '@marigold/system';
-
-interface SVGProps extends SVGPropsType {
-  className?: string;
-}
-
-export const Accessible = forwardRef<SVGSVGElement, SVGProps>((props, ref) => (
```
sarahgm, line 8: This will be a breaking change when we change the name, can we use the same name as before?

sarahgm: ah I see its not the one from icon package :D I will close this

sarahgm (approved):

PR #4890, opened 2025-10-16
refa([DST-1099]): Improve icon visibility

sebald:
# Description

- Adjusted some icons to fit their components better.
- Some other small improvements it stumpled upon while adjusting the icons.

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/packages/components/src/Radio/Radio.test.tsx
+++ b/packages/components/src/Radio/Radio.test.tsx
@@ -1,4 +1,4 @@
-import { act, fireEvent, screen, waitFor } from '@testing-library/react';
+import { screen } from '@testing-library/react';
 import React from 'react';
 import { Theme, ThemeProvider, cva } from '@marigold/system';
 import { setup } from '../test.utils';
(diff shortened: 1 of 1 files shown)


packages/components/src/Pagination/Pagination.test.tsx
```diff
@@ -23,101 +23,101 @@ describe('Pagination tests', () => {
...
-            xmlns="http://www.w3.org/2000/svg"
-          >
-            <path
-              d="m15 18-6-6 6-6"
-            />
-          </svg>
-        </button>
         <div
-          class="flex items-center justify-center space-x-2"
+          class="bg-bg-surface"
```
sarahgm, line 27: `bg-bg-surface` where is this coming from, we don't have a token like this anymore its `bg-surface-body`

packages/components/src/Radio/Radio.test.tsx
```diff
@@ -75,9 +75,7 @@ test('allows styling via theme', () => {
   expect(radioLabel.className).toMatchInlineSnapshot(`"text-base"`);
 
   const radio = getVisibleRadios()?.[0];
-  expect(radio?.className).toMatchInlineSnapshot(
-    `"bg-secondary-50 flex h-4 w-4 items-center justify-center border p-1 cursor-pointer rounded-xs border-solid checked:text-blue-700"`
-  );
+  expect(radio?.className).toMatchInlineSnapshot(`"hidden"`);
```
sarahgm: think the test outcome is wrong or? Should it really be `hidden` ?

packages/components/src/Pagination/Pagination.test.tsx
sebald: this is coming from the pagination.styles :D

packages/components/src/Radio/Radio.test.tsx
sebald: I think we can remove the whole test.

PR #4894, opened 2025-10-17
[DSTSUP-191]: Make XLoader more flexible and generic [BC]

aromko:
# Description

Add a new attribute `loaderType` which can be xloader (default) or cycle.

Reused ProgressCycle.

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/packages/components/src/Loader/Loader.test.tsx
+++ b/packages/components/src/Loader/Loader.test.tsx
@@ -32,14 +32,6 @@ test('render custom label', () => {
   expect(label).toBeInTheDocument();
 });
 
-test('fullsize uses "inverted" variant', () => {
-  render(<Basic mode="fullscreen">Loading...</Basic>);
-
-  const loader = screen.getByRole('progressbar');
-
-  expect(loader).toHaveClass('grid place-items-center text-secondary size-20');
-});
-
 test('inline uses "inverted" variant', () => {
   render(<Basic mode="section">Loading...</Basic>);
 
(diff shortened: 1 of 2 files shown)


packages/components/src/Loader/BaseLoader.tsx
```diff
@@ -47,157 +206,11 @@ export const BaseLoader = ({
...
-            attributeName="opacity"
-            attributeType="XML"
-            values="1; .01; 1; 1; 1;"
-            begin="0.0s"
-            dur="2.5s"
-            repeatCount="indefinite"
-          />
-        </path>
-      </svg>
+      {loaderType === 'xloader' ? (
```
sebald, line 209: I would rather have two separate loaders than cram everything into one component.

If we want to all loaders to have the same options, I think modularizing `<BaseLoader>` would be better.

aromko: @sebald Hm what is the benefit? I mean basically it just uses another SVG component based on the loaderType 🤔

sebald: "XLoader" is a very speicifc name though. `<XLoader loaderType="xloader">` feels redudant.

aromko: @sebald Now I'm confused - so the naming of the loaderType is the problem? Can call it "default" or "x". We don't want two separate loaders or like XLoader and CycleLoader?

.changeset/unlucky-fishes-matter.md
```diff
@@ -0,0 +1,9 @@
+---
+"@marigold/components": patch
```
sebald: major! :D

PR #4902, opened 2025-10-23
[Infra] bump minor and patch versions for dependencies

aromko:
# Description

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/packages/components/src/Button/Button.stories.tsx
+++ b/packages/components/src/Button/Button.stories.tsx
@@ -1,6 +1,6 @@
 import type { Meta, StoryObj } from '@storybook/react';
 import { useState } from 'storybook/preview-api';
-import { expect, fn, userEvent, within } from 'storybook/test';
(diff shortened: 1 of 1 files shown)


packages/components/src/Slider/Slider.stories.tsx
```diff
@@ -99,7 +99,9 @@ export const ValueFormatting: Story = {
       </Stack>
     </I18nProvider>
   ),
-  play: async ({ canvas }) => {
```
sebald, line 102: Is this a issue with the types or did they randomly change their API? I do not think it is a good idea to use `canvasElement` over `canvas`, especially because it requires the extra step with `within`.

aromko: Couldn't find out.

aromko: @sebald Found it and undo it 👍

PR #4906, opened 2025-10-23
[Infra]: upgrade Vitest and related dependencies to v4

aromko:
# Description

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


vitest.config.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-ignore
```
sebald: Is this currently necessary? If so, rather use `@ts-expect-error`

aromko: @sebald seems not to work with ts-expect error I get 
`vitest.config.ts:1:1 - error TS2578: Unused '@ts-expect-error' directive.

1 //@ts-expect-error TS2307`

From their docs this means "// @ts-expect-error wasn’t necessary."

sebald (approved):

PR #4907, opened 2025-10-23
[Infra]: NextJs16 upgrade

aromko:
# Description

- a lot of ts fixes necessary
- turbopack is now the default and we use it now. Contentlayer2 uses webpack which leads trouble with turbopack. I've splitted the tasks and let run them in parallel

# Test Instructions:

- test the docs
- do changes and check for HMR

# Reviewers:

Focus in review on package.jsons, config files

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,8 +9,8 @@ importers:
   .:
     dependencies:
       '@babel/core':
-        specifier: 7.28.4
-        version: 7.28.4
+        specifier: 7.28.5
+        version: 7.28.5
       '@changesets/changelog-git':
         specifier: 0.2.1
         version: 0.2.1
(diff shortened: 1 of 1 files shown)


docs/package.json
```diff
@@ -30,12 +32,13 @@
...
     "dotenv": "17.2.3",
     "eslint": "9.34.0",
-    "eslint-config-next": "15.5.2",
+    "eslint-config-next": "16.0.0",
     "framer-motion": "12.23.22",
     "github-slugger": "2.0.0",
     "lucide-react": "0.542.0",
     "mdast-util-toc": "7.1.0",
-    "next": "15.5.2",
+    "mdx-bundler": "10.1.1",
```
sebald, line 40: Why do we have both? `mdx-bundler` and `contentlayer` :-O

docs/next.config.js
```diff
@@ -1,13 +1,9 @@
-const { withContentlayer } = require('next-contentlayer2');
```
sebald, line 1: Are we not using `contentlayer` anymore?

sebald:
@aromko if contentlayer is an issue, shouldn't we migrate away from contentlayer then?

aromko:
> @aromko if contentlayer is an issue, shouldn't we migrate away from contentlayer then?

@sebald Is it an issue? They don't support turbopack. But I can write a research ticket for this. I think we do not need to replace it in this pr.

sebald (approved):

PR #4914, opened 2025-10-27
Enable Next.js Typed routes

max-programming:
# Description
This PR enables Next.js typed routes for the marigold ui docs. Now there is better intellisense for href in `Link` component and `router.push` and `router.replace` methods

image

# Test Instructions:
Go to usage of any `Link` component and pass in a non-existing `href`, it will throw a typeerror

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


sebald:
Thanks for the PR but `as Route` casting is certainly not the correct approach.

PR #4917, opened 2025-10-31
style([DST-1089]): Add expand/collapse animation to `<Accordion>`

sebald:
> [!CAUTION]
> This is a breaking change since the styling slightly changed. The styles for the `content` need to be on the `panel` now. `content` is still there and is the child of the panel. This is a save place to add additional padding or anything else that add height to the element.

# Description

- Add animation when accordion pannels are expaned or collapse for better usability and affordance

In order to make this work I had to change the structure of the `<Accordion.Panel>` slidly (padding needs to be on the child). Used this to also fix the naming (the panel was called content in the styles).

I forgot that add `ui` but I manually checked while doing the animation that it looks exactly like before.

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

- [x Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [] Updated visual regression tests (only necessary when ui changes in the PR)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "engines": {
     "node": "22.x"
   },
-  "packageManager": "pnpm@10.15.1",
+  "packageManager": "pnpm@10.19.0",
   "dependencies": {
     "@babel/core": "7.28.5",
     "@changesets/changelog-git": "0.2.1",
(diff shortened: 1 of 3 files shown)


themes/theme-rui/src/components/Accordion.styles.ts
```diff
@@ -46,8 +46,11 @@ export const Accordion: ThemeComponent<'Accordion'> = {
       },
     }
   ),
-  content: cva('in-data-[expanded]:pb-2'),
```
sarahgm, line 49: why is this class removed?

sebald: It is not removed. It is still there. (line 52)

sarahgm: `in-data-[expanded]`

sarahgm: but i don't see a difference there so i guess its fine

aromko (approved):

PR #4922, opened 2025-11-03
refa(DST-1109): Remove required indicator from the label's text content

sebald:
> [!CAUTION]
> This is a breaking change since we removed the `indicator` part from `<Label>`. It is no longer multi part component.

# Description

In order to make it easier to use helpers from testing-library, we remove the "*" indicator from the labels text content by applying it via CSS (pseudo element + `content`).

While this make testing easier, this is a breaking change regarding theming. The `indicator` part is no longer needed.

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


packages/components/src/Label/Label.tsx
```diff
@@ -9,14 +9,11 @@ export interface LabelProps extends Omit<RAC.LabelProps, RemovedProps> {
 }
 
 const _Label = ({ size, variant, children, ...props }: LabelProps) => {
-  const classNames = useClassNames({ component: 'Label', size, variant });
+  const className = useClassNames({ component: 'Label', size, variant });
```
aromko, line 12: Why singular?

sebald: Not an object anymore. Has only one className prop

PR #5256, opened 2026-04-01
fix(DST-1294): remove muted text color from Dialog content slot

sebald:
## Summary
- Removes `text-muted-foreground` from the Dialog `content` slot in `theme-rui`
- Dialog body text now inherits the default foreground color instead of being muted
- Fixes readability issues and removes the need for consumer overrides

## Jira
[DST-1294](https://reservix.atlassian.net/browse/DST-1294)

## Test plan
- [ ] Verify Dialog content text uses default foreground color
- [ ] Check existing Dialog stories in Storybook for visual regression

[DST-1294]: https://reservix.atlassian.net/browse/DST-1294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


themes/theme-rui/src/components/Dialog.styles.ts
```diff
@@ -19,7 +19,7 @@ export const Dialog: ThemeComponent<'Dialog'> = {
   }),
   header: cva({ base: 'flex flex-col gap-1 text-center sm:text-left' }),
   title: cva({ base: 'text-lg font-semibold mb-1' }),
-  content: cva({ base: 'text-muted-foreground text-sm' }),
+  content: cva({ base: 'text-sm' }),
```
aromko, line 22: VRT missing?

aromko (dismissed):

themes/theme-rui/src/components/Dialog.styles.ts
```diff
@@ -19,7 +19,7 @@ export const Dialog: ThemeComponent<'Dialog'> = {
   }),
   header: cva({ base: 'flex flex-col gap-1 text-center sm:text-left' }),
   title: cva({ base: 'text-lg font-semibold mb-1' }),
-  content: cva({ base: 'text-muted-foreground text-sm' }),
```
aromko, line 22: There was no change https://www.chromatic.com/build?appId=68679799bacf3b2432835044&number=587 .

Maybe we need another story here?

sebald: added a story

aromko (approved):

packages/components/src/Dialog/Dialog.stories.tsx
```diff
@@ -204,6 +204,21 @@ export const Confirmation = meta.story({
   ),
 });
 
+export const TextOnly = meta.story({
```
aromko, line 207: Just saw it afterwards but the dialog should be open. otherwise the snapshot is useless or?

sebald: ah true. will fix

PR #5277, opened 2026-04-07
🧹 Reduce ESLint warnings: clean up createRef, useContext, and component-hook-factories

OsamaAbdellateef:
# Description

Reduced the lint warnings by 40% .

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


packages/components/src/Button/Button.test.tsx
```diff
@@ -22,7 +22,7 @@ test('add icon in button works as expected', () => {
 });
 
 test('forwards ref', () => {
-  const ref = createRef<HTMLButtonElement>();
+  const ref = { current: null } as unknown as RefObject<HTMLButtonElement>;
```
aromko: @OsamaAbdellateef I think this kind of a bad hack 😅 With this all typechecking is bypassed.

I think for the test you can do something like: 

```
 let refValue: HTMLButtonElement | null = null;       
                                                                                                                     
  render(<Basic.Component ref={node => { refValue = node; }} />);                                                
                                                           
  expect(refValue).toBeInstanceOf(HTMLButtonElement);     
```

packages/icons/src/Icon.test.tsx
```diff
@@ -27,8 +27,8 @@ test('supports className', () => {
 });
 
 test('forwards ref', () => {
-  const ref = createRef<SVGSVGElement>();
-  render(<ArrowUp ref={ref} />);
+  const ref = { current: null } as unknown as RefObject<SVGSVGElement>;
+  render(<ArrowUp />);
```
aromko: Here the ref is missing so it isn't the same test like before.

packages/components/src/Multiselect/Multiselect.tsx
```diff
@@ -319,7 +319,7 @@ export const Multiselect = ({
           components={{
             Input,
             MultiValueRemove,
-            // eslint-disable-next-line @eslint-react/no-nested-component-definitions
+            // eslint-disable-next-line @eslint-react/no-nested-component-definitions, @eslint-react/component-hook-factories
```
aromko: This isn't needed.

packages/components/src/Button/Button.test.tsx
```diff
@@ -13,10 +12,16 @@ test('add icon in button works as expected', () => {
 });
 
 test('forwards ref', () => {
-  const ref = createRef<HTMLButtonElement>();
-  render(<Basic.Component ref={ref} />);
-
-  expect(ref.current).toBeInstanceOf(HTMLButtonElement);
+  let refValue: HTMLButtonElement | null = null;
```
aromko, line 15: @OsamaAbdellateef You should did this also for the other places. Sry I was too lazy to write for every file a comment. I thought this is clear 😅 

This `{ current: null } as unknown as RefObject` shouldn't be anywhere.

aromko (approved):

PR #5279, opened 2026-04-08
feat(DST-1239): migrate Inset to semantic spacing tokens

sebald:
## Summary

Refactors `Inset` to use semantic spacing tokens with a clear separation between all-sides padding and per-axis padding:

- **`space` prop**: Accepts `PaddingSpacingTokens` (`square-*`, `squish-*`, `stretch-*`) and numeric scale values. These are multi-value padding recipes that set all sides at once.
- **`spaceX`/`spaceY` props**: Accept `InsetSpacingTokens` (`inset-tight`, `inset-snug`, `inset-regular`, `inset-relaxed`, `inset-loose`) and numeric scale values. These are single-value tokens for composing asymmetric padding.
- Adds `--spacing-inset-*` CSS custom properties to the theme
- Preserves the discriminated union prop pattern (either `space` or `spaceX`/`spaceY`, not both)

Resolves [DST-1239](https://reservix.atlassian.net/browse/DST-1239)

## Test plan

- [x] TypeScript type checking passes
- [x] All unit tests pass
- [ ] Verify Inset story renders correctly in Storybook with padding tokens
- [ ] Verify `spaceX`/`spaceY` with inset tokens produce correct per-axis padding

[DST-1239]: https://reservix.atlassian.net/browse/DST-1239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/Inset/Inset.tsx
```diff
@@ -20,34 +15,39 @@ export type InsetProps =
       children: ReactNode;
       space?: never;
       /**
-       * Horizontal alignment for the children. You can see allowed tokens [here](../../foundations/design-tokens?theme=core#spacing).
-       * @remarks `PaddingSpacePropX<Tokens>`
+       * Horizontal padding for the children. You can see allowed tokens [here](../../foundations/design-tokens?theme=core#spacing).
+       * @remarks `Scale | SpacingTokens`
        */
-      spaceX?: PaddingSpacePropX['spaceX'];
+      spaceX?: SpaceProp<SpacingTokens>['space'];
```
sarahgm: hmm for me it doesn't make semantically sense to have the spaceing tokens like "Section, or group" as horizontally or vertical padding 🤔

sebald: Maybe. Just didn't want to be that restrictive at the start :D

Our docs currently has no real opinion on that, I think

packages/components/src/Inline/Inline.stories.tsx
```diff
@@ -14,8 +14,9 @@ const meta = preview.meta({
   argTypes: {
     space: {
       control: {
-        type: 'text',
+        type: 'select',
```
sarahgm, line 17: do we want that we now not can try out numbers?

packages/components/src/Inset/Inset.tsx
sarahgm: I see but I think it doesn't make sense to have it, when would you use group padding horizontally on the element? For what reason?

sebald: I think if we want to have an opionion here, we should do this in another PR maybe?

packages/components/src/Inline/Inline.stories.tsx
sebald: Did this to basically promoted the spacing tokens. IMHO the numbers should only be a fallback/expection?

packages/components/src/Inset/Inset.tsx
sarahgm: What exactly? make thought of X/Y Padding should use the Spacing Tokens?

sebald: to have separate types for vertical and horizontal padding and to only allow a certain set of tokens for horizontal padding.

sarahgm: ok for me, but still think we should not use them here, this is from the token.ts file: `[...] where naming reflects the cognitive connection between objects [...]` sound to me as the spacing tokens are especially for between elements when you have an inset you can not have between elements

sebald: alright, then lets update this in this PR.

sebald: updated

aromko:
https://www.chromatic.com/build?appId=68679799bacf3b2432835044&number=593

packages/components/src/Inset/Inset.tsx
sarahgm: Hm I think you don't understand me :D I mean the SpacingTokens ALLL are for between elements or not? e-g- related -> when using spacex or spaceY. What does that mean? Related to what?

sarahgm: Inset doesn't work like stack or inline with more elements between

> Relation describes the connection between elements. It defines hierarchy and grouping for margins, flex gaps, and grid gaps. The fundamental rule is:
The tighter the space, the stronger the relationship.
When elements are strongly related, the bond is tight, and they pull close together. When they are distinct, the bond relaxes, and they drift apart.

sebald: aaaah. now I understand. let me take a look and find a solution

sebald: @sarahgm we do not have single tokens for X/Y space only. But since the semantic tokens allow us to set BOTH values anyway:

- remove them and simplify the API (that would be my solution)
- allow non-semantic values only on X/Y spacings

sebald: actually just checked. we ourselves us a lof of only `spaceX` I dont think we offer another layout that only does this. so we need to keep it :D

docs/content/components/layout/inset/index.mdx
```diff
@@ -11,15 +11,15 @@ The component should be used whenever you need to add some whitespace around ele
...
+In order to add equal spacing to all sides use the `space` property with a padding spacing token (`square-*`, `squish-*`, `stretch-*`) or a numeric scale value.
 
 <ComponentDemo name="inset-equal" />
 
 ### Horizontal and vertical spacing
 
 A common use case is to give children of a content component some space, you can do this for example with wrapping an `<Inset>` around the card content like shown below.
 
-To set only horizontal or vertical spacing, or to define different values for each, use the `spaceX` property for horizontal spacing and the `spaceY` property for vertical spacing.
+To set only horizontal or vertical spacing, or to define different values for each, use the `spaceX` property for horizontal spacing and the `spaceY` property for vertical spacing. These props accept inset spacing tokens (`inset-tight`, `inset-snug`, `inset-regular`, `inset-relaxed`, `inset-loose`), as well as numeric scale values.
```
sarahgm: hmm when should i use the tokens? e.g. inset-snug? I need to set when?

packages/components/src/Inset/Inset.tsx
```diff
@@ -20,34 +15,39 @@ export type InsetProps =
...
   | (AriaRegionProps & {
       children: ReactNode;
       /**
-       * The space between the children. You can see allowed tokens [here](../../foundations/design-tokens?theme=core#spacing).
-       * @remarks `PaddingSpaceProp<Tokens>`
+       * The padding of the children. You can see allowed tokens [here](../../foundations/design-tokens?theme=core#spacing).
+       * @remarks `Scale | PaddingSpacingTokens`
        */
-      space?: PaddingSpaceProp['space'];
+      space?: SpaceProp<PaddingSpacingTokens>['space'];
```
sarahgm: why is it here called PaddingSpacingTokens and not InsetSpacingTokens? I think the other ones (spaceX spaceY) should use it. They can also be use then for other component paddings spaceX and spaceY.

packages/components/src/Inset/Inset.tsx
sarahgm: I think we need to think about a new naming for X/Y spacings, also document them.

packages/components/src/Inset/Inset.tsx
sebald: Good catch — the naming was actually backwards. `square`/`squish`/`stretch` are the classic Inset recipes, so those are now `InsetSpacingTokens`. The single-value `tight/snug/regular/relaxed/loose` scale is generic enough to use as padding on any component with `spaceX`/`spaceY`, so it owns the `PaddingSpacingTokens` name.

Renamed:
- `PaddingSpacingTokens` → `InsetSpacingTokens` (the recipes, used by `Inset` `space`)
- `InsetSpacingTokens` → `PaddingSpacingTokens` (the single-value scale, used by `spaceX`/`spaceY`)
- Token values: `inset-{tight|snug|regular|relaxed|loose}` → `padding-{tight|snug|regular|relaxed|loose}`
- CSS: `--spacing-inset-*` → `--spacing-padding-*`

docs/content/components/layout/inset/index.mdx
sebald: updated

sarahgm (approved):

PR #5280, opened 2026-04-08
fix(DST-1124): prevent clipped focus rings in scrollable components

sebald:
## Summary

- Switch the `ui-state-focus` utility from `box-shadow` (`ring-*`) to CSS `outline`, which is not clipped by `overflow: hidden/auto/scroll` on ancestor elements
- Update `ui-state-error` focus state to use `outline` consistently
- Update `Multiselect` manual focus ring duplication to use `outline`
- Remove `p-1 -m-1` padding workarounds from `ActionBar` toolbar and `TopNavigation` middle slot (no longer needed)
- Replace `transition-all` and `transition-colors` with explicit `transition-[color,background-color]` to prevent `outline-color` from animating on focus

## Why `outline` instead of `ring` (box-shadow)?

Tailwind's `ring-*` utilities use CSS `box-shadow` under the hood. This was originally necessary because `outline` did not follow `border-radius` in older browsers. That limitation was resolved in all modern browsers (Chrome 94+, Firefox 88+, Safari 16.4+), and the Marigold codebase already requires these browsers (it uses `oklch()` relative color syntax).

The Tailwind team themselves now recommend `outline-*` for focus rings. From [Adam Wathan in tailwindcss#14602](https://github.com/tailwindlabs/tailwindcss/discussions/14602):

> *"When we do use `ring-*` it's usually as an alternative to border, rather than as a focus ring. We use `outline-*` for focus rings most of the time now."*

Additionally, `box-shadow` is **invisible in Windows High Contrast Mode** (the CSS spec forces `box-shadow: none` in forced-colors mode), while `outline` is preserved and recolored to system colors. This is a significant accessibility improvement for users who rely on High Contrast Mode.

## What changed

**`ui-state-focus`** (the single source of truth for focus rings):
```css
/* Before */
@apply ring-ring/50 ring-[3px] outline-none;

/* After */
@apply outline-solid outline-ring/50 outline-3 outline-offset-0;
```

The visual result is the same: a 3px ring in `--color-ring` at 50% opacity, positioned at the border edge, following `border-radius`. The `--ui-border-color` gradient border effect is unchanged.

**Removed workarounds** that were only needed because `box-shadow` is clipped by overflow:
- `ActionBar` toolbar: removed `p-1 -m-1`
- `TopNavigation` middle: removed `p-1 -m-1`

**Fixed outline-color animation blink** caused by Tailwind v4's `transition-colors` including `outline-color`:
- Replaced `transition-all` with `transition-[color,background-color]` (Calendar cell, Accordion header)
- Replaced `transition-colors` with `transition-[color,background-color]` where hover/selection animations are needed (Tabs, ActionBar clearButton, Switch track, Sidebar nav items)
- Removed `transition-colors` entirely where no color transitions are needed (Calendar controllers, Badge, Slider thumb, Accordion header)

## Test plan

- [x] All 1026 unit tests pass
- [ ] Visual check in Storybook: keyboard-navigate through Dialog, Accordion, Drawer, ListBox, Select to confirm focus rings are visible and not clipped at scroll boundaries
- [ ] Verify focus ring appears instantly with no color blink on Accordion, Slider, Tabs, Calendar, Switch
- [ ] Verify hover/selection color transitions still animate smoothly on Calendar cells, Tabs, Switch track, Sidebar items
- [ ] Verify error state focus on form fields (e.g. TextField with error)


sebald:
@sarahgm running vrt!

sarahgm (approved):

aromko:
https://www.chromatic.com/build?appId=68679799bacf3b2432835044&number=590

PR #5297, opened 2026-04-09
🧹 Reduce ESLint warnings by fixing all react-hooks/rules-of-hooks violation

OsamaAbdellateef:
# Description

his PR clears every react-hooks/rules-of-hooks warning by making Storybook render functions valid hook call sites (named function where hooks run) and renaming internal _… component implementations so hooks are recognized as living in real components. A few small cleanups (e.g. unused props) are included where they overlapped with the same lint pass. No API changes for consumers; public exports stay the same.

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


aromko:
Cut into smaller pieces.

PR #5298, opened 2026-04-10
[DST-1307] improve code coverage

aromko:
# Description

Going though test coverage and improve it for certain components.

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/packages/components/src/ComboBox/ComboBox.stories.tsx
+++ b/packages/components/src/ComboBox/ComboBox.stories.tsx
@@ -417,24 +417,26 @@ export const Mobile: any = meta.story({
     viewport: { value: 'smallScreen' },
   },
   render: args => (
-    <ComboBox {...args}>
-      <ComboBox.Option id="red panda">Red Panda</ComboBox.Option>
-      <ComboBox.Option id="cat">Cat</ComboBox.Option>
-      <ComboBox.Option id="dog">Dog</ComboBox.Option>
-      <ComboBox.Option id="aardvark">Aardvark</ComboBox.Option>
-      <ComboBox.Option id="kangaroo">Kangaroo</ComboBox.Option>
-      <ComboBox.Option id="snake">Snake</ComboBox.Option>
-      <ComboBox.Option id="elephant">Elephant</ComboBox.Option>
-      <ComboBox.Option id="giraffe">Giraffe</ComboBox.Option>
-      <ComboBox.Option id="penguin">Penguin</ComboBox.Option>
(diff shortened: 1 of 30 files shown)


packages/components/src/ComboBox/ComboBox.stories.tsx
```diff
@@ -417,24 +417,26 @@ export const Mobile: any = meta.story({
...
-      <ComboBox.Option id="penguin">Penguin</ComboBox.Option>
-      <ComboBox.Option id="dolphin">Dolphin</ComboBox.Option>
-      <ComboBox.Option id="koala">Koala</ComboBox.Option>
-      <ComboBox.Option id="tiger">Tiger</ComboBox.Option>
-      <ComboBox.Option id="lion">Lion</ComboBox.Option>
-      <ComboBox.Option id="zebra">Zebra</ComboBox.Option>
-      <ComboBox.Option id="owl">Owl</ComboBox.Option>
-      <ComboBox.Option id="fox">Fox</ComboBox.Option>
-    </ComboBox>
+    <I18nProvider locale="en">
```
jim761, line 420: Why is there a I18nProvider?

packages/components/src/FileField/FileField.test.tsx
```diff
@@ -159,3 +173,37 @@ test('hidden input persists after file removal', async () => {
   expect(screen.queryByText('a.pdf')).not.toBeInTheDocument();
   expect(screen.getByText('b.pdf')).toBeInTheDocument();
 });
+
+test('handles file drop on dropzone', async () => {
```
jim761, line 177: Can this be a component test

packages/components/src/ComboBox/ComboBox.stories.tsx
aromko: I think the story is somewhere reused and the allowsEmptyCollection shows a message when no data is there. And it can fail locally when you have locale=de. To make sure it's always the same result the locale is set hardoced to "en".

packages/components/src/FileField/FileField.test.tsx
aromko: With this version no because there is no user interaction. But I can try to use the userEvent from vitest. Than maybe it works as component test.

aromko: Nope. I think this test can stay like it is. There is no way to simulate file drops from outside the browser.

jim761 (approved):

OsamaAbdellateef (approved):

PR #5300, opened 2026-04-10
feat(DST-1312): demo app wrapper for pattern examples

sebald:
## Summary

Consolidates all pattern demos into a unified demo app at `/examples`, giving visitors the experience of navigating a real Marigold application.

- **Shared shell layout** — A `ShellLayout` in `_shared/` renders the sidebar, breadcrumbs, and section switcher from a single declarative navigation tree (`NavNode[]`). Nodes cover leaf pages, collapsible parent items, flat group labels, and separators. One tree walk per render derives the active section, leaf, and ancestor path for breadcrumbs.
- **Section switching** — A ghost `Menu` with lucide icons in the sidebar footer lets users switch between "App Shell" (fake business navigation demonstrating sidebar patterns) and "Patterns" (filter, form, inventory demos).
- **Per-page docs links** — Each example page can link back to its corresponding pattern guide (e.g. Filter → Filter Guidelines, Form → Form Guidelines, App Shell pages → App Shell Pattern).
- **Colocated demo files** — Each demo page's components live in the same folder (e.g. `examples/filter/` contains both `page.tsx` and its toolbar, applied-filter, venues-view components).
- **Shared building blocks** — `PlaceholderPage`, `DashboardContent`, and `UserMenu` are extracted into `_shared/` to eliminate duplication.
- **Removed standalone pages** — Deleted `/pattern/filter`, `/pattern/form`, `/pattern/app-shell`, and `/inventory` standalone routes. Updated all MDX "View Demo" links to point to `/examples/*`.

## Known complexity

The shell is more elaborate than a single-nav demo strictly needs. The two-section split (App Shell + Patterns) and the section-switcher menu exist only to host the App Shell example, which is itself a stand-in for a real pattern page.

Hopefully we can simplify this once App Shell moves inside a dedicated pattern as its own example — at that point the shell collapses to a single section with a richer sub-nav, the section-switcher UI disappears, and `NavSection[]` can reduce to a plain `NavNode[]`. The current shape is designed to degrade gracefully: when `sections.length === 1` the switcher already hides itself, so dissolving App Shell is a matter of deleting it from the config array.

### Final structure
```
docs/app/(examples)/
├── _shared/                    ← Shell layout, NavNode types, building blocks
│   ├── shell-layout.tsx        ← Tree-walking renderer + breadcrumb derivation
│   ├── shell-types.ts          ← NavNode, NavSection, ShellConfig
│   ├── dashboard-content.tsx
│   ├── placeholder-page.tsx
[...]

PR #5306, opened 2026-04-13
feat(DST-1322): add `current` prop to Sidebar.Nav for auto active item

sebald:
## Summary

- Adds a `current` prop on `Sidebar.Nav` that resolves the active leaf automatically — pass `pathname` (string) for smart segment-aware matching, or a predicate `(href, key) => boolean` for full control.
- Removes the `active={pathname === '/...'}` boilerplate on every `Sidebar.Item`. Per-item `active` stays as a local override and stacks with the resolved match (no breaking changes).
- Branch auto-highlight (existing `findActiveBranch`) keeps working when a descendant matches.

```tsx
// Before
<Sidebar.Nav>
  <Sidebar.Item href="/overview" active={pathname === '/overview'}>Overview</Sidebar.Item>
  <Sidebar.Item href="/users" active={pathname.startsWith('/users')}>Users</Sidebar.Item>
</Sidebar.Nav>

// After
<Sidebar.Nav current={pathname}>
  <Sidebar.Item href="/overview">Overview</Sidebar.Item>
  <Sidebar.Item href="/users">Users</Sidebar.Item>
</Sidebar.Nav>
```

### String-mode matching rules

1. Normalize input — strip query (`?…`), hash (`#…`), trim trailing `/`.
2. Leaves only — branches don't participate (their derived href would double-match with the first-leaf descendant).
3. Exact match wins first.
4. Otherwise, longest segment-prefix match (`current` starts with `href + '/'`, segment boundary mandatory). Handles dynamic routes like `/users/abc-123` lighting up `/users`.
5. Root (`/`) is exact-only — never acts as a prefix.
6. Equal-length matches: first in document order wins.

Dev-mode `console.error` when two leaves share the same href (gated behind `process.env.NODE_ENV !== 'production'`).

### Decisions log

The full decision log (identifier strategy, function shape, prop name rationale, override semantics, branch handling) is in [DST-1322](https://reservix.atlassian.net/browse/DST-1322).

Linked: this is the follow-up to DST-1312.

## Test plan

- [x] `pnpm typecheck:only` — clean
- [x] `pnpm test:unit` — 1041/1041 passed (102/102 Sidebar)
- [x] 31 new tests cover normalization, exact/longest-prefix matching, root exact-only, segment boundary, query/hash/trailing slash, predicate mode, branch exclusion, dev-mode duplicate warning, override interaction, and branch auto-highlight via `current`
- [x] Stories (`Basic`, `Complex`, `WithActiveBranch`, `KeyboardNavigation`) updated to use `current` and still work
- [ ] Manual smoke test in Storybook

[DST-1322]: https://reservix.atlassian.net/browse/DST-1322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/content/components/navigation/sidebar/index.mdx
```diff
@@ -87,6 +87,21 @@ The sidebar supports one level of nesting. More levels would force users to reme
 
 <ComponentDemo name="sidebar-drilldown" />
 
+### Active item
+
+The sidebar marks the user's current page so they always know where they are. Pass the current pathname (typically from your router) to the `current` prop on `Sidebar.Nav`, and the matching item lights up automatically:
```
sarahgm, line 92: would update this also in all the demos which has sidebar or? This is the perefered way instead of using `active` ?

packages/components/src/Sidebar/Sidebar.test.tsx
```diff
@@ -1231,3 +1237,384 @@ test('keyboard shortcut Ctrl+B toggles sidebar', async () => {
...
+});
+
+test('normalizePath preserves root when input is empty after stripping', () => {
+  expect(normalizePath('?foo=bar')).toBe('/');
+});
+
+// --- resolveCurrent: string mode ---
+
+const navJsx = () => [
+  <SidebarItem key="home" id="home" href="/">
```
sarahgm: not sure I get it, why can't you reuse the story here?

packages/components/src/Sidebar/Sidebar.test.tsx
```diff
@@ -1231,3 +1237,384 @@ test('keyboard shortcut Ctrl+B toggles sidebar', async () => {
...
+
+  // Branch.href === "/users" via firstLeafHref. Should not trigger duplicate
+  // warning because branches are excluded from leaf duplicate-detection.
+  resolveCurrent(collection, '/users');
+
+  expect(spy).not.toHaveBeenCalled();
+  spy.mockRestore();
+});
+
+// --- applyCurrent ---
```
sarahgm: maybe its better for readablitity to move the tests about the collection in an own test file?

docs/content/components/navigation/sidebar/index.mdx
sebald: yes, wanted to do this afterwards maybe. together with the thing that triggered it #5300 

yes, it really bugged me during #5300 that this is not simpler.

packages/components/src/Sidebar/Sidebar.test.tsx
```diff
@@ -1231,3 +1237,384 @@ test('keyboard shortcut Ctrl+B toggles sidebar', async () => {
...
+  expect(nested?.type === 'item' && nested.active).toBe(true);
+});
+
+// --- Sidebar.Nav `current` integration ---
+
+const renderSidebar = (
+  current?: string | ((href: string | undefined, key: string) => boolean)
+) =>
+  render(
+    <RouterProvider navigate={() => {}}>
```
sarahgm: why can't you reuse here a story?

packages/components/src/Sidebar/Sidebar.test.tsx
sebald: this tests the collection helpers :)

packages/components/src/Sidebar/Sidebar.test.tsx
```diff
@@ -1231,3 +1237,384 @@ test('keyboard shortcut Ctrl+B toggles sidebar', async () => {
...
+        </Sidebar.Provider>
+      </MarigoldProvider>
+    </RouterProvider>
+  );
+
+test('Sidebar.Nav current prop marks matching leaf as active', () => {
+  renderSidebar('/analytics');
+
+  const link = linkByText('Analytics');
+  expect(link).toHaveAttribute('aria-current', 'page');
```
sarahgm: blank lines sometimes missing

packages/components/src/Sidebar/collection.ts
```diff
@@ -105,6 +105,142 @@ const firstLeafHref = (nodes: SidebarNode[]): string | undefined => {
...
+ * Normalize a URL path for matching:
+ * - strip query (`?…`) and hash (`#…`)
+ * - strip trailing slash, except the root (`/`)
+ */
+export const normalizePath = (path: string): string => {
+  const noQuery = path.split('?')[0].split('#')[0];
+  if (noQuery === '' || noQuery === '/') return '/';
+  // Negative lookbehind anchors the match at the first `/` of a trailing run,
+  // preventing polynomial backtracking on inputs like `'/'.repeat(n) + 'x'`.
+  const trimmed = noQuery.replace(/(?<!\/)\/+$/, '');
```
sarahgm: Drop the regex ((?<!\/)\/+$) and replace with a while loop that trims trailing slashes. Same behavior, no regex to reason about.

packages/components/src/Sidebar/collection.ts
```diff
@@ -105,6 +105,142 @@ const firstLeafHref = (nodes: SidebarNode[]): string | undefined => {
...
+            `item should have a unique href.`
+        );
+      } else {
+        seen.set(href, leaf.key);
+      }
+    }
+  }
+
+  // 1. Exact match (first in document order wins).
+  for (const leaf of leaves) {
```
sarahgm: A hint from claude: 

  These can be one loop:

 ``` let exact: string | null = null;
  let best: { key: string; len: number } | null = null;

  for (const leaf of leaves) {
    const href = normalizePath(leaf.href);

    if (href === target) {
      exact = leaf.key;
      break;
    }

    if (href !== '/' && target.startsWith(href + '/')) {
      if (!best || href.length > best.len) {
        best = { key: leaf.key, len: href.length };
      }
    }
  }

  return new Set(exact ? [exact] : best ? [best.key] : []);
```

packages/components/src/Sidebar/collection.ts
sebald: leaving this as a regex, the loop would basically reimplement the regex and is harder to maintain

packages/components/src/Sidebar/Sidebar.test.tsx
sebald: separated them :)

packages/components/src/Sidebar/Sidebar.test.tsx
sebald: These are pure unit tests that call `buildCollection()` / `resolveCurrent()` directly with raw `SidebarItem` JSX — stories render the full component tree so there's no way to extract just the children. The `navJsx` fixture is also specifically crafted with minimal href patterns to exercise edge cases (exact vs prefix match, root exclusion, segment boundaries) that the stories aren't designed for.

packages/components/src/Sidebar/Sidebar.test.tsx
sebald: updated

packages/components/src/Sidebar/Sidebar.test.tsx
sebald: updated

packages/components/src/Sidebar/Sidebar.test.tsx
sebald: moved everything into its own file

packages/components/src/Sidebar/collection.ts
sebald: implemented

docs/content/components/navigation/sidebar/index.mdx
sebald: If its fine doing this later, please resolve comment :)

packages/components/src/Sidebar/Sidebar.test.tsx
sarahgm: same for other tests in the file.

Should I mention in future every line where I think it can be reuse?

sebald: Went through every test in all three new/changed test files. Here's the breakdown:

### Tests already using stories ✅

Most of `Sidebar.test.tsx` already reuses stories — `Basic.Component`, `Complex.Component`, `WithActiveBranch.Component`, `DefaultCollapsed.Component`, `Controlled.Component`. This covers: keyboard navigation (arrow, Home, End, wrapping), roving tabindex, Ctrl+B / Cmd+B shortcuts, collapse/expand, cookie persistence, mobile sheet, sub-panel open/back, branch switching, data-state, inert, separator, focus management, item id/key, and the `current` auto-opens-branch test.

### Tests that can't use stories (with reasons)

**`collection.test.tsx`** (entire file) — Pure unit tests that call `buildCollection()`, `resolveCurrent()`, `normalizePath()`, `applyCurrent()`, `findActiveBranch()` directly with raw JSX. Stories render the full component tree; there's no way to extract just the children to pass to these functions. The test fixtures are also deliberately minimal to exercise specific edge cases (exact vs prefix match, root exclusion, segment boundaries, duplicate href warnings) that the stories aren't designed for.

**`useSidebarNav.test.tsx`** (entire file) — Pure utility/hook tests: `panelPosition` is a plain function, `useRovingItem` tests an error thrown outside provider context, `useLastDistinctValue` tests hook state transitions via `renderHook`. None of these involve rendering a Sidebar component.

**`Sidebar.test.tsx` inline fixtures** — Each of these needs something no story provides:

| Test | Why it can't use a story |
|---|---|
| `onPress callback fires on item click` | Needs to pass a `vi.fn()` mock as `onPress` and assert it was called |
| `branch item onPress fires` | Same — needs mock callback on a branch trigger |
| `nested branches derive href from deepest first leaf` | Tests 3-level nesting; no story has this structure |
| `branch without any leaf hrefs renders without href` | Tests edge case with hrefless children; no story covers this |
| `group label inside branch renders correctly` | Tests GroupLabel/Separator _inside_ a branch; stories only have them at root |
| `textValue auto-extracted from string children` | Needs a branch _without_ explicit `textValue` prop; all stories set it explicitly |
| `direct branch-to-branch switch via active prop` | Needs an external button that changes active state to trigger branch-to-branch transition |
| `panel transition focuses active item` | Needs an external button to activate a child mid-session (tests focus-on-reentry) |
| `panel transition falls back to back button` | Needs a branch where no child has `active`; stories always have active items |
| `marker components return null` | Calls `Sidebar.Item({})` etc. directly — no rendering at all |
| `current` integration tests (renderSidebar helper) | Need to control the `current` prop value directly per test; stories wire `current` to internal `useState` so there's no way to inject a specific value from outside |
| `explicit active prop wins as local override` | Tests conflict resolution between `current` and per-item `active` — not a scenario any story covers |
| `without current prop falls back to per-item active` | Tests the no-`current` fallback path; all stories now use `current` |

sarahgm: collection and useSidebar test i understand, but you changed one test with a story in Sidebar.test file so I thought it might be possible to change the other ones in this file also :)

sarahgm (approved):

aromko (approved):

PR #5307, opened 2026-04-13
feat([DSTSUP-246]): Add virtualizer to TagField component

aromko:
# Description

Add virtualization to the `<TagField>` dropdown to support large datasets (10k+ items).

Previously, the TagField rendered all items as DOM nodes when the dropdown opened. With 13k items this caused the browser to freeze on open and during filtering. This was reported via a [StackBlitz reproduction](https://stackblitz.com/edit/github-s4wn1l9h).

## Solution

Uses React Aria's built-in `<Virtualizer>` + `ListLayout` (already available in `react-aria-components@1.16.0`) to only render visible items in the DOM. Following [React Spectrum 2's approach](https://react-spectrum.adobe.com/react-aria/Virtualizer.html), virtualization is always-on inside the TagField — consumers don't need to opt in.

### Changes

**ListBox** (`packages/components/src/ListBox/`):
- Added `virtualized` and `layoutOptions` props to `ListBox`
- When `virtualized` is true, wraps the RAC ListBox with `<Virtualizer layout={ListLayout}>`
- `ListBoxItem` applies required virtualizer styles (`height: 100%`, `minHeight: 0`) via context
- Backward-compatible: non-virtualized ListBox behavior is unchanged

**TagField** (`packages/components/src/TagField/`):
- Passes `virtualized` to the ListBox in `TagFieldDropdown`
- Added `LargeDataset` story (800 items) with a component test for visual regression

**Test infra** (`vite.config.ts`, `vitest.config.shared.ts`):
- Added `define` for `process.env.NODE_ENV` / `process.env.VIRT_ON` in vitest browser-mode projects — `@react-aria/virtualizer` references `process.env` which doesn't exist in the browser test environment (Firefox via Playwright)
- Added `@react-aria/virtualizer` and `@react-stately/layout` to `browserDeps` for pre-bundling

# Test Instructions:

1. Run `pnpm sb`, open the **LargeDataset** story → dropdown should open instantly, scrolling should be smooth
2. Search for a specific item (e.g. "item-500") → results filter instantly
3. Select items → tags appear correctly, removal works
4. Verify **Basic** and **Controlled** stories still work as before
5. Run `pnpm test:unit` → all tests pass
6. Run `pnpm test:sb` → story tests pass (including new LargeDataset test)

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

### DIFF
--- a/packages/components/src/TagField/TagField.tsx
+++ b/packages/components/src/TagField/TagField.tsx
@@ -196,6 +196,7 @@ const TagFieldDropdown = ({
       </SearchField>
       <ListBox
         items={items}
+        virtualized
         renderEmptyState={() =>
           emptyState ?? (
             <div className="flex items-center">
--- a/packages/components/src/ListBox/Context.ts
+++ b/packages/components/src/ListBox/Context.ts
@@ -3,6 +3,7 @@ import type { ComponentClassNames } from '@marigold/system';
 
 export interface ListBoxContextProps {
   classNames: ComponentClassNames<'ListBox'>;
+  virtualized?: boolean;
 }
 
 export const ListBoxContext = createContext<ListBoxContextProps>({
--- a/packages/components/src/ListBox/ListBox.tsx
+++ b/packages/components/src/ListBox/ListBox.tsx
@@ -5,7 +5,8 @@ import {
   forwardRef,
 } from 'react';
 import type RAC from 'react-aria-components';
-import { ListBox } from 'react-aria-components';
+import { ListBox, ListLayout, Virtualizer } from 'react-aria-components';
+import type { ListLayoutOptions } from 'react-aria-components';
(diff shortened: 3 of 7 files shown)


packages/components/src/TagField/TagField.tsx
```diff
@@ -196,6 +196,7 @@ const TagFieldDropdown = ({
       </SearchField>
       <ListBox
         items={items}
+        virtualized
```
sebald, line 199: Do we want to enable this on other components that use `<Listbox>` too?

aromko: I think we could do it. Wouldn't know a disadvantage or we wait until someone needs it.

sebald (approved):

PR #5312, opened 2026-04-13
feat(DST-1323): enable virtualization for Select, ComboBox, and Autocomplete

sebald:
## Summary

Follow-up to #5307. Turns on the `<ListBox>` `virtualized` prop for the three remaining ListBox-based components so large datasets (hundreds to thousands of items) no longer freeze the browser on open or filter. Virtualization is always-on — no public API change, matching the `<TagField>` pattern.

- `<Select>`, `<ComboBox>`, `<Autocomplete>`: desktop dropdowns now virtualize
- `<ListBox>`: the virtualized list now has a bounded max-height (24rem / `max-h-96`) so the `Virtualizer` has a viewport to clip against when used inside `<Popover>`. Non-virtualized ListBoxes are unchanged.
- Adds a `LargeDataset` story (800 items) with a component-test play function for each of the three components, mirroring the `<TagField>` `LargeDataset` setup.

Ticket: [DST-1323](https://reservix.atlassian.net/browse/DST-1323)

## Test plan

- [x] `pnpm test:sb` → 116 passed (incl. the three new `LargeDataset` tests)
- [x] `pnpm test:unit` → 1011 passed
- [x] `pnpm typecheck:only` → clean
- [x] `pnpm lint` → 0 errors
- [x] Manual check in Storybook (Firefox): opened each new `LargeDataset` story, dropdown renders instantly, scrolls smoothly, selection works. Verified via devtools that only ~16 of 800 option elements are in the DOM (`listboxClientHeight: 384`, `listboxScrollHeight: 26407`).
- [x] Sanity-checked existing stories: `Basic` / `Sections` / `WithSections` / `LotsOfOptions` / `Mobile` for each component still pass.

## Pull Request Checklist

- [x] Added/updated unit tests and storybook for this change
- [ ] Added/Updated documentation (if it already exists for this component) — n/a, no public API change

[DST-1323]: https://reservix.atlassian.net/browse/DST-1323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/listbox-virtualization-everywhere.md
+++ b/.changeset/listbox-virtualization-everywhere.md
@@ -1,0 +1,10 @@
+---
+"@marigold/components": minor
+---
+
+feat(DST-1323): always-on virtualization for `<Select>`, `<ComboBox>`, and `<Autocomplete>` dropdowns
+
+Follow-up to #5307. Large datasets (hundreds to thousands of items) no longer freeze the browser when opening or filtering these components — the internal `<ListBox>` is now virtualized with `react-aria`'s `Virtualizer` + `ListLayout` (same pattern used by `<TagField>`).
+
+- `<Select>`, `<ComboBox>`, `<Autocomplete>`: virtualization is enabled by default on desktop, with no public API change
+- `<ListBox>`: the virtualized list now has a bounded height (`max-h: 24rem`) so the virtualizer has a viewport to clip against when used inside a `<Popover>`
--- a/packages/components/src/Autocomplete/Autocomplete.stories.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.stories.tsx
@@ -1,7 +1,7 @@
 import { screen } from '@testing-library/react';
 import { useState } from 'react';
 import { Text } from 'react-aria-components';
-import { expect, userEvent, waitFor } from 'storybook/test';
+import { expect, userEvent, waitFor, within } from 'storybook/test';
 import preview from '.storybook/preview';
 import { useAsyncList } from '@react-stately/data';
 import { Center } from '../Center/Center';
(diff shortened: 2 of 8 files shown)


sebald:
### Why the virtualized `<ListBox>` gets an explicit `max-height`

_Context for the inline `maxHeight: '24rem'` added to `ListBox.tsx` when `virtualized` is on._

#### How virtualization works

`<Virtualizer layout={ListLayout}>` from `react-aria-components` wraps the `<ListBox>` and only mounts the items currently visible in the scroll container. Internally it calls `useScrollView(scrollRef)`, where `scrollRef` is the `<ListBox>` DOM element. It reads that element's **`clientHeight`** — the visible viewport — and uses it to decide which slice of rows (e.g. items 0–15 of 800) to render into the DOM.

#### Why the `maxHeight` matters

For the virtualizer to clip anything, that scroll container needs a **bounded height**. If it's unbounded, `clientHeight` ends up equal to the content's natural height — i.e. "I can see all 800 rows", so none get clipped.

In `TagField` the ListBox lives inside a `<Dialog>` wrapper with constraints, so it's naturally bounded. In `Select`/`ComboBox`/`Autocomplete` the ListBox lives inside a plain `<Popover>`:

- `packages/components/src/Overlay/Popover.tsx` → no max-height
- `themes/theme-rui/src/components/Popover.styles.ts` → no max-height

React Aria's Popover component also doesn't auto-apply a `max-height` (checked `node_modules/.../react-aria-components/dist/Popover.mjs` — no `maxHeight` / `--max-height` there). So when you put 800 rows into it, the Popover grows to ~25,600px (800 × ~32px). The `<ListBox>` has `overflow-y-auto` but, without a bounded parent, it also grows to fit its children. Result: `clientHeight ≈ 25,600px`, the virtualizer sees everything as "visible", all 800 `<li>` nodes mount, and virtualization is a no-op.

This is the failure I reproduced in Playwright — the `Select` `LargeDataset` test saw 800 option elements in the DOM.

#### The fix

```ts
style={
  virtualized
    ? { display: 'block', padding: 0, maxHeight: '24rem' }
    : undefined
}
```

`maxHeight: '24rem'` (= 384px, matches Tailwind's `max-h-96`) gives the ListBox a concrete ceiling. The virtualizer then sees `clientHeight: 384`, computes that only ~12 rows fit at 32px apiece, and keeps ~16 rows in the DOM (with a small overscan buffer). Combined with the existing `overflow-y-auto` on the list, the user sees a normal scrolling dropdown.

Verified in the browser via Playwright MCP:

```
optionCount: 16                 // was 800
listboxClientHeight: 384        // the bounded viewport
listboxScrollHeight: 26407      // the full virtual size
listboxInlineStyle: "display: block; padding: 0px; max-height: 24rem;"
```

#### Why inline style, not a className

First attempt was `virtualized && 'max-h-96'` via `cn()`. An inline style is more robust: it can't be overridden by a later class in `classNames.list`, it doesn't depend on Tailwind's content scanner generating the utility, and it's co-located with the other virtualization-only style (`display: block; padding: 0`) that #5307 put inline for the same reason — those are layout concessions the virtualizer specifically needs, not theme decisions.

#### Why only when `virtualized`

Non-virtualized ListBoxes don't need a bound for correctness — a 5-item Select dropdown should size to its content, not be capped at 24rem. The height cap is specifically a prerequisite of virtualization, so it's gated on the same flag.

aromko (approved):

PR #5320, opened 2026-04-14
refa([DST-1162]): 🧹 Refa Card to Compound Component Pattern

OsamaAbdellateef:
- Introduced Card subcomponents: CardHeader, CardBody, CardFooter, and CardPreview for better organization and usage.
- Updated Card component to use a grid layout and removed deprecated padding props in favor of a more semantic approach.
- Adjusted stories and tests to reflect the new structure and ensure proper rendering of the Card component and its subcomponents.
- Enhanced styling and layout management through the CardContext for improved maintainability.

# Description

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


.changeset/tiny-baths-accept.md
```diff
@@ -0,0 +1,8 @@
+---
+'@marigold/components': major
+'@marigold/theme-rui': major
+'@marigold/system': major
+'@marigold/docs': major
+---
+
+refa([DST-1162]): **Breaking changes**: The `Card` component has been refactored into a compound component. Sub-components like `Card.Header`, `Card.Body`, `Card.Footer`, and `Card.Preview` must now be used explicitly to compose the card layout. This replaces the previous prop-based API.
```
sarahgm: maybe you can add here more details what exactly is breaking change

themes/theme-rui/src/components/Card.styles.ts
```diff
@@ -1,16 +1,29 @@
...
+    },
+  }),
+  header: cva({
+    base: 'px-4 pt-4 pb-2 text-lg font-semibold',
+  }),
+  body: cva({
+    base: 'px-4 py-2 text-sm',
+  }),
+  footer: cva({
+    base: 'flex items-center  gap-2 px-4 pt-2 pb-4',
```
sarahgm:
```suggestion
    base: 'flex items-center gap-2 px-4 pt-2 pb-4',
```

packages/components/src/Scrollable/Scrollable.stories.tsx
```diff
@@ -135,7 +135,7 @@ export const WithTable = meta.story({
 
 export const ListScrolling = meta.story({
   render: args => (
-    <Card p={3}>
+    <Card>
       <Scrollable height="200px" {...args}>
```
sarahgm: This demo is not using Card slots

packages/components/src/Card/CardFooter.tsx
```diff
@@ -0,0 +1,16 @@
+import type { ReactNode } from 'react';
+import { cn } from '@marigold/system';
+import { useCardContext } from './CardContext';
+
+export interface CardFooterProps {
+  children?: ReactNode;
```
sarahgm, line 7: descriptions missing for interfaces, also the other components

packages/components/src/Card/CardContext.tsx
```diff
@@ -0,0 +1,24 @@
+import { createContext, useContext } from 'react';
```
sarahgm: we can also use `use` instead of `useContext` now

themes/theme-rui/src/components/Card.styles.ts
```diff
@@ -1,16 +1,29 @@
...
-      default: 'ui-surface',
-      master: 'bg-access-master/40 border border-access-master-accent',
-      admin: 'bg-access-admin/40 border border-access-admin-accent',
+export const Card: ThemeComponent<'Card'> = {
+  container: cva({
+    base: 'rounded-md shadow-elevation-raised',
+    variants: {
+      variant: {
+        default: 'ui-surface',
+        master: 'bg-access-master/30 border border-access-master',
```
sarahgm:
```suggestion
        master: 'bg-access-master/40 border border-access-master-accent',
```

themes/theme-rui/src/components/Card.styles.ts
```diff
@@ -1,16 +1,29 @@
...
+        admin: 'bg-access-admin/30 border border-access-admin',
```
sarahgm:
```suggestion
        admin: 'bg-access-admin/40 border border-access-admin-accent',
```

So it should take the current tokens

docs/app/(examples)/inventory/ContactTiles.tsx
```diff
@@ -15,13 +15,13 @@ export const ContactTiles = () => (
...
-        <Card key={person.id} pt={8} pb={6} px={0} stretch>
-          <Stack space={12} alignY="between" stretch>
+        <Card key={person.id} stretch>
+          <Card.Body>
             <Stack space={8} alignX="center">
               <img
                 src={person.avatar}
                 alt={person.name}
-                className="block w-32 rounded-full"
+                className="mt-4 block w-32 rounded-full"
```
sarahgm: this margin isn't needed or?

docs/content/components/layout/tiles/tiles-complex.demo.tsx
```diff
@@ -4,12 +4,14 @@ import { Card, Headline, Stack, Text, Tiles } from '@marigold/components';
...
-      <Card key={crypto.randomUUID()} p={4}>
-        <Stack space={2} alignX="center">
-          <img src={venue.image} alt={venue.name} width={200} height={200} />
-          <Headline level={3}>{venue.name}</Headline>
-          <Text>{venue.description}</Text>
-        </Stack>
+      <Card key={crypto.randomUUID()}>
+        <Card.Body>
+          <Stack space={2} alignX="center">
+            <img src={venue.image} alt={venue.name} width={200} height={200} />
```
sarahgm: wouldn't this image be a Card.Preview Item?

docs/content/components/collection/card/index.mdx
```diff
@@ -9,7 +9,12 @@ Cards are ideal for grouping related information and making it easily digestible
 
 ## Anatomy
 
-The `<Card>` component consists of a container which is basically a `<div>`. You can pass any content and components as `children` to the component.
+The `<Card>` component is a compound component made up of several sub-components that map to distinct visual areas of the card:
```
sarahgm: I would expect in the docs more usage guidelines about the new parts, when and how should i use them, some dos and don'ts - and better demos for the card

docs/content/components/content/card/card-elevation.demo.tsx
```diff
@@ -1,76 +1,43 @@
 import { Card, Inline, Stack, Text } from '@marigold/components';
 
+const now = Date.now();
```
sarahgm: Think this is maybe even not needed anymore (we don't have sunken anyway), would make a more meaningful example

docs/content/components/collection/card/index.mdx
OsamaAbdellateef: I think is more related/included to this ticket  https://reservix.atlassian.net/browse/DST-876  or ? @sarahgm

sarahgm: Hmm can be both 🤷‍♀️ I think its fine to do it in the other ticket

sarahgm (approved):

sebald:
EDIT: @sarahgm showed me the discussion you two had :D so ignore this comment below.

---

Took a read through the Card docs and honestly, even on its own terms the page doesn't quite land what a Card *is*. A few thoughts:

**The framing is off.** The opener calls Card *"a versatile UI element used to display content"* and the frontmatter says *"Component to support the structuring of information"*, which is pretty broad. The word "collection" doesn't appear anywhere on the page, even though that's really the job. A Card is one item in a grid or list of similar things. Something like *"Card represents a single item in a collection, like a product in a grid, an event in search results, or a profile in a list."* would do a lot of work up top and set expectations correctly.

**No "when NOT to use Card" guidance.** The Usage section explains how to compose the slots but never tells you when Card is the wrong tool. A short "Card is for collection items, not for boxing off arbitrary page content" note would help.

**Small stuff:**
- No a11y section. Worth covering linked or pressable cards and collection semantics (wrapping in a list, `GridList`, etc.).
- `Card.Body` is the only slot without a code example.
- The "Example" demo is named `card-elevation`, which reads like an elevation demo rather than a canonical card.
- `admin` and `master` variants could use a one-liner of context before the pattern link.

**TL;DR:** the doc describes the *shape* of Card well and the *role* of Card not at all. Even pre-Panel, tightening the intro around "item in a collection" would make the page much clearer, and it sets us up nicely for adding a Card vs Panel section once Panel lands.

sebald (approved):

PR #5321, opened 2026-04-15
fix(DST-1325): rename spacing token `none` to `collapsed`

sebald:
## Summary

- Rename `NoSpacingToken` from `'none'` to `'collapsed'` and `--spacing-none` to `--spacing-collapsed`
- Fixes a Tailwind v4 collision where `leading-none` resolved to `var(--spacing-none)` (= 0) instead of the built-in `line-height: 1`
- Other spacing tokens (`tight`, `snug`, `relaxed`, `loose`) are **not affected** — they have matching `--leading-*` variables that take precedence over the `--spacing-*` fallback

### Why `collapsed`?

The token represents the absence of spacing — gap or padding collapsing to zero. "Collapsed" is an established CSS/design term (cf. margin collapse) and reads naturally in both contexts:

```tsx
<Stack space="collapsed">           // gap collapses to zero
<Panel.Content inset="collapsed">   // padding collapses to zero
```

### Follow-up to #5313

This is a targeted fix for the `--spacing-none` token introduced in #5313. The Panel component branch (`feat/DST-1326-panel-scaffold`) will rebase on this once merged.

## Test plan

- [ ] `pnpm sb` → verify `leading-none` renders `line-height: 1` (not 0)
- [ ] Stack/Inline/Inset story controls show `collapsed` option and it produces zero spacing


themes/theme-rui/src/theme.css
```diff
@@ -186,7 +186,7 @@
   --spacing-container: 50rem;
 
   /* Universal: no spacing (works for gap, padding, and inset recipes) */
-  --spacing-none: --spacing(0);
+  --spacing-collapsed: --spacing(0);
```
sarahgm, line 189: I think this token should be also mentioned in the Spacing Guideline

aromko, line 189: 🔍 **Visual Regression Tests**

You can trigger it by commenting `/run-chromatic` on this PR, or by manually dispatching the workflow.

packages/components/src/Inline/Inline.stories.tsx
```diff
@@ -16,7 +16,7 @@ const meta = preview.meta({
       control: {
         type: 'select',
       },
-      options: ['none', 'tight', 'related', 'regular', 'group', 'section'],
+      options: ['collapsed', 'tight', 'related', 'regular', 'group', 'section'],
```
aromko, line 20: 💡 This story (and Stack/Inset stories) doesn't have `tags: ['component-test']` or a `play` function. Not introduced by this PR, but the renamed token would benefit from an interaction test verifying that `collapsed` actually produces zero spacing.

themes/theme-rui/src/theme.css
sebald: Good call — added `collapsed` to both the Relation and Inset scale tables (and the CSS Tokens references) in the Spacing guideline in 67c512fce.

packages/components/src/Inline/Inline.stories.tsx
sebald: Added dedicated `Collapsed` stories for Inline, Stack, and Inset with `tags: ['component-test']` and play functions asserting `gap`/`rowGap`/`padding` resolve to `0px` (67c512fce).

PR #5326, opened 2026-04-16
feat(DST-1237): theme-owned breakpoints with CSS fallback

sarahgm:
# Description

Breakpoint resolution is now theme-driven: `useSmallScreen` and `useResponsiveValue` read `theme.screens` from the ThemeProvider context instead of relying on hardcoded values in `defaultTheme`. If no theme provides screens, the hooks fall back to reading Tailwind v4's `--breakpoint-*` CSS custom properties.

**Changes:**
- Added `screens` to `@marigold/theme-rui` (matches Tailwind v4 defaults)
- Removed `screens` from `defaultTheme` in `@marigold/system`
- Added `resolveScreens` utility for theme-first, CSS-fallback resolution
- Refactored `useSmallScreen` to use `useSyncExternalStore`
- Updated `useResponsiveValue` docs and story
- Updated all component tests to derive breakpoint query from theme

# Test Instructions:

- Run `pnpm typecheck:only` to verify type safety
- Run `pnpm test:unit` to verify all unit tests pass (1051 tests across 106 files)
- Run `pnpm sb` and check the `System/useResponsiveValue` story -- resize browser to see breakpoint changes

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


packages/components/src/DatePicker/DatePicker.stories.tsx
```diff
@@ -7,6 +7,12 @@ import { I18nProvider } from '@react-aria/i18n';
 import { Stack } from '../Stack/Stack';
 import { DatePicker } from './DatePicker';
 
+/**
+ * Must match the `sm` breakpoint provided by the theme.
+ * Used only in the matchMedia fallback decorator below.
+ */
+const smallScreenQuery = '(width < 40rem)';
```
OsamaAbdellateef: Do we need to hardcode the value `40rem` or we use the value from the theme so it more dynamic theme.screens.sm ?

docs/content/components/hooks-and-utils/useResponsiveValue/index.mdx
```diff
@@ -1,18 +1,14 @@
 ---
 title: useResponsiveValue
```
OsamaAbdellateef, line 2: I think it isn't clear that `useSmallScreen` is also in this page until we scroll down, Can we make the title useResponsiveValue and useSmallScreen what do you think  ?

packages/system/src/hooks/resolveScreens.ts
```diff
@@ -0,0 +1,16 @@
+const BREAKPOINT_NAMES = ['sm', 'md', 'lg', 'xl', '2xl'] as const;
+
+export function resolveScreens(
```
OsamaAbdellateef: Replace function with const so it is consistent with the others

.storybook/preview.tsx
```diff
@@ -8,6 +8,9 @@ import withMarigoldProviders from './decorators.js';
 // @ts-expect-error - CSS side-effect import handled by Vite bundler
 import './styles.css';
 
+// Storybook viewports only accept fixed pixel values, so we can't use the
```
sebald, line 11: 🔴 **Visual Regression Tests — stale**

The last successful `Visual-Regression-Tests` run was on commit `e395c50d`, but the branch has since moved to `e35bde41`.

Please re-run visual regression tests to cover the latest changes by commenting `/run-chromatic` on this PR.

packages/components/src/DatePicker/DatePicker.stories.tsx
sebald: 💡 This hardcodes `'(width < 40rem)'` while all test files derive the query from `theme.screens!.sm`. If the `sm` breakpoint value ever changes, this story will silently break. Consider importing from the theme for consistency:

```suggestion
import { theme } from '@marigold/theme-rui';
const smallScreenQuery = `(width < ${theme.screens!.sm})`;
```

packages/system/src/hooks/useResponsiveValue.ts
```diff
@@ -1,26 +1,21 @@
...
+ * using breakpoints from the theme (`theme.screens`). Note that this
+ * hook is client-side only.
  */
 export const useResponsiveValue = <T>(
   values: T[],
   defaultIndex: number = 0
 ) => {
   const theme = useTheme();
-  const screens = theme.screens || defaultTheme.screens;
+  const screens = useMemo(() => resolveScreens(theme.screens), [theme.screens]);
```
sebald, line 15: 💡 `useSmallScreen` was nicely refactored to `useSyncExternalStore` in this PR, but `useResponsiveValue` still uses `useState` + `useEffect` with `index` in the dependency array (causing an extra render cycle on mount). Consider aligning both hooks to `useSyncExternalStore` in a follow-up.

packages/system/src/hooks/useResponsiveValue.stories.tsx
```diff
@@ -1,41 +1,49 @@
 import preview from '.storybook/preview';
```
sebald, line 1: 💡 Pre-existing filename typo: `useResponiveValue.stories.tsx` is missing the 's' in "Responsive". Since the file is already being modified, it could be a good opportunity to rename it to `useResponsiveValue.stories.tsx`.

docs/content/components/hooks-and-utils/useResponsiveValue/index.mdx
sarahgm: That would be very long title... maybe we should consider whether we want to make this more general?

packages/system/src/hooks/useResponsiveValue.ts
sarahgm: I would do this then now

docs/content/components/hooks-and-utils/useResponsiveValue/index.mdx
sarahgm: I created an Issue about it: https://reservix.atlassian.net/browse/DST-1344

sebald (approved):
!

PR #5337, opened 2026-04-19
docs(DST-1349): centralize design-system alias display in AutoTypeTable

sebald:
Jira: [DST-1349](https://reservix.atlassian.net/browse/DST-1349)

## Summary

Improves how the docs' `AutoTypeTable` renders props whose types reference design-system aliases (`SpacingTokens`, `Scale`, `WidthProp`, `HeightProp`, `MaxWidthProp`, `InsetSpacingTokens`, `PaddingSpacingTokens`, `NonZeroPercentage`):

- **Main cell:** short, accurate alias summary (e.g. `SpacingTokens | Scale`).
- **Collapsible row:** fully-resolved literal union (e.g. `"96" | "80" | ... | "tight" | "related" | 0`).

Previously the cell showed the fabricated generic `SpacingTokens<Tokens>` (or similar) and the collapsible showed the same alias names — no new information on expand.

This replaces 27 per-prop `@remarks \`TypeName\`` JSDoc overrides with a single transform in `docs/ui/autoTypeTableTransform.ts`, so future components pick up the same behavior automatically.

### Before vs after

| Prop | Cell before | Cell after | Expanded after |
| --- | --- | --- | --- |
| `Stack.space` | `SpacingTokens<Tokens>` | `SpacingTokens \| Scale` | `"96" \| "80" \| ... \| "tight" \| 0` |
| `Inset.spaceX` | `Scale \| PaddingSpacingTokens` | `Scale \| PaddingSpacingTokens` | `"padding-loose" \| ... \| 0` |
| `Autocomplete.width` | `WidthProp` | `WidthProp` | `"container" \| ... \| "auto" \| 0` |
| `Grid.height` | `HeightProp<Tokens>` | `HeightProp` | `"11/12" \| ... \| "auto" \| 0` |
| `Aside.wrap` | `NonZeroPercentage` | `NonZeroPercentage` | `"99%" \| "98%" \| ... \| "1%"` |

No regressions on props without design-system aliases (`alignX`, `variant`, RAC-derived props, etc.) — the transform is a no-op for them.

## How the transform works

- **Primary path:** regex-extracts `import("…").Alias` references that TS preserves in `Type.getText()`, filtered to paths containing `/packages/system/`, `/packages/types/`, or the equivalent `node_modules/@marigold/…` paths. Sets the cell to those names; copies fumadocs' literal expansion (already in `simplifiedType` at transform-run time) into `type` for the collapsible.
- **Fallback:** when TS has flattened an indexed access (`WidthProp['width']` → `'auto' | 'full' | …` with no surviving alias reference), surfaces the wrapper name the author wrote, provided its declaration lives in a design-system package.
- **Escape hatch:** a `@remarks \`TypeName\`` tag on a prop still wins — fumadocs sets `simplifiedType` from the tag before the transform runs, and the transform skips props that have a `remarks` tag.

## Side changes

- `Multiselect.width` and `ComboBox.width` now use `WidthProp['width']` directly instead of `FieldBaseProps<'label'>['width']` — structurally identical, no runtime change. This lets the fallback path recognize them consistently.
- All 27 now-obsolete `@remarks \`TypeName\`` JSDoc tags removed.

## Verification

Captured `{component → prop → { simplifiedType, type } }` snapshots before and after across all 27 affected components (snapshot script was temporary and is not part of this PR). Every diff in the main cell is a strict improvement or cosmetic re-ordering; every diff in the expanded `type` gains literal values instead of losing them.

## Test plan

- [ ] Spot-check `Stack`, `Inset`, `Multiselect`, `Grid`, `Aside` pages in `pnpm start` to confirm cell + expand render correctly
- [ ] Verify `alignX` (no design-system alias) still shows its literal union unchanged

[DST-1349]: https://reservix.atlassian.net/browse/DST-1349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sebald:
Closing — the transform swaps the main-cell text correctly but the collapsed "Type" row still shows `SpacingTokens | Scale | undefined` instead of the fully-resolved literal union, so there's no real gain over main. Fumadocs' `entry.type` is generated with `UseAliasDefinedOutsideCurrentScope`, which preserves alias names in the collapsed view too.

Future attempt would need to also override `entry.type` inside the transform with the literal expansion (e.g. reusing fumadocs' own `getSimpleForm` output, which `simplifiedType` holds before this transform runs). Tracked in DST-1349.

sebald:
Reopened with the fix. The transform now also copies fumadocs' `getSimpleForm` output (the literal expansion that was in `simplifiedType` before the transform ran) into `entry.type`. Result:

- **Cell:** alias names (e.g. `SpacingTokens | Scale`, `WidthProp`, `NonZeroPercentage`)
- **Collapsed row:** fully-resolved literal union (e.g. `"96" | "80" | ... | "tight" | 0`, all 99 percentage literals for `NonZeroPercentage`, etc.)

Verified with the same snapshot approach across all 27 components.

aromko (approved):

PR #5343, opened 2026-04-21
feat(DST-1351): switch AppLayout to page-level scroll

sebald:
## Summary

- `AppLayout` delegates scrolling to the document. `<AppLayout.Main>` loses its scroll container; the grid uses `min-h-dvh` instead of `h-dvh` + `overflow-hidden`.
- `<Sidebar>` becomes `sticky top-0 h-dvh self-start` so it stays pinned as the page scrolls. The header stays pinned through `TopNavigation`'s own sticky positioning.
- `@marigold/theme-rui` adds a new `preflight.css` with two peer-dependency fixes on `<html>` / `<body>` that page-level scroll needs. The rules ride inside both `theme.css` and the pre-built `styles.css`; `postcss-prefix-selector` excludes `html`/`body` so they reach the document root while the rest of the bundle stays scoped to `[data-theme="rui"]`.
- `ui-scrollbar`'s track is now transparent so the themed scrollbar blends into any surface.
- **theme-rui CSS file layout refactor** (see below) to fix a token-resolution bug and clarify what each entry point is for.

## Why page-level scroll

- **Mobile URL bar** collapses on scroll (interior scroll keeps it expanded — ~8% wasted screen).
- **Pull-to-refresh** works.
- **Scroll restoration** on back/forward is reliable (interior containers produce "lost scroll position" bugs).
- **`Cmd+F` find-in-page** scrolls matches into view.
- **Anchor links, iOS status-bar scroll-to-top, keyboard nav** (PgUp/PgDn/Space/Home/End) behave predictably.
- **`IntersectionObserver` with default root, scroll-snap, sticky, `scroll-margin-top`** — all simpler with one scroll container.

## Why `preflight.css`

- `html { scrollbar-gutter: stable }` — when `@react-aria/overlays` locks the page (sets `overflow: hidden` on `<html>` and compensates scrollbar width), reserving the gutter prevents the 1 px reflow that would otherwise occur on every overlay open/close.
- `body { position: relative; overflow-x: clip }` — `@react-aria/live-announcer` portals into `document.body` at `top: -10000px; left: -10000px`. With a static body that absolute element can expand the document's scrollable area. `position: relative` contains it. `clip` (not `hidden`) avoids creating a new scroll context that would break `position: sticky` on descendants.

Both rules target the real `<html>` / `<body>`. They cannot ride inside the scoped parts of the theme (`[data-theme="rui"] html` never matches), so the postcss prefixer is configured to exclude `html`/`body` — everything else in the bundle stays scoped.

## theme-rui CSS restructure

Two motivations: fix a token-resolution bug where `postcss-prefix-selector` was rewriting `@theme`'s `:root, :host` output to `[data-theme="rui"], [data-theme="rui"] :host`, which stranded the design tokens inside the scoped bundle and broke unscoped base rules that tried to resolve `var(--color-background)`. And make the file layout legible so each file has one job.

**File layout**

- `tokens.css` (new) — the `@plugin` declaration and every design token in one place. Easy to find, easy to extend.
- `theme.css` (Tailwind-native entry) — imports `preflight.css` + `tokens.css` + `ui.css` + `variants.css` and paints `<body>` directly. No `[data-theme]` attribute required. A Marigold-first app works with zero extra markup.
- `styles.css` (pre-compiled entry) — imports the same base and paints `[data-theme="rui"]`. The consumer places the attribute on `<html>`, `<body>`, or a wrapper `<div>` to control where Marigold paints — whole-app or island.
- `global.css` (removed, never released) — the body paint now lives inline in the two entry points that actually need it, so a shared file with ambiguous semantics is no longer necessary.

**Two clean mental models**

- Use `theme.css` with your own Tailwind build when Marigold is the whole app. Body paints automatically, tokens live at `:root`, utilities tree-shake against your content.
- Use `styles.css` as a pre-compiled drop-in when Marigold is an island or you are not running Tailwind. Place `data-theme="rui"` wherever Marigold should paint.

**Prefixer exclude list**

`:root`, `:host`, and `[data-theme="rui"]` are now excluded from prefixing in addition to `html`/`body`. That keeps tokens globally resolvable while utility classes remain scoped. The `[data-theme="rui"]` exclude uses a quote-agnostic regex so prettier round-trips don't reintroduce double-prefixing.

**CSS exports carry a `style` condition**

Tailwind v4's CSS resolver uses `conditionNames: ["style"]`. Bare-string export entries without a matching condition fail under strict resolvers, so every `.css` subpath now declares both `style` and `default` targets. The unused `./*` JS catchall is removed. New subpath exports: `./tokens.css`, `./preflight.css`. Removed: `./global.css` (never released).

## Breaking changes

- Code reading `mainRef.current.scrollTop` must read `window.scrollY` / `document.documentElement.scrollTop` instead.
- Styles assuming a fixed-height main region (`height: 100%` on direct children of `<AppLayout.Main>`) must use `min-h-dvh` or drop the constraint.

Documented imports (`theme.css`, `styles.css`) continue to work unchanged.

## Trade-offs

- Pure app-shell look via `position: sticky` can flicker on iOS Safari momentum scroll (cosmetic).
- Sticky elements may show a brief re-paint when overlays close (cosmetic).

## Test plan

- [x] `pnpm typecheck:only`
- [x] `pnpm test:unit`
- [x] `pnpm test:sb`
- [x] `pnpm lint`
- [x] `pnpm --filter @marigold/theme-rui build` — confirms `dist/preflight.css` ships raw; `dist/theme.css` paints `body` unscoped; `dist/styles.css` paints `[data-theme="rui"]` scoped with no double-prefix
- [ ] Manual at `/examples/*`:
  - [ ] Page scrolls with the document (no interior scroll container)
  - [ ] Sidebar stays pinned on scroll; header stays pinned
- [ ] Mobile Safari: URL bar collapses on scroll, pull-to-refresh works
- [ ] Open an overlay while scrolled — no 1 px layout shift
- [ ] Docs app (island consumer of `styles.css`): fumadocs pages retain their own background; Marigold demos inside `<div data-theme="rui">` render with Marigold's paint

Jira: [DST-1351](https://reservix.atlassian.net/browse/DST-1351)

[DST-1351]: https://reservix.atlassian.net/browse/DST-1351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/AppLayout/AppLayout.tsx
```diff
@@ -33,7 +33,7 @@ interface AppLayoutComponent {
 const _AppLayout = (({ children, ref, ...props }: AppLayoutProps) => (
   <div
     ref={ref}
-    className="grid h-dvh grid-cols-[auto_1fr] grid-rows-[3.5rem_1fr] overflow-hidden [grid-template-areas:'sidebar_header'_'sidebar_main']"
+    className="grid min-h-dvh grid-cols-[auto_1fr] grid-rows-[3.5rem_1fr] [grid-template-areas:'sidebar_header'_'sidebar_main']"
```
sarahgm, line 36: 🔴 **Visual Regression Tests — not run**

themes/theme-rui/src/ui.css
```diff
@@ -242,7 +242,7 @@
  */
 @utility ui-scrollbar {
   scrollbar-width: auto;
-  scrollbar-color: var(--color-scrollbar) var(--color-scrollbar-track);
+  scrollbar-color: var(--color-scrollbar) transparent;
```
sarahgm: ⚠️ **Scrollbar track token bypassed**

The `ui-scrollbar` utility now hardcodes `transparent` instead of using `var(--color-scrollbar-track)`. The token `--color-scrollbar-track: transparent` still exists in `tokens.css` but is effectively dead — consumers who override it won't see the override applied.

Either remove the token from `tokens.css` or keep using `var(--color-scrollbar-track)` here and on line 258.

packages/components/src/Sidebar/Sidebar.tsx
```diff
@@ -68,7 +68,10 @@ const SidebarBase = ({
       ref={ref}
       aria-label={stringFormatter.format('sidebar')}
       data-state={state}
-      className={cn('[grid-area:sidebar]', classNames.root)}
+      className={cn(
+        classNames.root,
+        'sticky top-0 h-dvh self-start [grid-area:sidebar]'
```
sarahgm: 💡 The `cn()` argument order flipped: previously `cn('[grid-area:sidebar]', classNames.root)`, now `cn(classNames.root, 'sticky top-0 h-dvh self-start [grid-area:sidebar]')`. Since layout classes come last, Tailwind merge will prefer them over anything in `classNames.root` — a theme can no longer override `position`, `top`, `height`, or `align-self` on the desktop sidebar. Is this intentional?

packages/components/src/AppLayout/AppLayout.tsx
sarahgm: I close this, I know its on beta-release :D  My first time trying out this inline review

themes/theme-rui/package.json
```diff
@@ -33,14 +33,22 @@
...
+      "default": "./dist/theme.css"
+    },
+    "./tokens.css": {
+      "style": "./dist/tokens.css",
+      "default": "./dist/tokens.css"
+    },
+    "./preflight.css": {
+      "style": "./dist/preflight.css",
+      "default": "./dist/preflight.css"
+    }
```
sarahgm, line 59: what about the other css files like `ui.css` ?

themes/theme-rui/src/tokens.css
```diff
@@ -0,0 +1,455 @@
...
+  --color-charcoal-400: oklch(0.74 0.006 54);
+  --color-charcoal-500: oklch(0.62 0.007 54);
+  --color-charcoal-600: oklch(0.52 0.008 54);
+  --color-charcoal-700: oklch(0.42 0.008 54);
+  --color-charcoal-800: oklch(0.32 0.008 54);
+  --color-charcoal-900: oklch(0.22 0.008 54);
+  --color-charcoal-950: oklch(0.15 0.008 54);
+
+  /* high-contrast obsidian (warm neutral, hue 54, wider contrast range) */
+  --color-obsidian-50: oklch(0.99 0.002 54);
```
sarahgm: you need to update again, so this is removed already

sarahgm (approved):

PR #5345, opened 2026-04-21
fix(DST-1353): remove width="fit" from Select, ComboBox, and Autocomplete

sarahgm:
## Summary

- Exclude `"fit"` from the `width` prop type on Select, ComboBox, and Autocomplete using `Exclude<WidthProp['width'], 'fit'>`
- Replace all existing `width="fit"` usages with explicit widths or defaults
- The Virtualizer controls item sizing, so the dropdown can never grow wider than the trigger. With `width="fit"` the trigger shrinks to the selected value and the dropdown cuts off longer options. This matches how RAC handles Select (fixed width, dropdown locked to trigger width).

## Test plan

- [ ] Verify Select, ComboBox, and Autocomplete demos render correctly without `width="fit"`
- [ ] Confirm TypeScript errors when trying to pass `width="fit"` to Select, ComboBox, or Autocomplete
- [ ] Pagination page-size Select renders with explicit width

Closes DST-1353


packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -119,7 +119,7 @@ export interface AutocompleteProps
    * Sets the width of the field. You can see allowed tokens here: https://tailwindcss.com/docs/width
    * @default 'full'
    */
-  width?: WidthProp['width'];
+  width?: Exclude<WidthProp['width'], 'fit'>;
```
aromko, line 117: 🔴 **Visual Regression Tests — stale**

The last successful `Visual-Regression-Tests` run was on commit `607d5929`, but the branch has since moved to `ed354a1a` (3 commits later, including a merge from `main`). Stories and demos in this PR affect Select / ComboBox / Autocomplete layout — worth covering the current head before merging.

Please re-run visual regression tests by commenting `/run-chromatic` on this PR.

.changeset/remove-width-fit-from-select.md
```diff
@@ -0,0 +1,7 @@
+---
+'@marigold/components': patch
```
aromko: 🔴 Narrowing a public union type (removing `'fit'` from `width`) is a **breaking API change**: any consumer currently passing `width="fit"` to `Select`, `ComboBox`, or `Autocomplete` will get a TypeScript error on upgrade.

Precedent: [`tricky-times-invite.md`](https://github.com/marigold-ui/marigold/blob/main/.changeset/tricky-times-invite.md) marks an equivalent "remove a previously-accepted prop value" change as `major` with `**Breaking change**`. Either:

- Bump to `major` and add a `**BREAKING CHANGE**` marker to the body, **or**
- Explicitly justify why `patch` is appropriate (e.g. "`width=\"fit\"` was broken at runtime, no valid usage exists").

.changeset/remove-width-fit-from-select.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+fix: remove `width="fit"` from Select, ComboBox, and Autocomplete
```
aromko: ⚠️ Changeset title is missing the DST scope. Recent conventions (`align-breakpoints-rem.md`, `tricky-times-invite.md`) consistently include it.

```suggestion
fix(DST-1353): remove `width="fit"` from Select, ComboBox, and Autocomplete
```

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -119,7 +119,7 @@ export interface AutocompleteProps
    * Sets the width of the field. You can see allowed tokens here: https://tailwindcss.com/docs/width
```
aromko, line 113: 💡 The jsdoc still says "You can see allowed tokens here: https://tailwindcss.com/docs/width" without noting that `"fit"` is now excluded. Consider mentioning the exclusion so it's discoverable from the docstring without having to read the type signature.

Same applies to `Select.tsx` and `ComboBox.tsx` for consistency.

aromko (commented):
Automated review via Claude Code `review-pr` skill. 3 focused comments below.

sebald (approved):

PR #5354, opened 2026-04-23
fix(DST-1354): restore collapsing Table.EditableCell edit trigger

sebald:
## Summary

Reverts the overlay/ring affordance from #5250 (DST-1275) and restores the explicit pencil edit button on `Table.EditableCell`. The ring-based approach did not read as editable in user testing — sighted users did not associate the hover ring with inline editing, and there was no discoverable trigger for keyboard or touch paths.

This PR also drops a stale `eslint-disable` that was carried over from the reverted branch.

## Why revert back to the collapsing-button approach

- **Affordance**: users expect a visible control for inline edit. The pencil is the established idiom in spreadsheets, Airtable, Notion, etc.
- **Keyboard**: the pencil is a focusable `<Button>`, so the cell's edit path is reachable via `Tab` + `Enter`. The ring did not expose a trigger.
- **Touch**: the cell itself remains clickable (whole-cell click target), so touch users can tap anywhere in the cell to open the editor; the pencil is just a visual hint for pointer users.
- **Layout**: the trigger collapses to `w-0 overflow-hidden` at rest and expands on row hover / keyboard focus, so the cell's static layout stays clean and the content truncates consistently.

## Notes on the approach swap

- Enabled editable cells now always `truncate` their content to match the single-line editing controls (`TextField`, `Select`, `NumberField`). Disabled cells behave like a regular `Table.Cell`.
- The right-aligned edge case (content visually shifting on row hover when the button expands) is a known tradeoff of this approach — the alternative would truncate meaningful digits behind the button, which is worse for currency/numeric columns. Considered and accepted.
- `selectionMode` is read from `useTableOptions()` so that text selection inside editable cells stays enabled only when the table is not in selection mode (matches the behavior of plain `Table.Cell`).

## Test plan

- [ ] `pnpm sb` → `Components/Table/Editable Cell` — verify pencil appears on row hover, expands smoothly, stays reachable via keyboard (`Tab` focuses pencil, `Enter` opens popover).
- [ ] Verify clicking anywhere in the cell also opens the popover (touch-path).
- [ ] Verify the Balance column (right-aligned) still works; note the expected layout shift on hover.
- [ ] Verify cells with `disabled` behave like a regular cell (no pencil, no forced truncation, no cursor-pointer).
- [ ] `pnpm test:sb` passes for Table.
- [ ] `pnpm typecheck:only` is clean.

## Jira

DST-1354


packages/components/src/Table/TableEditableCell.tsx
```diff
@@ -210,28 +213,44 @@ export const TableEditableCell = ({
...
       {({ columnIndex }) => (
         <>
-          <TableCellContent
-            columnIndex={columnIndex}
-            alignX={alignX}
-            cellOverflow={cellOverflow}
+          <div
+            className={cn(
+              'group/editable-cell flex items-center',
+              !disabled && 'cursor-pointer'
```
sarahgm, line 223: 🔴 **Visual Regression Tests — not run**

packages/components/src/Table/TableEditableCell.tsx
```diff
@@ -210,28 +213,44 @@ export const TableEditableCell = ({
     </Button>
   );
 
+  const handleCellInteraction = () => {
+    if (!disabled) {
+      setOpen(true);
```
sarahgm: 💡 **Minor: could be inlined**

This is a one-liner used in one place. Could be simplified to `onClick={disabled ? undefined : () => setOpen(true)}`, though the current form is perfectly fine for readability.

themes/theme-rui/src/components/Table.styles.ts
```diff
@@ -91,11 +91,9 @@ export const Table: ThemeComponent<'Table'> = {
   cell: cva({
     base: [
       'px-(--cell-x-padding) py-(--cell-y-padding)',
-      'focus-visible:outline-2 focus-visible:-outline-offset-2 focus-visible:outline-ring/75',
-      'has-[[data-cell-content]:focus-visible]:outline-2 has-[[data-cell-content]:focus-visible]:-outline-offset-2 has-[[data-cell-content]:focus-visible]:outline-ring/75',
+      'focus-visible:outline-2 focus-visible:-outline-offset-2 focus-visible:outline-ring/50',
```
sarahgm: 💡 **Outline opacity change applies to all cells**

The focus-visible outline opacity was changed from `outline-ring/75` to `outline-ring/50` for all table cells, not just editable ones. This is a subtle visual change to the entire Table component. Consider noting this in the changeset description if intentional.

.changeset/editable-cell-trigger-focus-outline.md
```diff
@@ -0,0 +1,7 @@
+---
+'@marigold/components': patch
```
sarahgm: you now have two changesets :D

sebald: Merged the focus-outline notes into the main changeset in 9384714 — down to one changeset for this PR. ✅

packages/components/src/Table/TableEditableCell.tsx
sebald: Inlined in a3f77d2 — the handler is now `onClick={disabled ? undefined : () => setOpen(true)}` directly on the wrapper div. ✅

sarahgm (approved):

PR #5357, opened 2026-04-24
docs([DST-1314]): Full page data pattern

OsamaAbdellateef:
# Description

Adds a full-page data management pattern example at /examples/filter demonstrating real-world venue browsing with API integration and multi-criteria filtering. The example covers:

- A QueryProvider using React Query for server-driven data fetching
- A toolbar with search input and filter controls
- Applied filter chips showing active filters with individual dismiss actions
- A venues table displaying paginated, filtered results

# Test Instructions:

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


docs/app/(examples)/examples/filter/api.ts
```diff
@@ -0,0 +1,35 @@
+import type { Venue } from './types';
+
+export const API_BASE = 'https://69e9d8de15c7e2d51268dcf0.mockapi.io/api';
```
sebald: How does this work? can I see the full data somewhere? I just get a 404 going to that link.

Also since, we would still need to implement search etc. why not use our own venue data here?

docs/app/(examples)/examples/filter/api.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const VENUES_QUERY_KEY = ['venues'] as const;
+
+export interface FetchVenuesParams {
+  search?: string;
+  sortBy?: string;
+  order?: 'asc' | 'desc';
+}
+
+export const fetchVenues = async (
```
sebald: this really should be a `routs.ts`

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,340 @@
+'use client';
+
+import {
+  amenitiesOptions,
+  parkingOptions,
+  venueTypes,
+} from '@/lib/data/venues';
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
```
sebald: leaky abstraction. no good architecture, rather have a file that includes dedicated hooks (e.g. useVenues, useDeleteVenues, ...)

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,340 @@
...
+    if (filter.rating && venue.rating < filter.rating) return false;
+    if (filter.traits && filter.traits.length > 0) {
+      if (!filter.traits.some(t => venue.traits.includes(t))) return false;
+    }
+    return true;
+  });
+
+// Price sort stays client-side — MockAPI cannot sort nested fields (price.to)
+// ---------------
+const sortByPrice = (venues: Venue[], sort: VenueSortDescriptor) => {
```
sebald: leaky abstraction

docs/app/(examples)/examples/filter/utils.tsx
```diff
@@ -104,29 +105,25 @@ export const toDisplayValue = {
...
 // ---------------
 type FilterKeys = keyof typeof defaultFilter;
 
 export const useFilter = () => {
   const [filter, setFilter] = useQueryState(
-    'filter',
-    parseAsJson(urlSchema.parse).withDefault(defaultFilter).withOptions({
-      history: 'push',
-    })
+    'dm-filter',
```
sebald: nit: the `dm-` prefix is residue from when this lived under `/examples/data-management`. Now that the data-management route is gone and `/examples/filter` is the only consumer, consider renaming these query-state keys back to plain `q` / `filter` / `sort` / `page` so the URLs aren't carrying historical context.

.changeset/tasty-ravens-count.md
```diff
@@ -0,0 +1,13 @@
+---
+'@marigold/components': patch
```
sebald: This is a docs-only change — the diff only touches `docs/`. I think only `@marigold/docs` should bump here; the entries for `@marigold/components`, `@marigold/theme-rui`, `@marigold/system`, `@marigold/icons`, `@marigold/types`, and `@marigold/eslint-config` look like overreach and would publish empty patch releases for those packages.

docs/app/(examples)/examples/filter/utils.tsx
```diff
@@ -135,4 +132,36 @@ export const useFilter = () => {
...
+
+// Pagination hook
+// ---------------
+export const PAGE_SIZE = 5;
+
+export const usePage = () =>
+  useQueryState(
+    'dm-page',
+    parseAsInteger.withDefault(1).withOptions({ history: 'push' })
+  );
```
sebald: **Open question — probably out of scope for [DST-1314](https://reservix.atlassian.net/browse/DST-1314):** these four hooks feel non-idiomatic for `react-query` and we usually use `nuqs` for state management here. Two issues stand out:

1. **API inconsistency.** `useSearch` uses the legacy options-only form `useQueryState('dm-q', { defaultValue, history })` while `useFilter`/`useSort`/`usePage` use the parser-builder form. Mixing both styles in the same file is confusing.
2. **`parseAsJson` for `filter` and `sort` produces opaque URLs.** `?dm-sort=%7B%22column%22%3A%22name%22%2C%22direction%22%3A%22ascending%22%7D` instead of `?column=name&direction=ascending`. `nuqs` is built around per-field parsers + `useQueryStates` — that's how you get human-readable, shareable URLs.

**Reference implementation we already have:** the **organizer search app** in our internal `core` repo is a clean precedent. Its hooks directory contains `useFilter`, `useSorting`, `usePagination`, and `useSearch` — per-field parsers via `useQueryStates` for filter, `parseAsStringLiteral` for sort column + direction, and a thin `useQueryState` wrapper for search.

**Sketch of the idiomatic shape:**

```ts
import {
  parseAsArrayOf, parseAsInteger, parseAsString,
  parseAsStringLiteral, useQueryState, useQueryStates,
} from 'nuqs';

const history = { history: 'push' } as const;

export const useSearch = () =>
  useQueryState('q', parseAsString.withDefault('').withOptions(history));

const filterParsers = {
  type: parseAsInteger,
  capacity: parseAsInteger,
  price: parseAsInteger,
  traits: parseAsArrayOf(parseAsString).withDefault([]),
  rating: parseAsInteger,
};
export const useFilter = () => {
  const [filter, setFilter] = useQueryStates(filterParsers, history);
  const removeFilter = (keys: Set<keyof typeof filterParsers>) =>
    setFilter(Object.fromEntries([...keys].map(k => [k, null])));
  return { filter, setFilter, removeFilter } as const;
};

export const useSort = () =>
  useQueryStates({
    column: parseAsStringLiteral(['name','type','capacity','price','rating']).withDefault('name'),
    direction: parseAsStringLiteral(['ascending','descending']).withDefault('ascending'),
  }, history);

export const usePage = () =>
  useQueryState('page', parseAsInteger.withDefault(1).withOptions(history));
```

This also makes the `formSchema`/`urlSchema` Zod transforms much smaller (no JSON parsing, parsers do the type coercion).

My read: this is bigger than the ticket and probably belongs in a follow-up (something like *"polish nuqs query-state hooks in `/examples/filter`"*) — not a blocker for shipping the data-management pattern. WDYT?

docs/app/(examples)/examples/filter/utils.tsx
sebald: This is a bad storage key. not idomatic useQuery.

docs/app/(examples)/examples/filter/types.ts
```diff
@@ -0,0 +1,20 @@
+export interface Venue {
```
sebald: why do we need this? isn't it the same as in our venue data?

docs/app/(examples)/examples/filter/toolbar.tsx
```diff
@@ -1,152 +1,195 @@
...
+  traitOptions: string[];
+}
 
-  const submit = (e: FormEvent<HTMLFormElement>) => {
-    e.preventDefault();
-    const data = new FormData(e.currentTarget);
-    setSearch(data.get('q') as string);
-  };
+const FilterForm = ({ state, traitOptions }: FilterFormProps) => {
+  const [rating, setRating] = useState<Set<string>>(
```
sebald: this should not be necessary. bad to have state management here

docs/app/(examples)/examples/filter/toolbar.tsx
```diff
@@ -1,152 +1,195 @@
...
-// Component
+      {/* Hidden input so ToggleButton.Group value lands in FormData */}
+      <input
+        type="hidden"
+        name="rating"
+        value={[...rating][0]?.toString() ?? ''}
+      />
+      <Stack space={2}>
+        <Text weight="semibold">Min. Rating</Text>
+        <ToggleButton.Group
```
sebald: NOT a form component

docs/app/(examples)/examples/filter/query-provider.tsx
```diff
@@ -0,0 +1,20 @@
+'use client';
+
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { useState } from 'react';
+import type { PropsWithChildren } from 'react';
+
+export const QueryProvider = ({ children }: PropsWithChildren) => {
+  const [queryClient] = useState(
```
sebald: why is there a use state? there is no change?

sebald (changes requested):

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,340 @@
...
+        sortBy: currentSort.column,
+        order: serverOrder,
+      }),
+  });
+
+  const deleteMutation = useMutation({
+    mutationFn: (ids: string[]) => Promise.all(ids.map(id => deleteVenue(id))),
+    onSuccess: () =>
+      queryClient.invalidateQueries({ queryKey: VENUES_QUERY_KEY }),
+  });
```
sebald: **Open question — likely out of scope for [DST-1314](https://reservix.atlassian.net/browse/DST-1314):** the React Query plumbing here lives directly in the view component. That makes the example harder to read top-to-bottom (340 lines mixing URL state, API param derivation, query/cache keys, optimistic logic, *and* rendering) and means a few important pattern bits are missing.

**Reference:** the **organizer search app** in our internal `core` repo is the precedent we already have, and its `useOrganizer` hook (in the app's `hooks/` directory) is a much cleaner shape:

- `useOrganizerParams()` (private) — single place that translates URL state (search + filter + sort + page) into an API params object.
- `getOrganizersQueryKey(params)` — single source of truth for cache keys (used by both the query and invalidation).
- `useOrganizers()` — domain hook returning `useQuery` result. View consumes `{ data, isLoading, error, refetch }`.
- `useDeleteOrganizer()` — bundles **confirmation dialog** (`useConfirmation`), **optimistic update** with `onMutate`/`onError`/`onSettled` rollback, **error toast**, and key invalidation. Component just calls `mutate(id)`.

**Things this version is missing that the reference has:**

1. **Optimistic delete** — currently `onSuccess: invalidateQueries(...)` only, so the user clicks Delete and waits for a round-trip before rows disappear. For a docs *pattern* example, this is the wrong thing to copy.
2. **Confirmation dialog** — bulk-delete with no confirmation isn't what the example should teach.
3. **Error toast on mutation failure** — silent failure today.
4. **Inline query key construction** at line 146-149 — the array literal `[...VENUES_QUERY_KEY, { search, sortBy, order }]` should live in a `getVenuesQueryKey(params)` helper so the query and any future invalidations can't drift.
5. **Client-side compensation for backend limits** (`filterVenues`, `sortByPrice` at lines 38-55, applied at lines 165-169) is performed in the view. The reference keeps that inside the data hook so the component never sees the seams. The `// MockAPI cannot sort nested fields` comment at line 50 is exactly the kind of thing that belongs encapsulated.

**Sketch of the split:**

```ts
// useVenues.ts
const useVenuesParams = () => {
  const [search] = useSearch();
  const { filter } = useFilter();
  const [sort] = useSort();
  const [page] = usePage();
  return { search, filter, sort, page };
};

const getVenuesQueryKey = (params: ReturnType<typeof useVenuesParams>) =>
  ['venues', params] as const;

export const useVenues = () => {
  const params = useVenuesParams();
  return useQuery({
    queryKey: getVenuesQueryKey(params),
    queryFn: () => fetchVenues(toApiParams(params)),
    select: data => applyClientSideCompensation(data, params),
  });
};

export const useDeleteVenues = () => {
  const params = useVenuesParams();
  const queryKey = getVenuesQueryKey(params);
  const queryClient = useQueryClient();
  const confirm = useConfirmation();
  const toast = useToast();

  const mutation = useMutation({
    mutationFn: (ids: string[]) => Promise.all(ids.map(deleteVenue)),
    onMutate: async (ids) => {
      await queryClient.cancelQueries({ queryKey });
      const prev = queryClient.getQueryData<Venue[]>(queryKey);
      queryClient.setQueryData<Venue[]>(queryKey, old =>
        old?.filter(v => !ids.includes(v.id))
      );
      return { prev };
    },
    onError: (_err, _ids, ctx) => {
      queryClient.setQueryData(queryKey, ctx?.prev);
      toast.addToast({ variant: 'error', title: 'Failed to delete venues' });
    },
    onSettled: () => queryClient.invalidateQueries({ queryKey }),
  });

  return {
    ...mutation,
    mutate: async (ids: string[]) => {
      const ok = await confirm({
        variant: 'destructive',
        title: ids.length === 1 ? 'Delete venue' : `Delete ${ids.length} venues`,
        confirmationLabel: 'Delete',
      });
      if (ok === 'confirmed') mutation.mutate(ids);
    },
  };
};
```

The view then collapses to: `const { data, isLoading } = useVenues(); const deleteVenues = useDeleteVenues();` and gets to focus on table layout.

This is bigger than the ticket — probably belongs in a follow-up like *"polish data-management example: extract domain hooks, optimistic delete, confirmation"*. Not a blocker for shipping, but worth flagging because the example will be used as a reference. WDYT?

.changeset/tasty-ravens-count.md
```diff
@@ -0,0 +1,13 @@
...
+'@marigold/theme-rui': patch
+'@marigold/system': patch
+'@marigold/icons': patch
+'@marigold/types': patch
+'@marigold/eslint-config': patch
+'@marigold/docs': patch
```
aromko, line 2: 🔴 **Critical — Changeset bumps unrelated packages**

This PR only touches files under `docs/`, but the changeset declares `patch` bumps for `@marigold/components`, `@marigold/theme-rui`, `@marigold/system`, `@marigold/icons`, `@marigold/types`, and `@marigold/eslint-config`. None of those packages have any source changes here, so this would publish empty patch releases of all of them and pollute the changelog.

```suggestion
---
'@marigold/docs': patch
---
```

.changeset/tasty-ravens-count.md
```diff
@@ -0,0 +1,13 @@
...
+---
+
+docs(DST-1314): add full data management pattern with filtering:
+
+Add a full-page data management pattern example at `/examples/filter` demonstrating real-world venue browsing with API integration and multi-criteria filtering. The example includes a query provider using React Query, a toolbar with search and filter controls, applied filter chips for active filter display, and a data table with server-driven results. Also adds an inventory page at `/examples/inventory` showcasing multiple Marigold components in realistic Panel-based layouts, and an event form at `/examples/form` split across multiple Panel sections.
```
aromko: 🔴 **Critical — Description doesn't match the diff**

The body claims *"Also adds an inventory page at `/examples/inventory` showcasing multiple Marigold components in realistic Panel-based layouts, and an event form at `/examples/form` split across multiple Panel sections."* — but neither inventory nor form pages exist in this PR. Reads as if it was copied from a multi-PR description; please trim to what this PR actually does.

Also, the team format is `<type>([DST-XX]): <description>`. This file uses `docs(DST-1314):` (no brackets) while the PR title uses `docs([DST-1314]):` — pick one and apply consistently.

docs/app/(examples)/examples/filter/api.ts
aromko: 🔴 **Critical — Hard-coded third-party MockAPI endpoint**

This URL is published to the public docs site. If the MockAPI account expires, gets throttled, or the host goes down, the example breaks for every visitor and can't be debugged from the repo.

Prefer a Next.js Route Handler (e.g. `docs/app/api/venues/route.ts`) backed by the existing static `@/lib/data/venues` array. That keeps the React Query / server-driven story while making the example self-contained and reproducible.

docs/app/(examples)/examples/filter/page.tsx
```diff
@@ -1,30 +1,41 @@
...
-import { Headline, Inset, Panel, Stack, Text } from '@marigold/components';
+import {
+  Button,
+  Headline,
+  Inset,
+  Panel,
+  Stack,
+  Text,
+} from '@marigold/components';
+import { Add } from '@marigold/icons';
```
aromko, line 12: 🔴 **Critical — Unused imports**

Neither `Button` (line 4) nor `Add` (line 11) is referenced anywhere in this file. They look like leftovers from when the "Add Venue" button lived here before being moved into `Toolbar`.

```suggestion
import {
  Headline,
  Inset,
  Panel,
  Stack,
  Text,
} from '@marigold/components';
import { AppliedFilter } from './applied-filter';
import { QueryProvider } from './query-provider';
import { Toolbar } from './toolbar';
import { VenuesTable } from './venues-table';
```

docs/app/(examples)/examples/filter/toolbar.tsx
```diff
@@ -158,6 +201,16 @@ export const Toolbar = () => {
           </Form>
         </Drawer>
       </Drawer.Trigger>
+      <div className="ml-auto">
+        <Button
+          variant="primary"
+          onPress={() => {
+            /* TODO: open Add Venue dialog (DST-1288) */
```
aromko: 🔴 **Critical — Non-functional primary CTA shipped in a published example**

A primary "Add Venue" button that does nothing is misleading for users of the design system who copy this example. Options:

1. Remove the button until DST-1288 lands
2. Wire it to a placeholder dialog with a *"coming soon (DST-1288)"* message
3. Visibly disable it with `disabled` and a tooltip

A shipped TODO with no user-visible feedback is the worst of the three options.

docs/app/(examples)/examples/filter/toolbar.tsx
```diff
@@ -158,6 +201,16 @@ export const Toolbar = () => {
           </Form>
         </Drawer>
       </Drawer.Trigger>
+      <div className="ml-auto">
```
aromko: ⚠️ **Raw Tailwind className wrapper**

CLAUDE.md says *"Don't write inline styles or use `className` prop directly"*. Even though this is example code, the docs are the canonical reference consumers copy from. Express the layout with Marigold primitives — e.g. split into two `Inline` blocks, or use `alignX="space-between"` on the parent.

docs/app/(examples)/examples/filter/applied-filter.tsx
```diff
@@ -27,7 +26,7 @@ export const AppliedFilter = () => {
     >
       {appliedFilters.map(([name, value]) => (
         <Tag id={name} key={name}>
-          {toDisplayValue[name as FilterKeys](value as any)}
+          {toDisplayValue[name as FilterKeys](value as never)}
```
aromko: ⚠️ **`as never` cast replaces `as any`**

Switching `as any` → `as never` silences the diagnostic without addressing the underlying typing problem (heterogeneous formatter-by-key). Consider a typed `Record<FilterKeys, (value: ...) => ReactNode>` keyed lookup with per-key narrowing, so future readers don't propagate the cast.

docs/app/(examples)/examples/filter/utils.tsx
```diff
@@ -104,29 +105,25 @@ export const toDisplayValue = {
...
+    parseAsJson(urlSchema.parse)
+      .withDefault(defaultFilter)
+      .withOptions({ history: 'push' })
   );
 
   const removeFilter = (keys: Set<FilterKeys>) => {
     const next = { ...filter };
     keys.forEach(key => {
-      next[key] = defaultFilter[key] as any;
+      next[key] = defaultFilter[key] as never;
```
aromko: ⚠️ **`as never` cast**

Same concern as `applied-filter.tsx:29` — silences the diagnostic without fixing the underlying type. Consider a discriminated union or per-key handler that lets each value flow through with its actual type.

docs/app/(examples)/examples/filter/toolbar.tsx
```diff
@@ -1,152 +1,195 @@
...
+        </Tag.Group>
+      )}
 
-// Component
+      {/* Hidden input so ToggleButton.Group value lands in FormData */}
+      <input
+        type="hidden"
+        name="rating"
+        value={[...rating][0]?.toString() ?? ''}
+      />
```
aromko: ⚠️ **Hidden input bridge for `ToggleButton.Group`**

The pattern works but consumers will cargo-cult it. If `ToggleButton.Group` lacks form integration, the docs example shouldn't quietly encode that gap — at minimum add a comment explaining the workaround and link to the underlying issue (or component limitation), so readers know not to copy this pattern blindly into other forms.

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,340 @@
...
+  // Apply client-side filters, then price sort if needed (server already sorted other columns)
+  const filtered = filterVenues(venues, filter);
+  const display =
+    currentSort.column === 'price'
+      ? sortByPrice(filtered, currentSort)
+      : filtered;
+
+  const totalItems = display.length;
+  const totalPages = Math.ceil(totalItems / PAGE_SIZE);
+  const safePage = Math.min(page, Math.max(1, totalPages));
```
aromko: ⚠️ **`safePage` clamps locally but URL state isn't reset**

`Math.min(page, Math.max(1, totalPages))` clamps the local value but the `dm-page=99` query string survives even if there are only 5 pages. URL pasting / browser history can produce stale page state.

Add an effect:
```ts
useEffect(() => {
  if (page > Math.max(1, totalPages)) setPage(null);
}, [page, totalPages, setPage]);
```

docs/app/(examples)/examples/filter/venues-table.tsx
aromko: ⚠️ **`deleteMutation` has no `onError` handler**

A failed delete is silently swallowed. For an example demonstrating *"real-world venue browsing"*, showing an error toast / inline message is the lesson worth teaching:

```ts
const deleteMutation = useMutation({
  mutationFn: ...,
  onSuccess: () => queryClient.invalidateQueries({ queryKey: VENUES_QUERY_KEY }),
  onError: (err) => {
    // surface to user — toast, banner, etc.
  },
});
```

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,340 @@
...
+  const paged = display.slice((safePage - 1) * PAGE_SIZE, safePage * PAGE_SIZE);
```
aromko: ⚠️ **Derived values rebuild every render**

`filtered`, `display`, and `paged` (and `getSelectedVenues` further down, plus the `actionBar` render-prop closure on line 228) are all recreated on every render. Datasets are small, so it's fine functionally — but as a published example others copy, wrapping them in `useMemo` / `useCallback` sets a better baseline:

```ts
const filtered = useMemo(() => filterVenues(venues, filter), [venues, filter]);
const display = useMemo(
  () => (currentSort.column === 'price' ? sortByPrice(filtered, currentSort) : filtered),
  [filtered, currentSort]
);
```

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,340 @@
...
+import {
+  PAGE_SIZE,
+  defaultFilter,
+  useFilter,
+  usePage,
+  useSearch,
+  useSort,
+} from './utils';
+
+// Client-side filter — search and sort are handled server-side
```
aromko: 💡 **Misleading file comment**

The header says *"search and sort are handled server-side"*, but `filterVenues` ignores `search` (server handles it) **and** `sortByPrice` is client-side because MockAPI cannot sort nested fields. The split is fine, but the comment is misleading. Consider rewording to something like:

> Most filters are applied client-side over the server-fetched result set. Search and column sorting are passed to the API; price sort stays client-side because MockAPI cannot sort nested fields (`price.to`).

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,340 @@
...
+    'Rating',
+  ];
+  const rows = selected.map(v => [
+    `"${v.name.replace(/"/g, '""')}"`,
+    venueTypes[v.type] ?? 'Unknown',
+    v.city,
+    v.country,
+    v.capacity,
+    v.price.to,
+    v.rating,
```
aromko, line 89: 💡 **CSV escaping only on `name`**

`city`, `country`, `venueTypes[v.type]`, etc. could contain commas or quotes and would corrupt the CSV. Either quote-escape every string field with the same helper, or note the limitation in a comment so consumers don't ship this verbatim:

```ts
const csvCell = (v: unknown) => {
  const s = String(v ?? '');
  return /[",\n]/.test(s) ? `"${s.replace(/"/g, '""')}"` : s;
};
```

docs/app/(examples)/examples/filter/utils.tsx
```diff
@@ -1,10 +1,13 @@
...
-import { parseAsJson, useQueryState } from 'nuqs';
+import { venueTypes } from '@/lib/data/venues';
+import { parseAsInteger, parseAsJson, useQueryState } from 'nuqs';
 import type { FormEvent } from 'react';
 import { z } from 'zod';
 import { NumericFormat } from '@marigold/system';
 
-// Handling form data
+export const MAX_CAPACITY = 50_000;
+export const MAX_PRICE = 5_000;
```
aromko, line 17: 💡 **Constants disconnected from data source**

Previously these were derived from the venues array. Switching to constants risks the slider / number-field range falling out of sync with the API. Either compute from the cached query (`useQuery(VENUES_QUERY_KEY)`) or document the rationale in a comment.

aromko (changes requested):
## Automated Code Review

Great work assembling the full DST-1314 pattern — the React Query integration, drawer-based filter form, applied-filter chips, sort/pagination, action bar, and empty states all match the ticket's scope. The blocking concerns below are mostly publishing-side (changeset + dependency on a third-party host) plus a few leftover fragments from the refactor.

**One general note that can't be inlined:** the PR title says *"Full page **date** management pattern"* — should be *"**data** management pattern"*. Please fix on merge so the squashed commit and changelog are correct.

VRT is **not needed** for this PR — the changes are scoped to `docs/app/...` and don't touch `packages/`/`themes/`/stories.

---

docs/app/(examples)/examples/filter/api.ts
OsamaAbdellateef: Replaced MockAPI with a Next.js Route Handler at docs/app/api/venues/route.ts backed by @/lib/data/venues.

docs/app/(examples)/examples/filter/toolbar.tsx
OsamaAbdellateef: Added ticket for it 
https://reservix.atlassian.net/browse/DST-1392

.claude/settings.local.json
```diff
@@ -0,0 +1,17 @@
+{
```
aromko: 🔴 **Critical: do not commit this file**

This is a user-local Claude Code permissions file with hardcoded Windows paths (`C:\\Users\\osama\\...`) and machine-specific PowerShell commands. It should not be in the repo.

Please remove the file from this PR. As a follow-up, recommend adding `.claude/settings.local.json` to the project `.gitignore` — it currently isn't ignored at the project level (macOS contributors get it via their global ignore, which is why this slipped through CI but local Windows checkouts didn't).

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,324 @@
...
+  const safePage = Math.min(page, Math.max(1, totalPages));
+  const paged = useMemo(
+    () => display.slice((safePage - 1) * PAGE_SIZE, safePage * PAGE_SIZE),
+    [display, safePage]
+  );
+
+  // Keep the URL in sync when the requested page is out of range so a stale
+  // ?page=99 query string doesn't survive after filtering shrinks the result.
+  useEffect(() => {
+    if (page !== safePage) setPage(safePage === 1 ? null : safePage);
```
aromko: ⚠️ **Page-correction effect pushes a history entry**

`usePage` is configured with `history: 'push'`. When filtering shrinks the result set below the current page, this effect pushes a new history entry — the user's Back button now sends them back to the same view (history trap).

For URL-correction like this, prefer `history: 'replace'` (either as the default for `usePage`, or as a per-call override on this specific `setPage`).

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,324 @@
...
+        onSortChange={d =>
+          setSort({
+            column: String(d.column),
+            direction: d.direction,
+          })
+        }
+        actionBar={selectedKeys => (
+          <ActionBar>
+            <ActionBar.Button>
+              <Delete /> Delete
```
aromko: ⚠️ **Delete button has no `onPress`**

Non-functional button in a published docs example. Either disable it (mirror the `Add Venue` pattern with a `// Will be wired up in DST-XXXX` comment) or wire it to a stub.

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,324 @@
...
+  </div>
+);
+
+// Active filter check
+// ---------------
+const hasActiveFilters = (search: string, filter: VenueFilter) => {
+  if (search.length > 0) return true;
+  return Object.entries(filter).some(([name, value]) => {
+    if (name === 'traits') return Array.isArray(value) && value.length > 0;
+    return `${value}` !== `${defaultFilter[name as keyof VenueFilter]}`;
```
aromko: ⚠️ **Stringification comparison is opaque**

Since `defaultFilter` values are now all `undefined`, this branch reduces to `value !== undefined`. The template-string round-trip is hard to read and has subtle gotchas (e.g. `${null}` → `'null'`).

```suggestion
    return value !== defaultFilter[name as keyof VenueFilter];
```

docs/app/(examples)/examples/filter/utils.tsx
```diff
@@ -35,104 +42,145 @@ const urlSchema = z.object({
...
+  traits: data.traits ?? [],
   rating: String(data.rating ?? ''),
 })).parse;
 
-// Form -> URL
 export const toUrlSchema = formSchema.transform(data => ({
-  type: data.type ? Number(data.type) : undefined,
   capacity: data.capacity ? Number(data.capacity) : undefined,
-  price: data.price ? Number(data.price) : undefined,
+  price: Number(data.price) < MAX_PRICE ? Number(data.price) : undefined,
```
aromko: ⚠️ **`Number(data.price)` evaluated twice**

Compute once, then compare. Side note: `Number('')` is `0`, so an empty/missing slider value silently becomes `price: 0` rather than `undefined`. It still works because of the `if (filter.price)` guard in `filterVenues`, but the intent reads more clearly if 0 → undefined explicitly.

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,324 @@
...
+                </Inline>
+              </Table.Cell>
+            </Table.Row>
+          ))}
+        </Table.Body>
+      </Table>
+
+      {totalPages > 1 && (
+        <Inline alignX="center">
+          <Pagination
```
aromko: 💡 **Pagination `key` over-specifies**

`PAGE_SIZE` is a constant — `key={totalPages}` is sufficient.

docs/app/(examples)/examples/filter/toolbar.tsx
```diff
@@ -1,163 +1,188 @@
...
+        ref={ratingInputRef}
+        type="hidden"
+        name="rating"
+        defaultValue={state.rating}
+      />
+      <Stack space={2}>
+        <Text weight="semibold">Min. Rating</Text>
+        <ToggleButton.Group
+          selectionMode="single"
+          defaultSelectedKeys={state.rating ? new Set([state.rating]) : []}
```
aromko: 💡 **Inconsistent empty-selection initializer**

Uses `[]` for the empty case but `new Set([state.rating])` for the populated case. Use `new Set()` for both to keep the type stable:

```suggestion
          defaultSelectedKeys={state.rating ? new Set([state.rating]) : new Set()}
```

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,324 @@
...
+  const isFiltered = hasActiveFilters(search, filter);
+  const emptyState: () => ReactNode = isFiltered
+    ? () => <FilterEmptyState onClear={clearFilters} />
+    : () => <NoDataEmptyState />;
+
+  const getSelectedVenues = useCallback(
+    (selectedKeys: Selection) =>
+      selectedKeys === 'all'
+        ? display
+        : display.filter(v => (selectedKeys as Set<string>).has(v.id)),
```
aromko, line 184: 💡 **`as Set<string>` cast on `Selection`**

After the `=== 'all'` check, TS still widens to `Selection`. A small narrowing helper or destructuring avoids the cast — minor.

aromko (changes requested):
Linked to **DST-1314** (In Review). All CI checks pass; VRT not needed (docs example only).

**Note:** PR title says "**date** management pattern" — should be "**data**". The changeset is already correct.

---

aromko (approved):

docs/app/(examples)/examples/filter/utils.tsx
```diff
@@ -35,104 +42,150 @@ const urlSchema = z.object({
...
 export const useFilter = () => {
   const [filter, setFilter] = useQueryState(
     'filter',
-    parseAsJson(urlSchema.parse).withDefault(defaultFilter).withOptions({
-      history: 'push',
-    })
+    parseAsJson(urlSchema.parse)
+      .withDefault(defaultFilter)
+      .withOptions({ history: 'push' })
   );
```
sebald: **Re-raising my earlier comment ([#3159373437](https://github.com/marigold-ui/marigold/pull/5357#discussion_r3159373437)) — it was marked resolved but the code wasn't actually changed.**

`parseAsJson(urlSchema.parse)` produces opaque URLs like `?filter=%7B%22capacity%22%3A1000%2C%22price%22%3A2000%7D` instead of `?capacity=1000&price=2000`. For a docs example that will be copy-pasted, this is the wrong pattern to teach — `nuqs` is built around per-field parsers + `useQueryStates`.

```ts
import { parseAsArrayOf, parseAsInteger, parseAsString, useQueryStates } from 'nuqs';

const filterParsers = {
  type: parseAsInteger,
  capacity: parseAsInteger,
  price: parseAsInteger,
  traits: parseAsArrayOf(parseAsString).withDefault([]),
  rating: parseAsInteger,
};

export const useFilter = () => {
  const [filter, setFilter] = useQueryStates(filterParsers, { history: 'push' });
  const removeFilter = (keys: Set<keyof typeof filterParsers>) =>
    setFilter(Object.fromEntries([...keys].map(k => [k, null])));
  return { filter, setFilter, removeFilter } as const;
};
```

Side benefit: the `urlSchema` Zod transforms in `toFormSchema` / `toUrlSchema` get a lot smaller — parsers do the type coercion, no JSON round-trip needed.

docs/app/(examples)/examples/filter/utils.tsx
```diff
@@ -35,104 +42,150 @@ const urlSchema = z.object({
...
+      [...keys].map(key => [key, defaultFilter[key]])
+    );
+    setFilter({ ...filter, ...cleared });
   };
 
   return { filter, setFilter, removeFilter } as const;
 };
 
 export const useSearch = () =>
   useQueryState('q', { defaultValue: '', history: 'push' });
```
sebald: **Same thread as [#3159373437](https://github.com/marigold-ui/marigold/pull/5357#discussion_r3159373437) — API inconsistency still present.**

`useSearch` uses the legacy options-only form (`useQueryState('q', { defaultValue, history })`) while `useFilter`/`useSort`/`usePage` use the parser-builder form. Mixing both styles in the same file is confusing for readers copying the example.

```ts
export const useSearch = () =>
  useQueryState('q', parseAsString.withDefault('').withOptions({ history: 'push' }));
```

docs/app/(examples)/examples/filter/utils.tsx
```diff
@@ -35,104 +42,150 @@ const urlSchema = z.object({
...
+  direction: 'ascending',
+};
+
+export const useSort = () =>
+  useQueryState(
+    'sort',
+    parseAsJson(sortSchema.parse)
+      .withDefault(defaultSort)
+      .withOptions({ history: 'push' })
+  );
```
sebald: **Same thread as [#3159373437](https://github.com/marigold-ui/marigold/pull/5357#discussion_r3159373437) — opaque sort URL not addressed.**

`?sort=%7B%22column%22%3A%22name%22%2C%22direction%22%3A%22ascending%22%7D` is the kind of URL `nuqs` is designed to avoid. Use per-field parsers with `parseAsStringLiteral` so column/direction are validated and human-readable:

```ts
export const useSort = () =>
  useQueryStates(
    {
      column: parseAsStringLiteral(['name', 'capacity', 'price'] as const).withDefault('name'),
      direction: parseAsStringLiteral(['ascending', 'descending'] as const).withDefault('ascending'),
    },
    { history: 'push' }
  );
```

URL becomes `?column=price&direction=descending` — readable, shareable, and invalid values are rejected by the parser instead of blowing up at `sortSchema.parse`.

sebald (commented):
Re-raising the nuqs idiomaticity concerns from my earlier review ([thread #3159373437](https://github.com/marigold-ui/marigold/pull/5357#discussion_r3159373437)) — that thread is marked resolved, but the code in `useFilter`, `useSearch`, and `useSort` still has both issues I flagged (mixed `useQueryState` API styles + opaque `parseAsJson` URLs). Inline comments below pin the specific lines.

Since this is the canonical docs example for the filter pattern, the URLs it produces will be copied — worth getting them right before merge.

docs/app/(examples)/examples/filter/toolbar.tsx
```diff
@@ -1,163 +1,186 @@
...
+              ratingInputRef.current.value = [...keys][0]?.toString() ?? '';
+            }
+          }}
+        >
+          {['1', '2', '3', '4', '5'].map(val => (
+            <ToggleButton key={val} id={val}>
+              {val} ★
+            </ToggleButton>
+          ))}
+        </ToggleButton.Group>
```
sebald: **Re-raising my earlier comment ([#3159383480](https://github.com/marigold-ui/marigold/pull/5357#discussion_r3159383480)) — "NOT a form component". This is a hard no-go.**

Replacing `Radio.Group` with `ToggleButton.Group` + a hidden `<input>` bridge is not the right direction. **`ToggleButton.Group` is not a form element and will never be designed for use inside a `<Form>`** — it is a UI control for toggling state, not for collecting form data. The hidden-input workaround codifies a pattern we explicitly do not want to teach in our canonical filter example, and the DST-1392 reference is misleading: that ticket is not going to "fix" `ToggleButton.Group` to become a form component.

**Please revert this section to the previous `Radio.Group` implementation.** `Radio.Group` is the correct primitive for "pick one of N values inside a form" — it integrates with `FormData`, has built-in label/description support, and matches the pattern used by every other field in this drawer (`NumberField`, `Slider`, `Tag.Group`).

If the visual change to a star-button row is desired for design reasons, that is a separate conversation and belongs in a follow-up — not here, and not via a hidden-input hack.

sebald (commented):
Follow-up on the rating field — see inline comment. The `ToggleButton.Group` + hidden-input bridge needs to be reverted to `Radio.Group`.

docs/app/(examples)/examples/filter/utils.tsx
```diff
@@ -35,104 +42,150 @@ const urlSchema = z.object({
...
+
+// Pagination hook
+// ---------------
+export const PAGE_SIZE = 5;
+
+export const usePage = () =>
+  useQueryState(
+    'page',
+    parseAsInteger.withDefault(1).withOptions({ history: 'push' })
+  );
```
sebald: ## Architecture: align `useFilter` / `useSearch` / `useSort` / `usePage` with the organizer-search-app reference

Stepping back from the per-line comments above — the bigger ask is that this whole hooks block follows the architecture we already have in `apps/organizer-search-app/src/hooks/` (internal `core` repo). That is **the** canonical precedent for nuqs-driven URL state in our codebase, and the docs example is going to be the thing people copy-paste, so it needs to match.

Concretely this means:

1. **Per-field parsers via `useQueryStates`** (no `parseAsJson` round-trip)
2. **Consistent API shape** — every hook uses `useQueryStates` (or `useQueryState` with the parser-builder form), no mixing legacy options-only
3. **Cross-hook coordination** — `setFilter` and `setSearch` reset `page` to 1 (currently this is sprinkled into `Toolbar` and the `useEffect` in `venues-table.tsx`; it belongs in the hooks)
4. **`hasFilter()` helper on `useFilter`** so call sites don't reimplement "is anything active?" (currently done as `hasActiveFilters` in `venues-table.tsx` *and* via `Object.entries(filter).filter(...)` in `applied-filter.tsx`)

### `usePagination` — paste in directly

```ts
import { parseAsInteger, useQueryStates } from 'nuqs';

export const usePagination = () =>
  useQueryStates({
    page: parseAsInteger.withDefault(1),
    pageSize: parseAsInteger.withDefault(PAGE_SIZE),
  });
```

`PAGE_SIZE` becomes the default for the `pageSize` parser, which means the URL also stops carrying it when it matches the default. Replace the `usePage` call sites with `const [{ page, pageSize }, setPagination] = usePagination();`.

### `useSort` — paste in directly, narrow the column literal

```ts
import { parseAsStringLiteral, useQueryStates } from 'nuqs';

export const useSort = () =>
  useQueryStates({
    column: parseAsStringLiteral(['name', 'capacity', 'price'] as const).withDefault('name'),
    direction: parseAsStringLiteral(['ascending', 'descending'] as const).withDefault('ascending'),
  });
```

URL becomes `?column=price&direction=descending` instead of the JSON-encoded blob. The `sortSchema` zod object and the `defaultSort` constant can both be deleted — `parseAsStringLiteral` validates and the `withDefault` covers the default. The `String(d.column)` cast in `venues-table.tsx` also goes away because `column` is now correctly typed.

### `useSearch` — match the reference shape

```ts
import { parseAsString, useQueryState } from 'nuqs';
import { usePagination } from './usePagination';

export const useSearch = () => {
  const [pagination, setPagination] = usePagination();
  const [search, _setSearch] = useQueryState('q', parseAsString.withDefault(''));

  const setSearch = (next: string) => {
    setPagination({ ...pagination, page: 1 });
    _setSearch(next);
  };

  return [search, setSearch] as const;
};
```

This kills the `setPage(null)` calls that are currently scattered across `Toolbar` (`onSearch`, `onClear`).

### `useFilter` — adapted to the venues domain

The reference hook in core is the right shape. Adapted to the URL fields this PR actually uses:

```ts
import { parseAsArrayOf, parseAsInteger, parseAsString, useQueryStates } from 'nuqs';
import { usePagination } from './usePagination';

const filterParsers = {
  type: parseAsInteger,
  capacity: parseAsInteger,
  price: parseAsInteger,
  traits: parseAsArrayOf(parseAsString).withDefault([]),
  rating: parseAsInteger,
};

export type FilterKeys = keyof typeof filterParsers;

const defaultFilter = {
  type: null,
  capacity: null,
  price: null,
  traits: [],
  rating: null,
};

export const useFilter = () => {
  const [pagination, setPagination] = usePagination();
  const [filter, _setFilter] = useQueryStates(filterParsers);

  const setFilter = (next: Partial<typeof filter>) => {
    setPagination({ ...pagination, page: 1 });
    _setFilter(prev => ({ ...prev, ...next }));
  };

  const removeFilter = (keys: Set<FilterKeys>) => {
    _setFilter(Object.fromEntries([...keys].map(k => [k, null])) as Partial<typeof filter>);
  };

  const hasFilter = () =>
    (Object.keys(filterParsers) as FilterKeys[]).some(key => {
      const v = filter[key];
      return Array.isArray(v) ? v.length > 0 : v !== null;
    });

  return { filter, setFilter, removeFilter, hasFilter } as const;
};
```

Notes for adapting to this PR specifically:
- `traits` is already an array — `parseAsArrayOf(parseAsString)` serializes it as `?traits=foo,bar`, no JSON.
- `null` is nuqs's "absent from URL" signal for nullable parsers, which is why `defaultFilter` uses `null` instead of `undefined`. The Zod `urlSchema` and `formSchema`/`toFormSchema`/`toUrlSchema` transforms can be **deleted entirely** — the parsers do the type coercion, and the form's submit handler can build the partial directly:
  ```ts
  const onFilterSubmit = (e: FormEvent<HTMLFormElement>) => {
    e.preventDefault();
    const data = getFormData(e);
    setFilter({
      capacity: data.capacity ? Number(data.capacity) : null,
      price: data.price ? Number(data.price) : null,
      traits: Array.isArray(data.traits) ? data.traits : data.traits ? [data.traits as string] : [],
      rating: data.rating ? Number(data.rating) : null,
    });
  };
  ```
- `applied-filter.tsx` replaces its `Object.keys(filter).filter(...)` block with `hasFilter()` for the conditional, and the per-key checks in `renderFilterValue` keep working (just compare against `null` instead of `undefined`).
- `venues-table.tsx`'s `hasActiveFilters` helper can be replaced with `useFilter().hasFilter() || search.length > 0`.

### Reference

The four files in `apps/organizer-search-app/src/hooks/` are tiny — `usePagination.ts` is 8 lines, `useSorting.ts` is 14, `useSearch.ts` is 18, `useFilter.ts` is 60. Worth opening them side by side while doing this refactor; the shapes above mirror them 1:1 except for the venues-specific field list.

This is the architectural ask I had in my first review and on [thread #3159373437](https://github.com/marigold-ui/marigold/pull/5357#discussion_r3159373437) — please don't resolve it without actually applying the change. The docs example needs to *be* the reference, not the counter-example.

sebald (commented):
Architectural follow-up on the URL-state hooks — see inline comment on `utils.tsx`. The shape we want is the one already established in the organizer-search-app (`apps/organizer-search-app/src/hooks/` in `core`); the inline comment includes paste-ready code for `usePagination` and `useSort`, plus a venues-adapted `useFilter` and `useSearch`.

sebald:
## Architecture & maintainability follow-up

Beyond the hook-shape comment on `utils.tsx`, here are concrete improvements that would lift this from "a working example" to "the canonical reference we want consumers to copy". All of these are achievable inside this PR's scope and remove code rather than add it.

---

### 1. Use `parseFormData` from `@marigold/components` for the filter submit

`@marigold/components` ships a `parseFormData` helper that takes a `FormEvent<HTMLFormElement>` and returns a typed object keyed by field name, with arrays handled correctly for multi-value fields (`Tag.Group selectionMode="multiple"`, multi-select, etc.).

Once `useFilter` switches to per-field nuqs parsers (see prior comment), the parsers handle type coercion, which means the entire form↔URL transform layer in `utils.tsx` collapses:

- `getFormData` (the manual `FormData` aggregator) → **delete**, replaced by `parseFormData(e)`
- `formSchema` Zod object → **delete**, parsers validate
- `toFormSchema` URL→form transform → **delete**, hook returns the right shape
- `toUrlSchema` form→URL transform with `safeParse` → **delete**, parsers coerce on the way in
- The `if (!success) console.error(...)` error branch → **delete**, no schema to fail

`Toolbar`'s `onFilterSubmit` reduces to:

```tsx
const onFilterSubmit = (e: FormEvent<HTMLFormElement>) => {
  e.preventDefault();
  const data = parseFormData(e);
  setFilter({
    capacity: data.capacity ? Number(data.capacity) : null,
    price: data.price ? Number(data.price) : null,
    traits: Array.isArray(data.traits) ? data.traits : data.traits ? [data.traits] : [],
    rating: data.rating ? Number(data.rating) : null,
  });
};
```

Net change: ~80 lines deleted from `utils.tsx`, one helper import in `toolbar.tsx`, no behavior change. This is the single biggest readability win available in this PR.

---

### 2. `<NuqsAdapter defaultOptions={{ clearOnDefault: true }}>`

Wrap the example in a `NuqsAdapter` with `clearOnDefault: true`. With this option, query params automatically drop out of the URL when their value matches the parser's default — no more manual `setPage(null)` sprinkled across `Toolbar` (`onSearch`, `onClear`, `onFilterSubmit`) and the `useEffect` page-correction in `venues-table.tsx`.

```tsx
import { NuqsAdapter } from 'nuqs/adapters/next/app';

<NuqsAdapter defaultOptions={{ clearOnDefault: true }}>
  {/* example */}
</NuqsAdapter>
```

Concrete cleanups this enables:
- `setPage(null)` calls become `setPage(1)` — and since `1` is the parser default, the URL stays clean automatically. Even cleaner: don't call `setPage` from search/filter handlers at all once it's part of `useFilter`/`useSearch` (per the hook architecture comment).
- The page-correction effect in `venues-table.tsx` (the `useEffect` that resets a stale `?page=99`) keeps the `setPage(safePage === 1 ? null : safePage)` shape but no longer needs the `=== 1 ? null` branch — every defaulted value clears itself.
- The whole concept of "remember to use `null` to clear" stops being a thing the reader has to learn.

The PR currently relies on whatever `NuqsAdapter` the docs app provides at the layout level (so nuqs works at all), but the example never makes its URL-state setup visible. Making it explicit is also a teachable detail.

---

### 3. Extract a dedicated `Provider` component

The example needs three providers to be a complete data-management reference:

```tsx
// query-provider.tsx (rename to provider.tsx)
import { NuqsAdapter } from 'nuqs/adapters/next/app';
import { ConfirmationProvider, ToastProvider } from '@marigold/components';

export const Provider = ({ children }) => (
  <ConfirmationProvider>
    <NuqsAdapter defaultOptions={{ clearOnDefault: true }}>
      {children}
      <ToastProvider />
    </NuqsAdapter>
  </ConfirmationProvider>
);
```

Why all three:
- `NuqsAdapter` — covered above, makes the URL-state setup explicit.
- `ConfirmationProvider` — the example currently has a `Delete` action bar button stubbed with `alert('Delete element')`. A docs example showing how to delete from a table without a confirmation dialog is teaching the wrong pattern. With `ConfirmationProvider` present, `useConfirmation()` becomes available and the destructive flow can be wired correctly.
- `ToastProvider` — every "delete failed", "export complete" surface in the example needs somewhere to render. Currently there's nowhere.

Wrapping `page.tsx`'s tree in `<Provider>` makes the dependencies of the example self-documenting.

---

### 4. Extract pagination into its own component

The pagination block at the bottom of `venues-table.tsx` (currently 9 lines inside the table component) should live in `venues-pagination.tsx`. Three concrete UX details that go missing in the inline version:

**a. Page-size selector.** `PAGE_SIZE = 5` is hardcoded as a constant. A real data-management example should let the user change it. Add a `Select` next to the pagination control:

```tsx
<Select
  aria-labelledby="pageSize"
  width={20}
  selectedKey={pageSize}
  onChange={key => setPagination({ ...pagination, pageSize: Number(key) })}
>
  <Select.Option id={5}>5</Select.Option>
  <Select.Option id={10}>10</Select.Option>
  <Select.Option id={25}>25</Select.Option>
</Select>
```

This also justifies why `pageSize` belongs in the URL state (under `usePagination`), not as a constant.

**b. "Showing X – Y of Z" helper text.** Standard pattern for paginated tables, missing today:

```tsx
const from = (page - 1) * pageSize + 1;
const to = Math.min(page * pageSize, totalItems);
// → "Showing 1 – 5 of 23"
```

**c. Scroll-into-view on page change.** When the user paginates, the new rows are above their viewport. Pass a ref to the table region and call `scrollIntoView` on page change:

```tsx
const updatePage = (page: number) => {
  setPagination({ ...pagination, page });
  ref?.current?.scrollIntoView({ behavior: 'smooth' });
};
```

Layout: avoid raw `className="grid grid-cols-..."` — use `<Inline alignX="between">` with three slots (text on left, `<Pagination>` centered, page-size select on right) to stay consistent with the rest of the docs.

---

### 5. Empty-state components consume hooks directly

`FilterEmptyState` currently takes `onClear` as a prop, threaded down from `venues-table.tsx` where `clearFilters` is constructed. This is one degree of indirection too many — the empty state knows it needs to clear search + filter + page, so let it call those hooks itself:

```tsx
const FilterEmptyState = () => {
  const [, setSearch] = useSearch();
  const { filter, removeFilter } = useFilter();

  const clearAll = () => {
    setSearch('');
    removeFilter(new Set(Object.keys(filter) as FilterKeys[]));
  };

  return (
    <Center>
      <Text fontSize="lg" weight="bold">No venues match your criteria</Text>
      <Text variant="muted">Try adjusting your filters or search terms.</Text>
      <Button variant="ghost" onPress={clearAll}>Clear all filters</Button>
    </Center>
  );
};
```

`venues-table.tsx` no longer constructs `clearFilters` or passes `onClear` — the empty state is fully self-contained. The `hasActiveFilters` helper can also move into the empty-state file or become `useFilter().hasFilter() || search.length > 0` once the hook gains `hasFilter()`.

---

### 6. Region semantics for the data-updating area

The table region updates dynamically as the user types in the search field, applies filters, sorts, and paginates. Today there's no a11y signal that data is changing. Wrap the table in a live region:

```tsx
<div
  role="region"
  aria-label="Venues"
  aria-live="polite"
  aria-busy={isLoading}
>
  <VisuallyHidden role="status">
    {isLoading ? 'Loading venues' : 'Venues loaded'}
  </VisuallyHidden>
  <Table ...>...</Table>
</div>
```

Why this matters specifically for a *data-management example*:
- `aria-live="polite"` — screen readers announce row-count changes without interrupting.
- `aria-busy` — assistive tech knows a refetch is in flight; pairs naturally with a future `isPending` from a real data hook.
- The `<VisuallyHidden role="status">` — the user is told whether they're currently looking at stale or fresh data.

This is the kind of detail people copy-paste verbatim, so getting it right in the canonical example pays off across every consumer.

---

### 8. `Tag.Group` `removeAll` + `emptyState` for applied filters

`applied-filter.tsx` currently renders nothing when the filter is empty, and there's no "clear all" affordance — users have to remove tags one at a time or open the drawer and reset.

Two `Tag.Group` props you're already wiring solve both:

```tsx
<Tag.Group
  label="Applied Filters"
  onRemove={keys => removeFilter(keys as Set<FilterKeys>)}
  removeAll
  emptyState={() => (
    <Text variant="muted" fontSize="sm" fontStyle="italic">
      No filters applied
    </Text>
  )}
>
  {/* tags */}
</Tag.Group>
```

`removeAll` adds a built-in "clear all" button to the group; `emptyState` gives the component a graceful zero-tags state instead of rendering nothing. This also ties together with the `hasFilter()` helper proposed in the hooks comment — the chip area becomes self-narrating.

---

### 9. Wire `setSearch` directly to `<SearchField onSubmit>`

`Toolbar` currently wraps `<SearchField>` in a `<Form>` and intercepts at the form level (`onSearch` reads `FormData` and pulls `q`). `<SearchField>` already exposes an `onSubmit` callback that fires on Enter and on the clear button — passing `setSearch` directly is simpler:

```tsx
<SearchField
  aria-label="Search venues"
  description="Search by name"
  width={64}
  defaultValue={search}
  onSubmit={setSearch}
  onClear={() => setSearch('')}
/>
```

The surrounding `<Form onSubmit={onSearch} unstyled>` and the explicit `<Button type="submit">Search</Button>` can both go away — `SearchField` already has its own submit affordance built in. Net: ~15 lines removed from `toolbar.tsx`, identical UX, fewer concepts to explain in the docs.

---

### Summary

These changes net out as **substantial code deletion** (`getFormData`, the four Zod schemas/transforms, scattered `setPage(null)` calls, the form wrapper around the search field, the prop-threaded `clearFilters`) in exchange for **stronger demonstration of the patterns the example exists to teach** (URL-state defaults, confirmations, toasts, accessibility, applied-filter UX). None of it is out of scope for `[DST-1314]` — they're all things this PR already gestures at but doesn't quite close on.

[DST-1314]: https://reservix.atlassian.net/browse/DST-1314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

docs/app/(examples)/examples/filter/venues-table.tsx
```diff
@@ -0,0 +1,328 @@
...
+
+  // Keep the URL in sync when the requested page is out of range so a stale
+  // ?page=99 query string doesn't survive after filtering shrinks the result.
+  // Use `history: 'replace'` for this correction so we don't trap the user's
+  // Back button on the same view.
+  useEffect(() => {
+    if (page !== safePage) {
+      setPage(safePage === 1 ? null : safePage, { history: 'replace' });
+    }
+  }, [page, safePage, setPage]);
```
sebald: **This whole effect should be deleted as a consequence of the other architectural changes.**

The effect handles two scenarios that produce a stale `?page=N`:

1. **Filter/search change shrinks the result set.** With cross-hook coordination on `useFilter` / `useSearch` (see prior comment — `setFilter` and `setSearch` reset page to `1` at the hook layer), this scenario can never happen. Every filter/search mutation resets pagination *before* the URL ever ends up out of range. The bug class disappears by construction.

2. **User pastes / bookmarks a URL with an out-of-range page.** The local `safePage = Math.min(page, Math.max(1, totalPages))` clamp already shows the correct rows. The only thing the effect does in this case is rewrite the URL to match — and rewriting the user's URL on mount is a surprising side-effect, which is precisely why the comment has to explain `history: 'replace'` ("so we don't trap the user's Back button"). A simpler answer is to just not rewrite: clamp the displayed slice and let the URL be whatever the user typed. The next pagination click writes a fresh, valid value.

So the chain is:

- Hook coordination → kills the common case (#1) entirely.
- Drop the URL-rewrite for the paste/bookmark edge case (#2) — it's not a correctness issue, just URL hygiene the example doesn't need to teach.

Net deletion from `venues-table.tsx`:
- The `useEffect` block (5 lines) + its 4-line preamble comment (these 9 lines)
- `useEffect` from the `react` import (line 10)
- The `safePage === 1 ? null : safePage` and `{ history: 'replace' }` argument shape inside the call

The `safePage` derivation itself stays — it's pure, no effect needed, and protects the slice from rendering empty pages on stale URLs.

sebald (commented):
Follow-up tied to the architecture comment above — the page-correction `useEffect` becomes unnecessary once `useFilter`/`useSearch` coordinate with pagination at the hook layer. See inline comment.

sebald (changes requested):
## Requesting changes

Re-raising my review state to **changes requested**. Several substantive items from my earlier review (29 Apr) and the follow-ups today have not been addressed:

**Blocking:**

1. **`useFilter` / `useSearch` / `useSort` architecture** — `parseAsJson` for filter/sort still produces opaque URLs; mixed `useQueryState` API shapes; no cross-hook page reset; no `hasFilter()` helper. Comment thread [#3159373437](https://github.com/marigold-ui/marigold/pull/5357#discussion_r3159373437) was resolved without the change being applied. See inline comments on `utils.tsx` and the architectural follow-up comment.

2. **`ToggleButton.Group` + hidden-input bridge for the rating field** — `ToggleButton.Group` is not a form element and will not be made one. Revert this section to `Radio.Group`. See inline comment on `toolbar.tsx`.

3. **Form/URL transform layer can collapse to `parseFormData`** — once the hooks switch to per-field parsers, `getFormData`, `formSchema`, `toFormSchema`, and `toUrlSchema` all delete. See architectural follow-up comment, item 1.

4. **Page-correction `useEffect` in `venues-table.tsx`** — becomes unnecessary as a direct consequence of the hook coordination above. See inline comment.

**Strongly recommended (still in scope for [DST-1314]):**

5. Dedicated `Provider` with `<NuqsAdapter defaultOptions={{ clearOnDefault: true }}>`, `ConfirmationProvider`, `ToastProvider` — eliminates `setPage(null)` calls and unblocks the destructive-delete pattern.
6. Pagination extracted into its own component with page-size selector, "showing X – Y of Z" text, and scroll-into-view on page change.
7. `FilterEmptyState` reads hooks directly instead of taking `onClear` as a prop.
8. Region semantics (`role="region"`, `aria-live`, `aria-busy`, `<VisuallyHidden role="status">`) on the table area.
9. `Tag.Group` `removeAll` + `emptyState` props on the applied-filter chips.
10. `<SearchField onSubmit={setSearch}>` instead of the surrounding `<Form>` wrapper.

**Also still open:**

- PR title says *"Full page **date** pattern"* — should be *"**data**"*. Fix on merge.
- The PR's stated purpose is "data management pattern" but React Query was removed entirely, so the example no longer demonstrates server data fetching, mutations, optimistic updates, query keys, loading states, or error states. If that's intentional, retitle the example and update the description; if not, those patterns need to come back.

Happy to pair on any of the items above. The hook architecture (#1) and the form helper switch (#3) together net out as a substantial code *deletion*, not addition — this is a "less code, better example" refactor.

sebald (approved):

PR #5362, opened 2026-04-27
feat(DST-1375): make Panel.Header optional for title-only Panels

sebald:
## Summary

- Allow `Panel.Title` as a direct child of `Panel` when no description or actions are needed. The headerless authoring form was already accessibility-correct (the `aria-labelledby` slot/context plumbing lives on the `Panel` root, not on `Panel.Header`); this PR adds the matching horizontal panel padding via Tailwind v4's `not-in-data-panel-header:` variant and officially documents it.
- `Panel.Description` and `Panel.HeaderActions` continue to require a `Panel.Header` wrapper.
- Drops the now-redundant `Panel.Header` from 8 dashboard examples + 1 docs MDX snippet (all "Danger zone" panels and form/inventory headings that had only a title).

Implementation note: `PanelHeader` adds a `data-panel-header` attribute; `PanelTitle` carries `not-in-data-panel-header:px-(--panel-px)` so the padding only emits when no `[data-panel-header]` ancestor exists. No React context, no provider boundary — purely DOM-driven via Tailwind variant. Pattern mirrors `BooleanField` ↔ `Checkbox`/`Switch`.

Jira: [DST-1375](https://reservix.atlassian.net/browse/DST-1375)

Changeset: `.changeset/panel-header-optional.md` (patch).

## Test plan

- [ ] `pnpm typecheck:only` clean
- [ ] `pnpm test:unit -- Panel` — 34/34 Panel tests + new "Title-only Panel" describe block pass (1118/1118 overall)
- [ ] `pnpm test:sb -- Panel` — 11/11 Panel storybook tests including the new `TitleOnlyWithoutHeader` story
- [ ] Visual check (verified in 6007 Storybook): title `padding-inline: 12px` headerless, `0px` inside `Panel.Header` — same alignment in both forms, no double-padding
- [ ] `pnpm sb` and visit the new `TitleOnlyWithoutHeader` story
- [ ] `pnpm start` and review the updated dashboard examples (Security → Danger zone, General → Danger zone, Billing → Danger zone, etc.)
- [ ] Verify the new `panel-without-header` demo renders in `/components/layout/panel`

[DST-1375]: https://reservix.atlassian.net/browse/DST-1375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/Panel/Panel.stories.tsx
+++ b/packages/components/src/Panel/Panel.stories.tsx
@@ -110,7 +110,6 @@ Basic.test(
 
 export const TitleOnlyWithoutHeader = meta.story({
   args: { children: null as never },
-  tags: ['component-test'],
   render: args => (
     <Panel {...args}>
       <Panel.Title>Quick Settings</Panel.Title>
@@ -121,14 +120,6 @@ export const TitleOnlyWithoutHeader = meta.story({
       </Panel.Content>
     </Panel>
   ),
-  play: async ({ canvas }) => {
-    const title = canvas.getByRole('heading', { name: 'Quick Settings' });
-    const region = canvas.getByRole('region', { name: 'Quick Settings' });
-
-    expect(title.tagName).toBe('H2');
-    expect(region.getAttribute('aria-labelledby')).toBe(title.id);
-    expect(title.closest('[data-panel-header]')).toBeNull();
-  },
 });
 
 export const WithHeaderActions = meta.story(() => (
--- a/packages/components/src/Panel/Panel.stories.tsx
+++ b/packages/components/src/Panel/Panel.stories.tsx
@@ -108,6 +108,33 @@ Basic.test(
   }
 );
 
+export const TitleOnlyWithoutHeader = meta.story({
+  args: { children: null as never },
+  tags: ['component-test'],
+  render: args => (
+    <Panel {...args}>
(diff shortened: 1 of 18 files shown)


packages/components/src/Panel/Panel.stories.tsx
```diff
@@ -108,6 +108,33 @@ Basic.test(
...
+      <Panel.Content>
+        <Stack space="regular">
+          <TextField label="Display name" defaultValue="Marigold Events" />
+        </Stack>
+      </Panel.Content>
+    </Panel>
+  ),
+});
+
+TitleOnlyWithoutHeader.test(
```
jim761: This isn't a component test, should be a normal test

jim761 (changes requested):

packages/components/src/Panel/Panel.stories.tsx
jim761: you're not having any user interactions in this test, so shouldn't it be in the normal test file then?

sebald: good catch

jim761 (approved):

PR #5364, opened 2026-04-27
refa(DST-1213): migrate @marigold/icons to lucide-react proxy

sarahgm:
# Description

Migrates `@marigold/icons` from a curated catalog of hand-rolled SVG components to a thin `lucide-react` proxy plus 12 retained custom icons that don't have a Lucide equivalent (DesignTicket, Facebook, GiftCard, Google, Instagram, PDF, Resale, Scanner, Stadium, TicketInsurance, Turnstile, Twitter).

Highlights:

- `packages/icons/src/index.ts` is now `export * from 'lucide-react'` plus the 12 custom re-exports.
- All 12 custom icons are built via `lucide-react`'s `createLucideIcon` factory, so they share the exact same prop API and runtime behavior as the re-exported Lucide icons (same `LucideProps` shape, `color` → `stroke` mapping, `LucideProvider` context support, auto `aria-hidden`, ref forwarding).
- Peer dependency narrowed to `react: >=19.0.0`; the unused `react-dom` peer has been removed.
- Three docs demos and the Button icon test updated to track lucide's attribute serialization (`width="30"` vs `"30px"`) and `color` → `stroke` mapping.
- `docs/ui/IconList.tsx` widened to `string` + cast through `unknown` so the `foundations/icons` page keeps building until the catalogue MDX is rewritten.

**Note**: The `docs/content/foundations/icons/index.mdx` page (the "all icons" gallery) still references ~65 names that no longer exist after the Lucide migration. The runtime guard in `IconList` falls back to a `console.warn` and renders nothing for unknown names, so the page builds and doesn't crash. Rewriting that page to use the Lucide names is tracked separately as **DST-1308**.

**Breaking change**: any consumer importing `Add`, `BurgerMenu`, `Cancel`, `Close`, `Exclamation`, `FormatBold`, `FormatItalic`, `Logout`, `Parking`, `SettingDots`, `Underlined`, etc. from `@marigold/icons` will need to switch to the equivalent Lucide name (`Plus`, `Menu`, `X`, `TriangleAlert`, `Bold`, `Italic`, `LogOut`, `ParkingCircle`, `EllipsisVertical`, `Underline`, …). The full mapping and migration examples live in the changeset.

# Test Instructions:

1. `pnpm install && pnpm build`
2. `pnpm typecheck` and `pnpm typecheck:docs` — both should pass.
3. `pnpm test:unit` — 1131+ tests passing.
4. `pnpm sb` → open **Icons / All custom icons** and verify all 12 render at the controls' default size and respect `color`, `strokeWidth`, `size`.
5. Spot-check existing component stories that previously used names like `SettingDots`, `Logout`, `User`, `Star`, `Search`, `Edit` — they should all still render via the Lucide re-exports.
6. `pnpm start` → smoke-test the three updated docs demos:
   - `/patterns/feedback/feedback-messages/confirmation-dialog`
   - `/components/navigation/accordion/multi-select-accordion`
   - `/components/content/divider/divider-vertical`

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

- [ ] Added/Updated documentation (if it already exists for this component) — *the `docs/content/foundations/icons` catalogue rewrite is tracked under DST-1308*
- [ ] Updated visual regression tests (only necessary when ui changes in the PR) — *needs `/run-chromatic` once this PR is open*


jim761:
i cannot leave a comment in the docs so i leave it here, in the icon component page the icon is not loading

packages/icons/src/Icon.test.tsx
```diff
@@ -1,34 +1,64 @@
...
-import { Facebook } from './social/Facebook';
-import { ArrowUp } from './ui/ArrowUp';
+import { describe, expect, test } from 'vitest';
+import { DesignTicket } from './custom/DesignTicket';
+import { Search } from './index';
 
-test('supports default size', () => {
-  render(<Facebook data-testid="svg" />);
-  const svg = screen.getByTestId(/svg/);
+describe.each([
```
jim761, line 7: There's no test to see if fill works

jim761 (changes requested):

.changeset/icons-lucide-migration.md
```diff
@@ -0,0 +1,34 @@
...
+  Search,
+  Check,
+  Plus as Add,
+  Ticket,
+  Armchair as Seat,
+  BadgePercent as Deal,
+} from '@marigold/icons';
+```
+
+The 12 retained custom icons (no Lucide equivalent) live under `src/custom/`:
```
aromko: Is this helpful for the user? As user I would also just import via @marigold/icons or? Or how do I import e.g. DesignTicket?

packages/components/src/Accordion/Accordion.stories.tsx
```diff
@@ -1,6 +1,10 @@
+import {
+  Accessibility as Accessible,
```
aromko: Why aliasing (happens in different files) and why from lucide-react and not from @marigold/icons?

sarahgm:
> i cannot leave a comment in the docs so i leave it here, in the icon component page the icon is not loading

would do the icon page and the icons page in DST-1308

packages/icons/src/custom/DesignTicket.tsx
```diff
@@ -0,0 +1,12 @@
+import { type LucideIcon, createLucideIcon } from 'lucide-react';
```
aromko: General question: What about the custom icons under src/components? e.g. TriangleAlert we have now twice.

sarahgm: As far as I remember we wanted to have the components package and icons package not depending on each other, thats why we have own icons in components.

.changeset/icons-lucide-migration.md
sarahgm: updated to be more user centered

packages/icons/src/custom/DesignTicket.tsx
aromko: ok than we just can hope they stay consistent 🤞

aromko (approved):

aromko:
@jim761 what is missing. Can we merge this?

jim761 (approved):

PR #5366, opened 2026-04-27
feat(DST-1076): standardize `<SelectList>` API and refresh docs

sebald:
## Summary

[DST-1076](https://reservix.atlassian.net/browse/DST-1076) — turns `<SelectList>` into a first-class form field for picking one or many items from a fully visible list of rich rows.

### Component (breaking)

- `SelectList.Item` → **`SelectList.Option`**, matching `Select.Option` and the HTML `<option>` mental model.
- `SelectList.Action` is **gone** — drop your `<ActionMenu>` or `<IconButton>` directly inside `<SelectList.Option>`. The component positions, sizes, and styles the nested control via `ButtonContext`.
- Leading-image slot **removed**. Compose images inside `<Text slot="label">` or anywhere in children.
- `selectionMode="none"` no longer accepted. `SelectList` is a form field; default is `"single"`.
- `onChange` is strictly typed per `selectionMode`: `(key: Key | null) => void` for single, `(keys: Key[]) => void` for multiple — mirrors `Select<T, M>`.
- New theme entry `SelectList` with first-class `label`, `description`, `action` slots (no descendant selectors). Custom themes must add the entry.

### Component additions

- `orientation="horizontal"`, `emptyState`, `disallowEmptySelection`.
- Inset-style padding API: `p` / `px` / `py` (mirrors `<Inset>` / `<Panel>`).
- Visible selection indicator (radio circle for single, checkbox for multiple).
- Real form-field plumbing via `FieldBase`: hidden native `<select>` for native validation, `useFormValidation` returns focus to the list on invalid submit.
- Dev-mode warning when `textValue` is missing on rich-content options.

### Visual refresh

- Top-aligned rows, edge-to-edge items with inset focus ring, hover via theme background.
- Bordered variant rebuilt on the `ui-surface` primitive.
- Horizontal variant no longer stretches to full width.

### Documentation

Full rewrite of the SelectList page: anatomy diagram, decision table for picking between `<SelectList>` and lighter controls (`<Radio.Group>`, `<Checkbox.Group>`, `<Select>`, `<Combobox>`, `<TagField>`), and dedicated sections for multi-selection, per-row actions (decision-help and configuration patterns), horizontal orientation, and empty state. Replaces selected prose with Do/Don't guideline tiles. Tightens the accessibility section to what's specific to SelectList (keyboard model, label requirement, `textValue` for rich rows).

### Migration

```diff
- <SelectList selectionMode="none">
-   <SelectList.Item id="free">
-     <SelectList.Action>
-       <IconButton aria-label="Info"><Info /></IconButton>
-     </SelectList.Action>
-     Free
-   </SelectList.Item>
- </SelectList>
+ <SelectList selectionMode="single">
+   <SelectList.Option id="free">
+     Free
+     <IconButton aria-label="Info"><Info /></IconButton>
+   </SelectList.Option>
+ </SelectList>
```

## Test plan

- [ ] `pnpm test:unit` passes
- [ ] `pnpm test:sb` passes (browser-based play tests)
- [ ] Storybook renders all `SelectList` stories (default, bordered, horizontal, multi, decision-help, configuration, empty, padding)
- [ ] Docs page renders without errors at `/components/form/selectlist` (anatomy, decision table, demos, alternatives)
- [ ] No regressions on pages that reference `<SelectList>` (Radio, Checkbox, TagField, Multiselect)

[DST-1076]: https://reservix.atlassian.net/browse/DST-1076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sebald:
Has some breaking, but component usage is zero

packages/components/src/SelectList/SelectList.tsx
```diff
@@ -1,76 +1,364 @@
...
+    value: validationValue,
+    isInvalid: error,
+    validate,
+    validationBehavior,
+  });
+
+  const handleSelectionChange = useCallback(
+    (keys: Selection) => {
+      setSelection(keys);
+      validationState.commitValidation();
```
aromko, line 290: 🔴 **Visual Regression Tests — not run**

This PR is a complete visual refresh of `<SelectList>` (top-aligned rows, edge-to-edge layout, new selection indicators, bordered variant rebuilt on `ui-surface`) plus changes that ripple into Checkbox/Radio. The `Visual-Regression-Tests` workflow has not run for this branch.

Please trigger it by commenting `/run-chromatic` on this PR before merging.

packages/components/src/AppLayout/AppLayoutMain.tsx
```diff
@@ -3,5 +3,7 @@ import type { PropsWithChildren } from 'react';
 export type AppLayoutMainProps = PropsWithChildren;
 
 export const AppLayoutMain = ({ children }: AppLayoutMainProps) => (
-  <main className="relative overflow-y-auto [grid-area:main]">{children}</main>
+  <main className="relative flex flex-col gap-4 overflow-y-auto px-4 [grid-area:main]">
```
aromko: 🔴 **Out-of-scope change**

This adds `flex flex-col gap-4 px-4` to the docs app shell `<main>` — high blast radius across the docs site, and not mentioned in the PR description. Please either split into a separate PR or document the rationale (and confirm it doesn't shift other docs pages).

tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": "@marigold/tsconfig/react",
   "compilerOptions": {
+    "declaration": false,
```
aromko: 🔴 **Out-of-scope change**

`"declaration": false, "declarationMap": false` at the root tsconfig is unrelated to the SelectList rewrite. Please add a note in the PR description explaining why these are needed, or split into a separate PR.

themes/theme-rui/src/components/Checkbox.styles.ts
```diff
@@ -4,7 +4,7 @@ export const Checkbox: ThemeComponent<'Checkbox'> = {
   checkbox: cva({
     base: [
       'grid size-4 shrink-0 place-content-center rounded',
-      'border border-input shadow-elevation-border',
+      'border border-input shadow-elevation-border bg-background',
```
aromko, line 7: 🔴 **Out-of-scope change — visual primitive**

Adding `bg-background` to the base `Checkbox` style is a visual change to a primitive used everywhere (and the snapshot test at `Checkbox.test.tsx:47` was updated to reflect it). Even if it's needed for the new selection indicator, this should land as its own deliberate change with its own changeset entry, not folded into the SelectList PR.

Same applies to `Radio.styles.ts` (which has no snapshot test catching it).

themes/theme-rui/src/components/Radio.styles.ts
```diff
@@ -11,7 +11,7 @@ export const Radio: ThemeComponent<'Radio'> = {
   radio: cva({
     base: [
       'aspect-square size-4 rounded-full',
-      'border border-input shadow-elevation-border',
+      'border border-input shadow-elevation-border bg-background',
```
aromko, line 14: 🔴 **Out-of-scope change — visual primitive**

Same as `Checkbox.styles.ts`: this adds `bg-background` to the base `Radio` style, but unlike Checkbox there is no snapshot test catching the change. Please surface this as a deliberate Radio change with its own changeset, or roll it back if it isn't required for the SelectList work.

packages/components/src/SelectList/SelectList.tsx
aromko, line 290: 🟡 **Possible stale validation**

```tsx
const handleSelectionChange = useCallback((keys: Selection) => {
  setSelection(keys);
  validationState.commitValidation();
}, [setSelection, validationState]);
```

`useFormValidationState` derives validation from the `value` prop, which is `validationValue`, which is `useMemo`'d from `selection`. At the moment `commitValidation()` is called, `selection` hasn't flushed yet, so the committed validation reflects the **previous** selection. RAC's own `useFormValidation` typically commits via effect rather than inline.

Worth verifying this doesn't cause a one-event lag in error display, especially with `validationBehavior: "native"`.

packages/components/src/SelectList/SelectionIndicator.tsx
```diff
@@ -0,0 +1,46 @@
+import { cn } from '@marigold/system';
+import { Checkbox } from '../Checkbox/Checkbox';
+
+interface SelectionIndicatorProps {
+  selectionMode: 'single' | 'multiple' | 'none';
```
aromko: 🟡 **Dead code**

This PR explicitly removes `selectionMode="none"` support, and the render-prop from `RACGridListItem` won't pass `'none'` either. Drop `'none'` from the union and remove the `if (selectionMode === 'none') return null;` early return at line 35.

```suggestion
  selectionMode: 'single' | 'multiple';
```

aromko (commented):
## Automated Code Review

### Overview

| Field | Value |
|-------|-------|
| Files Changed | 47 |
| Additions | +3545 |
| Deletions | −457 |
| CI | 1 failure (`Storybook Tests 4/4`) — flaky module-loading error in `@storybook/addon-vitest`; all 33 tests in shard passed |

### Linked Ticket

[DST-1076](https://reservix.atlassian.net/browse/DST-1076) — Standardize `<SelectList>` API and item layout, separate visually from `<ListBox>`.

### Code Review Checklist

- ✅ TypeScript: strict types, generic `SelectionMode`, no `any` in component code
- ✅ RAC wrapping convention (`isDisabled` → `disabled`, no `className`/`style` exposed)
- ✅ `useClassNames` for theming; theme entry registered in `system/types/theme.ts`
- ✅ Compound component pattern (`SelectList.Option`)
- ✅ Tests use `userEvent`, accessible queries, story imports — match Marigold convention
- ✅ A11y: `aria-labelledby`, hidden native `<select>` for native validation, dev warning when `textValue` is missing on rich rows

### Recommendation

The SelectList rework itself is high-quality: clean type ergonomics (`SelectionMode` generic, mutually-exclusive padding props), real form-field plumbing, comprehensive tests, and a thoughtful docs rewrite.

Before merge:
1. Run VRT (comment `/run-chromatic`) — this is a full visual refresh.
2. Either split out the Checkbox/Radio/AppLayoutMain/tsconfig edits, or document them in the PR description with rationale.
3. Address the inline warnings on `validationState.commitValidation()` and the dead `selectionMode === 'none'` branch.

See inline comments for specifics.

---

packages/components/src/SelectList/SelectList.tsx
sebald: I thought I should not run them? 👀

packages/components/src/SelectList/SelectionIndicator.tsx
sebald: Good catch — applied. One nuance: RAC's `ItemRenderProps.selectionMode` is typed `'none' | 'single' | 'multiple'`, so just narrowing the prop here breaks typecheck at the `SelectListOption` call site. I narrowed `SelectionIndicator` to `'single' | 'multiple'` and added an `as 'single' | 'multiple'` cast at the call site since `SelectList` itself never passes `'none'` to RAC (`resolvedSelectionMode = (selectionMode ?? 'single') as SelectionMode`).

themes/theme-rui/src/components/Checkbox.styles.ts
sebald: Fair — split out into #5378 with its own changeset and the snapshot update.

themes/theme-rui/src/components/Radio.styles.ts
sebald: Same — covered by #5378 (Checkbox and Radio together, with a deliberate changeset entry).

packages/components/src/AppLayout/AppLayoutMain.tsx
sebald: Yeah, that one was an accident — reverted.

tsconfig.json
sebald: Reverted. I added these while working through some IDE errors but `tsc --noEmit --project tsconfig.json` reports identical errors with or without `declaration: false` (the real culprit is unhoisted `@react-stately/form` etc., unrelated to this PR). The override wasn't actually doing anything for us.

packages/components/src/SelectList/SelectList.tsx
sebald: Dug into this with the RAC source — there's no lag. The exposed `commitValidation()` doesn't commit synchronously, it just queues a flag:

```ts
// react-stately/src/form/useFormValidationState.ts:186
commitValidation() {
  // Wait until after the next render to commit so that the latest value has been validated.
  if (validationBehavior === 'native') {
    setCommitQueued(true);
  }
  setServerErrorCleared(true);
}
```

The actual commit runs via `useEffect(commitValidation)` after the next render. By that point, `clientError = useMemo(..., [validate, value])` has already recomputed with the **new** `value` — so the committed validation reflects the post-change selection.

The hook's own comment makes this explicit: _"Wait until after the next render to commit so that the latest value has been validated."_

And the inline `setX(); commitValidation();` pattern is exactly what RAC's own collection states use:

- `useRadioGroupState` ([radio/useRadioGroupState.ts:87–91](https://github.com/adobe/react-spectrum/blob/main/packages/react-stately/src/radio/useRadioGroupState.ts#L87-L91)):
  ```ts
  let setSelectedValue = (value) => {
    if (!props.isReadOnly && !props.isDisabled) {
      setSelected(value);
      validation.commitValidation();
    }
  };
  ```
- `useSelectState` ([select/useSelectState.ts:165–176](https://github.com/adobe/react-spectrum/blob/main/packages/react-stately/src/select/useSelectState.ts#L165-L176)) does the same inline call after `setValue(...)`.

So our `setSelection(keys); validationState.commitValidation();` matches RAC's own pattern 1:1 and works correctly under both `validationBehavior: "native"` and `"aria"`.

aromko: yes it from the review-pr skill :-D

aromko: I was too lazy to remove this so just resolve for now

aromko (approved):

PR #5367, opened 2026-04-27
feat(DST-1264): add @marigold/cli — terminal access to docs, discovery, and setup

jim761:
## Summary

Adds `@marigold/cli`, a new CLI package that brings Marigold component docs, discovery, and project setup into the terminal. Designed primarily so **AI coding agents** can fetch accurate Marigold API data instead of guessing from training data.

- New package `packages/cli` (`@marigold/cli`) with five commands: `docs`, `list`, `init`, `telemetry`, `completion`
- Docs site: consumes the static `/manifest.json` built by `docs/scripts/build-manifest.mjs` (now emits categorized components, top-level `@marigold/components` `version`, and a `generatedAt` timestamp), and adds an `/api/telemetry` ingest route
- `CLAUDE.md`: documents the CLI as the recommended way for agents to look up component APIs

## Commands

| Command | Purpose |
| --- | --- |
| `marigold docs <Component>` | Fetch component docs. Flags: `--section props\|usage\|examples`, `--format markdown\|json\|plain`, `--fresh`, `--offline` |
| `marigold list` | List components. Flags: `--category`, `--search`, `--format json` |
| `marigold init` | Interactive wizard: installs `@marigold/components`, `@marigold/system`, `@marigold/theme-rui`, patches Tailwind, wraps root layout with `MarigoldProvider` |
| `marigold telemetry <status\|enable\|disable>` | Manage anonymous usage telemetry. Honors `DO_NOT_TRACK`, `CI`, `MARIGOLD_TELEMETRY_DISABLED` |
| `marigold completion <bash\|zsh\|fish>` | Print a shell-completion script for the chosen shell. Source the output (or write it to your shell's completions dir) to enable tab completion of subcommands, flags, and component names. |

Docs are fetched from the docs site, cached for 24h in `$XDG_CACHE_HOME/marigold` (offline-friendly via `--offline`).

## Security hardening

Remote markdown is sanitized at the fetch boundary before it reaches the terminal. The previous `stripAnsi` only covered SGR color codes — a compromised or overridden docs origin could have injected OSC 52 (clipboard write), cursor moves, screen clears, or DCS payloads.

- `lib/strip-ansi.ts` covers the full ECMA-48 set (CSI, OSC, DCS/APC/PM/SOS, single-char escapes, 8-bit C1 variants) plus stray C0 controls.
- `getComponentDocs` and `transformManifest` sanitize remote content on ingest, so all output formats (`markdown` / `plain` / `json`) are safe.
- The formatter uses the same stripper as defense-in-depth on the `plain` path.

Follow-ups (separate PRs): validate `MARIGOLD_DOCS_URL`, add manifest schema validation, replace the telemetry temp-file URL channel.

## Test plan

To Test it locally, go to the cli folder cd packages/cli and run pnpm link --global, that way you can use the commands normally like marigold list, to uninstall it run pnpm uninstall -g @marigold/cli, also add an .env.local, in packages/cli, that has MARIGOLD_DOCS_URL="https://marigold-docs-git-marigold-cli-v1-marigold.vercel.app/" written in it. If you don't want to link it you can use node packages/cli/dist/bin/marigold.mjs docs Button.

- [ ] `pnpm install` succeeds and the postinstall build picks up the new package
- [ ] `pnpm --filter @marigold/cli build` produces a working `dist/` (tsdown)
- [ ] `pnpm --filter @marigold/cli test` covers cache, manifest, docs, format, detect-project, strip-ansi, **completion, init, edit-css, edit-tsx, edit-vite-config**
- [ ] `pnpm typecheck:only` clean
- [ ] `pnpm lint` clean
- [ ] `node packages/cli/dist/bin/marigold.mjs docs Button` prints markdown docs (with docs site running locally or against preview, add an .env.local that has MARIGOLD_DOCS_URL="https://marigold-docs-git-marigold-cli-v1-marigold.vercel.app/" written in it)
- [ ] `… docs Button --section props --format json` returns a structured JSON payload
- [ ] `… list --category form` filters correctly
- [ ] `… list --search date --format json` returns matching entries
- [ ] `… telemetry status` / `disable` / `enable` round-trips persisted state
- [ ] `… docs Button --offline` after a warm cache works without network; fails clean on cold cache
- [ ] `… init --yes` in a scratch Next.js app installs deps, patches Tailwind, wraps the layout, and the app boots
- [ ] `marigold completion bash | source` enables tab completion (likewise `zsh` / `fish`)
- [ ] `marigold completion <shell>` output passes shellcheck / has no unsanitized component names
- [ ] Static manifest: `curl http://localhost:3000/manifest.json` includes a top-level `version` (matching `@marigold/components`) and `generatedAt` ISO timestamp
- [ ] Docs API: `curl -X POST http://localhost:3000/api/telemetry -d '{...}'` accepts a payload (or no-ops gracefully when Redis env is unset)
- [ ] After running a CLI command with telemetry enabled, the corresponding event is visible in the Upstash KV database (the one bound to `KV_REST_API_URL` / `KV_REST_API_TOKEN`) — confirms the `/api/telemetry` ingest path end-to-end
- [ ] CI in CI=1 environment auto-suppresses telemetry (no network calls)
- [ ] Markdown response containing `\x1b]52;c;…\x07` and `\x1b[2J\x1b[H` is stripped from all output formats

## Notes for reviewers

- Library exports (`getComponentDocs`, `loadManifest`, …) are intentional — the MCP server will reuse them in a follow-up. The `Manifest` / `ManifestPage` types are marked `@internal` so outside consumers don't implicitly pin to the 0.1.0 shape.
- `engines.node` is `>=22` (vs the repo root's `24.x`) on purpose — consumer machines installing the CLI may not be on Node 24 yet.
- Changeset: `minor` bump for `@marigold/cli` and `@marigold/docs`.
- Jira: DST-1264.
- `packages/components/src/Grid/Grid.test.tsx` has a one-line camelCase fix for `gridTemplateAreas` that landed on this branch. Strictly outside CLI scope; flag if you'd prefer it in a separate PR.


packages/cli/README.md
```diff
@@ -0,0 +1,144 @@
+# @marigold/cli
+
+Marigold Design System CLI — component docs, discovery, and project setup, straight from the terminal.
```
aromko, line 3: How I can use it locally? Is nowhere documented or? and maybe a command in pnpm would be the easiest?

packages/cli/src/bin/marigold.ts
```diff
@@ -0,0 +1,293 @@
+#!/usr/bin/env node
+import pc from 'picocolors';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { parseArgs } from 'node:util';
+import { runCompleteSuggest, runCompletion } from '../commands/completion.js';
+import { runDocs } from '../commands/docs.js';
+import { runInit } from '../commands/init.js';
```
aromko: ⚠️ **Eager `init` import in the hot path**

This top-level import transitively loads `@babel/parser` (~1MB), `magic-string`, and `@clack/prompts` on **every** `marigold docs` / `list` invocation. AI agents will hit this dozens of times per session, and `docs` is the hot path the package is optimised for.

Dynamic-import the init module inside the `if (command === 'init')` branch so docs/list calls stay lean:

```ts
if (command === 'init') {
  const { runInit } = await import('../commands/init.js');
  // ...
}
```

Same applies to `@clack/prompts` inside `init.ts` if you want belt-and-braces.

packages/cli/src/lib/telemetry.ts
```diff
@@ -0,0 +1,136 @@
...
+  if (process.env.DO_NOT_TRACK === '1') return true;
+  if (ci.isCI) return true;
+  const config = readConfig();
+  if (config.telemetryEnabled === false) return true;
+  return false;
+};
+
+export const setTelemetryEnabled = (enabled: boolean): void => {
+  const config = readConfig();
+  config.telemetryEnabled = enabled;
```
aromko, line 55: ⚠️ **Opt-out still mints an anonymous UUID**

`setTelemetryEnabled(false)` persists a UUID into `~/.config/marigold/config.json`. A user explicitly opting out shouldn't end up with a stored identifier; the ID is only useful if they later re-enable.

`anonymousId()` already lazy-creates the UUID on `emit()`, which is the right path. Mint here only when `enabled === true`:

```suggestion
export const setTelemetryEnabled = (enabled: boolean): void => {
  const config = readConfig();
  config.telemetryEnabled = enabled;
  if (enabled && !config.anonymousId) config.anonymousId = crypto.randomUUID();
  writeConfig(config);
};
```

packages/cli/src/lib/cache.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (Date.now() - stat.mtimeMs > ttl) return null;
+    return fs.readFileSync(file, 'utf8');
+  } catch {
+    return null;
+  }
+};
+
+const writeEntry = (file: string, data: string): void => {
+  fs.mkdirSync(path.dirname(file), { recursive: true });
+  fs.writeFileSync(file, data);
```
aromko: ⚠️ **Non-atomic cache write — concurrent CLI invocations can corrupt entries**

Two parallel `marigold docs` calls (agents commonly fire several) can race here. The existing parse-failure self-healing in `fetchWithCache` covers truncation on read, but a partial-write window still surfaces as confusing errors.

Write-then-rename is atomic on POSIX:

```ts
const writeEntry = (file: string, data: string): void => {
  fs.mkdirSync(path.dirname(file), { recursive: true });
  const tmp = `${file}.${process.pid}.tmp`;
  fs.writeFileSync(tmp, data);
  fs.renameSync(tmp, file);
};
```

packages/cli/src/lib/edit-tsx.ts
```diff
@@ -0,0 +1,298 @@
...
+  walk(file, (n, parent) => {
+    if (found) return;
+    if (n.type !== 'JSXExpressionContainer') return;
+    const expr = n.expression as AnyNode | undefined;
+    if (!expr || expr.type !== 'Identifier') return;
+    if ((expr as { name?: string }).name !== 'children') return;
+    found = { node: n, parent };
+  });
+  return found;
+};
```
aromko, line 116: ⚠️ **First-match heuristic can wrap the wrong JSX**

`findChildrenExpressionContainer` (and `findRenderArgument` below) walks the entire AST and returns the *first* `{children}` expression / `.render(<JSX/>)` call, not anchored to the default export.

The tests cover the standard single-`{children}` layout shape, but a layout file that defines a helper component which also receives `children` would silently get the helper wrapped instead of the root. Same risk for `editViteEntry` if the entry has multiple `.render(...)` calls.

Worth scoping the search to the default-exported function's return value. Worst case today is a manual revert by the user, but it's the kind of silent edit that's frustrating to debug.

docs/app/api/telemetry/route.ts
```diff
@@ -0,0 +1,78 @@
...
+const getRedis = (): Redis | null => {
+  if (redis) return redis;
+  const url = process.env.KV_REST_API_URL;
+  const token = process.env.KV_REST_API_TOKEN;
+  if (!url || !token) return null;
+  redis = new Redis({ url, token });
+  return redis;
+};
+
+export async function POST(request: Request) {
```
aromko, line 52: ⚠️ **Public POST endpoint with no rate limiting or origin check**

Upstash free tier is 10K commands/day — easy to exhaust. The PR notes the temp-file URL channel as a follow-up but doesn't mention abuse mitigation.

Minimum: throttle per `anonymousId` (Upstash `INCRBY` + `EXPIRE` is one line). Or file a follow-up ticket so it's tracked.

packages/cli/src/commands/init.ts
```diff
@@ -0,0 +1,353 @@
...
+    } else if (options.yes) {
+      setupTailwind = true;
+    } else if (isInteractive) {
+      const ans = await prompts.confirm({
+        message:
+          'Tailwind not detected. Install Tailwind v4 and configure it for you?',
+      });
+      if (prompts.isCancel(ans)) {
+        prompts.cancel('Aborted.');
+        return { installed: false, project, nextSteps: '' };
```
aromko: 💡 **`marigold init` always exits 0, even on user cancel**

This returns `{ installed: false }` on cancel, and `bin/marigold.ts:233-236` ignores the return value. Scripts wrapping `marigold init` can't distinguish "completed" from "user aborted".

Either throw a typed cancel error, or surface a non-zero exit code via the return-value path in `bin/marigold.ts`.

aromko (commented):
## Automated Code Review

Findings grouped by severity. CI is green and nothing here is a ship-blocker — but items #1–#5 get harder to fix after `0.1.0` publishes, so worth handling before release or filing follow-ups.

---

packages/cli/src/lib/telemetry.ts
```diff
@@ -0,0 +1,136 @@
...
+    args: options.args,
+    anonymousId: anonymousId(),
+  };
+
+  const script = findSenderScript();
+  if (!script) return;
+
+  try {
+    const tmpFile = path.join(
+      os.tmpdir(),
```
aromko, line 151: 💡 **Tempfile-per-event can leak**

If the detached child dies before reaching `unlink` in `send-telemetry.ts`, the file leaks into `os.tmpdir()`. For a frequently-invoked CLI this can mount over time.

Either a startup sweep (`marigold-telemetry-*.json` older than 1 day → unlink), or move to a single rolling file. The PR already lists the temp-file URL channel as a follow-up; worth bundling cleanup into that work.

packages/cli/src/commands/list.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  const output = formatList(
+    manifest,
+    { category: options.category, search: options.search },
+    options.format ?? 'markdown'
+  );
+
+  // Cache-hit info isn't surfaced by loadManifest today; manifest is small
+  // enough that we don't bother to thread it through.
+  return { output, cacheHit: false };
```
aromko: 💡 **`cacheHit` hard-coded to `false`**

Under-reports cache effectiveness for `list` in telemetry. `loadManifest` already routes through `fetchWithCache` which returns `{ value, hit }` — just thread it through here and in `lib/manifest.ts:loadManifest` (returning the hit flag alongside the parsed manifest).

packages/cli/src/lib/format.ts
```diff
@@ -0,0 +1,156 @@
...
+import {
+  type Manifest,
+  type ManifestCategory,
+  type ManifestComponent,
+  normalize,
+} from './manifest.js';
+import { stripAnsi } from './strip-ansi.js';
+
+export type OutputFormat = 'markdown' | 'json' | 'plain';
+
```
aromko, line 12: 💡 **Hand-rolled markdown renderer is incomplete**

Covers `#`/`##`/`###`, fences, inline code, and bold — but bullets, links, blockquotes, tables, and nested formatting pass through raw. The Marigold docs origin will produce all of those.

Two paths: (a) widen the renderer (e.g. `marked-terminal`), or (b) document `--format plain` / `--format json` as the recommended paths for non-trivial docs and lean on the markdown path being best-effort. For AI agents `--format json` is preferred anyway, so this is mostly about the human-on-the-terminal experience.

packages/cli/src/commands/init.ts
```diff
@@ -0,0 +1,353 @@
...
+  if (framework === 'vite') {
+    return ['tailwindcss', '@tailwindcss/vite'];
+  }
+  return ['tailwindcss'];
+};
+
+const runInstall = (pm: PackageManager, pkgs: string[], cwd: string) =>
+  new Promise<void>((resolve, reject) => {
+    const [cmd, ...args] = installCommand(pm, pkgs);
+    const child = spawn(cmd, args, { cwd, stdio: 'inherit', shell: false });
```
aromko, line 75: 💡 **`runInstall` doesn't forward signals**

With `stdio: 'inherit'` and `shell: false`, SIGINT propagation across platforms isn't guaranteed for the spawned package manager. On Ctrl+C the parent dies and the child usually inherits via the terminal's foreground process group, but it's not reliable across Windows/CI/non-TTY contexts.

Add an explicit signal-forwarding handler for robustness.

packages/cli/src/lib/manifest.ts
```diff
@@ -0,0 +1,196 @@
...
+    baseUrl: raw.baseUrl,
+    categories,
+    pages: standalonePages,
+  };
+};
+
+export const loadManifest = async (
+  options: CacheOptions = {}
+): Promise<Manifest> => {
+  const url = `${docsUrl()}/manifest.json`;
```
aromko, line 127: 💡 **PR description drift**

The PR body says "Docs site: extends `/api/manifest.json`". The comment on line 32-34 above notes the route was replaced by static `public/manifest.json` in PR #5373, and this code fetches `${docsUrl()}/manifest.json`. Update the PR body so reviewers don't chase a route that isn't there.

packages/cli/src/lib/manifest.ts
```diff
@@ -0,0 +1,196 @@
...
+}
+
+export interface ManifestPage {
+  title: string;
+  slug: string;
+  description?: string;
+}
+
+export interface Manifest {
+  version?: string;
```
aromko, line 32: 💡 **Dead `Manifest` fields**

`version` and `generatedAt` are declared here but never populated by `transformManifest` (lines 65-106) and not serialized by `formatList` JSON output. Either remove them or wire them through from the docs side.

packages/cli/src/lib/config.ts
```diff
@@ -0,0 +1,63 @@
...
+import os from 'node:os';
+import path from 'node:path';
+
+const DEFAULT_DOCS_URL = 'https://www.marigold-ui.io';
+const DEFAULT_CACHE_TTL_MS = 24 * 60 * 60 * 1000;
+
+export const docsUrl = (): string =>
+  process.env.MARIGOLD_DOCS_URL?.replace(/\/$/, '') ?? DEFAULT_DOCS_URL;
+
+export const cacheTtlMs = (): number => {
```
aromko, line 11: 💡 **`MARIGOLD_CACHE_TTL_MS=0` semantics are surprising**

Returns `0`, then `cache.ts:36` reports every entry as stale but still writes the cache file. Probably intentional ("always-stale") but undocumented.

Either document the behavior here, or treat `0` as "caching disabled" and short-circuit the write in `cache.ts`.

packages/cli/src/bin/marigold.ts
```diff
@@ -0,0 +1,293 @@
...
+  '..',
+  '..'
+);
+
+// In local dev only, pick up packages/cli/.env.local. The file is excluded
+// from the published `files` array, so this is a no-op for installed copies.
+// loadEnvFile preserves existing process.env values, so explicit shell
+// exports still win.
+try {
+  process.loadEnvFile(path.join(packageRoot, '.env.local'));
```
aromko, line 31: 💡 **`process.loadEnvFile` runs unconditionally in production builds**

`.env.local` is excluded from the published `files` array, so this is a harmless try/catch in distribution — but dev-only behavior is baked into the production entry point.

Gate on `process.env.NODE_ENV !== 'production'` or check for the file before calling. Defense-in-depth.

packages/cli/src/index.ts
```diff
@@ -0,0 +1,14 @@
+export { getComponentDocs, type ComponentDocs } from './lib/docs.js';
+export {
+  loadManifest,
+  resolveComponent,
+  type Manifest,
```
aromko, line 5: 💡 **`Manifest` / `ManifestPage` exports lack stability markers**

Changeset notes these are intended for the MCP server to reuse. Fine for `0.1.0`, but with no `@internal` JSDoc or subpath separation, an outside consumer who imports them is implicitly pinned to this shape.

Add `@internal` JSDoc, or move under a `@marigold/cli/internal` subpath export.

packages/cli/src/lib/format.ts
```diff
@@ -0,0 +1,156 @@
...
+  const flushCode = () => {
+    if (codeBuffer.length === 0) return;
+    for (const line of codeBuffer) {
+      out.push(pc.dim('│ ') + pc.cyan(line));
+    }
+    codeBuffer.length = 0;
+  };
+
+  for (const rawLine of lines) {
```
aromko: 💡 No-op rename — drop or use `rawLine` directly.

```suggestion
  for (const line of lines) {
```

packages/cli/src/bin/marigold.ts
```diff
@@ -0,0 +1,293 @@
...
+        fail('Usage: marigold telemetry <status|enable|disable>');
+      }
+      const output = runTelemetry({ subcommand: sub });
+      process.stdout.write(output);
+    } else {
+      fail(`Unknown command: ${command}\n\nRun "marigold --help" for usage.`);
+    }
+  } catch (err) {
+    const message = err instanceof Error ? err.message : String(err);
+    process.stderr.write(pc.red(message) + '\n');
```
aromko: 💡 Duplicates the subcommand list maintained in `lib/commands-spec.ts`. Could reuse `TOP_LEVEL_NAMES` (filtered to drop `completion`) so the list doesn't drift when a new command is added.

packages/cli/src/lib/telemetry.ts
```diff
@@ -0,0 +1,136 @@
...
+const bucketDuration = (ms: number): TelemetryEvent['durationBucket'] => {
+  if (ms < 100) return '0-100';
+  if (ms < 500) return '100-500';
+  if (ms < 2000) return '500-2000';
+  return '2000+';
+};
+
+const detectAIAgent = (): boolean =>
+  Boolean(
+    process.env.CLAUDECODE ||
```
aromko, line 37: 💡 `Boolean(process.env.X)` treats `"0"` / `"false"` as truthy. Edge case in practice (agents don't usually unset to falsy strings) but `process.env.X === '1'` would mirror the `MARIGOLD_TELEMETRY_DISABLED` / `DO_NOT_TRACK` checks in `isTelemetryDisabled`.

packages/cli/package.json
```diff
@@ -0,0 +1,64 @@
...
+  "files": [
+    "dist"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/marigold-ui/marigold",
+    "directory": "packages/cli"
+  },
+  "engines": {
+    "node": ">=22"
```
aromko, line 41: 💡 `"engines": { "node": ">=22" }` vs repo root `"node": "24.x"`. Almost certainly intentional (the CLI has to run on consumer machines that may not yet be on 24), but worth a one-line note in the PR body.

sarahgm:
@jim761 how can I test this locally not in marigold root
- I created a new folder in marigold and try to test it there it worked untill i wanted to try out the `marigold completion`

packages/cli/README.md
jim761: either this: node packages/cli/dist/bin/marigold.mjs docs Button ,prints markdown docs (with docs site running locally or against preview, add an .env.local that has MARIGOLD_DOCS_URL="https://marigold-docs-git-marigold-cli-v1-marigold.vercel.app/" written in it), or go to the cli folder and type pnpm link --global, that way you can use the commands like it will be used later. But the .env.local needs to be added bc it cannot used the prod until its merged, in the .env add the preview link or localhost

jim761:
@sarahgm go in the cli folder then type pnpm link --global, that way you can use it anywhere to test, to get rid of it type pnpm uninstall -g @marigold/cli, also add a .env.local with MARIGOLD_DOCS_URL="https://marigold-docs-git-marigold-cli-v1-marigold.vercel.app/" or localhost

sarahgm:
> @jim761 how can I test this locally not in marigold root
> 
> * I created a new folder in marigold and try to test it there it worked untill i wanted to try out the `marigold completion`

Found a way to test it. But I have some questions regarding the commands: 

From the commands: 
docs -> can we have here also a search command like in list?
docs -> what if i don’t know the section I need? How to use --section effective?

jim761:
@sarahgm i dont understand the first question, i think list is already the search command, to see which components exist. or for what do you want it.
For the second if you run marigold completion and follow the steps, you can press tab to see which section exists, maybe you also wanted this for the first question. otherwise section is mostly for ai to only get relevant information

sarahgm:
> @sarahgm i dont understand the first question, i think list is already the search command, to see which components exist. or for what do you want it. For the second if you run marigold completion and follow the steps, you can press tab to see which section exists, maybe you also wanted this for the first question. otherwise section is mostly for ai to only get relevant information

I mean for example i want to search admin and master mark usage, how would I do this? because I don't now the title of the page.

packages/cli/src/bin/marigold.ts
jim761: Good catch — fixed in `c7c231aae`. `runInit` is now dynamic-imported inside the `init` branch, so `@babel/parser`, `magic-string` and `@clack/prompts` stay off the docs/list hot path. After rebuild, the init bundle is a separate ~23 KB chunk (`init-*.mjs`) and the main `bin/marigold.mjs` dropped to ~7.3 KB; a cold `import()` of the bin now measures ~11 ms locally. Skipped the belt-and-braces lazy-load inside `init.ts` itself — single-level was enough to take the heavy deps off the agent hot path.

packages/cli/src/lib/telemetry.ts
jim761: `setTelemetryEnabled` now only mints when `enabled === true`. Verified that after `marigold telemetry disable` the config file is `{ "telemetryEnabled": false }` with no `anonymousId`. `anonymousId()` continues to lazy-mint inside `emit()` for the opt-in path.

packages/cli/src/lib/cache.ts
jim761: Fixed in `c7c231aae` — `writeEntry` now writes to `${file}.${process.pid}.tmp` and `renameSync`s into place. The `.pid` suffix prevents sibling processes writing the same cache key from racing on the temp file itself, and a tmp-cleanup catch covers the partial-write-then-throw case.

packages/cli/src/commands/init.ts
jim761: Fixed in `c7c231aae`. Added an `InitCancelError` (name-tagged so it crosses the dynamic-import boundary safely) thrown from both cancel paths in `runInit`. `bin/marigold.ts` catches by `.name` and sets exit code 130 (SIGINT convention). The existing `prompts.cancel('Aborted.')` still prints, so the UX is unchanged for humans but scripts can now distinguish abort from completion.

packages/cli/src/commands/list.ts
jim761: Fixed in `c7c231aae`. `loadManifest` now returns `{ manifest, cacheHit }` (new `LoadManifestResult` type), and the `list` branch in `bin/marigold.ts` threads it through to the telemetry emit. `loadManifestSync` (sync completion path) is unchanged since it doesn't carry a hit flag.

packages/cli/src/lib/format.ts
jim761: Pushing back on the renderer rewrite for 0.1. Per your own framing, `--format json` is the agent path and we just spent commit `c7c231aae` (item #1 above) shaving the cold-start path — pulling `marked-terminal` adds a ~200 KB dep tree to the CLI we just made lean. Compromise in `c7c231aae`: added a one-line note under `marigold docs` in the README that the default markdown render is best-effort and `--format json`/`plain` are the recommended paths for non-trivial docs. Happy to revisit if a "rich human terminal" use case shows up in telemetry.

packages/cli/src/lib/manifest.ts
jim761: Fixed in the PR body — the summary now reads "consumes the static `/manifest.json` built by `docs/scripts/build-manifest.mjs` (now emits categorized components, top-level `@marigold/components` `version`, and a `generatedAt` timestamp)." Also updated the test-plan checklist item that referenced `/api/manifest.json`.

packages/cli/src/lib/manifest.ts
jim761: Fixed in `c7c231aae`. Took option (b) — `formatList` already serializes `manifest.version` in JSON output and the agent use case benefits from "which version of Marigold did these docs come from?", so I wired them in rather than dropping. `docs/scripts/build-manifest.mjs` now reads `packages/components/package.json` and emits `version` + `generatedAt` (ISO) at the top of the static manifest; `transformManifest` sanitizes and threads them through. Verified locally: `curl /manifest.json | head` returns `{"version":"17.4.0","generatedAt":"2026-…","baseUrl":…}`.

packages/cli/src/lib/config.ts
jim761: Fixed in `c7c231aae` — went with "0 = disabled". `fetchWithCache` now short-circuits both `readEntry` and `writeEntry` when `ttl === 0`, so the cache directory stays empty rather than filling with entries the next call will immediately treat as stale. `cacheTtlMs()` itself is unchanged; the policy lives in `cache.ts`.

packages/cli/src/index.ts
jim761: Fixed in `c7c231aae` — added `/** @internal */` to `Manifest`, `ManifestPage`, `ManifestComponent`, `ManifestCategory` in `lib/manifest.ts`. Subpath export (`@marigold/cli/internal`) is the cleaner long-term shape but I've deferred it; the JSDoc is enough to keep tools that respect `@internal` from auto-completing these and to signal intent to outside consumers.

packages/cli/src/lib/format.ts
jim761: Applied in `c7c231aae`. Thanks.

packages/cli/src/bin/marigold.ts
jim761: Fixed in `c7c231aae`. Added a `SubcommandName` type in `lib/commands-spec.ts`; `bin/marigold.ts` now derives the telemetry filter from `TOP_LEVEL_NAMES` (minus `completion`, which is excluded from telemetry by design and returns early before this block anyway).

packages/cli/src/lib/telemetry.ts
jim761: Going to push back on this one. The variables in `detectAIAgent` (`CLAUDECODE`, `CURSOR_AGENT`, `VSCODE_AGENT`, `CODEX_SANDBOX`, `AI_AGENT`) are *presence markers* set by the agent harness — agents signal "I'm running" by setting the var to anything, and we don't control what value they pick (Claude Code uses `=1`, others use `=true`, `=claude`, or just `=`). `Boolean(...)` is the correct semantics here — strict `=== '1'` would silently break detection for any agent that doesn't happen to use `1`. The parity with `isTelemetryDisabled` doesn't apply: that function checks **explicit user opt-outs** (`MARIGOLD_TELEMETRY_DISABLED=1`), which we *do* control and where strict equality is right.

packages/cli/package.json
jim761: Intentional — added a line to the "Notes for reviewers" section of the PR body calling this out: "`engines.node` is `>=22` (vs the repo root's `24.x`) on purpose — consumer machines installing the CLI may not be on Node 24 yet."

docs/app/api/telemetry/route.ts
jim761: as a per-`anonymousId` daily quota using Upstash `INCR` + conditional `EXPIRE`: cap 1000/day, returns 429 when exceeded. INCR is atomic so concurrent CLI invocations can't race past the limit, and `EXPIRE` only fires on the first hit of the day so the TTL doesn't get pushed out forever. The legit upper bound is "one event per command invocation," so 1000/day leaves plenty of room for power users (CI auto-suppresses telemetry anyway) while still bounding abuse to a knowable share of the Upstash daily budget.

jim761:
> > @sarahgm i dont understand the first question, i think list is already the search command, to see which components exist. or for what do you want it. For the second if you run marigold completion and follow the steps, you can press tab to see which section exists, maybe you also wanted this for the first question. otherwise section is mostly for ai to only get relevant information
> 
> I mean for example i want to search admin and master mark usage, how would I do this? because I don't now the title of the page.

will look for a way to adress this issue

packages/cli/src/lib/edit-tsx.ts
jim761: Agreed the risk is real, but punting this to a follow-up: scoping the search to the default-exported function's return value means walking export → function → return → JSX, which is a meaningful chunk of code for what (per your own framing) is a worst-case "manual revert by the user" today. The wizard's diff preview also gives the user a chance to spot a bad rewrite before it's written. Filed DST follow-up: `🩹 marigold init` should scope JSX rewrites to the default export.

aromko: adressed in https://reservix.atlassian.net/browse/DST-1425

docs/app/api/telemetry/route.ts
aromko: Fixed in `b6759e3` — per-`anonymousId` daily quota via `INCR` + `EXPIRE` (1000/day, returns 429 over the limit). Resolving.

packages/cli/src/lib/telemetry.ts
aromko: Fixed — `sweepStaleTelemetryTmpFiles()` (`telemetry.ts:106-124`) runs on every `emit()` and unlinks `marigold-telemetry-*.json` older than 24h. Resolving.

packages/cli/src/lib/telemetry.ts
aromko: I am not sure about the agent thingy but let it for now. But with this snippet the value doesn't matter at all - what matters is whether the variable is set or not. Is this intended?

packages/cli/src/bin/marigold.ts
jim761: Holding here. A `NODE_ENV !== 'production'` gate doesn't really help — CLIs don't get `NODE_ENV` set by their callers, so it would either always-pass or always-skip depending on what we pick as default. An `fs.existsSync` precheck just replaces one syscall with two. The current try/catch is the right shape: zero cost when the file is absent (the common case in published copies), zero cost when it's present + readable (the dev case), and `.env.local` is excluded from `files` anyway so it can't ship.

packages/cli/src/commands/init.ts
jim761: Good catch — fixed in `0870b7c`.                                                                                                     
                                                                                                                                          
  `runInstall` now registers explicit `SIGINT`/`SIGTERM` handlers on `process` for the lifetime of the child, forwards them via           
  `child.kill(signal)`, and removes the listeners on `exit`/`error` so they don't leak across runs. Signal-driven exits are mapped to the 
  existing `InitCancelError`, so `bin/marigold.ts` keeps producing the conventional exit code 130 without any changes there.

packages/cli/src/lib/telemetry.ts
jim761: Yes this is intented, if the value exist it was used by an AI agent and the value doesn't matter, because every agent sets the value to something else. If the value doesn't exist at all, we will know it was a normal user.

jim761:
@sarahgm i wrote a follow up ticket for the search, for now its not included

packages/cli/src/lib/edit-tsx.ts
```diff
@@ -0,0 +1,298 @@
...
+
+  if (!target) {
+    return {
+      kind: 'skipped',
+      reason: 'could not find {children} JSX in layout',
+    };
+  }
+
+  const parentName = target.parent ? jsxOpeningName(target.parent) : null;
+  const alreadyWrapped =
```
sarahgm, line 184: 💡 **Dead `Providers` import when layout already wraps with `<MarigoldProvider>` directly**

If a user has manually wrapped `{children}` in `<MarigoldProvider>` before running `marigold init` (rare but reachable via partial manual setup), this branch falls through inconsistently:

- `alreadyWrapped` is `true` (matches `'MarigoldProvider'` on line 185)
- `hasImport` is `false` (only checks for the `Providers` named import on line 187)
- The early-return on line 193 requires *both*, so it doesn't fire
- Line 200 inserts `import { Providers } from './providers';`
- Line 210 skips the wrap because `alreadyWrapped` is true

Result: the user's file ends up with an unused `Providers` import and no actual change to the JSX.

Easiest fix is to gate the import insertion on `!alreadyWrapped`:

```suggestion
  if (!hasImport && !alreadyWrapped) {
```

Edge case, low priority — typical `init` flow on a fresh project isn't affected.

---

packages/cli/src/commands/completion.test.ts
```diff
@@ -0,0 +1,228 @@
...
+
+afterEach(() => {
+  if (originalCacheDir === undefined) delete process.env.MARIGOLD_CACHE_DIR;
+  else process.env.MARIGOLD_CACHE_DIR = originalCacheDir;
+  fs.rmSync(tmpDir, { recursive: true, force: true });
+});
+
+describe('computeSuggestions — static surface', () => {
+  it('completes top-level subcommands on empty input', () => {
+    const out = computeSuggestions(['']);
```
aromko, line 64: In general you could format the test to our guidelines: https://reservix.atlassian.net/wiki/spaces/DST/pages/3626336415/Refactoring+writing+tests+in+storybook#Structure-of-a-test

packages/cli/src/commands/init.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    `import { createRoot } from 'react-dom/client';
+import App from './App.tsx';
+
+createRoot(document.getElementById('root')!).render(<App />);
+`
+  );
+};
+
+describe('runInit (integration, skipInstall)', () => {
+  it('wires up a Next.js project end-to-end', async () => {
```
aromko: This test is also not nice to read (https://reservix.atlassian.net/wiki/spaces/DST/pages/3626336415/Refactoring+writing+tests+in+storybook#Structure-of-a-test). At least you shit split up those tests or adding steps with AAA.

packages/cli/src/lib/detect-project.test.ts
```diff
@@ -0,0 +1,69 @@
...
+import os from 'node:os';
+import path from 'node:path';
+import { detectProject, installCommand } from './detect-project.js';
+
+const makeTmpDir = () =>
+  fs.mkdtempSync(path.join(os.tmpdir(), 'marigold-detect-'));
+
+describe('detectProject', () => {
+  it('detects Next.js + pnpm', () => {
+    const dir = makeTmpDir();
```
aromko: This can be optimized. Use beforeEach, afterEach for repetitive stuff.

packages/cli/src/lib/edit-css.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+afterEach(() => {
+  fs.rmSync(dir, { recursive: true, force: true });
+});
+
+describe('editCss', () => {
+  it('creates app/globals.css for a fresh Next.js project', () => {
+    const result = editCss({ cwd: dir, framework: 'nextjs' });
+    expect(result.kind).toBe('edited');
+    if (result.kind !== 'edited') return;
```
aromko: No - just no. Never ever use branches inside tests. Split it up and test the cases isolated.

aromko (commented):
@jim761 I didn't comment all places in the test files so please just go through all.

sebald:
## Follow-up thought: consider migrating to a CLI framework once the next command lands

**This is not a blocker — the PR is fine to ship as-is.** The package is well-tested, the library/CLI separation is clean (typed `runX(options)` functions in `commands/*` with the argv shell in `bin/marigold.ts`), and the surface is small enough that a later refactor is cheap. Filing this as a follow-up consideration for when `marigold doctor` / `add` / `search` / `examples` land (DST-1265, DST-1421, DST-1445, DST-1446).

### What I'd flag in the current shape

1. **`bin/marigold.ts` is a ~300-line if/else dispatcher.** Each command branch hand-rolls its own `parseArgs` block, builds a `telemetryArgs` record, validates `format`/`section` via type guards, and re-maps to the runner's option type. ~40 lines of boilerplate per command, scaling linearly with each new command.
2. **Command surface duplicated across three files.** `parseArgs` blocks in `bin/marigold.ts`, `lib/commands-spec.ts` (for completion), and the hand-formatted help string. The comment at the top of `commands-spec.ts` literally says *"Keep in sync with the parseArgs blocks in bin/marigold.ts and the help template."* Drift is inevitable as the surface grows.
3. **No per-subcommand `--help`.** `marigold docs --help` falls through to `parseArgs`; only the root help template exists.
4. **No typo suggestions / aliases.** `marigold docs Buttom` just errors.
5. **Env-var docs only live in the help string.** `MARIGOLD_DOCS_URL`, `MARIGOLD_CACHE_TTL_MS`, telemetry vars are scattered with no central config schema.
6. **`--format json` partial coverage.** `telemetry` and `completion` don't honor it. Future commands like `doctor` will want it.
7. **No `--quiet` / `--verbose` / central logger.** `picocolors` handles `NO_COLOR`, but there's no log level.

### What a framework would actually fix

Using **citty** (UnJS, tiny, ESM/TS-native, declarative `defineCommand`) as the reference:

| # | Fixed by framework? |
|---|---|
| 1. Dispatcher boilerplate | **Yes.** Each command becomes a `defineCommand({ args, run })` file. Dispatcher disappears; telemetry wrapping moves into a single hook. |
| 2. Triple-source command spec | **Yes.** The command definition *is* the spec — help and parser derive from it. `commands-spec.ts` can be deleted. |
| 3. Per-command help | **Yes.** `marigold docs --help` works out of the box, generated from the args definition. |
| 4. Did-you-mean / aliases | **Partial.** citty has aliases; typo suggestions aren't built in (add `didyoumean2`, ~2 kB). cac and clipanion have it natively. |
| 5. Central env-var/config | **No.** Frameworks parse argv, not env. Still want a small `lib/config.ts`. |
| 6. Uniform `--format json` | **No.** Design discipline issue. A shared `output(result, format)` helper fixes it regardless of framework. |
| 7. `--quiet` / `--verbose` / logger | **Partial.** Global flags become declarative and inherited; the logger itself you still write once. |

Bonus that also goes away: `isOutputFormat` / `isSection` / `isTelemetrySub` type guards (replaced by `enum`-style arg declarations), the hand-written `fail()` + `CliError` validation path, the `-v` / `--version` and `-h` / `--help` short-circuits in `main()`.

### Framework alternatives considered

- **citty** — best fit. Tiny, ESM, TS-first, declarative, same ecosystem as `c12`/`unbuild`. Used by Nuxt/Nitro CLIs.
- **cac** — solid, minimal. Has did-you-mean.
- **commander** — works, but heavier API, CJS legacy, less TS-idiomatic.
- **clipanion** — powerful (built-in completion, did-you-mean), but the class+decorator model is heavy for this size.
- **oclif** — overkill, don't.

### Things to keep regardless

- Library exports (`runDocs`, `runList`, `getComponentDocs`) — the framework should wrap, not replace, the typed `runX(options)` functions.
- The lazy-load trick for `init` — citty's `setup`/`run` separation supports this via dynamic import inside `run`.
- `lib/completion-scripts.ts` + `__complete` flow — citty doesn't generate completion, and dynamic component-name completion from the manifest is custom anyway. This is one of the cleaner pieces of the current code.

### Suggested timing

Land this PR. When the next command (`doctor` / `add` / `search`) starts, do the citty migration first — that's the inflection point where the duplication starts to bite. The conversion is mechanical because the runners are already isolated from argv parsing.

Two cheap improvements you could land *without* a framework in the meantime:
- Move each command's `parseArgs` block into its `commands/<x>.ts` file (export a `parse(argv)`). Dispatcher shrinks to a name→module map.
- Derive the help text from `SUBCOMMANDS`. Kills duplication #2 today.

packages/cli/src/lib/edit-tsx.ts
jim761: Good catch — confirmed the edge case. When the layout is already manually wrapped with `<MarigoldProvider>`, `alreadyWrapped=true` but `hasImport=false`, so the early-return at line 193 doesn't fire and we end up inserting a dead `Providers` import without wrapping anything.

Applied your suggestion in `2e9527d47`: gated the import insertion on `!alreadyWrapped`. Also tightened the bottom of `editNextLayout` to return `unchanged` when nothing actually got pushed to `changes[]`, so the function never reports `edited` with an empty edit set. Added a regression test in `edit-tsx.test.ts`.

packages/cli/src/commands/completion.test.ts
jim761: Done in `2e9527d47`. Switched to `test()`, dropped the `vitest` imports (relying on `globals: true`), added AAA blank-line spacing, and split `'exits 2 for unsupported shells'` into two tests (powershell + cmd) since those were two independent scenarios under one `it`.

packages/cli/src/commands/init.test.ts
jim761: Reformatted with AAA blank lines rather than splitting. The two `'wires up …'` tests intentionally verify the full end-to-end contract of `runInit` in one go — splitting them by file (layout / providers / css / postcss) would mean six identical `await runInit(...)` calls per scenario, and would lose the "single run produces consistent project state" semantic that's the whole point of the test. The blank-line separation between each file-assertion group should cover the readability ask. Let me know if you'd still prefer it split.

packages/cli/src/lib/detect-project.test.ts
jim761: Done in `2e9527d47`. Hoisted tmp-dir creation/cleanup to `beforeEach`/`afterEach`, dropped the `makeTmpDir` helper, switched to `test()`, AAA blank lines. The `installCommand` tests don't touch the filesystem so they stay in their own `describe` without setup.

packages/cli/src/lib/edit-css.test.ts
jim761: Agreed. Replaced the `if (result.kind !== 'edited') return;` pattern with a colocated `assertEdited` helper that throws when the kind is wrong — so the test fails loudly instead of silently skipping the rest of the body. Applied the same pattern in `edit-tsx.test.ts` and `edit-vite-config.test.ts`. No `if`/`else` left in any CLI test.

aromko (approved):

sarahgm (approved):

PR #5233, opened 2026-03-23
feat([DST-1166]): mcp server integration

sinan-rsvx:
# Description
Adds an MCP (Model Context Protocol) server to the Marigold docs at /mcp. Coding agents like Cursor, VS Code Copilot, and Claude Code can connect to it and semantically search the design system documentation using the search_docs tool.
An ETL pipeline (pnpm build:embeddings) chunks the MDX docs and generates vector embeddings via AWS Bedrock Titan v2, stored alongside the app as chunks_search.json. The MCP route at /mcp uses Streamable HTTP transport and performs cosine similarity search over these precomputed vectors at query time.
Authentication is handled via Keycloak

# Test Instructions:

Add the server to your MCP client config and test it through your IDE or AI agent.

**VS Code** (`.vscode/mcp.json`):
```json
{
  "servers": {
    "marigold-docs": {
      "url": "https://marigold-docs-git-dst-1166-mcp-server-integration-marigold.vercel.app/mcp"
    }
  }
}
```
or 
```bash
claude mcp add --transport http \
  --client-id dst-marigold-docs-mcp \
  --callback-port 6749 \
  marigold-docs https://marigold-docs-git-dst-1166-mcp-server-integration-marigold.vercel.app/mcp
```
A browser login via Keycloak will open automatically. After authenticating, try asking something like: "How do I use the Button component in Marigold?". It should invoke search_docs and return relevant docs.

## Test Prompts

**Prompt:**
```
How do I use the Button component in Marigold?
```
**Expected:** Props overview (`variant`, `size`, `disabled`), code examples, link to docs

**Prompt:**
```
How do I build a form with validation in Marigold?
```
**Expected:** Usage of `Form`, `TextField`, `Checkbox` components, validation logic, submit handling

**Prompt:**
```
What layout components does Marigold provide?
```
**Expected:** `Stack`, `Inline`, `Box`, `Grid` – each with spacing/alignment props and examples

**Prompt:**
```
How does theming work in Marigold?
```
**Expected:** Theme provider setup, token system, how to create or override custom themes

**Prompt:**
```
How do I display tabular data with Marigold?
```
**Expected:** `Table` component, column definitions, sorting, pagination

**Prompt:**
```
How do I show a modal dialog in Marigold?
```
**Expected:** `Dialog`/`Modal` component, open/close state management, trigger pattern

**Prompt:**
```
How do I use icons in Marigold?
```
**Expected:** Icon component usage, available icon names, size and color options

**Prompt:**
```
[...]

PR #5368, opened 2026-04-27
Revert "fix(docs): drop force-static on manifest route (#5365)"

sebald:
## Summary

Reverts #5365.

The `force-dynamic` swap was treating a symptom: the original Vercel
`Preview – marigold-docs` build failure on beta-release was reported as a
prerender error on `/api/manifest.json`, but production main deploys with
the same code (`force-static` + `NextResponse.json()`) had been succeeding
the entire time. The latest beta-release preview deploy (`7389e83ff`,
which has the `force-dynamic` fix merged in via this PR) is **still
failing** — confirming the fix wasn't addressing the real cause.

Suspected real root cause: env-var difference between the Vercel Production
and Preview environments after the MCP server integration (#5233) merged
to main. See `docs-vercel-preview-regression.md` for the full
investigation.

## Test plan

- [ ] Confirm production main deploy continues to succeed with the route
      back on `force-static`
- [ ] Investigate the underlying env-var/build issue separately and open a
      proper fix once the cause is understood

### DIFF
--- a/docs/app/api/manifest.json/route.ts
+++ b/docs/app/api/manifest.json/route.ts
@@ -1,9 +1,7 @@
 import { source } from '@/lib/source';
+import { NextResponse } from 'next/server';
 
-// `force-static` triggers a Node 22+ undici bug during Next.js prerender:
-// `TypeError: Cannot read private member #state` (undici#4290, node#58814).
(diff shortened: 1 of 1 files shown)


sebald:
Superseded by a single combined PR — see follow-up.

PR #5371, opened 2026-04-27
fix(docs): generate /api/manifest.json and /api/md/* as static assets

sebald:
## Summary

Both `/api/manifest.json` and `/api/md/[...slug]` were Next.js Route Handlers exporting `dynamic = 'force-static'`. On Node 22+ the bundled undici uses private `#state` fields on `Response`, and Next 16 wraps any `Response` returned from a `force-static` handler in a Proxy during static export. `Reflect.get` on that Proxy can't forward the private-field access, undici throws `Cannot read private member #state` ([undici#4290](https://github.com/nodejs/undici/issues/4290), [node#58814](https://github.com/nodejs/node/issues/58814)), and the prerender worker dies. The bug is **non-deterministic** — same code sometimes succeeds and sometimes fails, which is why production main deploys looked healthy while preview deploys flaked on beta-release.

#5365 only flipped `/api/manifest.json` to `force-dynamic`. The next merge into beta-release fired the same crash on `/api/md/components/.../selectlist.md`. `force-dynamic` is also a workaround anyway: I measured the md route at ~270ms cold on first request per region per deploy due to ts-morph init in the MDX parser pipeline. With ~130 routes × multiple Vercel regions × every deploy, that's a lot of user-visible cold renders we don't have today.

## What changed

- **`docs/scripts/build-manifest.mjs`** — globs `content/**/*.mdx`, parses frontmatter, writes `docs/public/api/manifest.json`. Same shape and filtering the route handler had (excludes `releases/`, `__internal__/`).
- **`docs/scripts/build-md.ts`** — iterates `content/**/*.mdx` through the existing `parseMdxToMarkdown` pipeline (`p-limit` concurrency = 4 to balance ts-morph init cost), writes `docs/public/api/md/<slug>.md`.
- **`docs/package.json`** — wires `build:manifest` and `build:md` into the `build` and `dev` scripts.
- **`docs/.gitignore`** — `public/api` is gitignored (build artifact).
- **Deletes** `docs/app/api/manifest.json/route.ts` and `docs/app/api/md/[...slug]/route.ts`.

Vercel serves the assets straight from the CDN — same end-state as `force-static` was aiming for (built once per deploy, served from the edge forever) without going through Next's prerender pipeline at all. No more `#state` proxy bug.

## Verification

- `pnpm --filter @marigold/docs run build:manifest` → 101 pages, output verified
- `pnpm --filter @marigold/docs run build:md` → 129 markdown files in 10.2s (1 skipped — root index)
- Sample diff against current production `/api/md/components/actions/button.md`: only cosmetic difference (variant union order from ts-morph)

## Test plan

- [ ] Vercel preview build completes (no `#state` error possible — no `force-static` route handlers)
- [ ] `curl -I https://<preview>/api/manifest.json` returns 200 with default static-asset cache headers
- [ ] `curl https://<preview>/components/actions/button.md` returns the markdown (via existing `/:path*.md → /api/md/:path.md` rewrite, falling through to the new static file)
- [ ] Build log shows neither `/api/manifest.json` nor `/api/md/[...slug]` in the route table

## Notes

Closes the saga of:
- #5365 (force-dynamic for manifest, merged) — its effect is superseded by deleting the route handler entirely.
- #5370 (force-dynamic for md route, closed) — replaced by this approach.


sebald:
Superseded by #5372, which carries the same build-script fix plus the fumadocs-typescript generator consolidation and the slug-mirror URL migration.

PR #5372, opened 2026-04-28
fix(docs): pre-compute prop tables and serve markdown/manifest as static assets

sebald:
## Issue

The Vercel preview build for the docs (`pnpm --filter @marigold/docs build`) was failing during prerender:

```
Error occurred prerendering page "/api/manifest.json".
TypeError: Cannot read private member #state from an object whose class did not declare it
    at Reflect.get (<anonymous>)
Export encountered an error on /api/manifest.json/route, exiting the build.
```

Same crash subsequently surfaced on `/api/md/[...slug]/route` once the manifest path was patched.

## Root cause

Three things compose into one bug:

1. **Node 22+** changed undici's `Response` to use **private class fields** (`#state`) instead of WeakMap-backed state. See [undici#4290](https://github.com/nodejs/undici/issues/4290), [node#58814](https://github.com/nodejs/node/issues/58814).
2. **Next.js 16's `force-static` route-handler prerender** wraps the returned `Response` in a JavaScript **`Proxy`** to inspect/transform it for static export.
3. The Proxy's handlers use **`Reflect.get`** to forward property access. `Reflect.get` cannot forward access to **private fields** declared on a class the receiver isn't an instance of — JavaScript's private-field semantics intentionally prevent that.

So when Next.js's static exporter touches the proxied Response (e.g. to read its body or status during prerender), undici's getters call `Reflect.get(target, '#state')`, which throws because the Proxy receiver isn't a `Response` instance and can't see the field. The route handler itself runs fine — the crash is on the way out, in the framework's prerender wrapper.

The bug is **environment-specific**: only triggers on Node 22+ × `force-static` route handlers × Vercel build. Local builds with older Node never tripped it; CI on Vercel running the Node 24 default did.

`revalidate = false` (what fumadocs uses for similar endpoints) does **not** escape the bug — same prerender Proxy wrapping. Verified via Next.js 16 source and tracker; not just inferred.

## Fix

The data both endpoints serve (per-page markdown, page-index manifest) is **fully static at build time**. Rather than re-introduce route handlers and dance around the prerender Proxy with `await connection()` / cache headers / `outputFileTracingIncludes`, they're now plain files in `public/` served directly by Next.js. **No route handler runs, no Proxy ever wraps the response — the bug surface is removed entirely.**

While the build pipeline was getting touched, ts-morph was eliminated from the runtime entirely:

- `<AutoTypeTable package="system">` (latent bug — empty markdown tables for SVG / Icon) is fixed.
- One shared `fumadocs-typescript` generator + `Project` between `<AutoTypeTable>` UI and the markdown build, so HTML and Markdown prop tables are generated through identical pipelines.
- Dead `resolve-design-tokens` plugin (crashed every markdown build with ENOENT after a content-rewrite removed its target files) is removed.
- ts-morph runs once at `pnpm build:types`, writing `.registry/props.json`. Both the `<AutoTypeTable>` RSC and the markdown plugin read from that JSON. ts-morph is now a devDependency only.

## What changed

| Concern | Before | After |
|---|---|---|
| Manifest delivery | `app/api/manifest.json/route.ts` (force-static, broken) | Static asset `public/manifest.json`, served at `/manifest.json` |
| Per-page MD delivery | `app/api/md/[...slug]/route.ts` (force-static, broken) | Static asset `public/<slug>.md`, served at `/<slug>.md` |
| Public MD URL | `/api/md/<slug>.md` (with rewrite from `/<slug>.md`) | `/<slug>.md` (no rewrite) |
| Public manifest URL | `/api/manifest.json` | `/manifest.json` |
| Manifest entry `url` field | `/api/md/<slug>.md` | `/<slug>.md` |
| Prop-table type extraction | Runtime ts-morph at every render of `<AutoTypeTable>`, plus a separate ts-morph instance in the markdown build | One ts-morph pass at `pnpm build:types`. JSON read at request/build time. ts-morph is `devDependency` only |
| `<AutoTypeTable package="system">` markdown output | Empty (latent bug) | Real props table |
| `<AppearanceTable>` markdown output | Custom ts-morph regex over `*Props`. Diverged from HTML for Menu / LinkButton / Panel / Link | Reads same `appearances` map as HTML. Aligned by construction |
| HTML vs Markdown type formatting | HTML used `getSimpleForm` + transform; MD used a hand-rolled `simplifyType()`. Diverged | Both go through fumadocs's `getSimpleForm` and `autoTypeTableTransform` (applied at build time, baked into JSON) |
| `resolve-design-tokens` plugin | Eagerly read deleted `ui/ColorTable.tsx` / `ui/ColorTokens.tsx`, crashed every markdown build | Deleted |
| `next.config.mjs` `serverExternalPackages` | `['ts-morph', 'typescript']` (forced runtime fallback) | Removed (no runtime ts-morph) |
| Route handlers / rewrites / `outputFileTracingIncludes` for these endpoints | All present | All removed |

## Architecture

```
pnpm build (or dev)
  ├─ build:registry      → .registry/demos.{tsx,json}     (unchanged)
  ├─ build:types  NEW    → .registry/props.json           (one ts-morph pass)
  ├─ build:changelog                                       (unchanged)
  ├─ build:manifest      → public/manifest.json           (static asset)
  ├─ build:md            → public/<slug>.md               (static assets)
  └─ next build
```

At request time:
- `<AutoTypeTable>` reads `.registry/props.json`, renders via `<TypeTable>` from `fumadocs-ui`. The fumadocs `/ui` entry doesn't import ts-morph, so ts-morph is fully out of the runtime bundle.
- `<AppearanceTable>` reads `@marigold/theme-rui/appearances` directly.
- `/manifest.json` and `/<slug>.md` are served from `public/` by Next.js as static assets — no function invocation, no prerender Proxy.

## Behavioral diffs in markdown output

`pnpm build:md` was run; the relevant changes:

- **`components/content/svg.md`, `components/content/icon.md`** — gained ~480-line props tables each (the `package="system"` bug fix).
- **`components/overlay/menu.md`** — `variant` now `default | destructive | ghost` (gained `destructive`, which lives on the `MenuItem` slot in cva and was missed by the old ts-morph regex).
- **`components/actions/link-button.md`** — `variant` now includes `destructive-ghost` (its `Props` interface omits it but the theme defines it).
- **`components/layout/panel.md`** — `size` now `default | form` (gained `default`).
- **`components/actions/link.md`** — `size` now `default | small` (was empty).
- **All AppearanceTable rows** — description column renders as plain text instead of `inlineCode` (descriptions aren't code; cosmetic alignment with HTML).
- **All AutoTypeTable prop rows** — `simplifiedType` formatting from `getSimpleForm` + `autoTypeTableTransform` (e.g. `MouseEventHandler<...>` → `function`, design-system aliases preserved). Matches what HTML already shows.

## Test plan

- [x] `pnpm --filter @marigold/docs lint` — 0 errors.
- [x] `pnpm --filter @marigold/docs types:check` — only the pre-existing `Panel` errors on `beta-release` (unrelated; Panel is used in MDX content but not yet exported from `@marigold/components`).
- [x] `pnpm --filter @marigold/docs build:registry && build:types && build:manifest && build:md` — all 121 prop tables built (cold ~17s, warm ~0.3s); 128 markdown files built (~8s).
- [x] Local dev server: `curl /manifest.json` → 200 `application/json`. `curl /components/actions/button.md` → 200 `text/markdown`. `curl /components/content/svg.md | wc -l` → 556 (the `package="system"` fix). All three are served as static assets — no route handler invoked.
- [ ] Vercel preview build succeeds (no route handlers means the Node 22+ undici Proxy bug can no longer be triggered).
- [ ] After deploy, verify `https://<preview>/manifest.json` and `https://<preview>/components/actions/button.md` resolve from the CDN with correct content types.


sebald:
## Why did this start firing now?

Not because Vercel changed anything — we did it ourselves, in three commits spread across nine months. The bug needs three ingredients stacked at once, and the third one only landed in late March:

| Date | Change | Effect |
|---|---|---|
| **2025-07-02** | Bumped `engines.node: 22.x` ([#4671](https://github.com/marigold-ui/marigold/pull/4671)) | Vercel switched to Node 22 → undici's `Response` now uses private `#state` ❶ |
| **2025-10-27** | Next.js 16 upgrade ([#4907](https://github.com/marigold-ui/marigold/pull/4907)) | New prerender code path that wraps `Response` in a `Proxy` ❷ |
| **2026-02-18** | Bumped `engines.node: 24.x` ([#5156](https://github.com/marigold-ui/marigold/pull/5156)) | Cosmetic — Node 22 already had the bug surface |
| **2026-03-31** | Added `app/api/manifest.json/route.ts` with `force-static` ([#5243](https://github.com/marigold-ui/marigold/pull/5243)) | Final ingredient — gave the prerender Proxy something to wrap ❸ |
| **2026-04-15** | First Vercel build with all three present | 💥 |

**Cause and effect:**

- The **bug surface** has existed since July 2025 (Node ≥22 + undici's private `#state`). Vercel honors `engines.node` in `package.json`, so that's when our builds moved off Node 20.
- The **Proxy wrapper** arrived in Oct 2025 (Next.js 16).
- But neither triggers a crash on its own — you need a `force-static` route handler returning a `Response` so the prerender Proxy has something to wrap. We didn't have one until [#5243](https://github.com/marigold-ui/marigold/pull/5243) added `/api/manifest.json` for AI discovery in late March.

So the question "did Vercel switch to Node 24 recently?" is the wrong frame — Node 22 in July 2025 was already enough to load the gun. What changed two weeks ago is **we pulled the trigger** by adding our first prerendered route handler that returns `Response.json(...)`.

(Bumping to Node 24 in Feb didn't help, but it didn't cause it either; the same crash would have eventually fired on Node 22 the moment we added a `force-static` route.)

sarahgm (approved):

PR #5377, opened 2026-04-29
fix([DST-1295]): continuous click area for CheckboxGroup and RadioGroup

sebald:
## Summary

- Replace the `gap` between `CheckboxGroup` / `RadioGroup` items with per-item padding so the full space between items is clickable. Vertical items now meet the 24 px target-size minimum; horizontal spacing keeps visual parity. Standalone `Checkbox` is unaffected.
- Align the row layout from `items-center` → `items-start` so the icon stays on the first line when the label wraps to multiple lines (instead of drifting to the geometric center of a multi-line label block).
- For `Radio`, the label slot now also uses `leading-4` to match `Checkbox`.

Jira: [DST-1295](https://reservix.atlassian.net/browse/DST-1295)

## Test plan

- [x] `pnpm test:unit` — 30/30 passing for Checkbox + Radio
- [x] `pnpm test:sb` — story tests passing
- [x] `pnpm lint` clean
- [ ] Visual regression review via Chromatic (`/vrt`)
- [ ] Manual check: vertical click between items registers on the closest item; horizontal items keep visual spacing; multi-line labels keep icon aligned to the first line

[DST-1295]: https://reservix.atlassian.net/browse/DST-1295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/Checkbox/CheckboxGroup.test.tsx
```diff
@@ -131,17 +131,6 @@ test('accepts error message', () => {
   expect(screen.getByText('This is an error')).toBeInTheDocument();
 });
```
aromko, line 132: 🟡 **Removed test has no replacement.**

The `horiziontal orientation style` test that lived here was correctly deleted because it asserted on the now-gone `gap-[1.5ch]`. But nothing replaces it — there's no longer any test guarding that horizontal mode renders `flex-row`, or that horizontal items pick up `px-1.5`. Worth keeping a slimmed-down version:

```suggestion
});

test('horizontal orientation applies flex-row and per-item padding', () => {
  render(<Basic.Component orientation="horizontal" />);
  const presentation = screen
    .getAllByRole('presentation')
    .filter(el => el.getAttribute('data-orientation') === 'horizontal');

  expect(presentation[0].className).toContain('flex-row');
});
```

.changeset/dst-1295-checkbox-radio-group-padding.md
```diff
@@ -0,0 +1,8 @@
+---
+'@marigold/components': patch
+'@marigold/theme-rui': patch
+---
+
```
aromko, line 5: 🟡 **Changeset is missing the team's title-line convention.**

Recent merged changesets all start with a `<type>([DST-XX]): <summary>` line (`refa([DST-1281]):`, `feat(DST-1246):`, `docs(DST-1385):`, etc.). This one jumps straight into the body. Suggested:

```suggestion
fix([DST-1295]): replace `gap` between `CheckboxGroup` and `RadioGroup` items with per-item padding so the full space between items is clickable. Vertical items now meet the 24px target-size minimum; horizontal spacing keeps visual parity. Standalone `Checkbox` is unaffected.
```

themes/theme-rui/src/components/Checkbox.styles.ts
```diff
@@ -14,9 +14,11 @@ export const Checkbox: ThemeComponent<'Checkbox'> = {
   }),
   container: cva({
     base: [
-      'grid grid-cols-[auto_1fr] gap-x-2 items-center',
+      'grid grid-cols-[auto_1fr] gap-x-2 items-start',
       'cursor-pointer read-only:cursor-default',
       'group-data-[booleanfield]/booleanfield:grid-cols-subgrid group-data-[booleanfield]/booleanfield:col-span-full',
+      'group-data-[orientation=vertical]/checkboxgroup:py-1',
+      'group-data-[orientation=horizontal]/checkboxgroup:px-1.5',
```
aromko, line 21: 💡 **Consider adding a test for the new modifier classes.**

The whole point of these `group-data-[orientation=…]/checkboxgroup:` modifiers is that an item gets padding **only when** wrapped in a group of the matching orientation. The existing Checkbox snapshot tests render standalone Checkboxes, so they never exercise these branches — a regression that silently strips the modifiers would pass.

A short test wrapping a Checkbox in `<CheckboxGroup orientation="horizontal">` and asserting `px-1.5` on the rendered container (and the same for vertical / `py-1`) would lock the contract.

Same idea applies to `Radio.styles.ts` for `radiogroup`.

packages/components/src/Radio/Radio.tsx
```diff
@@ -63,7 +63,7 @@ const _Radio = ({
       ref={ref}
       className={cn(
         'group/radio',
-        'relative flex items-center gap-[1ch]',
+        'relative flex items-start gap-[1ch]',
```
aromko: 💡 **Pre-existing inconsistency, not introduced here:** Radio uses inline `gap-[1ch]` while Checkbox's container uses theme-driven `gap-x-2`. Now that the alignment + leading is being unified across the two components, it would be nice to also align the icon-to-label gap (and ideally move it into the theme like Checkbox does). Out of scope for this PR — possible follow-up.

aromko (commented):
## Automated Code Review

### Overview
| Field | Value |
|-------|-------|
| Linked Ticket | [DST-1295](https://reservix.atlassian.net/browse/DST-1295) — In Review |
| CI | ✅ All passing (typecheck, lint, unit, storybook) |

### Summary

Clean CSS-only refactor: the `gap` → per-item padding swap correctly addresses both the Fitts's Law dead-zone problem and WCAG 2.5.8 target-size compliance. The new `group/radiogroup` class brings RadioGroup in line with how CheckboxGroup already exposes itself for descendant theming, and `items-start` + `leading-4` consistency between Checkbox and Radio is a nice touch.

No React structural changes — purely class-name churn — so no perf/render concerns.

### Issues

#### 🟡 Warnings (Should Fix)

1. **Removed orientation test has no replacement** — see inline on `CheckboxGroup.test.tsx`.
2. **Changeset doesn't follow team convention** — see inline on the changeset file.

#### 💡 Suggestions (Nice to Have)

3. **No test for the new padding modifiers** — see inline on `Checkbox.styles.ts`.
4. **Pre-existing `gap-[1ch]` vs `gap-x-2` inconsistency** — see inline on `Radio.tsx`.

### Recommendation

**Comment** — Two minor warnings and two follow-up suggestions. Nothing blocking; LGTM once warnings are addressed (or acknowledged).

---

aromko (approved):

PR #5379, opened 2026-04-29
docs(DST-1308): consolidate icon documentation onto the Iconography page

sarahgm:
# Description

Consolidate the icon documentation into a single page. The former `/components/content/icons` page is removed; its engineering content moves into a new `Using icons in code` section on `/foundations/iconography`, so designers and developers reach principles, catalog, and API in one place.

## What changed

- **One-page model**: `/foundations/iconography` now covers Principles → Catalog → Using icons in code.
- **New engineering section** `## Using icons in code` on the foundations page:
  - Installation + tree-shakeable note
  - Importing (Lucide proxy) + brief Brand icons mention pointing back to the catalog
  - Sizing demo (16/20/24/32 px)
  - Color: `currentColor` + `className` recommended; explicit warning that the `color` prop is literal CSS, not a token
  - Filled brand icons: `createFilledIcon` couples `fill` and `color` to keep solid shapes a single tint.
  - Accessibility: auto `aria-hidden` behavior + `aria-label` opt-out
  - Refs (forwarded to `SVGSVGElement`)
  - Migration link to the changeset
- **Three new demos** under `docs/content/foundations/iconography/`: `icon-sizes.demo.tsx`, `icon-className-vs-color.demo.tsx`, `icon-filled.demo.tsx`.
- **`/components/content/icons` deleted** and removed from `docs/content/components/content/meta.json`. The SVG page TeaserList link is updated to point at the consolidated page.
- **Catalog accuracy**: added `FileText` (Reports) to Info and `TwitterX` to Social so every legacy icon from the design assignment has a target.
- **Migration table changeset**: `.changeset/iconography-docs.md` (`@marigold/docs` patch) contains the full old → new mapping by section (UI / Info / Action / Ticketing / User / Social).

## Why a foundations page (not a components page)

The catalog is the largest and most-used part of the icon documentation, and it belongs in foundations. The remaining engineering content (~80 lines) fits comfortably as a final section. Two pages added cross-link maintenance and tempted duplication on sizing, color, and accessibility.

# Test Instructions:

1. `cd docs && pnpm dev`, open http://localhost:3000/foundations/iconography
2. Verify the page loads with three top-level sections in order: Principles, Catalog, Using icons in code.
3. Inside `Using icons in code`, confirm all three demos render: sizes row, color (className vs color prop), filled brand icons (default / fill).
4. Click each in-page anchor link (`#principles`, `#catalog`, `#visual-language-and-sizing`, `#touch-targets`, `#using-icons-in-code`) and confirm they jump correctly.
5. Click the [Token Overview](/foundations/token-overview) link in the Color subsection — it should resolve.
6. Visit http://localhost:3000/components/content/svg → scroll to Related → click the "Icons" tile → it should land on `/foundations/iconography`.
7. `pnpm types:check` in `docs/` — should pass.

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:


sarahgm:
Would change the target to beta-relase if the other one is merged

jim761 (dismissed):

aromko (commented):
## Automated Code Review

Docs-only PR consolidating two icon pages into `/foundations/iconography`. CI is green and visual regression is not needed (no `packages/components`, `packages/system`, `themes/`, `.storybook`, `*.stories.tsx`, or `*.styles.ts` touched).

### Summary of findings

No critical blockers. The main concerns are around **migration-table accuracy** in the changeset, since this becomes the public release-notes guidance for consumers upgrading.

**Warnings (should fix):**

1. The `Old` column in the migration table mixes kebab-case and lowercase (`arrow-down`, `search`, `accessibility`, `id`, `twitter`, …), but the actual `@marigold/icons` exports in `packages/icons/src/index.ts` are PascalCase (`ArrowDown`, `Search`, `Accessible`, `Id`, `Twitter`). Anyone grep'ing their imports for the old name won't find anything.
2. Seven legacy icons that exist in `@marigold/icons` have **no row** in the migration table — `Close`, `PauseAlt`, `PlayAlt`, `Upload`, `Phone`, `Print`, `Whatsapp`. The PR description says "every legacy icon from the design assignment has a target," so this should be reconciled (either add rows or note them as removed).
3. The catalog renders both `Twitter` _(custom)_ and `TwitterX`, but no migration row points at `TwitterX` and there's no guidance on which to pick.
4. PR description promises `createFilledIcon` / `stroke` decoupling docs in the "Filled brand icons" section, but the rendered page only shows the `fill` prop. Either add the engineering content or remove the claim.

**Suggestions:**

- Demo modes are inconsistent (`mode="preview"` for sizes, default Preview+Code for the other two).
- `ChevronDown`/`ChevronUp`/`CircleX` appear in two catalog sections; acceptable trade-off but worth a footnote.
- `cursor-pointer` in `IconList.tsx` was added on the absolutely-positioned overlay, not the parent `<button>`.
- The "Before" code example in **Migrating from the legacy icon set** repeats the same lowercase `search` typo as the changeset.

Recommendation: **Comment** — fix the migration-table issues before this lands in release notes; everything else is non-blocking.

---

.changeset/iconography-docs.md
```diff
@@ -0,0 +1,153 @@
...
+## Migration table
+
+Old names from the legacy `@marigold/icons` set and their Lucide replacement. Names marked _custom_ are retained brand icons that still ship from `@marigold/icons` directly.
+
+### UI
+
+| Old             | New                |
+| --------------- | ------------------ |
+| Add             | `Plus`             |
+| arrow-down      | `ArrowDown`        |
```
aromko: ⚠️ **Casing mismatch with actual exports.** The `Old` column here uses kebab-case / lowercase, but the legacy `@marigold/icons` package exports these in PascalCase (verified against `packages/icons/src/index.ts`). Consumers searching their codebase by old name won't find a match.

The rows that need to flip back to PascalCase:

- UI: `arrow-down` → `ArrowDown`, `arrow-left` → `ArrowLeft`, `arrow-right` → `ArrowRight`, `arrow-up` → `ArrowUp`, `check` → `Check`, `chevron-down/-left/-right/-up` → `ChevronDown/Left/Right/Up`, `search` → `Search`
- Info: `accessibility` → `Accessible`
- User: `id` → `Id`
- Social: `share` → `Share`, `twitter` → `Twitter`

Worth a single sweep across the whole table since the migration table is what ends up in the Marigold 18 release notes.

.changeset/iconography-docs.md
```diff
@@ -0,0 +1,153 @@
...
+| User                   | `UserRound`    |
+
+### Social
+
+| Old       | New                    |
+| --------- | ---------------------- |
+| Google    | `Google` _(custom)_    |
+| Instagram | `Instagram` _(custom)_ |
+| share     | `Share2`               |
+| twitter   | `Twitter` _(custom)_   |
```
aromko: ⚠️ **Missing migration rows.** The PR description says "every legacy icon from the design assignment has a target," but seven icons currently exported from `@marigold/icons` have no row in the table:

- UI: `Close`
- Info: `Phone`, `Print`
- Action: `PauseAlt`, `PlayAlt`, `Upload`
- Social: `Whatsapp`

Either map them to a Lucide replacement, mark them as kept brand/legacy icons, or call them out as removed — otherwise consumers using these have no migration path.

Also: the catalog now renders both `Twitter` _(custom)_ and `TwitterX`, but only `twitter → Twitter` is listed here. Add a row (or a one-line note) clarifying which mark to use post-rebrand.

docs/content/foundations/iconography/index.mdx
```diff
@@ -0,0 +1,336 @@
...
+// Recommended — className with a theme token
+<TriangleAlert className="text-destructive-accent" />
+```
+
+See [Token Overview](/foundations/token-overview) for token names.
+
+### Filled brand icons
+
+Some brand icons are solid shapes instead of outlines. For these, setting
+`fill` (or `color`) colors the entire icon at once.
```
aromko, line 286: ⚠️ **Engineering content from the PR description is missing here.** The PR description says:

> Filled brand icons: `createFilledIcon` couples `fill` and `color`; pass `stroke` to decouple

…but this section only shows `<DesignTicket fill={...} />`. There's nothing about `createFilledIcon` or how to decouple `stroke` from `fill`. Either drop that line from the description or add the explanation here so behavior matches what the changeset advertises.

docs/content/foundations/iconography/index.mdx
```diff
@@ -0,0 +1,336 @@
...
+
+The package name stays the same (`@marigold/icons`), so the import path does
+not change. What changes is the icon names and a small API tweak.
+
+**Names.** Most legacy names map to a Lucide equivalent. Lucide uses
+PascalCase consistently, so a few common renames look like this:
+
+```tsx
+// Before
+import { Add, Delete, Exclamation, search } from '@marigold/icons';
```
aromko: 💡 Same casing slip as the changeset — the legacy export was `Search` (PascalCase), not `search`.

```suggestion
import { Add, Delete, Exclamation, Search } from '@marigold/icons';
```

(The "After" line on the next line then needs no change since `Search` already appears there with `TriangleAlert`.)

docs/ui/IconList.tsx
```diff
@@ -54,7 +54,7 @@ const IconListItem = ({ icon }: IconListItemProps) => {
             <div
               className={cn(
                 isCopied ? 'opacity-100' : 'opacity-0',
-                'bg-fd-card absolute flex size-full items-center justify-center font-medium tracking-wider whitespace-nowrap select-none'
+                'bg-fd-card absolute flex size-full cursor-pointer items-center justify-center font-medium tracking-wider whitespace-nowrap select-none'
```
aromko: 💡 `cursor-pointer` is being applied to the absolutely-positioned overlay div rather than the parent `<button>` (line 51). The button already gets a pointer cursor by default, and this overlay sits on top of it, so the rule is effectively redundant.

If the goal was to make the whole tile feel clickable, applying `cursor-pointer` on the `<button>` (or the `Card`) would be the more direct fix. Not blocking.

.changeset/iconography-docs.md
sarahgm: Fixed in 90a850d6a — switched the Old column to the legacy `@marigold/icons` PascalCase export names (ArrowDown, Check, Chevron*, Search, Accessible, Id, Share, Twitter) so a grep across consumer imports finds them.

docs/content/foundations/iconography/index.mdx
sarahgm: Fixed in 90a850d6a.

docs/ui/IconList.tsx
sarahgm: Moved to the `<button>` in 429a6feb2. Heads-up for future review: Tailwind v4 Preflight resets `button { cursor: default }`, so the explicit `cursor-pointer` class is needed in this project — the browser default does not apply.

.changeset/iconography-docs.md
sarahgm: Cross-checked against the design assignment in `~/Downloads/Lucide Icons   2/{UI,Info,Action,Social,Ticketing,User}/` — `Close`, `Phone`, `Print`, `PauseAlt`, `PlayAlt`, `Upload`, and `Whatsapp` are not in the redesigned set. They were intentionally dropped during the redesign rather than missing from the migration table. The table fully covers the design assignment (28+33+4+14+23+10 = 112 rows == 112 SVGs).

For the TwitterX clarification: added a footnote under Social in 90a850d6a — `Twitter` _(custom)_ stays for the bird mark, `TwitterX` is available for the X mark.

docs/content/foundations/iconography/index.mdx
sarahgm: Good catch on this one — the PR description was wrong. I traced through `createFilledIcon` (`packages/icons/src/custom/createFilledIcon.tsx:14-18`): it does spread `...props` after `fill={c} color={c}`, so a `stroke` prop reaches the SVG element. **But** every filled icon hardcodes `stroke: none` on its path nodes (e.g. `DesignTicket.tsx:8`, `Stadium.tsx:9`, `Turnstile.tsx:8`), and path-level attrs override SVG-level ones — so passing `stroke="red"` renders nothing.

Updated the PR description to drop the false `pass stroke to decouple` claim. The existing docs section ("setting `fill` (or `color`) colors the entire icon at once") is accurate, so no doc change.

.changeset/iconography-docs.md
sarahgm: Can check again here: https://drive.google.com/drive/folders/1XbiI5m451TXuMRZgvjvlzEVRifIusG2_ If there is WhatApp for example needed, i would add it then...

aromko (approved):

PR #5380, opened 2026-04-30
feat([DST-901]): accept number + string scale values in styleProps

sebald:
## Summary

Closes [DST-901](https://reservix.atlassian.net/browse/DST-901). Co-lands the dimensional / spacing portion of [DST-1309](https://reservix.atlassian.net/browse/DST-1309) (CSS-var resolution for scale-based style props).

- StyleProps for `width`, `maxWidth`, `height`, `space`, `spaceX`, `spaceY`, `pr`, `pl`, `pt`, `pb` now accept both a numeric scale value (`4`) and its string equivalent (`"4"`). Public types are declarative (`Scale | Fraction | WidthKeyword`, etc.) instead of derived from internal class-name maps, so the type surface is independent of how the value is resolved at runtime.
- Components that previously resolved `width` / `maxWidth` / `height` through map lookups (`Form`, `Calendar`, legacy `Table` column-header / select-all-cell, `Slider`, `Scrollable`, `Switch`, `Grid`) now resolve them through CSS custom properties via `createWidthVar` / new `createHeightVar`, targeting `--width` / `--max-width` / `--height`. These vars — plus the existing `--container-width` and `--field-width` used by `FieldBase` — are registered as non-inheriting (`@property … inherits: false`) in the RUI theme so they cannot leak into descendants.
- `createWidthVar` gained the previously-missing keywords (`svh`, `lvh`, `dvh`, `px`, `container`); both helpers share a common factory and a base keyword set.
- **Breaking** (`@marigold/system`): the runtime class-name maps `width`, `maxWidth`, `height`, `gapSpace`, `paddingSpace`, `paddingSpaceX`, `paddingSpaceY`, `paddingRight`, `paddingLeft`, `paddingTop`, `paddingBottom` are no longer exported. Consume the prop types (`WidthProp`, `HeightProp`, …) and the CSS-var helpers (`createWidthVar`, `createHeightVar`, `createSpacingVar`) instead. Prop type names and shapes are unchanged.
- Docs: short note in `foundations/form-fields` that scale values accept both forms; `docs/ui/Token.tsx`'s spacing-table renderer (the only consumer of the deleted maps outside the system package) now uses a local scale array.

## Test plan

- [x] `pnpm lint` 0 errors
- [x] `pnpm test:unit` — 1104/1104
- [x] `pnpm test:sb` — passing on touched components
- [x] Build of `@marigold/system` and `@marigold/theme-rui` succeeds
- [ ] CI green
- [ ] Visual regression (Chromatic) — verify Form / Calendar / Slider / Scrollable / Switch / Grid / legacy Table render identically; `--container-width` / `--field-width` non-inheritance shouldn't change visuals but is worth a sanity check
- [ ] Smoke test docs site (`pnpm start`) — Token spacing table renders, all `pl-*` cells visible

## Notes for the reviewer

- The 9 lookup sites moved from `twMap[value]` (static class) to `cn('w-(--width)', …)` + `style={createWidthVar('width', \`${value}\`)}` (dynamic CSS var). All of them have a default value at destructure time (`width = 'full'` / `'auto'` / `'fit'`), so `\`${value}\`` never coerces `undefined`.
- The non-inheriting `@property` rules for `--container-width` and `--field-width` were a finding from the simplify pass on this PR, not a behaviour fix — `FieldBase` already worked correctly because no descendant currently re-uses those names. Registering them keeps the layout-var protection complete.
- DST-1309's other categories (alignment, textAlign, placeItems, aspect, cursorStyle, font*, etc.) are deliberately untouched here. Those need a different per-prop strategy (inline `style` / move-into-component-file) and are best done in their own PR.

[DST-901]: https://reservix.atlassian.net/browse/DST-901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DST-1309]: https://reservix.atlassian.net/browse/DST-1309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/Slider/Slider.tsx
+++ b/packages/components/src/Slider/Slider.tsx
@@ -7,12 +7,7 @@ import {
   SliderThumb,
   SliderTrack,
 } from 'react-aria-components';
-import {
-  WidthProp,
-  cn,
-  width as twWidth,
-  useClassNames,
-} from '@marigold/system';
+import { WidthProp, cn, createWidthVar, useClassNames } from '@marigold/system';
 import { FieldBase, FieldBaseProps } from '../FieldBase/FieldBase';
 import { Label } from '../Label/Label';
 
@@ -83,10 +78,10 @@ const _Slider = forwardRef(
       <FieldBase
         as={Slider}
         className={cn(
-          'grid grid-cols-[auto_1fr] gap-y-1',
-          classNames.container,
-          twWidth[width]
+          'grid w-(--width) grid-cols-[auto_1fr] gap-y-1',
+          classNames.container
         )}
+        style={createWidthVar('width', `${width}`)}
         ref={ref}
         {...props}
       >
--- a/packages/components/src/Slider/Slider.tsx
+++ b/packages/components/src/Slider/Slider.tsx
@@ -13,7 +13,10 @@ import { Label } from '../Label/Label';
 
 export interface SliderProps<T>
   extends
-    Omit<RAC.SliderProps<T>, 'children' | 'isDisabled' | 'orientation'>,
+    Omit<
+      RAC.SliderProps<T>,
+      'children' | 'isDisabled' | 'orientation' | 'style'
+    >,
     Pick<FieldBaseProps<'label'>, 'description'> {
   variant?: string;
   size?: string;
--- a/packages/components/src/Slider/Slider.tsx
+++ b/packages/components/src/Slider/Slider.tsx
@@ -84,7 +84,7 @@ const _Slider = forwardRef(
           'grid w-(--width) grid-cols-[auto_1fr] gap-y-1',
           classNames.container
         )}
-        style={createWidthVar('width', `${width}`)}
+        style={createWidthVar('width', width)}
         ref={ref}
         {...props}
       >
--- a/packages/components/src/Switch/Switch.tsx
+++ b/packages/components/src/Switch/Switch.tsx
@@ -1,12 +1,7 @@
 import { ReactNode, forwardRef } from 'react';
 import type RAC from 'react-aria-components';
 import { Switch } from 'react-aria-components';
-import {
-  WidthProp,
-  cn,
-  width as twWidth,
-  useClassNames,
-} from '@marigold/system';
+import { WidthProp, cn, createWidthVar, useClassNames } from '@marigold/system';
 import { Label } from '../Label/Label';
 
 type RemovedProps =
@@ -78,11 +73,10 @@ const _Switch = forwardRef<HTMLLabelElement, SwitchProps>(
         {...props}
         ref={ref}
         className={cn(
-          twWidth[width],
-          'group/switch',
-          'flex items-center gap-[1ch]',
+          'group/switch flex w-(--width) items-center gap-[1ch]',
           classNames.container
         )}
+        style={createWidthVar('width', `${width}`)}
       >
         {label && <Label elementType="span">{label}</Label>}
         <div className="relative">
--- a/packages/components/src/Switch/Switch.tsx
+++ b/packages/components/src/Switch/Switch.tsx
@@ -76,7 +76,7 @@ const _Switch = forwardRef<HTMLLabelElement, SwitchProps>(
           'group/switch flex w-(--width) items-center gap-[1ch]',
           classNames.container
         )}
-        style={createWidthVar('width', `${width}`)}
+        style={createWidthVar('width', width)}
       >
         {label && <Label elementType="span">{label}</Label>}
         <div className="relative">
--- a/packages/components/src/Switch/Switch.tsx
+++ b/packages/components/src/Switch/Switch.tsx
@@ -7,10 +7,10 @@ import { Label } from '../Label/Label';
 type RemovedProps =
   | 'children'
   | 'className'
+  | 'style'
   | 'isDisabled'
   | 'isReadOnly'
   | 'isSelected'
-  | 'children'
   | 'slot';
 
 export interface SwitchProps extends Omit<RAC.SwitchProps, RemovedProps> {
--- a/.changeset/dst-901-styleprops-numeric-strings.md
+++ b/.changeset/dst-901-styleprops-numeric-strings.md
@@ -4,10 +4,10 @@
 '@marigold/theme-rui': patch
 ---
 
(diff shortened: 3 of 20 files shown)


aromko (commented):
## Automated Code Review

### Overview
| Field | Value |
|-------|-------|
| Author | @sebald |
| Branch | `dst-901-styleprops-numeric-strings` → `main` |
| Files Changed | 20 |
| Additions | +335 |
| Deletions | -643 |
| CI | All checks passing |

### Linked Tickets
- **DST-901** *(In Review)* — "Extend styleProps to accept both numbers and their string equivalents". Delivers the `Scale | Fraction | WidthKeyword` public surface on the dimensional/spacing props.
- **DST-1309** *(In Review)* — "Move style-prop class resolution from @marigold/system to @marigold/components". Co-lands the dimensional/spacing portion via `createWidthVar` / `createHeightVar` / `createSpacingVar`, exactly per the 2026-04-30 comment.

### Issues Found

#### Warnings (Should Fix)

1. `packages/components/src/Switch/Switch.tsx` — `RemovedProps` does **not** include `'style'`. Pre-PR, `Switch` set no inline style, so a consumer-passed `style={{ ... }}` was harmlessly forwarded through `{...props}`. After this PR, `style={createWidthVar('width', ` + '`${width}`' + `)}` is set **after** `{...props}`, so any consumer-supplied `style` is silently dropped. CLAUDE.md is explicit: *"Don't expose `className` or `style` props - Remove from RAC props interface"*. Easiest fix: add `'style'` to `RemovedProps`. Same prop hygiene missing on `packages/components/src/Slider/Slider.tsx`; there `style={createWidthVar(...)}` comes **before** `{...props}`, so the opposite bug — consumer-passed `style` clobbers `--width`. The two components fail differently for the same reason.

2. `packages/system/src/utils/css-variables.utils.ts` — `keywords[value] || (isScale(value) && ...) || (isFraction(value) && ...)` works correctly because no keyword maps to an empty string. Mildly fragile if a future keyword is `''`. `Object.hasOwn(keywords, value)` would harden it.

3. `packages/system/src/style-props.tsx` — `WidthValue` and `MaxWidthValue` are structurally identical (`Scale | Fraction | WidthKeyword`); one alias would compress the public surface.

4. `docs/ui/Token.tsx` — `SPACING_SCALE` duplicates the `ScaleValue` union. Could be auto-generated, or at minimum unit-asserted that array length matches the type.

### Recommendation

**Comment** — Solid PR. Core design (declarative public types + var-based resolution + non-inheriting `@property` rules) is sound. The only real bug is the `Switch`/`Slider` `style`-forwarding regression — one-line fix per file.

---

packages/components/src/Switch/Switch.tsx
```diff
@@ -78,11 +73,10 @@ const _Switch = forwardRef<HTMLLabelElement, SwitchProps>(
         {...props}
         ref={ref}
         className={cn(
-          twWidth[width],
-          'group/switch',
-          'flex items-center gap-[1ch]',
+          'group/switch flex w-(--width) items-center gap-[1ch]',
           classNames.container
         )}
+        style={createWidthVar('width', `${width}`)}
```
aromko: ⚠️ `style` is not in `RemovedProps`. Because this `style={...}` is set **after** `{...props}`, any consumer-supplied `style` is silently dropped. Per CLAUDE.md (*"Don't expose `className` or `style` props - Remove from RAC props interface"*), add `'style'` to the `RemovedProps` union at the top of the file.

packages/components/src/Slider/Slider.tsx
```diff
@@ -83,10 +78,10 @@ const _Slider = forwardRef(
       <FieldBase
         as={Slider}
         className={cn(
-          'grid grid-cols-[auto_1fr] gap-y-1',
-          classNames.container,
-          twWidth[width]
+          'grid w-(--width) grid-cols-[auto_1fr] gap-y-1',
+          classNames.container
         )}
+        style={createWidthVar('width', `${width}`)}
```
aromko: ⚠️ Mirror of the issue in `Switch.tsx`, but worse: here `style={createWidthVar(...)}` is set **before** `{...props}`, so a consumer-supplied `style` clobbers `--width` and breaks the width prop entirely.

Fix: omit `'style'` from `SliderProps` (`Omit<RAC.SliderProps<T>, 'children' | 'isDisabled' | 'orientation' | 'style'>`).

packages/components/src/Switch/Switch.tsx
sebald: Good catch, thanks. Fixed in bff49de — added `'style'` to `RemovedProps`. The order (`style={...}` after `{...props}`) meant consumer-supplied `style` was silently dropped rather than breaking `--width`, but per CLAUDE.md neither outcome is acceptable. Also dropped the duplicate `'children'` entry in the same union while I was there.

packages/components/src/Slider/Slider.tsx
sebald: Right — this one was the actually-broken case (consumer `style` clobbering `--width`). Fixed in bff49de by adding `'style'` to the `Omit` exactly as suggested.

sebald:
Thanks for the review @aromko — addressed all four items.

| # | Item | Status |
|---|---|---|
| 1 | `Switch` / `Slider` consumer-`style` regression | Fixed in bff49de — both threads resolved |
| 2 | `keywords[value] \|\|` is fragile if a future keyword is `''` | Fixed in f3d7f69 — switched to `Object.hasOwn(keywords, value)`. As a bonus, this also stops `__proto__` / `constructor` / `toString` from silently resolving to the prototype object; added regression tests for that. |
| 3 | `WidthValue` and `MaxWidthValue` are structurally identical | Fixed in f3d7f69 — `MaxWidthValue` is now an alias of `WidthValue`. |
| 4 | `SPACING_SCALE` duplicates `ScaleValue` | Fixed in f3d7f69 — `SPACING_SCALE` is now `as const satisfies readonly \`${ScaleValue}\`[]`, so typos / invalid additions are a compile error. Missing entries still surface as a missing row in the rendered table; full type-level exhaustiveness felt over-engineered for docs UI. Happy to add it if you'd prefer.

aromko (approved):

PR #5382, opened 2026-04-30
chore: backport SelectList rework and main updates onto beta-release

sebald:
## Summary

Backports the `<SelectList>` API rework and other recent `main` work onto `beta-release`, with reconciliation against the renamed v18 design tokens.

### SelectList
- New API: `<SelectList.Option>` (replaces `Item`); single `SelectionIndicator` (radio for single, checkbox for multi); per-option action via `<IconButton>`/`<ActionMenu>` through `ButtonContext` (replaces `SelectList.Action`).
- Dedicated `theme-rui` `SelectList` style entry with `default` and `bordered` variants.
- Tokens reconciled to v18: `border-border`, `bg-surface`, `border-selected-bold`, `bg-selected-bold`, `text-disabled`, `text-secondary`, `--color-info`, etc.
- `SelectionIndicator` repainted with `border-border` / `shadow-elevation-border` / `bg-surface`; disabled-selected uses `border-disabled-surface!` / `bg-disabled-surface`.

### FieldBase
- Forwards `isInvalid`, `isRequired`, `isDisabled` to the underlying component when `as` is a RAC component, while still skipping them on plain DOM elements (avoids unknown DOM attribute warnings). Combines main's RAC validation forwarding with beta's React 19 ref-as-prop pattern.

### Docs
- SelectList docs moved from `components/collection/` to `components/form/` and rewritten around the new API (anatomy, decision table, multi-selection, per-row actions, horizontal, empty state).
- `selectlist-decision-help.demo.tsx` restructured: label sits in row 1, rich content (price/description/features) in row 2, so the selection indicator aligns with the label at the top of each option.
- Token updates in remaining demos: `text-muted-foreground` → `text-secondary`, `--color-*-muted` → `--color-*`, `text-success-muted-foreground` → `text-success-foreground`.

### Storybook
- Removed `parameters: { surface: false }` from `SelectList.stories.tsx` so the story renders inside the default Panel surface (matches docs presentation).

### Changesets
- `select-list-standard-layout.md` (existing) — describes the SelectList API rework.
- `fieldbase-rac-validation-forwarding.md` (new) — describes the FieldBase RAC validation forwarding behavior.

## Test plan

- [x] `pnpm test:unit` — 1138/1138 pass
- [x] `pnpm test:sb` — 139/139 pass
- [x] `pnpm lint` — 0 errors
- [x] `pnpm build` — components and `theme-rui` build cleanly
- [x] Manual visual check: docs `/components/form/selectlist` (Basic, Bordered, Multi-selection, Configuration, Horizontal, Empty, Decision-help) and Storybook `Components/SelectList` look correct
- [x] Reviewer to verify indicator alignment and bordered variant on hover/selected/focus across themes


themes/theme-rui/src/components/SelectList.styles.ts
```diff
@@ -0,0 +1,108 @@
...
+      // Inset focus ring so it can't be clipped by the list's `overflow`.
+      'focus-visible:inset-shadow-[0_0_0_2px_var(--color-ring)] transition-[border,color,background]',
+      'disabled:cursor-not-allowed disabled:text-disabled',
+      'group-orientation-horizontal/list:min-w-40',
+    ],
+    variants: {
+      variant: {
+        default: [
+          'min-h-14',
+          'selected:bg-selected hover:bg-hover',
```
sarahgm: Why no `hover:ui-state-hover` instead of bg-hover?

packages/components/src/SelectList/SelectList.tsx
```diff
@@ -1,68 +1,364 @@
...
+   * Content to render when the list has no items.
+   */
+  emptyState?: ReactNode;
+  /**
+   * Selection mode.
+   * - `single`: at most one option may be selected.
+   * - `multiple`: any number of options may be selected.
+   * @default 'single'
+   */
+  selectionMode?: M;
```
sarahgm: Why M? Mode will be clearer?

packages/components/src/SelectList/SelectList.stories.tsx
```diff
@@ -1,96 +1,734 @@
...
+      {...args}
+      label="Accepted payment methods"
+      description="Customers see these on checkout. Tap the info icon for help."
+      items={paymentMethods}
+    >
+      {(item: { id: string; name: string; description: string }) => (
+        <SelectList.Option textValue={item.name}>
+          <Text slot="label">{item.name}</Text>
+          <Text slot="description">{item.description}</Text>
+          <IconButton
```
sarahgm: IconButton should not be used, thought its for building Components only. You should use here `<Button variant=icon>`

packages/components/src/SelectList/SelectList.stories.tsx
```diff
@@ -1,96 +1,734 @@
...
+  render: args => (
+    <SelectList
+      {...args}
+      label="Select a payment method"
+      description="Each option is its own committed choice."
+      defaultSelectedKeys={['visa']}
+    >
+      {borderedMethods.map(({ id, label, description, Logo }) => (
+        <SelectList.Option key={id} id={id} textValue={label}>
+          <div className="col-start-2 row-span-2 flex items-start gap-3">
```
sarahgm: Can we reuse here a Marigold Layout component? Would be cool if the users can see how to reuse our stuff together

packages/components/src/FieldBase/FieldBase.tsx
```diff
@@ -44,9 +45,19 @@ const _FieldBase = <T extends ElementType>({
...
+  isDisabled,
   ref,
   ...rest
 }: FieldBaseProps<T> & DistributiveOmit<ComponentPropsWithRef<T>, 'as'>) => {
+  // Forward RAC validation props only when rendering through a RAC component;
+  // they would emit unknown DOM attribute warnings on a plain element.
+  const racValidationProps =
+    typeof Component === 'string'
+      ? null
+      : { isInvalid, isRequired, isDisabled };
```
sarahgm, line 60: ⚠ `typeof Component === 'string' ? null : { ... }` forwards validation props to *any* non-string `as` value, not just RAC. The changeset describes this change as RAC-specific. Either tighten the heuristic or add a JSDoc on this branch clarifying that any non-DOM `as` will receive these props — otherwise consumers passing custom non-RAC components via `as` may see unexpected forwarding.

packages/components/src/SelectList/SelectListHiddenSelect.tsx
```diff
@@ -0,0 +1,86 @@
...
+        required={validationBehavior === 'native' && required}
+        name={name}
+        form={form}
+        value={value}
+        onChange={onChange}
+      >
+        <option value="" />
+        {selectedKeys.map(key => (
+          <option key={key} value={key}>
+            {key}
```
sarahgm: ⚠ `{key}` as the option's text content leaks the raw key into native validation tooltips (`select.validationMessage`), `select.selectedOptions[i].text`, automation tooling, and DOM introspection. The `<select>` is `aria-hidden`, but option text isn't gated by that. If keys are opaque ids (UUIDs, `pm_4f2a-9b`), they shouldn't be the visible label.

```suggestion
        <option key={key} value={key} />
```

packages/components/src/SelectList/SelectListHiddenSelect.tsx
```diff
@@ -0,0 +1,86 @@
...
+        onSelectionChange(
+          target.value ? new Set([target.value as Key]) : new Set()
+        );
+      }
+    },
+    [onSelectionChange]
+  );
+
+  const selectedKeys =
+    selection === 'all' ? [] : Array.from(selection).map(String);
```
sarahgm, line 65: 💡 `selection === 'all' ? [] : ...` produces an empty `FormData` submission when SelectList state is `'all'`, but visually the user thinks everything is selected. The hidden select can't enumerate the full key set, so this can't really be fixed here — either reject `'all'` at SelectList's API surface, or add a comment documenting the limitation.

packages/components/src/SelectList/SelectListHiddenSelect.tsx
```diff
@@ -0,0 +1,86 @@
...
+}: SelectListHiddenSelectProps) => {
+  const selectRef = useRef<HTMLSelectElement>(null);
+
+  useFormValidation(
+    { validationBehavior, focus: () => focusRef.current?.focus() },
+    validationState,
+    selectRef
+  );
+
+  const onChange = useCallback(
```
sarahgm, line 45: 💡 This `onChange` is essentially unreachable in normal flow — the select is `tabIndex={-1}`, `aria-hidden`, inside `VisuallyHidden`. It's there for form reset and browser autofill. A one-line comment would save the next reader a confused re-read.

packages/components/src/SelectList/SelectListOption.tsx
```diff
@@ -0,0 +1,140 @@
...
+const OptionChildren = ({
+  children,
+  labelClassName,
+  descriptionClassName,
+  actionClassName,
+}: OptionChildrenProps) => {
+  const parentText = use(TextContext) as SlottedContextValue | undefined;
+  const parentButton = use(ButtonContext) as SlottedContextValue | undefined;
+  const parentTextSlots = parentText?.slots;
+  const parentButtonValue = parentButton ?? {};
```
sarahgm: 💡 `parentButton ?? {}` allocates a fresh object each render when no parent context is set, defeating the `useMemo` of `buttonContextValue` on line 72. Stabilize with a module-level constant (e.g. `const EMPTY_BUTTON_CTX = {};` at module scope, then `parentButton ?? EMPTY_BUTTON_CTX`).

packages/components/src/SelectList/SelectListOption.tsx
```diff
@@ -0,0 +1,140 @@
...
+}: SelectListOptionProps) => {
+  const { classNames, disabled: listDisabled } = useSelectListContext();
+  const resolvedTextValue =
+    textValue ?? (typeof children === 'string' ? children : undefined);
+
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    resolvedTextValue === undefined
+  ) {
+    console.warn(
```
sarahgm, line 108: 💡 Consider including the option's `id` in this warning when available — easier to debug in long lists than a generic message.

packages/components/src/SelectList/SelectList.tsx
```diff
@@ -1,68 +1,364 @@
...
+  (props: SelectListProps): ReactNode;
+  <M extends SelectionMode = 'single'>(props: SelectListProps<M>): ReactNode;
+  Option: typeof SelectListOption;
 }
 
-const SelectListBase = ({
-  onChange,
+// Module-level sentinel so unset padding props don't allocate a fresh object
+// each render — RAC's GridList compares `style` by identity in some places.
+const EMPTY_STYLE: CSSProperties = {};
```
sarahgm, line 183: 💡 The "RAC's GridList compares `style` by identity in some places" claim is non-obvious and uncited. Either link the RAC source or soften the wording (e.g. "avoid identity churn for downstream consumers"). Future readers will hunt for that comparison and not find it.

themes/theme-rui/src/components/SelectList.styles.ts
```diff
@@ -0,0 +1,108 @@
...
+      // `items-start` keeps the label/description anchored to the top of their
+      // grid cells, while `content-center` centers the row tracks vertically
+      // when `min-h-*` makes the option taller than its content. The indicator
+      // then sits on the label's first line even when the option content uses
+      // a row-spanning wrapper (e.g. a leading illustration).
+      'relative grid items-start content-center gap-x-3',
+      'grid-cols-[auto_1fr_auto]',
+      // Row 1 minimum matches the label's line-height (text-sm == 1.25rem) so
+      // the indicator stays aligned with the label's first line.
+      'grid-rows-[minmax(1.25rem,auto)_auto]',
```
sarahgm, line 44: 💡 `1.25rem` is hard-coded to match `text-sm`'s line-height. If Tailwind's `text-sm` ever changes, indicator alignment will silently drift. Worth a comment: "update if `text-sm` line-height changes."

docs/content/components/form/selectlist/selectlist-multiple-selection.demo.tsx
```diff
@@ -0,0 +1,306 @@
...
+      selectedKeys={selected}
+      onChange={keys => setSelected(keys as string[])}
+    >
+      {methods.map(method => (
+        <SelectList.Option
+          key={method.id}
+          id={method.id}
+          textValue={method.name}
+        >
+          <div className="col-start-2 row-span-2 flex min-w-0 items-start gap-3">
```
sarahgm: Isn't it somehow reachable to get the same layout with our components? So the User knows which layout component he/she should use? (Same for other demos)

themes/theme-rui/src/components/SelectList.styles.ts
sebald: Good catch — switched to `hover:ui-state-hover` so it matches `ListBox`, `Menu`, `Panel` and the rest of the system.

packages/components/src/SelectList/SelectList.tsx
sebald: Renamed the generic to `Mode` throughout.

packages/components/src/SelectList/SelectList.stories.tsx
sebald: Switched to `<Button variant='ghost' size='icon'>`. Note that `Button` exposes `size='icon'` (not `variant='icon'`) — this is the canonical icon-only pattern we already use elsewhere in stories/docs.

packages/components/src/SelectList/SelectList.stories.tsx
sebald: Replaced the raw flex/div with `<Inline>` + `<Stack>` to demonstrate composition. The outer wrapping div has to stay because it carries the SelectList option grid placement (`col-start-2 row-span-2`) — `Stack`/`Inline` don't accept a className escape hatch for that.

packages/components/src/FieldBase/FieldBase.tsx
sebald: Updated the comment to make the heuristic explicit: any non-string `as` receives `isInvalid`/`isRequired`/`isDisabled`. In practice all our consumers pass RAC components or wrappers that follow RAC's prop names; if a consumer passes a custom non-RAC component via `as`, they're expected to drop or ignore those props.

packages/components/src/SelectList/SelectListHiddenSelect.tsx
sebald: Done — options no longer carry text content (`<option key={key} value={key} />`). Added a comment so the next reader knows why.

packages/components/src/SelectList/SelectListHiddenSelect.tsx
sebald: Added a comment documenting the limitation. Rejecting `'all'` at the API surface would break consumers who select-all in the UI but don't care about form submission, so I left it as a soft trap that the comment now flags.

packages/components/src/SelectList/SelectListHiddenSelect.tsx
sebald: Added a comment explaining it's there for form reset / browser autofill (the `<select>` is `tabIndex={-1}`, `aria-hidden`, inside `VisuallyHidden`).

packages/components/src/SelectList/SelectListOption.tsx
sebald: Stabilized with a module-level `EMPTY_BUTTON_CTX` so the `useMemo` actually keeps a stable identity when no parent context is set.

packages/components/src/SelectList/SelectListOption.tsx
sebald: Done — the warning now reads `[SelectList.Option id="…"]` when the option has an id.

packages/components/src/SelectList/SelectList.tsx
sebald: Softened — the comment no longer claims a specific RAC behavior, just describes the intent (avoid identity churn for downstream consumers).

themes/theme-rui/src/components/SelectList.styles.ts
sebald: Added the suggested note next to the magic number.

docs/content/components/form/selectlist/selectlist-multiple-selection.demo.tsx
sebald: Restructured this demo to use `<Inline>` + `<Stack>` for label/badge/description and the fee column. The outer wrapping divs remain for SelectList grid placement (`col-start-N row-span-2`), since `Stack`/`Inline` don't expose a className escape hatch — but the actual layout is now expressed through Marigold layout components.

sarahgm (approved):

PR #5385, opened 2026-05-04
feat([DST-1134]): add `<RangeCalendar>` component

aromko:
## Summary

> [!CAUTION]
> If we want to give this the Portal Team fast we should cut a release onto main?
 
Adds a new `<RangeCalendar>` for picking a contiguous (or non-contiguous) date range, built on react-aria's `<RangeCalendar>` with Marigold conventions on top.

- **Multi-month layout** — `visibleDuration={{ months: 1 | 2 | 3 }}`. Months stack vertically below the `sm` (640px) breakpoint and flow side-by-side from `sm` upward (also applied to single-month `<Calendar>` for parity).
- **Range visuals** — cells render as a continuous bar between the `selection-start` and `selection-end` (rounded caps, `bg-brand/15` mid-range, hard `bg-brand` on the ends).
- **Form-component parity** — description and error rendering go through `<FieldBase>` so the help/error UI matches `<DateField>`/`<DatePicker>` (TriangleAlert icon + HelpText container). Because react-aria's `<RangeCalendar>` is a picker rather than a form input, the component wraps `FieldBase` in a `FieldErrorContext.Provider` so `HelpText` sees `isInvalid`.
- **Picker dropdown overlays** — month and year selection via listbox dropdowns. Native `pointerup` is stopped at the dropdown overlay so react-aria's window-level commit handler doesn't drop the first click on a `role="option"` ListBoxItem.
- **Shared with Calendar** — reuses `Calendar`'s context, header, grid, and listbox primitives. `RangeCalendar.styles.ts` spreads `Calendar.styles.ts` and overrides only `calendarCell` and `calendarGrid` (the two visual differences). `errorMessage` lives where it's used.
- **Stories** — `Basic`, `Controlled`, `Disabled`, `ReadOnly`, `WithError`, `Unavailable`, `NonContiguous`, `TwoMonths`, `ThreeMonths`, `TwoMonthsMobile`, `ThreeMonthsMobile`, `MultiMonthNavigation`, `WithDescription`. `MultiMonthNavigation` uses `step()` blocks for clearer test reporting.

## Test plan

- [ ] Visit `RangeCalendar/Basic` and select a range across two clicks; the strip renders with rounded caps and a `bg-brand/15` middle.
- [ ] `RangeCalendar/NonContiguous` — click a weekday, then a weekday past a weekend; the range spans the weekend with the unavailable days as gaps.
- [ ] `RangeCalendar/TwoMonthsMobile` and `ThreeMonthsMobile` — months stack vertically; reset the viewport and confirm side-by-side at ≥640px.
- [ ] `RangeCalendar/MultiMonthNavigation` — prev/next paging moves both visible months in lockstep; story tests cover this.
- [ ] `RangeCalendar/WithError` — TriangleAlert icon and message render below the calendar; aria-describedby wiring intact.
- [ ] Open the month or year dropdown and click an option; range selection isn't committed mid-pick.
- [ ] `pnpm typecheck:only` clean.
- [ ] `pnpm test:sb` clean for `Components/Calendar` and `Components/RangeCalendar`.

### DIFF
--- a/packages/components/src/RangeCalendar/RangeCalendar.stories.tsx
+++ b/packages/components/src/RangeCalendar/RangeCalendar.stories.tsx
@@ -133,19 +133,47 @@ export const Unavailable = meta.story({
 
 export const NonContiguous = meta.story({
   args: {
-    defaultValue: {
-      start: new CalendarDate(2025, 8, 4),
-      end: new CalendarDate(2025, 8, 8),
-    },
     allowsNonContiguousRanges: true,
   },
   render: args => {
     const { locale } = useLocale();
+    const [value, setValue] = useState<RangeValue<DateValue> | null>({
+      start: new CalendarDate(2025, 8, 4),
+      end: new CalendarDate(2025, 8, 8),
+    });
+
+    const isUnavailable = (date: DateValue) => isWeekend(date, locale);
+
+    const selectedDates: DateValue[] = [];
+    if (value) {
+      let cursor = value.start;
+      while (cursor.compare(value.end) <= 0) {
+        if (!isUnavailable(cursor)) selectedDates.push(cursor);
+        cursor = cursor.add({ days: 1 });
+      }
+    }
+
     return (
-      <RangeCalendar
-        {...args}
-        dateUnavailable={date => isWeekend(date, locale)}
-      />
+      <>
+        <RangeCalendar
+          {...args}
+          value={value}
+          onChange={setValue}
+          dateUnavailable={isUnavailable}
+        />
+        <pre style={{ marginTop: '1rem' }} data-testid="selectedRange">
+          <strong>RangeCalendar Value:</strong>
+          {value
+            ? ` ${value.start.toString()} → ${value.end.toString()}`
+            : ' (none)'}
+        </pre>
+        <pre style={{ marginTop: '0.5rem' }} data-testid="selectedDates">
+          <strong>Selected dates (excluding unavailable):</strong>
+          {selectedDates.length
+            ? ` ${selectedDates.map(d => d.toString()).join(', ')}`
+            : ' (none)'}
+        </pre>
+      </>
     );
   },
 });
@@ -183,19 +211,27 @@ export const MultiMonthNavigation = meta.story({
     onChange: fn(),
   },
   render: args => <RangeCalendar {...args} />,
-  play: async ({ canvasElement, userEvent }) => {
+  play: async ({ canvasElement, userEvent, step }) => {
     const canvas = within(canvasElement);
 
-    const nextButtons = canvas.getAllByRole('button', { name: /next/i });
-    await userEvent.click(nextButtons[0]);
+    await step('navigates forward two months', async () => {
+      const nextButtons = canvas.getAllByRole('button', { name: /next/i });
+      await userEvent.click(nextButtons[0]);
+    });
 
-    await expect(canvas.getByText('March 2025')).toBeInTheDocument();
-    await expect(canvas.getByText('April 2025')).toBeInTheDocument();
+    await step('shows March and April after navigating forward', async () => {
+      await expect(canvas.getByText('March 2025')).toBeInTheDocument();
+      await expect(canvas.getByText('April 2025')).toBeInTheDocument();
+    });
 
-    const prevButtons = canvas.getAllByRole('button', { name: /previous/i });
-    await userEvent.click(prevButtons[0]);
+    await step('navigates back two months', async () => {
+      const prevButtons = canvas.getAllByRole('button', { name: /previous/i });
+      await userEvent.click(prevButtons[0]);
+    });
 
-    await expect(canvas.getByText('January 2025')).toBeInTheDocument();
-    await expect(canvas.getByText('February 2025')).toBeInTheDocument();
+    await step('shows January and February after navigating back', async () => {
+      await expect(canvas.getByText('January 2025')).toBeInTheDocument();
+      await expect(canvas.getByText('February 2025')).toBeInTheDocument();
+    });
   },
 });
--- a/packages/components/src/RangeCalendar/RangeCalendar.stories.tsx
+++ b/packages/components/src/RangeCalendar/RangeCalendar.stories.tsx
@@ -207,7 +207,7 @@ export const ThreeMonths = meta.story({
   render: args => <RangeCalendar {...args} />,
 });
 
-export const TwoMonthsMobile: any = meta.story({
+export const TwoMonthsMobile = meta.story({
   globals: {
     viewport: { value: 'smallScreen' },
   },
@@ -221,7 +221,7 @@ export const TwoMonthsMobile: any = meta.story({
   render: args => <RangeCalendar {...args} />,
 });
 
-export const ThreeMonthsMobile: any = meta.story({
+export const ThreeMonthsMobile = meta.story({
   globals: {
     viewport: { value: 'smallScreen' },
   },
--- a/packages/components/src/RangeCalendar/RangeCalendar.stories.tsx
+++ b/packages/components/src/RangeCalendar/RangeCalendar.stories.tsx
@@ -200,6 +200,34 @@ export const ThreeMonths = meta.story({
   render: args => <RangeCalendar {...args} />,
 });
 
+export const TwoMonthsMobile: any = meta.story({
+  globals: {
+    viewport: { value: 'smallScreen' },
+  },
+  args: {
+    visibleDuration: { months: 2 },
+    defaultValue: {
+      start: new CalendarDate(2025, 2, 15),
+      end: new CalendarDate(2025, 3, 5),
+    },
+  },
+  render: args => <RangeCalendar {...args} />,
+});
+
+export const ThreeMonthsMobile: any = meta.story({
+  globals: {
+    viewport: { value: 'smallScreen' },
+  },
+  args: {
+    visibleDuration: { months: 3 },
+    defaultValue: {
+      start: new CalendarDate(2025, 5, 20),
+      end: new CalendarDate(2025, 7, 10),
+    },
+  },
+  render: args => <RangeCalendar {...args} />,
+});
+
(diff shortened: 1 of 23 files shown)


packages/components/src/RangeCalendar/RangeCalendar.tsx
```diff
@@ -0,0 +1,275 @@
...
+  // keeps the event from reaching `window` while still letting the option's
+  // own press handler run at the target phase.
+  const dropdownOverlayRef = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    const node = dropdownOverlayRef.current;
+    if (!node) return;
+    const stop = (event: PointerEvent) => event.stopPropagation();
+    node.addEventListener('pointerup', stop);
+    return () => node.removeEventListener('pointerup', stop);
+  }, []);
```
sarahgm, line 162: ⚠️ This `useEffect` reads `dropdownOverlayRef.current` once on mount with `[]` deps. The overlay div only renders when `isMultiMonth === false`, so if `visibleDuration` ever toggles single↔multi at runtime, the listener never (re)attaches. Either include `isMultiMonth` in the deps or switch to a ref callback.

Also worth a one-line note on *why* a native `addEventListener` is required here: react-aria's window-level `pointerup` runs before React synthetic events bubble, so `onPointerUpCapture` wouldn't suppress it.

packages/components/src/RangeCalendar/RangeCalendar.tsx
```diff
@@ -0,0 +1,275 @@
...
+    ),
+  } satisfies { [key in ViewMapKeys]: React.JSX.Element };
+
+  const fieldErrorValue = useMemo(
+    () =>
+      error
+        ? {
+            isInvalid: true,
+            validationErrors: [],
+            validationDetails: {} as ValidityState,
```
sarahgm: 💡 `{} as ValidityState` lies to the type system. Consider typing `fieldErrorValue` via `ContextType<typeof FieldErrorContext>` and providing a populated stub (e.g. `{ valid: false, ... }`) once, so future readers don't have to verify the cast.

packages/components/src/RangeCalendar/RangeCalendar.tsx
```diff
@@ -0,0 +1,275 @@
...
+   * @default { months: 1 }
+   */
+  visibleDuration?: { months: number };
+  /**
+   * Controls how the calendar pages when navigating.
+   * - 'single': Page by one month at a time
+   * - 'visible': Page by the number of visible months
+   * @default 'visible'
+   */
+  pageBehavior?: 'single' | 'visible';
```
sarahgm: 💡 The `'single' | 'visible'` literal duplicates RAC's `pageBehavior` type. Prefer `Pick<RAC.RangeCalendarProps<T>, 'pageBehavior'>` to avoid drift if RAC adds a third variant.

packages/components/src/RangeCalendar/RangeCalendar.tsx
```diff
@@ -0,0 +1,275 @@
...
+  width?: WidthProp['width'];
+  /**
+   * Callback that is called for each date of the calendar. If it returns true, then the date is unavailable.
+   */
+  dateUnavailable?: RAC.RangeCalendarProps<T>['isDateUnavailable'];
+  /**
+   * The number of months to display at once. Up to 3 months are supported.
+   * @default { months: 1 }
+   */
+  visibleDuration?: { months: number };
```
sarahgm: 💡 The JSDoc says "Up to 3 months are supported" but nothing prevents `{ months: 12 }`. An optional dev-only warning or clamp would make the constraint explicit; fine to leave as-is for alpha.

packages/components/src/RangeCalendar/RangeCalendar.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+});
+
+export const Disabled = meta.story({
+  ...Basic.input,
+  args: {
+    disabled: true,
+  },
+});
+
+export const ReadOnly = meta.story({
```
sarahgm, line 101: do we really need for every state an own Story?

packages/components/src/RangeCalendar/RangeCalendar.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+        defaultValue={{
+          start: new CalendarDate(2019, 6, 5),
+          end: new CalendarDate(2019, 6, 10),
+        }}
+        onChange={onChange}
+        readOnly
+      />
+    );
+
+    await user.click(screen.getByLabelText(/Saturday, June 15, 2019/i));
```
sarahgm: Could be a component test or not? Because of interaction?

packages/components/src/RangeCalendar/RangeCalendar.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+    ).not.toBeInTheDocument();
+
+    await user.click(screen.getByRole('button', { name: '2025' }));
+
+    expect(
+      screen.getByRole('listbox', { name: 'yearOptions' })
+    ).toBeInTheDocument();
+  });
+
+  test('selecting an option from the month dropdown closes it', async () => {
```
sarahgm: Can also be a component test because its interactive?

packages/components/src/RangeCalendar/RangeCalendar.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+    ).not.toBeInTheDocument();
+
+    await user.click(screen.getByRole('button', { name: 'Aug' }));
+
+    expect(
+      screen.getByRole('listbox', { name: 'monthOptions' })
+    ).toBeInTheDocument();
+  });
+
+  test('opens the year dropdown when the year button is clicked', async () => {
```
sarahgm: Same here, interaction so Story test?

packages/components/src/RangeCalendar/RangeCalendar.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+    await user.click(options[2]);
+
+    expect(
+      screen.queryByRole('listbox', { name: 'monthOptions' })
+    ).not.toBeInTheDocument();
+  });
+});
+
+describe('RangeCalendar - Multi-month', () => {
```
sarahgm, line 57: Same on this one, Would also be a component test

packages/components/src/RangeCalendar/RangeCalendar.tsx
aromko: Fixed in 79d526a05.

Switched to a callback ref so the listener (re)attaches whenever the overlay node mounts (handles the single↔multi toggle). Also expanded the comment block to call out *why* a native listener is required — react-aria's listener fires on the native event before any synthetic event would reach a React capture handler.

packages/components/src/RangeCalendar/RangeCalendar.tsx
aromko: Fixed in 79d526a05.

`fieldErrorValue` is now typed as `ContextType<typeof FieldErrorContext>`, and a fully-populated `ValidityState` stub (`EMPTY_VALIDITY_STATE`) lives at module scope, so future readers can see the shape at a glance instead of verifying a cast.

packages/components/src/RangeCalendar/RangeCalendar.tsx
aromko: Fixed in 79d526a05.

Swapped the literal for `RAC.RangeCalendarProps<T>['pageBehavior']` (and mirrored the same change in `<Calendar>` for parity). Kept the JSDoc body so docs still document each variant.

packages/components/src/RangeCalendar/RangeCalendar.tsx
aromko: Tightened the type to `{ months: 1 | 2 | 3 }` in 79d526a05 (mirrored in `<Calendar>`). Compile-time only — no runtime warning needed, and the constraint now matches the JSDoc instead of relying on it.

packages/components/src/RangeCalendar/RangeCalendar.stories.tsx
aromko: Keeping the per-state stories for now — they mirror the `Calendar` story file 1:1, the docs registry pulls each story as a named example, and `__screenshots__` snapshots them individually. Args-only variation through the controls panel wouldn't give the docs site or the snapshot suite a stable anchor.

Happy to revisit if we move toward a controls-driven docs convention across the design system.

packages/components/src/RangeCalendar/RangeCalendar.test.tsx
aromko: Moved in 25f53d4f1 — this case is now covered by the `ReadOnlyDoesNotCommit` story (`tags: ['component-test']`, real browser via Playwright).

packages/components/src/RangeCalendar/RangeCalendar.test.tsx
aromko: Moved in 25f53d4f1 — covered by the new `YearDropdown` story (and the month equivalent `MonthDropdown`), both tagged `component-test`.

packages/components/src/RangeCalendar/RangeCalendar.test.tsx
aromko: Moved in 25f53d4f1 — `MonthDropdown` story covers open + select + close in stepped `play` blocks.

packages/components/src/RangeCalendar/RangeCalendar.test.tsx
aromko: Moved in 25f53d4f1 — the multi-month range-selection case is now the `TwoMonthsRangeSelection` story, and forward/back navigation continues to be exercised by `MultiMonthNavigation`. The render-only assertions (grid count, month labels) stay as plain unit tests.

sarahgm (approved):

PR #5386, opened 2026-05-04
feat([DST-1366]): slot-configurable primitives

sebald:
## Summary

Phase 0 of the slot-configuration rollout (DST-1366). Introduces the role-and-cascade vocabulary that Phase 1+ container migrations will consume:

- **Six new exports**: `Title`, `Description`, `TextValue` (text-bearing role primitives) and `ActionButton`, `ActionGroup`, `ActionMenu` (action primitives + cluster).
- **Three Marigold-owned slot contexts**: `ActionButtonContext`, `ActionGroupContext`, `ActionMenuContext`. Text and heading slots reuse RAC's `TextContext` / `HeadingContext` directly.
- **Existing primitives hardened**: `LinkButton` is now slot-aware; `ActionBar`'s legacy `ActionButton` is internalized as `ActionBar.Button`.
- **Docs**: dedicated MDX page per new primitive on the public docs site, slot-driven composition section in Foundations / Component principles, Foundations / Typography stub.
- **Typography prep**: `HeadlineSize`, `TextSize`, `TextVariant` exported so a future typography-token PR can replace runtime classes without touching consumer-facing prop types.

Ticket: [DST-1366](https://reservix.atlassian.net/browse/DST-1366)

## Background

The DST-1366 audit surfaced seven container components (`Panel`, `Dialog`, `Drawer`, `Tray`, `ContextualHelp`, `SectionMessage`, `EmptyState`) that each re-implement "title + description + actions" with bespoke wrappers. Phase 0 ships the shared primitives those containers will adopt in Phase 1+. **No container is migrated in this PR.**

The architectural rule established here: use RAC's `HeadingContext` / `TextContext` for text/heading slots; introduce Marigold-owned contexts only where RAC has no equivalent. Containers cascade defaults to nested primitives via a single `Provider`.

### Text role primitives

All three are thin wrappers that default a `slot` and let `HeadingContext` / `TextContext` do the styling. None carries typography props of its own; the host container or selection item owns size/variant/color/level via a `Provider`.

- **`<Title>`** — renders RAC `<Heading>`. Defaults `slot="title"` and `level={2}`. Participates in `HeadingContext` via `useContextProps`.
- **`<Description>`** — renders RAC `<Text>`. Defaults `slot="description"` and `as="span"`. Picks up `TextContext` natively through RAC's `<Text>`.
- **`<TextValue>`** — renders RAC `<Text>`. Defaults `slot="label"` and `as="span"`. Same shape as `<Description>` apart from the slot. Selection-item containers (ListBox, Menu, Select, ComboBox) style their primary text via the `label` slot of `TextContext`.

### Action role primitives

All three participate in Marigold-owned slot contexts and a shared cascade.

- **`<ActionButton>`** — wraps RAC `<Button>`. Owns `ActionButtonContext`. Reuses `<Button>`'s theme classes for v1 (bespoke visuals tracked as DST-1368).
- **`<ActionGroup>`** — wraps RAC `<Toolbar>`. Cascades `size`, `variant`, `disabled` to nested children via `ActionGroupContext`. Per-prop precedence:
  - `size`: group wins (visual uniformity).
  - `variant`: local wins (mix-in destructive into a uniform group).
  - `disabled`: local wins, group provides the default. `disabled={false}` on a child re-enables it.
- **`<ActionMenu>`** — composes its own `MenuTrigger` + `<ActionButton>` + `<Popover>`/`<Tray>` + RAC `<Menu>`. Owns `ActionMenuContext`. The trigger is `<ActionButton>` so it picks up the same cascade.

### Modified primitives

- **`<LinkButton>`** reads `ActionButtonContext` and `ActionGroupContext` via `useSlottedContext` (read-only consumption sidesteps the anchor/button ref-type mismatch that `useContextProps` would have created). Adds `destructive-ghost` to the variant union.
- **`<Headline>`** now exports `HeadlineSize`. No runtime change.
- **`<Text>`** now exports `TextSize` and `TextVariant`. No runtime change.

### Internalized API

- `ActionBar`'s top-level `ActionButton` slot is internalized and re-exposed as `ActionBar.Button`. Avoids the name collision with the new top-level `<ActionButton>`.

### Docs

- New: `docs/content/components/content/{title,description,text-value}`, `docs/content/components/actions/{action-button,action-group}`, `docs/content/components/overlay/action-menu`.
- Updated: `actionbar`, `button`, `link-button`, `menu` pages plus surrounding `meta.json` updates.
- New: `docs/content/foundations/typography` (stub) explaining where the heading and body scales live and where they're heading.
- Updated: `docs/content/foundations/component-principles` — adds a "Slot-driven composition" section explaining the mechanism.

## Deviations from DST-1366

The original ticket draft (`JIRA_DRAFT_phase0_primitives.md`) specced a particular shape for Phase 0. The implementation diverges in the following ways. Each deviation has a rationale.

### 1. New `<Title>` primitive added; `<Headline>` className-merge fix dropped

**Ticket said**: Modify `<Headline>` to `forwardRef` and fix the className-merge bug so context-injected classes from `HeadingContext` survive.

**We shipped**: A new `<Title>` primitive that defaults to `slot="title"` and participates in `HeadingContext` via `useContextProps`. `<Headline>` is unchanged at runtime (only a `HeadlineSize` type alias is extracted).

**Why**: The two responsibilities — "the typography scale primitive" and "the heading role inside a container" — pull in different directions. `<Headline>` accepts a rich set of typographic props (`align`, `lineHeight`, `color`, `variant`, `size`); `<Title>` is a thin role wrapper that picks up most of its configuration from the container. Fusing them in one component meant either bloating `<Headline>` with slot semantics or hobbling the role primitive's API. The role-vs-scale split also mirrors the `<Description>` / `<Text>` split that emerged organically from the same exercise.

### 2. `<Description>` and `<TextValue>` shipped more minimal than specced

**Ticket said**:
- `Description`: `variant`, `size`, `color`, `as`, `slot` props with its own theme entry (`Description.styles.ts`).
- `TextValue`: `as`, `slot`, `id` (no typography props), no theme entry.

**We shipped**: Both `<Description>` and `<TextValue>` carry no typography props and have no theme entry. Both render RAC's `<Text>` directly (not Marigold's `<Text>`). Apart from the default `slot`, the two are structurally identical.

**Why**: The original asymmetry between `Description` (with typography props) and `TextValue` (without) was a draft accommodation, not a design decision. Once implementation started the asymmetry created friction: when does a consumer reach for `<Description variant="muted">` vs. let the container decide? Resolving both as fully role-only states one principle clearly: the container owns typography; the role primitive owns the role. If a consumer needs typography overrides outside a container, they reach for `<Text>` (which carries the full styling vocabulary).

### 3. `<ActionGroup>` is a top-level component, not `ActionButton.Group`

**Ticket said**: Wrap children in RAC `<Toolbar>` and attach as `ActionButton.Group` compound.

**We shipped**: `<ActionGroup>` as its own top-level export with its own folder, docs page, and Storybook entry. The context renamed to match: `ActionGroupContext` (was `ActionButtonGroupContext`).

**Why**: The compound `.Group` form implied a tight coupling, but the toolbar wrapper is conceptually independent — it holds `<ActionButton>`, `<LinkButton>`, and (future) `<ToggleButton>`. As a top-level export the API surface stays flat, the cluster-vs-button distinction surfaces as a navigation distinction in the docs, and the cluster gets its own dedicated docs page.

### 4. `<LinkButton>` made slot-aware

**Ticket said**: No mention of `<LinkButton>`.

**We shipped**: `<LinkButton>` reads `ActionButtonContext` and `ActionGroupContext` so it picks up the same cascade as `<ActionButton>`. Adds `destructive-ghost` to its variant union.

**Why**: Real row-action clusters mix navigation and press actions ("Edit" → detail page, "Delete" → confirm). If only `<ActionButton>` cascades, the navigating sibling drifts visually inside an `<ActionGroup>`. Wiring `<LinkButton>` into the same context closes the gap with one small change. Read-only context consumption via `useSlottedContext` (rather than `useContextProps`) sidesteps the anchor/button ref-type mismatch.

### 5. `ActionBar`'s `ActionButton` internalized as `ActionBar.Button`

**Ticket said**: No mention.

**We shipped**: The legacy `ActionBar.ActionButton` is internalized and re-exposed as `ActionBar.Button`.

**Why**: The new top-level `<ActionButton>` collides with `ActionBar`'s old `ActionButton` slot. Internalizing it as `.Button` matches the rest of `ActionBar`'s compound vocabulary (`ActionBar.Title`, etc.) and avoids the collision.

### 6. "Slot configuration pattern" docs page replaced by Foundations / Component principles section

**Ticket said**: Ship a dedicated Storybook MDX page titled "Slot configuration pattern" with six sections (Why, The rule, Anatomy, Consumer recipe, Building a new slot-configurable primitive, Forward link).

**We shipped**: A "Slot-driven composition" section inside the existing `foundations/component-principles` docs page, plus per-primitive docs pages that show the pattern in context.

**Why**: A standalone "pattern" page read as architectural meta-content (how the system works) rather than usage guidance. The mechanism belongs in component-principles where the rest of the system architecture lives. Per-primitive pages already cover when and how to use each primitive, including the slot wiring, in the context where consumers will actually look for it.

### 7. Public docs site shipped (vs. Storybook-only)

**Ticket said**: "No public docs site changes (`docs/`). Phase 0 ships Storybook docs only."

**We shipped**: Six new MDX pages on the docs site for the new primitives, plus updates to surrounding pages (Button, LinkButton, ActionBar, Menu).

**Why**: Storybook-only docs would have been a stop-gap. Phase 0 introduces six consumer-facing exports — shipping them without docs site coverage would leave external consumers (theme-rui adopters, app teams) without discoverable usage guidance. Writing the docs alongside the primitives also pressure-tested the role-vs-scale split (`<Title>` vs. `<Headline>`, `<Description>` vs. `<Text>`). Several smaller refactors in this branch came out of that exercise.

### 8. `<ActionMenu>` rebuilt to compose its own popover (not delegated to Marigold's `<Menu>`)

**Ticket said**: Switch `<ActionMenu>`'s trigger from raw RAC `<Button>` to Marigold's `<ActionButton>`. Open question: "may require a small `Menu.tsx` refactor to allow injecting the trigger component."

**We shipped**: `<ActionMenu>` composes its own `MenuTrigger` + `<ActionButton>` + `Popover`/`Tray` + RAC `Menu`. Marigold's `<Menu>` is untouched.

**Why**: The Menu refactor turned out heavier than expected — `<Menu>` had structural assumptions about the trigger that weren't easy to parameterize without a wider rework. Building `<ActionMenu>` standalone kept `<Menu>` untouched and gave `<ActionMenu>` a cleaner architecture (different composition: ActionButton trigger + popover + RAC menu, vs. Menu's own combined approach).

### 9. Typography type aliases (additional)

**Ticket said**: No mention of cross-primitive type linkage.

**We shipped**: `HeadlineSize` exported from `Headline`. `TextSize` and `TextVariant` exported from `Text`. The aliases aren't yet consumed by other primitives — the role primitives shipped without typography props (see deviation 2) — but exposing them now lets a future typography-token PR replace runtime classes without rewriting consumer-facing prop types.

### 10. ActionGroup cascade rules formalized

**Ticket said**: ActionButton.Group "cascades size/variant to nested `<ActionButton>`s" (informal).

**We shipped**: Explicit per-prop precedence rules baked into `ActionGroup.tsx` JSDoc, mirrored in `ActionMenu.tsx` JSDoc, and covered by unit + story tests.

**Why**: Cascade behavior is the whole point of the cluster primitive. Without explicit precedence rules consumers would have to read the source to know what overrides what. Each rule was picked to match its most common case (uniformity for `size`, mix-in flexibility for `variant`, default-with-exception for `disabled`).

## How to review

Suggested order:

1. **`packages/components/src/{ActionButton,ActionGroup,Title,Description,TextValue}/Component.tsx`** — the six new primitives. Each is short. Confirm `useContextProps` wiring, slot defaults, and that the role primitives stay genuinely minimal.
2. **`packages/components/src/Menu/ActionMenu.tsx`** + `ActionMenuContext.ts` — the standalone composition. Confirm the trigger uses `<ActionButton>` and the popover/tray dispatch logic still fires for small screens.
3. **`packages/components/src/LinkButton/LinkButton.tsx`** — the read-only slot-context consumption. Confirm the precedence chain (`local prop ?? slot context ?? group context`) and the `destructive-ghost` addition.
4. **`packages/components/src/ActionBar/`** — the rename from `ActionButton.tsx` to `ActionBarButton.tsx`. Confirm `ActionBar.Button` still works for existing consumers.
5. **Tests** — `*.test.tsx` for each new primitive plus `LinkButton.test.tsx`. Each cascade rule has a dedicated test.
6. **Docs** — start at `docs/content/components/actions/action-group/index.mdx` for the cluster reference; then per-primitive pages.

## Risk & backwards compatibility

- **No public API change to existing released components.** `ActionBar.Button` continues to work; `ActionBar.ActionButton` was a beta API.
- **`<LinkButton>` behavior change inside `<ActionGroup>`.** Standalone `<LinkButton>` usage is untouched. Consumers placing `<LinkButton>` inside an `<ActionGroup>` will now see it pick up the group's cascade. Intentional and previously inconsistent (`<ActionButton>` cascaded; `<LinkButton>` did not).
- **`<Headline>` and `<Text>` runtime unchanged.** Only type aliases extracted; the inline unions are structurally identical to what they replaced.
- **Existing `<ActionMenu>` consumers.** API surface preserved; internal composition rebuilt. Stories visually verified.
- **`<Menu>` untouched.** Phase 0 does not touch the existing Menu component.

## Test plan

- [x] `pnpm typecheck:only` passes.
- [x] `pnpm test:unit` passes.
- [x] Storybook component-test stories pass (`pnpm test:sb`).
- [x] Docs build cleanly with the new MDX pages.
- [x] Visual smoke test of each new docs page after the theme-rui preflight fix from #5396 (tooltip placement above triggers).
- [ ] Reviewer confirms `ActionBar.Button` consumers see no behaviour change.
- [ ] Reviewer confirms existing `<Headline>` and `<Menu>` stories are visually unchanged.
- [ ] Reviewer confirms cascade rules behave as documented in `<ActionGroup>` (size group-wins, variant local-wins, disabled local-wins-with-default).

## Follow-ups

- **DST-1367** — Phase 1: refactor `<Panel>` to consume the new primitives.
- **DST-1369** — Dialog/Drawer/Tray adopt the same pattern.
- **DST-1370** — ContextualHelp/SectionMessage/EmptyState adopt the same pattern.
- **DST-1364** — ListBox item label/description styling using RAC's `TextContext` directly.
- **DST-1368** — bespoke `<ActionButton>` visual design (currently reuses `<Button>`'s theme classes).

[DST-1366]: https://reservix.atlassian.net/browse/DST-1366


.changeset/slot-configuration-primitives.md
```diff
@@ -0,0 +1,27 @@
+---
+'@marigold/components': patch
+'@marigold/theme-rui': patch
+'@marigold/system': patch
+---
+
+feat(DST-1366): introduce slot-configurable primitives
+
+Phase 0 of the slot-configuration pattern. Adds three text-bearing role primitives — `Title`, `Description`, `TextValue` — and three action primitives — `ActionButton`, `ActionGroup`, `ActionMenu` — that participate in slot-keyed context. Text/heading slots use React Aria's `HeadingContext` / `TextContext` directly; action slots use Marigold-owned contexts (`ActionButtonContext`, `ActionGroupContext`, `ActionMenuContext`) consumed via `useContextProps`.
```
sarahgm: Nobody know what Phase 0 is

.changeset/slot-configuration-primitives.md
```diff
@@ -0,0 +1,27 @@
...
+
+`LinkButton` is now slot-aware: it picks up `ActionButtonContext` and `ActionGroupContext` so a navigating action can sit alongside `<ActionButton>` inside an `<ActionGroup>` and inherit the same cascade. A `destructive-ghost` variant is added to match `<ActionButton>`. Context is consumed read-only (via `useSlottedContext`) to sidestep the anchor/button ref-type mismatch that `useContextProps` would have created.
+
+`<ActionBar>`'s legacy top-level `ActionButton` slot is internalized and re-exposed as `ActionBar.Button`. Existing consumers that already use `<ActionBar.Button>` are unaffected.
+
+Typography prep: `Headline` exports `HeadlineSize`, `Text` exports `TextSize` and `TextVariant`. The aliases aren't yet consumed by other primitives, but exposing them now lets a future typography-token PR replace runtime classes without rewriting consumer-facing prop types.
+
+Consumer container migrations follow in Phase 1+ (Panel — DST-1367; Dialog/Drawer/Tray — DST-1369; ContextualHelp/SectionMessage/EmptyState — DST-1370; ListBox item — DST-1364; bespoke ActionButton visual — DST-1368).
```
sarahgm: wouldn't also mention this

packages/components/src/Menu/ActionMenu.tsx
```diff
@@ -1,14 +1,146 @@
...
+
+  /**
+   * Whether the menu trigger is disabled.
+   */
+  disabled?: boolean;
+
+  /**
+   * A slot to place the element in.
+   */
+  slot?: string | null;
```
sarahgm: ⚠️ **Slot type inconsistency.** Every other primitive declares `slot?: string;` (`ActionButton.tsx:43`, `ActionGroup.tsx:36`, `Title.tsx:19`, `Description.tsx:19`, `TextValue.tsx:19`). `ActionMenu` is alone in accepting `string | null`. Either normalize:

```suggestion
  slot?: string;
```

…or add a one-line JSDoc explaining what `null` means here.

packages/components/src/index.ts
```diff
@@ -16,9 +16,15 @@ export type {
...
-export type { ActionButtonProps } from './ActionBar/ActionButton';
 export { useActionBar } from './ActionBar/useActionBar';
+
+export { ActionButton } from './ActionButton/ActionButton';
+export type { ActionButtonProps } from './ActionButton/ActionButton';
+export { ActionButtonContext } from './ActionButton/Context';
+
+export { ActionGroup } from './ActionGroup/ActionGroup';
+export type { ActionGroupProps } from './ActionGroup/ActionGroup';
+export { ActionGroupContext } from './ActionGroup/Context';
```
sarahgm, line 27: ⚠️ **Public surface commitment.** Exporting `ActionButtonContext`, `ActionGroupContext`, and `ActionMenuContext` (lines 23, 27, 167) makes them part of the stable public API once `beta-release` exits. The PR description shows the intent (containers cascade via `<Provider values={…}>`), and there's precedent (`ConfirmationContext` is exported the same way). Worth a deliberate call now — and consider documenting them as extension points on the per-primitive docs pages, since they're not currently surfaced there.

packages/components/src/Title/Title.tsx
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * The element id.
+   */
+  id?: string;
+  children?: ReactNode;
+  ref?: Ref<HTMLHeadingElement>;
+}
+
+const _Title = ({ ref: refProp, ...inputProps }: TitleProps) => {
+  const [props, ref] = useContextProps(
```
sarahgm: 💡 **Could be simplified to match `Description` / `TextValue`.** RAC's `<Heading>` already reads `HeadingContext` internally when given a `slot`, so the explicit `useContextProps` call here is doing what RAC would do anyway. Aligning the three role primitives on one shape would make the "all three are thin wrappers" claim in the changeset literally true. Optional — both versions work.

packages/components/src/Title/Title.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+      control: { type: 'select' },
+      options: [1, 2, 3, 4, 5, 6],
+    },
+    slot: {
+      control: { type: 'text' },
+    },
+  },
+});
+
+export const Basic: any = meta.story({
```
sarahgm: 💡 **Missing `component-test` tag.** `TextValue.stories.tsx:31-50` tags a play function for browser-test coverage; `Title.stories.tsx` and `Description.stories.tsx` don't. Consider adding at least one tagged story per role primitive for parity, or document why these rely solely on `.test.tsx` unit coverage.

packages/components/src/Title/Title.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+test('receives "level" from a HeadingContext slot', () => {
+  render(
+    <Provider values={[[HeadingContext, { slots: { title: { level: 4 } } }]]}>
+      <Basic.Component data-testid="title">Hi</Basic.Component>
+    </Provider>
+  );
+
+  expect(screen.getByTestId('title').tagName).toBe('H4');
+});
```
sarahgm, line 32: 💡 **Combined precedence is untested.** The three tests cover default `level=2`, explicit `level=3`, and context-only `level=4` independently. Missing: explicit `level={3}` *inside* a context providing `level={4}` — i.e., does the local prop win? Expected behavior is yes (`useContextProps` default), but it's worth a four-line assertion to lock it in.

packages/components/src/Menu/ActionMenu.tsx
```diff
@@ -1,14 +1,146 @@
...
+
+  /**
+   * Visual variant of the trigger button.
+   */
+  variant?: 'default' | 'ghost' | (string & {});
+
+  /**
+   * Size of the trigger button.
+   */
+  size?: 'default' | 'small' | 'large' | 'icon' | (string & {});
```
sarahgm, line 58: 💡 **`large` / `icon` sizes in a cluster?** Inside an `<ActionGroup>`, `size` is group-wins (`ActionButton.tsx:74`), and clusters are typically uniform default/small. If the intent is to allow these sizes only when `<ActionMenu>` is used outside an `ActionGroup`, a short JSDoc note would clarify. Otherwise prune the union.

packages/components/src/Text/Text.tsx
```diff
@@ -25,6 +25,34 @@ import {
...
+ * token in a future PR.
+ */
+export type TextVariant = 'default' | 'muted';
+
+/**
+ * The shared body-text size scale. Other body primitives derive their
+ * size unions from this via `Extract<TextSize, ...>`. Will become a
+ * typography token in a future PR.
+ */
+export type TextSize =
```
sarahgm: Why not using `textSize` from `style-props.tsx` ?

docs/package.json
```diff
@@ -45,6 +45,7 @@
     "next": "16.2.3",
     "nuqs": "2.8.8",
     "react": "^19.2.3",
+    "react-aria-components": "^1.17.0",
```
sarahgm: why is this needed now ?

packages/components/src/ActionButton/ActionButton.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const Basic = meta.story({
+  tags: ['component-test'],
+  args: {
+    children: 'Action',
+    onPress: fn(),
+  },
+  render: args => <ActionButton {...args} />,
+  play: async ({ canvas, args }) => {
+    const button = canvas.getByRole('button', { name: 'Action' });
+    await userEvent.click(button);
```
sarahgm, line 32: missing whitespaces between each AAA block

packages/components/src/ActionButton/ActionButton.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+test('disabled prop maps to RAC isDisabled', async () => {
+  const onPress = vi.fn();
+  render(
+    <Basic.Component disabled onPress={onPress} data-testid="action">
+      Click
+    </Basic.Component>
+  );
+
+  await user.click(screen.getByTestId('action'));
```
sarahgm: If you have interactions should be a component test instead?

packages/components/src/ActionGroup/ActionGroup.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+      </ActionButton>
+      <ActionButton aria-label="Duplicate">Duplicate</ActionButton>
+      <ActionButton aria-label="Delete">Delete</ActionButton>
+    </ActionGroup>
+  ),
+  play: async ({ canvas }: any) => {
+    const toolbar = canvas.getByRole('toolbar', { name: 'Item actions' });
+    await expect(toolbar).toBeInTheDocument();
+
+    const buttons = canvas.getAllByRole('button');
```
sarahgm, line 37: this is not following AAA

packages/components/src/Description/Description.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+const _Description = ({
+  as = 'span',
+  slot = 'description',
+  children,
+  ref,
+  ...rest
+}: DescriptionProps) => (
+  <Text elementType={as} slot={slot} ref={ref} {...rest}>
```
sarahgm: Is Description not more like a `HelpText` ?

sebald: No, helptext is something related to a field and it can be either a description or an error. I had a conversation with the C dude about this :D

.changeset/slot-configuration-primitives.md
sebald: Good catch. Dropped the 'Phase 0' lead-in and the matching 'Phase 1+' enumeration at the bottom in 4d77547e0. Also stripped the suffix from the PR title.

.changeset/slot-configuration-primitives.md
sebald: Dropped together with the 'Phase 0' lead-in (4d77547e0).

packages/components/src/Menu/ActionMenu.tsx
sebald: Good catch. Went the other way — extended all eight slot-bearing primitives to RAC's canonical `slot?: string | null` so the opt-out escape hatch is consistently available. Centralized the prop in a new `SlotProps` interface in `packages/components/src/types.ts` so the JSDoc lives in one place and primitives just `extends SlotProps`. Forwarding sites cast through `string | undefined` because RAC's `HeadingProps` / `TextProps` extend `HTMLAttributes` (slot: string), but `useContextProps` accepts null at runtime so the semantic still works. Commit c2fa85619.

packages/components/src/index.ts
sebald: Confirming deliberate. The slot-configuration pattern's whole point is that containers publish context values to drive their nested primitives, so app code (and any custom container built on top of Marigold) needs the context references. Keeping it consistent with the `ConfirmationContext` precedent.

Docs follow-up: each primitive's MDX page should call out `<ActionButtonContext.Provider value={...}>` as the extension point. Tracking that for the docs pass when these graduate from alpha (DST-1369 / DST-1370 are the natural place since they introduce the first real consumer containers).

packages/components/src/Title/Title.tsx
sebald: Thought about this and want to keep the explicit `useContextProps` in Title. The reason is the level-precedence model.

With `useContextProps` called explicitly:
- precedence is `default ← context ← local`
- the destructure default `level = 2` is applied AFTER context resolution, so a container that publishes `{ level: 4 }` via `HeadingContext` can drive a stretch of nested Titles to `<h4>`

Without it (matching Description/TextValue):
- precedence flips to `default-as-local ← context`
- the destructure default `level = 2` is applied BEFORE `<Heading>` runs its own `useContextProps`, and since useContextProps treats anything non-undefined as a local override, that default 2 then wins over a context-provided level
- the `level=4 via HeadingContext` case would silently regress to `<h2>`

That's exactly the behavior comment 7 asks us to lock in with a test, so the explicit call earns its keep. Description and TextValue don't need this because neither has a value prop a container would inject (only a slot key, which RAC's `<Text>` handles).

Reworded the changeset to stop overselling the symmetry between the three role primitives in d525230b0 — the three are no longer described as uniformly thin wrappers, since Title does the extra step on purpose.

docs/content/components/actions/action-group/action-group-row-actions.demo.tsx
```diff
@@ -0,0 +1,63 @@
...
+  },
+  {
+    id: 'evt-3',
+    title: 'Oak Ridge Wedding Showcase',
+    subtitle: 'Sat, 5 Jul · Oak Ridge Barn',
+  },
+];
+
+export default () => (
+  <Stack space={1}>
```
sarahgm: SShould use spacing tokens

packages/components/src/Title/Title.stories.tsx
sebald: Added matching tagged play stories to `Title.stories.tsx` and `Description.stories.tsx` (08ad026f8). All three role primitives now carry equivalent browser-test coverage. Title's play asserts the default level (h2) renders, Description's asserts the text renders.

docs/content/components/actions/action-group/action-group-row-actions.demo.tsx
sarahgm: same for other demos

docs/content/components/actions/action-group/action-group-toolbar.demo.tsx
```diff
@@ -0,0 +1,46 @@
+import { AtSign, Image, Link, Paperclip, SmilePlus } from 'lucide-react';
```
sarahgm: Why not. @marigold/icons?

packages/components/src/Title/Title.test.tsx
sebald: Added in a7af60bc2. The new test renders Title with explicit `level={3}` inside a `HeadingContext` that publishes `level=4` and asserts the rendered element is `<h3>`. Combined with the existing three cases that pins the precedence contract: `default ← context ← local`.

docs/content/components/actions/action-group/action-group-row-actions.demo.tsx
```diff
@@ -0,0 +1,63 @@
+import { Eye, Pencil, Trash2 } from 'lucide-react';
```
sarahgm: why not @marigold/icons?

sarahgm: same for other demos

packages/components/src/Menu/ActionMenu.tsx
sebald: Kept the full `default | small | large | icon` union. The trigger is just an `<ActionButton>` underneath, and all four sizes are plausible for a standalone ActionMenu (e.g. `size="icon"` for the typical ellipsis-only trigger, `size="large"` for a CTA-style dropdown). Inside an `<ActionGroup>` the group's size wins anyway so the local value just becomes inert — that doesn't seem like a reason to forbid the wider palette outside the group.

Clarified the JSDoc on both `ActionMenu.size` and `ActionButton.size` (for symmetry) in 45b64d096 so the precedence rule is explicit at the prop site.

packages/components/src/Text/Text.tsx
sebald: Good catch. The union was hand-typed but `textSize` already lives in `style-props.tsx` with exactly those keys. Derived `TextSize = 'default' | keyof typeof textSize` in 1ee97a311 so adding a key to the record now automatically extends the type. `'default'` stays as an explicit literal because it is not in the record (the cva variants resolve it to the theme's default size class).

docs/package.json
sebald: The new Title docs page has an in-container demo (`docs/content/components/content/title/title-in-container.demo.tsx`) that imports `HeadingContext` and `Provider` from `react-aria-components` directly:

```tsx
import { HeadingContext, Provider } from 'react-aria-components';
import { Description, Stack, Title } from '@marigold/components';

<Provider values={[[HeadingContext, { slots: { title: { level: 3 } } }]]}>
  …
</Provider>
```

The demo is intentionally raw-RAC to show how Title hooks into the underlying `HeadingContext`. Marigold doesn't re-export RAC primitives (`HeadingContext`, `Provider`) because they have a stable home in `react-aria-components` and adding them to Marigold's public surface would just duplicate RAC.

Since `react-aria-components` is already a transitive dep via `@marigold/components`, the runtime cost is zero — the explicit line in `docs/package.json` is just to satisfy module resolution for the direct import.

packages/components/src/ActionButton/ActionButton.stories.tsx
sebald: Added blank lines between arrange / act / assert in the Basic play function (599a368cd). Matches the convention used in Switch, Checkbox, etc.

packages/components/src/ActionButton/ActionButton.test.tsx
sebald: Agreed. Moved the disabled-onPress assertion into the Disabled story as a tagged play function (599a368cd) and removed the jsdom test. The story now clicks the disabled button in a real browser and asserts onPress was not called, which is a more honest test of the disabled behavior anyway (jsdom's pointer-events handling can disagree with real browsers).

docs/package.json
sebald: Reconsidered — re-exporting `HeadingContext`, `TextContext`, and `Provider` from `@marigold/components` (matching the I18nProvider re-export pattern) in 761a95c08. The title-in-container demo now imports everything from `@marigold/components`, and `react-aria-components` is dropped from `docs/package.json`. Cleaner story for consumers too: anyone wiring a container can stay on a single `@marigold/components` import.

packages/components/src/ActionGroup/ActionGroup.stories.tsx
sebald: Fair. Split the story in 21cf028ba so each one reads as a single AAA block:

- `Basic` is now arrange + assert only (toolbar exists, three buttons present, no act step).
- `KeyboardNavigation` arranges the three buttons, acts by focusing the first and pressing `ArrowRight` twice, then asserts focus lands on the third.

Cleaner than the old combined block.

docs/content/components/content/description/description-with-title.demo.tsx
```diff
@@ -0,0 +1,13 @@
+import { Description, Stack, Title } from '@marigold/components';
+
+export default () => (
+  <div className="rounded border p-4">
+    <Stack space={2}>
```
sarahgm: use spacing tokens

packages/components/src/ActionButton/ActionButton.stories.tsx
sebald: Done in 599a368cd — added blank lines between the Arrange / Act / Assert blocks in the Basic story (and matched the same in the new Disabled play story).

packages/components/src/ActionButton/ActionButton.test.tsx
sebald: Agree. Moved the disabled-onPress assertion out of the unit test in 599a368cd — it now lives as a tagged play story `Disabled` in `ActionButton.stories.tsx`, asserting `onPress` was not called after clicking the disabled button. The unit test file no longer interacts with the DOM.

packages/components/src/ActionGroup/ActionGroup.stories.tsx
sebald: Split the original Basic story into two in 21cf028ba. The new `Basic` covers Arrange + Assert only (toolbar + 3 buttons). The interaction moved into a separate `KeyboardNavigation` story where Arrange / Act / Assert reads cleanly: focus the first button, press ArrowRight twice, assert the third has focus.

docs/content/components/actions/action-group/action-group-row-actions.demo.tsx
sebald: Done in 154d588e4. Replaced the raw scale values with semantic spacing tokens — `space="tight"` for title+meta stacks, `space="related"` for the rows list.

sebald: Same sweep applied across the other new demos in 154d588e4: action-group-toolbar, action-group-panel-chrome, button-vs-action-button, panel-header-actions, panel-appearance, panel-danger-zone-rows, description-with-title, title-in-container, text-value-in-menu, switch-card, card-appearance, card-elevation. Kept numeric `Inline` values where they encode visual proximity for specific layouts (image+text rows, price+rating clusters) — happy to revisit any specific one if you want them tokenised too.

docs/content/components/actions/action-group/action-group-toolbar.demo.tsx
sebald: Switched in 154d588e4. `@marigold/icons` re-exports lucide so all the icon names stay identical, but the demo import now reads `from "@marigold/icons"`.

docs/content/components/actions/action-group/action-group-row-actions.demo.tsx
sebald: Same change for this demo in 154d588e4.

sebald: Same swap across the other new demos in this PR that imported from `lucide-react`: action-button-in-context, actionbutton-appearance, button-vs-action-button, action-group-toolbar, action-group-panel-chrome, panel-header-actions, drawer-bulk-edit. All in 154d588e4.

packages/components/src/ActionGroup/ActionGroup.tsx
```diff
@@ -0,0 +1,80 @@
...
+  variant?: string;
+  /**
+   * Cascades the size to nested actions. The group's size wins over a
+   * child's local `size` so the cluster stays visually uniform.
+   */
+  size?: string;
+  /**
+   * Disables every nested action by default. A child can re-enable itself
+   * with `disabled={false}`.
+   * @default false
```
aromko: ⚠️ `@default false` is misleading — `disabled` propagates as `undefined` unless the consumer passes it (no default substitution in the destructure below). Either drop this JSDoc line or default `disabled = false` explicitly.

packages/components/src/Menu/ActionMenu.tsx
```diff
@@ -1,14 +1,146 @@
...
+    children,
+    variant,
+    size,
+    disabled,
+    open,
+    placement,
+    trigger,
+    defaultOpen,
+    onOpenChange,
+    'aria-label': ariaLabel,
```
aromko, line 99: ⚠️ Only `aria-label` is forwarded to the trigger; other ARIA labelling props (`aria-labelledby`, `aria-describedby`) fall through into `menuProps` and end up on the inner `RACMenu` instead of the `<ActionButton>`. Spread the full `aria-*` set onto the trigger or document the limitation.

packages/components/src/ActionGroup/ActionGroup.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+    },
+    size: {
+      control: { type: 'select' },
+      options: ['default', 'small', 'large', 'icon'],
+    },
+    disabled: { control: { type: 'boolean' } },
+  },
+});
+
+export const Basic: any = meta.story({
```
aromko: ⚠️ `: any` on the story export and `({ canvas }: any)` callbacks violate the project's no-`any` rule (`CLAUDE.md`). The inferred `StoryObj` type from `preview.meta(...).story(...)` should work without the cast — address the underlying generic rather than escaping with `any`. Same pattern repeats across the other four new story files.

packages/components/src/Description/Description.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+      control: { type: 'select' },
+      options: ['p', 'span', 'div'],
+    },
+    slot: {
+      control: { type: 'text' },
+    },
+  },
+});
+
+export const Basic: any = meta.story({
```
aromko: ⚠️ Drop `: any` on both the export and the `render` callback signature — same issue as in `ActionGroup.stories.tsx`.

packages/components/src/Title/Title.stories.tsx
aromko: ⚠️ Drop `: any` on the export and `render` signatures (lines 19, 23, 26) — same issue as in `ActionGroup.stories.tsx`.

packages/components/src/ActionBar/ActionBarButton.tsx
```diff
@@ -2,7 +2,7 @@ import type { ReactNode } from 'react';
 import RAC, { Button } from 'react-aria-components';
 import { useClassNames } from '@marigold/system';
 
-export interface ActionButtonProps {
+export interface ActionBarButtonProps {
```
aromko, line 5: ⚠️ The public surface is narrowed to `onPress` + `children` only — no `ref`, no `isDisabled`, no `data-*`, no `aria-*`. If any existing consumer was passing extra attributes through `ActionBar.ActionButton`, they'd be silently dropped. PR description marks the old API as beta — please confirm the reviewer checklist item ("`ActionBar.Button` consumers see no behaviour change") before merging.

packages/components/src/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,107 @@
...
+  slot?: string;
+  ref?: Ref<HTMLButtonElement>;
+}
+
+export const ActionButton = ({
+  ref: refProp,
+  ...inputProps
+}: ActionButtonProps) => {
+  const [merged, ref] = useContextProps(
+    inputProps as ActionButtonProps & { className?: string },
```
aromko: 💡 Consider exporting `ActionButtonContextValue = ActionButtonProps & { className?: string }` and reusing it both here and in `ActionButton/Context.ts`. `ActionMenuContext` already does this — the asymmetry between the two contexts is mildly confusing for `Provider` authors.

packages/components/src/Description/Description.tsx
```diff
@@ -0,0 +1,40 @@
...
+  id?: string;
+  children?: ReactNode;
+  ref?: Ref<HTMLElement>;
+}
+
+const _Description = ({
+  as = 'span',
+  slot = 'description',
+  children,
+  ref,
```
aromko, line 28: 💡 `Description` and `TextValue` are structurally identical apart from the default `slot`. The role distinction defended in PR deviation 2 stands, but a private `createSlottedText({ defaultSlot })` factory would prevent drift between the two implementations as the codebase evolves.

packages/components/src/ActionGroup/ActionGroup.tsx
```diff
@@ -0,0 +1,80 @@
...
+    className: contextClassName,
+    ...rest
+  } = merged;
+
+  // Publish a `ghost` baseline so any action consumer renders low-emphasis
+  // without each one repeating the fallback.
+  const ctx = useMemo(
+    () => ({ variant: variant ?? 'ghost', size, disabled }),
+    [variant, size, disabled]
+  );
```
aromko, line 68: 💡 Worth surfacing the hardcoded `'ghost'` baseline in the JSDoc on `variant` (around line 16) — consumers see an unset group and don't realise `ghost` is still being cascaded to descendants.

packages/components/src/ActionGroup/ActionGroup.tsx
sebald: Dropped the misleading `@default false` JSDoc line in 21cddc70c. `disabled` stays `undefined` when unset on purpose — the cascade contract in `ActionButton` is `ownDisabled ?? groupCtx?.disabled`, so a literal `false` default would suppress the group's disabled state.

packages/components/src/Menu/ActionMenu.tsx
sebald: Fixed in 21cddc70c. Destructured `aria-labelledby` and `aria-describedby` alongside `aria-label` and forwarded all three to the `<ActionButton>` trigger. Other `aria-*` (e.g. `aria-expanded`, `aria-haspopup`) are owned by RAC's `MenuTrigger` and stay untouched.

packages/components/src/ActionGroup/ActionGroup.stories.tsx
sebald: Stripped `: any` from `Basic`, `KeyboardNavigation`, `DisabledCascade`, `CascadePrecedence`, `WithActionMenu` (plus the three play callbacks) in 8c03bd5b2. `pnpm typecheck:only` and `pnpm lint` are clean.

For context on why the cast existed: the LSP surfaces TS2883 because RAC's internal `DOMRenderFunction` leaks into the inferred `StoryObj` type, and that type can only be named via `react-aria-components/dist/types/src/utils`. `tsc --noEmit` (no declaration emit) accepts it, the build accepts it, and the story files are not part of the package's emitted `.d.ts`, so the LSP noise is editor-only and doesn't block the no-`any` rule.

packages/components/src/Description/Description.stories.tsx
sebald: Same fix applied here in 8c03bd5b2 — see the reply on `ActionGroup.stories.tsx` for context on why the LSP `TS2883` warning lingers (editor-only, doesn't affect typecheck or build).

packages/components/src/Title/Title.stories.tsx
sebald: Stripped `: any` from `Basic`, `Levels`, `Renders` and the render callback in 8c03bd5b2. Same applies across `Description.stories.tsx`, `TextValue.stories.tsx`, `ActionButton.stories.tsx`.

packages/components/src/ActionBar/ActionBarButton.tsx
sebald: Confirmed: `ActionBar.Button` is a byte-for-byte port of the legacy `ActionBar.ActionButton` on `beta-release` (same `children` + `onPress` minimal surface). The rename does not narrow anything — no behaviour change for existing consumers.

Widening `ActionBar.Button` to forward `aria-*` / `disabled` / `ref` is a real ask but a separate concern (would need a deliberate look at ActionBar's interaction model). Filing a follow-up ticket. ✅ on the reviewer checklist item.

packages/components/src/ActionButton/ActionButton.tsx
sebald: Done in 21cddc70c. Exported `ActionButtonContextValue` from `ActionButton/Context.ts` and `ActionMenuContextValue` from `Menu/ActionMenuContext.ts`, and replaced both inline `as ActionXProps & { className?: string }` casts with the named aliases. The two contexts are now symmetric and `Provider` authors can import the type.

packages/components/src/Description/Description.tsx
sebald: Thought about the factory. Going to leave the duplication in place.

Both wrappers are explicitly role-locked (deviation 2: no typography props, no theme entry), so they're *supposed* to stay near-identical. A `createSlottedText({ defaultSlot })` factory would put an abstraction layer between the consumer-facing name and the rendering code for ~6 lines of dedupe — and the moment either role grows a role-specific prop (e.g. `errorState` on `Description`, `valueFormat` on `TextValue`), the factory would need to bail. The drift risk is low enough that the duplication earns its keep here.

packages/components/src/ActionGroup/ActionGroup.tsx
sebald: Done in 21cddc70c. Extended the `variant` JSDoc to call out the hidden `'ghost'` baseline and added `@default 'ghost'` so consumers see it in tooltips.

aromko (approved):

PR #5390, opened 2026-05-05
refa([DST-1324]): 🧹 disambiguate space prop semantics across layout components

sarahgm:
# Description

Resolves [DST-1324](https://reservix.atlassian.net/browse/DST-1324). Parent context: [DST-1306](https://reservix.atlassian.net/browse/DST-1306).

The `space` prop carried two different meanings in Marigold:

- **Gap between children** on `Stack`, `Inline`, `Columns`, `Grid`, `Tiles`, `Center`, `Container`, `Aside`, `Panel` (`SpacingTokens`).
- **Padding inside the container** on `Inset` (`InsetSpacingTokens` / `PaddingSpacingTokens`).

`Panel` had already split the two with `space` (gap) + `p`/`px`/`py` (padding). `Inset` was the lone outlier. This PR makes `Inset` match `Panel` so that across the design system **`space` always means gap and `p`/`px`/`py` always mean inner padding**.

## Migration

| Before | After |
| --- | --- |
| `<Inset space="…" />` | `<Inset p="…" />` |
| `<Inset spaceX="…" spaceY="…" />` | `<Inset px="…" py="…" />` |

Internal CSS custom properties also renamed (mirrors Panel's `--panel-*` namespacing): `--space` → `--inset-p`, `--spaceX` → `--inset-px`, `--spaceY` → `--inset-py`.

The discriminated union shape and token vocabularies (`InsetSpacingTokens` for `p`, `PaddingSpacingTokens` for `px`/`py`) are unchanged.

## Drive-by cleanups in the same files

- Forward `aria-*` / `id` / `role` from `AriaRegionProps` to the Inset div (was previously typed but never spread). Added a unit test.
- Replaced pre-DST-1162 `<Card>{children}</Card>` wrappers in `inset-hv.demo.tsx` and the `Inset` Basic story with `<div className="bg-muted rounded-md">`. The wrappers existed only to make the inset visible against a contrasting background; using a plain muted div keeps the demos focused on `Inset` itself and decouples them from the `Card` API (which has since moved to compound `Card.Header` / `Card.Body` / `Card.Footer`).
- Dropped a leftover `data-hello` debug attribute on `VisualInset`.
- Aligned `VisualInset` (docs visualization helper) so its props mirror Inset's `px`/`py`.

## Why no codemod / RFC doc

The ticket listed "deprecation strategy + codemod" and a separate decision record as scope items. Discussed during planning and reduced scope to: hard rename in v18-beta + regex migration in the changeset (no published `jscodeshift`, no separate RFC). The rename is mechanical and the in-repo surface is small.

# Test Instructions:

```bash
pnpm typecheck:only   # clean
pnpm test:unit        # 1111 passed (added 1 aria-forwarding test)
pnpm test:sb          # 132 passed
pnpm lint             # no new warnings on changed files
```

Visual smoke-test in the docs site:

1. `pnpm start` → `/components/layout/inset` — both demos (`inset-equal`, `inset-hv`) should render with visible padding axes.
2. `/foundations/spacing` — the `spacing-composition` and `spacing-inset` demos should render correctly (the second uses `VisualInset`).
3. `pnpm sb` → open `Components/Inset` — the `Basic` story should expose `p` / `px` / `py` controls (no longer `space`/`spaceX`/`spaceY`).

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

- [ ] Updated visual regression tests (N/A — VRT enforced on PRs into `main`, not `beta-release`)

[DST-1324]: https://reservix.atlassian.net/browse/DST-1324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DST-1306]: https://reservix.atlassian.net/browse/DST-1306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/Inset/Inset.stories.tsx
```diff
@@ -67,14 +66,14 @@ const meta = preview.meta({
...
   },
   args: {
-    space: 'square-regular',
+    p: 'square-regular',
     children: undefined,
   } as const,
 });
 
 export const Basic = meta.story({
   render: args => (
```
aromko, line 75: 💡 The PR description (and changeset migration note) attributes this swap to "`Card.Body` would stack its own `px-4 py-2` on top of Inset's, defeating the demos." But the original story used `<Card>` directly — `Card` root applies no padding (only `rounded-md shadow-elevation-raised`, see `themes/theme-rui/src/components/Card.styles.ts`). The swap itself is still reasonable (decouples the Inset demo from `Card`), but the stated reason doesn't match the code being replaced. Worth tightening the rationale before it lands in the migration changelog. Same applies to the matching change in `docs/content/components/layout/inset/inset-hv.demo.tsx`.

packages/components/src/Inset/Inset.tsx
```diff
@@ -13,37 +13,38 @@ export type InsetProps =
...
+export const Inset = ({ p, px, py, children, ...props }: InsetProps) => (
   <div
+    {...props}
     className={cn(
-      space !== undefined && 'p-(--space)',
-      spaceX !== undefined && 'px-(--spaceX)',
-      spaceY !== undefined && 'py-(--spaceY)'
+      p !== undefined && 'p-(--inset-p)',
+      px !== undefined && 'px-(--inset-px)',
+      py !== undefined && 'py-(--inset-py)'
```
aromko, line 42: 💡 Now that these CSS custom properties are namespaced (`--inset-p` / `--inset-px` / `--inset-py` instead of bare `--space`), they're effectively a public-ish API surface that consumers may inspect when debugging or extending. A one-liner JSDoc on the component — or a short note in the Inset MDX — calling out the variable names would help discoverability.

packages/components/src/Inset/Inset.stories.tsx
sarahgm: Good catch, the rationale was wrong. The original wrappers were bare `<Card>`, not `<Card.Body>`, and `Card`'s root applies no padding (only `rounded-md shadow-elevation-raised`, see `themes/theme-rui/src/components/Card.styles.ts`). The swap itself still stands: it decouples the demos from `Card` and makes the padding axes visible against a muted background. I've tightened the PR description to reflect the actual reason. Thanks!

packages/components/src/Inset/Inset.tsx
sarahgm: Thought about this one and I'd lean toward leaving it for now. `Panel` exposes a parallel set of namespaced vars (`--panel-px` / `--panel-py` / `--panel-gap`) and we don't document those in JSDoc or in its MDX either. Promoting only `Inset`'s set to a documented surface would diverge from that precedent and quietly commit us to maintaining the variable names as a contract. If we do want to expose these as a public CSS hook for consumers, I think it's worth its own cross-component decision so we can do `Panel` and `Inset` together. Resolving for now.

aromko (approved):

PR #5400, opened 2026-05-08
fix([DST-1412]): fix Calendar/RangeCalendar width handling

aromko:
## Summary

Fixes [DST-1412](https://reservix.atlassian.net/browse/DST-1412):

- **Single-month**: `RangeCalendar` was applying width twice (FieldBase + inner AriaRangeCalendar). Removed the duplicate by dropping `width={width}` from `FieldBase` so only the inner element handles the width. (commit 75d4f6268)
- **Multi-month at `width="1/2"`**: With three months, the calendar wrapper shrank to 50% of its parent while content needed ~750px, causing the third month to overflow visually. Switched the calendar's minimum from `min-w-[250px]` to `min-w-fit` so it always grows to fit its natural content (single-month ≈ 280px, three-month ≈ 750px).
- **Multi-month at `width="full"`**: Each `calendarMonth` column expanded with `flex-1` but the inner `<table>` (auto-width) stayed compact, leaving header text floating over empty space. Added `w-full` to `calendarGrid` so the date table fills its column.

## Test plan

- [x] `pnpm test:unit` — 1116 tests pass
- [x] `pnpm test:sb` — 136 storybook play tests pass
- [x] Visual check in Storybook:
  - RangeCalendar Basic (single, fit): natural size, range fill spans cells
  - RangeCalendar Basic at `width="full"`: cells expand, range fill continuous
  - RangeCalendar ThreeMonths at `width="1/2"`: calendar grows to fit all 3 months
  - RangeCalendar ThreeMonths at `width="full"`: 3 columns evenly distributed, dates fill columns
  - Calendar Basic at `width="full"`: 36×36 selected date stays centered in wider columns
  - DatePicker popover: calendar fits content naturally

[DST-1412]: https://reservix.atlassian.net/browse/DST-1412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/RangeCalendar/RangeCalendar.stories.tsx
```diff
@@ -209,7 +209,7 @@ export const ThreeMonths = meta.story({
 
 export const TwoMonthsMobile = meta.story({
   globals: {
-    viewport: { value: 'smallScreen' },
+    viewport: { value: 'mobile1' },
```
sarahgm, line 212: you said to me a few month ago that we should use our breakpoint for this :D

aromko: Maybe, but then you probably didn't have a good reason not to use them months ago 😅 

Why I use this?

Going in the direction of B2C I think we need to support 320px because they do it right now?! I use the breakpoint from storybook for now because one requirement for them was that the range calendar is usable on 320px. But like you see I didn't extend our marigold breakpoints so that we can decide this together or as part in the project.

sarahgm: I had a reason ;) I found it helpful hat Storybook supports device specific breakpoints already. When you find it helpful that we support 320px we should maybe think about a new breakpoint called xs for example?

sarahgm: I think we said that we want to support the value which is used when `useSmallScreen`.

aromko: "But like you see I didn't extend our marigold breakpoints so that we can decide this together or as part in the project." -> Like I've written we should discuss this with the B2C project in mind and I don't think it's helpful I think it is necessary if we support the B2C.

For now I would let it this way because I want to make sure that the RangeCalendar fits into 320px so the VRT should break if not. I will create a [ticket](https://reservix.atlassian.net/jira/software/projects/DST/boards/134?selectedIssue=DST-1417) for this to discuss and like you mentioned to adjust our breakpoints if we agree as a team.

sarahgm: Its no blocker for merging this but I would expect every component should fit into 320px screens. I would approve this, and add my thoughts on the ticket you wrote

sarahgm (approved):

PR #5407, opened 2026-05-11
refactor([DST-1367]): Panel adopts the slot-configuration pattern

sebald:
## Summary

Panel becomes the canonical real-component adopter of the slot-configuration pattern.

- **`Panel.Header` is a single `<Provider>`** that configures `HeadingContext`, `TextContext`, `ActionButtonContext`, and `ActionGroupContext` for everything nested inside it. Consumers drop slot-aware role primitives directly into the header — `<Title>`, `<Description>`, and any of `<ActionButton>`, `<ActionGroup>`, `<ActionMenu>`, `<LinkButton>` — and the Panel injects level, ids, ref wiring, grid-area positioning, and the size/variant cascade via context.
- **Five compound sub-components removed**: `Panel.Title`, `Panel.Description`, `Panel.HeaderActions`, `Panel.CollapsibleTitle`, `Panel.CollapsibleDescription`. The slot-aware role primitives subsume their responsibilities; the actions grid cell is published as a className on `ActionButtonContext` / `ActionGroupContext` so bare action primitives land in the right cell. Multiple actions belong inside `<ActionGroup>` so the cluster claims one cell as a toolbar.
- **`Panel.CollapsibleHeader` adopts the same shape**: drop a `<Title>` and an optional `<Description>` inside it. `Panel.CollapsibleHeader` publishes `HeadingContext` and `TextContext` slot configs that force the primitives to render as `<span>` inside the disclosure trigger button (heading elements can't legally nest inside a `<button>`); the structural `<hN>` semantics come from `Panel.CollapsibleHeader` itself.
- **Title-only-without-header shortcut preserved**: a bare `<Title>` used as a direct child of `<Panel>` still wires up correctly. `Panel.tsx` publishes a minimal `HeadingContext` at the root with just `{ level, id, ref }`; `Panel.Header` publishes the enriched value (adds the grid-area className) that wins for descendants inside the header.

Ticket: [DST-1367](https://reservix.atlassian.net/browse/DST-1367)

### Source

- `packages/components/src/Panel/Panel.tsx` — publishes a minimal `HeadingContext` at the Panel root (slot `title`: `level`, `id`, `ref`, plus the panel's horizontal padding className) so a bare `<Title>` used without `<Panel.Header>` still wires up. Drops `Panel.Title`, `Panel.Description`, `Panel.HeaderActions`, `Panel.CollapsibleTitle`, `Panel.CollapsibleDescription` attachments and imports; refreshes JSDoc to point at the slot-aware primitives.
- `packages/components/src/Panel/PanelHeader.tsx` — wraps the existing grid `<div>` in a `<Provider>` that publishes:
  - `HeadingContext` slot `title` — grid-area class, `headingLevel`, panel `titleId`, panel `titleSlotRef`. Replaces the panel root's value for descendants of the header.
  - `TextContext` slot `description` — grid-area class, `elementType: 'p'`.
  - `ActionButtonContext` (unslotted) — `[grid-area:actions] self-center …classNames.actions`, `size: 'icon'`.
  - `ActionGroupContext` (unslotted) — same className + size, so a toolbar claims the cell as a whole.
- `packages/components/src/Panel/PanelCollapsibleHeader.tsx` — replaces the bespoke `CollapsibleHeaderProvider` with `HeadingContext` + `TextContext` slot Providers inside the trigger button. Title's `as: 'span'` and Description's `elementType: 'span'` are delivered via slot config so the primitives nest cleanly inside the disclosure button. The structural outer `<Heading>` uses `slot={noSlot}` to opt out of the Panel root's title slot.
- `packages/components/src/Panel/PanelTitle.tsx`, `PanelDescription.tsx`, `PanelHeaderActions.tsx`, `PanelCollapsibleTitle.tsx`, `PanelCollapsibleDescription.tsx`, `CollapsibleContext.tsx` — **deleted**.
- `packages/components/src/Description/Description.tsx` — drops the `as` prop and the hard-coded `elementType` forwarding. Description now consumes `elementType` from its surrounding `TextContext` slot config (`<p>` in `Panel.Header`, `<span>` in `Panel.CollapsibleHeader`, RAC's default `<span>` elsewhere).
- `packages/components/src/Title/Title.tsx` — passes `slot={noSlot}` to the underlying RAC `<Heading>` so the slot config consumed by `useContextProps` isn't re-applied (preventing duplicated classNames). The `as` override is delivered via slot context only (`Panel.CollapsibleHeader` sets `as: 'span'`).
- `packages/components/src/utils/noSlot.ts` — new utility for the `null`-cast-to-`string | undefined` opt-out idiom; consolidates the pattern across 7 call sites (Title, Headline, ContextualHelp, Calendar, Accordion, Collapsible, PanelCollapsibleHeader).

### Stories

- `Panel.stories.tsx` migrated to bare slot primitives. The three `<Panel.HeaderActions>` blocks in `WithHeaderActions` map to: a single `<ActionButton>`, a single `<ActionMenu>`, and an `<ActionGroup>` of `<ActionButton>` + `<ActionMenu>`. Redundant `variant='ghost' size='icon'` props dropped where the cascade now supplies them.
- New `SlotsActionGroup` story tagged `component-test` — covers `role="toolbar"`, arrow-key navigation between three clustered actions, and confirms inner `<ActionButton>`s don't carry the grid-area className.
- `TitleOnlyWithoutHeader` story preserved (tagged `component-test`) — asserts `aria-labelledby` wires the panel region to the title when used without `<Panel.Header>`.

### Tests

- `Panel.test.tsx` rewritten to match the new structure: the "Panel.Title / Panel.Description / Panel.HeaderActions" describe-blocks become "Title in Panel.Header / Description in Panel.Header / Panel.Header" with assertions on the slot-aware rendering. The "labels the panel region when used without Panel.Header" test is kept and now runs against the slot-context wiring. Compound-exports test trimmed to the surviving sub-components.
- `Description.test.tsx` adds a "honors `elementType` from a TextContext slot" test covering the new context-driven element type behavior.

### Docs

- `docs/content/components/layout/panel/index.mdx` — "The header configures its children" section added; "Title-only Panels can skip the header" section preserved and updated to describe the slot-context behavior. `Panel.Title` / `Panel.Description` / `Panel.HeaderActions` AutoTypeTable entries removed.
- `docs/content/releases/blog/release-2026-06-30.mdx` — Panel snippets updated to the slot-primitive pattern.
- 11 Panel docs + action docs demos migrated. 14 example apps under `docs/app/(examples)/examples/` migrated.

### Changeset

`.changeset/dst-1367-panel-slot-configuration.md` — `patch` for `@marigold/components`.

## Architectural principle established

> **Positional `className` flows through slot contexts and is absorbed at the first layout boundary. Container slot-aware components MUST republish (scrub) the descendant slot contexts so positional className does not leak past their cell.**

`<ActionGroup>` enforces this at its own toolbar boundary. Future container refactors (Dialog, Drawer, Tray, ContextualHelp, SectionMessage, EmptyState, Card) follow the same pattern.

## Risk & backwards compatibility

- **Breaking change to `Panel`'s public API**: `Panel.Title`, `Panel.Description`, `Panel.HeaderActions`, `Panel.CollapsibleTitle`, and `Panel.CollapsibleDescription` are removed. No production consumers depend on them (Panel is on `beta-release` and unreleased).
- **Breaking change to `Description`'s public API**: the `as` prop is removed. Element type is now decided by the surrounding `TextContext` slot config — `<p>` inside `Panel.Header`, `<span>` inside `Panel.CollapsibleHeader`, RAC's default `<span>` everywhere else. No production consumers passed `as` (verified by codebase search).
- **Bare `<Button>` inside `<Panel.Header>` is a deliberate footgun**: it isn't slot-aware, doesn't pick up the cascade, and renders outside the grid cell. Consumers should use `<ActionButton>` / `<LinkButton>` / `<ActionMenu>` / `<ActionGroup>`.

## Test plan

- [x] `pnpm typecheck:only` passes (workspace + `pnpm --filter @marigold/docs types:check` clean apart from one pre-existing `@marigold/icons` `Add` import unrelated to this work).
- [x] `pnpm test:unit` — Panel + dependent suites pass.
- [x] New `SlotsActionGroup` story has component-test assertions for toolbar a11y.
- [x] `TitleOnlyWithoutHeader` story has component-test assertion for the aria-labelledby wiring.
- [x] `Description.test.tsx` covers `elementType` delivered via `TextContext` slot config.
- [ ] Reviewer confirms Chromatic visual parity for existing Panel stories.
- [ ] Reviewer confirms `pnpm test:sb` (Storybook component-test runner) is green — including the new `SlotsActionGroup` toolbar play function.

## Follow-ups

- **DST-1369** — Dialog / Drawer / Tray adopt the same pattern.
- **DST-1370** — ContextualHelp / SectionMessage / EmptyState adopt the same pattern.
- **DST-1364** — ListBox item label/description using RAC's `TextContext` directly.
- **DST-1368** — bespoke `<ActionButton>` visual design.

[DST-1367]: https://reservix.atlassian.net/browse/DST-1367


packages/components/src/Accordion/AccordionHeader.tsx
```diff
@@ -25,7 +26,7 @@ export const AccordionHeader = ({ children }: AccordionHeaderProps) => {
           : undefined
       }
     >
-      <Heading>
+      <Heading slot={noSlot}>
```
sarahgm, line 29: What has this change todo with the Panel Slot Configuration? Same fir the CalendarHeader?

packages/components/src/Panel/PanelCollapsibleHeader.tsx
```diff
@@ -42,18 +78,30 @@ export const PanelCollapsibleHeader = ({
...
-          slot="trigger"
-          aria-labelledby={titleId}
-          aria-describedby={hasDescription ? descriptionId : undefined}
-          className={cn('flex-1', classNames.collapsibleHeader)}
+    // Panel root publishes a slot-keyed HeadingContext (for the title-only
+    // shortcut); `slot={noSlot}` opts this structural Heading out of it.
+    <Heading level={level} slot={noSlot} className="flex">
+      <Button
+        slot="trigger"
+        aria-labelledby={titleId}
```
sarahgm, line 84: ⚠️ `aria-labelledby={titleId}` always references this ID, but `titleId` is only landed on a DOM node when a `<Title>` is rendered inside this header. The `Description` path tracks presence via `useSlot(false)` → `hasDescription` and conditionally adds `aria-describedby` — the title path doesn't do the same.

If a consumer renders `<Panel.CollapsibleHeader>` with only a `<Description>` (or no children at all), the button's accessible name resolution silently fails.

Options:
- Track local title presence with `useSlot` and either fall back or omit `aria-labelledby` when absent, **or**
- Tighten the JSDoc on line 23 from "Typically a `<Title>` and an optional `<Description>`" to make the `<Title>` requirement explicit.

packages/components/src/Panel/PanelHeader.tsx
```diff
@@ -1,28 +1,82 @@
...
+        description: {
+          className: cn('[grid-area:description]', classNames.description),
+          elementType: 'p' as const,
+        },
+      },
+    }),
+    [classNames.description]
+  );
+
+  const actionProps = useMemo(
```
sarahgm, line 64: 💡 `actionProps` is published as the same value on both `ActionButtonContext` and `ActionGroupContext`. This works because both contexts read a compatible shape, but it relies on convention rather than type. A short comment — similar to the helpful one on lines 22–26 above for `headingProps` — explaining that the two contexts intentionally share this value (and why `size: 'icon'` is the right cascade for both a bare action and a toolbar of actions) would help future readers.

sarahgm (commented):
Automated review via Claude Code `review-pr` skill — 1 warning, 2 suggestions, no critical issues. Targets `beta-release` so VRT isn't required.

packages/components/src/Accordion/AccordionHeader.tsx
sarahgm: Ok found the comment in Headline.tsx that explains this,

packages/components/src/Accordion/AccordionHeader.tsx
```diff
@@ -25,7 +26,7 @@ export const AccordionHeader = ({ children }: AccordionHeaderProps) => {
           : undefined
       }
     >
```
aromko, line 28: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files across `packages/components/src/Panel/**`, multiple header primitives (Accordion/Calendar/Collapsible/ContextualHelp/Headline), `Title`, `Description`, and 11+ docs demos. The `Visual-Regression-Tests` workflow has not run for this branch.

Please trigger it by commenting `/run-chromatic` on this PR — the test plan's "Chromatic visual parity" checkbox is still unchecked.

packages/components/src/Panel/PanelCollapsibleHeader.tsx
```diff
@@ -42,18 +78,30 @@ export const PanelCollapsibleHeader = ({
...
+        aria-describedby={hasDescription ? descriptionId : undefined}
+        className={cn('flex-1', classNames.collapsibleHeader)}
+      >
+        <Provider
+          values={[
+            // `HeadingContext` is typed against RAC's `HeadingProps`; Title
+            // reads an additional `as` slot key that lives outside that shape.
+            [
+              HeadingContext,
+              headingProps as ContextValue<HeadingProps, HTMLHeadingElement>,
```
aromko: ⚠️ Cross-element type cast on the Provider value.

`headingProps as ContextValue<HeadingProps, HTMLHeadingElement>` lies about the rendered element: when the slot supplies `as: 'span'` (line 54), the ref points at an `HTMLSpanElement`, not an `HTMLHeadingElement`. The comment explains why, but the same shape is reproduced in `Title.tsx:45`. Consider promoting `TitleSlotValue` to a shared type (or declaration-merging RAC's `HeadingProps`) so the `as` extension is typed once, instead of cast at each consumer site.

packages/components/src/Title/Title.tsx
```diff
@@ -18,25 +19,52 @@ export interface TitleProps extends AriaLabelingProps, SlotProps {
...
-  const { level = 2, slot, children, ...props } = merged;
+  // `slot` is consumed by `useContextProps` above; we drop it so the spread
+  // below cannot reintroduce it on the rendered element.
+  const {
+    level = 2,
+    slot: _slot,
+    children,
+    as,
+    ...props
+  } = merged as TitleSlotProps;
```
aromko: ⚠️ Same root cause as the `as ContextValue<…>` cast in `PanelCollapsibleHeader.tsx`. The `as` slot key is a Marigold extension to RAC's `HeadingProps`. Worth solving once at the type layer rather than repeating the cast at each Provider/consumer pair.

packages/components/src/Panel/PanelHeader.tsx
```diff
@@ -1,28 +1,82 @@
...
 
 export interface PanelHeaderProps {
   /**
-   * Content of the header. Typically a `Panel.Title`, optional
-   * `Panel.Description`, and optional `Panel.HeaderActions`.
+   * Content of the header. Typically a `<Title>`, optional `<Description>`,
+   * and optional action primitives (`<ActionButton>`, `<ActionGroup>`,
+   * `<ActionMenu>`, `<LinkButton>`).
    */
   children: ReactNode;
```
aromko, line 18: 💡 The PR body calls out that a bare `<Button>` inside `Panel.Header` is a deliberate footgun (won't pick up the size/grid-area cascade). That guidance isn't in the JSDoc or the docs page — anyone hovering `PanelHeaderProps.children` in their editor won't see it. A one-liner here would surface it:

```suggestion
  /**
   * Content of the header. Typically a `<Title>`, optional `<Description>`,
   * and optional action primitives (`<ActionButton>`, `<ActionGroup>`,
   * `<ActionMenu>`, `<LinkButton>`).
   *
   * Use the action primitives above for header chrome — a bare `<Button>`
   * is intentionally not slot-aware and won't pick up the size/grid cascade.
   */
  children: ReactNode;
```

aromko (commented):
Automated code review — 6 inline comments. Overall the refactor is high quality; the main blocker before merge is running Chromatic VRT.

*Generated with Claude Code review-pr skill.*

packages/components/src/Description/Description.tsx
```diff
@@ -22,16 +17,15 @@ export interface DescriptionProps
...
   ...props
 }: DescriptionProps) => (
-  // `slot` may be `null` (opt out of inherited slot context). RAC's
-  // `TextProps` narrows slot to `string`, but `useContextProps` (used
-  // internally by `<Text>`) accepts `null` at runtime.
-  <Text elementType={as} slot={slot as string | undefined} ref={ref} {...props}>
+  // The element type is decided by the surrounding `TextContext` slot config
+  // (`<p>` inside `<Panel.Header>`, `<span>` by RAC's default elsewhere).
+  // `slot` may be `null` to opt out of inherited slot context.
+  <Text slot={slot as string | undefined} ref={ref} {...props}>
```
aromko, line 30: 💡 `slot as string | undefined` may be redundant depending on the resolved type of `DescriptionProps['slot']` (via `SlotProps`). If `SlotProps['slot']` is already typed as `string | undefined`, the cast can go. Quick check; drop if so.

packages/components/src/Panel/Panel.test.tsx
```diff
@@ -203,29 +197,28 @@ describe('Panel.Header', () => {
...
     render(<WithHeaderActions.Component />);
 
     const description = screen.getByText(
       /People with access to this workspace/
     );
-    const actionsSlot = screen
-      .getByRole('button', { name: /Invite member/ })
-      .closest('[class*="[grid-area:actions]"]');
 
     expect(description.className).toContain('[grid-area:description]');
```
aromko: 💡 These assertions check for the literal Tailwind arbitrary-value class (`[grid-area:description]`, `[grid-area:actions]`). They're coupled to the current class syntax — a refactor that switches to `data-grid-area="description"` would break the test without breaking the behavior. Not blocking, but a `closest('[data-panel-header]')`-style assertion would survive that refactor.

packages/components/src/Description/Description.tsx
sebald: Checked — the cast isn't redundant. `SlotProps['slot']` is intentionally `string | null` (the `null` is how a consumer opts out of inherited slot context). After the destructure default `slot = 'description'`, the local `slot` is `string | null` — defaults only replace `undefined`, not `null`. RAC's `Text.slot` is typed `string | undefined`, so the cast bridges the nullable opt-out to RAC's narrower type; `useContextProps` inside `Text` accepts `null` at runtime.

I've tightened the inline comment to spell this out so future readers don't have to re-derive it.

sarahgm (approved):

packages/components/src/Accordion/AccordionHeader.tsx
sebald: no running against beta

aromko (approved):

PR #5417, opened 2026-05-15
refactor([DST-1419]): polish full-page data filter example

sebald:
## Summary

Polish pass on `docs/app/(examples)/examples/filter/` to make it a copy-worthy reference. Mirrors structure of the production `organizer-search-app` template.

- **Structure**: split into PascalCase components (`Toolbar`, `AppliedFilter`, `VenuesTable`, `VenuesPagination`) + a `hooks/` folder (`useFilter`, `useSearch`, `useSort`, `usePagination`, `useVenues`). Single page shell composes them.
- **State**: single `defaultFilter` constant as source of truth (`VenueFilter = typeof defaultFilter`); sentinel-based "no filter" semantics; `setFilterFromForm` encapsulates form coercion.
- **Pipeline**: `useVenues` owns search → filter → sort → paginate; called by both `VenuesTable` and `VenuesPagination` (nuqs deduplicates URL subscriptions).
- **A11y**: single live region (`role="status"` implies `aria-live="polite"`), removed duplicate "Venues" landmark, `prefers-reduced-motion` guard on scroll, region anchor for scroll-on-page-change.
- **Pagination layout**: now matches `pagination/index.mdx` docs pattern (results indicator | pagination | results-per-page Select), extracted into its own `Panel.Content` for correct horizontal padding while the table stays in a `bleed` content area.
- **Cleanup**: removed memo theatre on the 10-row static dataset; eliminated internal ticket references and placeholder buttons; trimmed `as readonly string[]` cast; CSV export with Safari race-condition fix.

Ticket: https://reservix.atlassian.net/browse/DST-1419

## Test plan

- [ ] Visit `/examples/filter` — toolbar, applied filters, table, pagination render with correct padding
- [ ] Search → URL updates `?q=`, page resets to 1
- [ ] Apply filter → URL updates, applied-filter tags appear, page resets to 1
- [ ] Sort by column → URL updates with `column`/`direction`
- [ ] Change page / page size → URL updates, table scrolls back to first row
- [ ] Refresh on a deep-linked URL → state restored
- [ ] Empty state (no matches) → "Clear all filters" works
- [ ] CSV export from row selection works in Safari
- [ ] `prefers-reduced-motion` → scroll on page change is instant


aromko (commented):
## Automated Code Review

### Overview
| Field | Value |
|-------|-------|
| Author | @sebald |
| Branch | `dst-1419-polish-filter-example` → `beta-release` |
| Files Changed | 14 |
| Additions | +617 / Deletions | -562 |
| CI | All green |
| Visual Regression | Not needed (docs example only — no `packages/**`, `themes/**`, `.storybook/**`, `*.stories.tsx`, or `*.styles.ts` changes) |

### Linked Ticket
**[DST-1419](https://reservix.atlassian.net/browse/DST-1419)** — 🧹 Polish full-page data filter example (PR #5357 follow-ups). Status: **In Review**.

Excellent ticket coverage — essentially every scope item is addressed. Highlights:

- ✅ Single source of truth via `type VenueFilter = typeof defaultFilter`
- ✅ `hasFilter()`/`activeFilterKeys()` iterate keys instead of naming fields
- ✅ `utils.tsx` split into `hooks/` + `AppliedFilter.tsx`
- ✅ Form coercion moved into `setFilterFromForm`; `key={JSON.stringify(filter)}` documented
- ✅ `<VenueRow>` and `<VenuesPagination>` extracted; three `useMemo`s removed (correct call for the static dataset)
- ✅ Single live region via `role="status"`; "Venues" landmark deduplicated; `prefers-reduced-motion` honored on scroll
- ✅ `exportToCsv` extracted to `csv.ts` with deferred-cleanup Safari fix
- ✅ Internal ticket comments and placeholder buttons stripped
- ✅ Page size now configurable via `Results per page` Select
- ✅ Remaining `as` casts all carry inline documenting comments

### Issues Found

#### Critical (Must Fix)
_None._

#### Warnings (Should Fix)

1. **`docs/app/(examples)/examples/filter/hooks/useVenues.ts:28`** — `safePage = Math.min(page, totalPages)` doesn't guard against `page < 1`. A URL like `?page=0` or `?page=-5` (typed manually, bookmarked, or from buggy back-button state) produces a negative `safePage`, and `display.slice((safePage - 1) * pageSize, safePage * pageSize)` becomes a negative-index slice that returns an unexpected window. Suggested fix:

   \`\`\`ts
   const safePage = Math.max(1, Math.min(page, totalPages));
   \`\`\`

1. **`VenuesTable.tsx:9`** and **`VenuesPagination.tsx:30`** — `VENUES_REGION_ID = 'venues-region'` and `id=\"page-size\"` are hardcoded module-scope IDs. Fine for a single-mount docs example, but copy-pasted into a real app where `<VenuesTable>` could be rendered twice they'd produce duplicate IDs. Since this example is meant to be copy-worthy, consider `useId()` to teach the safer pattern.

2. **`hooks/useFilter.ts:31`** — `isActive` via string interpolation (`` `${filter[key]}` !== `${defaultFilter[key]}` ``) is clever but subtle. It relies on Array→`.join(',')` coercion, which would silently break if the schema ever gains a non-array, non-primitive value. A slightly more explicit form:

   \`\`\`ts
   const isActive = (filter: VenueFilter, key: FilterKeys) => {
     const value = filter[key];
     return Array.isArray(value)
       ? value.length > 0
       : value !== defaultFilter[key];
   };
   \`\`\`

3. **`hooks/useFilter.ts:65`** — `price >= MAX_PRICE ? defaultFilter.price : price` is a no-op in value terms (since `defaultFilter.price === MAX_PRICE`), but functions as a \"URL param clear when at max\" hook. A one-line comment explaining the intent would help readers who'd otherwise try to simplify it.

### Recommendation
**Approve** ✅

Thorough, well-executed polish pass. The new structure is far more copy-worthy than what shipped in #5357, the a11y cleanup is on-target, and the documenting comments on the remaining `as` casts make the typing gaps actionable for primitive follow-ups. Only the `safePage` negative-index edge case is worth fixing before merge; everything else is polish-level.

---

docs/app/(examples)/examples/filter/AppliedFilter.tsx
```diff
@@ -0,0 +1,71 @@
+'use client';
```
sarahgm, line 1: 💡 Once `page.tsx` is marked `'use client'`, every component imported from it (`AppliedFilter`, `Toolbar`, `VenuesTable`, `VenuesPagination`) is automatically in the client tree — the directive isn't required on this file (or the other three child files).

Two reasonable readings:
- Drop the directive on children and rely on Next.js's tree-inheritance (less noise).
- Keep them and add a one-line comment explaining the intent is "every file is independently copyable" so readers don't cargo-cult the directive without understanding it.

Either is fine — just worth being deliberate about which message you want the example to send.

docs/app/(examples)/examples/filter/hooks/useSort.ts
```diff
@@ -0,0 +1,17 @@
...
+const sortDirections = ['ascending', 'descending'] as const;
+
+const sortParsers = {
+  column: parseAsStringLiteral(sortColumns).withDefault('name'),
+  direction: parseAsStringLiteral(sortDirections).withDefault('ascending'),
+};
+
+export type VenueSortDescriptor = inferParserType<typeof sortParsers>;
+
+export const useSort = () => useQueryStates(sortParsers, { history: 'push' });
```
sarahgm, line 19: 💡 `useSearch` and `useFilter` reset `page` to 1 on every change; `useSort` does not. The asymmetry is *correct* (changing sort order doesn't change item count, so the current page stays meaningful), but it isn't documented anywhere — which makes it look like an oversight at first read.

One line above the export would prevent the next reader from "fixing" it:

```suggestion
// Unlike useSearch/useFilter, sorting doesn't reset page — changing order
// preserves item count, so the current page stays meaningful.
export const useSort = () => useQueryStates(sortParsers, { history: 'push' });
```

docs/app/(examples)/examples/filter/VenuesTable.tsx
```diff
@@ -0,0 +1,191 @@
...
+      <Inline space="0.5">
+        {venue.amenities.map((a, i) => (
+          <Badge key={`${a}-${i}`}>{amenitiesOptions[a]}</Badge>
+        ))}
+      </Inline>
+    </Table.Cell>
+    <Table.Cell>
+      <Inline space="0.5">
+        {venue.parking.map((p, i) => (
+          <Badge key={`${p}-${i}`}>{parkingOptions[p]}</Badge>
```
sarahgm: Here is a Lint warning, did you saw it?: `Do not use item index in the array as its key`

docs/app/(examples)/examples/filter/AppliedFilter.tsx
sebald: Verified the Next.js inheritance claim — correct, the directive on `AppliedFilter` / `Toolbar` / `VenuesTable` / `VenuesPagination` is redundant once `page.tsx` is `'use client'`.

Went with the second option (keep, document the intent). The other examples in `docs/app/(examples)/examples/*` (e.g. `event-form`, `auto-save-settings`) also carry `'use client'` on every child file, so this matches the established convention. Added a comment block on `page.tsx` in 22e7d69 explaining that each file keeps its own directive so individual components stay copy-pasteable.

docs/app/(examples)/examples/filter/hooks/useSort.ts
sebald: Applied the suggested comment verbatim in 22e7d69. Good catch — without the doc the next reader would absolutely "fix" this.

docs/app/(examples)/examples/filter/VenuesTable.tsx
sebald: Confirmed — `eslint` flagged lines 102 and 109. Each venue's `amenities` and `parking` arrays carry unique indices (checked the dataset in `lib/data/venues.ts`), so the `${value}-${index}` key was only needed because the original codemod was over-cautious. Fixed in 22e7d69 by keying on the value alone (`key={a}` / `key={p}`).

sarahgm (approved):

aromko (approved):

PR #5418, opened 2026-05-15
fix([DST-1436]): drop background from hover transitions for snappier feel

sebald:
## Summary

- Removes `background` from `transition-[…]` declarations across `ui-surface`, `ui-surface-contrast`, and high-frequency interactive components — hover background flips are now instant.
- Fixes an in-family inconsistency: solid `Button` variants (primary/secondary, via `ui-surface`) transitioned the bg while ghost/destructive variants were already instant. Same hover, two feels. Now consistently instant.
- Color, border, box-shadow, and transform transitions are preserved (~150ms `ease`) — only the background flip is snapped.

## Motivation

Per **Emil Kowalski's frequency principle** ("interactions users see 100+ times/day shouldn't be animated") and Google's **RAIL model** (visual feedback within ~100ms feels instant). A 150ms bg color transition on Buttons, Tabs, Table rows, etc. sits at the edge of perception and makes the UI feel slower than it should. Linear / Raycast / Vercel / GitHub all flip hover backgrounds instantly for the same reason.

## Files changed

| File | Change |
|---|---|
| `themes/theme-rui/src/ui.css` | `ui-surface` & `ui-surface-contrast` → `transition-[border,box-shadow]` |
| `Tabs.styles.ts` | `transition-[color,background-color]` → `transition-[color]` |
| `Table.styles.ts` | Row: dropped `transition-[background-color]`. Edit buttons (×3): `transition-[color,background,transform]` → `transition-[color,transform]` |
| `LegacyTable.styles.ts` | `transition-colors` → `transition-[color,border-color]` (avoids the Tailwind shorthand silently including bg) |
| `Sidebar.styles.ts` | navLink + backButton: `transition-[color,background-color]` → `transition-[color]` |
| `Calendar.styles.ts` | calendarCellBase: `transition-[color,background-color]` → `transition-[color]` |
| `SelectList.styles.ts` | `transition-[border,color,background]` → `transition-[border,color]` |
| `ActionBar.styles.ts` | clearButton: `transition-[color,background-color]` → `transition-[color]` |

## Out of scope

- `FileField.dropZone` — bg transition is for drag-target / focus, not hover
- `Switch.track` — bg transition is for toggled state, not hover
- Introducing a `--ui-hover-transition` token — deferred until a real opt-in surface needs it

## Jira

[DST-1436](https://reservix.atlassian.net/browse/DST-1436)

## Test plan

- [ ] Chromatic VRT passes (will flag any unintended visual shift on the affected components)
- [ ] Eyeball pass in Storybook: Button (primary, secondary, ghost, destructive — should all feel identical on hover now), Tabs, Table row hover, Sidebar nav, Calendar date cells, SelectList items
- [ ] `pnpm typecheck:only` (already clean locally)

[DST-1436]: https://reservix.atlassian.net/browse/DST-1436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


OsamaAbdellateef (dismissed):
Test review body

themes/theme-rui/src/ui.css
```diff
@@ -48,7 +48,7 @@
  * sits on the page background and needs a clear visual boundary.
  */
 @utility ui-surface {
-  @apply rounded-surface relative transition-[background,border,box-shadow];
+  @apply rounded-surface relative transition-[border,box-shadow];
```
OsamaAbdellateef, line 54: ?? **Visual Regression Tests - not confirmed**

This PR changes the visual output of every component using `ui-surface` or `ui-surface-contrast` (Buttons, ActionBar, SelectList, ToggleButton, Menu, ListBox, and more). The Jira ticket [DST-1436](https://reservix.atlassian.net/browse/DST-1436) explicitly lists 'Visual regression via Chromatic' as part of the validation plan, but no Chromatic run appears in the CI checks.

Please trigger by commenting `/run-chromatic`, link the Chromatic build in the PR description, and confirm it passed before merging.

---

?? **No comment explaining the intentional `background` exclusion**

The ticket itself warns: 'without a deliberate stance documented, new components keep replicating the inconsistency.' A one-liner here prevents the next contributor from adding `background` back into this transition list.

themes/theme-rui/src/components/LegacyTable.styles.ts
```diff
@@ -52,7 +52,7 @@ export const LegacyTable: ThemeComponent<'LegacyTable'> = {
   body: cva({ base: '[&_tr:last-child]:border-0 bg-surface' }),
   row: cva({
     base: [
-      'border-b border-border transition-colors',
+      'border-b border-border transition-[color,border-color]',
```
OsamaAbdellateef, line 55: ?? **`transition-colors` ? `transition-[color,border-color]` makes row selection animation instant - outside the PR's stated hover scope**

Unlike `Table.row` (which has `data-hovered:ui-state-hover`), `LegacyTable.row` has **no hover background state at all**. The only `background-color` change on this row is `aria-selected:bg-muted` (selection state). `transition-colors` was animating that selection transition - switching to `transition-[color,border-color]` makes it snap instantly.

Selection is lower-frequency than hover; the RAIL model / '100+ times/day' rationale from the ticket does not apply to it. Since the row has no hover background, color state, or border-color state to animate, the cleanest fix is removing the transition class entirely:

`diff
- 'border-b border-border transition-[color,border-color]',
+ 'border-b border-border',
`"

If dropping the selection background animation is intentional, please add a comment noting it.

themes/theme-rui/src/components/Calendar.styles.ts
```diff
@@ -7,7 +7,7 @@ export const calendarCellBase = [
   'relative flex items-center justify-center whitespace-nowrap justify-self-center',
   'my-0.5',
   'border border-transparent p-0 text-sm font-normal text-foreground',
-  'outline-offset-2 duration-150 transition-[color,background-color]',
+  'outline-offset-2 duration-150 transition-[color]',
```
OsamaAbdellateef, line 10: ?? **`duration-150` now exclusively drives the `color` transition - was this deliberate?**

After removing `background-color`, `duration-150` applies only to the text-color transition on calendar cells, which are hover-heavy. A 150 ms color fade at that frequency can feel sluggish. Consider `duration-75` or removing `duration-150` entirely (Tailwind's default is also 150 ms, so dropping it keeps the same behavior). Not a blocker - just confirm it was a conscious choice.

OsamaAbdellateef (changes requested):
## Code Review - PR #5418

**Additional finding (file not in diff):** `themes/theme-rui/src/components/Breadcrumbs.styles.ts:25` uses `transition-colors hover:text-foreground`. `transition-colors` includes `background-color`, but this link never changes its background on hover. Should be `transition-[color]` for consistency with the no-background-transition stance this PR establishes.

---

themes/theme-rui/src/ui.css
sebald: Both points addressed:

1. **Comment added**: extended the block comments above `ui-surface` and `ui-surface-contrast` with a one-liner explaining that `background` is intentionally excluded (references DST-1436).
2. **Chromatic**: triggering `/vrt` after this push — will link the build once it's up.

themes/theme-rui/src/components/LegacyTable.styles.ts
sebald: Good catch — agreed this was out of stated scope. Went with a slight variant of your suggestion: kept the selection fade by switching to `transition-[background-color]` instead of dropping the transition entirely.

Rationale: selection feedback is low-frequency (same logic that excludes `Switch.track` and `FileField.dropZone` from this PR), so a brief fade on `aria-selected:bg-muted` is the appropriate behavior. Now only the selection bg animates — no leftover `color`/`border-color` animating nothing.

themes/theme-rui/src/components/Calendar.styles.ts
sebald: Verified — the premise doesn't quite hold here, so leaving as-is.

Traced what actually animates on hover:
- Cell base is `text-foreground`
- `data-hovered:ui-state-hover` expands to `bg-hover text-foreground` (defined at `ui.css:136-138`)

So on hover, text color stays `text-foreground` → `text-foreground`. No color transition fires on hover. The `transition-[color]` only runs on `selected:text-selected-bold-foreground` and `disabled:text-disabled`, neither of which is hover-frequency — 150ms there is the right amount of fade for "this date is now selected".

Happy to revisit if you spot a state I missed.

themes/theme-rui/src/components/LegacyTable.styles.ts
```diff
@@ -52,7 +52,7 @@ export const LegacyTable: ThemeComponent<'LegacyTable'> = {
   body: cva({ base: '[&_tr:last-child]:border-0 bg-surface' }),
   row: cva({
     base: [
-      'border-b border-border transition-colors',
+      'border-b border-border transition-[background-color]',
```
aromko: 🔴 **Critical — change is the inverse of the PR's stated goal**

The PR description says this should become `transition-[color,border-color]` (the safe expansion of `transition-colors` minus `background-color`). The actual change does the opposite: it keeps **only** `background-color` and drops `color` + `border-color` from the transition list.

This contradicts:
- The PR's own description ("avoids the Tailwind shorthand silently including bg")
- The [DST-1436](https://reservix.atlassian.net/browse/DST-1436) decision ("remove `background` from all hover-related transition lists")
- The "in family" consistency goal — every other touched component dropped bg; this one keeps only bg

```suggestion
      'border-b border-border transition-[color,border-color]',
```

---

aromko: @sebald not sure if it matters for the legacy one. But as long as we have it we can also change it there. If not ignore it.

sebald: Good catch — you're right, this was the inverse of what the PR description (and the rest of the diff) says it should be. Even though it's the legacy table, leaving `transition-[background-color]` would have shipped a regression (lost the color/border transitions that `transition-colors` was covering) and reintroduced the exact in-family inconsistency this PR is trying to fix.

Fixed in 38d85a6 — row now uses `transition-[color,border-color]` to match your suggestion and the PR description. Thanks!

OsamaAbdellateef (approved):

PR #5420, opened 2026-05-15
feat([DST-753]): SectionMessage announce prop, close button alignment, docs rewrite

sebald:
## Summary

Closes [DST-753](https://reservix.atlassian.net/browse/DST-753).

`<SectionMessage>` now announces its content to assistive technology through `@react-aria/live-announcer`'s singleton live region, replacing the previous (and spec-violating) `role="alert"` on the error variant.

- Adds `announce?: boolean` prop. Defaults to `true` for `variant="error"`, `false` otherwise.
- Polite priority for `info` / `success` / `warning`, assertive for `error`.
- Drops `role="alert"` from the wrapper. **Breaking** for tests/CSS selectors that target `[role="alert"]` on SectionMessage.
- Rewrites the docs Announcements section with two realistic demos (bulk-archive form + server-save error retry).
- Adds Storybook play tests covering each variant + announce combination.

## Why drop `role="alert"`?

The WAI-ARIA spec and MDN are explicit about two constraints:

1. The element with `role="alert"` must already exist in the DOM before content is injected.
2. It should be text-only, with no interactive children.

Marigold's previous implementation violated both: it conditionally mounted the alert together with its content, and the error variant can contain a close button plus action links inside `<SectionMessage.Content>`. Announcements were therefore unreliable across screen reader / browser combinations.

Routing through `@react-aria/live-announcer` sidesteps both problems. It maintains persistent polite and assertive `role="log"` regions at `document.body` and announces by appending a fresh text-only child. This is the same mechanism used across React Spectrum.

## Additional work bundled with this PR (beyond DST-753)

While doing the doc rewrite for the new prop, several adjacent issues surfaced and were fixed in the same branch:

- **Close button now matches the rest of the system.** SectionMessage's close button was visibly larger than every other close button in Marigold due to bespoke `size-8` + `[&_svg]:size-6` + negative-margin overrides in the theme. It now uses the shared `<CloseButton>` with no overrides. The `close` slot has been removed from the SectionMessage theme type (`@marigold/system` and `@marigold/theme-rui` bumped accordingly).
- **Dead component code.** Dropped a stale `useButton(props, buttonRef)` call that was applying div-level props to a button, plus its unused `buttonRef` and the `{...buttonProps}` spread. RAC's Button inside `CloseButton` already provides all the keyboard/press semantics.
- **Theme variant order.** Reordered variant slots so `info` (the default) is listed first across `container` / `content` / `icon`, matching the variant table in the docs.
- **Doc page overhaul.**
  - Anatomy rebuilt as inline SVG matching Card / Sidebar / SelectList; title and close button marked as optional, with content rules (no period in title, don't repeat title in body).
  - Two realistic announcement demos replace the old button-toggle.
  - Added a11y focus-management guidance for dynamic appearance and post-dismiss.
  - Added form-summary placement rule pairing SectionMessage with field-level validation.
  - Added action constraints (one primary action, verb+noun labels, descriptive link text).
  - Added two-line body rule with link-out overflow pattern.
  - Removed redundant Do/Don't tiles, merged Position into Usage, renamed subsections to Dismissal / Actions / Announcements.
- **Drive-bys.** Fixed "Accessililty" typo in the feedback-messages pattern doc.

## Migration

- Code relying on `getByRole('alert')` or `[role="alert"]` selectors against `SectionMessage` needs updating. The visible message DOM is unchanged; only the wrapper role is gone.
- Consumers who previously wrapped a dynamic `<SectionMessage>` in their own `<div role="status">` can replace that wrapper with `<SectionMessage announce>`.
- Custom themes that defined a `SectionMessage.close` slot will now get a type error. Remove the slot. Close button styling now comes entirely from the `CloseButton` theme.
- The SectionMessage's close button is visually smaller after this release (matches every other close button in Marigold). If you previously relied on the larger size, that was an inconsistency, not a feature.
- To re-announce the same message (e.g. user clicks "Retry" twice), pass a changing `key` to remount the component.

## Test plan

- [x] `pnpm test:unit SectionMessage` passes (11 tests, including 4 new announcer-specific tests; dropped the `role="alert"` assertion).
- [x] `pnpm test:sb SectionMessage` passes (4 new play-function tests covering assertive announces, polite announces, no-announce by default for non-error, opt-out for error).
- [x] Manual verification in the docs site at `/components/content/section-message#announcements`. Both demos fire announcements in the expected `[role="log"][aria-live="polite|assertive"]` region.
- [x] Manual verification of the new close-button look across the SectionMessage variants in Storybook.

[DST-753]: https://reservix.atlassian.net/browse/DST-753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


themes/theme-rui/src/components/SectionMessage.styles.ts
```diff
@@ -38,21 +38,14 @@ export const SectionMessage: ThemeComponent<'SectionMessage'> = {
...
     },
   }),
-  close: cva({
-    base: [
-      'size-8 text-foreground',
-      '[&_svg]:size-6',
-      '-my-1.5 -me-2', // align button with title
-    ],
-  }),
 };
```
aromko, line 51: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files (this theme file, the component source, stories) — most notably the close-button restyle (this file drops the `size-8` / `[&_svg]:size-6` / negative-margin overrides) and the variant order reorder. The `Visual-Regression-Tests` workflow has not run on this branch.

packages/components/src/SectionMessage/SectionMessage.tsx
```diff
@@ -67,7 +77,14 @@ export const SectionMessage = ({
...
   const isCurrentlyVisible = close ?? internalVisible;
 
-  const { buttonProps } = useButton(props, buttonRef);
+  // Announce on mount. Remount with a changing `key` to re-announce.
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => {
+    const shouldAnnounce = announceProp ?? variant === 'error';
+    if (!shouldAnnounce || !isCurrentlyVisible) return;
+    const text = containerRef.current?.textContent?.trim();
+    if (text) announce(text, variant === 'error' ? 'assertive' : 'polite');
```
aromko: ⚠️ The empty deps + suppressed exhaustive-deps lint masks a controlled-visibility edge case.

If a consumer renders with `close={true}` (hidden) and later flips to `close={false}` to show, the effect already ran during the hidden render and bailed on `!isCurrentlyVisible`. With empty deps it won't run again, so the announcement silently never fires.

Two options:
- Include `isCurrentlyVisible` in deps (so the announcement fires when the message first becomes visible).
- Or, document this in the `announce` JSDoc and tell consumers to remount via `key` when toggling controlled visibility.

The PR description covers the re-announce-same-message case via `key`, but this first-appearance case in the controlled API is silent today.

packages/components/src/SectionMessage/SectionMessage.tsx
```diff
@@ -67,7 +77,14 @@ export const SectionMessage = ({
   const [internalVisible, setInternalVisible] = useState(true);
   const isCurrentlyVisible = close ?? internalVisible;
 
-  const { buttonProps } = useButton(props, buttonRef);
+  // Announce on mount. Remount with a changing `key` to re-announce.
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => {
+    const shouldAnnounce = announceProp ?? variant === 'error';
+    if (!shouldAnnounce || !isCurrentlyVisible) return;
+    const text = containerRef.current?.textContent?.trim();
```
aromko, line 86: 💡 Minor readability — `variant === 'error'` is evaluated twice in the effect. Extract to a `priority` constant:

```suggestion
    const shouldAnnounce = announceProp ?? variant === 'error';
    if (!shouldAnnounce || !isCurrentlyVisible) return;
    const text = containerRef.current?.textContent?.trim();
    const priority = variant === 'error' ? 'assertive' : 'polite';
    if (text) announce(text, priority);
```

themes/theme-rui/src/components/SectionMessage.styles.ts
sebald: Visual Regression Tests (Chromatic) only run on PRs targeting `main`, not `beta-release`. The VRT workflow is intentionally scoped to `main` so the baseline stays stable while we stage beta work — running it from `beta-release` would compare against an inconsistent baseline. Resolving this thread; VRT will run automatically once the changes land on a `main`-targeted PR.

packages/components/src/SectionMessage/SectionMessage.tsx
sebald: Good catch — fixed in e592cbcc5.

Changed the effect deps from `[]` to `[isCurrentlyVisible]` so each hidden→visible transition (controlled or uncontrolled) fires the announcer. Mount-while-visible still announces once; mount-while-hidden announces on first show; re-announcing identical content without a visibility flip still needs a `key` remount.

Updated the `announce` JSDoc to say "on mount and whenever controlled visibility transitions from hidden to visible," and added a `AnnouncesWhenControlledShown` play test that renders the message hidden via `close={false}`, asserts the polite log is empty, clicks a "Show" button, and asserts the announcement lands.

packages/components/src/SectionMessage/SectionMessage.tsx
sebald: Applied in e592cbcc5 — `priority` is now extracted as a const, `variant === 'error'` is evaluated once. Folded into the same effect refactor as the controlled-visibility fix above.

aromko (approved):

PR #5443, opened 2026-05-18
feat([DST-1407]): Drawer enforces one open at a time

sebald:
## Summary

Opening a sibling `<Drawer>` while one is already open now dismisses the first. A `<Drawer.Trigger>` nested *inside* an open Drawer instead opens on top of its parent — see "Nested case" below. Closes [DST-1407](https://reservix.atlassian.net/browse/DST-1407).

Applies to both the desktop (`NonModal`) and mobile (`ModalOverlay`) paths. The dismissed Drawer's `onOpenChange(false)` fires through the normal `useControlledState` path so controlled-state consumers stay in sync.

## Implementation

Three internal pieces:

- **`drawerRegistry.ts`** — single-slot module-level singleton holding the currently active Drawer's close handler. Registering a new close invokes and replaces the previous one. Pattern follows react-aria's `ToastQueue` and `LandmarkManager`.

- **`useDrawerCoordination.ts`** — hook that subscribes the current Drawer to the registry when it opens. Reads `OverlayTriggerStateContext` (provided by `NonModal` on desktop and `ModalOverlay`/`DialogTrigger` on mobile) and `DrawerNestingContext`. Uses `useEffectEvent` from `@react-aria/utils` so the registered handler stays stable across renders while always invoking the latest `state.close` — robust against unstable `onOpenChange` references.

- **`Drawer.tsx`** — calls `useDrawerCoordination()` directly and wraps its children in `<DrawerNestingContext value={true}>`. The hook runs at the `Drawer` level (not in a child) so it reads the *ancestor's* nesting value, not the provider this Drawer publishes for its descendants.

### Nested case

A `<Drawer.Trigger>` rendered inside another Drawer's content can't dismiss its ancestor without taking itself down in the unmount cascade — closing the parent unmounts the nested trigger and the new Drawer with it. The hook detects this via `DrawerNestingContext` and skips registration; the nested Drawer opens over the parent and "one at a time" continues to apply between siblings only.

### Why module-level (vs Context)

Drawer is client-only (no SSR concern) and conceptually has one scope per app instance. A `<DrawerProvider>` would require consumers to wrap their app and adds no value over a module-level slot. Same call-site shape as react-aria's own `ToastQueue`.

### Why imperative (vs `useSyncExternalStore`)

Drawers don't need to *re-render* based on "am I active?" — they just need to receive a one-shot close call. `useSyncExternalStore` is the right tool when components must re-render on external state changes; here, imperative invocation is the correct shape.

### Concurrent React safety

`activeClose` is never read during render — only mutated inside effect callbacks. No tearing risk. Under `startTransition`, a deferred open registers only when the transition commits; if interrupted, the prior Drawer stays open. Strict-Mode double-invoke is safe: re-registering the same handler is a no-op and the matching cleanup only clears the slot if it still holds that handler.

## Behaviour notes

- Same-instance idempotency: re-opening the currently active Drawer is a no-op (no flicker).
- Animations are concurrent: A's exit and B's enter run in parallel, no sequential close-then-open delay.
- Nested Drawers stack over their parent; closing the nested one returns to the parent unchanged.
- No new public API surface.
- Silent close (no console warning) — matches the contract described in DST-1407.

## Test plan

- [x] **Unit tests** (`Drawer.test.tsx`): sibling-drawer dismissal, ESC scope, controlled-state preemption, mobile path. All passing.
- [x] **Storybook component-tests** (`OneAtATime`, `OneAtATimeNested`, `tags: ['component-test']`): run in real Firefox via Playwright. Sibling story asserts A closes when B opens; nested story asserts A stays mounted under B and Escape closes only B.
- [x] `pnpm typecheck:only` clean.
- [x] `pnpm lint` clean for new files.
- [x] Manual verification in Storybook desktop + mobile breakpoints, sibling and nested cases.

## Files

- `packages/components/src/Drawer/drawerRegistry.ts` (new)
- `packages/components/src/Drawer/useDrawerCoordination.ts` (new)
- `packages/components/src/Drawer/Context.tsx` (add `DrawerNestingContext`)
- `packages/components/src/Drawer/Drawer.tsx` (call hook, publish nesting context)
- `packages/components/src/Drawer/Drawer.stories.tsx` (add `OneAtATime`, `OneAtATimeNested`)
- `packages/components/src/Drawer/Drawer.test.tsx` (story-based tests)
- `.changeset/drawer-one-at-a-time.md` (major)

[DST-1407]: https://reservix.atlassian.net/browse/DST-1407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/Drawer/Drawer.test.tsx
```diff
@@ -112,3 +112,32 @@ test('uses modal on small screens', async () => {
   expect(screen.getByRole('dialog')).toBeInTheDocument();
   expect(screen.queryByRole('complementary')).not.toBeInTheDocument();
 });
+
+test('opening a second drawer closes the first one', async () => {
```
jim761: This test is already in the stories test, so its not necessary here

packages/components/src/Drawer/Drawer.test.tsx
```diff
@@ -112,3 +112,32 @@ test('uses modal on small screens', async () => {
...
+  expect(await screen.findByText('Title A')).toBeInTheDocument();
+
+  await user.click(screen.getByRole('button', { name: 'Open B' }));
+  expect(await screen.findByText('Title B')).toBeInTheDocument();
+  await waitFor(() =>
+    expect(screen.queryByText('Title A')).not.toBeInTheDocument()
+  );
+});
+
+test('escape closes only the visible drawer', async () => {
```
jim761: also the same as in stories test

jim761 (dismissed):

packages/components/src/Drawer/Drawer.stories.tsx
```diff
@@ -242,3 +242,44 @@ export const Controlled = meta.story({
...
+      expect(canvas.queryByText('Title A')).not.toBeInTheDocument()
+    );
+
+    // ESC closes only the visible drawer.
+    await userEvent.keyboard('{Escape}');
+    await waitFor(() =>
+      expect(canvas.queryByText('Title B')).not.toBeInTheDocument()
+    );
+  },
+});
```
jim761, line 285: ⚠️ **Missing test for the primary motivating case from DST-1407.** The ticket calls out the "icky failure mode" as a `<DrawerTrigger>` inside an open drawer opening a second drawer. This story tests *sibling* triggers; the AC explicitly asks for a story where "a trigger inside drawer A opens drawer B → A closes as B opens." Recommend adding a second story (or extending this one) where B's trigger lives inside A's `Drawer.Content`.

jim761 (commented):
## Automated Code Review

Nice implementation — clean, small public-API surface, and the concurrent-React analysis in `drawerRegistry.ts` is thorough. Findings below are about test coverage of DST-1407's acceptance criteria; nothing blocking.

---

packages/components/src/Drawer/Drawer.test.tsx
jim761: ⚠️ **No controlled-state sync assertion.** The PR description states `onOpenChange(false)` fires on auto-dismiss, and DST-1407 lists "controlled state sync" as a required test. The current test uses uncontrolled `Drawer.Trigger`s and only asserts DOM disappearance. Add a test with two controlled drawers asserting both `onOpenChange` mocks fire when B preempts A.

packages/components/src/Drawer/Drawer.test.tsx
```diff
@@ -112,3 +112,32 @@ test('uses modal on small screens', async () => {
...
+
+  await user.click(screen.getByRole('button', { name: 'Open A' }));
+  expect(await screen.findByText('Title A')).toBeInTheDocument();
+
+  await user.click(screen.getByRole('button', { name: 'Open B' }));
+  expect(await screen.findByText('Title B')).toBeInTheDocument();
+  await waitFor(() =>
+    expect(screen.queryByText('Title A')).not.toBeInTheDocument()
+  );
+});
```
jim761, line 176: ⚠️ **No mobile-path coordination test.** Coordination is claimed to work on both `NonModal` (desktop) and `ModalOverlay` (mobile) paths, but no test exercises *coordination* on the mobile path. Add a `mockMatchMedia([smallScreenQuery])` variant of this test.

packages/components/src/Drawer/Drawer.test.tsx
```diff
@@ -112,3 +112,32 @@ test('uses modal on small screens', async () => {
...
+
+  await user.click(screen.getByRole('button', { name: 'Open B' }));
+  expect(await screen.findByText('Title B')).toBeInTheDocument();
+
+  await user.keyboard('{Escape}');
+  await waitFor(() =>
+    expect(screen.queryByText('Title B')).not.toBeInTheDocument()
+  );
+  expect(screen.queryByText('Title A')).not.toBeInTheDocument();
+});
```
jim761, line 200: 💡 **Same-instance idempotency test.** Behaviour notes mention re-opening the currently active drawer is a no-op. A small test toggling the same trigger twice would protect the `if (activeClose !== close)` guard from regression.

packages/components/src/Drawer/drawerRegistry.ts
```diff
@@ -0,0 +1,38 @@
...
+ * - `activeClose` is never read during render — only mutated inside effect
+ *   callbacks. There is no tearing risk and `useSyncExternalStore` is not
+ *   needed.
+ * - Under `startTransition`, a deferred drawer-open won't register until the
+ *   transition commits; if interrupted, the prior drawer stays open, which is
+ *   the desired outcome.
+ * - Strict Mode double-invoke is safe: registering the same `close` twice is
+ *   a no-op, and the matching cleanup only clears the slot if it still holds
+ *   that handler.
+ */
```
jim761, line 13: 💡 **Heavy module doc.** This is on the edge of CLAUDE.md's "no comments unless WHY is non-obvious" rule. The concurrent-React notes are defensible, but consider trimming to a 3–4 line summary and keeping deeper rationale in the PR/Jira.

packages/components/src/Drawer/Drawer.tsx
```diff
@@ -102,6 +103,7 @@ export const Drawer = ({
             classNames.container
           )}
         >
+          <DrawerCoordinator />
```
jim761: 💡 **One-line context comment.** Consider `// enforces one-Drawer-at-a-time` so future readers don't have to grep for why this renderless element is here.

sebald:
Addressed all review feedback in 2ac0584dd. Summary:

**Fully integrated (7 of 8):**
- ✅ Removed both duplicated unit tests (covered by the `OneAtATime` story play function).
- ✅ Added `OneAtATimeNested` story — DST-1407's primary motivating AC where B's trigger lives inside A's `Drawer.Content`.
- ✅ Added `'controlled drawers receive onOpenChange(false) when preempted'` test — verifies the controlled-state sync claim from the PR description with two controlled drawers + `vi.fn()` mocks.
- ✅ Added `'coordination also applies on small screens (mobile path)'` test — exercises the `ModalOverlay` path via `mockMatchMedia([smallScreenQuery])`.
- ✅ Added `'registry no-ops on same-handler re-registration'` — direct unit test of the `if (activeClose !== close)` guard against regression.
- ✅ Trimmed `drawerRegistry.ts` module doc — dropped the `ToastQueue`/`LandmarkManager` analogy, kept the substantive concurrent-React safety notes (Strict Mode, `startTransition`, `useSyncExternalStore` rationale).

**Declined (1 of 8):**
- ⏭️ One-line comment at `<DrawerCoordinator />` — `DrawerCoordinator.tsx` already has a JSDoc covering its purpose; clicking into the component gives the answer immediately. Happy to add it if you'd prefer.

**Test results:**
- `pnpm test:unit Drawer.test.tsx` → 12 passed (was 11; net +1 after −2 dup, +3 new)
- `pnpm test:sb Drawer.stories.tsx` → 3 passed (`LongContent`, `OneAtATime`, `OneAtATimeNested`)
- `pnpm typecheck:only`, `pnpm lint` → clean

packages/components/src/Drawer/Drawer.tsx
sebald: Skipping this one — `DrawerCoordinator.tsx` already has a JSDoc explaining its purpose, so clicking into the component gives the answer immediately. Resolving, but happy to revisit if you prefer the inline pointer.

jim761 (approved):

PR #5449, opened 2026-05-19
fix(DST-1447): honor prefers-reduced-motion to stop VRT focus flicker

sarahgm:
# Description

Chromatic visual tests keep flagging 1 to 3 stories on Autocomplete, Select, ComboBox, and other components with focus state diffs on every run, even on unrelated PRs (Renovate updates, etc.). This PR fixes the root cause. Refs [DST-1447](https://reservix.atlassian.net/browse/DST-1447).

## What was happening

Many parts of the theme have **CSS transitions** on focus and hover states. For example, the focus border on a Select trigger fades in over 150ms when you click it.

Chromatic uses a setting called `pauseAnimationAtEnd` that pauses CSS animations before it takes a screenshot. The catch: **CSS transitions are not the same as CSS animations**, and that setting does not pause them. So when a story's play function ended with focus on a button or input, Chromatic could take the screenshot while the transition was still happening: sometimes before, sometimes during, sometimes after. The captured pixels changed run by run.

## What changed

One rule in the theme: when the user (or Chromatic) asks for reduced motion, every CSS transition runs in 0 seconds.

```css
@media (prefers-reduced-motion: reduce) {
  *,
  *::before,
  *::after {
    transition-duration: 0s !important;
    transition-delay: 0s !important;
  }
}
```

This single rule replaces what would otherwise be dozens of per-component opt-outs. It covers every transition that exists in the theme today and every transition anyone adds later.

## Why prefers-reduced-motion is the right lever

- Chromatic already turns on `prefers-reduced-motion: reduce` for every snapshot it takes (configured in `.storybook/preview.tsx:28`). The theme just needs to honor that preference.
- It is also an accessibility improvement. People with vestibular issues or motion sensitivity set this preference on their device, and until now the theme ignored it for these transitions. WCAG 2.1 SC 2.3.3 recommends honoring it.
- Real users without the preference see no change. The 150ms transitions still play exactly as before.

## CSS animations are not affected

The rule only zeroes out CSS transitions. CSS animations (`@keyframes`) keep working, so things that should still feel like motion are untouched:

- Sidebar slide in and out
- Drawer slide in and out
- ActionBar entrance
- Spinners

If a specific animation should also stop under reduced motion, it can be opted out per component with `motion-reduce:animate-none`, the way the Sidebar already does.

## Bonus fix: Select.WithImages avatar race

The `Select.WithImages` story loads avatars from `i.pravatar.cc`. Its play function did not wait for those images to finish loading, so Chromatic could capture the snapshot before the avatars appeared. The play function now waits on `img.complete` for every avatar in the listbox before the snapshot is taken.

This is a separate root cause from the transition issue (network timing, not CSS), but it was showing up as a flicker on the same PR, so it is included here.

# Test Instructions:

**Local manual check (about a minute):**

1. `pnpm sb`
2. Open any Select / Autocomplete / ComboBox / Tray story.
3. Chrome DevTools: `Cmd+Shift+P` -> "Show Rendering" -> "Emulate CSS media feature `prefers-reduced-motion`" -> set to **reduce**.
4. Click the trigger. The focus border should snap in instantly with no fade.
5. Turn emulation off. The 150ms transition should still play.

**Real success signal in Chromatic:**

- [ ] First Chromatic run on this PR rebaselines once (focus state is now deterministic). Accept the affected snapshots and merge.
- [ ] Re-run VRT on the same merged commit two or three times: snapshots should be byte-identical.
- [ ] Watch the next few unrelated Renovate PRs: zero spurious diffs on Autocomplete, Select, ComboBox, Tray.

**Other checks already green locally:**

- [x] `pnpm typecheck:only`
- [x] `pnpm lint` (0 errors)
- [x] `pnpm test:sb` (130 passed)
- [x] `pnpm build:themes` (compiled CSS contains the global reduced-motion block at the `[data-theme="rui"]` scope)

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

[DST-1447]: https://reservix.atlassian.net/browse/DST-1447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/themes/theme-rui/src/theme.css
+++ b/themes/theme-rui/src/theme.css
@@ -5,6 +5,25 @@
 @import './variants.css';
 
 /*
+ * Honor the user's reduced-motion preference globally: disable every CSS
+ * transition. CSS @keyframes animations are left alone so intentional motion
+ * (entering:animate-*, exiting:animate-*, spinners) keeps working; components
+ * that want to disable specific animations can still opt in with
+ * motion-reduce:animate-none.
+ *
+ * This also stabilizes Chromatic VRT snapshots, since Chromatic enables
+ * prefers-reduced-motion: reduce globally during capture.
+ */
+@media (prefers-reduced-motion: reduce) {
+  *,
+  *::before,
+  *::after {
+    transition-duration: 0s !important;
+    transition-delay: 0s !important;
+  }
+}
+
+/*
     - create tokens if we think:
         - it is necessary for a component
         - it can be used by other teams (if they build their own component)
--- a/themes/theme-rui/src/theme.css
+++ b/themes/theme-rui/src/theme.css
@@ -5,14 +5,9 @@
 @import './variants.css';
 
 /*
- * Honor the user's reduced-motion preference globally: disable every CSS
- * transition. CSS @keyframes animations are left alone so intentional motion
- * (entering:animate-*, exiting:animate-*, spinners) keeps working; components
- * that want to disable specific animations can still opt in with
- * motion-reduce:animate-none.
- *
- * This also stabilizes Chromatic VRT snapshots, since Chromatic enables
- * prefers-reduced-motion: reduce globally during capture.
+ * Keep @keyframes animations intact so entering:/exiting:/spinner motion still
+ * works; components can opt out per-animation via motion-reduce:animate-none.
+ * Also stabilizes Chromatic VRT, which forces prefers-reduced-motion: reduce.
  */
 @media (prefers-reduced-motion: reduce) {
   *,
--- a/.changeset/dst-1447-prefers-reduced-motion.md
+++ b/.changeset/dst-1447-prefers-reduced-motion.md
@@ -1,0 +1,5 @@
+---
+'@marigold/theme-rui': patch
+---
+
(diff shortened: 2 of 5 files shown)


sarahgm:
Accepted Tray Chromatic change, the focus should be visible after escaping with Keyboard

jim761 (approved):

themes/theme-rui/src/theme.css
```diff
@@ -4,6 +4,20 @@
...
+ * Keep @keyframes animations intact so entering:/exiting:/spinner motion still
+ * works; components can opt out per-animation via motion-reduce:animate-none.
+ * Also stabilizes Chromatic VRT, which forces prefers-reduced-motion: reduce.
+ */
+@media (prefers-reduced-motion: reduce) {
+  *,
+  *::before,
+  *::after {
+    transition-duration: 0s !important;
+    transition-delay: 0s !important;
```
OsamaAbdellateef, line 17: `prefers-reduced-motion` also should close scroll-behaviour , e.g 
`scroll-behavior: auto !important;`

OsamaAbdellateef:
Do we need to do the same for `Combobox` and `Autocomplete` ? @sarahgm

sarahgm:
> Do we need to do the same for `Combobox` and `Autocomplete` ? @sarahgm

its globally, so it affects already Combobox and Autocomplete

themes/theme-rui/src/theme.css
sarahgm: Thanks @OsamaAbdellateef, good a11y point. I'd like to keep this PR scoped to the VRT focus-flicker root cause (transition timing) and ship the scroll-behavior change separately so it gets its own changeset and test plan. I'll open a follow-up ticket for it.

PR #5451, opened 2026-05-19
fix(DSTSUP-253): make `cva` results assignable to `ComponentStyleFunction` under strict TS

sebald:
# Description

`@marigold/system@17.5.0` shipped `cva@1.0.0-beta.4`, whose narrowed return type stopped satisfying `ComponentStyleFunction<Variants, Sizes>` once consumers enabled `strictFunctionTypes` (implied by `strict: true`). Function-parameter contravariance rejected both the dead `variant?: Variants | null` slot and the wider `string` vs literal-union variant, so the documented `cva(...) as ThemeComponent<'Foo'>` pattern failed to compile in every external project. The Marigold monorepo hid the bug because `config/tsconfig/base.json` sets `strictFunctionTypes: false`.

`ComponentStyleFunction` now types `variant`/`size` as `any` — honestly reflecting the runtime contract (`useClassNames` passes `string | undefined`, `cva` falls back to base styles for unknown values) and letting narrower `cva` returns satisfy the wider theme contract. The `Variants`/`Sizes`/`Additional` generics are retained for backward source-compatibility with existing annotations. No runtime change.

Also adds a type-level regression test colocated with the source it exercises:
- `packages/system/src/types/theme.test-d.ts` mirrors the canonical theming pattern from the ticket.
- A `StrictlyAssignable<Source, Target>` conditional-type helper verifies parameter-contravariance / return-covariance structurally, so the regression is caught by the regular `pnpm typecheck` regardless of the monorepo's `strictFunctionTypes: false` base config — no separate tsconfig or CI step required.

No UI change — Screenshots/VRT not applicable.

Closes DSTSUP-253

## Test Instructions

1. `pnpm install`
2. `pnpm typecheck:only` — passes (regression test included).
3. (Optional) Temporarily revert the change to `packages/system/src/types/theme.ts` and re-run `pnpm typecheck:only` — it should fail at `packages/system/src/types/theme.test-d.ts` with `Type 'false' does not satisfy the constraint 'true'`, matching the regression in DSTSUP-253.
4. `pnpm --filter @marigold/system test` — 205/205 pass.
5. `pnpm --filter @marigold/system build` and `pnpm --filter @marigold/theme-rui build` — clean.

## Breaking Changes

No

## Checklist


sarahgm (commented):
## Automated Code Review

The technical fix itself is correct and CI is green — but the PR went through 4 commits with two pivots, and **the changeset and PR body still describe the abandoned approach**. Two materially inaccurate consumer-facing surfaces should be fixed before merge.

### Commit history vs final state

| Commit | What it did |
|--------|-------------|
| 1 | `bivariantHack` method-shorthand + `tsconfig.types-test.json` + `pnpm typecheck:strict` script + CI step |
| 2 | refactor: replace bivariance hack with `any`-typed `variant`/`size` |
| 3 | refactor: fold strict check into regular typecheck via `StrictlyAssignable<S, T>` conditional-type helper — **removes** the separate tsconfig, script, and CI step |
| 4 | colocate test file → `src/types/theme.test-d.ts` |

Final state (diff vs `main`):
- `variant: any` / `size: any` in `ComponentStyleFunction`
- One new file: `packages/system/src/types/theme.test-d.ts` using `StrictlyAssignable<Source, Target>`
- **No** new tsconfig, **no** new script, **no** new CI step

### 🔴 Changeset is materially inaccurate (blocker)

`.changeset/dstsup-253-cva-strict-function-types.md` ends with:

> A new `pnpm typecheck:strict` script compiles a focused regression test under `strictFunctionTypes: true`, so the monorepo's loose base config no longer hides this class of bug.

There is no `pnpm typecheck:strict` script in the merged diff (commit 3 removed it). This changeset gets published in `CHANGELOG.md` and `@marigold/system@17.5.1` release notes — it would ship a false claim to downstream consumers.

Suggested replacement for the final sentence:

> A dedicated type-level regression test (`packages/system/src/types/theme.test-d.ts`) uses a `StrictlyAssignable<Source, Target>` conditional-type helper to verify the contract under TypeScript's structural function checking, so this class of bug is caught by the regular `pnpm typecheck` even though the monorepo's base config keeps `strictFunctionTypes: false`.

### 🟡 PR body / Test Instructions describe the abandoned approach

> This switches `ComponentStyleFunction` to a method-shorthand call signature (`{ bivariantHack(...): string }['bivariantHack']`)…

That was commit 1's approach. Commit 2 replaced it with `variant: any` / `size: any`.

> - `packages/system/src/__type-tests__/themeContract.test-d.ts` mirrors the canonical theming pattern from the ticket.
> - `packages/system/tsconfig.types-test.json` compiles it with `strictFunctionTypes: true`.
> - `pnpm typecheck:strict` (wired into `.github/workflows/typecheck.yml`) makes the loose base config no longer hide similar regressions.

Of those four references, only the test file survives (and it was renamed and moved to `src/types/theme.test-d.ts` in commit 4). The tsconfig, script, and workflow change were all removed in commit 3.

Test Instructions step 3 (`pnpm typecheck:strict — strict regression check passes.`) is not runnable — the script doesn't exist. Step 4 also references it.

### Things that are fine

- The in-file documentation is **correct** for the final state — the JSDoc on `ComponentStyleFunction` in `theme.ts` accurately explains the `any` choice and why, and the header of `theme.test-d.ts` explicitly says _"this file is picked up by the regular `pnpm typecheck` — no separate tsconfig or CI step required."_
- `any` is a defensible pragmatic choice given the runtime contract (`useClassNames` already passes `string | undefined`, `cva` falls back to base styles for unknown values). A narrower `string | undefined` would still fail contravariance vs `cva`'s literal-union returns.
- The `StrictlyAssignable` conditional-type trick is genuinely nice — catches the regression structurally regardless of `strictFunctionTypes`, so no extra CI surface needed.
- Generics retained for backward source-compatibility with appropriate `eslint-disable` comments and JSDoc note.
- Correctly self-identified as type-only — VRT not applicable.
- All CI checks pass.

### Recommendation

The code change is good and ready. Please **rewrite the changeset body** to remove the reference to the nonexistent `pnpm typecheck:strict` script (it will otherwise ship into the public release notes), and update the PR description / Test Instructions to match the final implementation.

---

sebald:
@sarahgm thanks for the careful review! Both items are now integrated:

- **Changeset** (`77e16f485`): rewrote the final sentence to describe the `StrictlyAssignable<Source, Target>` conditional-type helper in `packages/system/src/types/theme.test-d.ts` instead of the nonexistent `pnpm typecheck:strict` script — used your suggested wording almost verbatim.
- **PR description**: updated the technical description and Test Instructions to reflect the final implementation (`variant`/`size: any` + the colocated `theme.test-d.ts`), and dropped the references to the bivariance hack, the separate tsconfig, the extra script, and the CI step that were all removed in commits 2/3.

sarahgm (approved):

PR #5452, opened 2026-05-19
fix([DST-1363]): bridge <TagGroup> errorMessage to validation

sebald:
## Summary

- `<TagGroup>` accepted `errorMessage` via `<FieldBase>` but never rendered it because RAC's `TagGroup` doesn't populate `FieldErrorContext`. Bridges the validation lifecycle the same way `<SelectList>` does: `useFormValidationState` + `useFormValidation` against a hidden `<select>`, plus `FormContext` inheritance for `validationBehavior`.
- Extracts the hidden form control into a shared `HiddenSelection/` component used by both `<SelectList>` and `<TagGroup>`. Replaces the previous `<input type=\"checkbox\">` shim.
- Normalises the public API to Marigold conventions: `error`/`required`/`disabled` (was `isInvalid`/`isRequired`/`isDisabled`), `onChange` (was `onSelectionChange`), `selectionMode` defaults to `'multiple'`. Adds `validate`, `validationBehavior`, and `form` props.

## Test plan

- [x] `pnpm typecheck:only`
- [x] `pnpm vitest run packages/components/src/HiddenSelection packages/components/src/TagGroup packages/components/src/SelectList` — 70/70 pass
- [x] `pnpm exec eslint packages/components/src/HiddenSelection/ packages/components/src/TagGroup/TagGroup.tsx` clean
- [ ] Verify `errorMessage` renders in Storybook (`Tag > WithError`, `Tag > Required` after submit)
- [ ] Native form submission emits `categories=<key>` entries (`Tag > WithForm`)
- [ ] `<SelectList>` regression check (story tests)

[DST-1363](https://reservix.atlassian.net/browse/DST-1363)

[DST-1363]: https://reservix.atlassian.net/browse/DST-1363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/TagGroup/TagGroup.tsx
```diff
@@ -42,41 +89,157 @@ export interface TagGroupProps
...
+        description={description}
+        errorMessage={errorMessage}
+        isInvalid={validationState.displayValidation.isInvalid}
+        isRequired={required}
+        isDisabled={disabled}
+      >
+        <RACTagGroup
+          {...(rest as RAC.TagGroupProps)}
+          aria-labelledby={label ? labelId : rest['aria-labelledby']}
+          aria-disabled={disabled || undefined}
```
sarahgm, line 210: 🔴 **`disabled` prop is only partially honored.**

RAC's `<TagGroup>` doesn't accept `isDisabled` — it expects per-tag `disabledKeys`. Passing only `aria-disabled` here gives screen readers the right hint, but tags remain clickable and selection still goes through. Meanwhile the hidden `<select>` *is* disabled (line 240), so the visible UI and form behaviour disagree.

Either disable interaction for real (e.g. derive `disabledKeys` to cover every tag) or document `disabled` as 'form-disabled only'.

packages/components/src/TagGroup/TagGroup.tsx
```diff
@@ -42,41 +89,157 @@ export interface TagGroupProps
...
-      <div className={classNames.container}>
-        <TagList
-          items={items}
-          className={classNames.listItems}
-          renderEmptyState={emptyState}
+    <Provider
+      values={[
+        [LabelContext, { id: labelId, elementType: 'span' }],
+        [FieldErrorContext, validationState.displayValidation],
+        [FormContext, { validationBehavior }],
```
sarahgm: ⚠️ **`FormContext` is replaced, not merged.**

`FormContext` carries more than `validationBehavior` (`action`, `method`, `encType`, `onSubmit`, `onReset`, `onInvalid`, …). Passing only `{ validationBehavior }` erases the rest for descendants inside `<RACTagGroup>`. Today nothing inside reads those fields, but the safer pattern is:

```suggestion
        [FormContext, { ...formCtx, validationBehavior }],
```

packages/components/src/TagGroup/TagGroup.tsx
```diff
@@ -42,41 +89,157 @@ export interface TagGroupProps
...
+        </RACTagGroup>
+        {selectionMode !== 'none' ? (
+          <HiddenSelection
+            name={name}
+            form={form}
+            disabled={disabled}
+            required={required}
+            selectionMode={selectionMode === 'multiple' ? 'multiple' : 'single'}
+            selection={selection}
+            onSelectionChange={setSelection}
```
sarahgm: ⚠️ **`HiddenSelection`'s `onSelectionChange` doesn't commit validation.**

The `RACTagGroup` handler does `setSelection(keys); validationState.commitValidation()`, but this path (form reset / browser autofill) skips the commit. An autofill that satisfies a `required` rule would update state but leave the displayed error in place until the next user interaction.

```suggestion
            onSelectionChange={keys => {
              setSelection(keys);
              validationState.commitValidation();
            }}
```

sarahgm (commented):
## Automated Code Review

## PR Review: #5452 — fix TagGroup error message

### Overview
| Field | Value |
|-------|-------|
| Branch | `fix/DST-1363-taggroup-field-error` → `beta-release` |
| Files Changed | 9 (+521 / −93) |
| CI | All green |
| Visual Regression | Not required (target is `beta-release`) |

### The ticket: DST-1363
**🐛 TagGroup: errorMessage / FieldError never renders**

The bug: `<TagGroup>` accepted an `errorMessage` prop but never showed it. The reason is that React Aria's `TagGroup` doesn't set up the `FieldErrorContext` that Marigold's `HelpText` looks for. So the error message path was a silent no-op.

The ticket recommended two things:
1. **Option B** — fix it properly by wiring `useFormValidationState` + `useFormValidation` to a hidden form control.
2. **Follow-up** — extract a shared `<HiddenSelection>` component, since `SelectList` will need the same plumbing.

This PR does **both**.

### What's good
- The bug is actually fixed. `errorMessage` now shows up.
- `<TagGroup>` now works like a real form field: `required` blocks submit, `validate` runs, and `<Form validationErrors>` reaches the user.
- The hidden form control is now a shared `<HiddenSelection>` used by both `SelectList` and `TagGroup`. Future fixes happen in one place.
- The public API now matches Marigold's style: `disabled`, `required`, `error`, `onChange` instead of the React Aria names.
- Good test coverage: 9 new unit tests for `HiddenSelection`, plus new story tests (`WithForm`, `Required`) that run in a real browser.
- Changeset is there and the description is clear.

### Things I'd change

#### Warnings (worth fixing — see inline comments)

1. **`disabled` prop doesn't actually disable the tags** — passing `aria-disabled` tells screen readers the group is disabled, but users can still click tags. Meanwhile the hidden `<select>` *is* disabled, so the form won't submit. (Inline on `TagGroup.tsx:204`.)
2. **`FormContext` override drops the parent's other fields** — `[FormContext, { validationBehavior }]` erases everything else from a parent `<Form>`. Spreading `formCtx` first keeps it safe for future descendants. (Inline on `TagGroup.tsx:186`.)
3. **Autofill / form reset won't clear the error** — `HiddenSelection`'s `onSelectionChange` skips `commitValidation()`, so an autofill that satisfies `required` wouldn't clear the displayed error until the user does something else. (Inline on `TagGroup.tsx:235`.)

#### Suggestions (small / optional)

4. `useId()` runs even without a label (`TagGroup.tsx`, near the top of the component body).
5. `aria-labelledby` is replaced instead of merged — if a consumer passes both `label` and `aria-labelledby`, the second is silently dropped.
6. The `{...(rest as RAC.TagGroupProps)}` cast shouldn't be needed; refining the rest-props type would let it go away.
7. Story `play` functions in `WithForm` and `Required` could use a blank line between Act and Assert for clearer AAA structure.
8. **Branch coverage on TagGroup dropped to 61.53%** and line 115 is uncovered (per the coverage bot). Many new branches (validation paths, `selectionMode='none'`, `selectedKeys==='all'`) aren't exercised. Not a blocker for this PR, but worth a follow-up to bring branch coverage back closer to the 85% threshold.

### AAA structure
The new tests follow AAA with blank-line separation between phases. CLAUDE.md doesn't mandate AAA, so it's a guideline. A few pre-existing tests in `TagGroup.test.tsx` don't strictly separate Act from Assert, but those weren't changed by this PR.

### Checklist
- [x] No `any` types (even removed one from the demo file)
- [x] `import type` used
- [x] Marigold prop naming (`disabled` etc.)
- [x] `useClassNames`, no inline styles
- [x] Vitest + `userEvent` (no Jest, no `fireEvent`)
- [x] Accessible queries; the one `querySelector('select')` is correctly explained in a comment
- [x] Story tests tagged `component-test`
- [x] Changeset present (`minor` — fits the beta-release context)

### Summary
The fix is on the right track and matches what the ticket asked for. The three warnings above are real — especially the `disabled` one, which a user would actually notice — so I'd resolve those before merging into `beta-release`.

---

packages/components/src/HiddenSelection/HiddenSelection.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+// The <select> is intentionally `aria-hidden` and lives inside VisuallyHidden,
+// so it's invisible to accessible queries — `querySelector` is the right tool.
+const getSelect = () =>
+  // eslint-disable-next-line testing-library/no-node-access
+  document.querySelector('select') as HTMLSelectElement | null;
+
+test('renders a hidden <select> that is not user-focusable', () => {
+  render(<Harness selectionMode="single" />);
+
+  const select = getSelect();
```
sarahgm, line 77: blank lines missing

sebald:
Thanks for the review! Just pushed `817841bc` covering the three warnings + the `aria-labelledby` merge + AAA blank-lines:

- ✅ **(1) `disabled` cascade** — added a `TagGroupContext` that each `<Tag>` consumes (mirrors `SelectListContext`). Tags now pick up `data-disabled` and ignore press events, matching the form-disabled hidden `<select>`. Covered by a new unit test.
- ✅ **(2) `FormContext` merge** — `[FormContext, { ...formCtx, validationBehavior }]`.
- ✅ **(3) `commitValidation` on the hidden-select path** — `HiddenSelection`'s `onSelectionChange` now goes through the same `handleSelectionChange` as the RAC path.
- ✅ **(5) `aria-labelledby` merge** — joined with the generated `labelId` instead of replaced.
- ✅ **(8) AAA blank-lines** — added to the new story play functions and `HiddenSelection` unit tests.

On the two remaining nits:

- ⏭️ **(4) `useId()` runs even without a label** — leaving as is. `useId` is cheap, and conditionalising hook calls is brittle / against React's rules; the unconditional call also keeps the value stable if a consumer toggles `label`.
- ⏭️ **(6) `{...(rest as RAC.TagGroupProps)}` cast** — leaving as is. The cast sits at the RAC boundary where Marigold's `TagGroupProps` and `RAC.TagGroupProps` diverge (different `onSelectionChange` shape, no `isInvalid`/`isRequired`/`isDisabled`). Refining the rest-props type to make the cast structurally valid would be a larger type-plumbing change with no runtime benefit.

Will follow up on **(7) branch coverage** in a separate ticket — the new validation/`selectionMode='none'`/`'all'` paths need their own targeted tests.

packages/components/src/TagGroup/TagGroup.tsx
sebald: Fixed in 817841bc. Added a `TagGroupContext` ([Context.ts](https://github.com/marigold-ui/marigold/blob/fix/DST-1363-taggroup-field-error/packages/components/src/TagGroup/Context.ts)) that carries `disabled` down to each `<Tag>` — mirrors the `SelectListContext` pattern. `<Tag>` now picks up `isDisabled: disabled ?? groupDisabled` so the cascade is real: each tag gets `data-disabled` (theme already styles that), and presses on disabled tags are ignored by RAC. Covered by a new unit test in 817841bc and tightened in 466d25fc.

packages/components/src/TagGroup/TagGroup.tsx
sebald: Fixed in 817841bc — `[FormContext, { ...formCtx, validationBehavior }]` now preserves any parent `<Form>` fields and only overrides `validationBehavior`.

packages/components/src/TagGroup/TagGroup.tsx
sebald: Fixed in 817841bc — both the `RACTagGroup` and the `HiddenSelection` paths now go through the same `handleSelectionChange` (which does `setSelection` + `validationState.commitValidation()`), so autofill / form-reset clears the displayed error consistently with user interaction.

packages/components/src/HiddenSelection/HiddenSelection.test.tsx
sebald: Fixed in 817841bc — added blank-line separation between Arrange / Act / Assert phases across the `HiddenSelection.test.tsx` cases (and matching separation in the new `TagGroup` story `play` functions).

sarahgm (approved):

PR #5453, opened 2026-05-20
fix(DST-1447): restore focus to Tray trigger on close

sarahgm:
# Description

When the mobile Autocomplete/ComboBox tray closes (via `Escape` or the close button), focus should return to the trigger button. It wasn't — `document.activeElement` ended up on `<body>`, so the trigger never showed a focus ring after dismissal. This was a latent a11y bug surfaced by [#5449](https://github.com/marigold-ui/marigold/pull/5449), which made VRT snapshots deterministic and exposed two snapshots in Chromatic Build 19 ("Mobile Autocomplete interaction", "Mobile ComboBox keyboard navigation") that consistently captured the trigger *without* focus.

## Root cause

`TrayModal` wraps RAC's `ModalOverlay` in framer-motion's `AnimatePresence` so the tray can slide in/out and support drag-to-dismiss. `AnimatePresence` delays React's unmount until the exit animation completes — and during that delay, RAC's `FocusScope contain` keeps focus trapped inside the modal. By the time the unmount finally commits and `FocusScope.restoreFocus` runs, the captured restore target is wrong and focus lands on `<body>`.

## Fix

Under `prefers-reduced-motion: reduce` (which Chromatic forces), `TrayModal` now skips `AnimatePresence` and the motion wrappers entirely and renders a plain RAC `ModalOverlay` + `Modal`. RAC unmounts synchronously, `FocusScope` runs its restore step cleanly, and the trigger reliably regains focus + `data-focus-visible="true"`.

**This is a partial fix.** Users without `prefers-reduced-motion: reduce` still hit the same race; the broader fix is tracked as follow-up. Scope here is making Chromatic + a11y snapshots deterministic and giving reduced-motion users correct focus behavior.

Uses `motion/react`'s built-in `useReducedMotion` hook to read the preference. Also migrates `SidebarToggleIcon` away from its inline module-level `matchMedia` constant to the same hook, so the repo has a single source of truth for the preference.

Closes DST-1447

## Screenshots / Preview

Diff is visible in the Chromatic snapshots for **Autocomplete: Mobile Autocomplete interaction** and **ComboBox: Mobile ComboBox keyboard navigation** — trigger now has a visible focus ring after dismissal.

| Before | After |
| ------ | ----- |
| Trigger has no focus ring after Escape (focus on \`<body>\`) | Trigger shows focus ring with \`data-focus-visible=\"true\"\` |

## Test Instructions

**Manual check with reduced motion:**

1. \`pnpm sb\` and open the Mobile Autocomplete or Mobile ComboBox story at the small-screen viewport (639px).
2. Chrome DevTools → \`Cmd+Shift+P\` → \"Show Rendering\" → set \"Emulate CSS media feature \`prefers-reduced-motion\`\" to **reduce**.
3. Click the trigger to open the tray, then press \`Escape\` to close.
4. The trigger should have a visible focus ring afterwards.
5. Turn emulation off — the tray should slide as before, and drag-to-dismiss should still work.

**Real success signal in Chromatic:**

- [ ] First Chromatic run rebaselines the two affected snapshots — accept them and merge.
- [ ] Re-run VRT on the merged commit: snapshots should be byte-identical and show the trigger focused.

**Other checks green locally:**

- [x] \`pnpm typecheck:only\`
- [x] \`pnpm lint\`
- [x] \`pnpm test:unit\` (1084 tests)
- [x] \`pnpm test:sb\` (130 story tests, incl. the previously failing Mobile Autocomplete/ComboBox tests)

## Breaking Changes

No

## Checklist

- [ ] Stories added/updated (with \`component-test\` tag where applicable) — existing Mobile Autocomplete/ComboBox tests cover the fix; no new stories needed
- [x] Unit tests added/updated — added \`useReducedMotion\` to the existing \`vi.mock('motion/react', ...)\` stub in \`Tray.test.tsx\`
- [ ] Component documentation added/updated (if it exists) — N/A (internal Tray behavior)
- [x] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) (for new/changed interactive components) — this PR fixes a focus-management bug; behavior now matches the WAI dialog pattern (focus returns to the trigger on close)
- [x] Visual regression tests updated (for UI changes) — Chromatic will rebaseline the two Mobile snapshots
- [x] Changeset added (\`pnpm changeset\`) — \`.changeset/dst-1447-tray-focus-restoration.md\`


packages/system/src/hooks/useReducedMotion.ts
```diff
@@ -0,0 +1,26 @@
...
+    ? window.matchMedia(REDUCED_MOTION_QUERY).matches
+    : false;
+
+const getServerSnapshot = () => false;
+
+/**
+ * Returns whether the user has requested reduced motion via the
+ * `prefers-reduced-motion: reduce` media query. Reactive to changes.
+ */
+export const useReducedMotion = (): boolean =>
```
sebald: **Do we need this hook at all?** `motion@12.34.3` (already a direct dep, already imported in `TrayModal.tsx`) ships:

```ts
declare function useReducedMotion(): boolean | null;
declare function useReducedMotionConfig(): boolean | null;
// and: <MotionConfig reducedMotion="always" | "never" | "user">
```

(verified in `node_modules/.../framer-motion/dist/types/index.d.ts`)

To be clear: `MotionConfig reducedMotion="user"` is **not** a substitute for the structural fix in `TrayModal.tsx` — even at zero duration, `AnimatePresence` still resolves its exit asynchronously and races with `FocusScope.restoreFocus`. The bypass branch you added is the right call. The question is only about the hook itself.

A related concern: we already have **three** ways to read this preference in the repo after this PR lands:

1. `packages/components/src/utils/reducedMotion.ts` — module-level constant, used by `SidebarToggleIcon.tsx` and `MorphCaret.tsx`
2. This new `@marigold/system` hook
3. `motion/react`'s built-in `useReducedMotion`

Can we collapse to one? Two options I'd float:

- **(a)** Drop this hook, `import { useReducedMotion } from 'motion/react'` at the `TrayModal` call site, and migrate `SidebarToggleIcon` + `MorphCaret` + delete `utils/reducedMotion.ts` (either here or in a follow-up).
- **(b)** Keep this hook as Marigold's single source of truth and migrate the existing two callers in this PR so we don't end up with three flavors permanently.

packages/system/src/hooks/useReducedMotion.ts
```diff
@@ -0,0 +1,26 @@
...
+  mq.addEventListener('change', callback);
+  return () => mq.removeEventListener('change', callback);
+};
+
+const getSnapshot = () =>
+  typeof window !== 'undefined' && window.matchMedia
+    ? window.matchMedia(REDUCED_MOTION_QUERY).matches
+    : false;
+
+const getServerSnapshot = () => false;
```
sebald: SSR semantics divergence worth noting: this returns `false` on the server; `motion/react`'s `useReducedMotion()` returns `boolean | null` (null during SSR). `false` matches the convention `useSmallScreen` uses, so it's internally consistent — just calling it out so we don't trip over the difference if we ever swap to motion's hook.

packages/system/src/hooks/useReducedMotion.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Returns whether the user has requested reduced motion via the
+ * `prefers-reduced-motion: reduce` media query. Reactive to changes.
+ */
+export const useReducedMotion = (): boolean =>
+  useSyncExternalStore(subscribe, getSnapshot, getServerSnapshot);
```
sebald: Is the live subscription actually useful for the caller? `TrayModal` reads `reducedMotion` to choose between two completely different render branches — if the OS preference flips while the tray is open, we'd swap from `<AnimatePresence><MotionModalOverlay>...` to `<ModalOverlay>...` mid-flight, which RAC's `FocusScope` won't love.

The existing `utils/reducedMotion.ts` philosophy ("sample once, full reload to change — preference rarely flips mid-session") has a real argument behind it. Worth a sentence in the JSDoc on which contract Marigold is committing to going forward — reactive (current) or sticky.

packages/components/src/Tray/TrayModal.tsx
```diff
@@ -71,6 +64,24 @@ export const TrayModal = ({
     }
   };
 
+  // Skip framer-motion under reduced motion: AnimatePresence delays unmount,
+  // which races with RAC's FocusScope cleanup and prevents the trigger from
+  // receiving focus on close. Plain ModalOverlay unmounts synchronously, so
+  // FocusScope restores focus as designed.
+  if (reducedMotion) {
```
sebald, line 73: The PR body is candid that real users without `prefers-reduced-motion` see no change — same `AnimatePresence`, same race. So this fixes the Chromatic snapshot + a11y test determinism, but the underlying focus-restoration bug still ships for the majority of mobile users.

Not asking to expand scope here, but could we either:

- Add a `// TODO(DST-xxxx): ...` here pointing at a follow-up ticket, or
- Note in the JSDoc / changeset that this is a partial fix?

Otherwise the next person reading this branch will reasonably assume DST-1447 closed the whole bug.

packages/system/src/hooks/useReducedMotion.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const mq = {
+    matches: false,
+    addEventListener: vi.fn((_event: string, handler: () => void) => {
+      changeHandler = handler;
+    }),
+    removeEventListener: vi.fn(),
+  };
+  window.matchMedia = vi.fn().mockReturnValue(mq);
+  const { result } = renderHook(() => useReducedMotion());
+  const { act } = await import('@testing-library/react');
```
sebald: Nit: dynamic import is unnecessary — `@testing-library/react` is already imported on line 1 for `renderHook`. Move to the top-level import:

```ts
import { renderHook, act } from '@testing-library/react';
```

and drop the `await import` here.

sebald (commented):
Strong fix for the snapshot determinism + Chromatic regression — skipping `AnimatePresence` entirely (instead of just zeroing the transition) is the right structural call. Comments inline; the main thing I'd like to discuss is whether we need a Marigold-owned `useReducedMotion` at all given `motion/react` already exports one (plus `useReducedMotionConfig` and `<MotionConfig reducedMotion="user">`).

packages/system/src/hooks/useReducedMotion.ts
sarahgm: Went with option (a) — deleted `@marigold/system/useReducedMotion` + test, replaced `TrayModal`'s import with `motion/react`'s `useReducedMotion`, and migrated `SidebarToggleIcon.tsx` away from its inline module-level constant to the same source. Single source of truth, no maintenance burden on our side.

packages/system/src/hooks/useReducedMotion.ts
sarahgm: Resolved by deletion — we follow motion/react's `boolean | null` contract directly now. `TrayModal`'s `if (reducedMotion)` check already treats `null` as falsy (same as our old `false` default), so behavior is unchanged.

packages/system/src/hooks/useReducedMotion.ts
sarahgm: Resolved by deletion — motion/react's hook owns its own contract (reactive, per their docs). The mid-flight switch concern you raised is still valid in theory, but it's now their hook's behavior to document/change, not ours to commit to.

packages/components/src/Tray/TrayModal.tsx
sarahgm: Extended the inline comment in `TrayModal` to make this explicit: _"Users without the preference still hit this race — follow-up needed for a full fix."_ Leaving the follow-up Jira ticket out of scope here.

packages/system/src/hooks/useReducedMotion.test.tsx
sarahgm: Resolved by deletion — the test file is gone now that the hook is gone.

sebald (approved):

PR #5455, opened 2026-05-21
feat(DST-1460): use MorphCaret in Accordion header

sebald:
# Description

Replaces the static `ChevronDown` icon in `Accordion.Header` with the new `MorphCaret` icon. `MorphCaret` animates between the closed (down) and open (up) states by interpolating its SVG `d` path with a 250ms `cubic-bezier(0.4, 0, 0.2, 1)` transition. The animation is suppressed when the user has `prefers-reduced-motion` enabled.

The now-unused `ChevronDown` icon (`packages/components/src/icons/ChevronDown.tsx`) has been removed; no other component referenced it.

Closes DST-1460

## Screenshots / Preview

| Before (static chevron) | After (morphing caret) |
| ----------------------- | ---------------------- |
| `v` swaps to `^` instantly on open | `v` smoothly morphs to `^` on open (and back) |

## Test Instructions

1. Run `pnpm sb` and open the **Components/Accordion → Basic** story.
2. Click each header to toggle. The caret should smoothly morph between down and up.
3. Enable “Reduce motion” in your OS and re-toggle: the caret should swap without an animation.

## Breaking Changes

No

## Checklist

- [x] Stories added/updated (with `component-test` tag where applicable) — existing Accordion stories cover the change
- [x] Unit tests added/updated — existing tests still pass
- [x] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) — icon is `aria-hidden`, no semantic change


jim761 (commented):
Nice cleanup — `ChevronDown` is verified unused and the deleted `rotate-180` workaround takes the stale "group-aria-expanded bug" comment with it, which is good. All CI green.

One thing worth noting (not a blocker for this PR):

#### Hydration mismatch from `reducedMotion`

The `reducedMotion` constant in `packages/components/src/utils/reducedMotion.ts` is sampled once at module load. On the server `window` is undefined → resolves to `false` → SSR HTML emits the full `transition: d 250ms cubic-bezier(...)`. On a client with `prefers-reduced-motion: reduce`, it resolves to `true` → React tries to emit `transition: none` on first render. React 19 logs this as a hydration **error** (not just a warning) for every Accordion header on the page:

> A tree hydrated but some attributes of the server rendered HTML didn't match the client properties. This won't be patched up.
> This can happen if a SSR-ed Client Component used: A server/client branch \`if (typeof window !== 'undefined')\`.

Reproduced locally on `/components/navigation/accordion` in dev mode by enabling Reduce Motion before page load — the error fires once per `MorphCaret` on the page.

This is a pre-existing issue with the util (introduced in DST-1447) and applies equally to `Panel.Header`. Since the util's docstring claims it's "SSR-safe", I'd suggest either:
1. Driving the transition opt-out from CSS (`@media (prefers-reduced-motion: reduce)`) so server and client emit the same `style`, or
2. Reading the preference in an effect/state so first render always matches SSR and reduced-motion clients downgrade after hydration.

Happy to file a separate ticket — doesn't need to land here.

jim761 (approved):

PR #5456, opened 2026-05-21
feat(DST-1462)!: support selectionMode="multiple" in ComboBox and align prop names with React Aria

sebald:
## Summary

`<ComboBox>` now supports multi-select via `selectionMode="multiple"`, mirroring the API on `<Select>`. While at it, the props for selection vs. input filtering are renamed so both components stay consistent with React Aria Components.

- Generic signature: `ComboBoxProps<T, M extends SelectionMode>` with `M` defaulting to `'single'`.
- New `renderValue?: (selectedItems: T[]) => ReactNode` prop to customize the mobile trigger when items are selected.
- Mobile trigger now uses `<ComboBoxValue>` so a comma-separated list of selected items renders in both single and multi modes.
- Exposes `<ComboBox.Value>` (re-export of react-aria's `ComboBoxValue`) for rendering selected items in custom layouts.

## Breaking changes

Prop names on `<ComboBox>` now match React Aria Components and Marigold's `<Select>`:

| Before              | After               | Purpose                      |
| ------------------- | ------------------- | ---------------------------- |
| `value`             | `inputValue`        | Controlled input filter text |
| `defaultValue`      | `defaultInputValue` | Default input filter text    |
| `onChange`          | `onInputChange`     | Input filter change handler  |
| `onSelectionChange` | `onChange`          | Selection change handler     |

In multiple selection mode `onChange` receives `Key[]`; in single mode it receives `Key | null` (matching React Aria's `ChangeValueType<M>`).

## Docs & stories

- New `combobox-multiple` demo plus a Multiselection section in the ComboBox docs.
- Existing demos (`combobox-action`, `combobox-controlled`, `combobox-async`) and the MDX prose are migrated to the new prop names.
- New `Multiple` and `WithRenderValue` stories, both tagged `component-test`.

## Test plan

- [x] `pnpm test:unit packages/components/src/ComboBox` — 17/17
- [x] `pnpm test:sb packages/components/src/ComboBox` — 12/12 (incl. `Multiple` and `WithRenderValue`)
- [ ] Manual smoke test in Storybook (`pnpm sb` → ComboBox → Multiple)
- [ ] Manual smoke test on mobile viewport (`WithRenderValue`)

Closes [DST-1462](https://reservix.atlassian.net/browse/DST-1462).

[DST-1462]: https://reservix.atlassian.net/browse/DST-1462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/content/components/form/combobox/combobox-controlled.demo.tsx
+++ b/docs/content/components/form/combobox/combobox-controlled.demo.tsx
@@ -8,7 +8,6 @@ export default () => {
       <ComboBox
         inputValue={currentValue}
         onInputChange={setCurrentValue}
-        defaultValue="dog"
         label="Animals"
       >
         <ComboBox.Option id="red panda">Red Panda</ComboBox.Option>
--- a/docs/content/components/form/combobox/combobox-controlled.demo.tsx
+++ b/docs/content/components/form/combobox/combobox-controlled.demo.tsx
@@ -6,9 +6,9 @@ export default () => {
   return (
     <Stack>
       <ComboBox
-        value={currentValue}
-        onChange={setCurrentValue}
-        defaultSelectedKey={3}
+        inputValue={currentValue}
+        onInputChange={setCurrentValue}
+        defaultValue="dog"
         label="Animals"
       >
         <ComboBox.Option id="red panda">Red Panda</ComboBox.Option>
--- a/packages/components/src/ComboBox/ComboBox.stories.tsx
+++ b/packages/components/src/ComboBox/ComboBox.stories.tsx
@@ -553,6 +553,66 @@ export const WithRenderValue: any = meta.story({
   },
 });
 
+export const MobileMultiple: any = meta.story({
+  tags: ['component-test'],
+  globals: {
+    viewport: { value: 'smallScreen' },
+  },
+  args: {
+    label: 'Animals',
+  },
+  render: args => {
+    const [selected, setSelected] = useState<Key | Key[] | null>([]);
+    return (
+      <Stack>
+        <ComboBox {...args} selectionMode="multiple" onChange={setSelected}>
+          <ComboBox.Option id="red panda">Red Panda</ComboBox.Option>
+          <ComboBox.Option id="cat">Cat</ComboBox.Option>
+          <ComboBox.Option id="dog">Dog</ComboBox.Option>
+          <ComboBox.Option id="aardvark">Aardvark</ComboBox.Option>
+          <ComboBox.Option id="kangaroo">Kangaroo</ComboBox.Option>
+        </ComboBox>
+        <pre data-testid="selected">selected: {JSON.stringify(selected)}</pre>
+      </Stack>
+    );
+  },
+  play: async ({ canvas, step }: any) => {
+    const trigger = await canvas.findByRole('button');
+
+    await step('Open tray', async () => {
+      await userEvent.click(trigger);
+      await waitFor(() =>
+        expect(canvas.getByRole('dialog')).toBeInTheDocument()
+      );
+    });
+
+    await step('Pick first option and verify tray stays open', async () => {
+      await userEvent.click(await canvas.findByText('Dog'));
+      expect(canvas.getByRole('dialog')).toBeInTheDocument();
+    });
+
+    await step('Pick second option and verify tray stays open', async () => {
+      await userEvent.click(await canvas.findByText('Kangaroo'));
+      expect(canvas.getByRole('dialog')).toBeInTheDocument();
+    });
+
+    await step('Verify both selections persisted', async () => {
+      await waitFor(() =>
+        expect(canvas.getByTestId('selected')).toHaveTextContent(
+          'selected: ["dog","kangaroo"]'
+        )
+      );
+    });
+
+    await step('Dismiss tray so the story renders cleanly', async () => {
+      await userEvent.keyboard('{Escape}');
+      await waitFor(() =>
+        expect(canvas.queryByRole('dialog')).not.toBeInTheDocument()
+      );
+    });
+  },
+});
+
 export const Mobile: any = meta.story({
   tags: ['component-test'],
   globals: {
--- a/packages/components/src/ComboBox/ComboBox.stories.tsx
+++ b/packages/components/src/ComboBox/ComboBox.stories.tsx
@@ -171,7 +171,7 @@ export const Controlled: any = meta.story({
     const [id, setId] = useState<Key | Key[] | null>(null);
     return (
       <Stack>
-        <ComboBox {...args} onSelectionChange={setId} label="Animals">
+        <ComboBox {...args} onChange={setId} label="Animals">
           <ComboBox.Option id="red panda">Red Panda</ComboBox.Option>
           <ComboBox.Option id="cat">Cat</ComboBox.Option>
           <ComboBox.Option id="dog">Dog</ComboBox.Option>
@@ -227,8 +227,8 @@ export const AsyncLoading: any = meta.story({
     });
     return (
       <ComboBox
-        value={list.filterText}
-        onChange={list.setFilterText}
+        inputValue={list.filterText}
+        onInputChange={list.setFilterText}
         items={list.items}
         label="Star Wars Character Lookup"
         {...args}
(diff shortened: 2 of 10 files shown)


packages/components/src/ComboBox/ComboBox.stories.tsx
```diff
@@ -461,6 +461,98 @@ export const LargeDataset: any = meta.story({
...
+          <ComboBox.Option id="cat">Cat</ComboBox.Option>
+          <ComboBox.Option id="dog">Dog</ComboBox.Option>
+          <ComboBox.Option id="aardvark">Aardvark</ComboBox.Option>
+          <ComboBox.Option id="kangaroo">Kangaroo</ComboBox.Option>
+        </ComboBox>
+        <pre data-testid="selected">selected: {JSON.stringify(selected)}</pre>
+      </Stack>
+    );
+  },
+  play: async ({ canvas }: any) => {
```
jim761, line 489: ⚠️ **No mobile multi-select play test.**

This play function covers desktop multi-select. `WithRenderValue` covers the mobile `renderValue` render path, but no test picks two options on the mobile tray and asserts both selections persist while the tray stays open. The Jira ticket (DST-1462) called out "prevent the tray from auto-closing on selection" as in scope — worth either extending the `Mobile` story with a `selectionMode="multiple"` variant or adding a `MobileMultiple` story that picks two items in a row and verifies the tray remains open with both selections reflected.

docs/content/components/form/combobox/combobox-controlled.demo.tsx
```diff
@@ -6,9 +6,9 @@ export default () => {
   return (
     <Stack>
       <ComboBox
-        value={currentValue}
-        onChange={setCurrentValue}
-        defaultSelectedKey={3}
+        inputValue={currentValue}
+        onInputChange={setCurrentValue}
+        defaultValue="dog"
```
jim761: ⚠️ **`defaultValue` now means *default selection*, not default input value.**

This demo lives under the "Controlled Usage with custom Filter" section, whose topic is controlling input *filtering* via `inputValue`/`onInputChange`. Adding `defaultValue="dog"` here mixes in a default *selection* and may lead readers to infer that `defaultValue` is the default filter text (as it used to be before this PR). Either drop this line, or move the selection-default illustration into a separate demo.

packages/components/src/ComboBox/ComboBox.test.tsx
```diff
@@ -133,3 +137,48 @@ test('supports specific empty state text', async () => {
...
+
+  expect(onChange).toHaveBeenLastCalledWith(['dog', 'kangaroo']);
+});
+
+test('exposes ComboBoxValue as ComboBox.Value', () => {
+  expect(ComboBox.Value).toBeDefined();
+});
+
+test('renderValue customizes the mobile trigger with selected items', () => {
+  window.matchMedia = mockMatchMedia([smallScreenQuery]);
```
jim761, line 177: 💡 **`window.matchMedia` reassignment leaks beyond this test.**

Line 12 sets `window.matchMedia = mockMatchMedia([])` once at module load. This test reassigns it to the small-screen variant but never restores it. It happens to be the last test in the file today, but any test added below would silently inherit the small-screen mock. Either reset in an `afterEach`, or restore at the end of the test:

```ts
const prev = window.matchMedia;
window.matchMedia = mockMatchMedia([smallScreenQuery]);
try {
  // ...assertions
} finally {
  window.matchMedia = prev;
}
```

packages/components/src/ComboBox/ComboBox.tsx
```diff
@@ -100,34 +89,35 @@ export interface ComboBoxProps
...
    */
   loading?: boolean;
+
+  /**
+   * Render the trigger value when one or more options are selected (mobile
+   * trigger only). Replaces the default trigger render. The placeholder still
+   * shows when nothing is selected. Must not contain focusable or interactive
+   * elements, since the trigger is itself a button.
+   */
+  renderValue?: (selectedItems: T[]) => ReactNode;
```
jim761, line 99: 💡 The JSDoc scopes `renderValue` to the mobile trigger, but the MDX section now also tells desktop users to render `<ComboBox.Value>` for custom selected-item layouts. Calling out the desktop-vs-mobile split here would mirror `Select`'s `renderValue` JSDoc and help IDE users discover the right escape hatch per viewport.

jim761 (commented):
## Automated Code Review

Re-review after the prop rename + cast cleanup. Three of the five prior items are resolved; the four below remain.

### Warnings (Should Fix)

**1. No mobile multi-select play test** — `WithRenderValue` proves `renderValue` *renders*, but no test picks two options on the mobile tray and asserts both selections persist + the tray stays open across picks. The Jira ticket called out "prevent the tray from auto-closing on selection" as in scope. Default `Tray.Trigger` + ComboBox state plumbing should handle this (mobile single-select only closes on Escape), but a behavioral play test would lock it in before someone refactors the tray later.

**2. `combobox-controlled.demo.tsx` adds `defaultValue="dog"` to the "Controlled Usage with custom Filter" section.** `defaultValue` now means *default selection*, not default input value. The MDX section's topic is controlled *input filtering* (`inputValue` / `onInputChange`); mixing a default selection in there muddies what the demo teaches. Either drop `defaultValue="dog"`, or move the selection-default illustration to its own demo.

### Suggestions (Nice to Have)

**3. `window.matchMedia = mockMatchMedia([smallScreenQuery])` is not restored** — `ComboBox.test.tsx:176`. It's the last test in the file today, so it works, but any test added below this point would silently inherit the small-screen mock. Wrap in `beforeEach`/`afterEach`, or restore the previous matcher at the end of the test.

**4. `renderValue` JSDoc still scopes itself to "(mobile trigger only)"** — The MDX now also tells desktop users to render `<ComboBox.Value>` for custom layouts. Mentioning the desktop-vs-mobile split in the JSDoc would match `Select`'s `renderValue` doc and help IDE users discover the right tool per viewport.

---

packages/components/src/ComboBox/ComboBox.stories.tsx
sebald: Added `MobileMultiple` in `ComboBox.stories.tsx` (commit 75aa8db) — opens the tray, picks two options, asserts the dialog stays open between picks, verifies both keys in `onChange` state, then dismisses with Escape. The Escape teardown is deliberate: `userEvent.click()` on tray options bleeds into framer-motion's `drag="y"` listener and leaves the modal transformed ~85px below the viewport, so without it the story URL would render with the Close button clipped. Locks in DST-1462's "tray stays open on selection" invariant.

docs/content/components/form/combobox/combobox-controlled.demo.tsx
sebald: Dropped `defaultValue="dog"` (commit 75aa8db). Agreed the demo's purpose is controlled input filtering, and the rename made the selection default read as a default input value at first glance.

packages/components/src/ComboBox/ComboBox.test.tsx
sebald: Wrapped the reassignment in try/finally with restore (commit 75aa8db). Went with try/finally over `afterEach` to keep the fix local — it's a one-off restore at the bottom of the file, an `afterEach` hook would be machinery for a single test.

packages/components/src/ComboBox/ComboBox.tsx
sebald: Pushing back on this one. `renderValue` is mobile-only by architecture: the desktop ComboBox trigger is an `<input>` element, which can only display a string — there's no slot to render arbitrary `ReactNode`. The MDX guidance to render `<ComboBox.Value>` for desktop layouts is the right escape hatch, not an alternative API surface to mirror in JSDoc. The Select comparison doesn't carry over because Select's desktop trigger is a `<button>`, so `renderValue` fits both viewports there. I'd rather keep the JSDoc accurate to the implementation than imply a desktop parity that doesn't exist.

sebald:
Closing without merge — won't ship. Tracking ticket DST-1462 transitioned to Won't Do.

PR #5461, opened 2026-05-22
feat(DST-1373): adopt slot-configuration pattern in Card

sebald:
# Description

Bring Card in line with the slot-configuration pattern that Panel established in DST-1367.

`Card.Header` is now a slot provider for `<Title>` and `<Description>` — drop them directly inside the header and the heading level, id, `aria-labelledby` wiring, and theme classes are handled for you. A bare `<Title>` placed directly inside `<Card>` (no `Card.Header` wrapper) is also picked up by the root, so title-only cards can skip the wrapper and still get padding and labelling. `<Card>` itself renders an `<article>` landmark, labelled by its `<Title>` (or by an explicit `aria-label` when there's no visible title). A new `headingLevel` prop (default `3`) controls the underlying heading tag.

Scope, by explicit choice: **Title + Description only**. No action-context publishing in `Card.Header` or `Card.Footer`. Action menus and button sizing on Card are deferred to a follow-up.

Theme `Card` slot map gains `title` and `description` entries — the typography that lived on the `header` slot moves to `title`, with a new `--card-accent` custom property so variant text colors (`master`, `admin`) pick up the matching accent automatically.

Consumer-facing docs are aligned with the new API: four existing Card docs demos (`card-grid`, `card-minimal`, `card-elevation`, `card-appearance`) move to `<Title>` / `<Description>` instead of ad-hoc `<Headline>` / `<Stack>` headers; a new `card-header` demo backs the docs Header section; the Bleed snippet now uses `<Title>`; the Inventory full-page example (`docs/app/(examples)/examples/inventory`) is migrated; and the Card anatomy diagram gains a Description bar so the visual matches the new slot vocabulary.

Closes DST-1373

## Preview

Storybook → Components → Card. Key stories: `Basic`, `WithFooter`, `WithMedia`, `MasterAndAdmin`, `HeadingLevels`, `AriaLabeled`, `TitleOnlyWithoutHeader`, plus existing padding / bleed / stretch / anti-pattern stories. Docs preview: `/components/card`. Pattern preview: `/examples/inventory`.

## Test Instructions

1. `pnpm test:unit` → `Card.test.tsx` (28 tests) passes.
2. `pnpm test:sb` → Card story-tests pass.
3. `pnpm sb` → walk through `Components/Card` stories; verify the `<article>` landmark is labelled by the Title, `headingLevel` overrides the heading tag, and `aria-label` works when no Title is present.
4. `pnpm start` → open `/components/card`; verify the Anatomy diagram, Header demo, and migrated `card-grid` / `card-minimal` / `card-elevation` / `card-appearance` demos render and read correctly. Also open `/examples/inventory` and verify the venue Card still looks right with the new Title/Description wiring.

## Breaking Changes

No — but worth flagging for consumers: bare strings inside `<Card.Header>` are no longer styled (typography moved off the `header` slot onto `title`). The migration is `<Title>…</Title>` inside the header. Raw composition still renders, just without the slot wiring.

## Checklist

- [x] Accessibility reviewed (article landmark with `aria-labelledby`/`aria-label`, configurable heading level)
- [x] Changeset added (`.changeset/card-slot-configuration-pattern.md`)


packages/components/src/Card/CardContext.tsx
```diff
@@ -1,18 +1,24 @@
...
+  headingLevel: number;
+  hasTitle: boolean;
+  titleSlotRef: RefCallback<Element>;
 }
 
-const Context = createContext<CardContext | null>(null);
+export const CardContext = createContext<CardContextValue | null>(null);
 
-export const CardProvider = Context.Provider;
+export const CardProvider = CardContext.Provider;
```
sarahgm: ⚠️ `CardProvider` is dead code after this PR. The new `Card.tsx` uses RAC's `<Provider values={[[CardContext, contextValue], …]}>` directly and `CardProvider` is no longer referenced anywhere in source (grep confirms). It's also not re-exported from the package barrel, so it can't be relied on by external consumers either. Consider removing it.

packages/components/src/Card/Card.tsx
```diff
@@ -64,22 +81,59 @@ export const Card = ({
...
-      <div
+    <Provider
+      values={[
+        [CardContext, contextValue],
+        [HeadingContext, rootHeadingProps],
+      ]}
+    >
+      <article
         {...props}
+        aria-labelledby={!ariaLabel && hasTitle ? titleId : undefined}
```
sarahgm: ⚠️ When a consumer renders `<Card>` without a `<Title>` **and** without `aria-label`, both `aria-labelledby` and `aria-label` resolve to `undefined`, producing an unnamed `<article>` landmark. The `BareChildrenAntiPattern` story now sidesteps this by passing `aria-label`, which suggests awareness — but there's no dev-time guardrail. Options: (a) add a `console.warn` in development when neither is present, or (b) call this out explicitly in the new Accessibility section of `index.mdx`.

Side note: this `!ariaLabel && hasTitle` guard is actually stricter (and more correct) than `Panel.tsx`'s `!ariaLabel` — worth backporting to Panel in a follow-up so it doesn't dangle an `aria-labelledby` on an absent title.

packages/components/src/Card/Card.test.tsx
```diff
@@ -177,6 +199,71 @@ describe('Card', () => {
     });
   });
 
+  describe('Title in Card.Header', () => {
```
sarahgm, line 201: 💡 The `Title in Card.Header` describe block overlaps with the top-level `Rendering` block. Three tests assert essentially the same fact — that the Title renders as `<h3>` by default and labels the article:

- line 20: `renders an <article> labelled by the Title` (asserts H3 + aria-labelledby)
- line 203: `defaults to an <h3>` (asserts H3)
- line 231: `mirrors the titleId onto the heading id` (asserts `title.id === article.aria-labelledby`)

Consider collapsing into one test, or trimming the duplicated assertions so each test has a single focused intent (sharper AAA — each test arranges, acts, and asserts one specific behavior).

packages/components/src/Card/Card.stories.tsx
```diff
@@ -8,27 +11,43 @@ const meta = preview.meta({
...
-        type: 'boolean',
-      },
+      control: { type: 'boolean' },
       description: 'Whether the card stretches to fill its container.',
     },
   },
 });
 
 export const Basic = meta.story({
+  args: { children: null as never },
```
sarahgm: 💡 `args: { children: null as never }` is repeated on almost every story. It's already a workaround for Storybook's `children` ArgType (which is also `disable`d in `argTypes` at line 20–22). Consider lifting it into `meta.args` once instead of repeating it on each story — less noise, single source of truth.

sarahgm (commented):
## Automated Code Review

Nice, focused follow-on to the Panel slot-configuration work. The pattern is correctly transplanted, the deviations from Panel are deliberate (default `headingLevel=3` vs Panel's `2`, stricter `!ariaLabel && hasTitle` guard on `aria-labelledby`), and the changeset accurately flags the soft-breaking typography move off the `header` slot.

No critical issues; two warnings and four suggestions inline. Approve once the `CardProvider` dead-code and unnamed-landmark notes are either addressed or explicitly deferred.

---
*Generated with Claude Code review-pr skill — focus: Card-related changes + AAA test structure*

docs/content/components/collection/card/card-grid.demo.tsx
```diff
@@ -24,27 +24,26 @@ export default () => (
...
+            </Text>
           </Stack>
-        </Card.Header>
-        <Card.Body>
-          <Text variant="muted">
-            {venue.city}, {venue.country} · up to{' '}
-            {venue.capacity.toLocaleString()} guests
-          </Text>
         </Card.Body>
         <Card.Footer>
```
sarahgm, line 42: 💡 The previous version of this demo had a `<Button variant="ghost" size="small">View details</Button>` in the footer. The PR description lists this file as “migrated to `<Title>` / `<Description>`” — please confirm dropping the “View details” action was intentional and not collateral. If intentional, no action needed.

docs/content/components/collection/card/card-elevation.demo.tsx
```diff
@@ -18,26 +19,24 @@ export default () => (
...
+      </Description>
+    </Card.Header>
+    <Card.Body>
       <Stack space="tight">
-        <Headline level={3}>Annual Conference 2025</Headline>
         <Inline space="related">
           <Badge variant="info">Music</Badge>
           <Badge variant="success">Tickets available</Badge>
         </Inline>
+        <Text variant="muted">Join us for our flagship annual event.</Text>
```
sarahgm: 💡 The original single-sentence body (“Join us for a day of inspiring talks, workshops, and networking at our flagship annual event.”) is now split across `<Description>` (“A day of inspiring talks, workshops, and networking.”) and `<Text>` (“Join us for our flagship annual event.”). The split reads slightly stilted — minor copy polish, not blocking.

packages/components/src/Card/Card.stories.tsx
```diff
@@ -43,10 +62,74 @@ export const Basic = meta.story({
...
+        <Text>The Title renders as an h4 in the document outline.</Text>
+      </Card.Body>
+    </Card>
+  ),
+});
+
+HeadingLevels.test(
+  'renders the Title at the configured heading level',
+  async ({ canvas }) => {
+    const title = canvas.getByRole('heading', { name: 'Heading at level 4' });
```
sarahgm, line 98: whitespace missing

packages/components/src/Card/Card.stories.tsx
```diff
@@ -43,10 +62,74 @@ export const Basic = meta.story({
...
+        </Text>
+      </Card.Body>
+    </Card>
+  ),
+});
+
+AriaLabeled.test(
+  'uses aria-label as the accessible name and omits aria-labelledby',
+  async ({ canvas }) => {
+    const article = canvas.getByRole('article', { name: 'Quick stats card' });
```
sarahgm, line 121: whitespace missing

themes/theme-rui/src/components/Card.styles.ts
```diff
@@ -2,20 +2,26 @@ import { ThemeComponent, cva } from '@marigold/system';
...
   container: cva({
-    base: 'rounded-md shadow-elevation-raised',
+    base: 'rounded-md shadow-elevation-raised [--card-accent:currentColor]',
     variants: {
       variant: {
         default: 'ui-surface',
-        master: 'bg-access-master/40 border border-access-master-accent',
-        admin: 'bg-access-admin/40 border border-access-admin-accent',
+        master:
+          'bg-access-master/40 border border-access-master-accent [--card-accent:var(--color-access-master-foreground)]',
```
sarahgm: why change of color tokens? Still should be 
```suggestion
          'bg-access-master/40 border border-access-master-accent [--card-accent:var(--color-access-master-accent)]',
```
or for what is this used?

themes/theme-rui/src/components/Card.styles.ts
```diff
@@ -2,20 +2,26 @@ import { ThemeComponent, cva } from '@marigold/system';
...
+        admin:
+          'bg-access-admin/40 border border-access-admin-accent [--card-accent:var(--color-access-admin-foreground)]',
```
sarahgm: same here then 
```suggestion
          'bg-access-admin/40 border border-access-admin-accent [--card-accent:var(--color-access-admin-accent)]',
```

themes/theme-rui/src/components/Card.styles.ts
```diff
@@ -2,20 +2,26 @@ import { ThemeComponent, cva } from '@marigold/system';
...
     },
     defaultVariants: {
       variant: 'default',
     },
   }),
-  header: cva({
-    base: 'text-lg font-semibold',
+  header: cva({}),
+  title: cva({
+    base: 'text-lg font-semibold leading-none text-(--card-accent)',
```
sarahgm: ah I wouldn't call text-(--card-accent) this would mean to me to use the accent color token not the foreground

themes/theme-rui/src/components/Card.styles.ts
sebald: Nothing about the existing tokens changed here — the border still uses `border-access-master-accent`. What this PR *adds* is a new `--card-accent` CSS var that drives the **title text** color, set to the variant `-foreground` token.

Using `-foreground` (not `-accent`) for text is deliberate:

- **Contrast:** `-foreground` is the darker step (`orange-800` / `purple-900`) vs `-accent` (`orange-700` / `purple-800`), so it has more contrast on the `/40` background.
- **Convention:** `Badge.styles.ts` already pairs `text-access-*-foreground` (text) with `border-access-*-accent` (border) — text = foreground.
- **Parallel to Panel:** `Panel.styles.ts` does the identical thing today (`[--panel-accent:var(--color-access-*-foreground)]`) to tint its title.

Switching the title to the `-accent` token would lower contrast and diverge from both Badge and Panel, so I’m keeping `-foreground` here. Naming is a fair point — replied separately on the `title` line.

themes/theme-rui/src/components/Card.styles.ts
sebald: Fair — the name reads ambiguously. It’s intentional parallel to Panel: `Panel.styles.ts` already exposes `--panel-accent` set to the variant `-foreground` color and applies it to the title the same way. So `--card-accent` is meant as "the card’s per-variant title color" (`currentColor` for `default`, the access palette’s text color for `master`/`admin`) rather than literally the `-accent` token.

I’d rather not rename just `--card-accent` here, since that desyncs it from `--panel-accent`. If the "accent name holding a foreground color" bothers us (it’s a fair critique), I think the right move is a small follow-up that renames **both** `--panel-accent` and `--card-accent` to something foreground-accurate in one go, so the two stay consistent. Keeping as-is in this PR.

packages/components/src/Card/CardContext.tsx
sebald: Confirmed dead — it was the single-context JSX wrapper (`<CardProvider value={…}>`) right up until this PR swapped to RAC’s multi-context `<Provider values={[…]}>`. Removed in c72b48191. Only `Card`/`CardProps` are exported from the barrel, so no consumer impact.

packages/components/src/Card/Card.tsx
sebald: Did both: c72b48191 adds a dev `console.warn` when a Card has neither `<Title>` nor `aria-label`, **and** a note in the Accessibility section of `index.mdx`. Good catch on the side note too — backported the stricter `!ariaLabel && hasTitle` guard to Panel in #5486 (DST-1480).

packages/components/src/Card/Card.test.tsx
sebald: Trimmed the redundant `defaults to an <h3>` test in c72b48191 — the `headingLevel` `test.each` already asserts level 3. Kept `mirrors the titleId` since it has a distinct intent (id wiring, not the tag).

packages/components/src/Card/Card.stories.tsx
sebald: Hoisted `children: null as never` into `meta.args` (with a short explanatory comment) in c72b48191 and removed it from every story.

docs/content/components/collection/card/card-grid.demo.tsx
sebald: Intentional — the footer now carries the price, and the "View details" action was dropped on purpose as part of the Title/Description migration. No change.

docs/content/components/collection/card/card-elevation.demo.tsx
sebald: Polished in c72b48191: Description is now a short tagline ("Our flagship annual event.") and the body keeps the full sentence, so the two no longer echo each other.

packages/components/src/Card/Card.stories.tsx
sebald: Added the missing blank line (AAA spacing) in c72b48191.

packages/components/src/Card/Card.stories.tsx
sebald: Added the missing blank line (AAA spacing) in c72b48191.

themes/theme-rui/src/components/Card.styles.ts
sarahgm: but panel-accent is then also wrong when its for foreground color?!

sarahgm: My problem is that accent should be used for accents, and foreground for text colors, but you call the variable -accent althrough its using foreground color. But thats missleading as you noticed in my previous comments. So call it card-text ? panel-text ?

packages/components/src/Card/Card.tsx
```diff
@@ -64,22 +81,67 @@ export const Card = ({
...
```
sarahgm: 💡 `{...props}` is spread above this line, so an explicit `aria-labelledby={… : undefined}` will override a consumer-supplied `aria-labelledby` with `undefined` when no `<Title>` is present. Edge case (the slot pattern owns labelling), and it matches Panel, so low priority — flagging for awareness.

packages/components/src/Card/CardContext.tsx
```diff
@@ -1,18 +1,22 @@
+import type { RefCallback } from 'react';
 import { createContext, use } from 'react';
 import type { ThemeComponent } from '@marigold/system';
 
-export interface CardContext {
+export interface CardContextValue {
   classNames: {
     [Key in keyof ThemeComponent<'Card'>]: string;
   };
+  variant?: string;
```
sarahgm: 💡 `variant` is added to `CardContextValue` and populated in `contextValue`, but none of the sub-components (`CardBody`/`CardFooter`/`CardHeader`/`CardMedia`) read it — they only consume `classNames` (+ heading wiring in the header). If this is scaffolding for the deferred action slots, a short comment would help; otherwise it can be dropped.

packages/components/src/Card/Card.test.tsx
```diff
@@ -177,6 +199,62 @@ describe('Card', () => {
...
+        render(<Basic.Component headingLevel={level} />);
+
+        const title = screen.getByRole('heading', {
+          name: 'Professor Severus Snape',
+        });
+        expect(title.tagName).toBe(`H${level}`);
+      }
+    );
+
+    test('renders at level 4 in the HeadingLevels story', () => {
```
sarahgm: 💡 This overlaps with both the `test.each([2,3,4,5,6])` above (which already covers level 4) and the story-level `HeadingLevels.test`. One of the three is redundant — same spirit as the earlier AAA-trim feedback.

packages/components/src/Card/CardFooter.tsx
```diff
@@ -15,6 +15,7 @@ export interface CardFooterProps {
 
 export const CardFooter = ({ children, bleed }: CardFooterProps) => {
   const { classNames } = useCardContext();
+
```
sarahgm: 💡 Whitespace-only change (blank line after `useCardContext()`). Harmless, but it's noise in an otherwise focused diff.

themes/theme-rui/src/components/Card.styles.ts
sebald: Agreed — went with your suggestion. `--card-accent` is now `--card-text`, and I renamed `--panel-accent` → `--panel-text` to match. The color values are unchanged (still the variant `-foreground`), it’s purely the misleading name that’s fixed.

- Card: pushed here (`themes/theme-rui/src/components/Card.styles.ts`).
- Panel: pushed in #5486 (DST-1480), changeset updated to include `@marigold/theme-rui`.

No visual or behavioural change in either.

packages/components/src/Card/Card.tsx
sebald: Good flag — addressed in 5fdcd0d. `aria-labelledby` now falls back to the consumer-supplied value (`props['aria-labelledby']`) instead of overwriting it with `undefined` when there's no `<Title>`. The slot-owned `titleId` still wins when a Title is present. Mirroring the same fallback onto Panel in #5486.

packages/components/src/Card/CardContext.tsx
sebald: Dropped in 5fdcd0d — removed `variant` from both `CardContextValue` and the `contextValue` object. It was genuinely unread by every sub-component (the `variant` prop feeding `useClassNames` on the root is untouched). Will re-introduce via context only if/when the deferred action slots actually need it.

packages/components/src/Card/Card.test.tsx
sebald: Agreed — removed in 5fdcd0d. The `headingLevel` `test.each` already asserts level 4 and the `component-test` story covers it in-browser, so the standalone `HeadingLevels` test was pure duplication. Same AAA-trim spirit as the earlier round.

packages/components/src/Card/CardFooter.tsx
sebald: Reverted in 5fdcd0d.

packages/components/src/Card/Card.tsx
sebald: Correction / update: the consumer-`aria-labelledby` fallback didn't land in #5486 (that PR only shipped the stricter `hasTitle` guard). #5486's Panel still spread no props, so there was no consumer value to preserve — Panel simply didn't accept arbitrary HTML attributes.

I've now closed that gap on this branch in 5d16f9e (DST-1480): `Panel` extends `HTMLAttributes<HTMLElement>`, spreads `...props` onto its root `<section>`, and uses the same `props['aria-labelledby']` fallback as Card here. So Card and Panel are now genuinely in parity. Changeset: `.changeset/dst-1480-panel-html-attributes.md` (minor).

sarahgm (approved):

PR #5472, opened 2026-05-26
feat(DST-1238): align Dialog with the shared ui-panel-* pattern

sarahgm:
# Description

Aligns Dialog with the shared `ui-panel-*` utility pattern already used by Drawer, Tray, and Sidebar. Dialog now adopts `ui-panel-content` and `ui-panel-actions`; its actions gain a `border-t` divider, content gains consistent `py-4` vertical padding, and actions padding becomes symmetric. Responsive button stacking (`flex-col-reverse sm:flex-row`) and the original header treatment (`px-6 pt-6`, no border) remain Dialog-specific.

A new `ui-panel-content` utility is introduced for shared scrollable content rhythm. Tray retains its compact `p-2` content padding and Sidebar retains its tighter `px-3 py-1` nav padding — both opt out of `ui-panel-content` because those areas have intentionally tighter densities.

Drawer migrates inline padding/scroll classes to the new `ui-panel-content` utility (identical CSS output, no visual change). Tray and Sidebar style files are untouched. The `ui-panel-*` block in `ui.css` now has an inline docstring clarifying it's for modal-style panels and distinct from the `Panel` component.

Closes DST-1238

## Design notes

**Borders are decided per-border, based on functional value.** The ticket asks to keep borders configurable so borderless looks remain possible. Rather than add a `bordered`/`borderless` prop on Dialog, each border in the utility set is evaluated on its own merit:

- *Actions border-top* (kept on Dialog): marks the transition from informational content to interactive choices — a useful mental pause matching macOS/iOS/Material action sheet conventions, helpful at any dialog size.
- *Header border-bottom* (skipped on Dialog): would sit directly under the title and visually disconnect it from the message it introduces. Title and message are conceptually one unit ("the what"), and a divider between them feels chopped on compact confirmation dialogs — exactly the "rigid styles preventing Dialog from scaling to smaller sizes" the ticket warns against.

A future component needing a different combination can compose its own classes; the utilities don't lock anyone in.

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Open Storybook → Components/Dialog and exercise the stories at both mobile (≤640px) and desktop viewports. Confirm:
   - Actions row has a top border and symmetric `py-4` padding
   - Content area has comfortable vertical breathing room
   - Mobile button layout stacks vertically (column-reverse); desktop is horizontal
   - Header padding and treatment is unchanged from before (no new border under the title)
2. Open Components/Drawer, Components/Tray, and Components/Sidebar — confirm visual output is unchanged.

## Breaking Changes

No

## Checklist


themes/theme-rui/src/components/Dialog.styles.ts
```diff
@@ -18,11 +18,11 @@ export const Dialog: ThemeComponent<'Dialog'> = {
     },
   }),
   header: cva({
-    base: 'flex flex-col gap-1 text-center sm:text-left px-6 pt-6',
```
OsamaAbdellateef, line 21: Why did we remove the gap ? @sarahgm

sarahgm: Good catch checking on that! The `gap-1` on the `header` slot was actually a no-op — the `Header` element only ever renders a single child (the `Heading`), and flex `gap` only takes effect between siblings:

```tsx
// DialogTitle.tsx
<Header className={cn('[grid-area:title]', classNames.header)}>
  <Heading slot="title" className={classNames.title}>
    {children}
  </Heading>
</Header>
```

The spacing you might be thinking of — between the title and the message/content below it — is unchanged (well, slightly larger now):
- `title` still has `mb-1`
- `content` went from `py-1` to `py-4` via `ui-panel-content`

So the visible gap above content actually grew from `0.5rem` to `1.25rem`, in line with the consistent content rhythm the `ui-panel-*` utilities give us. Removing the dead `gap-1` was just cleanup.

OsamaAbdellateef (approved):

PR #5476, opened 2026-05-28
feat(DST-1395): ✨ Make SelectList responsive (flip horizontal → vertical)

OsamaAbdellateef:
# Description

Horizontal `<SelectList>` layouts previously stayed in a row regardless of available space, breaking in narrow containers (drawers, sidebars, narrow cards). This change adds an automatic flip to a vertical stack when the wrapping container is narrower than `28rem` (~448px), driven by a CSS `@container` query on the `FieldBase` wrapper.

The flip is always-on for `orientation="horizontal"` — no new prop needed. Keyboard navigation continues to work across both axes because `RACGridList` uses `layout="grid"` which supports all four arrow keys regardless of orientation.

**What changed:**
- `SelectList.styles.ts` — added `@max-md/selectlist:` Tailwind variants to `container`, `list`, and `item` slots; per-corner border-radius overrides on first/last items to fix the flip transition (side-based `rounded-l-none` was clobbering the top-left corner due to Tailwind's cascade sort order)
- `SelectList.tsx` — moved `@container/selectlist` from a child `<div>` to `FieldBase`'s `className`; putting it on the `w-fit` surface itself would have applied size containment and collapsed the element
- `SelectList.stories.tsx` — added `HorizontalResponsive` story with wide (640px) and narrow (320px) containers and a `play` function that asserts flex direction in each
- `SelectList.test.tsx` — replaced a stale orientation test with a container-scope assertion
- `docs/…/selectlist/index.mdx` — extended the "Horizontal orientation" section to document the auto-flip behaviour

## Test Instructions

1. Run `pnpm sb` and navigate to **Components / SelectList / Horizontal Responsive**
2. Confirm items appear in a **row** in the wide (640px) container
3. Confirm items appear in a **column** in the narrow (320px) container
4. Resize the browser viewport to verify the flip triggers on the *container* width, not the viewport width (e.g. render inside a 300px sidebar while the viewport is wide)
5. Tab into the list and use arrow keys — confirm Up/Down still work in the flipped vertical layout
6. Run `pnpm test:unit` and `pnpm test:sb` — all tests should pass

## Checklist


themes/theme-rui/src/components/SelectList.styles.ts
```diff
@@ -6,6 +6,10 @@ export const SelectList: ThemeComponent<'SelectList'> = {
       'flex',
       'has-orientation-vertical:w-full',
       'has-orientation-horizontal:w-fit',
+      // When horizontal and the @container/selectlist scope is narrower than
+      // the flip threshold (md ≈ 28rem), the surface fills its parent so the
+      // stacked layout reads as a regular vertical list.
+      'has-orientation-horizontal:@max-md/selectlist:w-full',
```
sarahgm: 🔴 **Must fix — the list spills out of its box at medium widths.**

The flip only happens below **448px** (`@md`), but a horizontal row can't shrink below ~**640px** (each item is `min-w-40` and won't shrink, and the surface is `w-fit`). So between **448–640px** the list is too wide to fit but not narrow enough to flip, and it pushes past the edge of the card. This shows up at the Storybook "small screen" size (639px) and in real sidebars/drawers.

Also, the flip point (448px) is narrower than what the design system calls "small" (`--breakpoint-sm: 40rem` = 640px), so it waits longer to flip than feels natural.

**Suggested fix** — cap the surface and list at the container width so they can never grow past it (the inner list then scrolls instead of overflowing):
```ts
// container slot (line ~8)
'has-orientation-horizontal:w-fit has-orientation-horizontal:max-w-full',
// list slot (line ~26)
'orientation-horizontal:w-fit orientation-horizontal:max-w-full orientation-horizontal:flex-row orientation-horizontal:overflow-x-auto orientation-horizontal:overflow-y-hidden',
```
If a scrollbar isn't wanted in that range, also raise the flip threshold so it switches to vertical before it would overflow. (Note: the right flip point depends on item count, since min-content grows with the number of items.)

packages/components/src/SelectList/SelectList.stories.tsx
```diff
@@ -490,6 +490,73 @@ export const Horizontal = meta.story({
...
+            <Text slot="label">Express</Text>
+            <Text slot="description">1–2 business days</Text>
+          </SelectList.Option>
+          <SelectList.Option id="overnight" textValue="Overnight">
+            <Text slot="label">Overnight</Text>
+            <Text slot="description">Next business day</Text>
+          </SelectList.Option>
+        </SelectList>
+      </div>
+      <div data-testid="narrow-container" style={{ width: 320 }}>
```
sarahgm: 🔴 **Must fix — this test hides the bug above.**

This story only checks a 320px box (flips ✅) and a 640px box (row ✅). It skips the 448–640px range where the list actually overflows, so the `play` test passes even though the layout breaks there.

Please add a case at a middle width (e.g. a 500px container) that asserts the rendered list never grows wider than its container (compare `getBoundingClientRect()` of the list vs. the container).

docs/content/components/form/selectlist/index.mdx
```diff
@@ -139,6 +139,8 @@ Set `orientation="horizontal"` to switch the layout.
 
 <ComponentDemo name="selectlist-horizontal" />
 
+A horizontally arranged list automatically flips to a vertical stack once its **wrapping container** is narrower than `28rem` (≈448px). The flip is driven by a CSS container query on the list's own wrapper, so it reacts to the parent it is rendered into a sidebar, a drawer, a narrow card not just the viewport width. Keyboard navigation continues to work in both axes after the flip.
```
sarahgm: 🟡 **Should fix — hard to read.** The dashes were removed but no other punctuation replaced them, so this runs together ("...the parent it is rendered into a sidebar, a drawer, a narrow card not just the viewport width").

```suggestion
A horizontally arranged list automatically flips to a vertical stack once its **wrapping container** is narrower than `28rem` (≈448px). The flip is driven by a CSS container query on the list's own wrapper, so it reacts to the parent it is rendered into, whether a sidebar, a drawer, or a narrow card, not just the viewport width. Keyboard navigation continues to work in both axes after the flip.
```

.changeset/chilly-hotels-wonder.md
```diff
@@ -0,0 +1,7 @@
+---
+'@marigold/components': patch
+'@marigold/theme-rui': patch
+'@marigold/docs': patch
```
sarahgm: 💡 **Nice to have.** `@marigold/docs` is a private package, so it doesn't need a changeset entry. Safe to drop this line.
```suggestion
```

sarahgm (changes requested):
## Automated Code Review

### Summary
Great direction — flipping based on the **container** width (not the window) is the right call, and the tricky bits are well commented. But there's a real gap: in medium-width containers the horizontal list spills out of its box instead of adapting, and the new test doesn't catch it. Details below.

### 🔴 Must fix

**1. The horizontal list spills out of its box at medium widths.**
The list flips to vertical only when its container is narrower than **448px**. But a row of items can't get smaller than about **640px** (each item has a fixed minimum width and won't shrink). So when the container is between **448px and 640px**, the list is too wide to fit but not narrow enough to flip — and it pushes out past the edge of the card. This happens at the Storybook "small screen" size (639px) and in real sidebars/drawers — the exact cases this ticket targets.

**2. The flip point is narrower than people expect "small" to be.**
The flip happens at 448px (container `@md`), but the design system treats "small screen" as 640px (`--breakpoint-sm: 40rem`). So the list waits much longer to flip than feels natural — which is what causes problem #1. The flip point should be a deliberate choice tied to how wide the items actually are, not just Tailwind's default `@md`.

**3. The test hides the bug.**
The new `HorizontalResponsive` test only checks a 320px box (flips) and a 640px box (stays in a row). It never checks the in-between sizes where the list breaks, so it passes even though the problem is real. Add a test at a middle width (e.g. 500px) asserting the list never grows wider than its container.

### 🟡 Should fix
- **Docs:** two sentences are hard to read because the dashes were removed without adding other punctuation. Add commas so they read cleanly.

### 💡 Nice to have
- **Changeset:** remove `@marigold/docs` (private, not needed) and add a final newline.
- **Story test:** uses `getByTestId`; accessible queries are preferred.
- **Keyboard check:** the ticket asks to confirm keyboard nav after the flip — worth a quick test, not just manual notes.
- **Unit test:** relies on a class from a different component (`FieldBase`), which could break for unrelated reasons.

### Recommendation
**Request Changes.** The main feature doesn't fully work yet: in narrow boxes the list overflows instead of adapting, and the tests don't catch it. The fix is small. The docs wording should also be cleaned up.

---

sarahgm (commented):
Automated review focused on documentation correctness. The main theme: the flip breakpoint is documented as 28rem (448px) in several places, but the code actually flips at 40rem (640px). See inline comments.

themes/theme-rui/src/components/SelectList.styles.ts
```diff
@@ -5,7 +5,11 @@ export const SelectList: ThemeComponent<'SelectList'> = {
     base: [
       'flex',
       'has-orientation-vertical:w-full',
-      'has-orientation-horizontal:w-fit',
+      'has-orientation-horizontal:w-fit has-orientation-horizontal:max-w-full',
+      // When horizontal and the @container/selectlist scope is narrower than
+      // 40rem, the surface fills its parent so the stacked layout reads as a
+      // regular vertical list.
+      'has-orientation-horizontal:@max-[40rem]/selectlist:w-full',
```
sarahgm, line 12: 🔴 The flip is set to trigger below **40rem (640px)** here. But the changeset, the docs, and the PR description all say **28rem (448px)**. These need to match. Please confirm the intended value — and note the medium-container story (500px) suggests 448px was the original intent.

docs/content/components/form/selectlist/index.mdx
```diff
@@ -139,6 +139,8 @@ Set `orientation="horizontal"` to switch the layout.
 
 <ComponentDemo name="selectlist-horizontal" />
 
+A horizontally arranged list automatically flips to a vertical stack once its **wrapping container** is narrower than `28rem` (≈448px). The flip is driven by a CSS container query on the list's own wrapper, so it reacts to the parent it is rendered into, whether a sidebar, a drawer, or a narrow card, not just the viewport width. Keyboard navigation continues to work in both axes after the flip.
```
sarahgm: 🔴 This says the list flips below **28rem (≈448px)**, but the actual code flips below **40rem (640px)** (`SelectList.styles.ts`). Readers building a ~500px sidebar would expect a row but get a vertical stack. Update this number once the intended breakpoint is confirmed.

docs/content/components/form/selectlist/index.mdx
```diff
@@ -148,8 +150,8 @@ Set `orientation="horizontal"` to switch the layout.
   </Do>
   <Dont>
     <DontDescription>
-      Don't use horizontal for long lists, rich rows, or narrow viewports.
-      Options scrolled out of view is the worst of both shapes.
+      Don't use horizontal for long lists or rich rows that wouldn't read well
+      stacked either pick the orientation the content fits.
```
sarahgm: ⚠️ This sentence is missing punctuation and reads as a run-on: *“…wouldn't read well stacked either pick the orientation the content fits.”* Suggested: *“…rich rows that wouldn't read well stacked. Pick the orientation the content fits.”*

.changeset/chilly-hotels-wonder.md
```diff
@@ -0,0 +1,6 @@
+---
+'@marigold/components': patch
+'@marigold/theme-rui': patch
+---
+
+feat([DST-1395]): **SelectList** horizontal layouts now automatically flip to a vertical stack when the wrapping container is narrower than `28rem` (~448px).
```
sarahgm: 💡 Same breakpoint mismatch here — says **28rem (~448px)** but the code uses **40rem (640px)**. Also: ticket should be `DST-1395` (not `[DST-1395]`) per convention, and the file is missing a trailing newline.

packages/components/src/SelectList/SelectList.stories.tsx
```diff
@@ -490,6 +490,105 @@ export const Horizontal = meta.story({
...
+            <Text slot="label">Overnight</Text>
+            <Text slot="description">Next business day</Text>
+          </SelectList.Option>
+        </SelectList>
+      </div>
+      <div data-testid="medium-container" className="max-w-125">
+        <SelectList
+          {...args}
+          label="Shipping speed (medium container)"
+          description="500px parent items should never overflow the container."
```
sarahgm: 🔴 This 500px “medium” example says items *“should never overflow”*, implying they stay side by side. But with the shipped 640px breakpoint, a 500px container actually flips to a vertical stack. The `play` test only checks width, so it won't catch this mismatch. This points to 448px being the intended breakpoint — please reconcile, and add a `flexDirection` assertion here to lock in the expected behavior.

OsamaAbdellateef:
Closing in favour of #5491 — this branch accidentally included a merge of `beta-release` which caused unrelated test failures in DatePicker, Dialog, Drawer, and TagField. The new PR contains only the SelectList changes rebased cleanly onto `main`.

sarahgm (approved):

PR #5490, opened 2026-06-01
fix(DSTSUP-255/257): respect min/maxValue and harden Calendar year/month pickers

sarahgm:
# Description

This branch fixes three things in the Calendar and RangeCalendar year/month pickers.

**1. The year picker now respects `minValue` / `maxValue`.**
Before, it always showed 41 years (the focused year ±20) and just greyed out the ones outside the allowed range. With `minValue` set, it opened on a list of disabled past years you had to scroll past first. Now both pickers get their lists from react-aria's `CalendarYearPicker` / `CalendarMonthPicker`, so out-of-range years are not shown at all (the month picker still shows all twelve months).

**2. The picker opens centered on the selected option.**
Before, the selected year/month appeared at the bottom of the list when the picker opened. Now it is scrolled to the middle.

**3. RangeCalendar month/year dropdown no longer breaks a range selection.**
This is a follow-up to DSTSUP-257. If you had started picking a range and then used the dropdown to jump to another month or year, that tap could wrongly finish the range. The dropdown now only changes the view, and picking an option still works on touch.

Closes DSTSUP-255

## Screenshots / Preview

| Before | After |
| ------ | ----- |
| Selected year opened at the bottom of the list | Selected year opens in the middle |

## Test Instructions

1. Open the `Calendar` / `RangeCalendar` stories in Storybook.
2. Set a `minValue` (e.g. `YearDropdownWithMinMax`) and open the year picker — only in-range years should appear.
3. Open the year/month picker and confirm the selected option is centered, not at the bottom.
4. In `RangeCalendar`, click a start date, open the month dropdown, pick another month — the half-finished range must NOT be committed, and the grid should switch months.
5. Repeat step 4 with touch (DevTools mobile emulation) — picking a month/year should still work.

## Breaking Changes

No

## Checklist


packages/components/src/Calendar/ListBox.tsx
```diff
@@ -17,39 +18,62 @@ interface ListBoxProps<T> {
...
+      const listRect = list.getBoundingClientRect();
+      const selectedRect = selected.getBoundingClientRect();
+      // Center within the list's own scroll area only — never scroll ancestors/the page.
+      list.scrollTop +=
+        selectedRect.top -
+        listRect.top -
+        (list.clientHeight - selected.clientHeight) / 2;
+    });
+    return () => cancelAnimationFrame(raf);
+  }, []);
```
sebald, line 54: **Centering effect — works, but consider `useLayoutEffect` to kill the one-frame flash.**

I traced the timing against react-aria: RAC scrolls the focused item from a passive `useEffect` that schedules its *own* `requestAnimationFrame` (`useSelectableCollection`), so the `useEffect` + `rAF` here is a deliberate way to queue *after* RAC and win. That's correct.

The cost is a one-frame flash: the dropdown's first paint shows the list at `scrollTop: 0` (selected item possibly off-screen), then it jumps to centered on the next frame — on every open.

A synchronous `useLayoutEffect` (set `scrollTop` before paint, no `rAF`) would remove the flash *and* the rAF bookkeeping, and it survives RAC's later scroll: once the item is centered it's fully visible, and RAC's `scrollIntoView` is conditional (`shouldScrollBlock = scrollAreaTop < scrollPortTop || scrollAreaBottom > scrollPortBottom`) so it no-ops when the item is already in view (true in test env too). `aria-selected` comes from `selectedKeys` at render, so the node + layout are available in a layout effect.

Trade-off: the current `rAF` is slightly more defensive (wins even if RAC ever force-recenters); `useLayoutEffect` is simpler/flash-free but leans on "RAC's scroll stays conditional". Either is fine — not a defect — but if you switch, add a one-line note about the dependency. Worth a quick browser check.

packages/components/src/Calendar/ListBox.tsx
```diff
@@ -17,39 +18,62 @@ interface ListBoxProps<T> {
...
         return (
           <RACListBoxItem
             key={index}
             id={String(index)}
-            textValue={String(format(item, index))}
+            textValue={String(label)}
             isDisabled={disabled}
-            aria-current={selected ? 'true' : undefined}
-            aria-label={`${format(item, index)}${
+            aria-label={`${label}${
```
sebald: A few small cleanups clustered in this component (all optional, none blocking):

1. **Redundant label suffix (this line).** With `selectionMode="single"` + `selectedKeys`, RAC already exposes `aria-selected`, and `isDisabled` already sets `aria-disabled`. Appending `' selected'` / `' not selectable'` to `aria-label` can double-announce. Since the old `aria-current` was dropped, consider relying solely on RAC's state and removing the suffix.

2. **`dataTestid` is now dead weight (L10, L20, L60).** Stories were migrated to `getByRole('listbox', { name: 'year'|'month' })`, so `data-testid` has no remaining story consumer — only `Calendar.test.tsx` still uses `getByTestId('monthOptions'/'yearOptions')`, which contradicts our "prefer accessible queries" rule. If you migrate that test to `getByRole`, the whole `dataTestid` prop can be deleted from `ListBoxProps`, `ListBox`, `MonthListBox`, `YearListBox`. (While there: `Calendar.test.tsx:198` has a typo `queryByTestId('yearhOptions')` — extra "h" — making that assertion a silent no-op, and the `scrollIntoView` mock at L172/186 is now stale since this code uses `scrollTop`.)

3. **`cn()` around a single string literal is a no-op (L57–59).** Replace with the plain `className="grid h-full max-h-[300px] …"`.

4. **Unused `index` in the callback signatures (L13–16).** Neither `MonthListBox` nor `YearListBox` uses the second `index` arg in `isDisabled`/`isSelected`/`onSelect`/`format`. Dropping it from the public signature removes noise (the internal `.map(item, index)` keeps its own index for keys).

5. **`autoFocus={true}` → `autoFocus`** (JSX boolean shorthand), to match the codebase style. Minor.

packages/components/src/Calendar/MonthListBox.tsx
```diff
@@ -1,43 +1,44 @@
...
+          dataTestid="monthOptions"
+          ariaLabel={ariaLabel}
+          items={items}
+          isDisabled={item => item.id < minMonth || item.id > maxMonth}
+          isSelected={item => item.id === value}
+          onSelect={item => {
+            onChange(item.id);
+            setSelectedDropdown(undefined);
+          }}
+          format={item => item.formatted.substring(0, 3)}
```
sebald: **`.substring(0, 3)` is redundant here and a latent i18n bug.** `CalendarMonthPicker` already emits abbreviated names (its `format` defaults to `'short'`, e.g. `"Jan"`), so truncating to 3 chars on top of that is a no-op for English but can mangle non-Latin / longer localized abbreviations (some locales' "short" month names aren't 3 chars). Suggest dropping the `.substring(0, 3)` and relying on the picker's format — pass `format="short"` explicitly on `<CalendarMonthPicker>` if you want it pinned:

```tsx
format={item => item.formatted}
```

(Carried over from the old `useFormattedMonths` code, not introduced here — but this is the natural spot to fix it.)

packages/components/src/Calendar/YearListBox.tsx
```diff
@@ -1,40 +1,56 @@
...
 
-  const minYear = minValue ? minValue.year : -Infinity;
-  const maxYear = maxValue ? maxValue.year : Infinity;
+  // Localized accessible name, matching what react-aria's CalendarYearPicker
+  // would expose (e.g. "year"/"Jahr"). Computed inline — this only mounts when
+  // the dropdown opens, so memoizing the lookup isn't worth a hook.
+  const { locale } = useLocale();
+  const ariaLabel =
+    new Intl.DisplayNames(locale, { type: 'dateTimeField' }).of('year') ??
+    'year';
```
sebald, line 44: **Hand-rolling the year list is the right call — verified.** I read RAC 1.18's `useCalendarYearPicker`: its loop is `while (date.compare(maxDate) < 0)` with `maxDate` clamped to exactly `maxValue`, so the `maxValue` year is dropped when the iterating date lands on it; and its `visibleYears` cap (default 20) means it structurally can't render an arbitrary in-range span. So `CalendarYearPicker` genuinely isn't usable for a bounded calendar today. 👍 I've written up a detailed upstream fix spec so we can eventually delete this workaround and adopt the component.

Two small notes:

- **Label duplication (this block).** `new Intl.DisplayNames(locale, { type: 'dateTimeField' }).of('year')` is copied verbatim from RAC's internals, so the year label can drift from the month label (which comes "for free" from `CalendarMonthPicker`). It's a standard platform API (fine to keep inline), but a tiny shared helper — `dateFieldLabel(locale, 'year' | 'month')` — would unify both. Optional.

- **`isDisabled={() => false}` (L52).** Every rendered year is in range, so if `ListBoxProps.isDisabled` were made optional with a `() => false` default in `ListBox`, this consumer could simply omit the prop and the intent ("nothing disabled") becomes self-documenting.

packages/components/src/RangeCalendar/RangeCalendar.tsx
```diff
@@ -140,37 +140,24 @@ const _RangeCalendar = <T extends DateValue>({
...
     const stop = (event: PointerEvent) => {
-      if ((event.target as Element)?.closest('[role="option"]')) return;
-      event.stopPropagation();
+      if (node.contains(event.target as Node | null)) event.stopPropagation();
     };
-    node.addEventListener('pointerup', stop);
-    return () => node.removeEventListener('pointerup', stop);
+    ownerDocument.addEventListener('pointerup', stop);
+    return () => ownerDocument.removeEventListener('pointerup', stop);
   }, []);
```
sebald, line 159: **Verified this is correct and minimal — keep it.** I checked the relevant react-aria internals:

- `useRangeCalendar` commits on a **`window`** `pointerup` when `isFocusWithin(calendarRoot) && (!nodeContains(calendarRoot, target) || !target.closest('button, [role="button"]'))`. The overlay lives *inside* the calendar root and autoFocus moves focus into it, and our items are `role="option"` (not buttons) → it commits. That's the bug.
- The clean RAC-idiomatic fix (overlay *outside* the calendar root, so `isFocusWithin`/`nodeContains` both go false) **isn't possible here** — `MonthListBox`/`YearListBox` depend on `CalendarStateContext`, which only exists inside `<AriaRangeCalendar>`. So the overlay must stay inside, and some interception is unavoidable.
- `usePress`'s touch press-end fallback listens on **`document`** (`getOwnerDocument`), while the range-commit listens on **`window`**. Listening on `document` + `stopPropagation()` (not `stopImmediatePropagation`) lets *both* usePress and this guard fire, then starves the `window` handler — which is exactly why touch selection keeps working (DSTSUP-257). A React-synthetic `onPointerUp` would stop the bubble before `document` and re-break touch, so the native document listener is the right mechanism.

Only suggestion: extend the comment to name the dependency explicitly — *"usePress listens on `document`, range-commit on `window`; we stop at `document` so both fire but `window` never does"* — so the next reader knows why `document` (not the node, not `window`) is the correct target.

sebald (commented):
Reviewed with a focus on the new react-aria month/year picker components and idiomatic usage. **Nothing blocking** — the change is correct and green locally (Calendar 16/16, RangeCalendar 11/11, unit 22/22, `typecheck:only` clean), and I verified the two workarounds against react-aria 1.18 source.

### Verified ✅
- **Year picker hand-rolls instead of `CalendarYearPicker` — justified.** RAC's `useCalendarYearPicker` drops the `maxValue` year (half-open `< 0` loop on a clamped `maxValue`) and caps at `visibleYears`, so it can't expose an arbitrary in-range span. I've drafted a detailed upstream fix spec; once it lands we can delete the workaround. No action here.
- **`min`/`max` for the month picker works correctly** — confirmed against `MonthSelectionWithMinMax` and the disabling logic (out-of-range months disabled in place; the picker only navigates the view).
- **RangeCalendar `pointerup` guard (D2)** and the **picker centering effect (D1)** — both traced through react-aria internals; see inline notes.

### Inline comments (all optional / non-blocking)
- `ListBox.tsx` — `useLayoutEffect` to remove the one-frame centering flash; redundant `aria-label` suffix vs `aria-selected`; dead `dataTestid` (+ migrate `Calendar.test.tsx` off `getByTestId`, fix the `yearhOptions` typo, drop stale `scrollIntoView` mock); `cn()` no-op; unused `index` args; `autoFocus` shorthand.
- `MonthListBox.tsx` — drop the redundant `.substring(0, 3)` (latent i18n truncation).
- `YearListBox.tsx` — optional shared label helper; make `isDisabled` optional so this can omit `() => false`.
- `RangeCalendar.tsx` — keep the guard (it's correct & minimal); just sharpen the comment to name the `document`-vs-`window` listener dependency.

Highest-value, lowest-risk batch if you want to fold a few in: the `MonthListBox` `.substring` fix, the `dataTestid`/test cleanup, and the `cn()` no-op. Nice work on the regression coverage for the touch + dropdown-navigation cases. 🙌

packages/components/src/Calendar/ListBox.tsx
sarahgm: Good catch on the flash. I went with a hybrid: \`useLayoutEffect\` sets \`scrollTop\` before first paint (no scrollTop:0 flash), plus a \`requestAnimationFrame\` that re-asserts the center afterwards. The browser check showed pure \`useLayoutEffect\` alone is not enough here — RACs \`useSelectableCollection\` re-scrolls the focused item to the nearest edge *after* the layout effect runs, so the year ended up ~130px off-center (caught by the `YearPickerScrollsToFocused` story test). The rAF wins that race; the layout effect kills the flash. Added a note explaining the dependency. Fixed in ca1453c2c.

packages/components/src/Calendar/MonthListBox.tsx
sarahgm: Done — dropped \`.substring(0, 3)\` and rely on \`item.formatted\` (the picker already emits \`short\` names). Added a comment noting the truncation was a latent i18n bug. ca1453c2c.

packages/components/src/RangeCalendar/RangeCalendar.tsx
sarahgm: Extended the comment to name the mechanism explicitly: usePress listens on \`document\`, range-commit on \`window\`; we stop at \`document\` so both fire but \`window\` never does. Kept the guard as-is. ca1453c2c.

packages/components/src/Calendar/ListBox.tsx
sarahgm: All five folded in (ca1453c2c):

1. **aria-label suffix** — removed; relying on RACs \`aria-selected\`/\`aria-disabled\`. Updated `MonthSelectionAccessibility` to assert via \`toHaveAccessibleName\` + the state attributes instead of the suffix.
2. **dead \`dataTestid\`** — deleted from \`ListBoxProps\`/\`ListBox\`/\`MonthListBox\`/\`YearListBox\`. Migrated \`Calendar.test.tsx\` to \`getByRole("listbox", { name: "month"|"year" })\`, fixed the \`yearhOptions\` typo, and dropped the now-stale \`scrollIntoView\` mocks.
3. **\`cn()\` no-op** — replaced with the plain \`className\` string.
4. **unused \`index\` args** — dropped from the \`isDisabled\`/\`isSelected\`/\`onSelect\`/\`format\` signatures (the internal \`.map\` keeps its own index for keys).
5. **\`autoFocus\`** — switched to the boolean shorthand.

packages/components/src/Calendar/YearListBox.tsx
sarahgm: Thanks for verifying the upstream loop behavior — agreed on keeping the hand-rolled list until the \`CalendarYearPicker\` fix lands.

- **\`isDisabled={() => false}\`** — done: made \`ListBox\`s \`isDisabled\` optional with a \`() => false\` default, so \`YearListBox\` now omits the prop and the intent is self-documenting.
- **shared label helper** — left the inline \`Intl.DisplayNames\` for now (the month label still comes free from \`CalendarMonthPicker\`); happy to extract \`dateFieldLabel(locale, ...)\` in a follow-up if we want both unified. ca1453c2c.

sebald (approved):
Re-reviewed `ca1453c2` against my earlier feedback — all five threads addressed faithfully and resolved. ✅

- **Centering (D1):** the `useLayoutEffect` (pre-paint) + `requestAnimationFrame` (re-assert) hybrid is correct, and the `scrollTop +=` keeps the double `center()` idempotent so the rAF call can't mis-scroll. Better than my original suggestion — pure layout-effect alone loses the race with RAC's `useSelectableCollection`.
- **a11y suffix** removed in favor of RAC's `aria-selected`/`aria-disabled`; story now asserts `toHaveAccessibleName`.
- **`dataTestid`** deleted everywhere; `Calendar.test.tsx` migrated to `getByRole('listbox', …)`, `yearhOptions` typo + stale `scrollIntoView` mocks fixed.
- **`.substring(0,3)`** i18n bug dropped; **`cn()` no-op**, **unused `index` args**, **`autoFocus` shorthand**, and the **optional `isDisabled` default** all done.
- **RangeCalendar** `document`-vs-`window` comment expanded; guard unchanged (correct & minimal).

All 18 checks green. Two acknowledged follow-ups remain (adopt `CalendarYearPicker` once the upstream span fix lands; optionally extract a shared `dateFieldLabel` helper) — neither blocking. Nice work. 🙌

PR #5492, opened 2026-06-03
fix(DST-1482): make the width prop size field components again

sarahgm:
# Description

Setting the `width` prop on Marigold field components (`<Select>`, `<TextField>`, `<NumberField>`, …) had no visible effect in `18.0.0-beta.2` — the field shrink-wrapped to its content, so consumers had to wrap it in an extra fixed-width `div`.

`FieldBase` sets the `--field-width` CSS variable for its child field element to consume via `w-(--field-width)`, but DST-901 registered the variable with `@property { inherits: false }`. With inheritance off, the value set on `FieldBase` never reached the child, so `w-(--field-width)` resolved to its guaranteed-invalid initial value and `width` fell back to `auto`. This PR registers `--field-width` with `inherits: true`, restoring the intended **parent→child handoff**. The same-element layout vars (`--width`, `--max-width`, `--height`, `--container-width`) keep their non-inheriting leak protection. A single CSS change fixes every affected field component (Select, TextField, TextArea, NumberField, SearchField, ComboBox, Autocomplete, …).

**Behavior change:** fields that previously sized to their content now honor a fixed/keyword `width` directly — no wrapper needed. Fractions (`width="1/2"`) and the full-width default are unchanged.

Also documents that numeric `width` values are spacing-scale **tokens, not pixels** (`width={64}` → `calc(var(--spacing) * 64)` ≈ 16rem/256px) on the prop JSDoc and the Form Fields foundations page, and adds computed-width regression guards.

Closes DST-1482

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. `pnpm build:themes`, then `pnpm sb`.
2. Render `<Select width={64}>` — confirm the trigger is ~256px (16rem) and no longer content-sized; no wrapper required.
3. Spot-check `TextField`, `TextArea`, `NumberField`, `SearchField`, `ComboBox`, `Autocomplete` with a fixed `width`, with `width="1/2"` (still responsive), and with no width (full-width default).
4. `pnpm test:sb` — the new `FixedWidth` play stories on Select/TextField assert the rendered field width equals 16rem.
5. `pnpm test:unit` — existing class-based assertions still pass.

## Breaking Changes

No

## Checklist

- [x] Visual regression tests updated (for UI changes)

### DIFF
--- a/packages/components/src/Checkbox/CheckboxGroup.tsx
+++ b/packages/components/src/Checkbox/CheckboxGroup.tsx
@@ -63,6 +63,9 @@ export interface CheckboxGroupProps
   /**
    * Sets the width of the field. You can see allowed tokens here: https://tailwindcss.com/docs/width
    *
+   * Numeric/scale values are spacing-scale tokens, not pixels: `width={64}`
+   * resolves to `calc(var(--spacing) * 64)` ~= 16rem (256px), not 64px.
+   *
    * @default full
    */
   width?: WidthProp['width'];
--- a/packages/components/src/Checkbox/CheckboxGroup.tsx
+++ b/packages/components/src/Checkbox/CheckboxGroup.tsx
@@ -65,7 +65,6 @@ export interface CheckboxGroupProps
    *
    * Numeric/scale values are spacing-scale tokens, not pixels: `width={64}`
    * resolves to `calc(var(--spacing) * 64)` ~= 16rem (256px), not 64px.
-   *
    * @default full
    */
   width?: WidthProp['width'];
--- a/packages/components/src/TextField/TextField.stories.tsx
+++ b/packages/components/src/TextField/TextField.stories.tsx
@@ -219,14 +219,13 @@ export const FixedWidth = meta.story({
   tags: ['component-test'],
   args: {
     label: 'Name',
-    description: '',
     width: 64,
   },
   render: args => <TextField {...args} placeholder="Type" />,
   play: async ({ canvas, step }) => {
     await step('Input width matches the requested scale value', async () => {
       // The --field-width consumer is the `group/input` wrapper of the input.
-      const wrapper = canvas.getByRole('textbox').parentElement!;
+      const wrapper = canvas.getByRole('textbox').closest('.group\\/input')!;
       const rem = parseFloat(
         getComputedStyle(document.documentElement).fontSize
       );
--- a/packages/components/src/TextField/TextField.stories.tsx
+++ b/packages/components/src/TextField/TextField.stories.tsx
@@ -1,4 +1,5 @@
 import { useState } from 'react';
+import { expect } from 'storybook/test';
 import preview from '.storybook/preview';
 import { Button } from '../Button/Button';
 import { Form } from '../Form/Form';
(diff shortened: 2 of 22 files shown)


packages/components/src/TextField/TextField.stories.tsx
```diff
@@ -207,3 +208,34 @@ export const WithFormValidation = meta.story({
...
+  args: {
+    label: 'Name',
+    description: '',
+    width: 64,
+  },
+  render: args => <TextField {...args} placeholder="Type" />,
+  play: async ({ canvas, step }) => {
+    await step('Input width matches the requested scale value', async () => {
+      // The --field-width consumer is the `group/input` wrapper of the input.
+      const wrapper = canvas.getByRole('textbox').parentElement!;
```
aromko: 💡 This relies on `getByRole('textbox').parentElement!` (non-null assertion + DOM-structure coupling) to reach the `group/input` wrapper that carries `w-(--field-width)`. Solid as a regression guard, but if the `Input` markup ever gains or loses a nesting level this would silently measure the wrong node. A more resilient lookup (e.g. `closest`-based) or an inline note would harden it. Optional.

packages/components/src/TextField/TextField.stories.tsx
```diff
@@ -207,3 +208,34 @@ export const WithFormValidation = meta.story({
...
+ * Regression guard for DST-1482: a fixed `width` must size the input element
+ * itself, not just the FieldBase wrapper. `width={64}` maps to the spacing
+ * scale, i.e. `calc(var(--spacing) * 64)` = 16rem (256px at default root
+ * font-size), and must drive the input's layout without an outer wrapper.
+ */
+export const FixedWidth = meta.story({
+  tags: ['component-test'],
+  args: {
+    label: 'Name',
+    description: '',
```
aromko: 💡 `description: ''` is passed in args but unused by the assertion — can be dropped for clarity.

packages/components/src/Checkbox/CheckboxGroup.tsx
```diff
@@ -63,6 +63,9 @@ export interface CheckboxGroupProps
   /**
    * Sets the width of the field. You can see allowed tokens here: https://tailwindcss.com/docs/width
    *
+   * Numeric/scale values are spacing-scale tokens, not pixels: `width={64}`
```
aromko, line 66: 💡 Minor cosmetic: the new "spacing-scale tokens" paragraph here sits with a surrounding blank line before `@default`, whereas the other field components (Autocomplete, DateField, …) insert it directly before `@default`. Renders fine either way — just an inconsistency.

aromko (commented):
## Automated Code Review

Clean, well-scoped fix. The root cause is correctly diagnosed: `FieldBase` sets `--field-width` on its root element, while every field consumer reads it as a *descendant* via `w-(--field-width)` (`Input.tsx`, `Select.tsx`, `TextArea.tsx`, `DateInput.tsx`, `TimeField.tsx`, `TagField.tsx`). With `@property … { inherits: false }` the value never reached those descendants, so width fell back to `auto`. Flipping `--field-width` to `inherits: true` restores the parent→child handoff, while the same-element layout vars keep their non-inheriting leak protection. Verified the test math: `width={64}` → `calc(var(--spacing) * 64)` = 16rem (256px).

Both DST-1482 acceptance criteria are met, all CI is green, coverage is intact (changed component files at 100%), and visual regression tests passed on the current head. No blocking issues — the three inline notes below are all optional polish.

---

packages/components/src/TextField/TextField.stories.tsx
sarahgm: Done in 0bb730ccf — switched to `closest('.group\\/input')` so the lookup targets the `group/input` wrapper directly instead of relying on the DOM nesting level.

packages/components/src/TextField/TextField.stories.tsx
sarahgm: Done in 0bb730ccf — dropped the unused `description` arg.

packages/components/src/Checkbox/CheckboxGroup.tsx
sarahgm: Done in 0bb730ccf — removed the blank line so the JSDoc matches the other field components.

aromko (approved):

PR #5493, opened 2026-06-03
chore(DST-1364): migrate ListBox item label/description styling off descendant selectors

sarahgm:
# Description

Moves `ListBox.Item`'s `label` / `description` styling off the fragile `[&_[slot=description]]:…` descendant selector on `item.base` and onto first-class `label` / `description` theme entries. `ListBox.Item` now injects those classNames into react-aria's `TextContext` via an `ItemChildren` provider that **merges** (rather than replaces) the parent slot config, so RAC's `aria-describedby` wiring on the description `id` is preserved for screen readers.

The shared merge logic (and its `SlottedContextValue` type) is extracted into a `useMergedTextSlots` hook, also consumed by `SelectList.Option`. Because `Select.Option`, `ComboBox.Option`, and `Autocomplete.Option` re-export `ListBox.Item`, they all benefit automatically.

No public API change in `@marigold/components`. Visually near-identical — `description` keeps the same `col-start-2 row-start-2 text-xs text-secondary` styling, just applied via the slot instead of a descendant selector; it additionally sets `font-normal` now (parity with `SelectList`), which only matters if a description previously inherited a heavier weight. VRT is not required here (PR targets `beta-release`).

Closes DST-1364

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. `pnpm install && pnpm typecheck:only` (already green locally).
2. `pnpm test:unit` → `ListBox.test.tsx` asserts the description slot gets the theme classNames directly and that `aria-describedby` still points at the rendered description's `id`.
3. Open Storybook → `Components/ListBox/WithDescription` and confirm label/description render unchanged.

## Breaking Changes

Type-level only: the `Theme` type in `@marigold/system` now requires `label` and `description` keys on the `ListBox` record, so external/custom themes implementing `ListBox` must add these entries (covered by a `minor` changeset for `@marigold/system`).

## Checklist


packages/system/src/types/theme.ts
```diff
@@ -140,7 +140,13 @@ export type Theme = {
...
     Link?: ComponentStyleFunction<string, string>;
     ListBox?: Record<
-      'container' | 'list' | 'item' | 'section' | 'header',
+      | 'container'
+      | 'list'
+      | 'item'
+      | 'section'
+      | 'header'
+      | 'label'
+      | 'description',
```
aromko, line 149: ⚠️ **Missing `@marigold/system` changeset.**

This adds `label`/`description` as **required** keys on the `ListBox` theme record — a change to the public `Theme` type exported from `@marigold/system`. The PR ships changesets for `@marigold/components` and `@marigold/theme-rui`, but none for `@marigold/system`, so this package won't get a version bump.

Since this is a breaking type change for any external/custom theme implementing `ListBox` (they'll now fail typecheck without these keys), please add a `@marigold/system` changeset entry — arguably `minor` rather than `patch`.

themes/theme-rui/src/components/ListBox.styles.ts
```diff
@@ -30,9 +30,12 @@ export const ListBox: ThemeComponent<'ListBox'> = {
       'disabled:cursor-not-allowed disabled:text-disabled',
       'focus-visible:ui-state-focus outline-none focus-visible:z-1 transition-[border,color]',
       'cursor-default data-selection-mode:cursor-pointer',
-      '[&_[slot=description]]:col-start-2 [&_[slot=description]]:row-start-2 [&_[slot=description]]:text-xs [&_[slot=description]]:text-secondary',
     ],
   }),
+  label: cva({ base: '' }),
+  description: cva({
+    base: 'col-start-2 row-start-2 text-xs text-secondary font-normal',
```
aromko, line 37: 💡 `font-normal` is new here — the old descendant selector was just `…text-xs text-secondary`. This is intentional per DST-1364 and brings parity with `SelectList.styles.ts`, so no change needed to the code. Just flagging that the changeset/PR body state "No … visual change," which is slightly inaccurate if any description previously inherited a heavier weight. Consider softening that wording.

aromko (commented):
## Automated Code Review

### Overview
| Field | Value |
|-------|-------|
| Branch | `chore/DST-1364-listbox-slot-styling` → `beta-release` |
| Files Changed | 9 (+144 / −28) |
| CI | ✅ All green (Typecheck, Lint, Unit, Storybook, Coverage) |
| Visual Regression | Not needed (targets `beta-release`) |
| Linked ticket | [DST-1364](https://reservix.atlassian.net/browse/DST-1364) — In Review |

### What's good
- ✅ **Accessibility preserved correctly.** The merge (spread parent slot, then override `className`) keeps RAC's `descriptionProps.id`, so `aria-describedby` wiring survives. Both the new unit test and the story `play` function guard this — exactly the regression DST-1364 called out.
- ✅ **Clean DRY refactor.** `useMergedTextSlots` deduplicates the merge logic and its `SlottedContextValue` type across `ListBox.Item` and `SelectList.Option`, with clear docblocks on *why* merge (not replace) is required.
- ✅ **Follows conventions** — imports stories in tests, accessible queries, no `className`/`style` exposure, no dangling imports after the SelectList refactor.
- ✅ **Parity with SelectList** — the new `description` style now matches `SelectList.styles.ts`.

### Issues Found

**Warnings (Should Fix):**
- `packages/system/src/types/theme.ts` — **no changeset for `@marigold/system`.** The public `Theme` type changed: `label`/`description` are now **required** keys on the `ListBox` record, but the PR only bumps `@marigold/components` and `@marigold/theme-rui`. External themes implementing `ListBox` will fail typecheck without these keys — worth a `@marigold/system` changeset (arguably `minor` rather than `patch`). See inline comment.

**Suggestions (Nice to Have):**
- `themes/theme-rui/src/components/ListBox.styles.ts` — `font-normal` is a (minor) visual change vs. the old selector. Intentional per DST-1364 + SelectList parity, but the "no visual change" wording in the changeset/PR body is slightly misleading. See inline comment.
- `ListBox.test.tsx` — the assertion could also cover the grid-position classes (`col-start-2 row-start-2`) that the migration actually moved off the descendant selector. Optional.

### Recommendation
**Comment** (close to Approve). Implementation is correct, well-tested, and faithful to the ticket; accessibility is properly preserved. The only substantive item is the missing `@marigold/system` changeset — quick to add.

---

packages/system/src/types/theme.ts
sarahgm: Done in 43da02012 — added `'@marigold/system': minor` to the existing changeset (`listbox-label-description-slots.md`, matching the precedent from `card-slot-configuration-pattern.md`) and documented the required `label`/`description` keys in the changeset body and the PR's Breaking Changes section.

themes/theme-rui/src/components/ListBox.styles.ts
sarahgm: Good catch — softened the wording in the changeset and PR body in 43da02012: it now says "visually identical except `description` explicitly sets `font-normal` (parity with `SelectList`)" instead of claiming no visual change.

aromko (approved):

PR #5494, opened 2026-06-03
feat(DST-1483): remove ActionButton in favor of a slot-aware Button (rename ActionGroup → ButtonGroup)

sebald:
# Description

Removes the **beta-only** `<ActionButton>` in favor of a **slot-aware `<Button>`**, and renames `<ActionGroup>` → `<ButtonGroup>`. A single Marigold-owned `ButtonContext` now drives the whole subsystem (mirroring the existing `ToggleButtonGroup → ToggleButtonContext → ToggleButton` trio), so users write `<Button>` everywhere and it adapts automatically inside a button container.

`ActionButton.styles` was a strict subset of `Button.styles` (its `default` ≡ Button's `ghost`), so the visual unification is essentially free. Since the standalone `<ActionButton>` never shipped to a stable consumer (it existed only on the beta line + docs), removing it has **zero stable-consumer impact** and needs no migration notes.

Closes DST-1483

## What changed

- **`<Button>` is slot-aware** — consumes a Marigold-owned `ButtonContext` (a distinct `createContext`, **not** RAC's) via `useContextProps`; still forwards `slot` to RAC's `<Button>` so `slot="close"`/`"increment"`/`"decrement"` keep working.
- **Uniform precedence: a local prop always wins** over the container for `variant`/`size`/`disabled` (drops the old `ActionGroup` `size`-group-wins outlier).
- **`<ActionGroup>` → `<ButtonGroup>`** — reads the incoming `ButtonContext` to position itself, then re-provides a fresh context to children (cascades `variant: 'secondary'` when unset, the same baseline as a standalone `<Button>`). `ActionGroupContext` is eliminated.
- **`<Panel.Header>`** publishes one `ButtonContext` (`variant: 'ghost'` + `size: 'icon'` + `[grid-area:actions]`), preserving the old `ActionButton` look for a bare `<Button>`.
- **`<ActionMenu>`** kept (public name unchanged); its trigger is a slot-aware `<Button>` that **inherits** the cascade rather than hardcoding a variant — `secondary` standalone (the `<Button>` baseline), `ghost` inside `<Panel.Header>`/`<SelectList.Option>`/`<ButtonGroup>`. A local `variant` still wins.
- **`LinkButton`** reads the single `ButtonContext`; always uses `Button` styles, so it tracks the same cascade as `<Button>`.
- Theme: deleted `ActionButton.styles.ts`, removed the `ActionButton` theme type + theme/appearances entries, added `ButtonGroup` (shares Button's vocabulary, narrowed to cluster-safe variants), regenerated `appearances.ts`.
- **Docs:** `action-button` folded into the `Button` page (new "Surface actions" section covering ghost chrome + slot-awareness); `action-group` → new `button-group` page; updated `link-button`, `panel`, `action-menu`, `menu`, `component-principles`, the inventory example, `meta.json`; regenerated registry/manifest/types. The `button-group` page is marked **beta**, opens with plain-language purpose copy, lists `<ActionMenu>` as an interchangeable anatomy action, narrows the appearance picker to cluster-safe variants (`secondary`/`ghost`/`destructive-ghost`, dropping `primary`/`destructive`/`link`), rewrites the slot-aware copy to be component-agnostic, restructures Usage into scenarios + cross-cutting guidance with demos, sets icon-only demos to `size="icon"`, and drops the redundant `### ButtonGroup` Props heading to match the other single-component pages. Also refreshes the future-dated release blog (`release-2026-06-30.mdx`) to the final `Button`/`ButtonGroup` naming (the versioned `CHANGELOG.md`/`release.mdx` beta entries are left as accurate per-beta history; the rename is documented in this PR's changeset). `typecheck:docs` passes.
- **Cleanup:** trimmed verbose inline comments across the Button subsystem (`Button`, `ButtonGroup`, `LinkButton`, `ActionMenu`, `Panel.Header`, `SelectList.Option`, `ResetButtonContext`), keeping only load-bearing rationale — no behavior change.

## 🔬 Context-leak fix (the core review concern)

React context flows through portals, so a `<ButtonGroup>`/`<Panel.Header>` cascade in the React tree **above** an overlay's trigger would otherwise leak into the overlay's portaled buttons (`slot="close"`, `Dialog.Actions` footers) — e.g. an `<ActionMenu>` placed in a `<Panel.Header>` leaking into its mobile Tray's close button. The separate Marigold context does **not** prevent this (it's the Marigold context reaching a Marigold `<Button>`); `slot="close"` can't opt out either (plain-object context ignores the slot arg).

**Fix:** a small `<ResetButtonContext>` wraps the content root of `Popover`, `Modal`, `Tray`, and `Drawer`, re-rooting `ButtonContext` to empty. A `<ButtonGroup>`/`<Panel.Header>` *inside* an overlay re-establishes its own context below the reset. Verified empirically (portal propagation reproduced, then closed by the reset) and covered by a regression test.

## Test Instructions

1. `pnpm sb` → check `Components/Button`, `Components/ButtonGroup`, `Components/LinkButton`, and `Panel` "With Header Actions": buttons adapt inside groups/headers; local props win.
2. Open an `<ActionMenu>` inside a `<Panel.Header>` on a small screen → its Tray close button is a normal button (no `[grid-area:actions]`/icon leak).
3. A standalone `<ActionMenu>` renders as a `secondary` (bordered) kebab; inside a ghost group/header it reads as `ghost`.
4. `pnpm test:unit packages/components/src/{Button,ButtonGroup,LinkButton,Panel,Menu,Tray,Drawer,Overlay}` → green.
5. `pnpm --filter @marigold/docs build:assets && pnpm typecheck:docs` → green.

## Breaking Changes

Yes (beta line only; **zero stable-consumer impact**).

- `<ActionButton>` → `<Button>` (`default` variant maps to `variant="ghost"`)
- `<ActionGroup>` → `<ButtonGroup>`
- `ActionButtonContext` / `ActionGroupContext` → `ButtonContext`

## ⚠️ Remaining work

- [ ] Refresh Chromatic/VRT snapshots for Button, ButtonGroup, Panel header, LinkButton, ActionMenu (deferred until after reviews).

> **Resolved during this PR — standalone `<ButtonGroup>` gap:** investigation showed `ActionGroup`/`Panel.styles.actions` provided **no** gap at all (clusters rendered flush, both standalone and in-panel). `<ButtonGroup>` now owns a structural `flex gap-1` layout (orientation-aware), merged with any container positional className via tailwind-merge.

> **Resolved during this PR — `<ButtonGroup>` default variant:** the zero-config cascade is now `secondary` (the same baseline as a standalone `<Button>`) rather than `ghost`, since `secondary` is the most common neutral default. `primary` is reserved for page-level commitments and form submits, so it is excluded from the ButtonGroup appearance picker. Low-emphasis surfaces opt into `ghost` explicitly: `<Panel.Header>` and `<SelectList.Option>` override the cascade, and the row-actions demo sets `variant="ghost"` to recede.

> **Resolved during this PR — `<ActionMenu>` trigger regression:** the trigger initially hardcoded `variant ?? 'ghost'`, which (a) changed the **standalone** look from `secondary` (its pre-unification default via the `Menu` button slot) to `ghost`, and (b) defeated the variant cascade, leaving it `ghost` even inside a non-ghost `<ButtonGroup>`. Fixed by passing `variant` straight through (`variant={variant}`) so the trigger `<Button>` resolves it via `ButtonContext`: `secondary` standalone (restored), the parent's variant in context. Ghost inside `<Panel.Header>`/`<SelectList.Option>` is preserved by those parents' own cascade.

## Checklist

- [x] Accessibility reviewed (Toolbar semantics preserved; RAC `ButtonContext` untouched)


docs/content/components/actions/link-button/index.mdx
```diff
@@ -48,16 +48,15 @@ In contrast, a regular button is meant for performing an action on the current p
...
-<Callout title="Inside an ActionGroup">
-  The variant vocabulary narrows to match `<ActionButton>` (`default`,
-  `secondary`, `destructive-ghost`, `link`) and the cascade default becomes
-  `default`. The bold variants (`primary`, `destructive`) are only available
-  when `<LinkButton>` is used standalone, where it stays aligned with
-  [`<Button>`](/components/actions/button).
+<Callout title="Inside a ButtonGroup">
+  `<LinkButton>` shares the full `<Button>` variant vocabulary. With no local
+  `variant`, it inherits the group's cascade, which defaults to `ghost` so the
+  cluster reads as low-emphasis chrome. A local `variant` or `size` always wins
```
aromko: ⚠️ **Cascade default is wrong here.** `ButtonGroup` cascades `secondary` when `variant` is unset (`ButtonGroup.tsx`: `variant ?? 'secondary'`), matching the changeset and the `button-group` page. As written, a reader following this callout expects a low-emphasis ghost cluster but gets bordered `secondary` buttons.

```suggestion
  `variant`, it inherits the group's cascade, which defaults to `secondary`, the
  same baseline as a standalone button. A local `variant` or `size` always wins
```

sebald: Good catch, you're right. `ButtonGroup.tsx` cascades `variant ?? 'secondary'` (JSDoc `@default 'secondary'`), and the changeset and button-group page agree. The callout was the only place still saying `ghost`. Accepted the suggestion in 34736d13f. Thanks!

aromko (approved):
@sebald conflicts

PR #5499, opened 2026-06-05
feat(DST-1370): migrate ContextualHelp, SectionMessage, and EmptyState to the slot-configuration pattern

sarahgm:
# Description

Migrates the inline-message family to the slot-configuration pattern (Phase 0/1 references: Panel, Dialog/Drawer/Tray):

- **ContextualHelp**: `.Title` switches from `slot={noSlot}` to `slot="title"`, so the RAC dialog wires `aria-labelledby` automatically (a11y improvement; title tag changes `h3` → `h2`, same as `Dialog.Title`, visuals unchanged). New `ContextualHelp.Description` sub-component, fed by a `TextContext` description slot at the root.
- **SectionMessage**: `.Title` now renders a real heading via the `Title` primitive (fixes the a11y gap where the title was a plain `<div>`), configurable through the new `headingLevel` prop (default 3). When a title is present, the container becomes a `role="group"` labelled by the title via `aria-labelledby` (a bare `<div>` must not carry `aria-labelledby`; see code comment). New optional `SectionMessage.Description`. Dismiss button stays `CloseButton`.
- **EmptyState**: keeps its flat-props API; internally renders the title through the `Title` primitive (new `headingLevel` prop, default `h3`) and the description through the `Description` primitive (identical DOM via `elementType: 'div'`).
- **Action/button context**: all three roots publish a `ButtonContext` (the shared `RESET_BUTTON_CONTEXT`) alongside `HeadingContext`/`TextContext`, completing the slot-configuration set. It scopes action buttons placed inside (e.g. `SectionMessage.Content`, the `EmptyState` `action`, ContextualHelp content) to a clean baseline so they never inherit a surrounding container's button cascade (e.g. a `Panel.Header` ghost/small). No variant or positioning is imposed, so existing usage renders unchanged. Mirrors how `Panel.Header` publishes `ButtonContext` inline in its single Provider.
- **Description rhythm/color**: all three descriptions hug the title (4px above, 8px below). SectionMessage drives the description color per variant via a `--section-message-description` CSS variable (Card master/admin mechanism), consumed once at 80% opacity. Optional grid rows follow the Toast mechanism (static template, no row gap, margins own the spacing), so description-less messages render pixel-identical to before.

**Deviation from the ticket text**: the ticket lists four pattern contexts (including `ActionMenuContext`) and a "dismiss button becomes `<ActionButton>`" item. `ActionButton` was removed in #5494 in favor of the slot-aware `Button`, so the live equivalent is `ButtonContext` — now published (see above). `ActionMenuContext` is intentionally **not** published: it matches the implemented siblings (`Panel.Header` publishes only `ButtonContext`; `Card` publishes neither) and there is no `ActionMenu` usage in these components. The dismiss button stays `CloseButton`.

Note: a raw RAC `<Text>`/`<Heading>` without a `slot` prop inside these containers throws RAC's slot error — same trade-off as Panel; all Marigold primitives (`Text`, `Headline`, `Title`) are unaffected.

Closes DST-1370

## Screenshots / Preview

Storybook: `SectionMessage` → "With Description" / "With Action" / "Title Is Semantic Heading", `ContextualHelp` → "With Description", `EmptyState` → "Basic" (title now inspects as `<h3>`).

## Test Instructions

1. `pnpm sb` → SectionMessage "With Description": title inspects as `<h3>` (switch `headingLevel` control), description renders muted between title and content; existing stories without description are visually unchanged.
2. SectionMessage "With Action" / EmptyState "With Action": action buttons render at their own baseline even when the component is nested under a container that provides a `ButtonContext` (covered by the new scoping unit tests).
3. ContextualHelp "With Description": open the popover, the dialog has `aria-labelledby` pointing at the title (`h2`).
4. EmptyState "Basic": title inspects as `<h3>`; embedding consumers (Table/SelectList/Autocomplete empty states) render unchanged.
5. `pnpm test:unit` and `pnpm test:sb` — all green (new `component-test` stories incl. the `getByRole('heading')` AC test and the `ButtonContext` scoping tests for SectionMessage/EmptyState).

## Breaking Changes

No. All APIs are additive; the only consumer-visible changes are semantic tags (`div` → `h3` for SectionMessage/EmptyState titles, `h3` → `h2` for ContextualHelp) with unchanged visuals.

## Checklist

- [ ] Visual regression tests updated (not required for `beta-release` PRs)


jim761 (approved):

packages/system/src/types/theme.ts
```diff
@@ -200,7 +200,7 @@ export type Theme = {
       ComponentStyleFunction<string, string>
     >;
     SectionMessage?: Record<
-      'container' | 'icon' | 'title' | 'content',
+      'container' | 'icon' | 'title' | 'description' | 'content',
```
aromko, line 203: ⚠️ **Discussion — semver of the `Theme` type change.**

Adding `'description'` as a **required** key to the `SectionMessage` style `Record` (and `ContextualHelp` at line 82) is a compile-time breaking change for any *downstream/external* theme that defines styles for these components — their theme object will no longer typecheck without a `description` entry.

The `@marigold/system` bump is `minor` and the changeset documents the requirement, which is reasonable given themes are effectively first-party and this targets `beta-release`. Just flagging so the `minor` (vs `major`) classification is a conscious choice rather than an oversight.

packages/components/src/SectionMessage/SectionMessage.stories.tsx
```diff
@@ -43,6 +50,96 @@ export const Basic = meta.story({
...
+      'All files were copied to the archive.'
+    );
+    await expect(description).toBeInTheDocument();
+    // Element type comes from the root's `TextContext` slot config.
+    await expect(description.tagName).toBe('P');
+
+    // The description is muted (variant color via
+    // `--section-message-description` at 80%), so its computed color must
+    // differ from the title's full variant foreground.
+    await expect(getComputedStyle(description).color).not.toBe(
```
aromko: 💡 **Suggestion — brittle assertion.**

Comparing `getComputedStyle(description).color` string-inequality against the heading (and reading the resolved `--section-message-description` variable below) works in the real-browser runner, but computed-color string comparison is fragile to serialization and future theming changes.

Consider asserting on the presence of the variant variable / class instead of comparing resolved color strings, so the test pins behaviour rather than a specific rendered color value. Not blocking.

aromko (commented):
## Automated Code Review

**Recommendation: Approve** ✅ — high-quality, well-documented, well-tested migration that fixes a real a11y gap and fully meets DST-1370's acceptance criteria. CI is green and changed-file coverage is ~100%.

### What's good
- **Real a11y fix**: `SectionMessage.Title` goes `<div>` → semantic `<h3>`; container becomes `role="group"` labelled by the title **only when a title is present** (correctly avoids `aria-labelledby` on a bare `<div>`). ContextualHelp dialog now gets `aria-labelledby` → `<h2>`; EmptyState title is a navigable heading. All asserted by tests.
- **Pattern correctness**: `ContextualHelpTitle` deliberately keeps RAC `<Heading slot="title">` to consume the Dialog's own `HeadingContext` instead of re-providing it — the inline comment explains why re-providing would clobber the `aria-labelledby` wiring. `useSlot(false)` cleanly drives the conditional `role`/`aria-labelledby`.
- **Deviation note is accurate**: `ActionButton` was removed in #5494, so `ButtonContext` is the live equivalent; `ActionMenuContext` intentionally not published (matches `Panel.Header`/`Card`); dismiss stays `CloseButton`. Consistent with the ticket leaving the exact context set to PR discretion.
- Testing follows project conventions (Vitest, `userEvent`, `getByRole`/`getByText`, `component-test` tags). Good coverage of the `ButtonContext` scoping behaviour.

### Acceptance criteria (DST-1370)
All satisfied — Provider published on all three roots; semantic heading + `getByRole('heading')` test; `.Description` added to ContextualHelp & SectionMessage; EmptyState refactored internally (ticket option **b**, flat-props preserved); no consumer API break; changeset added.

### Notes (non-blocking)
See the two inline comments — one discussion point on the `Theme` type semver, one suggestion on a brittle computed-color assertion. Neither blocks merge.

---

packages/system/src/types/theme.ts
sarahgm: Conscious choice — keeping it required. Every sibling theme record in `theme.ts` (`Card`, `Panel`, `ContextualHelp`, …) uses required keys, so a `Partial`/optional `description` here would break that pattern for one entry. The changeset documents the new requirement explicitly, the `@marigold/system` bump is `minor`, and this targets `beta-release` where themes are first-party, so the impact is contained. Leaving as `minor`.

packages/components/src/SectionMessage/SectionMessage.stories.tsx
sarahgm: Good catch — done in d437d08cb. Replaced the computed-color string comparison with an assertion on the description slot's class (`text-(--section-message-description)`), which pins that the description renders through the muted variant-driven variable without depending on a resolved color string. Kept the existing assertion that the variable wins the cascade over the `currentColor` fallback.

aromko (approved):

PR #5501, opened 2026-06-05
feat([DST-1369]): adopt slot-configuration pattern in Dialog, Drawer, and Tray components

OsamaAbdellateef:
# Description

Bring `Dialog`, `Drawer`, and `Tray` in line with the slot-configuration pattern that `Panel` (DST-1367) and `Card` (DST-1373) established. The three overlays each reimplemented the same bespoke title/header structure in parallel; they now converge on the shared slot vocabulary.

Each overlay root now mints its own `titleId` (`useId`), detects the title with `useSlot`, sets `aria-labelledby` explicitly, and publishes the slot contexts via a single `<Provider>` — the same shape as `Card.tsx` / `Panel.tsx`:

- **`*.Title`** is a thin wrapper over the slot-aware `<Title slot="title">`; **`*.Description`** (new) wraps `<Description slot="description">`.
- **`*.Header`** (new) is an **optional** layout wrapper that groups a title + description into one header zone. A bare `<Title>` (or `<*.Title>`) without a header is a first-class, accessible form — `aria-labelledby` resolves to it automatically.
- The bespoke `*Context` per component now carries `{ classNames, titleId, hasTitle, titleSlotRef }` so `*.Header` can re-publish the slot config.
- The RAC `<Header>` element that previously wrapped the title is gone — the title carries the header chrome directly, with no change to rendered visuals.

The compound-component API is unchanged and still supported. Theme `Dialog` / `Drawer` / `Tray` slot maps gain a `description` entry.

Two deliberate divergences from a literal Card/Panel clone, both forced by these components wrapping RAC's `<Dialog>` rather than a plain element:
1. The slot `<Provider>` lives *inside* the RAC `<Dialog>` (required to override RAC's own `HeadingContext`).
2. No custom "missing accessible name" dev-warning on Drawer/Tray — they run their hooks while still *closed* (title unmounted), so the check would fire spuriously; RAC already warns at open time. Dialog keeps the warning since its slot logic is in the open-only inner component.

Closes DST-1369

## Screenshots / Preview

Storybook → Components → Dialog / Drawer / Tray. Key stories: `SlotPrimitives` (direct primitives + `<ActionGroup>` in the Actions slot) and `TitleOnlyWithoutHeader` (bare `<Title>`, asserts `aria-labelledby`). Docs: `/components/overlay/dialog`, `/components/overlay/drawer` — new "Title and description" sections with a title-only demo.

No visual changes intended.

## Test Instructions

1. `pnpm test:unit --run Dialog Drawer Tray Provider` → 48 tests pass.
2. `pnpm test:sb --run Dialog Drawer Tray` → 11 story tests pass (incl. the title-only `aria-labelledby` checks).
3. `pnpm sb` → open Dialog / Drawer / Tray; confirm titles, headers, and the new `SlotPrimitives` / `TitleOnlyWithoutHeader` stories render correctly.
4. `pnpm start` → `/components/overlay/dialog` and `/components/overlay/drawer`; verify the new "Title and description" section and title-only demo.

## Breaking Changes

No. The compound form (`<Dialog.Title>`, etc.) is unchanged. Internally the `<header>` element wrapping the title is removed (chrome moves onto the title slot) — same rendered visuals.

## Checklist


packages/components/src/Drawer/DrawerHeader.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * header zone. A `<Drawer.Title>` works without it. Use `<Drawer.Header>`
+ * when you also want a description laid out beneath the title.
+ *
+ * It re-publishes the title/description slot config with the header-local
+ * classNames (the chrome lives on this container) while re-supplying the `id`
+ * and `ref` from the root so the accessible-name wiring keeps working.
+ */
+export const DrawerHeader = ({ children }: DrawerHeaderProps) => {
+  const { variant, size, titleId, titleSlotRef } = useDrawerContext();
+  const classNames = useClassNames({ component: 'Drawer', variant, size });
```
aromko: 💡 `DrawerHeader` recomputes the theme classes here, but the PR now publishes `classNames` on `DrawerContext`. The Dialog and Tray headers read `classNames` straight from context (`useDialogSlotContext` / `useTrayContext`) instead of re-running the hook. Reading it from context here would drop the redundant `useClassNames` call and align all three headers.

```suggestion
  const { classNames, titleId, titleSlotRef } = useDrawerContext();
```

(and drop the `variant`/`size` destructure, since they're no longer needed — the `useClassNames` import can go too.)

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -29,27 +33,81 @@ const InnerDialog = ({
...
+      },
+    }),
+    [classNames.header, classNames.title, titleId, titleSlotRef]
+  );
+
+  const textProps = useMemo(
+    () => ({
+      slots: {
+        description: {
+          className: classNames.description,
```
aromko: 💡 The root-level `description` slot gets `classNames.description` but no `[grid-area]`, while the Dialog grid only defines `'title'_'content'_'actions'`. A `<Dialog.Description>` dropped directly into `<Dialog>` (without `<Dialog.Header>`) would become an unplaced grid item landing in an implicit row. All demos/stories nest `Description` inside a `Header`, so this path is untested.

Low severity since the documented bare form is title-only — but worth either a doc note that `Description` requires a `Header`, or a grid-area fallback here. (Same applies to the root `textProps` in `Drawer.tsx` and `Tray.tsx`.)

packages/components/src/Drawer/DrawerHeader.tsx
OsamaAbdellateef: ✅

packages/components/src/Dialog/Dialog.tsx
OsamaAbdellateef: Resolved and added for Drawer, Dialog and Tray ✅ @aromko

aromko (commented):
## Automated Code Review (review-pr skill)

Clean, well-documented refactor — faithful to DST-1369's acceptance criteria, CI-green. **Approve** overall; the inline notes below are non-blocking (1 warning, 3 suggestions). VRT intentionally excluded since this targets `beta-release` (the workflow only gates merge-to-`main`).

packages/components/src/Dialog/Context.tsx
```diff
@@ -1,8 +1,41 @@
...
 export const DialogContext = createContext<DialogContextProps>({});
+
+/**
+ * Carries the resolved theme classes and the title-slot wiring from the
+ * `<Dialog>` root down to its sub-components (`Dialog.Header`,
+ * `Dialog.Content`, `Dialog.Actions`). Mirrors `CardContext` / `PanelContext`
+ * from the rest of the slot-configuration family.
+ */
+export interface DialogSlotContextValue {
+  classNames: {
```
aromko: ⚠️ **Inconsistent slot-`classNames` typing across the three contexts.** This uses a mapped type `{ [Key in keyof ThemeComponent<'Dialog'>]: string }`, while `Drawer/Context.tsx` uses `ComponentClassNames<'Drawer'>` and `Tray/Context.tsx:5` keeps a hand-written object literal. Since this PR is about convergence, consider aligning all three on `ComponentClassNames<'X'>` (the Drawer approach) so the Tray literal can't drift from the theme type.

```suggestion
  classNames: ComponentClassNames<'Dialog'>;
```
(plus swapping the `ThemeComponent` import for `ComponentClassNames`)

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -29,27 +33,81 @@ const InnerDialog = ({
...
+      },
+    }),
+    [classNames.header, classNames.title, titleId, titleSlotRef]
+  );
+
+  const textProps = useMemo(
+    () => ({
+      slots: {
+        description: {
+          className: cn('[grid-area:content]', classNames.description),
```
aromko: 💡 The root `description` slot is mapped to `[grid-area:content]`. A bare `<Dialog.Description>` used header-less *and* a `<Dialog.Content>` would both occupy the `content` grid area and could overlap. The docs/stories only show description-inside-`Header`, so this is an edge authoring case — a short doc note (or a guard) would cover it. Same applies to Drawer/Tray.

packages/components/src/Dialog/DialogHeader.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * inside `<Dialog>` and it still claims the header area. Use `<Dialog.Header>`
+ * when you also want a description laid out beneath the title.
+ *
+ * It re-publishes the title/description slot config with the header-local
+ * classNames (the chrome — padding — lives on this container) while
+ * re-supplying the `id` and `ref` from the root so the accessible-name wiring
+ * keeps working. `level`, `id`, and `ref` are re-published (not inherited)
+ * because Provider values do not merge.
+ */
+export const DialogHeader = ({ children }: DialogHeaderProps) => {
```
aromko, line 28: 💡 `DialogHeader` / `DrawerHeader` / `TrayHeader` — and the root `rootHeadingProps`/`textProps` memos in each overlay — are near-identical. A shared helper (e.g. `useOverlaySlotProviders`) could remove the remaining cross-component duplication this ticket targets, now one level up. Optional, given the deliberate per-component divergences (Provider placement, dev warning).

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -62,7 +120,16 @@ const InnerDialog = ({
           onPress={state?.close}
         />
       )}
-      {children}
+      <Provider
+        values={[
+          [DialogSlotContext, contextValue],
+          [HeadingContext, rootHeadingProps],
+          [TextContext, textProps],
+          [ActionMenuContext, {}],
```
aromko, line 108: 💡 DST-1369's AC lists four contexts to publish, including `ActionButtonContext`. This publishes `ActionMenuContext` (+ slot contexts) and the new stories use `ButtonGroup` rather than `ActionButton.Group`. If dropping `ActionButtonContext` is intentional, all good — just flagging the deviation from the AC.

packages/components/src/Dialog/DialogHeader.tsx
OsamaAbdellateef: Added `useOverlayHeaderSlotProps` to utils ✅ @aromko

packages/components/src/Dialog/Dialog.tsx
OsamaAbdellateef: It is intentional, The AC referenced `ActionButtonContext` which doesn't exist i think,  @aromko

packages/components/src/Dialog/Dialog.tsx
OsamaAbdellateef: Added comments for the 3 of them ✅  @aromko

docs-internal/ai-usage-proposal.html
```diff
@@ -1,332 +1,1013 @@
 <!doctype html>
```
aromko, line 1: @OsamaAbdellateef Was this formatting intentional? Consider to just format files you touched maybe a setting in your IDE.

OsamaAbdellateef: Reverted ✔️  @aromko

aromko: Still shown as change but you didn't changed here something intentionally or?

OsamaAbdellateef: ✅  @aromko

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -29,27 +33,81 @@ const InnerDialog = ({
...
+    typeof children === 'function'
+      ? children({ close: state?.close ?? (() => {}) })
+      : children;
+
+  // Published at the Dialog root so a `<Dialog.Title>` (or a bare
+  // `<Title slot="title">`) works without a wrapping `<Dialog.Header>`. The
+  // `title` slot carries the header chrome (padding) so a title-only dialog
+  // still looks like a titled bar; `<Dialog.Header>` re-publishes a stripped
+  // variant when it owns the chrome itself.
+  const rootHeadingProps = useMemo(
```
aromko: ⚠️ **W1 — Root slot-prop construction is triplicated across the three overlays.**

This `rootHeadingProps` + `textProps` `useMemo` block is duplicated verbatim in `Drawer.tsx` (`const rootHeadingProps` ~L115) and `Tray.tsx` (~L135) — only the `ComponentClassNames<...>` generic differs. You already extracted `useOverlayHeaderSlotProps` for the header path; the root path is the same idea minus the `[grid-area:*]` chrome.

Since the whole point of DST-1369 is to collapse three parallel implementations into one shared mechanism, consider a sibling hook:

```ts
// useOverlayRootSlotProps(classNames, titleId, titleSlotRef)
//   → same as useOverlayHeaderSlotProps but prepends
//     [grid-area:title] to title and [grid-area:content] to description
```

Not blocking, but it removes the last "three places to fix a bug" spot the ticket calls out.

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -29,27 +33,81 @@ const InnerDialog = ({
...
   const state = use(OverlayTriggerStateContext);
+  const titleId = useId();
   const classNames = useClassNames({
     component: 'Dialog',
     variant,
     size,
   });
+  const [titleSlotRef, hasTitle] = useSlot(!ariaLabel);
+
+  if (process.env.NODE_ENV !== 'production' && !ariaLabel && !hasTitle) {
```
aromko: ⚠️ **W2 — Dev-warning false-positives when the dialog is labeled via `aria-labelledby`.**

This guard only checks `!ariaLabel && !hasTitle`, so a perfectly accessible `<Dialog aria-labelledby="some-id">` (no title, no `aria-label`) still emits the "Renders a dialog without an accessible name" warning in dev. Include an explicitly-provided `aria-labelledby` in the check:

```suggestion
  if (
    process.env.NODE_ENV !== 'production' &&
    !ariaLabel &&
    !hasTitle &&
    !props['aria-labelledby']
  ) {
```

aromko (commented):
## Automated Code Review

### PR #5501 — feat([DST-1369]): adopt slot-configuration pattern in Dialog, Drawer, and Tray

#### Overview
| Field | Value |
|-------|-------|
| Branch | `dst-1369_dialog-drawer-slot-configration` → `beta-release` |
| Files Changed | 31 (+918 / −97) |
| CI | ✅ All green (Typecheck, Lint, Unit 4/4, Storybook 4/4, Coverage, Vercel previews) |
| Visual Regression | Not applicable (base: `beta-release`, not `main`) |

#### Linked Ticket — DST-1369 (In Review)
🧹 Migrate Dialog / Drawer / Tray to the slot-configuration pattern. Acceptance criteria verified against the diff:

- ✅ Each overlay publishes one root `<Provider>` (not on `*.Header`)
- ✅ Each `*Title` reduced to a thin `<Title slot="title">` wrapper; new `*Description` wrappers added
- ✅ Each `*Header` is a pure, optional layout wrapper owning no root Provider
- ✅ Title-only authoring tested with `aria-labelledby` assertions (all 3) + `SlotPrimitives` stories
- ✅ Docs "title-only" examples (Dialog & Drawer); changeset added (`patch`)

#### Checklist
- **TypeScript** ✅ No `any`; `import type` used; one documented `as` cast for the `DrawerContext` default.
- **Component patterns** ✅ `useClassNames` + `cn()`; slot config follows the Card/Panel family; no `className`/`style` exposed.
- **React performance** ✅ No inline components; component code imports direct paths (no barrels).
- **Testing** ✅ Vitest + `storybook/test`, `userEvent`, accessible queries, `component-test` tags.
- **Accessibility** ✅ `aria-labelledby` wiring verified and asserted across all three overlays.

#### Issues Found

**Warnings (Should Fix):**

**W1 — Root slot-prop construction is triplicated verbatim across the three overlays.** `Dialog.tsx`, `Drawer.tsx`, and `Tray.tsx` each contain an identical `rootHeadingProps` + `textProps` `useMemo` block (only the `ComponentClassNames<...>` type differs). The author extracted `useOverlayHeaderSlotProps` for the *header* path but left the *root* path copied three times — which runs against the ticket's stated motivation ("three parallel implementations means three places to fix a bug"). Consider a sibling `useOverlayRootSlotProps(classNames, titleId, titleSlotRef)` that adds the `[grid-area:title]` / `[grid-area:content]` chrome. *(See inline comment.)*

**W2 — Dialog dev-warning false-positives when labeled via `aria-labelledby`.** The warning fires on `!ariaLabel && !hasTitle`, ignoring an explicitly supplied `aria-labelledby`. A valid `<Dialog aria-labelledby="...">` (no title, no `aria-label`) would emit the "Renders a dialog without an accessible name" warning in dev. *(See inline comment with suggested fix.)*

#### Recommendation
**Comment / Approve-with-nits.** A high-quality, well-scoped refactor that meets the acceptance criteria with genuine `aria-labelledby` test coverage and green CI. Nothing blocks merge; addressing W1 and W2 before merge is recommended.

---

packages/components/src/Dialog/Dialog.tsx
OsamaAbdellateef: Integrated @aromko

packages/components/src/Dialog/Dialog.tsx
OsamaAbdellateef: Integrated @aromko

aromko (approved):

packages/components/src/Drawer/Drawer.tsx
```diff
@@ -1,5 +1,7 @@
 import type { CSSProperties } from 'react';
-import { use, useRef } from 'react';
+import { use, useId, useMemo, useRef } from 'react';
+import { useOverlayRootSlotProps } from '../utils/useOverlayRootSlotProps';
```
sebald: ⚠️ **Import order** — this relative import is placed between `react` and `react-aria-components`. The repo's `@trivago/prettier-plugin-sort-imports` config sorts relative imports (`^[./]`) last, so this ordering will be flagged by the format check. Running `pnpm format` will reorder it (move the two `react-aria-components` imports above the relative `../utils/...` import). Dialog and Tray imports are fine.

packages/components/src/utils/useOverlayRootSlotProps.ts
```diff
@@ -0,0 +1,47 @@
...
+      },
+    }),
+    [classNames.header, classNames.title, titleId, titleSlotRef]
+  );
+
+  const textProps = useMemo(
+    () => ({
+      slots: {
+        description: {
+          className: cn('[grid-area:content]', classNames.description),
```
sebald, line 38: ⚠️ **Root-level `<*.Description>` overlaps `<*.Content>`** — a description published at the root lands in `[grid-area:content]`, the same grid area as `<*.Content>`. So dropping `<Dialog.Description>` (or the Drawer/Tray equivalents) directly under the root instead of inside `<*.Header>` makes the two overlap. This is currently only guarded by a JSDoc note on the `*.Description` components. Since `*.Description` is public compound API, consider making misuse harder (or emitting a dev warning) rather than relying on docs alone.

.changeset/dialog-drawer-tray-slot-configuration.md
```diff
@@ -0,0 +1,15 @@
+---
+'@marigold/components': patch
```
sebald: 💡 **Semver bump** — this PR adds new public API (`Dialog.Header` / `Dialog.Description` and the Drawer/Tray equivalents) and is titled `feat(...)`, which is conventionally a **`minor`** bump rather than `patch`. If `patch` is intentional because the base branch is `beta-release`, feel free to ignore — just flagging the mismatch.

OsamaAbdellateef: Integrated @sebald

packages/components/src/utils/useOverlayRootSlotProps.ts
OsamaAbdellateef: Integrated @sebald

packages/components/src/Drawer/Drawer.tsx
sebald: Reopening — the externals are sorted correctly now, but `prettier --check` (repo config, `@trivago/prettier-plugin-sort-imports`) still reports `Drawer.tsx` as unformatted. The relative group isn't alphabetised: `../utils/useOverlayRootSlotProps` and `../utils/useSlot` need to move to the end, after `../intl/messages`. Expected order:

```
import { ResetButtonContext } from '../Button/ResetButtonContext';
import { CloseButton } from '../CloseButton/CloseButton';
import { ActionMenuContext } from '../Menu/ActionMenuContext';
import { intlMessages } from '../intl/messages';
import { useOverlayRootSlotProps } from '../utils/useOverlayRootSlotProps';
import { useSlot } from '../utils/useSlot';
```

Running `pnpm format` fixes it.

sebald: Re-opened this thread — it was marked resolved but `prettier` still flags the file.

**How to reproduce:**

\`\`\`bash
# check just this file
pnpm prettier --check packages/components/src/Drawer/Drawer.tsx
# or the whole repo, as CI does
pnpm format --check
\`\`\`

Current output is `[warn] …Drawer.tsx — Code style issues found`, exit code 1, because the relative-import group is not alphabetised (`../utils/*` is sorted ahead of `../CloseButton`, `../Menu`, `../intl`). The `@trivago/prettier-plugin-sort-imports` plugin sorts alphabetically **within** each import group, so it wants `../utils/useOverlayRootSlotProps` and `../utils/useSlot` moved to the end, after `../intl/messages`.

**Fix:**

\`\`\`bash
pnpm format        # rewrites the import order
pnpm prettier --check packages/components/src/Drawer/Drawer.tsx   # now exits 0
\`\`\`

Worth confirming green locally before merge — the CI format gate runs the same check.

OsamaAbdellateef: Fixed — moved ../utils/useOverlayRootSlotProps and ../utils/useSlot after ../intl/messages so the relative group is alphabetically ordered. Confirmed pnpm prettier --check packages/components/src/Drawer/Drawer.tsx exits 0. @sebald

sebald (approved):

PR #5514, opened 2026-06-11
fix(DSTSUP-261): guard Tray against the react-aria HiddenContext split (mobile Select tray)

sebald:
# Description

On a touch/mobile viewport, `Select` (and the other components that fall back to a `Tray`: `ComboBox`, `Autocomplete`, `TagField`, `DatePicker`, `Menu`) could open an **empty bottom sheet** next to the real one: tapping an option selected nothing, the trigger kept showing the placeholder, and the page was left `inert` until reload.

**Root cause — a dependency-resolution split, not component logic** (confirmed against the installed packages):

- `react-aria-components@1.18.0` pins `react-aria@3.49.0` / `react-stately@3.47.0` **exactly**. Its `Select` became a `createHideableComponent` in v1.18 (RAC PR #10019), which *sets* the `HiddenContext` during the collection's hidden build pass.
- `@react-aria/collections@3.1.1` — the source of `Tray`'s `useIsHidden` — depends on `react-aria@^3.48.0` (**caret**).
- Both read the *same* `HiddenContext` from `react-aria/private/collections/Hidden`, so they only agree when `react-aria` resolves to one version. When a consumer's lockfile resolves those two edges to different `react-aria` generations, the Tray's hidden-pass guard reads a context the `Select` never sets → the guard never fires → a duplicate empty `TrayModal` portals into `<body>` → the two modals `inert` each other → options are no longer hit-testable.

Thanks to the reporter for the precise diagnosis.

**Fix: `Tray` now verifies the hidden pass through the DOM, independent of context identity.** react-aria renders the hidden pass into a `<template>` element whose content lives in a **detached `DocumentFragment`** — the duplicate modal only ever appears because RAC's `Modal` portals to `<body>`, escaping that template. `Tray` renders a `<span hidden>` probe next to its modal and checks `isConnected` in a layout effect at mount:

- **Real pass** → probe is in the live document → modal renders exactly as before (zero behavior change on the healthy path).
- **Hidden pass with a split context** → probe sits in the detached fragment (`isConnected === false`) → `Tray` flips to its children-only branch **at mount, before the tray can ever open**, so the empty modal never mounts.

`useIsHidden()` stays as the render-time fast path; the probe is the safety net that holds **regardless of how a consumer's lockfile resolves `react-aria`** — frozen lockfiles, fresh installs, and future react-aria releases alike. No dependency pins required.

Closes DSTSUP-261

## Why not pin `react-aria`/`react-stately`?

An earlier iteration of this branch pinned both to the versions RAC pins. Deeper research showed the pins cannot actually protect consumers:

1. **They don't constrain the edge that splits.** `Tray` reads `useIsHidden` through `@react-aria/collections`, whose *own* `react-aria@^3.48.0` range is resolved independently by the consumer's package manager. pnpm reuses existing lockfile resolutions for unchanged packages, so the affected consumer (lockfile: `@react-aria/collections → react-aria@3.48.x`) would upgrade Marigold and **still be split**. A direct pin is just a third edge — not an override.
2. **The split re-arms on upstream's next release — even for fresh installs.** Today `react-aria@3.49.0` is the latest on npm, so the caret happens to dedupe. The moment upstream ships 3.50 (~6-week release cadence), `^3.48.0` fresh-resolves to 3.50 while RAC 1.18 stays pinned to 3.49 → **every fresh install splits**, pins or not.
3. **The asymmetry is permanent by design.** The react-spectrum team's [2025 dependencies RFC](https://github.com/adobe/react-spectrum/blob/main/rfcs/2025-dependencies.md) explicitly keeps caret ranges on the `@react-aria/*` shell packages for backward compatibility.

Also verified: there is **no split-proof import path** for the context — RAC does not export `useIsHidden` (checked `exports/index.ts` and every subpath export in 1.18.0), and `react-aria` only exposes it via the explicitly private `react-aria/private/collections/Hidden`.

## The durable fix is upstream

RAC's `Popover`/`Menu` already guard the hidden pass internally — via RAC's own exact-pinned `react-aria`, so they can never split — which is why the desktop `Popover` path was never affected. RAC's `Modal` (what `Tray` builds on) has no such guard. A small upstream PR mirroring `Popover`'s `useIsHidden` guard in `ModalOverlay`/`Modal` is split-proof by construction (both ends live inside RAC's dependency closure). Once that lands, `Tray`'s guard and the `@react-aria/collections` dependency can be dropped entirely. Related upstream context: [#6021](https://github.com/adobe/react-spectrum/issues/6021) (ListBox empty inside a Modal — same missing-guard family), [#8777](https://github.com/adobe/react-spectrum/issues/8777) (context singletons vs version ranges).

Follow-up (separate ticket): `packages/components/src/index.ts` re-exports `I18nProvider` from `@react-aria/i18n` — the same split class. RAC re-exports `I18nProvider`, so switching that import to `react-aria-components` makes it split-proof for free.

## Test Instructions

1. `pnpm test:unit packages/components/src/Select/Select.test.tsx packages/components/src/Select/Select.hiddenContext.test.tsx packages/components/src/Tray/Tray.test.tsx` — all pass.
2. `Select.hiddenContext.test.tsx` is the regression test for the split itself: it blinds `useIsHidden` (mocked to `false` — exactly what reading an orphan context does) and asserts the probe keeps it to **one** dialog, the selection commits, and the tray closes. Before the fix, the same mock produced the two-dialog leak.
3. `pnpm test:sb` — the `Components/Select` `MobileControlled` story test passes.
4. All Tray consumers pass: `pnpm test:unit packages/components/src/ComboBox packages/components/src/Autocomplete packages/components/src/DatePicker packages/components/src/Menu packages/components/src/TagField packages/components/src/Tray` (75 tests).
5. Manual (mobile viewport): open a `Select`, tap an option → it commits, the trigger updates, the tray closes, exactly one overlay exists.

## Breaking Changes

No

## What changed

- `packages/components/src/Tray/Tray.tsx` — DOM-probe safety net (`<span hidden>` + `isConnected` layout-effect check) alongside the existing `useIsHidden()` guard; comments document the `HiddenContext` coupling, why the guard exists (RAC `Modal` lacks the `Popover`/`Menu` guard), and the upstream path that lets us drop it.
- `packages/components/package.json` / `pnpm-lock.yaml` — net-unchanged vs `main` (the interim `react-aria`/`react-stately` pins were removed again; see "Why not pin").
- **Tests:**
  - `Select.hiddenContext.test.tsx` — simulates the split and asserts the probe prevents the duplicate modal and the selection commits (was: executable documentation of the leak).
  - `Select.test.tsx` — mobile tray invariant: exactly one dialog, holding the listbox; controlled `value` round-trips.
  - `Tray.test.tsx` — locks the hidden-pass guard contract (`isHidden` → children render without a mounted modal).
  - `Select.stories.tsx` — `MobileControlled` story (`component-test`) verifies `onChange`/`value` selection on the tray.

## Checklist

### DIFF
--- a/packages/components/src/Select/Select.hiddenContext.test.tsx
+++ b/packages/components/src/Select/Select.hiddenContext.test.tsx
@@ -1,10 +1,8 @@
-import { render, screen, waitFor, within } from '@testing-library/react';
+import { screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { vi } from 'vitest';
-import { theme } from '../../../../themes/theme-rui/src/index';
-import { MarigoldProvider } from '../Provider/MarigoldProvider';
-import { mockMatchMedia } from '../test.utils';
-import { Select } from './Select';
+import { mockMatchMedia, renderWithOverlay } from '../test.utils';
+import { Basic } from './Select.stories';
 
 // DSTSUP-261 regression test.
 //
@@ -34,15 +32,12 @@ vi.mock('@react-aria/collections', async importActual => {
 const user = userEvent.setup();
 window.matchMedia = mockMatchMedia(['(width < 640px)']);
 
-const renderSelect = () =>
-  render(
-    <MarigoldProvider theme={theme}>
-      <Select label="Favorite" placeholder="Select Item">
-        <Select.Option id="Star Wars">Star Wars</Select.Option>
-        <Select.Option id="Star Trek">Star Trek</Select.Option>
-      </Select>
-    </MarigoldProvider>
-  );
+// Import the existing story instead of hand-rolling a fixture (CLAUDE.md:
+// "Don't create test fixtures/themes — import stories"). `Basic` is labelled
+// "Favorite", ships the Star Wars / Star Trek options, and wires the theme via
+// its decorator path; `renderWithOverlay` adds the portal container the tray
+// modal needs.
+const renderSelect = () => renderWithOverlay(<Basic.Component />);
 
 test('split HiddenContext (useIsHidden blind to hidden pass) does not leak a duplicate tray modal', async () => {
   // `false` everywhere mimics the Tray reading a different react-aria
--- a/packages/components/src/Select/Select.hiddenContext.test.tsx
+++ b/packages/components/src/Select/Select.hiddenContext.test.tsx
@@ -1,0 +1,69 @@
+import { render, screen, waitFor, within } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { vi } from 'vitest';
+import { theme } from '../../../../themes/theme-rui/src/index';
+import { MarigoldProvider } from '../Provider/MarigoldProvider';
+import { mockMatchMedia } from '../test.utils';
+import { Select } from './Select';
+
+// DSTSUP-261 regression test.
+//
+// On mobile, `Select` builds its collection in a hidden pass before rendering
+// for real. `Tray` skips its modal during that pass via `useIsHidden()`, which
+// reads the `HiddenContext` that react-aria-components' `Select`
+// (`createHideableComponent`) sets. Provider and consumer only share that
+// context when `react-aria` resolves to a single version. If two `react-aria`
+// generations are installed, `useIsHidden()` reads an orphan context that is
+// never flipped to `true`. Without further protection the guard never fires,
+// a second (empty) tray modal leaks, and the two modals `inert` each other so
+// the option becomes unselectable.
+//
+// We can't install two react-aria generations inside a single test, so we
+// simulate the split: force `useIsHidden` to stay `false` (blind to the hidden
+// pass), exactly like reading a context the collection renderer never sets.
+// `Tray`'s DOM-probe safety net must catch this: the hidden pass renders into
+// the collection's `<template>` (a detached DocumentFragment), so the probe is
+// never connected to the document and the duplicate modal must not mount —
+// regardless of how react-aria versions resolve.
+const hidden = vi.hoisted(() => ({ isHidden: false }));
+vi.mock('@react-aria/collections', async importActual => {
+  const actual = await importActual<typeof import('@react-aria/collections')>();
+  return { ...actual, useIsHidden: () => hidden.isHidden };
+});
+
+const user = userEvent.setup();
+window.matchMedia = mockMatchMedia(['(width < 640px)']);
+
+const renderSelect = () =>
+  render(
+    <MarigoldProvider theme={theme}>
+      <Select label="Favorite" placeholder="Select Item">
+        <Select.Option id="Star Wars">Star Wars</Select.Option>
+        <Select.Option id="Star Trek">Star Trek</Select.Option>
+      </Select>
+    </MarigoldProvider>
+  );
+
+test('split HiddenContext (useIsHidden blind to hidden pass) does not leak a duplicate tray modal', async () => {
+  // `false` everywhere mimics the Tray reading a different react-aria
+  // generation's context than the Select's collection renderer sets.
+  hidden.isHidden = false;
+  renderSelect();
+
+  await user.click(screen.getByLabelText(/Favorite/i));
+  const dialog = await screen.findByRole('dialog');
+
+  // The broken DOM signature from the ticket was two overlay dialogs, only one
+  // of which holds the listbox. The DOM probe must suppress the hidden-pass
+  // copy, leaving exactly one populated tray modal.
+  expect(screen.getAllByRole('dialog')).toHaveLength(1);
+  expect(within(dialog).getByRole('listbox')).toBeInTheDocument();
+
+  // No leaked twin modal means the real tray is not `inert`-ed: tapping an
+  // option commits the selection and closes the tray.
+  await user.click(within(dialog).getByRole('option', { name: 'Star Trek' }));
+  await waitFor(() =>
+    expect(screen.queryByRole('dialog')).not.toBeInTheDocument()
+  );
+  expect(screen.getByLabelText(/Favorite/i)).toHaveTextContent('Star Trek');
+});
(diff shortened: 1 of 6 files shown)


packages/components/src/Tray/Tray.tsx
```diff
@@ -1,6 +1,25 @@
...
+// internally (via RAC's bundled `react-aria`, so they can never split) — which
+// is why the desktop `Popover` path is immune. RAC's `Modal` (what `Tray` builds
+// on) has no such guard, so we add one here.
+//
+// The `react-aria`/`react-stately` pins in `@marigold/components` are an INTERIM
+// fix that keeps a single generation (keep them in sync when bumping RAC). The
+// durable fix is upstream — RAC guarding `Modal`'s hidden pass like `Popover`, or
+// RAC exporting `useIsHidden` — after which this guard and the
+// `@react-aria/collections` dependency can be dropped. See DSTSUP-261.
 import { useIsHidden } from '@react-aria/collections';
```
aromko, line 27: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files (new `MobileControlled` story, Tray comments, package pins) but the `Visual-Regression-Tests` workflow has not run for this branch.

packages/components/src/Select/Select.hiddenContext.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+const user = userEvent.setup();
+window.matchMedia = mockMatchMedia(['(width < 640px)']);
+
+const renderSelect = () =>
+  render(
+    <MarigoldProvider theme={theme}>
+      <Select label="Favorite" placeholder="Select Item">
+        <Select.Option id="Star Wars">Star Wars</Select.Option>
+        <Select.Option id="Star Trek">Star Trek</Select.Option>
+      </Select>
```
aromko: ⚠️ This builds a custom inline fixture (`MarigoldProvider` + hand-rolled `<Select>`) instead of importing the existing story. CLAUDE.md says *"Don't create test fixtures/themes — Import stories instead"*.

`Basic.Component` from `Select.stories.tsx` already provides Star Wars / Star Trek options and theme wiring via the decorator path. Consider:

```suggestion
const renderSelect = () => renderWithOverlay(<Basic.Component />);
```

(and drop the `MarigoldProvider` / `theme` / inline `<Select>` imports).

packages/components/src/Tray/Tray.test.tsx
```diff
@@ -3,8 +3,20 @@ import userEvent from '@testing-library/user-event';
 import { renderWithOverlay } from '../test.utils';
 import { Basic, DismissControlsWithCallbacks } from './Tray.stories';
 
+// Lets individual tests drive the `useIsHidden()` guard (DSTSUP-261). Defaults
+// to `false` so the existing tests keep exercising the real modal path.
+const hiddenState = vi.hoisted(() => ({ isHidden: false }));
+vi.mock('@react-aria/collections', async importActual => {
+  const actual = await importActual<typeof import('@react-aria/collections')>();
+  return { ...actual, useIsHidden: () => hiddenState.isHidden };
```
aromko, line 15: ⚠️ This file-level mock applies to **every** test in `Tray.test.tsx`, not just the new hidden-pass test. With the default `isHidden: false` it's behavior-preserving today, but it silently couples unrelated tests to the mock and can mask future regressions in real `useIsHidden` behavior.

Consider scoping the mock + `beforeEach` reset to a `describe('hidden collection pass', …)` block that contains only the new test.

packages/components/src/Select/Select.hiddenContext.test.tsx
sebald: Done in 22941f7a8. Replaced the hand-rolled `MarigoldProvider` + inline `<Select>` fixture with `renderWithOverlay(<Basic.Component />)` and dropped the `theme` / `MarigoldProvider` / `Select` imports. `Basic` is labelled "Favorite" and ships the Star Wars / Star Trek options, and `renderWithOverlay` adds the portal container the tray needs — so the assertions are unchanged and pass. Thanks for the catch.

packages/components/src/Tray/Tray.test.tsx
sebald: Addressed in 22941f7a8 — with one caveat on the exact mechanism. Vitest hoists every `vi.mock` factory to module scope, so the mock itself can't be nested inside a `describe` block. What I could do (and did): move the hidden-pass test **and** its `useIsHidden` reset into a `describe('hidden collection pass', …)` block, so the `true` toggle is set and reset only there and can never leak into the tests above. I also removed the file-wide `beforeEach` reset and documented why the mock stays file-scoped: the partial mock keeps all other `@react-aria/collections` exports real and defaults `useIsHidden` to `false`, which is exactly what the real hook returns outside a hidden pass — so the other tests are behaviorally untouched. If you'd prefer a fully isolated mechanism we'd have to drop `vi.mock` for a private `HiddenContext` provider, which reaches into `react-aria/private/...` — I think the scoped block is the better trade-off, but happy to revisit.

packages/components/src/Tray/Tray.tsx
sebald: Agreed — the new `MobileControlled` story adds a snapshot, so VRT should run. Triggering Chromatic now; I'll leave this thread open until the run reports back. (The probe element is `<span hidden>`, so no existing snapshot should move — the only expected delta is the new story.)

sebald: Chromatic run dispatched: https://github.com/marigold-ui/marigold/actions/runs/27489580875 — leaving this thread open until it reports back.

sebald: VRT reviewed on Chromatic ([build 57](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=57)): workflow passed (425 stories / 86 components). The 14 flagged changes are all baseline drift (stale inherited baseline from the beta-release churn on `main`), not Select/Tray regressions — reviewed and accepted. Resolving.

aromko (approved):

PR #5515, opened 2026-06-12
docs(DST-1288): add Table Record Management pattern

sebald:
# Description

Adds a new **Table Records** pattern page under `docs/content/patterns/user-input/`, an opinionated, demo-backed guide for creating, correcting, and maintaining records in data heavy tables. It also ships the `Table.EditableCell` SSR fix that the inline-edit demo depends on.

The page is written from the user's intent rather than the components. Each section opens with the situation the user is in and the problem it creates, then arrives at the interaction that solves it (problem and context first, solution second). The sections are:

- **Choosing an approach** — a decision table framed by two questions: how much input the task needs, and how much of the table the user must still see while doing it. Includes an explicit stance against modals for record creation.
- **Fixing a single value** — the user spots one wrong value while scanning, where a full form is out of proportion to the change. Solved with inline editing (`Table.EditableCell`).
- **Creating and editing whole records** — the user needs a real form but keeps referring back to the table as reference. Solved with a `Drawer` and its split screen model, including a guard against discarding unsaved changes.
- **Entering records in a batch** — the user works through a list in one sitting, where reopening the drawer and retyping shared values repeats per record. Solved with "Save and add another" plus selective field retention.
- **When a saved record disappears** — filtering, sorting, and pagination decide visibility, and a freshly saved record can drop out of view (the filter paradox, and sorting/pagination displacement). Solved with filter inheritance, a proactive `SectionMessage` warning, and confirming toasts.
- **When the table hides form fields** — hidden columns make the form and table drift apart. Solved with field ordering and progressive disclosure.

Four interactive demos, all Marigold only, each wrapping the table in a `<Panel>` with a `bleed` content area and rendering on the page background:

- **Inline editing** with `Table.EditableCell` (Name, Capacity, Status)
- **Drawer create and edit** — a single drawer handles both, a per-row "Edit" button opens it pre-filled, a `ConfirmationDialog` warns before discarding unsaved changes, and a success toast confirms each save
- **Save and add another** — a continuous entry loop with contextual field retention and focus return, using `EmptyState` for the empty table
- **Filter paradox** — the form inherits the active filter and warns via `SectionMessage` when a chosen value would hide the new record

New records append to the bottom of the table across the demos, toast messages are descriptive, and the larger demos use Shiki highlight markers to draw attention to the relevant lines in the Code tab.

## Included component fix

This branch also contains `fix(DST-1507)`: `Table.EditableCell` was inert after SSR hydration until an unrelated re-render occurred. The editing state and overlay now live in an inner component rendered inside the `Cell`, so interaction reconnects on its own after hydration. A changeset is included (`@marigold/components` patch).

> **Base branch:** this targets `beta-release` rather than `main`, because the page depends on `Panel` (with `bleed`), `Title`, `Description`, and the current `Drawer`/`SectionMessage`/`EditableCell` APIs, which are not yet on `main`.

Closes DST-1288

## Test Instructions

1. Run `pnpm start` and open the docs site.
2. Navigate to **Patterns → User Input → Table Records**.
3. **Fixing a single value:** select a Name/Capacity/Status cell, edit, commit with Enter, cancel with Escape. Confirm an empty Name is rejected.
4. **Creating and editing whole records:** use "Add venue" to create a record, then use a row's "Edit" button to revise it. Confirm the success toast and the table update. Edit a field, then try to close the drawer, and confirm the discard warning appears. "Keep editing" resumes, "Discard" closes and leaves the row unchanged.
5. **Entering records in a batch:** with the table empty, use the `EmptyState` action to add the first venue, then create several in a row with "Save and add another". Confirm City and Type are retained, record specific fields clear, and focus returns to Name.
6. **When a saved record disappears:** with the filter on "Active only", open "Add venue" and switch Status to Draft. Confirm the warning appears. Switch the filter to "All statuses" to see hidden records reappear.
7. **EditableCell SSR fix:** confirm inline editing works on first interaction in the docs (a server-rendered Next.js app) without needing a resize or other re-render.

## Breaking Changes

No

## Checklist

- [ ] Stories added/updated (with `component-test` tag where applicable) — N/A (pattern docs)
- [ ] Unit tests added/updated — N/A
- [ ] Visual regression tests updated (for UI changes) — N/A (SSR behavior fix, no visual change)


aromko (commented):
Automated review via Claude Code `/review-pr` — see inline comments. Recommendation: **Comment** (no blockers).

docs/content/patterns/user-input/table-record-management/table-record-inline-edit.demo.tsx
```diff
@@ -0,0 +1,166 @@
...
+    setVenues(prev =>
+      prev.map(venue =>
+        venue.id === id ? { ...venue, [field]: value } : venue
+      )
+    );
+  };
+
+  return (
+    <Panel>
+      <Panel.Header>
```
aromko, line 96: ⚠️ **Inconsistent `Panel` header structure across the four demos.**

This demo uses `<Panel.Header><Title/><Description/></Panel.Header>`, while `table-record-drawer-crud`, `table-record-add-another`, and `table-record-filter-paradox` use `<Panel.Content><Inline alignX="between" alignY="center"><Title/>…</Inline></Panel.Content>`.

Since all four demos render on the same page, the visual rhythm shifts when scrolling between them. Pick one structure and apply it to all four.

docs/content/patterns/user-input/table-record-management/table-record-drawer-crud.demo.tsx
```diff
@@ -0,0 +1,239 @@
...
+      );
+      addToast({ title: `“${data.name}” updated`, variant: 'success' });
+    }
+
+    setOpen(false);
+  };
+
+  return (
+    <Panel aria-label="Venues">
+      <ToastProvider position="bottom-right" />
```
aromko, line 240: ⚠️ **Only this demo renders a `ToastProvider`, and it sits *inside* `<Panel>`.**

The other two `useToast()` demos (`table-record-add-another`, `table-record-filter-paradox`) rely on an ambient provider higher up. Decide on one strategy — either give every demo its own provider for isolation, or drop this one and rely on the docs-site provider everywhere. Rendering `ToastProvider` inside a `Panel` is also unusual placement.

docs/content/patterns/user-input/table-record-management/table-record-filter-paradox.demo.tsx
```diff
@@ -0,0 +1,195 @@
...
+            </Select>
+          </Inline>
+          <Drawer.Trigger open={open} onOpenChange={setOpen}>
+            <Button variant="primary" onPress={openCreate}>
+              Add venue
+            </Button>
+            <Drawer size="medium" closeButton>
+              <Form
+                unstyled
+                key={open ? 'open' : 'closed'}
```
aromko: ⚠️ **Fragile form-reset key.**

`key={open ? 'open' : 'closed'}` remounts the form on every drawer toggle. If the close animation overlaps with user input, field state is wiped mid-keypress. The other demos use an incrementing `formKey` ref (`table-record-add-another`) or `${mode}-${id}` (`table-record-drawer-crud`). Match that pattern.

docs/content/patterns/user-input/table-record-management/table-record-inline-edit.demo.tsx
```diff
@@ -0,0 +1,166 @@
...
+  {
+    active: 'success',
+    draft: 'info',
+    archived: 'default',
+  };
+
+export default () => {
+  const [venues, setVenues] = useState(initialVenues);
+
+  const update = (
```
aromko: 💡 **Loose typing in `update`.**

`value` ends up as `string | number | File | null` and is assigned via `[field]: value`. Since users copy demo code, a typed per-field handler (or a generic over `Pick<Venue, K>`) would make this a better teaching example. Not blocking.

docs/content/patterns/user-input/table-record-management/table-record-drawer-crud.demo.tsx
```diff
@@ -0,0 +1,239 @@
...
+  const [selected, setSelected] = useState<Venue | null>(null);
+  const nextIdRef = useRef(initialVenues.length + 1);
+
+  const openCreate = () => {
+    setMode('create');
+    setSelected(null);
+    setOpen(true);
+  };
+
+  const openEdit = (id: React.Key) => {
```
aromko: 💡 **`onRowAction` id comparison.**

`openEdit = (id: React.Key) => venues.find(v => v.id === id)` — `React.Key` is `string | number`, but `venue.id` is always `string` here. Works today, but `String(id)` (or narrowing the parameter to `string`) makes the contract explicit so copy-pasters who switch to numeric ids don't get a silent `find` miss.

docs/content/patterns/user-input/table-record-management/index.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+- **Order visible fields first.** Fields that map to currently visible columns belong at the top of the form, ideally in the same left to right order as the columns. The form then mirrors the table, which gives the user cognitive continuity.
+- **Group hidden fields behind progressive disclosure.** Fields that map to hidden columns go below, inside a collapsed [`<Accordion>`](/components/navigation/accordion) section such as "Additional fields". This honors the user's intent while still surfacing every field.
+- **Keep required fields visible.** If a required field maps to a hidden column, leave it in the main section with its required indicator rather than burying it in the accordion. Only optional hidden fields belong there.
+
+For the form layout and progressive disclosure details, see the [Forms](/patterns/user-input/forms) pattern.
+
+## Related
+
+<TeaserList
```
aromko, line 151: 💡 **Inline SVG bloat in `<TeaserList>`.**

The Related section inlines ~60 lines of raw SVG markup. If the docs already standardize an icon source (heroicons import / shared icon component), reaching for that would shrink this section and avoid drift across patterns.

docs/content/patterns/user-input/table-record-management/index.mdx
```diff
@@ -0,0 +1,237 @@
...
+Inline editing keeps the correction where the eye already is. The cell switches from a read view to an input view, the user commits with the save button or by pressing Enter, and cancels with the cancel button or by clicking outside. Because the user never leaves the table, neighboring rows and columns stay in view the whole time.
+
+Use [`<Table.EditableCell>`](/components/collection/table) for the columns that should be editable, and leave the rest as plain cells. The `field` prop holds the input that appears while editing, and the children hold the value shown at rest. On save, read the submitted value from the form data and update your record.
+
+<ComponentDemo file="./table-record-inline-edit.demo.tsx" />
+
+A few rules keep inline editing safe and predictable:
+
+- **Reserve it for single, low risk fields.** A status, a name, or a quantity is a good fit. Anything with complex validation, foreign key relationships, or cascading side effects belongs in a drawer.
+- **Make editable cells discoverable.** The cell shows a subtle edit affordance on hover and focus, so users can tell which values are interactive without clicking every cell to find out.
```
aromko: 💡 **Verify the claimed edit affordance.**

*"The cell shows a subtle edit affordance on hover and focus"* — worth a manual check that `Table.EditableCell` actually renders this affordance in the current theme, since the prose sets user expectation.

docs/content/patterns/user-input/table-record-management/table-record-drawer-crud.demo.tsx
sebald: Outdated — resolving. The demo no longer uses an `onRowAction`/`React.Key` lookup. Editing is now triggered by a per-row `<Button onPress={() => openEdit(venue)}>Edit</Button>` that passes the fully-typed `Venue` object directly, so `openEdit(id: React.Key)` and the `venues.find` comparison no longer exist. No id coercion is needed.

docs/content/patterns/user-input/table-record-management/index.mdx
sebald: Outdated — resolving. The prose that claimed a "subtle edit affordance on hover and focus" has been removed. The inline-editing section no longer makes any claim about a hover/focus affordance, so there is nothing left to verify against the theme.

docs/content/patterns/user-input/table-record-management/table-record-inline-edit.demo.tsx
sebald: Fixed in 1e25cba6d. Aligned `inline-edit` to the `Panel.Content` + `Inline` structure the other three demos use. Note: `Panel.Header` could not host all four uniformly — `filter-paradox` keeps a `<Select>` status filter in its header, which `Panel.Header` has no grid slot for — so converging on the shared `Panel.Content` + `Inline` layout was the consistent choice.

docs/content/patterns/user-input/table-record-management/table-record-drawer-crud.demo.tsx
sebald: Fixed in 1e25cba6d. This turned out to be a latent bug, not just inconsistency: there is no ambient `ToastProvider` for docs content demos (`ComponentDemo` provides only `MarigoldProvider` + `OverlayContainerProvider`), so `filter-paradox` called `addToast` with no provider in scope and its save toasts never rendered. Added a `<ToastProvider position=\"bottom-right\" />` mirroring `table-record-drawer-crud`, so both toast demos now use the same strategy. (`table-record-add-another` uses no toasts.)

docs/content/patterns/user-input/table-record-management/table-record-filter-paradox.demo.tsx
sebald: Fixed in 1e25cba6d. `filter-paradox` now resets the form with an incrementing `formKey` (bumped in `openCreate`) instead of `key={open ? \"open\" : \"closed\"}`, matching `table-record-add-another`.

docs/content/patterns/user-input/table-record-management/table-record-inline-edit.demo.tsx
sebald: Done in 1e25cba6d. `update` is now generic over `K extends keyof Venue`, so the parsed value is typed as `Venue[K]` rather than `string | number | File | null`. A single cast remains at the `FormData` boundary (which is inherently stringly typed), with a comment explaining it.

docs/content/patterns/user-input/table-record-management/index.mdx
sebald: Not changing this one. The inline SVG follows the established docs convention: 53 of 55 `TeaserList` usages across the docs inline heroicons-style SVG, and only `panel/index.mdx` imports icon components. Switching just this page to `@marigold/icons` would make it the outlier. A repo-wide migration of `TeaserList` icons to `@marigold/icons` (which re-exports lucide) would be worthwhile, but belongs in its own cross-cutting change rather than this pattern PR.

docs/content/patterns/user-input/table-record-management/table-record-add-another.demo.tsx
```diff
@@ -0,0 +1,171 @@
...
+    if (addAnotherRef.current) {
+      // [!code highlight:8]
+      // Keep the drawer open, retain the contextual fields, and remount the
+      // form so the record specific fields clear and focus returns to "Name".
+      setRetained({ city: venue.city, type: venue.type });
+      setFormKey(key => key + 1);
+    } else {
+      setOpen(false);
+    }
+    addAnotherRef.current = false;
```
aromko: 🟡 This reset only runs when the submit succeeds. If **Save & add another** is pressed with an empty required **Name**, validation blocks the submit, `handleSubmit` never runs, and `addAnotherRef.current` stays `true`. The next plain **Save** then takes the "add another" branch — keeping the drawer open and resetting the form instead of closing it.

Consider deriving the intent from the submit event's `submitter` (`(e.nativeEvent as SubmitEvent).submitter`) instead of a ref that can go stale, or resetting the ref on validation failure too. Since this is reference demo code, consumers will copy the pattern as-is.

docs/content/patterns/user-input/table-record-management/table-record-drawer-crud.demo.tsx
```diff
@@ -0,0 +1,329 @@
...
+            <Table.Column rowHeader>Name</Table.Column>
+            <Table.Column>City</Table.Column>
+            <Table.Column>Type</Table.Column>
+            <Table.Column alignX="right">Capacity</Table.Column>
+            <Table.Column>Status</Table.Column>
+            <Table.Column alignX="right">Actions</Table.Column>
+          </Table.Header>
+          <Table.Body>
+            {venues.map(venue => (
+              <Table.Row key={venue.id}>
```
aromko, line 300: 💡 The pattern page states *"Indicate which record is being edited — highlight the matching row in the table behind the drawer"* (`index.mdx`), but this demo doesn't highlight the row currently open in the drawer. Consider applying a subtle highlight when `selected?.id === venue.id` so the demo matches its own guidance — or soften the rule to "recommended".

docs/content/patterns/user-input/table-record-management/table-record-add-another.demo.tsx
```diff
@@ -0,0 +1,171 @@
...
+  };
+
+  return (
+    <Panel aria-label="Venues">
+      <Panel.Header>
+        <Title>Venues</Title>
+        <Drawer.Trigger open={open} onOpenChange={setOpen}>
+          <Button variant="primary" onPress={openCreate}>
+            Add venues
```
aromko: 💡 Minor wording: this demo uses "Add venues" (plural) for the trigger and drawer title, while the other demos use "Add venue". If the plural is a deliberate nod to batch entry, ignore this — otherwise align for consistency.

aromko (commented):
Automated review (Claude Code `review-pr`). VRT skipped — this PR targets `beta-release`, not `main`. No critical issues; one warning and two polish suggestions inline below. The `Table.EditableCell` SSR fix is well-reasoned and the pattern page fully satisfies DST-1288.

docs/content/patterns/user-input/table-record-management/table-record-add-another.demo.tsx
sebald: Changed my mind — aligned this to the singular "Add venue" (trigger + drawer title) in 16afaef7d, matching the other demos. Done in the same place; thanks for flagging.

docs/content/patterns/user-input/table-record-management/table-record-add-another.demo.tsx
sebald: Good catch — fixed in `04fc733e7`. You're right that the ref went stale: with the `Form`'s default native validation, an empty required **Name** blocks the submit, so `handleSubmit` never ran and `addAnotherRef` stayed `true`, mis-routing the next plain **Save**.

Switched to deriving the intent from the submit event instead of a ref:

```ts
const submitter = (e.nativeEvent as SubmitEvent).submitter as HTMLButtonElement | null;
const addAnother = submitter?.value === 'add-another';
```

The "Save & add another" button now carries `value="add-another"` (part of RAC's `AriaButtonProps`, forwarded to the native button), and `addAnotherRef`/the `onPress` setter are gone. A blocked submit no longer leaves stale intent behind, so the copyable pattern is now correct.

docs/content/patterns/user-input/table-record-management/table-record-drawer-crud.demo.tsx
sebald: Resolved in `04fc733e7` by softening the guidance rather than adding the highlight. There's no in-system way to highlight a single row right now: the row `variant` type advertises `muted` but the rui theme only implements it for cells/columns, `Table.Row` strips `className`, and turning on selection forces a checkbox column (`selectionBehavior` is hardcoded to `toggle`). So a real highlight would need either a new DS primitive or custom CSS, neither of which belongs in this pattern PR.

I changed the rule in `index.mdx` from a directive to a recommendation:

> **Help the user keep their bearings while editing.** The table stays visible behind the drawer as a reference. Where your table supports it, highlighting the matching row is a nice touch, but it is a recommendation rather than a requirement.

That removes the contradiction with the demo. A proper row-highlight affordance on `Table.Row` would be a good follow-up DS feature.

sarahgm (approved):

aromko (approved):

PR #5517, opened 2026-06-12
fix(DSTSUP-260): make Pagination follow the controlled page prop

sebald:
# Description

The `page` prop of `Pagination` is documented as controlled, but it was only used to seed internal `useState`. After mount, external updates to `page` (e.g. resetting to page 1 when a filter changes) were silently ignored — the page indicator and the `<nav>` `aria-label` kept announcing the stale page.

`Pagination` now uses `useControlledState` from `@react-stately/utils` (added as a direct dependency, already in the tree transitively): when `page` is provided it is the single source of truth and `onChange` reports requested page changes; `defaultPage` keeps covering the uncontrolled case. Two deliberate behavior refinements come with this:

- `onChange` no longer fires when the already-active page is selected again (matches react-aria conventions).
- The reset-to-page-1 effect now guards on an actual `pageSize` change instead of a first-render flag, so an unstable `onChange` identity can no longer retrigger it.

The inventory example in the docs passed `page={2}` without `onChange` (relying on the old broken seeding behavior) and was switched to `defaultPage={2}`.

No visual changes — Before/After screenshots are not applicable.

Closes DSTSUP-260

## Test Instructions

1. Run `pnpm sb` and open `Components/Pagination` → `Controlled`.
2. Click "Page 3" — the parent state ("Selected Page: 3") and the indicator update.
3. Click "Reset to first page" — the indicator must move back to page 1 (previously it stayed on page 3).
4. Run `pnpm vitest run --config vite.config.ts --project=unit-tests packages/components/src/Pagination/` and `pnpm vitest run --config vite.config.ts --project=storybook-tests packages/components/src/Pagination/Pagination.stories.tsx`.

## Breaking Changes

No — this restores the documented contract. Consumers that passed `page` without ever updating it (as a pseudo-initial value) should switch to `defaultPage`.

## Checklist

### DIFF
--- a/packages/components/src/Pagination/Pagination.stories.tsx
+++ b/packages/components/src/Pagination/Pagination.stories.tsx
@@ -115,6 +115,7 @@ export const Basic = meta.story({
 });
 
 export const Controlled = meta.story({
+  tags: ['component-test'],
   render: ({ totalItems, pageSize, ...rest }: Partial<PaginationProps>) => {
     const [basicPage, setBasicPage] = useState(1);
 
@@ -122,6 +123,7 @@ export const Controlled = meta.story({
       <div>
         <h1>Pagination Example</h1>
         <p>Selected Page: {basicPage}</p>
+        <button onClick={() => setBasicPage(1)}>Reset to first page</button>
         <Pagination
           {...rest}
           totalItems={totalItems!}
@@ -130,6 +132,34 @@ export const Controlled = meta.story({
           onChange={setBasicPage}
         />
       </div>
+    );
+  },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement);
+
+    await step('Selecting a page updates the parent state', async () => {
+      await userEvent.click(canvas.getByLabelText('Page 3'));
+
+      await expect(canvas.getByText('Selected Page: 3')).toBeInTheDocument();
+      await expect(canvas.getByLabelText('Page 3')).toHaveAttribute(
+        'data-selected',
+        'true'
+      );
+    });
+
+    await step(
+      'An external change to the page prop updates the active page',
+      async () => {
+        await userEvent.click(
+          canvas.getByRole('button', { name: 'Reset to first page' })
+        );
+
+        await expect(canvas.getByLabelText('Page 1')).toHaveAttribute(
+          'data-selected',
+          'true'
+        );
+        await expect(canvas.getByText('Selected Page: 1')).toBeInTheDocument();
+      }
     );
   },
 });
--- a/packages/components/src/Pagination/Pagination.test.tsx
+++ b/packages/components/src/Pagination/Pagination.test.tsx
@@ -195,6 +195,40 @@ describe('Pagination tests', () => {
     );
   });
 
+  test('controlled: a pageSize change reports the reset via onChange(1) while the page prop stays the source of truth', () => {
+    const onChange = vi.fn();
+    const { rerender } = render(
+      <Basic.Component
+        totalItems={100}
+        pageSize={10}
+        page={3}
+        onChange={onChange}
+      />
+    );
+
+    expect(screen.getByLabelText('Page 3')).toHaveAttribute(
+      'data-selected',
+      'true'
+    );
+
+    rerender(
+      <Basic.Component
+        totalItems={100}
+        pageSize={20}
+        page={3}
+        onChange={onChange}
+      />
+    );
+
+    // In controlled mode the reset is requested via onChange(1); the parent owns
+    // `page`, so the active page stays 3 until the parent updates the prop.
+    expect(onChange).toHaveBeenCalledWith(1);
+    expect(screen.getByLabelText('Page 3')).toHaveAttribute(
+      'data-selected',
+      'true'
+    );
+  });
+
   test('use control labels', async () => {
     render(<WithButtonLabels.Component />);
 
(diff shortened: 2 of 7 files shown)


packages/components/src/Pagination/Pagination.stories.tsx
```diff
@@ -115,13 +115,15 @@ export const Basic = meta.story({
 });
 
 export const Controlled = meta.story({
+  tags: ['component-test'],
```
sarahgm, line 118: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files (`packages/components/src/Pagination/**`, `*.stories.tsx`) but the `Visual-Regression-Tests` workflow has not run for this branch. Please trigger it by commenting `/run-chromatic` on this PR.

_(I expect a zero-diff result — the change is purely behavioral and the inventory `page={2}`→`defaultPage={2}` edit renders identically — but the run should still confirm before merge into `main`.)_

packages/components/src/Pagination/Pagination.test.tsx
```diff
@@ -99,6 +99,102 @@ describe('Pagination tests', () => {
...
+    rerender(
+      <Basic.Component
+        totalItems={100}
+        pageSize={20}
+        defaultPage={5}
+        onChange={onChange}
+      />
+    );
+
+    expect(onChange).toHaveBeenCalledWith(1);
```
sarahgm, line 191: 💡 This test covers the **uncontrolled** reset path (`defaultPage`). Consider adding a controlled-mode counterpart — set `page` and change `pageSize`, then assert `onChange` is called with `1` — to close the remaining branch in the new `prevPageSize` effect. Low priority.

packages/components/src/Pagination/Pagination.tsx
```diff
@@ -107,12 +100,12 @@ const InnerPagination = ({
...
     <>
       <NavigationButton
-        onClick={() => handlePageChange(Math.max(1, currentPage - 1))}
+        onClick={() => setCurrentPage(Math.max(1, currentPage - 1))}
         aria-label={stringFormatter.format('pagePrevious')}
         isDisabled={isFirstPage}
         controlLabel={controlLabels?.[0]}
         position="left"
-        onKeyDown={handleKeyDown(() => handlePageChange(currentPage - 1))}
+        onKeyDown={handleKeyDown(() => setCurrentPage(currentPage - 1))}
```
sarahgm: 💡 Not introduced by this PR, but worth noting: the keyboard handlers call `setCurrentPage(currentPage - 1)` / `currentPage + 1)` without the `Math.max(1, …)` / `Math.min(totalPages, …)` clamps the `onClick` paths use (lines 103/134). The boundary buttons are `isDisabled` so it isn't reachable in practice — just a tidy-up for next time.

sarahgm (changes requested):
## Automated Code Review

The fix itself is correct and well-tested — `useControlledState` is the right tool, the `prevPageSize` guard is a genuine improvement over the `isFirstRender` flag, the `InnerPaginationProps` narrowing is accurate, and the `@react-stately/utils` dependency is correctly promoted to a direct dep. I verified the other `page={…}` usages in docs/stories are all properly controlled, so nothing else regresses.

The only blocking item is that Visual Regression Tests have not run for this branch (UI-affecting files changed). I expect a zero-diff result, but the run should confirm before merge into `main`. Two low-priority suggestions are inline.

---

packages/components/src/Pagination/Pagination.test.tsx
sebald: Added in 0addf66a5. New test `controlled: a pageSize change reports the reset via onChange(1) while the page prop stays the source of truth` — renders with `page={3}` + `onChange`, bumps `pageSize` 10→20 on rerender, and asserts `onChange` is called with `1` while page 3 stays `data-selected` (the parent owns `page`). Complements the existing uncontrolled `defaultPage` reset test. 18 unit tests pass.

packages/components/src/Pagination/Pagination.tsx
sebald: Done in 0addf66a5 — clamped both keyboard handlers with `Math.max(1, currentPage - 1)` / `Math.min(totalPages, currentPage + 1)` to match the onClick paths. You're right it isn't reachable via the disabled boundary buttons, but it's slightly more than cosmetic now: since `setCurrentPage` routes through `useControlledState`, an unclamped boundary call would emit an out-of-range `onChange(0)` / `onChange(totalPages + 1)`. So it's a cheap defensive fix on top of the consistency win. Tests stay green.

packages/components/src/Pagination/Pagination.stories.tsx
sebald: Agreed — triggering Chromatic now; I'll leave this thread open until it reports back. Expecting a zero/near-diff result as you noted (`page={2}`→`defaultPage={2}` renders identically and the rest is behavioral).

sebald: Chromatic run dispatched: https://github.com/marigold-ui/marigold/actions/runs/27489778503 — leaving this thread open until it reports back.

sebald: VRT reviewed and accepted on Chromatic ([build 58](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=58)): workflow passed (424 stories / 86 components). The flagged changes are baseline drift plus the one expected diff — the new "Reset to first page" button in the `Controlled` story — all accepted. Resolving.

sarahgm (approved):

PR #5519, opened 2026-06-12
feat(DST-1445): surface non-component docs pages in the CLI

sarahgm:
# Description

The Marigold CLI's `docs` and `list` commands previously resolved only component documentation, even though the docs site also publishes Foundations, Patterns, and Getting-Started pages (all already indexed in `manifest.json`). This PR generalizes both commands to reach those pages.

- `marigold docs <name|slug>` now resolves any docs page. Full slugs route by prefix (`components/…` to a component, otherwise a page); bare names try components first, then fall back to pages.
- `marigold list` lists pages under `Foundations`, `Patterns`, and `Getting Started` headings and accepts `--category foundations|patterns|getting-started`.
- `--format json` gains an additive top-level `pages` array; the existing `categories` shape is unchanged.
- Shell completion suggests page slugs and the new page categories.
- `--section props` on a non-component page emits a friendly note instead of failing.

The package README was updated to document the new page support, and a changeset is included.

Closes DST-1445

## Test Instructions

1. Build the CLI: `pnpm --filter @marigold/cli build`
2. `node packages/cli/dist/bin/marigold.mjs docs foundations/spacing` renders the page
3. `node packages/cli/dist/bin/marigold.mjs list --category foundations` (also `patterns`, `getting-started`) lists the pages
4. `node packages/cli/dist/bin/marigold.mjs list --search install` matches `getting-started/installation`
5. `node packages/cli/dist/bin/marigold.mjs docs Button` confirms component docs still work
6. `pnpm --filter @marigold/cli test` all tests pass

## Breaking Changes

No

## Checklist


packages/cli/src/lib/docs.ts
```diff
@@ -45,29 +49,89 @@ export const sliceSection = (
...
+  manifest: Awaited<ReturnType<typeof loadManifest>>['manifest'],
+  input: string
+): Pick<
+  PageDocs,
+  'kind' | 'name' | 'slug' | 'category' | 'description'
+> | null => {
+  if (input.includes('/')) {
+    if (input.startsWith('components/')) {
+      const resolved = resolveComponent(manifest, input);
+      if (!resolved) return null;
```
sebald: **Bug: a `components/…` slug that is a standalone page (not a 3-segment component) is unreachable via `docs`.**

The manifest contains standalone pages whose slug starts with `components/` but has only two segments — e.g. the **`components/form`** overview page (`transformManifest` files anything under `components/` with `slugParts.length < 3` into `manifest.pages` with `category: 'components'`). `marigold list` happily lists it, but `marigold docs components/form` fails because this branch routes every `components/`-prefixed input exclusively through `resolveComponent` and returns `null` on miss — it never falls back to `resolvePage`.

Reproduced against the live manifest on this branch:

```
$ marigold list --category components
Components
  components/form — Wrap your fields to submit user data...

$ marigold docs components/form
No docs page "components/form" found. Did you mean: ComboBox, components/form, foundations/component-principles?
```

Note the "Did you mean" even suggests `components/form` — the exact slug the user typed. A page that `list` advertises should be fetchable by `docs`.

Suggested fix — fall back to the page resolver instead of bailing:

```ts
if (input.startsWith('components/')) {
  const resolved = resolveComponent(manifest, input);
  if (resolved) {
    return {
      kind: 'component',
      name: resolved.component.name,
      slug: resolved.component.slug,
      category: resolved.category.name,
      description: resolved.component.description,
    };
  }
  // fall through to the page resolver below (e.g. `components/form`)
}
```

(then let the shared `resolvePage(...)` block at the end handle it). A regression test for `docs components/form` → `kind: 'page'` would lock this in.

packages/cli/src/lib/manifest.ts
```diff
@@ -92,6 +94,7 @@ const transformManifest = (raw: RawManifest): Manifest => {
       standalonePages.push({
         title: clean(page.name) ?? slug,
         slug,
+        category: slugParts[0] ?? slug,
```
sebald, line 97: **Minor / UX:** because `category` is the top-level slug segment, the `components/form` overview page gets `category: 'components'`, which then surfaces as a standalone **"Components"** group in `marigold list` (and as a valid `--category components` value in completion) — sitting right next to the real component categories (`Actions`, `Form`, …). It reads a bit oddly to have a generic "Components" page bucket alongside the proper component taxonomy.

Not blocking, but worth a decision: either fold `components/`-rooted standalone pages into the component listing, give them a clearer group label, or exclude them. Tied to the `docs` routing gap flagged in `docs.ts`.

packages/cli/src/lib/docs.ts
```diff
@@ -45,29 +49,89 @@ export const sliceSection = (
...
   section?: Section;
 }
 
-export const getComponentDocs = async (
+// Resolve the positional query to either a component or a standalone page.
+// Full slugs route by prefix (components/… → component, else page); bare
+// names try the component cascade first, then fall back to pages.
+const resolveQuery = (
+  manifest: Awaited<ReturnType<typeof loadManifest>>['manifest'],
+  input: string
```
sebald: **Nit:** `Awaited<ReturnType<typeof loadManifest>>['manifest']` is a roundabout way to spell `Manifest`, which is already imported in this file. `manifest: Manifest` would read much more directly.

sebald (commented):
## Code review — DST-1445 (CLI non-component docs pages)

Overall this is a clean, well-tested change that delivers the ticket's scope. The resolver cascade, the additive `pages` JSON shape, the friendly `--section props` note, and the completion updates all match the spec, and the test coverage is genuinely good (139 tests pass; build + lint clean). I verified the behaviour end-to-end against the live docs site — `docs`/`list` for foundations, patterns, and getting-started all work, including bare-name fallback, `--search`, suggestions, and JSON output.

One functional gap worth fixing before merge, plus two minor notes (inline):

1. **🐞 `docs components/form` is unreachable** (`docs.ts`) — a standalone page whose slug starts with `components/` but isn't a 3-segment component is listed by `list` yet can't be fetched by `docs`, because the `components/`-prefixed branch never falls back to `resolvePage`. Repro + suggested fix inline. This is the only thing I'd consider blocking.
2. **UX:** the same page produces a generic "Components" group in `list` / completion alongside the real component categories (`manifest.ts`).
3. **Nit:** convoluted `Awaited<ReturnType<...>>` type that could just be `Manifest` (`docs.ts`).

Against the ticket's "Included" scope, everything is covered — the only caveat is that the resolver generalization isn't *fully* general for `components/`-rooted non-component pages (item 1).

sebald:
## Please address the `docs components/form` routing gap before merge 🙏

Great PR — scope from DST-1445 is covered and the test coverage is solid. I exercised the commands end-to-end against the live docs site (foundations / patterns / getting-started all resolve, bare-name fallback, `--search`, `--section props` note, JSON output, completion, and typo suggestions all behave). One functional gap is worth fixing first.

### Finding

The resolver generalizes component-vs-page routing, but **not fully for `components/`-rooted standalone pages**. `transformManifest` files anything under `components/` with fewer than 3 slug segments into `manifest.pages` (with `category: 'components'`) rather than the component taxonomy. The live manifest has exactly one such page today — **`components/form`** (the Form overview page).

`marigold list` advertises it, but `marigold docs components/form` can't fetch it, because `resolveQuery` routes every `components/`-prefixed input solely through `resolveComponent` and returns `null` on miss instead of falling back to `resolvePage`:

```
$ marigold list --category components
Components
  components/form — Wrap your fields to submit user data...

$ marigold docs components/form
No docs page "components/form" found. Did you mean: ComboBox, components/form, foundations/component-principles?
```

Note the "Did you mean" even echoes `components/form` — the exact slug typed. A page `list` shows should be fetchable by `docs`.

### Suggested fix (`packages/cli/src/lib/docs.ts`, the `input.startsWith('components/')` branch)

Don't bail on a `resolveComponent` miss — fall through to the page resolver:

```ts
if (input.startsWith('components/')) {
  const resolved = resolveComponent(manifest, input);
  if (resolved) {
    return {
      kind: 'component',
      name: resolved.component.name,
      slug: resolved.component.slug,
      category: resolved.category.name,
      description: resolved.component.description,
    };
  }
  // fall through to resolvePage below (e.g. `components/form`)
}
```

A regression test asserting `docs components/form` → `kind: 'page'` would lock it in. (See the inline comments for the related "Components" list-group UX note and a small type nit.)

packages/cli/src/lib/docs.ts
```diff
@@ -45,29 +49,89 @@ export const sliceSection = (
...
-export interface GetComponentDocsOptions extends CacheOptions {
+export interface GetPageDocsOptions extends CacheOptions {
   section?: Section;
 }
 
-export const getComponentDocs = async (
+// Resolve the positional query to either a component or a standalone page.
+// Full slugs route by prefix (components/… → component, else page); bare
+// names try the component cascade first, then fall back to pages.
+const resolveQuery = (
```
sebald: **Simplify (reuse/DRY):** `resolveQuery` spells out the same `PageDocs` mapping literal **four times** — twice for components (the `components/` branch + the bare-name branch) and twice for pages. Two tiny adapters collapse the whole function and make the routing intent obvious:

```ts
const fromComponent = (r: ResolveResult): QueryResult => ({
  kind: 'component',
  name: r.component.name,
  slug: r.component.slug,
  category: r.category.name,
  description: r.component.description,
});

const fromPage = (p: ManifestPage): QueryResult => ({
  kind: 'page',
  name: p.title,
  slug: p.slug,
  category: p.category,
  description: p.description,
});

const resolveQuery = (manifest: Manifest, input: string): QueryResult | null => {
  if (input.startsWith('components/')) {
    const c = resolveComponent(manifest, input);
    if (c) return fromComponent(c);
    // fall through to page lookup (e.g. `components/form`)
  }
  const c = resolveComponent(manifest, input);
  if (c) return fromComponent(c);
  const p = resolvePage(manifest, input);
  return p ? fromPage(p) : null;
};
```

This also makes the `components/form` routing fix I flagged in the other review a non-event — there's only one page-fallback path to maintain. (`QueryResult` = the `Pick<PageDocs, …>` you already define.)

packages/cli/src/lib/format.ts
```diff
@@ -123,20 +167,31 @@ export const formatList = (
     }))
     .filter(cat => cat.components.length > 0);
 
+  const filteredPages = manifest.pages.filter(p =>
+    matchesPageFilter(p, filter)
+  );
+  const pageGroups = groupPages(filteredPages);
```
sebald: **Simplify (efficiency/altitude):** `groupPages(filteredPages)` runs here, but the `format === 'json'` branch right below returns without ever using `pageGroups` — so it's wasted work on the JSON path. Move the grouping below the JSON early-return (the JSON branch only needs `filteredPages`):

```ts
const filteredPages = manifest.pages.filter(p => matchesPageFilter(p, filter));

if (format === 'json') {
  return JSON.stringify({ /* … uses filteredPages … */ });
}

const pageGroups = groupPages(filteredPages);
```

Minor, but it keeps the cost on the path that actually consumes it.

packages/cli/src/lib/manifest.ts
```diff
@@ -210,3 +213,56 @@ export const suggestComponents = (
 
   return scored;
 };
+
+// Resolve a standalone (non-component) page using the same match cascade as
+// resolveComponent, matching against the page title and slug.
+export const resolvePage = (
```
sebald, line 237: **Simplify (reuse — lower confidence, optional):** `resolvePage` is a near-verbatim copy of `resolveComponent`'s 4-step cascade (exact → case-insensitive → normalized → slug-tail), and `suggestPages`/`suggestComponents` share the same scoring shape. The only real difference is the field accessor (`p.title`/`p.slug` vs `c.name`/`c.slug`) and that components are nested under categories while pages are flat.

A small generic could remove the duplication:

```ts
const matchCascade = <T>(items: T[], input: string, fields: (t: T) => string[]): T | null => { … };
// resolveComponent → matchCascade(flat, input, c => [c.name, c.slug])
// resolvePage      → matchCascade(pages, input, p => [p.title, p.slug])
```

Same idea applies to `matchesPageFilter` vs `matchesFilter` in `format.ts`. Worth weighing against the indirection cost — the cascades are short, so this is a judgment call, not a must-fix. Flagging it because the page-side helpers landed as straight copies.

sebald (commented):
## `/simplify` dry-run — quality-only pass (no code changed)

Reuse / simplification / efficiency / altitude review of the diff. **No correctness bugs here** (those are in the separate review) — these are cleanups, none blocking:

1. **`docs.ts` — `resolveQuery` repeats the `PageDocs` mapping literal 4×.** Two `fromComponent` / `fromPage` adapters collapse the function and incidentally make the `components/form` fix trivial. *(highest value)*
2. **`format.ts` — `groupPages()` runs before the JSON early-return** that doesn't use it; move it below. *(efficiency nit)*
3. **`manifest.ts` — `resolvePage`/`suggestPages` are near-copies of the component equivalents** (and `matchesPageFilter` mirrors `matchesFilter`); an optional shared matcher would dedupe them. *(lower confidence — judgment call)*

Details inline. Item 1 is the one I'd actually act on; 2 and 3 are take-or-leave.

packages/cli/src/lib/docs.ts
sarahgm: Fixed in `0a2349b06`. `resolveQuery` now runs the component cascade first and **falls through to `resolvePage`** on a miss, so `docs components/form` resolves as `kind: 'page'`. Added a regression test in `docs.test.ts` (seeded `components/form` fixture → asserts `kind: 'page'`).

packages/cli/src/lib/docs.ts
sarahgm: Done in `0a2349b06` — extracted `fromComponent`/`fromPage` adapters and collapsed `resolveQuery` to a single component-then-page path. As you noted, this made the `components/form` routing fix fall out for free (one page-fallback path).

packages/cli/src/lib/docs.ts
sarahgm: Done in `0a2349b06` — the param is now typed as the imported `Manifest`.

packages/cli/src/lib/format.ts
sarahgm: Done in `0a2349b06` — `groupPages(filteredPages)` now runs *below* the `format === 'json'` early-return, so it's off the JSON path.

packages/cli/src/lib/manifest.ts
sarahgm: Addressed in `0a2349b06` — added a `PAGE_GROUP_LABELS` override mapping the `components` page-group to **"Form Overview"**, so it no longer reads as a generic "Components" bucket next to the real taxonomy. The page stays listed and is now fetchable via `docs`. Display-only: the `--category components` completion token is unchanged, so the label and filter token diverge slightly — happy to revisit if you'd prefer renaming the underlying category instead.

packages/cli/src/lib/manifest.ts
sarahgm: Done in `0a2349b06` — added a generic `matchCascade<T>(items, input, name, slug)`; `resolveComponent` and `resolvePage` now both delegate to it. Also extracted `matchesListFilter` so `matchesFilter`/`matchesPageFilter` are thin callers. Left `suggestComponents`/`suggestPages` as-is — folding their scoring didn't come out clean enough to justify the indirection.

sarahgm:
Thanks for the thorough review @sebald 🙏 All points addressed in `0a2349b06`:

- 🐞 **`docs components/form` unreachable** — `resolveQuery` now falls through to the page resolver on a component-cascade miss. Regression test added.
- **DRY `resolveQuery`** — `fromComponent`/`fromPage` adapters; single page-fallback path.
- **Type nit** — param is now `Manifest`.
- **`groupPages` efficiency** — moved below the JSON early-return.
- **"Components" group UX** — relabeled to **"Form Overview"** via a `PAGE_GROUP_LABELS` override (display-only; see inline note).
- **Generic matcher** — added `matchCascade` (shared by `resolveComponent`/`resolvePage`) and `matchesListFilter`.

Verification: 141 tests pass (139 + 2 new), build clean, lint 0 errors, prettier applied. Replies inline on each thread.

sebald (approved):
All review findings from the earlier code review and `/simplify` pass are integrated in 0a2349b06:

1. 🐞 `docs components/form` now resolves as `kind: 'page'` — `resolveQuery` runs the component cascade then falls through to `resolvePage`. Regression test added.
2. UX: `PAGE_GROUP_LABELS` relabels the components page-group to "Form Overview".
3. `Manifest` type used directly instead of `Awaited<ReturnType<…>>`.
4. `fromComponent`/`fromPage` adapters collapse the repeated mapping literal.
5. `groupPages` moved below the JSON early-return.
6. Generic `matchCascade<T>` shared by `resolveComponent`/`resolvePage`; `matchesListFilter` shared by the filters. The generic matcher correctly matches `components/`-rooted standalone pages since `resolvePage` feeds it `manifest.pages`.

The `--category components` token vs "Form Overview" label divergence is acknowledged and fine to leave as-is. LGTM 🚀

PR #5520, opened 2026-06-12
fix(DST-1507): make Table.EditableCell inline editing work after SSR

sebald:
# Description

`Table.EditableCell` inline editing was **inert after SSR hydration**. In a server-rendered app (Next.js, including our own docs under `next dev`), clicking an editable cell or its pencil button did nothing and no editor opened — with **no hydration mismatch warning**. The cell only became interactive after an unrelated re-render (e.g. a window resize) forced the table collection to rebuild.

**Root cause:** React Aria builds the table collection in a *separate* render pass from the one that renders the visible cells. The editing state (`open`) previously lived in `TableEditableCell`, which renders in that **build** pass. The actual `<td>` content — the `onClick`/`onPress` → `setOpen` handlers and the Popover/Dialog — is rendered from the collection `Document` in the **content** pass. During SSR the build runs synchronously; after hydration React Aria switches to client/portal rendering, but the already-rendered cell content stays bound to the **server pass's closures** and never re-registers, so `setOpen` calls are no-ops. (`react-aria-components@1.18.0` is the latest published version — no upstream fix.)

**Fix:** adopt the same structure React Spectrum's S2 `TableView` uses — keep `TableEditableCell` **stateless** and move the editing state + overlay into a `TableEditableCellInner` that is rendered **inside the `Cell`'s render prop**. That inner component lives in the collection's *content* pass (the live, portal-rendered tree), so its state and the handlers/overlay it renders are co-located in one instance and reconnect on their own after hydration — no forced re-render or `clientOnly` escape hatch needed. `useTableOptions` stays in the outer cell (it is only available in the build pass) and passes `hasSelection` down.

This supersedes the docs-only `clientOnly` workaround (DST-1506, closed as Won't do); that workaround is being removed from the docs separately on the DST-1288 branch.

Closes DST-1507

## Test Instructions

1. Run the docs and open the **Table** component page, **Inline editing** section (`/components/table#inline-editing`).
2. On first load — no resize, no prior interaction — click a cell in the inline-editing table.
3. Confirm the inline editor opens immediately, you can edit the value (Name / Capacity / Status), and saving updates the table. No console errors.
4. `pnpm test:unit --run packages/components/src/Table` and `pnpm test:sb --run packages/components/src/Table/Table.stories.tsx` — all green (42 unit + 14 story tests).

> **Automated regression coverage:** `TableEditableCell.ssr.test.tsx` reproduces the bug directly — it `renderToString`s the table, `hydrateRoot`s that exact server markup, then clicks a server-rendered trigger and asserts the editor opens. The other editable-cell tests `createRoot` (client render), which never exercises React Aria's collection build → content → hydration transition; this one does. Verified it fails on the pre-fix component and passes on the fix.
>
> Also verified manually in a real SSR `next dev` build.

## Breaking Changes

No

## Checklist

- [ ] Visual regression tests updated (for UI changes) — N/A, no visual change


aromko (commented):
The path in the test instructions doesn't exsists - which one is it?
Wrong: /patterns/user-input/table-record-management

sebald:
Good catch, thanks @aromko — that path was wrong, there's no such pattern. The inline-editing demo lives on the **Table** component page in the **Inline editing** section. The `collection` folder is flattened in the docs routing, so the correct URL is `/components/table#inline-editing`. I've updated step 1 of the Test Instructions accordingly.

aromko (approved):

PR #5522, opened 2026-06-12
docs(DST-1434): promote Filter pattern from alpha to beta

sebald:
# Description

Resolves every open TODO on the Filter pattern page and promotes it from `alpha` to `beta`.

> [!IMPORTANT]
> **Depends on #5669 (OverflowRegion) — merge that first.**
> The filter bar now stays a single row by composing its quick filters inside the new `OverflowRegion` component from #5669, which hides trailing quick filters into the "All filters" panel as the bar narrows instead of wrapping to a second line. To build the example and the Bar Layout demo, #5669 is merged into this branch, so its commits currently ride along here. Please review and merge #5669 first, after which this branch reduces to just the filter changes. One small fix to `OverflowRegion` also lives here (it belongs conceptually to #5669): demoted items are now hidden with `opacity-0` on top of `invisible`, because react-aria date segments re-assert `visibility: visible` on themselves and would otherwise keep painting a ghost field once demoted.

**Guidance.** The page is restructured into task-focused chapters: Structure (themed inline SVG anatomy), Choosing Filters, Filter Types, Applying (instant bar vs batched panel, result count on Apply), Designing Options (many options, option counts with per-option query semantics, select all, exclusions), Scaling Up (quick filters, single-row bar layout, labeling, date presets, thematic grouping), Small Screens, Filter State (applied labels, active count, reverting, empty states), Accessibility, and Implementation.

Notable decisions, each research-backed and cited inline (Baymard, NN/g, Polaris, Carbon, Semrush Intergalactic, PatternFly, Helios):

- A `SegmentedControl` is framed as a **scope switch**, not a quick filter: it answers which set the user is viewing, sits at the start of the bar apart from the field cluster, and its label names the view.
- Date quick filters use the built-in **DateRangePicker quick select presets** (DST-1511) instead of a userland Select.
- Panel grouping is **thematic**: sections bundle several related filters and never wrap a single filter, whose long option list is a Many Options problem instead.
- The **filter bar stays a single row** with explicit widths and no visible labels. It uses [`OverflowRegion`](https://www.marigold-ui.io/components/layout/overflow-region) (from #5669) to measure the row and demote quick filters back into the panel when space runs out, so the bar never wraps.
- The Implementation chapter documents the non-obvious practices from `/examples/filter`: URL persistence, pagination reset, re-keying the uncontrolled form, **previewing the draft** (debounced FormData snapshots feeding one `pageSize: 1` + facets query), sentinel "no filter" values, and nuqs `history: push` vs `replace`.

**Reference implementation** (`/examples/filter`). Extended so every row of the Filter Types table has a live counterpart: three quick filters in the bar (multi-select Type with count badge, single-select Rating, Availability via DateRangePicker presets), a searchable **City** TagField and an **Availability** date range in the panel (canonical copies of the bar shortcuts), live option counts and section badges driven by the draft preview, a result-count Apply button, and grouped panel sections matching the thematic guidance.

Follow-up tickets: [DST-1509](https://reservix.atlassian.net/browse/DST-1509) (add `collapseAt` to `Tag.Group` for applied-filter overflow), linked from the prose.

Closes DST-1434

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run the docs app (`pnpm start`) and open `/patterns/user-input/filter`.
2. Confirm the page badge reads **beta** and the chapters match the description above.
3. Open `/examples/filter`: the bar shows search, Type (count badge), Rating, Availability (with preset list in the popover), and "All filters". Quick filters apply instantly and stay in sync with the panel.
4. Open "All filters": Essentials holds Type, City (searchable), Available between, Min. Capacity, Max. Price. Changing a draft updates the option counts, section badges, and the "Show N results" button before committing.
5. Apply filters and verify applied tags (one per filter, ranges as one tag), the badge on the filter button, URL persistence, and pagination reset.
6. Run `pnpm --filter @marigold/docs test` (route tests cover city and availability filtering).

## Breaking Changes

No

## Checklist

- [x] Unit tests added/updated (venues route tests for city and availability)
- [ ] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) (for new/changed interactive components) — N/A, no new components
- [ ] Visual regression tests updated (for UI changes) — N/A, docs and example app only
- [x] Changeset added (`pnpm changeset`) — `@marigold/docs` patch for the pattern, `@marigold/system` patch for the formatter hydration fix

[DST-1509]: https://reservix.atlassian.net/browse/DST-1509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sebald:
### 📋 Follow-up ToDo — filter a11y hardening (don't forget)

Surfaced while verifying the applied-filters behavior in #5584 against UX/a11y best practices. Logging it here on the Filter-pattern PR so it isn't lost:

- [ ] **Make the venues table's `role="status"` live region a single, always-mounted node.** Today the count region (`<VisuallyHidden role="status">{summary}</VisuallyHidden>` in `VenuesTable`) is swapped between the loading-skeleton subtree and the loaded subtree. Recreating a live-region container (rather than repopulating a pre-placed one) is a known failure mode where **NVDA+Chrome / JAWS+Chrome drop the announcement**, so "Loading…" and the result count may not be read aloud reliably.
  - **Fix:** keep one pre-placed, visually-hidden `role="status"` (`aria-live="polite"`) node mounted at all times across loading / loaded / filtered states, and only update its **text content** — never unmount/remount the container.
  - **Refs:** [W3C ARIA22](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22) · [TetraLogical – why are my live regions not working](https://tetralogical.com/blog/2024/05/01/why-are-my-live-regions-not-working/) · [Sara Soueidan – accessible notifications with ARIA live regions](https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-2/).
  - **Severity:** behavioral / AT-specific, **not** a hard WCAG 4.1.3 failure — nice-to-have hardening, not a blocker.

docs/content/patterns/user-input/filter/filter-anatomy.tsx
```diff
@@ -0,0 +1,430 @@
+export const FilterAnatomy = () => (
+  <svg
+    viewBox="0 0 820 470"
```
aromko, line 3: ⚠️ **Anatomy SVG has no accessible name**

The old `<ImageZoom>` carried `alt="Filter Anatomy"`. This inline SVG exposes neither a name nor a hidden marker, so screen readers encounter the five stray `<text>` fragments ("Filter Bar", "Applied Filter", ...) without context.

Since the bullet list right below the diagram describes every area, the lightest fix is to hide it:

```suggestion
    viewBox="0 0 820 470"
    aria-hidden="true"
```

(Alternatively `role="img"` plus an `aria-label` if you want it announced.)

docs/lib/data/venues.ts
```diff
@@ -255,10 +255,31 @@ export const venues = [
...
+// dates for the availability filter and its relative presets. Spread across
+// the next 12 weeks so every preset range matches at least some venues. Only
+// the API's server-computed values are ever rendered, so the client copy of
+// this module never causes hydration mismatches.
+const DAY_MS = 24 * 60 * 60 * 1000;
+
+export const venues = rawVenues.map((venue, index) => ({
+  ...venue,
+  /** Next available date as an ISO date (`YYYY-MM-DD`). */
+  nextAvailable: new Date(Date.now() + ((index * 11) % 84) * DAY_MS)
```
aromko: 💡 **UTC vs. local-timezone edge in derived dates**

`toISOString().slice(0, 10)` yields a UTC date, while the DateRangePicker presets ("Next 7 days", "This month") resolve in the visitor's local timezone. For users west of UTC (or near midnight), the earliest venue can fall just outside a preset that should include it. Demo-only nitpick; if you want it exact, derive the ISO string from local date parts instead of `toISOString()`.

aromko (commented):
## Automated Code Review

Thorough, research-backed promotion. Every DST-1434 item is resolved or explicitly dispositioned: all four inline TODOs got real demos, the trailing TODO list is answered in prose (including a well-reasoned rejection of RAC `Toolbar` semantics for the filter bar), the label-helper drop is noted in the docs as required, and the four `/examples/filter` best practices are now documented. The facet-count semantics (own-dimension ignored for match-any, preview-add for match-all) are both explained in the docs and covered by route tests.

The `suppressHydrationWarning` on `DateFormat`/`NumericFormat` matches the established pattern for expected Intl server/client mismatches, and pinning `I18nProvider locale="en-US"` in `ComponentDemo` addresses the root cause for demos.

Two small findings, both non-blocking:

1. ⚠️ The new inline anatomy SVG lost the accessible name the old `alt="Filter Anatomy"` image had; add `aria-hidden="true"` (or `role="img"` with a label). See inline comment.
2. 💡 `nextAvailable` derives from UTC while the date presets resolve in local time, so edge-of-day users can see one venue slip out of "Next 7 days". Demo-only. See inline comment.

Note: the `Vercel Preview Comments` check is failing, meaning there is unresolved feedback on the preview deployment to clear before merge.

**Verdict: Comment (approve-worthy, one one-line a11y fix requested)**

---

docs/content/patterns/user-input/filter/filter-anatomy.tsx
sebald: Good catch — hidden it with `aria-hidden="true"` since the bullet list right below describes each area. Fixed in ef0cf105b.

docs/lib/data/venues.ts
sebald: Fixed in ef0cf105b — `nextAvailable` now derives from local date parts instead of `toISOString()`, so the derived day matches the local timezone the DateRangePicker presets resolve in.

aromko (approved):

sebald:
### 📌 Reminder: rewire the filter bar onto `<OverflowRegion>` here

The "demote quick filters, don't wrap" behavior this pattern prescribes now has a real primitive: **`<OverflowRegion>`** landed in #5669 (DST-1646, beta). Before this PR merges, wire the reference implementation onto it so the live example matches the guidance instead of just describing it:

- [ ] `/examples/filter` — put the quick filters inside an `<OverflowRegion>` (pinned Search + "All filters" trigger as siblings in a `<Inline noWrap>` row), so they demote into the panel as width runs out rather than wrapping. The panel stays canonical (renders every filter), so demotion is lossless.
- [ ] The `filter-bar-layout` / `filter-promoted` docs demos — same treatment.
- [ ] Keep the badge on "All filters" counting **applied** filters (not hidden/demoted ones) — see #5669 for the rationale.

**Sequencing:** `OverflowRegion` ships from `@marigold/components` in #5669 (same `beta-release` base). Merge/rebase this branch after #5669 so the import resolves.

No separate ticket for this — it lives here. The original DST-1646 scope also floated a bespoke `<FilterBar>` **compound component**; we intentionally did *not* build one, since the bar composes cleanly from `OverflowRegion` + existing components (see the filter-bar recipe in #5669's docs/stories). We can revisit a compound later only if real usage shows people re-implementing the same composition. Context: [DST-1646](https://reservix.atlassian.net/browse/DST-1646).

[DST-1646]: https://reservix.atlassian.net/browse/DST-1646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

aromko (approved):

docs/app/(examples)/examples/filter/filter.marigold-pattern.yaml
```diff
@@ -27,10 +31,11 @@ mock_data:
         traits: string[];
         amenities: number[];
         parking: number[];
+        seatingTypes: number[];
```
sarahgm, line 39: ⚠️ The `mock_data` stub is missing two symbols that `key_files` rely on, so the standalone/registry snippet for this example is type-incomplete:

- `export const venueCities: readonly string[];` — imported by `FilterBar.tsx`.
- `nextAvailable: string;` on the `venues` object — read by `VenuesTable.tsx` and `csv.ts`.

The live docs build is unaffected (it imports the real module); this only affects the copyable snippet. The suggestion adds the field; `venueCities` still needs adding alongside the other `export const … : readonly string[];` lines.

```suggestion
        seatingTypes: number[];
        nextAvailable: string;
```

packages/system/src/components/Formatters/DateFormat.tsx
```diff
@@ -18,7 +18,8 @@ export const DateFormat = ({ value, tabular, ...props }: DateFormatProps) => {
   });
 
   return (
-    <span className={tabular ? 'tabular-nums' : ''}>
+    // Intl output can legitimately differ between server and client ICU/locale.
+    <span suppressHydrationWarning className={tabular ? 'tabular-nums' : ''}>
```
sarahgm, line 34: ⚠️ `suppressHydrationWarning` here (and the matching change in `NumericFormat.tsx`) suppresses hydration warnings for **every** `DateFormat`/`NumericFormat` across the design system, which could mask genuine future mismatches. The rationale (Intl locale/ICU output can legitimately differ server vs client) is sound, and demos are already pinned via `I18nProvider locale="en-US"` in `ComponentDemo.tsx`. Since this lands in a shared package, worth a conscious sign-off that the global suppression is preferred over a narrower fix.

sarahgm (commented):
Automated review (Claude Code `review-pr`). Strong docs promotion — DST-1434 acceptance criteria met, CI green, VRT not required for a `beta-release` target. Two non-blocking points inline; a couple of minor suggestions were kept out to stay high-signal. Recommendation: **Approve**.

docs/app/(examples)/examples/filter/filter.marigold-pattern.yaml
sebald: Fixed in 3ba38a286, and thanks — chasing this turned up something much worse sitting underneath it.

**Both symbols added**, and a sweep for the same class of gap found two more. `Venue` is imported as a type by `csv.ts:2` and `VenuesTable.tsx:8` — both `key_files` — and wasn't declared either; `MAX_CAPACITY`/`MAX_PRICE` are imported by `hooks/useFilter.ts`. So the stub now adds `venueCities`, `nextAvailable`, `MAX_CAPACITY`, `MAX_PRICE` and `export type Venue`.

**`@/lib/data/venues-query` had no `mock_data` entry at all**, despite `FilterBar.tsx:11` importing `VenueFacets` from it. Added, with `VenueQueryParams`, `VenueQueryResult`, `VenueFacets` and the two sort unions.

### The bigger problem

While checking which files needed which symbols I found that `build-examples.mjs` only read the **top level** of an example folder:

```js
// All examples are flat by convention; if one ever nests source in a subfolder,
// those files are not collected. Make this a recursive walk if that convention changes.
```

That convention stopped holding when this example moved its ten hooks into `hooks/`. The shipped payload was 7 files, and **all six key files import from `./hooks/…`** — so the snippet referenced ten modules that weren't in it. The stub being type-incomplete was real, but it was the smaller half: the snippet couldn't be built at all. `bulk-actions` was in the same state, missing 8 of its 16 files.

The walk is recursive now, paths POSIX-relative so they match the import specifiers, and `key_files` may point into a subdirectory. `filter` ships 17 files, `bulk-actions` 16.

### Verifying rather than eyeballing it

The point of your comment is that the snippet should typecheck, so I checked that directly instead of re-reading the stub: materialised the built payload into a temp project, wrote each `mock_data` shape at its aliased path as a `.d.ts`, and ran the repo's `tsc` over it. The only unresolved modules left are external peer deps — nothing under `@/lib/data/`. Then I mutation-tested each symbol by deleting it and confirming `tsc` complains:

```
src/FilterBar.tsx(7,3): error TS2305: Module '"@/lib/data/venues"' has no exported member 'venueCities'.
src/csv.ts(2,15): error TS2724: '"@/lib/data/venues"' has no exported member named 'Venue'.
src/hooks/useFilter.ts(1,10): error TS2305: ... has no exported member 'MAX_CAPACITY'.
```

That harness also caught `peer_deps` being wrong in both directions: `@internationalized/date`, `@tanstack/react-query` and `react-error-boundary` are imported but weren't declared, and `zod` was declared but appears nowhere in the example except that line. Fixed.

### Why the tests didn't catch any of this

`build-examples.test.mjs` has never run. `docs/vitest.config.mjs` and `docs/vitest.config.ts` both existed; vitest resolves the `.ts`, whose `include` covers only `app/` and `lib/`. So `pnpm --filter @marigold/docs test` — what the **Docs Script Tests** CI job runs — passed on an empty selection, and has done since DST-1421 added those tests. Merged into one config, keeping the `.ts` because the app tests need its `@` alias (the `.mjs` one lacks it, so they fail under it). The suite goes from 8 files/32 tests to 10/87.

There's a test for the nested walk and one for a subdirectory `key_files` ref, plus the old `collects only top-level` assertion inverted.

### One thing I left alone

`bulk-actions` has the same two sidecar gaps — no `@/lib/data/events-query` stub, missing `Event`/`EventStatus`, and the identical `peer_deps` drift. It's **pre-existing**, not something the walk change causes: `BulkEditDrawer.tsx` and `EventsPagination.tsx` are top-level key files that already imported the unstubbed module. Different example, so I'd rather not fold it into this PR — worth its own ticket, and the tsc harness above transfers to it directly.

packages/system/src/components/Formatters/DateFormat.tsx
sebald: Signing off on the global suppression, and I've written the reasoning into both files in 3ba38a286 so the next reader doesn't have to re-derive it from a one-line comment.

The thing that makes me comfortable with it being global is that **it isn't actually global in effect**. `suppressHydrationWarning` only works one level deep — React is explicit that it's an escape hatch scoped to the element's own attributes and text children, and it does not recurse into composed children. Both formatters render exactly one text child, and that child is entirely `Intl` output. So the blast radius is precisely the string that legitimately varies; it cannot mask a mismatch in anything rendered inside or around a `DateFormat`.

The narrower alternatives are all worse here:

- **An opt-in prop.** The mismatch is the *default* for any SSR consumer who hasn't pinned a locale — react-aria reads `navigator.language` in the browser and falls back to `en-US` on the server without an `I18nProvider`, so a German visitor gets a mismatch on every date. A prop would just make each consumer hit the warning and go find the prop.
- **Deferring render until mount** (`useIsSSR` or a mounted flag). Removes dates and numbers from the SSR HTML — layout shift, worse for no-JS, and a design system shouldn't impose that.
- **Pinning a locale server-side.** Not a library's call, and react-aria already owns locale resolution.

Worth adding that locale detection isn't the only source: Node's ICU can disagree with the browser's for the *same* locale, which shows up as currency spacing (NBSP vs narrow NBSP) and range separators. So even a fully pinned `I18nProvider` app can mismatch, which rules out "just tell consumers to pin the locale" as a substitute.

The one genuine risk your comment points at, and the reason I wrote it down rather than just replying: because the suppression is one level deep, someone later wrapping the formatted value in another element — a `<bdi>`, or splitting into `formatToParts` spans — would silently move the real output out from under it. Both comments now say so explicitly.

The residual cost is that `<DateFormat value={new Date()} />` won't warn. That was already true before this change for anyone who'd pinned a locale, and it's a misuse a hydration warning was never a reliable guard against.

sebald:
@sarahgm — following up because this merged (`012a02506`) with your `suppressHydrationWarning` thread resolved by me, as the PR author, rather than by you. You raised it, I argued for keeping the suppression, and then I marked my own answer as settled. Flagging that explicitly so it stays your call rather than something that quietly closed: if you're not satisfied, reopen it and I'll revert the formatter change on its own — it's independent of the rest of the Filter work.

The reasoning now lives in the code rather than only in this thread, in `DateFormat.tsx` and `NumericFormat.tsx`. Short version: the suppression is one level deep, so it covers each span's own attributes and its single text child — which is entirely `Intl` output — and nothing composed inside. Both comments also record the failure mode your point implies: if anyone later wraps the formatted value in another element, the real output moves out from under the suppression and the guard silently stops applying.

Two things you should weigh that argue for your caution rather than mine:

- **VRT never ran on this branch.** We merged without dispatching Chromatic, so the formatter change hasn't been visually verified anywhere. Since these two components render across the whole system, that's the least-covered part of the PR. The next `beta-release` run is the first real signal.
- **Re-reading these two files turned up a second, unrelated bug**, which predates this PR: both document `tabular` as `@default true`, but only `NumericFormat` actually defaulted it. `DateFormat` left it `undefined`, so every one of the 38 `DateFormat` call sites in the repo — none of which passed the prop — rendered without `tabular-nums`, while the docs site told consumers otherwise. Filed as DST-1666 and fixed in #5687. Worth a look from you since it lands a visible digit-width change everywhere `DateFormat` is used, and Chromatic hasn't run there either yet.

PR #5527, opened 2026-06-15
feat(DST-1421): add marigold examples CLI commands

sarahgm:
# Description

Adds `marigold examples list` and `marigold examples get <slug>` so AI agents (and humans) can discover and retrieve Marigold's application-level reference patterns from the terminal, mirroring the library-first architecture of the existing `docs`/`list` commands. The unit is the **pattern**, not the file: the CLI ships the source plus the metadata an agent needs to adapt it (mapped pattern docs, mock-data shapes, key vs. scaffolding files, peer deps), and adaptation is delegated to the agent.

- **CLI:** new `listExamples()` / `getExample(slug)` library functions and an `examples` command wrapping them, with `--format markdown|json|plain`, `--fresh` and `--offline` (reusing the existing cache layer). Tab completion and telemetry are extended to cover the new command and example slugs. `suggestComponents` is refactored onto a shared `suggestByScore` helper.
- **Docs:** a new `build-examples.mjs` registry step emits `public/mcp/examples.json` and `public/mcp/examples/<slug>.json` from colocated `*.marigold-pattern.yaml` sidecars. Examples are discovered by sidecar presence (placeholder pages are excluded automatically), and a malformed/schema-invalid sidecar fails the build. Sidecars ship for the `filter`, `form` and `inventory` examples.
- A framework-transformation note (`getting-started/examples-for-agents`) documents porting examples from the Next.js App Router to other frameworks (Vite, etc.) once, rather than per example.

Closes DST-1421

## Test Instructions

1. `pnpm install && pnpm --filter @marigold/cli build`
2. Build the registry assets: `pnpm --filter @marigold/docs build:examples` — confirm it reports `3 examples` and skips the placeholder folders.
3. Run the CLI against the generated payloads (or the live docs site): `marigold examples list`, then `marigold examples get filter` and `marigold examples get filter --format json`.
4. Run the suites: `pnpm --filter @marigold/cli test` (140 tests), `pnpm typecheck:only`, `pnpm lint`.

## Breaking Changes

No

## Checklist

- [ ] Stories added/updated (with \`component-test\` tag where applicable)
- [ ] Visual regression tests updated (for UI changes) — N/A, no UI/component/theme changes


jim761 (dismissed):

jim761 (changes requested):
## Requesting changes — `examples` telemetry events are rejected by the server

This branch wires up the CLI side of telemetry correctly (`CommandName` includes `'examples'`, `bin/marigold.ts` sets the telemetry args, `emit()` fires), but the docs `/api/telemetry` route was **not** updated to accept the new command — so running `marigold examples …` from this branch never records a telemetry event.

`docs/app/api/telemetry/route.ts` validates the incoming command against:

```ts
command: z.enum(['docs', 'list', 'init', 'telemetry']),
```

`'examples'` is missing, so `EventSchema.safeParse` fails and the route returns **400**. The CLI's sender (`send-telemetry.ts`) swallows the response by design, so the failure is invisible — the command works, but nothing is recorded.

### Required change

Add `'examples'` to the enum in `docs/app/api/telemetry/route.ts` (the file isn't currently part of this PR — it needs to be):

```ts
command: z.enum(['docs', 'list', 'examples', 'init', 'telemetry']),
```

Please also add a test asserting an `examples` event is accepted, so the CLI `CommandName` union and the server schema can't silently drift apart again.

---

sarahgm:
Thanks for the catch — fixed in 89d47d158.

- Added `'examples'` to the `command` enum in `docs/app/api/telemetry/route.ts`, so `examples` events now pass `EventSchema` instead of 400ing.
- Added `docs/app/api/telemetry/route.test.ts` asserting every command in the CLI's `CommandName` union (`docs`, `list`, `examples`, `init`, `telemetry`) is accepted, plus that an unknown command is rejected with 400. The `it.each` list mirrors the union, so the CLI type and the server schema can't silently drift apart again.
- Extended `docs/vitest.config.mjs` `include` to pick up `app/**/*.test.ts` so the route test runs in CI.

`pnpm --filter @marigold/docs test`, `pnpm typecheck:only`, and `pnpm typecheck:docs` all pass.

jim761 (approved):

PR #5529, opened 2026-06-15
feat(DST-1513): Select renderValue gets selection count + works with static children

sebald:
# Description

Improves `<Select>`'s `renderValue` so it can summarise a multi-select trigger, and fixes it for static options. Surfaced during the Filter pattern beta work (DST-1434, quick filters).

Two changes:

1. **`renderValue` now receives a second `details` argument** with the selection `count`, so you can render a compact summary instead of listing every value:
   ```tsx
   <Select selectionMode="multiple" renderValue={(items, { count }) => `${count} selected`}>
   ```
   `count` is correct even for static `<Select.Option>` children.
2. **`renderValue` now runs with static children.** Previously it was silently skipped unless options came from the `items` prop: react-aria derives the render-prop `selectedItems` from each option's `.value` (`null` for static children), and `TriggerValue` filtered those out and fell back to the default. It now gates on `isPlaceholder` (the raw selection) instead of the filtered length.

**Default trigger rendering is unchanged** — without `renderValue`, a multi-select still lists the selected values. The "N selected" summary is opt-in via `renderValue` + `count`.

Closes DST-1513

## Test Instructions

1. `pnpm test:unit --project=unit-tests Select/Select.test` — 17 pass, incl. two new tests for `renderValue` + `count` with static children.
2. Storybook → `Components/Select/MultiSelectSummary`: open, select two options, trigger shows "2 selected".
3. Confirm single-select and the `WithRenderValue` (items-driven) stories are unchanged.

## Breaking Changes

No. The `renderValue` second argument is additive (existing 1-arg callbacks keep working), and the default trigger rendering is unchanged.

## Checklist


packages/components/src/Select/Select.test.tsx
```diff
@@ -174,6 +175,31 @@ test('renderValue receives all selected items in multi-select mode', () => {
   expect(within(button).getByText('Bob Smith')).toBeVisible();
 });
 
+test('renderValue receives the selection count with static children (multiple)', async () => {
```
sarahgm: 💡 Minor redundancy: the `'2 selected'` assertion here is also covered by the `MultiSelectSummary` play function (both run in a real browser). The single-selection test below is genuinely new coverage; this 2-selection unit test is largely duplicative. Fine to keep for explicitness, or trim to just the single-selection case.

sarahgm (commented):
## Automated Code Review

**Recommendation: ✅ Approve** — clean, well-scoped bug fix plus an additive `renderValue` API.

### What it does
`renderValue` gains a second `details: { count }` argument, and the real fix swaps the `items.length === 0` gate (which filtered out the `null` values of static `<Select.Option>` children and silently skipped `renderValue`) for an `isPlaceholder` gate on the raw selection. `count` is taken pre-filter so it's correct even when values are `null`. Applied to both the desktop and mobile (Tray) trigger branches.

### Strengths
- Additive, non-breaking; default trigger rendering unchanged.
- Excellent JSDoc + inline comments explaining the `null`-value gotcha at the point it matters.
- No `any`, proper exported `SelectValueDetails` interface.
- Tests: Vitest, `userEvent`, accessible queries, `component-test` tags, story-imported components. AAA structure respected, with the single-selection unit test adding coverage the play function lacks.
- Correct `minor` changeset.
- CI green; VRT not required (base is `beta-release`).

### Suggestions (optional)
See the two inline comments — minor test redundancy and an accessible-name note on the `QuickFilter` example pattern. Both are nice-to-have; neither blocks merge.

---

packages/components/src/Select/Select.stories.tsx
```diff
@@ -643,6 +643,103 @@ export const WithRenderValue = meta.story({
...
+ */
+export const QuickFilter = meta.story({
+  tags: ['component-test'],
+  render: () => (
+    <Select
+      aria-label="Status"
+      selectionMode="multiple"
+      placeholder="Status"
+      width={48}
+      renderValue={(_items, { count }) => (
```
sarahgm, line 829: 💡 The accessible name of this trigger resolves to roughly "Status 2", which reads ambiguously for screen-reader users. Since this is a pattern others will copy from the Filter beta work, consider demonstrating a clearer SR summary (e.g. visible `Badge` but an accessible "Status, 2 selected"). Usage/example concern only — the component API itself is correct.

packages/components/src/Select/Select.test.tsx
```diff
@@ -174,6 +175,31 @@ test('renderValue receives all selected items in multi-select mode', () => {
...
+  renderWithOverlay(<MultiSelectSummary.Component />);
+
+  const button = screen.getByRole('button');
+  await user.click(button);
```
sarahgm: can also think about if the tests should be interaction tests

sarahgm: same for the other tests with `user.click`

packages/components/src/Select/Select.test.tsx
sebald: Good catch — trimmed the desktop `2 selected` unit test in aa4110e0c. The single-selection unit test plus the small-screen tray test keep the unique coverage, and the `2 selected` assertion still runs in the `MultiSelectSummary` play function (same Firefox env).

packages/components/src/Select/Select.stories.tsx
sebald: Fixed in aa4110e0c. The visual `Status [n]` chrome is now `aria-hidden`, and a `VisuallyHidden` "n selected" summary makes the trigger's accessible name read "Status … 2 selected" instead of a bare, unitless `2`. The `QuickFilter` play now asserts this (`getByRole("button", { name: /2 selected/i })`). Applying the same fix to the filter-pattern docs demo on the DST-1434 branch.

packages/components/src/Select/Select.test.tsx
sebald: Kept these as unit tests deliberately: they force the small-screen `Tray` branch via `mockMatchMedia`, which the browser/interaction (play) tests can only approximate through viewport globals — `MobileControlled` and the DSTSUP-261 guard already follow this unit + matchMedia pattern. The one purely duplicative assertion is now removed (see thread above); the remaining unit tests exercise paths the play functions do not (single-selection count, tray branch). Happy to migrate them to stories if you’d prefer.

sarahgm: Makes sense to keep the matchMedia unit tests for the deterministic small-screen branch control. Just to align on one detail though: the tray itself is reachable from play functions today. `MobileControlled` is a `component-test` play story that sets `globals.viewport: "smallScreen"`, asserts `window.innerWidth < 640`, and exercises the full tray round-trip (open → select → auto-close → controlled `value`) in real Firefox.

One nuance worth noting since it might affect how we reason about these: reduced motion is *not* on in the play/interaction runner. The `prefersReducedMotion: "reduce"` in `preview.tsx` lives under the `chromatic` parameter, so it only applies to VRT snapshots. The Vitest browser runner has no reduced-motion emulation, so play tests actually hit the animated `AnimatePresence` tray path (and the FocusScope close race called out in `TrayModal`), not the synchronous one. `MobileControlled` handles that fine via `waitFor`.

So no objection to the unit-test approach here, just flagging that "play tests can only approximate the tray" is a bit stronger than reality.

sebald: Fully agreed on both counts — thanks for the precision here.

You're right that the tray is genuinely reachable from play tests: `MobileControlled` already drives the full round-trip (open → select → auto-close → controlled `value`) from a `component-test` story via the `smallScreen` viewport global in real Firefox, so "play tests can only approximate the tray" was overstated. The only thing I'm leaning on these unit tests for is deterministic per-test breakpoint control — toggling `(width < 640px)` via `mockMatchMedia` mid-file without a story-level viewport global.

And good catch on reduced motion: since `prefersReducedMotion: "reduce"` is scoped to the `chromatic` parameter, the Vitest browser runner has no reduced-motion emulation and hits the animated `AnimatePresence` tray path (+ the `TrayModal` FocusScope close race), so the `waitFor`s here are load-bearing rather than defensive. Noted — resolving.

packages/components/src/Select/Select.test.tsx
```diff
@@ -186,40 +196,12 @@ test('default trigger render hides description slot', () => {
...
-});
+// The small-screen (`Tray`) branch of `renderValue` is covered by play stories
+// running in real Firefox under the `smallScreen` viewport global:
+//   - `MobileControlled`         — controlled `value` round-trip + DSTSUP-261
+//                                  single-tray-modal invariant
+//   - `MultiSelectSummaryMobile` — `count` summary with static children
+//   - `WithRenderValueMobile`    — custom trigger render, description slot hidden
+// Those were previously `mockMatchMedia`-forced unit tests here; the tray is
+// genuinely reachable from play functions, so they now live alongside their
+// desktop counterparts as interaction tests.
```
sarahgm:
```suggestion
```

sarahgm (approved):

PR #5530, opened 2026-06-15
build(DST-1315): unbundle components build for tree-shaking

sebald:
## Summary

Resolves [DST-1315](https://reservix.atlassian.net/browse/DST-1315). Makes `@marigold/components` tree-shakeable so consumers (e.g. the Portal team's 15 rspack apps, which use ~30 of 74+ components) only pay for what they import.

Previously the package shipped its whole ESM build as one concatenated barrel (`index.mjs`), so bundlers couldn't prove which exports were unused — importing a single leaf pulled in essentially the entire library. The build now emits **one file per source module** (`unbundle: true`) while keeping the `.` barrel import fully backward compatible.

### Measured impact (rspack, deps externalized)

| Consumer imports | Before | After |
| --- | --- | --- |
| `Stack + Text + Card` | 57 KB (≈82% of lib) | **0.84 KB** |
| `Button` | 57 KB | **1.68 KB** |

`size-limit` (brotli, includes deps): leaf 10.9 KB · Button 23 KB · Calendar 60 KB · full barrel 281 KB — all under budget.

---

## ⚠️ Heads-up for reviewers: this is NOT just a build-config change

A reasonable expectation is "to make a library tree-shakeable you only flip a build flag." That's the *start*, but flipping `unbundle: true` alone delivers little benefit. The reason is subtle: the **old single-barrel build concatenated every module into one file**, which masked a class of problems that only surface once each module stands on its own and the bundler starts deciding — per module — what it can safely delete.

So this PR touches component source, not just `tsdown.config.ts`. Here is the exhaustive list of *why* each non-build change was required.

### 1. Toast queue — removed a module-scope side effect (`ToastQueue.ts`, `ToastProvider.tsx`)

`ToastProvider` did this at module top level:

```ts
export const queue = new ToastQueue({ wrapUpdate(fn) { document.startViewTransition(...) } });
```

That constructor **runs the instant the module is imported**, and it touches `document`. Meanwhile the package declares `"sideEffects": false`, which is a promise to bundlers: *"importing any module in here does nothing observable, so if you don't use its exports you may delete it entirely."*

A top-level `new` that touches `document` directly **breaks that promise**:
- It's a genuine side effect on import (and would throw during SSR, where `document` doesn't exist).
- It makes `sideEffects: false` *dishonest* — depending on the bundler that leads to either needed code being dropped or unused code being kept. Either way the contract that powers tree-shaking is unreliable.

**Fix:** construct the queue lazily on first use via `getToastQueue()`. Same singleton, but the module is now genuinely side-effect-free. Stories/tests/`ToastProvider` were updated to call `getToastQueue()`.

### 2. `motion` — stopped using the non-shakeable `motion` proxy (`ActionBar.tsx`, `Tabs/Tab.tsx`, `Tabs/TabList.tsx`, `Tray/TrayModal.tsx`, new `lazyMotion.tsx` + `motionFeatures.ts`)

`import { motion } from 'motion/react'` pulls in motion's **entire** feature set, and the `motion.*` proxy is *by design* not tree-shakeable — referencing `motion.div` drags in the whole renderer (~34 KB).

In the old concatenated barrel this cost was paid once and amortized across the whole library, so it was invisible. Under `unbundle`, the cost attaches to **every module that imports `motion`** (`ActionBar`, `Tabs`, `Tray`). So importing *any one* of those would re-pull motion's full bundle — defeating the entire purpose of the change.

**Fix:** use the lightweight `m` components from `motion/react-m` (tiny core, zero features) and load the `domMax` feature set through a `LazyMotion` boundary via a dynamic `import()`. Bundlers split that into its own async chunk that only downloads when an animated component actually renders.

> Implementation note: the dynamic import targets a **local** module (`motionFeatures.ts` → `export { domMax } from 'motion/react'`) rather than `import('motion/react')` directly. Importing the dependency dynamically made vite's dep-optimizer re-bundle mid-run and drop named exports during browser tests; routing through a local file avoids that.

### 3. `hooks` barrel — `export *` → explicit named re-exports (`index.ts`)

`export * from './hooks'` forces the bundler to pull in and consider that module's *entire* namespace. Explicit named re-exports let it trace precisely which symbols are reachable. `export *` chains are a classic way to silently anchor unused code into every consumer.

### 4. `react-select` — externalized and declared as a dependency (`tsdown.config.ts`, `package.json`)

Under `unbundle`, rolldown copies any **non-externalized** dependency into the output, per importing-module. `react-select` (~2 MB with `@emotion`) was being **bundled into dist without even being a declared dependency** — invisible in the old barrel, but under unbundle it bloated every chunk that referenced it.

**Fix:** mark it `external` and add it to `dependencies` so it resolves transitively at install time. It's used only by the deprecated `Multiselect`; drop both together in the next major.

### 5. Test/mock updates that follow the source changes (`Tray.test.tsx`, `Toast.test.tsx`, `Toast.stories.tsx`)

- `Tray.test.tsx`'s `vi.mock('motion/react')` gained `LazyMotion`/`domMax`, plus a new `vi.mock('motion/react-m')` providing `create`, because `TrayModal` now imports `create` from `motion/react-m`.
- Toast tests/stories switched from the removed module-level `queue` export to `getToastQueue()`.

### 6. `size-limit` CI gate (`.size-limit.json`, `turbo.json`, `.github/workflows/size-limit.yml`)

Added so these wins can't silently regress — e.g. someone re-introducing a `motion` proxy import or a module-scope side effect would trip the budget.

**Bottom line:** `unbundle: true` is the lever, but tree-shaking only actually fires when (a) the package is *honestly* side-effect-free, (b) each module's own imports are themselves shakeable, and (c) heavy deps aren't copied into output. Points 1–4 are exactly those three conditions.

---

### Validation

- ✅ Build (esm + cjs + dts) · `typecheck:only` · lint (0 errors)
- ✅ 1100 unit tests · 11 storybook component tests (real-browser animations)
- ✅ publint clean · attw green for the `.` barrel

No public API changes — every import from `@marigold/components` works exactly as before.

---

## ⚠️ Note: this is barrel-only — no per-component subpath imports (yet)

This PR makes the existing barrel import (`import { Button } from '@marigold/components'`) fully tree-shakeable. It does **not** add per-component subpath entry points like `import { Button } from '@marigold/components/Button'`. Here's exactly why, in plain terms.

**1. Subpath imports would add nothing to bundle size here.**
Tree-shaking and subpath imports solve the *same* problem — "don't ship code the consumer didn't use" — but in two different ways:

- *Subpath imports* are a manual, physical split: each component is its own importable file, so the bundler never even sees the others.
- *Tree-shaking* is automatic: the bundler reads one barrel, figures out which exports are actually reachable, and discards the rest.

We benchmarked both on rspack, vite, and esbuild. With `unbundle: true`, **the tree-shaken barrel produces byte-identical output to importing from subpaths.** Modern bundlers shake the per-module barrel perfectly. So subpaths would be pure ergonomics (a different import style), with zero additional savings — and the whole point of DST-1315 was the savings.

**2. Our build tool can't generate clean subpaths automatically for this output mode.**
`tsdown`'s automatic `exports` generation only emits `.` and `./legacy` when `unbundle` is on (upstream bug [tsdown#752](https://github.com/rolldown/tsdown/issues/752)), and it also drops the `types` conditions in the process. So we can't just flip a flag — getting subpaths would require either:

- hand-maintaining ~73 `exports` entries in `package.json` (one per component, kept in sync forever, easy to break), or
- adding a codegen step that writes those entries on every build.

Both are real, ongoing maintenance cost for **no bundle-size benefit** (see point 1).

**3. Subpaths are a one-way door for the public API.**
Once we publish `@marigold/components/Button`, consumers will import it, and we can never move or rename that file without a breaking change. We'd be locking in the internal folder layout as a public contract. Not worth committing to for an ergonomic-only change.

**In short:** the barrel is now as small as subpaths would make it, subpaths can't be auto-generated for our build mode, and they'd lock in API surface for no payoff. If we later want the subpath *import style* for DX reasons, it can be added on top of this work as a follow-up (via codegen) — it is not a prerequisite for the tree-shaking win this PR delivers.

[DST-1315]: https://reservix.atlassian.net/browse/DST-1315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/Tray/TrayModal.tsx
```diff
@@ -85,37 +87,39 @@ export const TrayModal = ({
...
+    <MotionFeatures>
+      <AnimatePresence>
+        {open && (
+          <MotionModalOverlay
+            // Force the modal to be open when AnimatePresence renders it.
+            isOpen
+            onOpenChange={handleOpenChange}
+            isDismissable={dismissable}
+            isKeyboardDismissDisabled={!keyboardDismissable}
+            className={cn(classNames.overlay)}
```
sarahgm:
```suggestion
            className={classNames.overlay}
```

.github/workflows/size-limit.yml
```diff
@@ -0,0 +1,32 @@
+name: Size Limit
+
+on: ['push']
```
sarahgm: ⚠️ `on: ['push']` is unscoped — it runs on every push to every branch (including `main`) with no `paths` filter, so docs-only or changeset-only pushes still trigger a full `pnpm install` + `pnpm build`. Consider scoping to `pull_request` and/or adding a path filter:

```yaml
on:
  pull_request:
    paths:
      - 'packages/components/**'
      - 'pnpm-lock.yaml'
```

Every other workflow in the repo uses path filters.

packages/components/.size-limit.json
```diff
@@ -0,0 +1,30 @@
...
+    "path": "dist/index.mjs",
+    "import": "{ Stack, Text, Card }",
+    "ignore": ["react", "react-dom", "react/jsx-runtime"],
+    "limit": "16 kB"
+  },
+  {
+    "name": "Button",
+    "path": "dist/index.mjs",
+    "import": "{ Button }",
+    "ignore": ["react", "react-dom", "react/jsx-runtime"],
```
sarahgm, line 13: 💡 The budgets carry ~30–45% headroom vs. the measured sizes in the PR description (Button 23/30 KB, barrel 281/340 KB). That avoids flaky failures, but a real regression — e.g. a re-introduced `motion.*` proxy adding ~10 KB to Button — could land without tripping the gate this PR adds specifically to catch that. Tightening toward ~15–20% headroom would make it bite.

packages/components/package.json
```diff
@@ -74,7 +74,8 @@
     "@react-types/shared": "^3.35.0",
     "@react-types/table": "~3.13.0",
     "motion": "12.39.0",
-    "react-aria-components": "^1.18.0"
+    "react-aria-components": "^1.18.0",
+    "react-select": "^5.10.2"
```
sarahgm, line 78: 💡 `react-select` is now a hard dependency for all consumers even though only the deprecated `Multiselect` uses it. Since it's externalized there's no bundle bloat, but it does land in every consumer's `node_modules` at install. The "drop both together in the next major" plan (in `tsdown.config.ts`) is the right call — just flagging the transitive install cost as a known tradeoff.

packages/components/src/ActionBar/ActionBar.tsx
```diff
@@ -1,4 +1,4 @@
-import { motion } from 'motion/react';
+import * as m from 'motion/react-m';
```
sarahgm, line 1: 💡 VRT recommendation (not a blocker): the `motion` proxy → `m` + lazy `domMax` switch here, in `Tabs`, and in `Tray` changes first-interaction animation timing (LazyMotion loads features async, so the very first render before the chunk resolves may skip its animation). The VRT workflow runs post-merge / on-demand in this repo, so consider commenting `/run-chromatic` before merge to snapshot-check the animated components.

sarahgm (commented):
## Automated Code Review

**PR #5530 — build(DST-1315): unbundle components build for tree-shaking**

### Recommendation: ✅ Approve (with a pre-merge `/run-chromatic` suggested)

Exemplary PR. It solves [DST-1315](https://reservix.atlassian.net/browse/DST-1315) with the ticket's prioritized low-maintenance approach (**Strategy 1 / preserveModules**, validated on rspack), keeps the `.` barrel import backward compatible as required, and every source change is individually justified rather than incidental. CI is fully green (Lint, Typecheck, Size Limit, 1100 unit + 11 SB tests, Packages, CodeQL).

### Correctness checks I verified
- ✅ **Hooks `export *` → explicit re-exports is lossless** — `src/hooks.ts` exports exactly `useAsyncList, useListData, useDragAndDrop, DateFormat, NumericFormat`, identical to the new explicit list. No public export dropped.
- ✅ **`export * from './types'` correctly retained** — types are erased at runtime, no tree-shaking cost.
- ✅ **Toast lazy singleton is correct + SSR-safe** — `getToastQueue()` preserves the shared instance and the new `typeof document !== 'undefined'` guard fixes a latent SSR crash.
- ✅ **`.js` dynamic-import extension is correct**, not a typo — required under `moduleResolution: nodenext`; routing through a local `motionFeatures.ts` is a sound fix for the vite dep-optimizer issue.

### Findings
**Warnings**
- `size-limit.yml` uses unscoped `on: ['push']` — runs on every branch/push with no `paths` filter. Consider scoping to `pull_request` + relevant paths.

**Suggestions**
- `.size-limit.json` budgets carry ~30–45% headroom — a real regression could land without tripping the new gate; tighten toward ~15–20%.
- `react-select` is now a hard dependency for all consumers though only the deprecated `Multiselect` uses it (externalized, so install-cost only — documented, acceptable).
- Lazy `domMax` loading changes first-interaction animation timing for ActionBar/Tabs/Tray — the main reason to run VRT before merge.

### Visual Regression
Not run on the branch, which is **expected** here — `visual-regression-tests.yml` runs on merge-to-main + on-demand `/run-chromatic`, not on open PRs. Because this PR rewrites animation code paths (`motion` proxy → `m` + lazy `domMax`), commenting `/run-chromatic` before merge to snapshot-check the animated components is recommended (not a blocker).

---

packages/components/src/Tray/TrayModal.tsx
sebald: Going to skip this one in this PR. The `className={cn(classNames.overlay)}` line is pre-existing — my only change to `TrayModal.tsx` is the import refactor (`react-aria-components` barrel → `/Dialog` + `/Modal` subpaths); this line isn't part of the diff. It also appears identically in the reduced-motion path just above (L77), so changing only one would be inconsistent.

Worth noting `cn()` isn't strictly a no-op here either — it runs `twMerge`, which resolves conflicting Tailwind utilities *within* a single string, so dropping it isn't guaranteed to be equivalent. Happy to do the cleanup (both call sites) in a separate hygiene PR since it's unrelated to the tree-shaking work here.

packages/components/package.json
sebald: Agreed — this is the intentional tradeoff. Under `unbundle`, a non-externalized `react-select` got copied into the output in full (~2 MB, untree-shaken), so externalizing + declaring it as a dependency is what removes the bundle bloat; the cost is that it lands in consumers' `node_modules` at install. Since it's only used by the deprecated `Multiselect`, the plan is to drop both together in the next major (noted in `tsdown.config.ts`). Flagging it as a known, tracked tradeoff — no change in this PR.

packages/components/src/ActionBar/ActionBar.tsx
aromko, line 1: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files (component source + stories) but the `Visual-Regression-Tests` workflow has not run for this branch. This is especially relevant here because the `motion` → `LazyMotion` refactor changes *when* animation features load. Please trigger it by commenting `/run-chromatic` on this PR.

packages/components/src/lazyMotion.tsx
```diff
@@ -0,0 +1,23 @@
...
+const loadFeatures = () =>
+  import('./motionFeatures.js').then(res => res.domMax);
+
+/**
+ * Wraps a motion subtree with lazily-loaded features. Use together with the
+ * `m` components from `motion/react-m` (never the full `motion` proxy — that
+ * would defeat the lazy split and trip `strict` mode).
+ */
+export const MotionFeatures = ({ children }: { children: ReactNode }) => (
+  <LazyMotion features={loadFeatures} strict>
```
aromko, line 20: ⚠️ With `features={loadFeatures}` loaded via dynamic `import()`, the first render of an animated subtree occurs before features resolve, so the initial entrance/layout/exit animation may be skipped until the chunk loads (motion caches features afterward). The storybook animation tests passing is reassuring — worth a final visual confirmation via VRT. Standard Motion code-splitting trade-off; just flagging for a conscious sign-off.

.github/workflows/size-limit.yml
aromko: 💡 `on: ['push']` runs this on every push to every branch, including docs-only/unrelated work. Consider scoping with `paths: ['packages/components/**']` and/or `pull_request` to cut CI noise. Great to have the budget gate, though.

packages/components/package.json
aromko, line 78: 💡 `react-select` becoming a hard dependency installs ~2 MB transitively for every consumer, though it's externalized so it stays out of bundles unless the deprecated `Multiselect` is imported (correct trade-off, and it fixes the prior undeclared-but-bundled bug). Please make sure the "drop both with `Multiselect` in the next major" intent is tracked in a ticket.

aromko (commented):
Automated code review (review-pr skill). Code is high quality and faithfully implements DST-1315. The only blocker is the missing Visual Regression Tests run — see inline comments.

packages/components/src/Tray/TrayModal.tsx
sarahgm: I would say you can do this also in this PR

sebald: Done in 3dac714a0 — dropped `cn()` at both call sites (the reduced-motion path and the animated path) so they stay consistent.

Confirmed it's safe rather than just plausible: `cn` is `cva`'s `cx` configured with a `twMerge` `onComplete` hook, and `classNames.overlay` has *already* been run through that same `twMerge` when `cva` resolves it. Its base (`bg-black/50 fixed inset-0 z-40 flex items-end justify-center`) has no conflicting utilities, so `cn(classNames.overlay)` was a genuine no-op here. Thanks for the nudge.

.github/workflows/size-limit.yml
sebald: Added `paths-ignore` for `.changeset/`, `.github/`, and `docs/` in 87ba3c198 to cut the noise you flagged.

I kept it on `push` rather than switching to `pull_request` on purpose: the sibling check workflows — `lint.yml`, `test.yml`, and `typecheck.yml` — all use `on: ['push']`, so matching them keeps Size Limit consistent with the rest of CI. (`build.yml` and the VRT workflow are the two exceptions that use `pull_request`.) Happy to revisit if we ever standardize all the checks on `pull_request`.

packages/components/package.json
sebald: The removal is tracked in **DST-1283** ("Remove `<Multiselect>` and the `react-select` dependency from `@marigold/components`") — already staged as a breaking change for the next major. So the transitive install cost is a known, ticketed tradeoff; externalizing keeps `react-select` out of consumer bundles in the meantime (and fixes the prior undeclared-but-bundled bug).

packages/components/src/lazyMotion.tsx
sebald: Agreed — this is the standard Motion code-split tradeoff: the first animated subtree can render before the feature chunk resolves, after which features are cached for the rest of the session. The 11 storybook animation tests (real-browser) pass, and I've just triggered Chromatic via `/run-chromatic` so we get the visual confirmation before merge. I'll sign off here once VRT comes back clean.

sebald: Signing off with empirical verification, not just the VRT pass. @aromko @sarahgm — you were both right to flag this, and Chromatic alone doesn't actually prove it: it captures *static* snapshots, and the storybook tests can take the reduced-motion branch (which in `TrayModal` is a separate early return that never mounts `LazyMotion` at all). So I tested the first render directly.

**Setup:** built the package and ran a throwaway Vite app consuming the **built `dist`** (with the real dynamic `import("./motionFeatures.mjs")` chunk), driven by a real Chromium with `prefers-reduced-motion: false`, sampling the panel's `translateY` every animation frame.

**Worst case tested:** a `Tray` not in the tree until a click, then mounted with `open=true` — so `LazyMotion` mounts in the *same commit* the slide-up entrance fires (this is the `ActionBar`-style mount-at-trigger race, but on a clearly motion-driven, visible animation).

**Result:**

| frame t (ms) | 16 | 26 | 42 | 76 | 126 | 192 | 309 | 492 |
|---|---|---|---|---|---|---|---|---|
| translateY (px) | 805 | 805 | 760 | 614 | 305 | 105 | 26 | 0 |

- Frame 1 is at **805px** = the `initial={{ y: h }}` off-screen state, applied **synchronously** — the panel never flashes at its final position.
- Smooth ramp to 0 over ~492ms (the 0.5s transition), **53 distinct intermediate values** → a genuine tween, not a snap.
- The realistic eager-mount path (`Tray.Trigger`) was byte-for-byte identical.

**Why it holds:** Motion's `m` applies the `initial` prop as a static style on the very first render, *before* features load. So the sequence is always render-off-screen → features resolve (1–2 frames on localhost) → tween in. There is no code path that skips the animation or pops the element to its final state — the worst the lazy load can do is *delay the start* while the element waits, parked, at its initial position.

**Honest caveat:** on a cold load over a slow network the ~30 kB feature chunk could take longer than 1–2 frames; during that window the panel sits parked at its initial position and the (non-animated) backdrop shows immediately, so the slide *starts late* but still plays in full — a deferred, never a missing, animation. Chunk is cached for the rest of the session after first use. If we ever want to eliminate even that deferral for `ActionBar`, hoisting its `<MotionFeatures>` boundary above the `shouldRender` gate (so it mounts before first appearance, the way `Tray` already does) would pre-warm the chunk — a nicety, not a defect fix.

Considering this resolved. 👍

.github/workflows/size-limit.yml
sebald: Scoped in 87ba3c198 by adding `paths-ignore` for `.changeset/`, `.github/`, and `docs/`. I kept the trigger on `push` rather than `pull_request` deliberately: `lint.yml`, `test.yml`, and `typecheck.yml` all use `on: ['push']`, so this stays consistent with the rest of CI (`build.yml` and the VRT workflow are the two `pull_request` exceptions). Happy to move everything to `pull_request` together if we ever standardize. Resolving.

packages/components/.size-limit.json
sebald: Tightened in 87ba3c198 to ~15–20% headroom: leaf **13 kB** (was 16), Button **27 kB** (was 30), Calendar **72 kB** (was 80), full barrel **325 kB** (was 340) — against measured 10.9 / 23 / 60 / 281. A re-introduced `motion.*` proxy (~+10 kB on Button) now trips the gate, which was the point. Resolving.

packages/components/src/ActionBar/ActionBar.tsx
sebald: Ran it — `/run-chromatic` triggered Chromatic **Build 72**, which passed clean (425 stories across 86 components, 416 snapshots, no visual changes). And since static snapshots can't actually catch a *skipped* first-render animation, I verified the `motion`→`LazyMotion` timing empirically in a real browser too — see the sign-off on the `lazyMotion.tsx` thread. Resolving.

sebald: Done — commented `/run-chromatic`; Chromatic **Build 72** passed clean (425 stories across 86 components, 416 snapshots, no visual changes). On the `motion`→`LazyMotion` load-timing concern specifically, I went further and verified the first render in a real browser against the built dist (frame-by-frame `translateY` sampling) — full write-up on the `lazyMotion.tsx` thread; the entrance tweens correctly even in the worst mount-at-trigger case. Resolving.

sarahgm (approved):

aromko (approved):

PR #5537, opened 2026-06-16
refa([DST-1502]): 🧹 Extract shared `resolveInsetAxes` helper for inset padding (fixes `<Card p={number}>`)

OsamaAbdellateef:
## Description

Centralises the `p` → `px`/`py` axis-resolution logic that was copy-pasted across `Page`, `Panel`, `Card`, and `SelectList`. The duplication was the root cause of the `<Card p={number}>` silent bug (resolving to a non-existent `var(--spacing-4-x)`) — the same class of bug that had already been fixed independently in `Panel` (DST-1501) and `Page` (DST-1360).

A single `resolveInsetAxes({ p, px, py, defaultInset })` helper is added to `@marigold/system` and adopted in all four components. `SelectList` carries an inline fix using `isScale` directly because its pattern is structurally different: it has no built-in default, supports partial-axis overrides, and sets CSS vars that cascade to child options rather than the root element.

Closes DST-1502

## Test Instructions

1. Run `pnpm test:unit` — new tests for `resolveInsetAxes` (in `css-variables.utils.test.ts`) and the numeric-`p` regression test for `Card` and `SelectList` should all pass.
2. In Storybook, render `<Card p={4}>` and confirm padding is applied on both axes (previously produced no padding).
3. Confirm `<Panel p={4}>` and `<Page p={4}>` still work as before (covered by existing tests in DST-1501 / DST-1360).

## Breaking Changes

No

## Checklist

### DIFF
--- a/.changeset/dst-1502-inset-padding-helper.md
+++ b/.changeset/dst-1502-inset-padding-helper.md
@@ -1,5 +1,5 @@
 ---
-'@marigold/system': patch
+'@marigold/system': minor
 '@marigold/components': patch
 ---
 
--- a/.changeset/dst-1502-inset-padding-helper.md
+++ b/.changeset/dst-1502-inset-padding-helper.md
@@ -1,0 +1,12 @@
+---
+'@marigold/system': patch
+'@marigold/components': patch
+---
+
+Extract `resolveInsetAxes` helper to centralise inset-padding axis resolution.
+
+The `p` → `px`/`py` resolution logic (branching on whether the value is a numeric scale or a named token) was copy-pasted across `Page`, `Panel`, and `Card`. This duplication caused the `<Card p={number}>` silent bug (resolving to a non-existent `var(--spacing-4-x)`), the same class of bug that had already been fixed independently in `Panel` (DST-1501) and `Page` (DST-1360).
+
+- Adds `resolveInsetAxes({ p, px, py, defaultInset })` to `@marigold/system` alongside `createSpacingVar`.
+- Adopts the helper in `Card`, `Panel`, and `Page`, fixing the live `<Card p={number}>` bug as part of the refactor.
+- Fixes the same numeric-`p` bug in `SelectList` (inline, since its conditional-axis pattern differs).
--- a/packages/components/src/SelectList/SelectList.tsx
+++ b/packages/components/src/SelectList/SelectList.tsx
@@ -21,7 +21,7 @@ import type {
   SpaceProp,
   WidthProp,
 } from '@marigold/system';
-import { cn, createSpacingVar, useClassNames } from '@marigold/system';
+import { cn, createSpacingVar, isScale, useClassNames } from '@marigold/system';
 import { FieldBase } from '../FieldBase/FieldBase';
 import { HiddenSelection } from '../HiddenSelection/HiddenSelection';
 import { SelectListContext } from './Context';
@@ -242,9 +242,15 @@ const SelectList = <Mode extends SelectionMode = 'single'>({
     if (p === undefined && px === undefined && py === undefined) {
       return EMPTY_STYLE;
     }
+    if (p !== undefined) {
+      const inset = `${p}`;
+      const scale = isScale(inset);
+      return {
+        ...createSpacingVar('selectlist-item-px', scale ? inset : `${inset}-x`),
+        ...createSpacingVar('selectlist-item-py', scale ? inset : `${inset}-y`),
+      };
+    }
     return {
-      ...(p !== undefined && createSpacingVar('selectlist-item-px', `${p}-x`)),
-      ...(p !== undefined && createSpacingVar('selectlist-item-py', `${p}-y`)),
       ...(px !== undefined && createSpacingVar('selectlist-item-px', `${px}`)),
       ...(py !== undefined && createSpacingVar('selectlist-item-py', `${py}`)),
     };
--- a/packages/components/src/SelectList/SelectList.tsx
+++ b/packages/components/src/SelectList/SelectList.tsx
@@ -242,6 +242,14 @@ const SelectList = <Mode extends SelectionMode = 'single'>({
     if (p === undefined && px === undefined && py === undefined) {
       return EMPTY_STYLE;
     }
+    // Intentionally diverges from resolveInsetAxes (used by Card/Page/Panel):
+    // 1. `p` short-circuits here — px/py are ignored when p is set. resolveInsetAxes
+    //    does the opposite (explicit px/py win over p via `?? `).
+    // 2. No theme-default fallback — the CSS vars cascade from the theme via the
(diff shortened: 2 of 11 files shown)


packages/components/src/SelectList/SelectList.tsx
```diff
@@ -242,9 +242,15 @@ const SelectList = <Mode extends SelectionMode = 'single'>({
     if (p === undefined && px === undefined && py === undefined) {
       return EMPTY_STYLE;
     }
+    if (p !== undefined) {
```
jim761, line 253: 💡 **Precedence diverges from `resolveInsetAxes`.** Here `p` short-circuits and wins — `px`/`py` are ignored entirely when `p` is set. The helper used by Card/Page/Panel does the opposite (`px ?? …`), so explicit `px`/`py` win over `p` there.

This is currently unreachable through the typed API (the `p` vs `px`/`py` props form a discriminated union, so both can't be set together), so it isn't a bug — just a latent inconsistency. Worth a short comment noting the intentional difference, or aligning the two precedences.

.changeset/dst-1502-inset-padding-helper.md
```diff
@@ -0,0 +1,12 @@
+---
+'@marigold/system': patch
```
jim761: 💡 **Bump level for `@marigold/system`.** This PR adds a new public export (`resolveInsetAxes`, exported from `packages/system/src/index.ts`), which under semver is a `minor` rather than a `patch`. The `@marigold/components` `patch` is correct (it's a bug fix).

If Marigold intentionally treats these system utilities as internal and keeps them at `patch`, feel free to ignore — just confirming it's deliberate.

packages/system/src/utils/css-variables.utils.ts
```diff
@@ -140,6 +140,38 @@ const makeDimensionVar =
...
 
+/**
+ * Resolves `p` / `px` / `py` inset-padding props to per-axis values.
+ *
+ * When `p` is a numeric scale value (e.g. `"4"`) the same value is used on
+ * both axes (`calc(var(--spacing) * 4)`). When it is a named token (e.g.
+ * `"square-regular"`) the conventional `-x` / `-y` suffixes are appended so
+ * the right axis-specific spacing variable is referenced.
+ *
+ * @param p - The shorthand padding prop (`InsetSpacingTokens` or a scale number).
```
jim761: 💡 **JSDoc describes positional params, but the function takes a single object.** The `@param p` / `@param px` / … tags read as if the signature were `resolveInsetAxes(p, px, py, defaultInset)`, but it's `resolveInsetAxes({ p, px, py, defaultInset })`. Harmless, but tidying the tags (or documenting the options object) avoids confusion.

Separately (not specific to this line): consider adding a `<Card p={4}>` story — this bug class has now surfaced three times (Page → Panel → Card), and a story would give it lasting visual coverage.

jim761 (commented):
A few optional, non-blocking suggestions from review (none block merge). The core fix looks correct and behaviour-preserving for Page/Panel.

aromko (commented):
## Automated Code Review

**Recommendation: ✅ Approve**

Clean, well-scoped refactor that fixes a real shipping bug (`<Card p={number}>`) while eliminating the copy-paste duplication that caused it to recur three times (Page, Panel, Card). Behavior is provably preserved for all existing paths, test coverage is solid (helper unit tests + per-component regression tests for `Card` and `SelectList`), the `Inset`/`SelectList` audit from DST-1502 was honored, and CI is fully green.

### Verification
- **Card** previously had no `isScale` branch → `<Card p={4}>` produced `var(--spacing-4-x)` (broken). Now fixed; named-token and `px`/`py` paths are equivalent to before.
- **Page / Panel**: helper reproduces the existing `isScale ? inset : ${inset}-x` logic exactly — pure dedup, no behavior change.
- **SelectList**: helper genuinely can't be reused (no default, partial-axis overrides, child-cascading vars). The inline `isScale` fix is the correct minimal change.
- **Inset**: uses a structurally different API (`space`/`spaceX`/`spaceY` via Tailwind `p-/px-/py-` utilities, no `-x`/`-y` suffix logic) — correctly needs no change.

### Checklist
- TypeScript ✅ explicit return type, no `any`
- Component patterns ✅ no new props, theming untouched
- Testing ✅ Vitest, story-imported components, accessible queries; helper covers named/numeric/string-scale/explicit-override/default cases
- Accessibility ✅ N/A (padding refactor)
- Visual Regression — Not applicable (base: `beta-release`, not `main`)

Two optional polish notes added as inline comments.

---

packages/system/src/utils/css-variables.utils.ts
```diff
@@ -140,6 +140,38 @@ const makeDimensionVar =
...
+ * @param px - Explicit horizontal padding (takes precedence over `p`).
+ * @param py - Explicit vertical padding (takes precedence over `p`).
+ * @param defaultInset - Fallback token used when `p` is `undefined`.
+ */
+export const resolveInsetAxes = ({
```
aromko, line 157: 💡 The helper types `p`/`px`/`py` as loose `string | number`, while call sites pass `InsetSpacingTokens` / `SpaceProp` values. This matches the ticket's suggested signature so it's fine as-is — but a tighter type (or a shared `InsetAxisInput` alias) would catch typos at the call sites. Optional.

packages/components/src/SelectList/SelectList.tsx
aromko, line 253: 💡 Consider a short comment here pointing back to `resolveInsetAxes` and noting *why* SelectList intentionally diverges (no default, partial-axis overrides, child-cascading vars). The rationale is in the PR description but not in the code — it'll help the next maintainer avoid incorrectly "consolidating" this.

packages/system/src/utils/css-variables.utils.ts
OsamaAbdellateef: The call sites are already constrained by SpaceProp<InsetSpacingTokens> / SpaceProp<PaddingSpacingTokens> the helper's string | number is just what those collapse to, so a tighter alias wouldn't add safety here. @aromko

packages/components/src/SelectList/SelectList.tsx
OsamaAbdellateef: Done @aromko

packages/system/src/utils/css-variables.utils.ts
OsamaAbdellateef: done ✔️  @jim761

packages/components/src/SelectList/SelectList.tsx
OsamaAbdellateef: Added ✅  @jim761

jim761 (approved):

aromko (approved):

PR #5539, opened 2026-06-17
feat(DST-1446): add `marigold search` to find components by docs content

sarahgm:
# Description

Adds `marigold search <query>`, which ranks components by what their docs
actually say (title, description, section headings, and section prose) rather
than just by name. This collapses the "list → guess → docs → retry" discovery
loop (3–5 calls) that AI agents run today into a single ranked, snippet-bearing,
deep-linked result.

- **CLI:** new `loadSearchIndex()` / `searchComponentDocs()` library functions
  and a `search` command wrapping them, with `--limit`,
  `--format markdown|json|plain`, `--fresh`, and `--offline` (reusing the
  existing cache and `sanitizeRemote`, no new dependencies). Scoring weights
  title ×3, description ×2, each matching heading ×2, and each matching section
  snippet ×1. Tab completion and telemetry cover the new command. No-match
  exits 0 (`[]` for `--format json`).
- **Docs:** `build-manifest.mjs` now also emits `public/component-search.json`,
  a content index over the component MDX (per-component `headings` plus
  prose-bearing `{ heading, snippet }` sections, with JSX/imports/code-fences
  stripped). It is written after `manifest.json` so a content-index bug can
  never block the manifest that `list`/`docs` depend on.

## Token cost vs. the discovery loop it replaces

Measured against the current docs content (78 components) using the real CLI
output, converted at roughly 4 characters per token. A `search` at the default
`--limit 5` consistently produces ~2.0–2.4 KB of markdown.

| Step (markdown output) | Size | ≈ tokens |
| ---------------------- | ---- | -------- |
| `marigold search "field validation"` | ~2.3 KB | **~570** |
| `marigold list` | 8.5 KB | ~2,130 |
| `marigold docs TextField` (full page) | 17.5 KB | ~4,380 |
| `marigold docs Select` (larger page) | 32.5 KB | ~8,120 |

**Before (agent that does not yet know the component name):**
`list` + one full `docs` is the best case (~6,500 tokens, 2 calls). A wrong
first guess pushes it to `list` + two `docs` (~10,900 tokens, 3 calls).

**After:** a single `search` call (~570 tokens, 1 call) returns the ranked
candidates with snippets and deep links. The agent then optionally fetches one
targeted `docs <Component> --section props` only after it has committed to a
component.

That is roughly an **11–19× reduction** in discovery-phase tokens, in line with
the 5–20× estimate in DST-1446.

### On the `--section` flag (descoped)

The `--section` flag from the original scope is intentionally not implemented.
`search` is a discovery command, so section-level refinement is already served
by `marigold docs <Component> --section props`. A `--section` filter here would
duplicate that surface and, because ranking reads every section while only the
single best hit is emitted, would mostly suppress otherwise-relevant components
for marginal value.

Closes DST-1446

## Test Instructions

1. `pnpm --filter @marigold/cli build`
2. `marigold search "field validation"` — expect ranked results with snippet + deep link
3. `marigold search "field validation" --format json` — expect `[{ name, slug, score, hits }]`
4. `marigold search zzzznope` — expect a no-match message (exit 0); `--format json` returns `[]`
5. Automated: `pnpm --filter @marigold/cli test` and `pnpm --filter @marigold/docs test`

### Reliable test queries

Each query exercises one behavior of the ranker. Expected outcomes list the stable parts (top hit / presence-absence) rather than full orderings.

| Query | What it demonstrates | Expected |
| ----- | -------------------- | -------- |
| `marigold search "form"` | Whole-word precision (no substring false positives) | `Form` / `parseFormData` near the top; **`DateFormat` and `NumericFormat` do _not_ appear** (substring `form`∈`Format` is gone) |
| `marigold search "field validation"` | camelCase-split titles | The `*Field` components surface (`TagField`, `DateField`, `TextField`) — `field` matches the title even though it's one token |
| `marigold search "date picker"` | Exact-phrase bonus | `DatePicker` ranks **#1** |
| `marigold search "buttons"` | Plural folding | Matches `Button` |
| `marigold search "modal dialog"` | Multi-word relevance | `Dialog` ranks **#1** |
| `marigold search "navigation"` | Exact-intent over rarity | `TopNavigation` ranks **#1** |
| `marigold search zzzznope` | No-match contract | Friendly message, exit `0`; `--format json` returns `[]` |

Add `--format json` to inspect the `{ name, slug, score, hits }` contract, or `--limit N` to widen the result set. `marigold search` ranks client-side, so a CLI built from this branch produces the tuned order against whichever index `component-search.json` is served — point `MARIGOLD_DOCS_URL` at this PR's Vercel preview to test before merge.

## Breaking Changes

No

## Checklist

- [ ] Visual regression tests updated (for UI changes) — n/a, no UI files changed

---


jim761 (changes requested):
## Automated Code Review

**feat(DST-1446): add `marigold search` to find components by docs content**

Strong PR overall — word-aware ranking, thorough tests, security-conscious index sanitization, and a clean `invokedDirectly` guard so the search index can't regress the manifest. CI is green and no UI files change (VRT not needed). One must-fix found below: the new command's telemetry is dropped server-side.

### 🔴 Critical (Must Fix)

**`search` telemetry never reaches Upstash — server schema rejects it**

The CLI now emits `command: 'search'` (`packages/cli/src/lib/telemetry.ts:14`, wired via `commands-spec.ts` → `marigold.ts:399`) and POSTs it to `${docsUrl()}/api/telemetry`. But the route handler — `docs/app/api/telemetry/route.ts:10`, **not touched by this PR** — still validates with:

```ts
command: z.enum(['docs', 'list', 'examples', 'init', 'telemetry']),   // no 'search'
```

So every `search` event fails `EventSchema.safeParse()` and the route returns **400 `invalid event`** (route.ts:66-68), never reaching the `client.lpush()` to Upstash at line 90. Verified against the route's own enum: `search` is the only command rejected. All other commands still upload fine — only the new one is silently dropped.

CI stayed green because the route's drift-guard test — whose stated purpose is to catch exactly this CLI-union ↔ server-schema drift (`route.test.ts:38-40`) — also wasn't extended to include `search`.

**Fix (two one-line changes, both outside the current diff):**

```ts
// docs/app/api/telemetry/route.ts:10
command: z.enum(['docs', 'list', 'search', 'examples', 'init', 'telemetry']),

// docs/app/api/telemetry/route.test.ts:40
it.each(['docs', 'list', 'search', 'examples', 'init', 'telemetry'])(
```

Follow-up worth considering: the CLI `CommandName` union and the server zod enum are a hand-synced contract across two packages with no shared source of truth — a shared constant would make this drift impossible.

### 💡 Suggestions (Nice to Have)

1. **`docs/scripts/build-manifest.mjs:149`** — `cleanProse`'s `.replace(/-{2,}/g, ' ')` (intended for markdown table/rule scaffolding) also collapses double-hyphens in real prose, so a documented flag like `--offline` shows up as `offline` in the snippet evidence. Cosmetic only — ranking is unaffected (the tokenizer splits on `[a-z0-9]+`, so `--offline` and `offline` yield the same token). Best-effort extraction is already documented.

2. **`packages/cli/src/lib/format.ts:1008-1013`** — a title-/heading-only match produces `hits: []` (JSON) and the markdown view falls back to `r.description`. Sensible behavior, but it's an unstated edge of the documented `hits` contract and that fallback path isn't directly tested (every fixture matches a section). Consider one test to lock it.

### Recommendation
**Request Changes** — the Critical telemetry fix should land before merge; the suggestions are optional.

---

aromko (commented):
## Automated Code Review

**Recommendation: Approve** ✅

This is a high-quality, well-tested PR. The implementation is faithful to [DST-1446](https://reservix.atlassian.net/browse/DST-1446), the code is unusually well-commented (the *why* behind regex-based extraction, scoring weights, phrase/coverage tuning, and manifest-before-index write ordering are all explained), security sanitization is thorough, and error handling degrades gracefully. CI is green across the board.

### Highlights
- `FetchError` carrying `.status` lets callers branch (404 → friendly "index not found" guidance) instead of string-matching error messages.
- `cleanIndex` runs `sanitizeRemote` on every string from the fetched index (defense-in-depth), verified by the escape-sequence test.
- Whole-word + camelCase tokenization fixes the classic `form` ∈ `Format` substring false positive.
- `invokedDirectly` guard + manifest-written-before-search-index ordering means a content-index bug can never block the `manifest.json` that `list`/`docs` depend on.
- Excellent test matrix: scoring, coverage scaling, phrase bonus, tie-break, plural folding, camelCase split, empty/partial index, cold/warm cache, 404, sanitization, no-match contract.

1. **Descoped `--section` flag vs. ticket scope.** DST-1446 lists `--section` (and a `--section` filter test) in scope. The PR's rationale for dropping it (discovery vs. refinement; `docs --section` already serves that surface) is sound and well-argued — recommend updating the Jira ticket's scope/AC so the ticket and the shipped command agree before close.

2. **`packages/cli/src/lib/search.ts` — phrase bonus uses the unstemmed query.** `phrase` derives from the raw normalized query while term matching is stemmed, so `search "date pickers"` matches both terms but won't earn `PHRASE_BONUS` against text reading "date picker". Reasonable simplicity tradeoff; flagging the asymmetry only.

3. **Telemetry records the raw `--limit` value** (e.g. `"0"`). Consistent with how `--format`'s enum value is kept, and a numeric limit isn't sensitive — just confirming it's intentional (the test asserts it, so it is).

### Notes
- **Visual Regression Tests:** Not needed — no UI files changed (CLI/docs-script only).
- **React performance:** Not applicable — no React components in this PR.

No critical or blocking issues found.

---

.changeset/dst-1446-cli-search.md
```diff
@@ -0,0 +1,11 @@
+---
+'@marigold/cli': minor
+'@marigold/docs': minor
+---
+
+feat(DST-1446): `marigold search` to find components by docs content
+
+Adds `marigold search <query>`, which ranks components by what their docs actually say — title, description, section headings, and section prose — rather than just by name. This collapses the "list → guess → docs → retry" discovery loop (3–5 calls) that AI agents run today into a single ranked, snippet-bearing, deep-linked result.
```
aromko: What this mean? "Adds `marigold search <query>`, which ranks components by what their docs actually say..."

packages/cli/src/bin/marigold.ts
```diff
@@ -138,6 +146,18 @@ const parseListCommand = (argv: string[]) =>
     },
   });
 
+const parseSearchCommand = (argv: string[]) =>
```
aromko, line 149: what is argv? Can we just call it args?

sarahgm: Good question. `argv` is the consistent name across all five `parse*Command` helpers and the main dispatcher in this file (it's the conventional Node name for the raw argument vector passed to `parseArgs`, distinct from the parsed `args`/`values` result). I kept it as-is here so the helper signatures stay uniform, since renaming just this one would make it the odd one out. Happy to do a file-wide `argv` to `args` rename in a follow-up if you'd prefer that convention everywhere.

.changeset/dst-1446-cli-search.md
sarahgm: Fair point, that phrasing was vague. Reworded in 268ee98: "ranks components by matching the query against their docs content (title, description, section headings, and section prose), not just the component name." The intent is that `search` scores against the actual documentation text, not only the component name like `list --search` does.

sarahgm:
## Review feedback addressed (`268ee98`)

Thanks for the thorough reviews. Summary of how each point was handled:

### 🔴 Must-fix
- **`search` telemetry rejected server-side** (@jim761) — added `search` to the telemetry route's command enum (`docs/app/api/telemetry/route.ts`) and to the drift-guard test's `it.each` list (`route.test.ts`), so `marigold search` events pass `EventSchema` validation instead of returning 400. Verified the route test now exercises all six commands.

### Suggestions adopted
- **Title-only-match fallback untested** (@jim761) — added a `formatSearchResults` test block covering the empty-`hits` case: the JSON contract emits `hits: []`, and the rendered view falls back to the description.
- **`cleanProse` collapsing prose flags** (@jim761) — scoped the hyphen-run strip to whitespace-bounded runs, so markdown rules (`---`) and table separators (`| --- |`) are still removed while real flags like `--offline`/`--limit` keep their dashes. Locked with a regression test.
- **Vague changeset wording** (@aromko) — reworded the lead line to state that `search` ranks by matching the query against docs content (title, description, headings, prose), not just the component name.

### Inline comments
- Both inline threads (`argv` naming, changeset phrasing) have been replied to and resolved.

### Acknowledged, no change
- **`--section` descope vs. DST-1446 scope** (@aromko) — agreed the ticket and shipped command should agree; will update the Jira scope/AC rather than the code.
- **Phrase bonus uses unstemmed query** / **raw `--limit` in telemetry** (@aromko) — intentional simplicity tradeoffs as noted; leaving as-is.

jim761 (approved):
## Automated Code Review — Re-review ✅

All three findings from my previous review have been addressed in `268ee9839` and CI is green on `eae1852`. Clearing the requested changes.

| Previous finding | Status |
|---|---|
| 🔴 **Critical** — `search` telemetry rejected server-side (`EventSchema` enum missing `search` → 400 before the Upstash `lpush`) | ✅ Fixed — `route.ts` enum + `route.test.ts` drift-guard list both include `search` |
| 💡 `cleanProse` collapsed `--flag` dashes in snippets | ✅ Fixed — `/-{2,}/` → `/(^\|\s)-{2,}(?=\s\|$)/` (whitespace-bounded only) + regression test |
| 💡 `hits: []` / description-fallback path untested | ✅ Fixed — new `formatSearchResults` tests cover JSON empty-hits and the plain/markdown description fallback |

Verified the new `cleanProse` regex locally: markdown rules/table separators still collapse, while real prose flags like `--offline` / `--fresh` keep their leading dashes. The telemetry fix resolves the issue end-to-end — the server now accepts and stores `search` events.

Nice work on the quick turnaround. 🎉

---

aromko (approved):

PR #5550, opened 2026-06-23
feat(DST-765): add SegmentedControl component

sebald:
## Summary

Adds a new **`<SegmentedControl>`** component for compact, single-select view switching and quick filters, plus its full documentation page.

It is built on react-aria's `RadioGroup` / `RadioField` / `RadioButton` with a sliding `SelectionIndicator`, so it is a real form field: `value` / `defaultValue` / `onChange`, the `name` attribute (submits like a radio group), `required`, `error` + `errorMessage`, `description`, `readOnly`, and validation all route through `<FieldBase>` like the other form components.

```jsx
<SegmentedControl label="View" defaultValue="list">
  <SegmentedControl.Option value="list">List</SegmentedControl.Option>
  <SegmentedControl.Option value="grid">Grid</SegmentedControl.Option>
</SegmentedControl>
```

## Component

- **Variants:** `default` (a `bg-control` track with a raised `ui-surface` thumb, mirroring `Switch`) and `ghost` (track-less, translucent indicator for dense toolbars).
- **Single `default` size** (matches the `h-control` Input height). The `small` size was dropped for now; the `size` prop stays as `'default' | (string & {})` for theme extensibility.
- **Full width** via the standard `width` prop (`width="full"`) — no separate `fullWidth` prop.
- **Overflow:** when segments exceed the available width the control scrolls horizontally (selected option kept in view, reduced-motion aware) with a scroll-driven edge fade.
- **Keyboard:** single Tab stop, arrow keys move and select (selection follows focus), focus rings reuse shared `ui-*` utilities and are not clipped at the track edges.
- `ToggleButtonGroup` now logs a dev-only warning when used with `selectionMode`, steering single-select use cases toward `SegmentedControl`.

## Docs

- Use-case and intent-driven page: Anatomy (with an SVG anatomy diagram in the house style), Appearance, Usage, Accessibility, Props, Alternative components.
- A dedicated **Segmented control vs tabs** section with a plain-language litmus test (could two options ever show the same item?), grounded in Apple HIG / Material / Carbon / Polaris / NN/g guidance.
- Follows the house pattern: short intro, standard `## Alternative components` list, full `## Accessibility` section.

## Tests

- 14 unit tests (AAA pattern) and 5 Storybook interaction tests, all passing.

## Notes for reviewers

- Targets `beta-release` (repo is in changesets pre/beta mode).
- Changeset: `add-segmented-control-component.md` (`@marigold/components`, `@marigold/theme-rui`, `@marigold/system` — minor).

[DST-765](https://reservix.atlassian.net/browse/DST-765)

[DST-765]: https://reservix.atlassian.net/browse/DST-765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/SegmentedControl/SegmentedControl.stories.tsx
```diff
@@ -0,0 +1,222 @@
+import { useState } from 'react';
+import { Form } from 'react-aria-components/Form';
+import { expect, userEvent, waitFor } from 'storybook/test';
+import preview from '../../../../.storybook/preview';
```
jim761: ../../../../ can be removed

jim761 (changes requested):

jim761 (commented):
## Automated Code Review

Reviewed the component against Marigold's house patterns — it mirrors the `Radio/RadioGroup` wrapper faithfully (`FieldBase as={RadioGroup}`, `error → isInvalid`, prop renames, context for variant/size), tests are thorough, and the dev-warning on `ToggleButtonGroup` follows the existing `Card`/`SelectListOption` precedent. A few non-blocking items below.

### 🟡 Warnings (Should Fix)

- **`SegmentedControlOptionProps` publicly exposes `variant?: string` and `size?: string`** (`packages/components/src/SegmentedControl/SegmentedControl.tsx:1078-1079`). These are internal — fed via `SegmentedControlContext` — and `variant ?? context.variant` means a consumer *can* override a single segment's variant, which is undocumented and would break visual consistency. Recommend dropping them from the public interface (keep them context-only), unless per-option override is a deliberate feature.

### 💡 Suggestions (Nice to Have)

- **`expandWidth = !!width`** (`SegmentedControl.tsx:979`) makes *any* truthy `width` — including a fixed value or a fraction like `"1/2"` — apply `grow basis-0` and divide segments equally. The docs only describe `width="full"`. Worth confirming the intent for fixed/fraction widths, and either documenting it or narrowing the condition to `width === 'full'`.

- **One unit test builds an inline `MarigoldProvider` fixture** — `disables only the option marked as disabled` (`SegmentedControl.test.tsx:813-830`) — instead of importing a story, which the other tests (and the "import stories, don't create fixtures" rule) follow. Consider adding a `DisabledOption` story and rendering `<DisabledOption.Component />`.

- **Compound export uses `Object.assign`** rather than the typed compound-component interface pattern. Works fine; just slightly weaker typing on `SegmentedControl.Option`.

---

packages/components/src/SegmentedControl/SegmentedControl.stories.tsx
sebald: Fixed — the import is now `import preview from '.storybook/preview';`, matching every other `*.stories.tsx` in the repo (the `.storybook` path alias resolves it). Resolving.

sebald:
Thanks for the review — addressed the four items:

**🟡 `variant?`/`size?` exposed on `SegmentedControlOptionProps`** — Fixed. Dropped both from the public interface; they're now context-only (read from `SegmentedControlContext`). `SegmentedControlOption` no longer accepts a per-segment `variant`/`size`, so a single option can never diverge from the control. Removed the `variant ?? context.variant` fallback.

**💡 `expandWidth = !!width`** — Fixed. Narrowed to `width === 'full'`. Only `width="full"` now stretches the segments to divide the available width equally; any other width value (fixed token, fraction) just constrains the field box like every other form component, leaving segments at their natural size. This matches what the docs describe. (Side note: there is no `"100%"` token — `full` → `w-full` is the keyword.)

**💡 inline `MarigoldProvider` fixture in `disables only the option marked as disabled`** — Fixed. Added a `DisabledOption` story (tagged `component-test`) and the test now renders `<DisabledOption.Component onChange={…} />`, matching the import-stories rule the other tests follow.

**💡 `Object.assign` compound export** — Kept as-is, intentionally. `Object.assign` infers a precise intersection type (`typeof SegmentedControlBase & { Option: typeof SegmentedControlOption }`), so `SegmentedControl.Option` is already fully typed with no `as` cast. The `interface … as XComponent` pattern would *introduce* a cast, so switching wouldn't strengthen typing here.

jim761 (approved):

PR #5552, opened 2026-06-23
fix(DST-1555): omit style on Label and fix dead Panel space control option

sarahgm:
# Description

Two small API-consistency leftovers found in the `beta-release` audit:

- `Label` only removed `className` from its react-aria props, leaving `style` in the public interface. It now omits both, matching the convention used by `Description` and `TextValue`. Theme the label via `variant`/`size` instead.
- `Panel`'s `space` Storybook control still offered a `'none'` option. The universal zero-spacing token was renamed `none` → `collapsed`, so selecting `'none'` resolved to nothing. The control now offers `'collapsed'` (the other five options were already valid).

No rendered change in either case — the `style` prop was never meant to be public, and the Panel fix only corrects a Storybook control option.

Closes DST-1555

## Test Instructions

1. `pnpm typecheck:only` — passes (verified locally).
2. `<Label style={{}}>` no longer typechecks (intended); `variant`/`size` still work.
3. Storybook → Components/Panel: the `space` control now lists `collapsed` instead of `none`, and selecting it collapses spacing.

## Breaking Changes

No — removing the never-functional `style` prop from `Label` is a type-only narrowing; ships as a `patch`. The Panel change is Storybook-only.

## Checklist

- [ ] Visual regression tests updated (for UI changes) — N/A, PR targets `beta-release` and has no rendered change

### DIFF
--- a/.changeset/label-omit-style-prop.md
+++ b/.changeset/label-omit-style-prop.md
@@ -1,0 +1,10 @@
+---
+'@marigold/components': patch
+---
+
+fix: stop exposing `style` on `Label`
+
+`Label` only removed `className` from its react-aria props, leaving `style` in
+the public interface. It now omits both, matching the `className`/`style`
+removal convention used across the design system (e.g. `Description`,
+`TextValue`). Theme the label via `variant`/`size` instead.
--- a/packages/components/src/Label/Label.tsx
+++ b/packages/components/src/Label/Label.tsx
@@ -2,7 +2,7 @@ import type RAC from 'react-aria-components';
 import { Label } from 'react-aria-components/Label';
 import { cn, useClassNames } from '@marigold/system';
 
-type RemovedProps = 'className';
+type RemovedProps = 'className' | 'style';
 export interface LabelProps extends Omit<RAC.LabelProps, RemovedProps> {
   size?: string;
   variant?: string;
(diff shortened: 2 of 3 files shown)


jim761 (approved):
## Automated Code Review — ✅ Approve

Tightly scoped, correct, and consistent with existing conventions. Both claims in the description check out against the source.

### Verified
- **`Label`** now omits `'className' | 'style'`, matching `Description` and `TextValue` (which both already omit `style`). The component body never read `style`, so this is a pure type-only narrowing — no runtime change. Correctly shipped as a `patch`.
- **`Panel` `space` control** options `['collapsed', 'tight', 'related', 'regular', 'group', 'section']` now exactly match the six `SpacingTokens` in `themes/theme-rui/src/tokens.css`. The dead `'none'` option was a leftover from the `none → collapsed` token rename (confirmed in the theme-rui CHANGELOG — Tailwind v4 `leading-none` collision).
- CI is green (Typecheck, Lint, Unit, Storybook, Coverage, Size Limit) and a changeset is included. No rendered change; VRT correctly skipped.

### Follow-up (not blocking)
The dead option existed because the story's `options` array is a hand-written duplicate of the `SpacingTokens` type, with nothing tying them together at compile time. Worth a future `🧹` cleanup to make it drift-proof — e.g. `as const satisfies readonly SpacingTokens[]` on the array (fails typecheck on a removed/renamed token), or deriving both the type and the control from a single `as const` token array. Applies to any Storybook control that manually enumerates a design-system union.

---

OsamaAbdellateef (approved):

PR #5556, opened 2026-06-23
feat(DST-1551): add `<DateRangePicker>` component

sebald:
## Summary

Adds the `<DateRangePicker>` component, completing the Marigold date-input family (Calendar, RangeCalendar, DateField, DatePicker, TimeField). Users can enter or select a start–end date range through a single field, mirroring `<DatePicker>`'s API and behaviour.

Closes DST-1551.

## What's included

- **Component** — `<DateRangePicker>` wraps RAC `DateRangePicker`: two date inputs (`start`/`end`) in one field group with a calendar button that opens a `<RangeCalendar>` in a popover on desktop and a tray on small screens.
- **Capabilities** — per-input paste (ISO/EU/US formats), `granularity` (inline time segments), `visibleDuration` (up to three months), and the usual Marigold field props (`disabled`, `readOnly`, `required`, `error`, `errorMessage`, `description`, `minValue`, `maxValue`, `dateUnavailable`, `width`, `variant`, `size`).
- **Theme** — matching `DateRangePicker` entry in `theme-rui`.
- **Tests** — unit tests + story `play` functions tagged `component-test` (open popover, select range, paste, keyboard nav, error/disabled/readonly states).
- **Docs** — intent- and use-case-driven usage page with a themed anatomy SVG and demos: comparison vs. two fields, limiting the range, range length, blocking unavailable dates, quick ranges, multiple months, including a time, and accessibility.

## Also in this branch

- **RangeCalendar hover fix** — days outside the selected range now round their hover/focus highlight to match the selected state; in-range cells stay square so the fill connects seamlessly.

## Changesets

- `@marigold/components` minor, `@marigold/theme-rui` minor — new component
- `@marigold/theme-rui` patch — RangeCalendar hover/focus radius fix

## Verification

- `pnpm --filter @marigold/docs types:check` ✓
- Docs registry rebuilt (383 items) ✓
- `theme-rui` build ✓


sebald:
Note on the **Quick ranges** section (the preset / quick-select demo): I'm not fully sure it belongs here yet. [DST-1511](https://reservix.atlassian.net/browse/DST-1511) ("Add relative date-range presets (quick ranges) to the date picker") is planned to implement quick ranges as a first-class feature anyway, so we may want to trim this section and document it properly once that lands. Open to feedback on whether to keep it as a usage recipe for now or drop it.

[DST-1511]: https://reservix.atlassian.net/browse/DST-1511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

sarahgm (commented):
## Automated Code Review

**PR #5556 — feat(DST-1551): add `<DateRangePicker>` component**

Reviewed against Marigold standards, the `DatePicker` pattern, and the DST-1551 scope. CI is fully green and VRT is not enforced for a `beta-release` target.

### Summary
The implementation faithfully delivers the ticket scope: RAC `DateRangePicker` wrapped via `<FieldBase>`, two `start`/`end` `DateInput`s with an `aria-hidden` separator, `RangeCalendar` in a `Popover`/`Dialog` on desktop and a `Tray` on mobile, per-input paste, `granularity` pass-through, multi-month support, theme entry, exports, tests, and docs. The two documented out-of-scope items (in-popover TimeField, full range-string paste) are correctly omitted.

### Checklist
- ✅ TypeScript — no new `any` in source; props mirror RAC with proper `Omit`/`Pick` (`: any` in stories matches the existing `DatePicker.stories.tsx` convention)
- ✅ Component patterns — React 19 `ref` prop, RAC prop renames, no `className`/`style` exposed, `useClassNames` for theming, `IconButton` used only internally
- ✅ Reuse — `parseDateFromString` extracted/exported from `DateField/DateInput.tsx` and reused verbatim
- ✅ Testing — `userEvent`, accessible queries, `component-test`-tagged play functions, unit tests import stories
- ✅ Accessibility — separator `aria-hidden`; segments/inputs surface as `spinbutton`/`group`; overlay a11y inherited from RAC
- ✅ Theme — `DateRangePicker.styles.ts` re-exports `DatePicker` styles with a clear divergence comment; `Theme` type + appearance build script updated

### Issues

**Critical:** None.

**Warnings:** None.

**Suggestions (non-blocking):**
1. `DateRangePicker.tsx` — paste uses `ctx.setDateTime(part, …)` rather than the ticket's suggested `setDate`. This is a *correct* deviation and the inline comment explains why. Worth confirming behaviour when `granularity` includes a time, since `parseDateFromString` yields a date-only `CalendarDate`.
2. The paste `try/catch` + `console.warn` duplicates `DateField/DateInput.tsx`. Consistent with existing code; could share a small helper later. Low priority.
3. `variant?: string; size?: string;` uses bare `string` rather than the `'default' | (string & {})` pattern in CLAUDE.md — but this matches `DatePicker.tsx` exactly, so it's an existing convention, not a regression.
4. 👍 `visibleDuration`/`pageBehavior` are forwarded only to the desktop `RangeCalendar`, not the `Tray` one — the correct resolution of the ticket's "single-month in the tray" open question.

### React Performance
No concerns — thin RAC wrapper with no data fetching, lists, or render-time work warranting `memo`/`useMemo`.

### Recommendation
**Approve / Comment.** High-quality, well-scoped addition that mirrors `DatePicker` and fully satisfies DST-1551. Findings are all minor suggestions, none blocking.

---

sarahgm:
> Note on the **Quick ranges** section (the preset / quick-select demo): I'm not fully sure it belongs here yet. [DST-1511](https://reservix.atlassian.net/browse/DST-1511) ("Add relative date-range presets (quick ranges) to the date picker") is planned to implement quick ranges as a first-class feature anyway, so we may want to trim this section and document it properly once that lands. Open to feedback on whether to keep it as a usage recipe for now or drop it.

I would leave it out since we know how we want to provide the quick ranges? At least for me it feels you can have many ways (in a Select, like Buttons as you did, with Calendar or not) So should we align first? @sebald 

[DST-1511]: https://reservix.atlassian.net/browse/DST-1511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

sarahgm:
Why is this targeting `beta-release` ? I think this could also be shipped within main or not? @sebald

packages/components/src/DateRangePicker/DateRangePicker.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+    test('renders the calendar when the picker is open', async () => {
+      render(<Basic.Component open />);
+
+      await waitFor(() => {
+        expect(screen.getByRole('application')).toBeVisible();
+      });
+      expect(screen.getByRole('grid')).toBeVisible();
+    });
+
+    test('opens the calendar when the trigger is pressed', async () => {
```
sarahgm: This would make sense as a component interaction test in storybook or not?

packages/components/src/DateRangePicker/DateRangePicker.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+    test('marks unavailable dates via dateUnavailable', async () => {
+      render(<UnavailableDate.Component open />);
+
+      await waitFor(() => {
+        expect(screen.getByRole('application')).toBeVisible();
+      });
+    });
+  });
+
+  describe('paste handling', () => {
```
sarahgm: These also should be component tests in stories

packages/components/src/DateRangePicker/DateRangePicker.test.tsx
sebald: Moved this to a story component-test — `Basic.test("opens the calendar popover and selects a range")` in `DateRangePicker.stories.tsx`, which runs in a real browser. Removed the duplicate jsdom test here. (commit ccacb90a0)

packages/components/src/DateRangePicker/DateRangePicker.test.tsx
sebald: Done — paste is now covered by the story component-test `Basic.test("pastes dates into the start and end inputs")`, extended to cover ISO, European and invalid-ignored input so coverage is preserved. Removed the duplicate jsdom paste tests here. (commit ccacb90a0)

sebald:
@sarahgm thanks for the review! Pushed changes addressing everything. Summary:

**Quick ranges** — agreed, dropped the section and the presets demo. We can document it properly once [DST-1511](https://reservix.atlassian.net/browse/DST-1511) lands and we've aligned on how quick ranges should work.

**Why `beta-release` and not `main`?** — Technically it *could* ship via `main` (all runtime deps already exist there), but it's deliberately on `beta-release` so it rides the same release train as `RangeCalendar` and the rest of the in-flight beta batch, rather than cutting a lone stable minor while that batch is still baking. We'll backport to `main` separately once this PR is closed — tracked in **[DST-1561](https://reservix.atlassian.net/browse/DST-1561)**.

**From the preview feedback:**
- **320px / narrow screens** — the field now fills the available width on small screens and holds a `min-w-fit` floor, so the inputs no longer crush together. It still needs room for its content (~260px), so I also added a doc note that very narrow viewports like 320px aren't a good fit.
- **"how do I know which time is picked"** — time/granularity support is dropped for now, so the section and demo are gone.
- **"already in the do and don't"** — removed the redundant "use two separate fields" paragraph.
- **"badge missing"** — added the `beta` badge to the sidebar entry.

**Tests** — the trigger-open and paste flows moved into story component-tests (replies on those threads above).

[DST-1511]: https://reservix.atlassian.net/browse/DST-1511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DST-1561]: https://reservix.atlassian.net/browse/DST-1561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

sebald:
> Why is this targeting `beta-release`? I think this could also be shipped within main or not?

@sarahgm I dug into this. It *could* build on `main` (all runtime deps already exist there), but it's intentionally on `beta-release` so it ships with the same release train as `RangeCalendar` and the rest of the in-flight beta batch instead of cutting a lone stable minor while that batch is still baking.

I created a follow-up to backport it to `main` once this PR is closed: **[DST-1561](https://reservix.atlassian.net/browse/DST-1561)** (blocked by DST-1551). It's a clean cherry-pick of the DateRangePicker commits onto a `main`-based branch.

[DST-1561]: https://reservix.atlassian.net/browse/DST-1561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

packages/components/src/DateRangePicker/DateRangePicker.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+      render(<WithError.Component />);
+
+      expect(screen.getByText('Whoopsie')).toBeVisible();
+    });
+
+    test('marks unavailable dates via dateUnavailable', async () => {
+      render(<UnavailableDate.Component open />);
+
+      await waitFor(() => {
+        expect(screen.getByRole('application')).toBeVisible();
```
sarahgm, line 70: ⚠️ This test is named `marks unavailable dates via dateUnavailable` but only asserts the calendar popover is visible — the same thing the `renders the calendar when the picker is open` test already checks. It never verifies a date is actually unavailable, so it would still pass if `dateUnavailable` were a no-op. Consider asserting a Sunday cell is disabled/non-selectable, e.g. query the cell and check `aria-disabled`.

packages/components/src/DateRangePicker/DateRangePicker.stories.tsx
```diff
@@ -0,0 +1,323 @@
...
+      const grid = await canvas.findByRole('grid');
+      await waitFor(() => expect(grid).toBeVisible());
+    });
+  }
+);
+
+Basic.test(
+  'pastes dates into the start and end inputs',
+  async ({ canvas, step }: any) => {
+    const firePaste = (element: Element, text: string) => {
```
sarahgm: 💡 Manually dispatching a synthetic `paste` event is reasonable here (clipboard `getData` can't be driven through `userEvent` in-browser). Since this is the only paste test in the date-input family, consider extracting `firePaste` into a shared test util so a future `DatePicker` paste test can reuse it.

packages/components/src/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,257 @@
...
+      as={DateRangePicker}
+      variant={variant}
+      size={size}
+      width={resolvedWidth}
+      {...props}
+      ref={ref}
+    >
+      <DateRangeInput
+        action={
+          <IconButton className={classNames}>
```
sarahgm, line 159: 💡 The calendar trigger `IconButton` has no explicit `aria-label`. This matches `DatePicker` exactly so it's not a regression, but worth confirming RAC's `DateRangePicker` exposes an accessible name for this toggle button to screen readers. If it does, no change needed.

packages/components/src/DateRangePicker/DateRangePicker.test.tsx
sebald: Good catch — fixed in ee30ae0e5. The test now actually exercises `dateUnavailable`: it queries the day cells and asserts that some carry `data-unavailable="true"` while *not all* do (the story blocks Sundays), so it would fail if `dateUnavailable` were a no-op. Mirrors the existing `data-unavailable` assertion in `DatePicker.test.tsx`.

packages/components/src/DateRangePicker/DateRangePicker.stories.tsx
sebald: Done in ee30ae0e5. Extracted `firePaste` into a shared module — and it was actually already copy-pasted in three places (`DateField.test.tsx`, `DatePicker.test.tsx`, and this story), so all three now import the single helper.

One important detail: I put it in its own **vitest-free** module `packages/components/src/firePaste.ts` rather than `test.utils.tsx`. `test.utils.tsx` imports `vitest`, and pulling that into a story’s import graph regresses DST-1564 (breaks the Storybook preview with the `customEqualityTesters` error) — same reasoning as `FileField/makeFile.ts`. Verified the story paste test still passes in-browser.

packages/components/src/DateRangePicker/DateRangePicker.tsx
sebald: Confirmed RAC exposes an accessible name, so no change needed. `useDateRangePicker` returns `buttonProps` with `aria-label: stringFormatter.format('calendar')` (→ localized "Calendar"), and RAC’s `DateRangePicker` injects those into `ButtonContext`. Our `IconButton` is a RAC `Button` that consumes that context and is the only button descendant, so it picks up the label automatically — the trigger is announced as "Calendar" to screen readers.

sarahgm (approved):

PR #5557, opened 2026-06-24
test(chromatic): disable redundant story snapshots (snapshot audit)

aromko:
# Description

Disables Chromatic snapshots on stories a snapshot audit flagged as visually redundant — pixel-confirmed duplicates of a base or showcase frame, or closed-trigger states whose open/expanded visual is captured by another (kept) story. Every play test and docs example is kept; only the snapshot is turned off via `parameters.chromatic.disableSnapshot`.

This is the **complete audit cut**, landed in three reviewable slices:
- **High-confidence** (`7540cedf0`) — pixel-identical duplicates.
- **Safe-bulk, medium/low** (`9312fe08a`) — near-identical frames.
- **Recheck rows** (`4c535d02e`) — higher-diff frames + the canonical `Dialog`/`Drawer`/`Select` `Basic` cuts, each verified to have its open state kept elsewhere (e.g. `Dialog.Basic` → `TextOnly`/`VeryLongContent`, `Select.Basic` → `SelectedScroll`/`WithImages`).

**Totals:** ~108 snapshots disabled across 39 story files, plus 3 deliberate **re-enables** (`disableSnapshot: false`) protecting kept `.test()` children whose named parent is now disabled (`Calendar` ×2, `Toast` ×1).

Mechanisms:
- **Named/base stories** — add the parameter directly. Disabling a base also covers its `.test()` children (they inherit params via `combineParameters`) — e.g. ComboBox `Mobile`.
- **Generated `.test()` stories** — the 3-arg `.test(name, annotations, fn)` overload disables only the derivative, not its base.
- **Badge** — `All` + `Icon` merged into the `Basic` showcase and removed; `Badge.test.tsx` updated to assert against the showcase.

Out of scope: 5 stories already disabling at the meta level (`Popover`, `Underlay`), and 2 audit rows that no longer exist on `main` (`Tooltip.ControlledTooltipTrigger` / `Single` — Tooltip now only has `Basic`/`OpenViaTooltipProp`).

## Screenshots / Preview

The only rendered change is the **Badge** stories (`Basic` is now the full variant showcase; `All`/`Icon` removed). Everything else is snapshot-config only and renders identically.

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. `pnpm install && pnpm typecheck:only` — passes clean.
2. `pnpm test:sb` — story tests pass (verified across all modified `.test()` components + Badge unit tests).
3. On the Chromatic build for this PR: confirm ~108 fewer snapshots, and that **kept** stories still have baselines — especially `Badge.Basic` (showcase), the three re-enabled `.test()` children, and the open-overlay stories that now solely cover Dialog/Drawer/Select.

## Breaking Changes

No

## Checklist

- [x] Stories added/updated (with \`component-test\` tag where applicable)
- [ ] Changeset added (\`pnpm changeset\`) — N/A, stories/tests only


sarahgm (commented):
## Automated Code Review

### Overview
| Field | Value |
|-------|-------|
| Files Changed | 39 |
| Additions | +297 |
| Deletions | -169 |
| Visual Regression | **Passed (up to date)** — ran on current head `4c535d0`, Chromatic: 424 tests unchanged |
| CI | All checks pass (typecheck, lint, unit, storybook tests, coverage, size-limit) |

### What this PR does
Adds `parameters.chromatic.disableSnapshot: true` to ~108 stories flagged as visually redundant, across 39 files. Three patterns:
- **Named/base stories** — parameter added directly (covers inherited `.test()` children).
- **Generated `.test()` stories** — uses the 3-arg `.test(name, annotations, fn)` overload to disable only the derivative.
- **3 deliberate re-enables** (`disableSnapshot: false`) on `Calendar` ×2 and `Toast` ×1, protecting `.test()` children whose now-disabled parent is the only thing that would otherwise carry the open-state snapshot.

The only rendered change is **Badge**: `All` merged into `Basic`, `Icon` removed, `Badge.test.tsx` re-pointed at the showcase.

### Checklist
- ✅ **Mechanical correctness** — all `disableSnapshot` additions are syntactically valid and CI-green (typecheck + tests pass).
- ✅ **`.test()` 3-arg overload** used correctly throughout; the `disableSnapshot: false` re-enables are well-reasoned (parent disabled, but the test's open/expanded state is the unique visual worth capturing).
- ✅ **Badge merge** — variant order verified: index `[4]` = info variant, matching the `bg-info-muted` inline snapshot; lock-icon count of 2 (master + admin) correct.
- ✅ **Badge test reformat** drops the `data-testid` in favor of accessible/text queries.
- ✅ **Changeset** correctly N/A (no published code change; stories/tests only).
- ✅ **Docs unaffected** — removing `Badge.All`/`Icon` stories did not break the docs build (Vercel docs deploy + Docs Script Tests pass).

### Issues Found

**Warnings (Should Fix):** None.

**Suggestions (Nice to Have):**
- **`Badge.test.tsx:5`** — `screen.getAllByText('Status')[4]` is positional: it silently asserts whatever variant sits at index 4. The inline snapshot (`bg-info-muted`) does protect against a wrong match, so it's self-correcting, but a reorder of the showcase would make the test's intent ("the info variant") and reality drift. Optional: assert on the variant explicitly rather than a raw index.
- **`Select.stories.tsx` (`LongItems`)** — `chromatic: { disableSnapshot: true, delay: 300 }` keeps `delay: 300`, now dead config since no snapshot is taken. Harmless; could be dropped for clarity.

### Note on coverage tradeoff
This permanently reduces visual coverage on ~108 frames. The premise (each is a pixel-confirmed duplicate of a kept frame) is sound and the current Chromatic build shows no regressions, but a future regression unique to a disabled story won't be caught. The audit's per-row verification and the re-enable safeguards mitigate this well.

### Recommendation
**Approve.** Clean, low-risk, well-documented test-config change. CI fully green, VRT up to date with no visual diffs, and the one behavioral change (Badge) is verified correct. The two suggestions are non-blocking.

---

sarahgm:
@aromko Can you write down in our Developer Giudelines when we should disable a Story from Chromatic? I think it would help to prevent running in the issue again 🤔

aromko:
@sarahgm Done — added a **"Disabling Chromatic snapshots"** section to the Development Guidelines (`docs/content/components/__internal__/component-guidelines.mdx`, under *Writing Stories*) in 1f6a9b40e.

It documents:

- **When to disable** — the story is a visual duplicate of another kept story, a closed/default trigger state whose open/expanded visual is already captured elsewhere, or an inherently flaky frame whose surface is covered by a stable story.
- **When to keep** — the story is the only one capturing a given visual (unique variant/size/overlay/edge state). When in doubt, keep it.
- That `disableSnapshot` only turns off the snapshot — the `play` test and any docs example stay.
- Mechanics — adding the param to a base (covers inherited `.test()` children), the 3-arg `.test(name, annotations, fn)` overload to disable a single derivative, re-enabling a kept child with `disableSnapshot: false`, and not leaving dead `chromatic.delay` config behind.

That gives us a single rule to point at next time so we don't end up re-litigating it per PR.

aromko:
Addressed both non-blocking suggestions from the review in 1f6a9b40e:

- **`Badge.test.tsx`** — replaced the bare positional `getAllByText('Status')[4]` with a named `VARIANT_ORDER` lookup (`getAllByText('Status')[VARIANT_ORDER.indexOf('info')]`) and renamed the test to `renders the info variant correctly`. The intent is now explicit and won't silently point at the wrong variant if the showcase is reordered — and the inline snapshot (`bg-info-muted`) still guards the markup. Unit tests pass.
- **`Select.stories.tsx` (`LongItems`)** — dropped the now-dead `delay: 300` (nothing is captured once the snapshot is disabled) and rewrote the stale comment, which still described stabilizing a Chromatic capture that no longer happens.

sarahgm (approved):

PR #5558, opened 2026-06-24
fix(DST-1559): remove dead/mis-named props from TextField, FileTrigger, Loader

sarahgm:
# Description

Cleans up three component-API defects found while auditing `beta-release` against `main`. The fixes have **no runtime behavior change** — the affected props never did anything — though two are type-level breaking changes (see below).

- **TextField** — removed `min`/`max`. They were never forwarded to the underlying `<input>` (react-aria filters them out of both the input and wrapper props), so they had no effect. Numeric bounds belong on `NumberField` (`minValue`/`maxValue`/`step`).
- **FileTrigger** — removed the mis-named singular `acceptedFileType`. Its key did not match react-aria's `acceptedFileTypes`, so file-type filtering silently never applied. The inherited (and working) `acceptedFileTypes` remains; `FileField` already uses it correctly.
- **Loader** — fixed the `loaderType` JSDoc, which documented a non-existent `cycle` value. The accepted values are `xloader` and `circle` (default `circle`), as the existing tests already assert.

All three are pre-existing (identical on `main`, not beta regressions). Versioned as a `patch`, matching the sibling cleanups DST-1553 / DST-1555 / DST-1557 — the major bump happens at the beta→main promotion, not per-changeset.

Closes DST-1559

## Test Instructions

1. `pnpm typecheck:only` — passes (no consumer in the repo referenced the removed props).
2. `pnpm test:unit` for FileField and Loader — file-type filtering (`FileField`/`fileUtils`) and `loaderType` (`xloader`/`circle`) remain green.

## Breaking Changes

**Type-level only — no runtime impact.** Removing the `min`/`max` props from `TextField` and the `acceptedFileType` prop from `FileTrigger` is a breaking change to the TypeScript surface: code that passes those props will no longer compile. Nothing changes at runtime, because the props were inert (filtered out before reaching the DOM / mis-named so react-aria never read them). This ships with the next major via `beta-release`, which is why it targets `beta-release` rather than a stable `main` patch.

Migration:
- `TextField` `min`/`max` → use `NumberField` with `minValue`/`maxValue`/`step` for numeric bounds.
- `FileTrigger` `acceptedFileType` → use the (already-working) `acceptedFileTypes`.

## Checklist

- [ ] Visual regression tests updated (for UI changes) — N/A, no visual changes

### DIFF
--- a/.changeset/dst-1559-fix-dead-mis-named-props.md
+++ b/.changeset/dst-1559-fix-dead-mis-named-props.md
@@ -1,0 +1,9 @@
+---
+'@marigold/components': patch
+---
+
+fix(DST-1559): remove dead and mis-named props from `TextField`, `FileTrigger`, and `Loader`.
+
+- `TextField`: drop the `min`/`max` props. They were never forwarded to the underlying `<input>` (react-aria filters them out of both the input and wrapper props), so they had no effect. Numeric constraints belong on `NumberField`.
+- `FileTrigger`: remove the mis-named singular `acceptedFileType` prop. Its key did not match react-aria's `acceptedFileTypes`, so file-type filtering silently never applied. Use the inherited `acceptedFileTypes` instead.
+- `Loader`: fix the `loaderType` JSDoc, which documented a non-existent `cycle` value. The accepted values are `xloader` and `circle` (default `circle`).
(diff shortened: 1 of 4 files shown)


.changeset/dst-1559-fix-dead-mis-named-props.md
```diff
@@ -0,0 +1,9 @@
+---
+'@marigold/components': patch
```
sebald:
```suggestion
'@marigold/components': major
```

API removed! :D

sebald (approved):

PR #5563, opened 2026-06-24
docs(DST-1474): document the ActionBar component

sebald:
# Description

Replaces the boilerplate ActionBar docs page with full, use-case-driven documentation. The page now frames `<ActionBar>` as a toolbar of bulk actions for the current selection in a collection (not a generic floating toolbar), and explains when to reach for it versus a `<ButtonGroup>` or a single `<Button>`.

Adds an anatomy diagram, a Do/Don't tile, three demos (bulk actions in a `<Table>` via the `actionBar` render prop, bulk editing through a `<Drawer>`, and driving the bar from custom selection state outside a Table), an accessibility section (labelled toolbar, live-region announcement, Escape-to-clear, focus restoration), and an "Alternative components" list. Prose is intent-led and kept light on API and implementation detail; content is deduplicated so each idea lives in one place.

Documentation only — no changes to the `ActionBar` component, `useActionBar`, `<Table>` integration, or theme styles.

Closes DST-1474

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm start` and open `/components/collection/actionbar`.
2. Confirm the anatomy diagram renders and the intro/usage/accessibility prose reads correctly.
3. Exercise each demo: select rows in the table demo (count + clear + Escape, and "All items selected" via the header checkbox), open the drawer demo's bulk-edit form, and toggle checkboxes in the standalone demo to show/hide the bar.

## Breaking Changes

No

## Checklist


jim761:
is it intentionally targeting the beta branch?

jim761:
This has nothing to do with the pr, but wouldn't it be good to show when all items are selected how many items were selected. In the docs it says it can select even more from other rows they cannot see , worst case they delete more than intended. also if it isn't a table it wont show all selected when theres no max value

jim761 (approved):

PR #5566, opened 2026-06-24
refactor(DST-1565): one surface model where the boundary encodes role and elevation encodes depth

sebald:
# Description

`theme-rui` drew edges two different ways with no rule for which to use: a `background-clip` gradient border on surfaces, and an opaque `border` on controls. DST-1565 replaces both with one model built on two independent axes.

- **The boundary encodes _role_.** A surface (Card, Panel, Dialog, Menu) wears `ui-surface`: a quiet translucent hairline. A control (Input, Select, field, the neutral Button family, SegmentedControl) wears `ui-surface-control`: the same charcoal stroke about 2.5× denser, so it reads as something to operate. Interactivity is not the divider. A selectable card is still a surface. The question is the element's role.
- **Elevation encodes _depth_.** The three `shadow-elevation-*` tiers carry only lift, independent of the boundary. Controls sit on the lowest tier because a surface contains them.

The two axes compose freely and never stand in for each other.

## What changed

**The boundary is a translucent ring, not a gradient-clip border.** The old surface border reserved a `1px solid transparent` border and showed a `border-box`-clipped gradient through it. That hijacked `background`, forced `ui-state-error` to override a `border-*` color, and got covered by edge-to-edge children (a `Table` in `Panel.Content bleed` ate the panel's left/right edge). It is now a 1px Tailwind `ring` in the `box-shadow` chain: translucent so it composites over any ground at consistent contrast, outset so a bleeding child can no longer cover it.

**Two boundary tokens, split by function.**

- `--color-surface-border` — the decorative rim of a surface. Whisper-light, below the contrast threshold, never load-bearing.
- `--color-control-border` — the functional edge of a control, about 2.5× denser. Translucent, so it adapts to its ground, including the SegmentedControl track where one fixed opaque gray went weak.
- `--color-border` narrows to its structural role (dividers, grid lines, table rules) and stays opaque, because crossing translucent strokes double-darken at their intersections.

**Elevation carries only lift.** Now that the ring owns the crisp edge, the shadows drop the harsh contact layer and spread across lighter, warm-tinted layers with negative spread, so each stays contained and a dense field of elevated elements reads calm.

**The dark surface uses the same model, inverted.** `ui-surface-contrast` (primary Button, ActionBar) is a crisp dark ring a shade darker than the fill, plus light on the face: a top-pooled glow, a gentle top-down gradient, and a soft top-edge highlight. Every layer derives from `--ui-background-color`, so hover retints the whole surface by swapping one token. `ui-surface-destructive` is the same recipe retinted red.

**Overlay content flattens inside its frame.** The `Popover` now owns the overlay surface (fill, rim, shadow). ListBox, Calendar, Menu, and Dialog render flat inside it and keep their own surface only when standalone. This fixes the double border a dropdown listbox drew on top of the popover.

## Review feedback addressed

- Disabled controls keep a dimmed boundary rather than a full opaque ring or none.
- The overlay double border is fixed by the Popover-owns-surface change above.
- Composite controls (`DateField`, `NumberField`) now set the control boundary on their wrapper, so they match the inputs beside them.
- `token-overview` and `elevation` docs are rewritten around the two-axis rule.

## Test instructions

1. `pnpm build:themes`
2. `/examples/filter` — the Venues panel keeps a full border while its table bleeds edge-to-edge.
3. `/examples/event-form` — control boundaries read consistent and firm; the date fields match the inputs beside them.
4. Storybook — Fields (plus error and disabled), Surfaces (Card / Panel / Dialog / Menu / Toast), Buttons (primary / secondary / destructive with hover, ToggleButton on-state, SegmentedControl), Overlays (dropdowns show a single frame with flat content).

## Risks / review notes

- **Outset ring under an `overflow-hidden` ancestor** could clip 1px. The ToggleButton group and the SegmentedControl indicator are the only such places and both have inner inset. Confirm in VRT.
- **The control border measures about 2:1 on white**, below WCAG 1.4.11's 3:1. Deliberate: the opaque keyboard focus ring remains the perceivable affordance. Tracked separately.
- **The overlay change touches the shared `Popover`** (Select, ComboBox, Autocomplete, TagField, Menu, DatePicker, ContextualHelp). Verified across all of them in Storybook; a Chromatic pass is still worthwhile.

## Breaking changes

None. Visual refinement plus additive tokens (`--color-surface-border`, `--color-control-border`, `--color-toggle-selected*`, `--contrast-*`). Component APIs and the `--ui-border-color` contract are unchanged.

## Checklist

- [x] Component documentation added/updated — `token-overview` and `elevation` rewritten around the boundary/elevation rule
- [x] Accessibility reviewed — focus ring unchanged; resting hairline decorative; disabled keeps a dimmed edge; control-border 3:1 gap documented and flagged
- [ ] Visual regression tests updated (for UI changes) — VRT pending


themes/theme-rui/src/ui.css
```diff
@@ -47,28 +47,26 @@
...
-        var(--ui-border-color, var(--color-border)),
-        oklch(from var(--ui-border-color, var(--color-border)) calc(l - 0.1) c h)
-      );
-
-  background-clip: padding-box, border-box;
-  background-origin: padding-box, border-box;
-  border: 1px solid transparent;
+  @apply rounded-surface relative transition-[box-shadow];
+  background: var(--ui-background-color, var(--color-surface));
+  @apply ring-1 ring-[var(--ui-border-color,var(--color-border-surface))];
```
jim761: ⚠️ **Implementation is a reasoned superset of DST-1565 — reconcile the ticket.** The ticket's Expected Outcome specifies an *inset*-ring keyed off `--color-border`, "visual result near-identical… expect only bevel-removal + corner-AA diffs." This ships an *outset* ring on a new translucent `--color-border-surface` token, plus a full elevation-shadow retune — a meaningfully larger visual change. The deviations are well-justified in the PR description, but the ticket's acceptance criteria are now stale. Please update DST-1565 (or note the scope change on it) so VRT approvers check against the right spec.

.changeset/dst-1565-ui-surface-hairline-ring.md
```diff
@@ -0,0 +1,20 @@
...
+
+`ui-surface` previously faked its border by reserving a `1px solid transparent` border and showing a `border-box`-clipped background gradient (with a darken-toward-bottom bevel) through it. That hijacked the entire `background` property and forced `ui-state-error` to override the border via a `border-*` color.
+
+The border is now a 1px Tailwind `ring` (routed through `--tw-ring-shadow`), so it composes into the same `box-shadow` chain as `shadow-elevation-*` without clobbering them. `background` is a plain fill again, and `ui-state-error` just swaps `--ui-border-color`. The `--ui-border-color` theming contract is unchanged.
+
+Two deliberate properties of the new hairline:
+
+- **Translucent.** The resting color is a new `--color-border-surface` token — the charcoal hue at low alpha (`oklch(from var(--color-charcoal-950) l c h / 0.12)`). Because it is translucent, the edge composites over whatever ground the surface sits on, so its contrast stays consistent on white, the page background, or a tinted panel — where a fixed gray reads heavy on white and washes out on gray. State overrides (focus → `--color-ring`, error → `--color-destructive-accent`) remain opaque on purpose.
```
jim761: ⚠️ **Alpha mismatch in the published release note.** `tokens.css` ships `/ 0.1` and the docs say `10%`, but this line (and the PR body) still say `0.12`. The changeset becomes a public changelog entry — please correct to `0.1`.

```suggestion
- **Translucent.** The resting color is a new `--color-border-surface` token — the charcoal hue at low alpha (`oklch(from var(--color-charcoal-950) l c h / 0.1)`). Because it is translucent, the edge composites over whatever ground the surface sits on, so its contrast stays consistent on white, the page background, or a tinted panel — where a fixed gray reads heavy on white and washes out on gray. State overrides (focus → `--color-ring`, error → `--color-destructive-accent`) remain opaque on purpose.
```

jim761 (commented):
Automated review (selected findings). The CSS mechanic is sound — verified the ring composes with `shadow-elevation-*` in one `box-shadow` chain rather than clobbering it. Notes below are scope/consistency/maintainability; the VRT gate was discussed separately.

themes/theme-rui/src/tokens.css
```diff
@@ -144,24 +155,31 @@
...
+    The ui-surface ring owns the crisp edge, so these carry only lift. Tuned
+    "crisp": tight offsets + negative spread keep each shadow contained inside
+    its own footprint, so a dense field of elevated elements stays calm instead
+    of bleeding into a haze. Light from above (positive y, no x), low contact
+    alpha, and a warm charcoal-hued tint (instead of pure black) so the lift
+    sits into the warm-neutral ground rather than punching against it.
   */
   --shadow-elevation-border:
-    0px 0.3px 0.3px oklch(0 0 0 / 0.14), 0px 1px 1.2px -3px oklch(0 0 0 / 0.12);
+    0px 1px 1px -0.5px oklch(0.15 0.02 54 / 0.108),
```
jim761: 💡 The shadow color literals the charcoal hue (`oklch(0.15 0.02 54 …)`) instead of deriving from `--color-charcoal-*`. If the palette hue ever shifts, these shadows silently drift off-palette. Either derive via `oklch(from var(--color-charcoal-950) …)` or add a brief comment noting `54` = charcoal hue, kept literal because L/C are tuned independently of the palette step.

themes/theme-rui/src/ui.css
```diff
@@ -120,9 +118,13 @@
...
-  /* Overrides the transparent border. Gradient would get too dark. */
-  @apply border-destructive-accent shadow-destructive-accent/20;
-  @apply focus:border-destructive-accent focus:outline-destructive-accent/20;
+  /* Recolors the ui-surface ring via --ui-border-color (opaque, so an invalid
+     field reads clearly, not as a faint translucent hairline). The focus:
+     re-assert keeps a focused-invalid field destructive instead of letting
+     ui-state-focus recolor the ring to --color-ring. */
+  --ui-border-color: var(--color-destructive-accent);
+  @apply shadow-destructive-accent/20;
+  @apply focus:outline-destructive-accent/20 focus:[--ui-border-color:var(--color-destructive-accent)];
```
jim761, line 160: 💡 The focused-invalid override relies on source order: `focus:[--ui-border-color:destructive]` only wins over `focus-visible:ui-state-focus`'s `--ui-border-color:ring` because it emits later at equal specificity. It works and the comment explains the intent, but it's order-fragile — worth an explicit eyeball on the focused-invalid story in VRT to confirm the ring stays destructive.

.changeset/dst-1565-ui-surface-hairline-ring.md
sebald: Fixed. The changeset now reads `oklch(… / 0.1)` and the PR body was corrected to `0.1` as well. (The token has since been renamed `--color-border-surface` → `--color-surface-border`, so the suggestion text is superseded.) Resolving.

themes/theme-rui/src/ui.css
sebald: Done — posted a scope-update note on DST-1565 documenting the deviations (outset vs inset ring, new translucent `--color-surface-border` token, reworked `ui-surface-contrast`, retuned elevation) plus the later additions (control-border darkening, button depth surfaces, edge-token contract), and flagged the VRT diff as wide-but-intended: https://reservix.atlassian.net/browse/DST-1565?focusedCommentId=570908 . The PR description has also been rewritten to the full scope. Resolving.

themes/theme-rui/src/tokens.css
sebald: Added an explainer in 9c1b7f045: the comment now notes `54` is the charcoal hue kept literal because the shadow's L/C are tuned independently of any palette step (a shadow isn't "charcoal-950 at low alpha"), with a reminder to update `54` if the palette hue shifts. Resolving.

themes/theme-rui/src/ui.css
sebald: Acknowledged — accurate. `ui-state-error` re-asserts the destructive color under `focus:` at equal specificity, so it relies on source order (there's an inline comment noting the intent). Leaving this thread open as a VRT-verification item: the focused-invalid story needs an explicit eyeball once Chromatic runs (VRT is still pending on this PR). Not changing the code — bumping specificity here would be awkward in the `@apply` context for a low risk.

sebald: Dug into the compiled CSS — it's actually **not** order-fragile; the re-assert wins by **specificity**, not source order.

Compiled selectors (dist), specificity as (id, class+attr+pseudo, element), `:where()` contributes nothing:
- `ui-state-focus` → `[data-theme="rui"] .focus\:ui-state-focus …[data-focused]` → ring = 1 class + 2 attrs = **(0,3,0)**
- the `focus:` re-assert → a *compound* `[data-theme="rui"] .invalid\:ui-state-error …[data-invalid] …[data-focused]` → destructive = 1 class + 3 attrs = **(0,4,0)**

The destructive ring outranks the focus ring by the extra `[data-invalid]` attribute, regardless of source order. The `focus:` re-assert *is* that specificity bump (without it, the level-1 `[data-invalid]` rule and the `focus:` rule would tie at (0,3,0) and order would decide — presumably what prompted the flag). Clarified the code comment in faeaf4622 so the re-assert isn't mistaken for a redundant duplicate and deleted. Resolving.

themes/theme-rui/src/tokens.css
```diff
@@ -73,6 +73,10 @@
   /* --- Primary --- */
   --color-primary: var(--color-charcoal-900);
   --color-primary-foreground: var(--color-charcoal-50);
+  /* Hover for the primary Button / contrast surface: the fill brightens a touch
+     so the glow intensifies ("lights up"). Darkening instead crushed the
+     gradient + glow into a flat near-black. */
+  --color-primary-hover: oklch(from var(--color-primary) calc(l + 0.05) c h);
```
sarahgm: this token is not documented, should be in token overview

themes/theme-rui/src/tokens.css
```diff
@@ -82,6 +86,16 @@
   --color-selected: var(--color-charcoal-300);
   --color-selected-bold: var(--color-charcoal-900);
   --color-selected-bold-foreground: var(--color-charcoal-50);
+
+  /* ToggleButton "on" = a sunken dark-gray surface. A lighter charcoal than
+     selected-bold/primary (charcoal-900) so a pressed toggle isn't mistaken for a
+     primary button. The edge is darker than the fill (derived from it, so they
+     stay coupled) to stay visible on the dark face — the faint surface-border
+     hairline vanishes here, making a grouped segment look shorter than its siblings. */
+  --color-toggle-selected: var(--color-charcoal-800);
```
sarahgm: this is a component specific token, we don't have currently, can this be also used for other components in the future? (also this is not documented in Token Overview)

themes/theme-rui/src/tokens.css
```diff
@@ -99,6 +122,10 @@
   --color-destructive-accent: var(--color-red-600);
   --color-destructive-bold: var(--color-red-600);
   --color-destructive-bold-foreground: var(--color-white);
+  /* Hover for the destructive Button: brightens to match primary's "light up". */
+  --color-destructive-bold-hover: oklch(
```
sarahgm: is this token really needed? We have ui-state-hover (using hover token), ui-state-hover-ghost (which uses  current/10 and also the destructive ghost button variant), would assume to think when destructive button uses  this token then the destructive ghost button also?

themes/theme-rui/src/ui.css
```diff
@@ -44,63 +48,85 @@
 }
 
 /*
- * Use for interactive UI elements like inputs, selects, and cards. Anything that
- * sits on the page background and needs a clear visual boundary.
+ * Interactive surfaces that sit on the page background and need a clear
+ * boundary: inputs, selects, cards, panels.
```
sarahgm: panels is not a interactive surface or?

themes/theme-rui/src/ui.css
```diff
@@ -44,63 +48,85 @@
...
+ * box-shadow chain — and adds the same top-lit gesture inverted for a light
+ * face: a faint top-down gradient plus a 1px white gloss on the top edge, so
+ * the surface reads as gently raised without going dark. The gradient starts a
+ * hair off-white (charcoal-50, i.e. --color-muted) so the white gloss is
+ * visible.
+ */
+@utility ui-surface-muted {
+  @apply ui-surface;
+  /* prettier-ignore */
+  background: linear-gradient(to bottom, var(--color-charcoal-50), var(--color-charcoal-100));
```
sarahgm: ui-surface-muted sounds to me as here would be the muted token be used.  Should we use it here?

themes/theme-rui/src/components/ToggleButton.styles.ts
```diff
@@ -17,15 +17,30 @@ export const ToggleButton: ThemeComponent<'ToggleButton'> = {
   button: cva({
     base: [
       'ui-button-base gap-2',
-      'ui-surface shadow-elevation-border',
+      // Neutral button = the secondary Button look (muted gloss surface).
+      'ui-surface-muted shadow-elevation-border',
+      // Flip on/off in one frame: the fill is instant, so an animated box-shadow
```
sarahgm: too long comments in this file, can you shorten/remove it?

themes/theme-rui/src/ui.css
```diff
@@ -132,7 +163,12 @@
   @apply text-disabled cursor-not-allowed;
   background: var(--color-disabled-surface);
   background-clip: border-box;
-  box-shadow: none;
+  /* No top highlight when disabled: the inset gloss/sheen of ui-surface-muted and
+     ui-surface-contrast would otherwise sit on the flat disabled fill. Zeroing the
+     `--tw-inset-shadow` var is reliable where a raw `box-shadow: none` lost the
+     cascade to the surface utilities' own composite box-shadow. Plain surfaces and
+     inputs have no inset, so this is a no-op for them — their boundary ring stays. */
+  @apply inset-shadow-none;
```
sarahgm, line 175: ⚠️ **Disabled controls now keep their boundary ring — this contradicts both the ticket and the PR's test instructions.**

Switching the old `box-shadow: none` to `inset-shadow-none` is the right move for preserving the composite box-shadow cascade — but it only zeroes `--tw-inset-shadow`. Nothing resets `--tw-ring-shadow` (no `disabled:ring-0` anywhere), so the outset ring **persists on disabled**. For `ui-input` controls that's the opaque, freshly-darkened `--color-border` (L 0.78) on a `charcoal-200` fill.

Three-way disagreement worth resolving explicitly:
- **DST-1565**: *"its `box-shadow: none` now also removes the ring — matches prior behavior"*
- **PR test instructions**: *"disabled (flat fill, **no ring**)"*
- **This comment + implementation**: *"their boundary ring stays"*

If the ring should stay, fix the ticket + test-instruction wording. If it shouldn't, add `disabled:ring-0` (or reset `--tw-ring-shadow`). Please confirm against the Chromatic diff before promoting toward `main`.

themes/theme-rui/src/ui.css
```diff
@@ -44,63 +48,85 @@
...
+ * the edge and glow together.
  *
- * `background` is intentionally excluded from the transition list: hover/state
- * bg flips should feel instant on high-frequency interactions (see DST-1436).
+ * `background` is excluded from the transition so hover/state bg flips feel
+ * instant on high-frequency interactions (DST-1436).
  */
 @utility ui-surface-contrast {
-  @apply rounded-surface relative transition-[border,box-shadow];
+  @apply rounded-surface relative transition-[box-shadow];
```
sarahgm, line 91: ⚠️ **Scope note (process, not code):** DST-1565 explicitly scoped `ui-surface-contrast` as *"Not included … left unchanged"*, and the ticket's declared scope was just the `ui-surface` ring + `ui-state-error` + `ui-state-disabled` comment. This PR reworks `ui-surface-contrast`, adds `ui-surface-muted`/`ui-surface-destructive`, retunes elevation, darkens `--color-border` system-wide, and migrates the whole Button family + Menu + ToggleButton. The PR body is transparent about the growth ("keep open"), but please update the ticket description (or split the surface work into its own ticket) so the scope/appetite record stays honest.

themes/theme-rui/src/ui.css
```diff
@@ -44,63 +48,85 @@
...
-        var(--ui-border-color, var(--color-border)),
-        oklch(from var(--ui-border-color, var(--color-border)) calc(l - 0.1) c h)
-      );
-
-  background-clip: padding-box, border-box;
-  background-origin: padding-box, border-box;
-  border: 1px solid transparent;
+  @apply rounded-surface relative transition-[box-shadow];
+  background: var(--ui-background-color, var(--color-surface));
+  @apply ring-1 ring-[var(--ui-border-color,var(--color-surface-border))];
```
sarahgm, line 66: 💡 Deliberate deviation from the ticket's suggested solution, flagging so a reviewer reading DST-1565's code block doesn't read it as a mistake: the ticket proposed `inset-ring-1`; this ships outset `ring-1`. The justification (paints above edge-to-edge children, fixes the disappearing-border bug) is sound and an improvement. Note the disabled-ring behavior in the other comment would exist with `inset-ring` too — it's a separate, non-`--tw-shadow` var either way.

themes/theme-rui/src/ui.css
sebald: Correct — fixed in `905d646e7`. Reworded the comment header from "Interactive surfaces" to just "Surfaces"; panels/cards aren't interactive. Resolving.

themes/theme-rui/src/ui.css
sebald: Good catch — aligned in `905d646e7`. `ui-surface-muted`'s gloss gradient now builds on `var(--color-muted)` (was a raw `charcoal-50` literal), so the name matches the token it uses. Resolving.

themes/theme-rui/src/components/ToggleButton.styles.ts
sebald: Done in `905d646e7` — trimmed the verbose block comments in `ToggleButton.styles.ts` down to one-liners. Resolving.

themes/theme-rui/src/ui.css
sebald: Resolved the three-way disagreement by **keeping the ring but dimming it**, rather than removing it (`a478d4856`). `ui-state-disabled` now sets `--ui-border-color: var(--color-disabled)`, so a disabled control keeps a faint, clearly-inactive boundary instead of going borderless — a borderless disabled input read as broken. Wording aligned on both sides: DST-1565's Scope bullet and the PR test instructions now say "flat fill, **dimmed boundary ring**" (was "no ring"). Resolving.

themes/theme-rui/src/ui.css
sebald: Done — DST-1565's description was updated to the grown scope, with a scope-update note on the ticket ([focusedCommentId=570908](https://reservix.atlassian.net/browse/DST-1565?focusedCommentId=570908)) documenting the additions: `ui-surface-contrast` rework, new `ui-surface-muted`/`-destructive`, elevation retune, system-wide `--color-border` darkening, and the Button + Menu + ToggleButton migration. Keeping it on one ticket per the "keep open" note rather than splitting. The PR body's new "Review feedback addressed" section mirrors this. Resolving.

themes/theme-rui/src/ui.css
sebald: Thanks — agreed, and it's intentional. The ticket's `inset-ring-1` sketch ships as an outset `ring-1` precisely for the reasons you note (paints above edge-to-edge children, fixes the disappearing-border bug). It's flagged as a deliberate deviation in the PR body and in the scope-update note on DST-1565, so the ticket's code block isn't mistaken for the source of truth. Resolving.

themes/theme-rui/src/tokens.css
sebald: You're right, it's not in the Token Overview. `--color-primary-hover` (and its sibling `--color-destructive-bold-hover` below) are *derived* hover colors — `oklch(from <fill> calc(l + 0.05) c h)` — added so the button-family "lift/light-up" on hover isn't a magic literal buried in the utility. They read as implementation detail rather than a consumer-facing knob, which is why they weren't listed. I'm happy to add them to the overview's color section if we'd rather treat them as public semantic tokens — leaving this open for a quick call on whether the hover tokens belong in the overview or stay internal.

themes/theme-rui/src/tokens.css
sebald: Correct — `--color-toggle-selected` / `--color-toggle-selected-edge` are ToggleButton-specific today: the sunken dark-gray "on"/pressed surface (a *lighter* charcoal than `selected-bold`/primary, so a pressed toggle isn't mistaken for a primary button). Nothing else needs it yet, so I scoped it narrowly rather than generalizing prematurely. It could absolutely become a shared "pressed/selected control surface" token later — if/when a second consumer shows up, we'd rename it to something non-component-specific at that point. And yes, not in the Token Overview (same open doc question as the hover tokens above). Leaving open pending that decision.

themes/theme-rui/src/tokens.css
sebald: It drives the **solid** destructive Button's hover: it brightens the bold red fill (`l + 0.05`) so the surface "lights up" like the primary button, instead of darkening (which crushed these already-saturated fills toward flat). The destructive **ghost** variant is a different hover mechanism — it stays on `ui-state-hover-ghost` (the `current/10` translucent overlay on a transparent button), so it doesn't consume, and doesn't need, this token. So it isn't redundant with the ghost hover; they're two different surfaces. The token exists (rather than an inline `oklch()` calc in the utility) purely for parity with `--color-primary-hover`. If you'd prefer, I can inline it and drop the token — your call. Leaving open.

sarahgm: since its only used for one button variant, i would avid making so much new tokens, would go for inline like before

themes/theme-rui/src/tokens.css
sarahgm: maybe a note as comment that its an internall token and should not be documented in the token overview page?

themes/theme-rui/src/tokens.css
sarahgm: so `--color-primary-hover` is also just used by one component? Does it then make sense to make it a token?

sebald: Agreed and done — inlined in `63f72aed8`. The primary hover is now `hover:[--ui-background-color:oklch(from_var(--color-primary)_calc(l-0.15)_c_h)]` directly on the Button, so there's no single-use token and nothing to add to the Token Overview.

themes/theme-rui/src/tokens.css
sebald: Removed in `63f72aed8`. The ToggleButton "on" state now reuses the existing `--color-selected-bold` token (plus `shadow-none`) instead of a component-specific one, so there's no internal token left to document.

themes/theme-rui/src/tokens.css
sebald: Agreed — inlined like the primary in `63f72aed8` (`hover:[--ui-background-color:oklch(from_var(--color-destructive-bold)_calc(l-0.15)_c_h)]`). No new token.

sebald:
Hey @sarahgm @aromko 👋

Sorry for the long wait on this one! But I think I've finally figured out a good "systematic" way to do the surfaces stuff that actually makes sense. I really hope you see this as a good direction as well :D

The PR sadly got a little more updated now. The idea is explained on the Elevation page:
👉 https://marigold-docs-git-refa-dst-1565-ui-surface-shadow-ring-marigold.vercel.app/foundations/elevation

The short version: the **boundary** encodes an element's *role* (surface vs. control), and **elevation** encodes its *depth*. Two independent axes that compose freely.

themes/theme-rui/src/tokens.css
sebald: Removed. `--color-primary-hover` no longer exists. The primary Button derives its hover inline with `oklch(from var(--color-primary) calc(l - 0.15) c h)`, so there is no standalone token left to document. Resolving.

themes/theme-rui/src/tokens.css
sebald: Removed. The ToggleButton on-state no longer uses a dedicated token. It now reuses the shared `--color-selected-bold` fill and `--color-hover`, so there is nothing component-specific left to document or generalise. Resolving.

themes/theme-rui/src/tokens.css
sebald: Removed and inlined, as you suggested. The solid destructive Button derives its hover inline with `oklch(from var(--color-destructive-bold) calc(l - 0.15) c h)`, matching the primary Button. No new token. Resolving.

sarahgm (approved):

PR #5571, opened 2026-06-26
chore(DST-1503): migrate Checkbox, Switch, Radio off deprecated react-aria-components exports

sarahgm:
# Description

Migrates `Checkbox`, `Switch`, and `Radio` off the deprecated react-aria-components single-element exports (`Checkbox`/`Switch`/`Radio`) to the `*Field` + `*Button` composition introduced in `react-aria-components@1.18.0`. This removes the `ts(6385)` deprecation warnings.

The migration unavoidably adds one wrapper `<div>` per control (the RAC `*Field`), because `*Button` reads a private context that only `*Field` provides. The `group/*` class, row layout, label, indicator, and forwarded `ref` stay on `*Button`, so the `HTMLLabelElement` ref contract and theme `group-*/…` selectors are preserved. No change to the public Marigold API, behavior, or visual output.

Closes DST-1503

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm typecheck:only` — confirm no `ts(6385)` deprecation warnings remain for these three components.
2. Run `pnpm test:unit` and `pnpm test:sb` — Checkbox/Switch/Radio suites pass (incl. the new Checkbox `aria-describedby` test and updated Switch snapshots).
3. In Storybook, exercise `Components/Checkbox`, `Components/Switch`, `Components/Radio` (and their groups) — verify layout, width-constrained variants, focus, hover, and disabled states are unchanged.

## Breaking Changes

No

## Checklist

### DIFF
--- a/packages/components/src/Radio/Radio.test.tsx
+++ b/packages/components/src/Radio/Radio.test.tsx
@@ -7,9 +7,10 @@ test('takes full width by default', () => {
   render(<Basic.Component />);
 
   const radio = screen.getByText('Option 1');
+  // Width lives on the RadioField wrapper, the RadioButton label's parent.
   // eslint-disable-next-line testing-library/no-node-access
-  const container = radio.closest('[data-rac]') || radio.parentElement!;
-  expect(container).toHaveClass('w-full');
+  const field = radio.closest('label')?.parentElement;
+  expect(field).toHaveClass('w-full');
 });
 
 test('forwards ref', () => {
--- a/packages/components/src/Radio/Radio.test.tsx
+++ b/packages/components/src/Radio/Radio.test.tsx
@@ -6,10 +6,12 @@ import { Basic, Error } from './Radio.stories';
 test('takes full width by default', () => {
   render(<Basic.Component />);
 
-  const radio = screen.getByText('Option 1');
-  // Width lives on the RadioField wrapper, the RadioButton label's parent.
+  // Width lives on the RadioField wrapper (a `div`). RadioButton renders a
+  // `label`, so `closest('div[data-rac]')` lands on the field regardless of how
+  // deeply RAC nests the label internally.
+  const radio = screen.getByRole('radio', { name: 'Option 1' });
   // eslint-disable-next-line testing-library/no-node-access
-  const field = radio.closest('label')?.parentElement;
+  const field = radio.closest('div[data-rac]');
   expect(field).toHaveClass('w-full');
 });
 
--- a/packages/components/src/Switch/Switch.tsx
+++ b/packages/components/src/Switch/Switch.tsx
@@ -72,6 +72,9 @@ const _Switch = forwardRef<HTMLLabelElement, SwitchProps>(
       ...rest,
     } satisfies RAC.SwitchFieldProps;
     return (
+      // Width is split across two levels: the outer SwitchField sizes itself to
+      // the `width` prop via the `--width` CSS variable, and the inner
+      // SwitchButton fills that width with `w-full`.
       <SwitchField
         {...props}
         className="w-(--width)"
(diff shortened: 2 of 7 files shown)


packages/components/src/Radio/Radio.test.tsx
```diff
@@ -7,9 +7,10 @@ test('takes full width by default', () => {
   render(<Basic.Component />);
```
OsamaAbdellateef, line 8: 🔴 **Critical — Visual Regression Tests not run**

This PR changes UI-affecting component source files (`Checkbox.tsx`, `Radio.tsx`, `Switch.tsx`) but the `Visual-Regression-Tests` workflow has not run on this branch at all.

packages/components/src/Radio/Radio.test.tsx
```diff
@@ -7,9 +7,10 @@ test('takes full width by default', () => {
...
 
   const radio = screen.getByText('Option 1');
+  // Width lives on the RadioField wrapper, the RadioButton label's parent.
   // eslint-disable-next-line testing-library/no-node-access
-  const container = radio.closest('[data-rac]') || radio.parentElement!;
-  expect(container).toHaveClass('w-full');
+  const field = radio.closest('label')?.parentElement;
+  expect(field).toHaveClass('w-full');
```
OsamaAbdellateef, line 15: ⚠️ **Warning — fragile DOM traversal**

`radio.closest('label')?.parentElement` will break silently if RAC changes `RadioField`'s render structure in a future release. Consider a more stable selector — e.g. targeting a `data-rac` attribute or the nearest role boundary — rather than relying on raw parent traversal.

packages/components/src/Switch/Switch.tsx
```diff
@@ -70,24 +70,28 @@ const _Switch = forwardRef<HTMLLabelElement, SwitchProps>(
...
+      <SwitchField
         {...props}
-        ref={ref}
-        className={cn(
-          'group/switch flex w-(--width) items-center gap-[1ch]',
-          classNames.container
-        )}
+        className="w-(--width)"
         style={createWidthVar('width', width)}
       >
```
OsamaAbdellateef, line 82: 💡 **Suggestion — clarify the two-level width split**

The `w-(--width)` on `SwitchField` + `w-full` on `SwitchButton` pattern is non-obvious without the PR description. A short inline comment would help future maintainers understand that the outer wrapper sizes to the width prop and the inner button fills it.

OsamaAbdellateef (commented):
## Automated Code Review

Clean, well-scoped migration. The `*Field` + `*Button` composition is correctly wired for all three components, `forwardRef`/ref contracts are preserved, prop renaming is intact, and the new `aria-describedby` test for `Checkbox` is a nice addition.

One blocker before merging: **VRT has not run** on this branch (see inline comment). Once Chromatic passes, this is good to go.

---

packages/components/src/Radio/Radio.test.tsx
sarahgm: Triggered the **Visual-Regression-Tests** workflow via `workflow_dispatch` (the `/run-chromatic` comment did not start a run on this branch). Run in progress: https://github.com/marigold-ui/marigold/actions/runs/28424148531

packages/components/src/Radio/Radio.test.tsx
sarahgm: Fixed in c2e873602. The assertion now anchors on the radio role and targets the `RadioField` via `closest('div[data-rac]')` instead of `closest('label').parentElement`. Since `RadioField` renders a `div` and `RadioButton` a `label`, this lands on the field regardless of how RAC nests the label internally.

packages/components/src/Switch/Switch.tsx
sarahgm: Added an inline comment in c2e873602 explaining the two-level split: the outer `SwitchField` sizes to the `width` prop via the `--width` CSS variable, and the inner `SwitchButton` fills it with `w-full`.

OsamaAbdellateef (approved):

PR #5572, opened 2026-06-26
feat(DST-1566): add a general-purpose Toolbar component

sebald:
# Description

Adds a general-purpose, accessible **`Toolbar`** — a single navigable region for the controls that act on a view (search, filters, actions). It gives mixed controls one tab stop with arrow-key roving focus (`role="toolbar"`, horizontal), and collapses overflowing actions into a measured **More** menu so the bar never wraps or scrolls.

This generalizes the `ButtonGroup` pattern (a RAC `Toolbar` hard-wired to buttons) to host mixed controls — a `SearchField`, a `Select`, a `Switch`, a `Menu`, alongside actions — which the documented Filter pattern (DST-1434) needs but had to assemble ad-hoc with `<Inline>` (no toolbar role, no roving focus, no accessible name).

Closes DST-1566

## Public API

```tsx
<Toolbar aria-label="Library" variant? size?>
  <SearchField … />            {/* static controls stay put; adapt via context */}
  <Switch … /> <Menu … />
  <Toolbar.Separator />
  <Toolbar.Action id="export" label="Export" icon={<Download />} />  {/* collapses */}
</Toolbar>
```

- **`Toolbar`** — container. Single tab stop + arrow roving, never wraps/scrolls. `variant`/`size` theme the bar only (not cascaded to children). Warns in dev if it has no accessible name.
- **`Toolbar.Action`** — a collapsing action, described as data. With an `icon` it renders an icon-only `<Button>` and auto-wires its `aria-label` + a tooltip from `label`; otherwise a text button. Folds right-to-left into **More** and returns as space allows. Always the secondary Button variant.
- **`Toolbar.Group`** — semantic cluster (`role="group"`), cascades `disabled` to nested buttons via `ButtonContext`.
- **`Toolbar.Separator`** — vertical divider (`role="separator"`).

## Design decisions

The whole design hinges on one question: **how does an author declare an action that can fold into a “More” menu?** We tried three answers before landing on the descriptor.

**Why `Toolbar.Action` is a descriptor (not real children we sniff).** An overflowing action must render in two forms — a **button** in the bar or a **menu item** in More. Earlier versions sniffed real `<Button>` children by component type and reconstructed menu items from them. That is lossy (icon/handler/ref don't map cleanly) and broke silently the moment a button was wrapped (e.g. in `<Tooltip.Trigger>`, which `child.type === Button` no longer matches — the exact case icon actions need). A descriptor (`<Toolbar.Action>` renders `null`) makes the toolbar the single source of truth: one set of props, rendered as a button *or* a menu item, with the identity check (`child.type === ToolbarAction`) reliable because a descriptor is never wrapped. This matches **`Breadcrumbs.Item`** and **`Sidebar.Item`**, the existing null-rendering collapse descriptors in the library.

**Why not a `Toolbar.Collapsible` / `Toolbar.Actions` wrapper holding real `<Button>`s?** Considered and rejected. A wrapper of real buttons forces one of two paths: (a) **reconstruct** menu items from the buttons — this *is* the brittle, Tooltip-breaking path above (we shipped it briefly as `Toolbar.Actions` and removed it); or (b) **relocate** the live buttons into a **popover** instead of a menu — robust, but it trades away `menu`/`menuitem` accessibility semantics and adds focus/state-management hazards (moving a focused, possibly-stateful element in and out of a closed overlay). A popover-based `Toolbar.Collapsible` is a reasonable *future addition* if we ever need to collapse heterogeneous stateful controls (a `Select`, a `SearchField`) — not a replacement for the action menu.

**Is the descriptor a detour from slot-aware components (`Button`, `Title`)?** No — slot-aware adaptation and collection descriptors are orthogonal tools. Slot context adapts a self-rendering control's *props* (variant/size/disabled); it cannot change *what* a control renders as. The toolbar uses slot-aware composition for everything static (real `<Button>`/`<SearchField>` children; `Toolbar.Group` cascades `disabled` via `ButtonContext`), and the descriptor only for the one thing slots can't express: button-or-menu-item. `Toolbar.Action` also *renders the real slot-aware `<Button>`* in the bar — it chooses, it doesn't re-implement.

**No control sub-component aliases.** No `Toolbar.SearchField`/`Toolbar.Select`. A survey of the library shows nothing aliases a standalone control as a sub-component (ActionBar wraps a button only to inject theming, which the toolbar deliberately doesn't do). Membership — “any single focusable control belongs” — stays in the docs, not a faked closed set in the type system.

## Test Instructions

1. `pnpm sb` → **Components/Toolbar**. Try `Basic`, `Groups`, `MixedControls`, `Overflow`, `IconActions`.
2. In `Overflow`/`IconActions`, drag the resizable container narrower — actions fold right-to-left into **More** and return as you widen; the bar never wraps or scrolls; the search field is never squished.
3. Keyboard: focus a control and use ←/→ to rove across all controls as one tab stop; separators are skipped.
4. `pnpm test:sb --run Toolbar` → 7/7 story tests pass. `pnpm typecheck:only` clean.

> Note: the overflow menu open/close can't be asserted in `test:sb` (overlays portal to `#storybook-root`, absent in the vitest-browser env); collapse is verified via button-count/scrollWidth instead. Works in a real browser.

## Breaking Changes

No. New component. The `pinned` prop briefly added to `Button`/`IconButton`/`LinkButton` during development was removed in the same branch and never released.

## Checklist

- [x] Accessibility reviewed against ARIA APG (toolbar pattern: roving focus, group/separator roles, accessible-name warning)
- [ ] Visual regression tests updated (not run — will request before dispatching)
- [x] Changeset added
- [ ] Storybook preview and Marigold docs preview are available (CI)


sebald:
Stopping work on the general-purpose `<Toolbar>` and closing this as **Won't do**.

## Why we're stopping

The Filter pattern (DST-1434 / #5522) was the **only concrete consumer** that justified this component. Revisiting it for the quick-filter requirement surfaced that the filter "toolbar" is **not actually a WAI-ARIA `toolbar` widget** — it's an input-dominated *filter bar* (search field + several selects + a switch + an "All filters" button), and the ARIA toolbar model is the wrong fit for it:

- A `role="toolbar"` widget collapses its controls into a **single tab stop with arrow-key roving focus** between them. That directly conflicts with a `SearchField` (and `ComboBox`, `DatePicker` segments, `NumberField`), where the arrow keys must drive the text cursor / value, not move focus.
- Verified at the source level: in react-aria 3.49.0, `useToolbar`'s `onKeyDown` unconditionally calls `focusNext` / `focusPrevious` + `preventDefault()` on the arrow keys with **no exception for text inputs**. So a search field placed in our toolbar loses its cursor keys. The headline use case is broken by design, not by our wrapper.
- The Filter pattern already documents this correctly and deliberately: it composes the bar with `<Inline>` (normal tab stops, each control owns its own keyboard) and explicitly warns *"Do not turn the toolbar into a `toolbar` widget."* The WAI-ARIA APG agrees — an input-dominated region should not be a toolbar.

So the premise in this ticket ("consumers who need to group search + filters + actions with toolbar semantics have nothing to reach for") does not hold: the filter bar does not *want* toolbar semantics, and `<Inline>` is the correct tool for it.

## What about the legitimate use case?

There is a real but narrower one — **action / command bars** (rows of buttons, icon actions, toggles, menus) where roving focus *is* correct and collapsing overflow into a "More" menu adds value. But:

- We have **no concrete consumer** asking for that today; this ticket was driven entirely by the filter pattern.
- The filter pattern itself is **still being refined**, so nothing is currently pulling on a toolbar.

Rather than ship a component whose only motivating consumer turned out not to need it, we stop now. If a real action-bar need appears later, we would re-open with a tighter scope: an **action / command bar** (buttons only, where roving is correct), explicitly **not** a search / filter bar, with the docs built around actions instead of search.

## Cleanup

- Local branch `dst-1566-toolbar-component` deleted.
- Draft PR #5572 closed and its remote branch deleted.

PR #5577, opened 2026-06-29
feat(DST-1574): surface example apps in the docs cmdk search

sebald:
# Description

The interactive example demos under `docs/app/(examples)/` were not discoverable from the docs ⌘K command menu — the fumadocs content index only covers MDX docs, and the "Jump to …" quick actions were built solely from the docs page tree.

This flattens the `(examples)` navigation config into the existing `PagesContext` quick-action path, so example demos now appear in the cmdk search:

- They show as **`Example: …`** quick actions (distinct icon/label from the docs `Jump to`).
- They match by name **or** by typing `example` (or any substring of it).
- Only **standalone demos** are surfaced. A self-contained app (App Shell) collapses to a single entry; its internal screens (Billing, Users, Teams, …) are intentionally not listed as separate examples. Pure sidebar groupings (Form) expand to their individual demos.

Resulting entries: App Shell, Filter, Event Form, Settings Form, Auto-Save Settings, Inventory.

Closes DST-1574

## Preview

Verified in-browser against the running docs app: searching `example`, an example name (e.g. `filter`), or `app shell` surfaces the demos and navigates correctly (e.g. `Example: App Shell` → `/examples`). No empty/internal entries (Billing, Users, …) appear.

## Test Instructions

1. Run `pnpm start` and open [localhost:3000](http://localhost:3000).
2. Open the search dialog (⌘K / click **Search**).
3. Type `example` — confirm the six standalone demos appear as `Example: …` actions.
4. Type `app shell` — confirm `Example: App Shell` appears and selecting it navigates to `/examples`.
5. Type `billing` — confirm no `Example:` entry appears (it's an App Shell internal screen, only the docs result remains).

## Breaking Changes

No

## Checklist

- [ ] Visual regression tests updated (for UI changes) — N/A, no component/theme changes

### DIFF
--- a/docs/app/(examples)/examples/navigation.tsx
+++ b/docs/app/(examples)/examples/navigation.tsx
@@ -1,5 +1,5 @@
 import { people } from '@/lib/data/people';
-import type { NavSection, ShellConfig } from '../_shared';
+import type { NavNode, NavSection, ShellConfig } from '../_shared';
 
 const appShellDocs = {
   docsHref: '/patterns/layout/app-frame',
@@ -99,3 +99,35 @@ export const config: ShellConfig = {
   // Label the dynamic `users/[id]` segment with the member's name.
   resolveLabel: id => people.find(person => person.id === id)?.name,
 };
+
+// Flatten the nav tree into the set of *standalone* example demos for the docs
+// cmdk search. Only standalone demos are surfaced — not the internal screens of
+// a larger demo. GroupLabels and Separators are not routes, so we skip them.
+//
+// A group with an index child (empty slug) is a single self-contained app
+// (e.g. "App Shell", whose Users/Billing/… screens are internal navigation, not
+// standalone examples): we collapse it to one entry pointing at its landing. A
+// group without an index is just a sidebar grouping (e.g. "Form"), so we recurse
+// and surface each child as its own demo.
+const collectDemos = (
+  items: NavNode[],
+  base: string
+): { name: string; url: string }[] =>
+  items.flatMap(item => {
+    if (item.kind !== 'Item') return [];
+    if (item.children) {
+      const isSelfContainedApp = item.children.some(
+        child => child.kind === 'Item' && !child.children && child.slug === ''
+      );
+      return isSelfContainedApp
+        ? [{ name: item.label, url: base }]
+        : collectDemos(item.children, base);
+    }
+    return [
+      { name: item.label, url: item.slug ? `${base}/${item.slug}` : base },
+    ];
+  });
+
+/** All standalone example demos as `{ name, url }` entries from the nav config. */
+export const examplePages = (): { name: string; url: string }[] =>
+  config.sections.flatMap(section => collectDemos(section.items, config.base));
--- a/.changeset/examples-in-cmdk-search.md
+++ b/.changeset/examples-in-cmdk-search.md
@@ -1,0 +1,7 @@
+---
+'@marigold/docs': patch
+---
+
+feat(DST-1574): surface example apps in the docs ⌘K search
+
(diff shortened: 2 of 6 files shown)


docs/app/(examples)/examples/navigation.tsx
```diff
@@ -99,3 +99,35 @@ export const config: ShellConfig = {
...
+  base: string
+): { name: string; url: string }[] =>
+  items.flatMap(item => {
+    if (item.kind !== 'Item') return [];
+    if (item.children) {
+      const isSelfContainedApp = item.children.some(
+        child => child.kind === 'Item' && !child.children && child.slug === ''
+      );
+      return isSelfContainedApp
+        ? [{ name: item.label, url: base }]
```
jim761, line 123: 💡 The self-contained-app collapse hardcodes `url: base`, assuming the empty-slug index sits exactly at `/examples`. Correct for the single App Shell today, but a second self-contained app would also collapse to `/examples` and collide. Deriving the URL from the empty-slug child (or a short comment noting the one-app assumption) would be more robust against future nav edits.

docs/app/(examples)/examples/navigation.tsx
```diff
@@ -99,3 +99,35 @@ export const config: ShellConfig = {
...
+        : collectDemos(item.children, base);
+    }
+    return [
+      { name: item.label, url: item.slug ? `${base}/${item.slug}` : base },
+    ];
+  });
+
+/** All standalone example demos as `{ name, url }` entries from the nav config. */
+export const examplePages = (): { name: string; url: string }[] =>
```
jim761, line 132: 💡 `collectDemos`/`examplePages` is pure with non-trivial branching (self-contained detection, recursion, GroupLabel/Separator skipping) — worth a small unit test locking the six-entry expectation so a future nav edit can't silently drop or duplicate a demo.

Minor: since `config` is static, `examplePages` could be a `const` array rather than a function that recomputes on each call — negligible, purely cosmetic.

jim761 (commented):
Automated review — non-blocking suggestions. Traced the flatten logic against the live nav config; it produces exactly the six documented demos. Two optional notes below.

aromko (approved):

docs/app/(examples)/examples/navigation.tsx
sebald: Added the unit test in 1f01b2614 — `docs/app/(examples)/examples/navigation.test.ts` (5 cases):

- `toEqual` on the full six-entry `{ name, url }` list — locks the exact contract, so a nav edit that drops/duplicates a demo breaks the build.
- self-contained-app collapse (App Shell → single entry at `config.base`)
- internal screens *not* surfaced (Users/Teams/Billing/Analytics/General/Security)
- sidebar-grouping recursion (Form expands to its three children; "Form" itself isn't an entry)
- non-route nodes skipped (the "Settings" `GroupLabel`)

It's picked up by the existing **Docs Script Tests** CI job (`pnpm --filter @marigold/docs test`, matches the `app/**/*.test.ts` glob).

On the `const` vs function point: leaving it as a function intentionally — it's the exported API called once from `layout.tsx`, and a `const` would just move evaluation to import time for no real gain. Since you flagged it as cosmetic anyway, resolving this thread.

docs/app/(examples)/examples/navigation.tsx
sebald: Looked into this — decided not to change it. The detection keys on `child.slug === ''`, and an empty slug *is* the index route at `base` (`/examples`). In Next.js only one route can be the `/examples` index, so a second self-contained app would have to live at a nested path (`/examples/foo`) with a non-empty index slug (`slug: 'foo'`) — it wouldn't match `slug === ''` at all, so the described collision can't actually occur. Deriving the URL from the empty-slug child would also just yield `base` again (empty slug = base), so it adds no robustness.

The one genuine (but speculative) gap is a *nested* self-contained sub-app: its non-empty index slug wouldn't be detected as self-contained and its screens would leak as demos. No such app exists today, and the new unit test would immediately catch that regression if one is ever added. Happy to add a one-line comment noting the "sits at the base index" assumption if you'd prefer — resolving for now.

PR #5578, opened 2026-06-30
feat(DST-1282): scroll Tabs row horizontally when it overflows

sarahgm:
# Description

When more tabs are rendered than fit the available width, `Tabs.List` now scrolls horizontally instead of wrapping onto multiple lines or pushing the page wide. Tabs keep their natural width (`shrink-0`), snap gently into place (`proximity`), and the adjacent tab stays peeking past the edge so the row reads as scrollable. A vertical mouse wheel scrolls the row horizontally for pointer users without a trackpad, without hijacking page scroll or browser zoom (`ctrl`/`⌘`+wheel is left to the browser). On browsers that support scroll-driven animations the overflowing edges fade out (`ui-scroll-mask-x`); elsewhere (currently Firefox) it falls back to a plain scroll container with a native scrollbar. When all tabs fit, nothing changes visually.

The change lives in a new optional `tabsListScroll` theme slot and an `m.div` `layoutScroll` wrapper so the sliding selection indicator stays correct while react-aria scrolls an off-screen tab into view. No public API change.

Closes DST-1282

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Open Storybook → `Components/Tabs/Mobile` (extra-small viewport) — the tab row overflows and scrolls horizontally instead of wrapping.
2. Click an off-screen tab (e.g. "Audit log"); it scrolls into view and the indicator stays aligned.
3. With a mouse (no trackpad), hover the row and scroll the wheel vertically — the row scrolls horizontally.
4. Hold `ctrl`/`⌘` and wheel — the browser zooms (scroll is not hijacked).
5. In Chrome/Safari the overflowing edges fade; in Firefox a native scrollbar appears instead.

## Breaking Changes

No

## Checklist


themes/theme-rui/src/components/Tabs.styles.ts
```diff
@@ -2,15 +2,31 @@ import { ThemeComponent, cva } from '@marigold/system';
...
   container: cva({ base: 'flex flex-col gap-2' }),
+  // Scroll container for the tab row: scrolls horizontally with an edge-fade
+  // affordance (`ui-scroll-mask-x`) when the tabs overflow, and is visually
+  // inert when they fit. `py-1 -my-1` reserves vertical room for the tab focus
+  // rings and the indicator that `overflow-x`'s implied `overflow-y` would
+  // otherwise clip; `scroll-px-6` keeps the adjacent tab peeking when a tab is
+  // scrolled into view; `proximity` snap avoids fighting momentum and
+  // react-aria's auto-scroll.
+  tabsListScroll: cva({
+    base: 'ui-scroll-mask-x -my-1 snap-proximity scroll-px-6 overscroll-x-contain py-1 motion-safe:scroll-smooth',
```
sebald: **`snap-proximity` is a no-op here without an axis utility.** In Tailwind v4, `snap-proximity` only sets `--tw-scroll-snap-strictness: proximity` — it does **not** emit `scroll-snap-type`. The axis utility (`snap-x`) is what actually turns snapping on (`scroll-snap-type: x var(--tw-scroll-snap-strictness)`). As written, `scroll-snap-type` stays unset, so `snap-start` on each tab and the `scroll-px-6` peek offset are both inert — the "snaps gently into place (`proximity`)" behavior described in the PR/changeset never happens.

```diff
-    base: 'ui-scroll-mask-x -my-1 snap-proximity scroll-px-6 overscroll-x-contain py-1 motion-safe:scroll-smooth',
+    base: 'ui-scroll-mask-x -my-1 snap-x snap-proximity scroll-px-6 overscroll-x-contain py-1 motion-safe:scroll-smooth',
```

Worth adding a quick assertion or visual check so this can't silently regress.

packages/components/src/Tabs/TabList.tsx
```diff
@@ -18,12 +18,61 @@ export type TabListProps = Omit<
...
+    if (!el) {
+      return;
+    }
+    const onWheel = (event: WheelEvent) => {
+      // Let the browser zoom: ctrl/⌘+wheel (and trackpad pinch, which the
+      // browser reports as ctrl+wheel) must not be hijacked into scroll.
+      if (event.ctrlKey || event.metaKey) {
+        return;
+      }
+      if (el.scrollWidth <= el.clientWidth) {
```
sebald, line 68: **Scroll trap at the ends.** The overflow guard only bails when the row isn't overflowing at all. Once it *is* overflowing and the user has scrolled to the far end, `scrollWidth > clientWidth` is still true, so `preventDefault()` fires but `scrollBy` has nowhere left to go — the wheel event is swallowed and the **page can't scroll past the tab bar**. Same happens at the left edge scrolling up.

Consider only hijacking the wheel when there's actually remaining room in the wheel's direction, e.g.:

```ts
const max = el.scrollWidth - el.clientWidth;
const atStart = el.scrollLeft <= 0;
const atEnd = el.scrollLeft >= max;
if ((delta < 0 && atStart) || (delta > 0 && atEnd)) return; // let the page scroll
```

(compute `delta` before the guard, after the deltaMode normalization).

packages/components/src/Tabs/TabList.tsx
```diff
@@ -18,12 +18,61 @@ export type TabListProps = Omit<
...
-        </TabList>
+        {/*
+         * `layoutScroll` keeps the motion indicator (`layoutId`) correct when an
+         * off-screen tab is scrolled into view, otherwise it projects against a
+         * stale scroll offset.
+         */}
+        <m.div
+          ref={scrollRef}
+          layoutScroll
+          className={classNames.tabsListScroll}
```
sebald, line 108: **The `tabsListScroll` ↔ `tabsList` pairing isn't enforceable.** `tabsListScroll` is typed as an *optional* slot (`Partial<Record<…>>` in `theme.ts:277`), but `tabsList`'s `w-max` only behaves if this slot supplies the `overflow-x`. A theme that sets `w-max` on `tabsList` but omits `tabsListScroll` makes `useClassNames` return no `tabsListScroll` key → `className={undefined}` here → no `overflow-x` → the row stretches the page wide instead of scrolling, with no type error or runtime warning.

theme-rui ships both so it's fine today, but for external theme authors this is a footgun. Options: make the slot required (so TS catches the omission), or move the `w-max`/overflow ownership so they can't drift apart. At minimum, the JSDoc "the two slots are a pair" is doing load-bearing work the type system should be doing.

packages/components/src/Tabs/TabList.tsx
```diff
@@ -18,12 +18,61 @@ export type TabListProps = Omit<
...
   const layoutId = useId();
+  const scrollRef = useRef<HTMLDivElement>(null);
+
+  // Translate a vertical mouse wheel into horizontal scroll so pointer users
+  // without a trackpad can reach overflowing tabs. We only hijack the wheel
+  // when the row actually overflows and the gesture is primarily vertical, so
+  // a trackpad's native horizontal scroll and normal page scrolling are left
+  // untouched. Attached natively (not via React's passive onWheel) so
+  // preventDefault works.
+  useEffect(() => {
```
sebald: **This whole effect can be `useEvent` from `@react-aria/utils`.** It's already a dependency we use elsewhere (`SidebarLink`, `TagGroup`, `SelectList`), and it does exactly this — native `addEventListener`/`removeEventListener` with cleanup, plus it keeps the handler current via `useEffectEvent` (no stale-closure risk), and it accepts the same `options`:

```ts
import { useEvent } from '@react-aria/utils';

const WHEEL_OPTS = { passive: false } as const; // module constant — keep options identity stable

// inside the component:
useEvent(scrollRef, 'wheel', event => {
  if (event.ctrlKey) return;
  // …existing guard + scrollBy…
}, WHEEL_OPTS);
```

Drops the ref-null check, the listener boilerplate, and the cleanup return. Hoisting `WHEEL_OPTS` to a module constant matters — `useEvent` lists `options` in its effect deps, so an inline object would re-subscribe every render.

packages/components/src/Tabs/TabList.tsx
```diff
@@ -18,12 +18,61 @@ export type TabListProps = Omit<
...
+         * `layoutScroll` keeps the motion indicator (`layoutId`) correct when an
+         * off-screen tab is scrolled into view, otherwise it projects against a
+         * stale scroll offset.
+         */}
+        <m.div
+          ref={scrollRef}
+          layoutScroll
+          className={classNames.tabsListScroll}
+        >
+          <TabList {...props} className={classNames.tabsList}>
```
sebald: **Minor / theme-contract:** this used to be `cn('flex', classNames.tabsList)`, so the component guaranteed `flex` regardless of theme. Now it's the theme's responsibility. theme-rui adds `flex` to its `tabsList` base so we're fine in-repo, but any external theme that relied on the injected `flex` will silently render tabs stacked vertically (and break the indicator) after upgrading. If that's an intentional contract change it's worth a line in the changeset; otherwise keeping `cn('flex', classNames.tabsList)` preserves the old guarantee.

packages/components/src/Tabs/TabList.tsx
```diff
@@ -18,12 +18,61 @@ export type TabListProps = Omit<
...
+  // preventDefault works.
+  useEffect(() => {
+    const el = scrollRef.current;
+    if (!el) {
+      return;
+    }
+    const onWheel = (event: WheelEvent) => {
+      // Let the browser zoom: ctrl/⌘+wheel (and trackpad pinch, which the
+      // browser reports as ctrl+wheel) must not be hijacked into scroll.
+      if (event.ctrlKey || event.metaKey) {
```
sebald: **Nuance:** the `metaKey` half of this guard doesn't match the stated intent. Browser zoom (and trackpad pinch, which is reported as `ctrl`+wheel) is `ctrlKey` only — no major browser uses `⌘`+wheel for zoom. So guarding on `metaKey` doesn't protect any zoom gesture; it just silently disables wheel-scroll for anyone holding Cmd (some remapped mice / scroll accelerators). Dropping `|| event.metaKey` matches the comment's reasoning. Harmless either way, just noting the `⌘` comment is slightly off.

packages/components/src/Tabs/Tabs.test.tsx
```diff
@@ -96,3 +96,80 @@ test('supports render prop children on Tabs.Item', async () => {
...
+  const setup = ({
+    scrollWidth = 1000,
+    clientWidth = 200,
+  }: { scrollWidth?: number; clientWidth?: number } = {}) => {
+    render(<Basic.Component />);
+
+    // The scroll container is the wrapper around the tablist; it has no role of
+    // its own, so we reach it from the tablist.
+    // eslint-disable-next-line testing-library/no-node-access
+    const el = screen.getAllByRole('tablist')[0].parentElement as HTMLElement;
```
sebald: **Test-fragility nit (low):** reaching the scroll container via `tablist.parentElement` couples the test to the current DOM nesting. If `MotionFeatures`/`LazyMotion` ever renders an intermediate wrapper, `parentElement` would be the wrong element, `scrollBy` would be spied on the wrong node, and all five wheel tests would keep passing while testing nothing. If switching to `useEvent` (see `TabList.tsx`), the listener moves to `scrollRef` either way. Not blocking — the comment already flags the assumption — but a `data-testid` on the scroll wrapper (or asserting it's the element with the listener) would make this robust.

packages/components/src/Tabs/Tabs.stories.tsx
```diff
@@ -172,3 +172,68 @@ export const WithRenderProps = meta.story({
...
+    await step('Act', async () => {
+      await userEvent.click(lastTab!);
+    });
+
+    await step('Assert', async () => {
+      await expect(
+        canvas.getAllByText('Audit log settings panel.')[0]
+      ).toBeVisible();
+      const indicator = await waitFor(
+        () => canvas.getAllByTestId('tab-indicator')[0]
```
sebald, line 234: **Nit (low):** `getAllByTestId('tab-indicator')` goes against the repo's "avoid `getByTestId`, prefer accessible queries" guideline. I'll grant the indicator is purely decorative (no role/label) so there's no accessible handle, and the test-id is a pre-existing pattern — so this is borderline acceptable. Just flagging: if the indicator is ever renamed/removed, the `waitFor` resolves on an empty array silently rather than failing. A behavioral assertion (e.g. the selected tab carries the expected indicator style/class) would be less brittle.

sebald (commented):
Nice, focused change — the `ui-scroll-mask-x` / `py-1 -my-1` / `shrink-0` reuse from SegmentedControl is exactly right, and the `layoutScroll` note is a good catch. High-effort review below; inline comments are mostly suggestions, with two I'd call real bugs (`snap-x` missing → snapping never engages; wheel scroll-trap at the row ends).

### Shared logic with SegmentedControl (possible follow-up, not blocking)

SegmentedControl already solved the same horizontal-overflow-scroll problem, and this PR correctly reuses its visual building blocks. Two opportunities to converge further, worth a separate ticket rather than this PR:

- **Reduced-motion handling diverges.** SegmentedControl gates smooth scroll in JS (`window.matchMedia('(prefers-reduced-motion: reduce)')` + `requestAnimationFrame`); this PR does it declaratively with `motion-safe:scroll-smooth`. The CSS approach here is cleaner and I'd like SegmentedControl to adopt it. I'm opening a separate PR for that.
- **The wheel→horizontal translation is new and Tabs-only.** SegmentedControl has the same overflow-scroll but no wheel support. This is the natural candidate for a shared hook (e.g. `useWheelScrollX`) so both components get the behavior — and so the scroll-trap / `deltaMode` fixes live in one place instead of being duplicated later. Flagging as a future extraction; no need to do it here.

(One behavioral nuance on the wheel handler: `scrollBy({ behavior: 'auto' })` defers to the element's CSS `scroll-behavior`, which is `motion-safe:scroll-smooth` — so wheel scrolling animates rather than tracking the wheel 1:1. If 1:1 tracking is intended, `behavior: 'instant'` would be needed. Debatable; leaving it to your judgement.)

themes/theme-rui/src/components/Tabs.styles.ts
sarahgm: Fixed in 607248d. Added `snap-x` to the `tabsListScroll` base so `scroll-snap-type` is actually emitted. Verified in the browser that it now computes to `x` and snapping engages.

packages/components/src/Tabs/TabList.tsx
sarahgm: Fixed in 607248d. The normalized `delta` is now computed before the guard, and the handler returns when there is no room left in the wheel direction (`scrollLeft <= 0` going left, `>= max` going right) so the page scrolls past the tab bar. Added edge-of-row unit tests for both ends.

packages/components/src/Tabs/TabList.tsx
sarahgm: Fixed in 607248d. Made `tabsListScroll` a required slot in `theme.ts` (dropped the `Partial`). TS now flags a theme that ships `tabsList` without the scroll container. It already caught an in-repo fixture, which I updated.

packages/components/src/Tabs/TabList.tsx
sarahgm: Done in 607248d. Switched to `useEvent` from `@react-aria/utils` with a hoisted `WHEEL_OPTS` module constant so the options identity stays stable across renders. Dropped the ref-null boilerplate and the cleanup return.

packages/components/src/Tabs/TabList.tsx
sarahgm: Fixed in 607248d. Restored `cn('flex', classNames.tabsList)` so the `flex` guarantee no longer depends on the theme.

packages/components/src/Tabs/TabList.tsx
sarahgm: Fixed in 607248d. Dropped the `metaKey` half of the guard and corrected the comment. Zoom is `ctrlKey`-only.

packages/components/src/Tabs/Tabs.test.tsx
sarahgm: Fixed in 607248d. The scroll wrapper now carries a stable `data-testid="tabs-list-scroll"` and the test reaches it via `getAllByTestId` instead of `parentElement`, so it no longer couples to the motion wrapper's DOM nesting.

packages/components/src/Tabs/Tabs.stories.tsx
sarahgm: Leaving this as-is for now. The indicator is decorative with no accessible handle and the test-id is a pre-existing pattern, so I kept the existing query rather than widen scope in this PR. Noted as a future cleanup.

sarahgm:
Thanks for the high-effort review. All inline threads addressed in 607248d and resolved. Two notes on the summary points:

**Wheel `behavior` (the nuance you flagged).** This turned out to matter more than expected once snapping was actually on. With `snap-x` engaged, `scrollBy({ behavior: 'auto' })` deferred to the container's `scroll-behavior: smooth`, so each notch started a fresh animation that snapping kept interrupting. Rapid notches did not compose and the row scrolled erratically. I verified this in the browser: the same 5 wheel notches landed at `[64,141,247,271,365]` on one run and `[375,451,460,460,460]` on the next. Switching to `behavior: 'instant'` makes the wheel track 1:1 and behave deterministically, while `scroll-behavior: smooth` still drives react-aria's programmatic scroll-into-view. So your suggestion is now the shipped behavior.

**Follow-ups (both out of scope here, agreed).**
- Reduced-motion CSS convergence for SegmentedControl: already covered by your #5588.
- Shared `useWheelScrollX` hook: worth doing once #5588 lands so the extraction has both call sites settled. Happy to file a DST ticket for it if you'd like.

sarahgm:
Filed the shared-hook extraction as [DST-1587](https://reservix.atlassian.net/browse/DST-1587) (🧹 Extract shared `useWheelScrollX` hook for Tabs and SegmentedControl). It's linked as blocked by DST-1585 (#5588) so it gets picked up once both this PR and the reduced-motion convergence land in `beta-release`.

[DST-1587]: https://reservix.atlassian.net/browse/DST-1587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

packages/system/src/types/theme.ts
```diff
@@ -266,8 +266,19 @@ export type Theme = {
...
+    // `overflow-x`. It is required and paired with `tabsList`: when `tabsList`
+    // is wider than its container (e.g. `w-max`) the row scrolls, otherwise it
+    // overflows the page. Making the slot required means a theme cannot ship
+    // `tabsList` without the scroll container that makes it behave. See
+    // theme-rui's Tabs.styles.ts.
     Tabs?: Record<
-      'container' | 'tabsList' | 'tabpanel' | 'tab' | 'tabIndicator',
+      | 'container'
+      | 'tabsList'
+      | 'tabsListScroll'
```
aromko, line 278: 🟠 **Required theme slot — confirm semver intent**

Adding `tabsListScroll` as a **required** key on the exported `Tabs` `Record` is a compile-time **breaking change** for any consumer maintaining a custom theme that defines a `Tabs` block — it won't type-check until they add this slot. The changeset labels `@marigold/system` as **`minor`**.

Since this targets `beta-release` (likely a batched pre-release train), that may well be intended, and the "required on purpose" rationale in the comment above is sound. Just flagging to confirm the semver label matches the release train — otherwise consider making the slot optional to stay backward-compatible within a minor.

sarahgm: Good catch — the `minor` label was inaccurate for a required-slot change. The intent is to keep the slot **required** (it is what makes `tabsList`’s `w-max` safe, and it already caught an in-repo fixture), so the right fix is the label, not the API.

Bumped this changeset’s `@marigold/system` to `major` and added an explicit **Breaking change** note describing the new required `tabsListScroll` slot — matching how the other breaking changes in this beta train are labeled (`DST-1324`, `DST-1549`, `DST-1559`).

No effect on the released version either way: `@marigold/system` is already `major` in this train (`align-breakpoints-rem`, `little-mirrors-burn`, `tiny-baths-accept`), and changesets takes the highest bump. This just makes the changelog line item semantically correct.

packages/components/src/Tabs/TabList.tsx
```diff
@@ -18,12 +30,75 @@ export type TabListProps = Omit<
...
-        </TabList>
+        {/*
+         * `layoutScroll` keeps the motion indicator (`layoutId`) correct when an
+         * off-screen tab is scrolled into view, otherwise it projects against a
+         * stale scroll offset.
+         */}
+        <m.div
+          ref={scrollRef}
+          layoutScroll
+          className={classNames.tabsListScroll}
```
sebald, line 108: **Suggestion (non-blocking) — the required-slot guarantee is compile-time only.** Making `tabsListScroll` required in `theme.ts` + the major bump is the right call, but the sibling below still keeps a literal fallback (`cn('flex', classNames.tabsList)`) while this wrapper is bare `className={classNames.tabsListScroll}`. `useClassNames` only checks that `Tabs` exists, not each slot key, so a consumer who reaches runtime without the slot — a JS theme, an `as Theme` cast, or a theme not yet migrated — gets `className={undefined}` here → no `overflow-x`/`snap-x`, and the row silently blows out the page instead of scrolling. No error, no warning.

There's a real design tension, so I'll leave the choice to you:
- **Accept as-is** — the required type + major bump + migration note is a deliberate, defensible contract; un-migrated JS themes are on the consumer.
- **Defense-in-depth fallback** — `className={cn('overflow-x-auto', classNames.tabsListScroll)}`, mirroring the `tabsList` `flex` fallback. Downside: it masks a missing slot and softens the "you must add it" contract.
- **Dev-only warning** when the slot is absent — keeps the strict contract but fails loudly instead of silently.

I'd lean toward the dev-warning or accepting as-is; flagging mainly so the silent-failure mode is a conscious decision.

sebald (commented):
Round 2 — the fixes look great. snap-x, the both-ends scroll-trap guard, the required `tabsListScroll` slot with the major bump + migration note, `useEvent`, and `behavior: 'instant'` for the wheel are all clean, and the added scroll-trap tests are a nice touch. One non-blocking suggestion inline about the runtime side of the now-required slot; everything else I looked at (deltaMode edge, subpixel end-guard, RTL) is minor and left off. Nice work.

packages/components/src/Tabs/TabList.tsx
sarahgm: Went with the **dev-only warning** in 3ac9a6f92. `TabList` now emits a `console.warn` (dev builds only, matching the `Card` pattern) when `classNames.tabsListScroll` is absent, calling out that the row will overflow the page and pointing to the required slot. Keeps the strict required-slot contract — I deliberately avoided the `overflow-x-auto` fallback since it would mask a genuinely missing slot — but turns the silent page-blowout into a loud, actionable message. Left it untested to match the repo convention for the sibling warnings (`Card`, `SelectList`, `ToggleButton`) and the no-theme-fixtures guideline.

sebald (approved):

PR #5581, opened 2026-06-30
fix(DST-1576): resolve ProgressCircle named size tokens to numeric SVG dimension

sebald:
# Description

`ProgressCircle` accepts named `size` tokens (`default` | `large` | `fit`), but `ProgressCircleSvg` forwarded the raw token to the underlying `<SVG>`, which renders `width`/`height` as `` `${size}px` ``. For a named token this produced invalid attributes like `width="defaultpx"` / `height="defaultpx"` and emitted runtime console errors (visible on the **Filter** example page, where `FetchingIndicator` uses `<ProgressCircle size="default">`).

`ProgressCircleSvg` now resolves `size` to a numeric intrinsic dimension (falling back to `24` for non-numeric tokens) for the SVG `width`/`height` attributes and the stroke-width math. The **rendered** size is unchanged — it comes from the theme `size-*` classes via `useClassNames`, not the attribute.

Closes DST-1576

## Test Instructions

1. Open the Filter example (`/examples/filter`) and trigger a refetch (search/sort/paginate) — the "Updating…" spinner appears with **no** `<svg> attribute width/height: Expected length, "defaultpx"` console errors.
2. `npx vitest run --project=unit-tests ProgressCircle.test` — includes a new regression test asserting named tokens resolve to a numeric `width`/`height`.

## Breaking Changes

No

## Checklist

### DIFF
--- a/.changeset/progresscircle-named-size-svg.md
+++ b/.changeset/progresscircle-named-size-svg.md
@@ -1,0 +1,13 @@
+---
+'@marigold/components': patch
+---
+
+fix(ProgressCircle): resolve named `size` tokens to a numeric SVG dimension
+
+A named `size` token (`default` | `large` | `fit`) was forwarded unchanged to the
+underlying `<SVG>` element, which renders `width`/`height` as `` `${size}px` `` —
+producing invalid attribute values like `width="defaultpx"` and emitting console
+errors. The rendered size already comes from the theme size classes, so the SVG now
+resolves `size` to a numeric intrinsic dimension (falling back to `24` for non-numeric
+tokens) for its `width`/`height` attributes and stroke-width math. Visual output is
+unchanged.
--- a/.changeset/progresscircle-named-size-svg.md
+++ b/.changeset/progresscircle-named-size-svg.md
@@ -7,7 +7,12 @@ fix(ProgressCircle): resolve named `size` tokens to a numeric SVG dimension
 A named `size` token (`default` | `large` | `fit`) was forwarded unchanged to the
 underlying `<SVG>` element, which renders `width`/`height` as `` `${size}px` `` —
 producing invalid attribute values like `width="defaultpx"` and emitting console
-errors. The rendered size already comes from the theme size classes, so the SVG now
-resolves `size` to a numeric intrinsic dimension (falling back to `24` for non-numeric
-tokens) for its `width`/`height` attributes and stroke-width math. Visual output is
-unchanged.
+errors. The SVG now resolves named tokens to the pixel dimension of their theme
+`size-*` class (`default` → 80, `large` → 144) for its `width`/`height` attributes
+and stroke-width math, so the stroke stays proportionate and the rendered output is
+unchanged. `fit` is content-sized and has no intrinsic dimension, so it falls back to
+the `<SVG>` default of `24`.
+
+The stroke-width comparison also switched from a string compare (`size <= '24'`) to a
+numeric one, which additionally fixes multi-digit sizes that were previously
+mis-classified (e.g. `size="100"` computed a `strokeWidth` of `2` instead of `4`).
--- a/docs/content/components/content/loader/loader-section.demo.tsx
+++ b/docs/content/components/content/loader/loader-section.demo.tsx
@@ -1,4 +1,4 @@
-import type { Venue } from '@/lib/data/venues';
+import type { VenueQueryResult } from '@/lib/data/venues-query';
 import {
   QueryClient,
   QueryClientProvider,
@@ -18,7 +18,7 @@ import {
 } from '@marigold/components';
 
 const Venues = () => {
-  const { data, status, isFetching } = useQuery<Venue[]>({
+  const { data, status, isFetching } = useQuery<VenueQueryResult>({
     queryKey: ['venues'],
     queryFn: async () => {
       const response = await fetch(`/api/venues?delay=1500&q=`);
@@ -52,7 +52,7 @@ const Venues = () => {
   return (
     <Scrollable height="300px">
       <Stack space={2}>
-        {data.map(v => (
+        {data.items.map(v => (
           <Card key={v.id}>
             <Card.Body>
               <Aside sideWidth="160px" space={8}>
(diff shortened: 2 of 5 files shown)


sarahgm (commented):
## Automated Code Review

### Overview
| Field | Value |
|-------|-------|
| Branch | `fix/DST-1576-progresscircle-svg-size` → `beta-release` |
| Files Changed | 4 (+60 / -9) |
| CI | All green (Typecheck, Lint, Unit, Storybook, Coverage, Size) |
| Visual Regression | Not required (base is `beta-release`) — but see Warning #1 |

### What the change does (verified)
- `SVG.tsx` renders `width={`${size}px`}` with a default of `24`. Passing a named token like `'default'` produced `width="defaultpx"` → the runtime error. The diagnosis is correct.
- The fix resolves `size` to a number via `Number.parseFloat(String(size))`, falling back to `24` (matching `SVG`'s own default) for non-numeric tokens, and switches the stroke-width comparisons from **string** (`size <= '24'`) to **numeric** (`svgSize <= 24`).

### Warnings (Should Fix)

**1. "Visual output is unchanged" is not accurate — the Loader's stroke thickness halves for named tokens.**

Box dimensions are driven by the theme `size-*` classes, so those are unchanged. But **stroke-width and radius are computed inside this component**, and they change in the rendered DOM. The `Loader.test.tsx` snapshot in this PR shows it:

```diff
- stroke-width="4"   r="calc(50% - 2px)"
+ stroke-width="2"   r="calc(50% - 1px)"
```

Why: the theme maps named tokens to **large** rendered sizes — `default → size-20` (80px), `large → size-36` (144px). The old string-comparison bug *coincidentally* produced `strokeWidth = 4` for every named token (`'default' >= '32'` is true because `'d' > '3'`), which happened to suit those large circles. Collapsing all named tokens to `svgSize = 24` now yields `strokeWidth = 2` for **all** of them, so the default (80px) and large (144px) spinners render with a visibly thinner stroke. Since `SVG` has no `viewBox`, user units map ~1:1 to CSS px, so this is a real on-screen change.

Recommendation: resolve named tokens to their intended numeric dimension (e.g. `default → 80`, `large → 144`) so the stroke scales as the `>= 32 → 4` branch intends — or, if a uniform thin stroke is genuinely desired, correct the PR description and changeset to state that the Loader's stroke thickness changes. This delta would be caught by VRT when forward-ported into `main`; worth a Chromatic run before then.

**2. Tests enshrine the thinned-stroke behavior.** `ProgressCircle.test.tsx` asserts `width/height = '24px'` and `stroke-width = '2'` for named tokens. Mechanically fine, but these lock in the behavior questioned in #1 — if the team decides named tokens should scale the stroke, these expectations need to change alongside.

### Suggestions (Nice to Have)
- The switch to numeric comparison is a quiet bonus fix: the old string compare mis-handled multi-digit sizes too (e.g. `'100' <= '24'` was `true` → strokeWidth 2; now correctly `4`). Worth a one-line mention in the changeset.

### Positives
- Console-error bug is genuinely fixed; fallback of `24` matches `SVG`'s own default.
- Changeset present, scoped `patch`, accurate.
- Tests use accessible queries (`getByRole('progressbar')`), `test.each` is clean, no `any`, `import type` respected.

### Recommendation
**Comment** (not a hard block). The fix correctly resolves the runtime errors and is well-tested, but it changes the Loader's stroke thickness for named tokens — which the description says it doesn't. Before merge, please either (a) resolve named tokens to their real dimensions so the stroke stays proportionate, or (b) confirm the thinner stroke is intended and update the description/changeset accordingly.

---

sebald:
Thanks for the thorough review @sarahgm — all three points addressed in d7fe14ea3.

**Warning #1 (stroke thinned for named tokens) — fixed.** You were right that collapsing every named token to `svgSize = 24` dropped the stroke from 4→2 on the large presets, so "visual output unchanged" wasn't accurate. Named tokens now resolve to the pixel dimension of their theme `size-*` class instead of a flat `24`:

```ts
const NAMED_SIZE_DIMENSIONS: Record<string, number> = {
  default: 80, // size-20
  large: 144,  // size-36
};
const numericSize = Number.parseFloat(String(size));
const svgSize = Number.isNaN(numericSize)
  ? (NAMED_SIZE_DIMENSIONS[size as string] ?? 24)
  : numericSize;
```

`default` (80) and `large` (144) hit the `>= 32 → strokeWidth 4` branch, so their rendered stroke/radius match pre-bug output exactly. `fit` is content-sized with no intrinsic dimension, so it keeps the `24` fallback (`strokeWidth 2`).

**Warning #2 (tests locked in the thin stroke) — updated.** `ProgressCircle.test.tsx` now asserts per-token: `default`/`large` → `80px`/`144px` + `stroke-width 4`, `fit` → `24px` + `stroke-width 2`. The `Loader.test.tsx` snapshot is back to `stroke-width 4` / `r="calc(50% - 2px)"`.

**Suggestion (multi-digit string compare) — mentioned.** Changeset now notes the numeric-comparison switch also fixes `size="100"` (was `strokeWidth 2`, now correctly `4`), and no longer claims a flat `24` fallback.

The PR description's "visual output unchanged" line is now genuinely true for `default`/`large`. This is still a visual-adjacent change, so I'll run Chromatic before it forward-ports to `main`.

sebald:
@aromko good catch on the thin ring — that was actually a side effect of an earlier state of this very PR, and it's already fixed in the latest commit (d7fe14ea3).

**What happened:** the first iteration resolved every named `size` token to a flat `24px`, which pushed the stroke onto the `<= 24 → strokeWidth 2` branch. The `CircleLoader` story renders at `size="default"` (80px, `size-20`), so its ring came out at half the intended thickness — most visible on the `inverted` variant used by the `section`/`fullscreen` modes (the dark-background ones you saw).

**Fix:** named tokens now resolve to the pixel dimension of their theme `size-*` class (`default → 80`, `large → 144`), so they land back on the `>= 32 → strokeWidth 4` branch. Once the Storybook preview redeploys from the latest commit the ring is back to the thicker `4px` stroke it had before this PR.

**On accessibility:** no hard failure here. The loader exposes `role="progressbar"` + `aria-label="Loading..."`, so it's announced to assistive tech regardless of stroke. WCAG has no stroke-thickness rule; the relevant criterion (1.4.11 Non-text Contrast) is about the ring *color* vs its background, which this PR doesn't change — so "too thin" was a perceptibility concern, now improved by the restored 4px stroke. If we want to be rigorous about the `inverted` ring's contrast (`stroke-secondary` on the overlay), I'm happy to run that as a separate check, since it's independent of this fix.

sarahgm (approved):

PR #5582, opened 2026-06-30
feat(DST-1577): surface failed-invoice recovery path on Billing example

sebald:
# Description

On the **Billing** example, a `failed` invoice (INV-2026-002 — February 2026, €420.00) was the most important fact on the page yet was surfaced only as a small red `Badge` in row 5 of 6, with no recovery affordance and the recovery action ("Update payment method") sitting in a separate panel with no thread to the problem.

Now, when any invoice has failed:

- The page **leads with a `SectionMessage variant="error"`** (`role="alert"`; since the banner is present on initial load rather than inserted dynamically, screen readers announce it when the user reaches it) that names the failed period + amount and offers a **primary "Update payment method"** action.
- The **failed row** gets an in-context **"Retry payment"** button alongside the download icon.

Renders only when an invoice is `failed`. Built entirely with shipped Marigold components (`SectionMessage`, `Button`, `Inline`, `Stack`) — no custom CSS.

Closes DST-1577

## Screenshots / Preview

Open `/examples/billing`: a red "Payment failed" banner appears above the panels with an "Update payment method" CTA, and the February row shows a "Retry payment" button.

## Test Instructions

1. Run the docs app and open `/examples/billing`.
2. Confirm the red **Payment failed** `SectionMessage` appears above the three panels, naming "February 2026 payment of €420.00", with a primary **Update payment method** button.
3. Confirm the **INV-2026-002** (failed) row shows a **Retry payment** button next to the download icon; other rows do not.

## Breaking Changes

No

## Checklist


aromko (commented):
Automated review (Claude Code `review-pr`) — 1 item to verify + 2 suggestions, all non-blocking. Change fully satisfies DST-1577 and CI is green.

docs/app/(examples)/examples/billing/page.tsx
```diff
@@ -23,108 +24,137 @@ const statusVariant = {
...
+                    />
+                  </Table.Cell>
+                  <Table.Cell alignX="right">{invoice.amount}</Table.Cell>
+                  <Table.Cell>
+                    <Badge variant={statusVariant[invoice.status]}>
+                      {invoice.status}
+                    </Badge>
+                  </Table.Cell>
+                  <Table.Cell alignX="right">
+                    <Inline space="tight" alignY="center">
```
aromko: 🟠 **Actions column may shift from right → left**

`Table.Cell alignX="right"` right-aligns content via a `text-right` class on the cell-content wrapper (`TableCellContent.tsx:77`). A bare `<button>` (inline-block) honored that, but `Inline` is a `display:flex` block that fills the cell width and **ignores `text-align`**, laying its children out at `flex-start` (left). `Inline` exposes only `alignY` — no `alignX`/justify prop — so right alignment can't be restored on it. Net effect: the Download icon (and the `[Retry][Download]` pair on the failed row) likely moves to the **left** of the column on every row, versus the deliberately right-aligned `&nbsp;` header.

Visual Regression Tests don't run against `beta-release`, so this won't be caught automatically — please open `/examples/billing` and confirm the alignment is intended.

docs/app/(examples)/examples/billing/page.tsx
```diff
@@ -23,108 +24,137 @@ const statusVariant = {
...
+                      {invoice.status === 'failed' && (
+                        <Button size="small">Retry payment</Button>
```
aromko: 💡 Consider giving this a per-row accessible name, matching the Download button's per-invoice `aria-label` just below:

```suggestion
                        <Button
                          size="small"
                          aria-label={`Retry payment for ${invoice.period}`}
                        >
                          Retry payment
                        </Button>
```

Low priority — table row context already helps screen readers — but it keeps the two row actions consistent.

docs/app/(examples)/examples/billing/page.tsx
```diff
@@ -23,108 +24,137 @@ const statusVariant = {
...
+      <Page.Header>
+        <Title>Billing</Title>
+        <Description>
+          Manage your plan, payment method, and invoices.
+        </Description>
+        <Button variant="primary">Change plan</Button>
+      </Page.Header>
+
+      {failedInvoice && (
+        <SectionMessage variant="error">
```
aromko, line 41: 💡 `SectionMessage variant="error"` renders `role="alert"` (implicit `aria-live="assertive"`), which reliably announces content inserted *after* the region mounts. Here the banner is present on initial page load, so most screen readers won't auto-announce it — it's read when the user reaches it. Code is fine; just flagging that the description's "auto assertive announce" is optimistic for a statically-present banner.

docs/app/(examples)/examples/billing/page.tsx
sebald: Good catch on the flex/`text-align` interaction. One correction: `Inline` *does* now expose `alignX` (maps to `justify-*`), so I restored right-alignment directly on it with `alignX="right"` rather than needing a wrapper. Fixed in a588ebce.

docs/app/(examples)/examples/billing/page.tsx
sebald: Applied — the Retry button now has `aria-label={`Retry payment for ${invoice.period}`}` to match the Download button. Fixed in a588ebce.

docs/app/(examples)/examples/billing/page.tsx
sebald: Agreed — the code is fine, the "auto assertive announce" phrasing in the description was optimistic for a statically-present banner. I'll adjust the PR description to say it's announced when the user reaches it rather than auto-announced on load. No code change needed here.

aromko (approved):

PR #5584, opened 2026-06-30
fix(DST-1579): tidy Filter toolbar — hide empty applied-filters, right-size spinner

sebald:
# Description

Two small craft fixes on the **Filter** example toolbar:

- **`AppliedFilter` renders `null` when no filters are active.** Previously it always showed an "Applied Filters / None" block — a full-width row of dead space on every fresh visit (the default state), training users to ignore the region. It now appears only once filters are applied.
- **`FetchingIndicator` uses a small `ProgressCircle size="16"`** instead of `size="default"` (which maps to the ~80px `size-20` class) so the inline "Updating…" indicator is appropriately sized.

The fetching indicator stays `aria-hidden` on purpose: `VenuesTable` already exposes a `role="status"` live region announcing result counts, so the visual spinner doesn't need to double-announce.

Closes DST-1579

## Test Instructions

1. Open `/examples/filter` with no filters applied — there is **no** "Applied Filters / None" row; the table sits directly under the toolbar.
2. Apply a filter via the Filter drawer — the applied-filter chips now appear.
3. Trigger a refetch (search/sort/paginate) — the "Updating…" spinner is a small inline indicator.

## Breaking Changes

No

## Checklist

### DIFF
--- a/docs/app/(examples)/examples/filter/AppliedFilter.tsx
+++ b/docs/app/(examples)/examples/filter/AppliedFilter.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import type { ReactNode } from 'react';
-import { Tag, Text } from '@marigold/components';
+import { Tag } from '@marigold/components';
 import { NumericFormat } from '@marigold/system';
 import {
   type FilterKeys,
@@ -47,6 +47,13 @@ const getLabel = (filter: VenueFilter, key: FilterKeys): ReactNode => {
 
 export const AppliedFilter = () => {
   const { filter, removeFilter, activeFilterKeys } = useFilter();
+  const activeKeys = activeFilterKeys();
+
+  // Nothing to show at rest (the default, no-filter state) — render nothing
+  // rather than a permanent "None" row of dead space.
+  if (activeKeys.length === 0) {
+    return null;
+  }
 
   return (
     <Tag.Group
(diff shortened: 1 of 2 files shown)


sebald:
## ✅ Best-practice verification: applied-filters display

I ran a multi-source review (Baymard, NN/g, Shopify Polaris, IBM Carbon, Ant Design, W3C ARIA — 19 sources, adversarially verified) to check this PR's applied-filters change. **Verdict: sound and on best practice.**

- **Hiding the region when empty is correct.** Polaris renders its filter-pills bar conditionally on `appliedFilters` (nothing when empty); Carbon and Ant Design have no "None" placeholder either. No authoritative source recommends a persistent empty-state / "None" row — the exact thing this PR removes. (Attempts to find support for an always-visible empty placeholder were refuted across the board.)
- **The chips + per-tag remove + `removeAll` we keep is *the* recommended pattern.** Baymard (72% of benchmarked sites provide an applied-filters overview; omitting it is flagged as a defect) and Polaris both prescribe discrete, individually-removable chips plus a global "clear all".
- **No layout-shift objection.** The "reserve space above results to avoid CLS" argument did not hold up under verification.

Sources: [Baymard – How to design Applied Filters](https://baymard.com/blog/how-to-design-applied-filters) · [Polaris Filters](https://polaris.shopify.com/components/selection-and-input/filters) · [Carbon Filtering](https://carbondesignsystem.com/patterns/filtering/) · [NN/g Empty States](https://www.nngroup.com/articles/empty-state-interface-design/).

**A11y note (not a regression in this PR):** the research flags that conditionally mounting/unmounting a *live region* makes NVDA/JAWS miss announcements. This PR is clean on that front — `AppliedFilter` isn't a live region, and filter changes are already announced by the venues table's always-present `role="status"` result-count region. A separate, optional hardening of that count region is tracked as a ToDo on #5522.

aromko (approved):

PR #5585, opened 2026-06-30
fix(DST-1580): cap tag-pill columns in the Venues table

sebald:
# Description

On the **Filter** venues table, the Traits / Amenities / Parking columns rendered one `Badge` per value, so a rich row (e.g. Grand Avenue Ballroom: 5 amenities + 3 parking) stacked ~10 badges and ballooned to several lines — row height became data-correlated and hurt vertical scannability.

A small `BadgeList` helper now shows up to **3** badges per cell and collapses the remainder into a single **"+N"** badge — the same truncation idiom `AppliedFilter` already uses ("+N more"). Row heights stay consistent.

Closes DST-1580

## Test Instructions

1. Open `/examples/filter`.
2. Confirm cells with many values (e.g. Grand Avenue Ballroom — Amenities) show 3 badges + a "+N" badge rather than the full stacked list, and rows are a more uniform height.

## Breaking Changes

No

## Checklist


docs/app/(examples)/examples/filter/VenuesTable.tsx
```diff
@@ -88,6 +88,28 @@ export const VenuesTableSkeleton = ({ rows = 5 }: { rows?: number }) => (
...
+}: {
+  items: readonly string[];
+  max?: number;
+}) => {
+  const overflow = items.length - max;
+
+  return (
+    <Inline space="0.5">
+      {items.slice(0, max).map(item => (
+        <Badge key={item}>{item}</Badge>
```
aromko, line 108: 💡 **Keying by display label instead of a stable id.** The original code keyed by the stable option id (`key={a}` / `key={p}`) and rendered the label as children; `BadgeList` now receives already-mapped display strings and keys by the label. If two option ids ever resolved to the same label, React would hit duplicate keys. Traits/Amenities/Parking are almost certainly unique per row so the practical risk is ~zero — just flagging the subtle behavioral change. Fine to leave as-is for a docs example.

docs/app/(examples)/examples/filter/VenuesTable.tsx
```diff
@@ -88,6 +88,28 @@ export const VenuesTableSkeleton = ({ rows = 5 }: { rows?: number }) => (
...
+      ))}
+      {overflow > 0 && <Badge>{`+${overflow}`}</Badge>}
```
aromko: 💡 **"+N" badge is count-only for screen readers.** It announces just "+3" without indicating what's hidden. Revealing the overflow (popover/tooltip) is an explicit ticket non-goal, so this is acknowledged scope — but a lightweight `aria-label`/`title` like `"+3 more"` would improve SR clarity without pulling in the deferred popover work. Optional.

docs/app/(examples)/examples/filter/VenuesTable.tsx
```diff
@@ -88,6 +88,28 @@ export const VenuesTableSkeleton = ({ rows = 5 }: { rows?: number }) => (
...
+// so multi-value cells stay a consistent height instead of ballooning the row.
+// Mirrors the "+N more" truncation used by AppliedFilter.
+const BadgeList = ({
+  items,
+  max = 3,
+}: {
+  items: readonly string[];
+  max?: number;
+}) => {
+  const overflow = items.length - max;
```
aromko: 💡 **Off-by-one tradeoff at `items.length === max + 1`.** With `max=3` and exactly 4 items, the cell shows 3 badges + a "+1" badge — the "+1" occupies the space the 4th badge would have. This matches the `AppliedFilter` idiom and keeps the code simple, so it's a deliberate tradeoff; some implementations show the extra item when overflow is exactly 1. Purely optional.

aromko (commented):
Automated review (Claude Code `review-pr`): small, focused docs-only change that matches DST-1580 exactly. CI green, types check. No blocking issues — 3 optional nice-to-haves inline below.

aromko (commented):
@sebald Before I could see all badges. With this solution now it is maybe space saving but I have no chance to see whats behind the e.g. +2. I think when I do filtering I will wonder hm why this still appears because I can't see everything. For me this is only a half-made solution. And I think the +2 isn't accessible in this way or?

docs/app/(examples)/examples/filter/VenuesTable.tsx
sebald: Fixed in ac8ef5f2d. The overflow badge is now wrapped in a `<span title="…">` and paired with a `VisuallyHidden` sibling, so it announces e.g. "+2 more: Rooftop, Garden" instead of a bare "+2", and hovering reveals the same labels natively. (Went via a wrapper rather than `aria-label`/`title` on `<Badge>` directly because `BadgeProps` is a closed interface — no passthrough for those attrs — and widening a published component for a docs example felt disproportionate.)

docs/app/(examples)/examples/filter/VenuesTable.tsx
sebald: Leaving as-is: matches the `AppliedFilter` "+N" idiom and keeps `BadgeList` a single, obvious rule (`slice(0, max)` + remainder). Showing the extra item only when overflow is exactly 1 would add a branch to a docs helper for a marginal gain. Deliberate tradeoff, noted.

docs/app/(examples)/examples/filter/VenuesTable.tsx
sebald: Leaving as-is: Traits/Amenities/Parking labels are unique per row, so duplicate keys cannot occur with this data, and it stays a docs example. Threading `{id, label}` pairs through `BadgeList` would make it collision-proof but adds shape for zero practical benefit here. Flag acknowledged.

sebald:
@aromko you're right — hiding the values with no way back was the real gap, thanks for pushing on it. I did **not** want to pull in the full popover reveal (that's an explicit DST-1580 non-goal), but the two concerns you raised — "I can't see what's behind the +2" and "the +2 isn't accessible" — are cheap to fix without it:

**a) Hover reveal.** The `+N` badge is now wrapped in `<span title="Rooftop, Garden, …">`, so hovering it surfaces the collapsed labels natively — no popover, no extra state.

**b) Accessibility.** A `VisuallyHidden` sibling gives it a real accessible name: it announces **"+2 more: Rooftop, Garden"** instead of a meaningless "+2".

So the collapsed values are now reachable by both mouse and screen-reader users; only the *rich* interactive reveal (click-to-expand popover) stays deferred to a follow-up, per the ticket scope. Does that address the "half-made" concern, or do you feel the click-to-expand reveal needs to land in this PR rather than a follow-up? Pushed as `ac8ef5f2d`.

aromko (approved):

PR #5589, opened 2026-06-30
style(DST-1586): quieter sidebar nav hierarchy, AA labels & aligned back action

sebald:
**Jira:** [DST-1586](https://reservix.atlassian.net/browse/DST-1586)

## Summary

Reworks the example app-shell sidebar and top navigation into a modern, minimalistic shell that fits the surrounding page. Two things happened here. First, the sidebar's hierarchy was re-grounded on semantic tokens and brought to WCAG AA. Second, a design review turned that into a broader "seamless shell" where regions separate on whitespace, with exactly one structural line (the sidebar divider) and scroll-revealed seams everywhere else.

Filed as unplanned (DST-1586). It surfaced while finishing the DST-1565 ui-surface work.

## What changed

**Sidebar hierarchy, token-driven and WCAG AA**

- Active row uses `text-foreground` on a `bg-selected` inset pill. Idle rows use `text-secondary`, and hover previews the pill in `bg-hover`. No raw `charcoal-*`.
- Section labels moved off sub-AA `charcoal-500` (3.3:1 ✗) to `text-secondary` (4.97:1 ✓). The tier reads through treatment (uppercase, smaller, heavier, tracked).
- Denser rows (fixed 30px height, tighter inset). The drill-in back action shares the nav-row pill geometry and weight.

**Seamless app shell**

- Exactly one structural line, the sidebar divider, at full-strength `border-surface-border`. It bounds the nav column so branch-row chevrons anchor, and it stays perceivable at low vision. Every other shell line is gone, including the sidebar header and footer hairlines and the `TopNavigation` bottom border.
- `TopNavigation` reveals a bottom hairline only once page content scrolls under it (`ui-scroll-edge`, a scroll-driven progressive enhancement).
- Scrolling sidebar seams: the sticky header and footer reveal a hairline driven by the nav's scroll position. Because the nav is a grid sibling between them, this uses a named scroll timeline (`--sidebar-nav`) hoisted via `timeline-scope`, with an always-on hairline `@supports not` fallback. Utilities are `ui-sidebar-seam-*`, marked non-reusable.

**Breadcrumbs**

The separator chevron now scales with the size variant (`small`, `default`, `large`) via the theme instead of a hardcoded `size={16}`. The current page reads a tier up (medium weight, `foreground`).

**Sidebar toggle**

The icon animation settles on `ease-out-quint` with no overshoot, matching the theme's motion register.

## Component surface touched

This ships a `@marigold/components: patch`. Three component source files change, so this is not theme and docs only.

- `Sidebar/SidebarToggleIcon.tsx` for the easing bezier rename and curve.
- `Breadcrumbs/Breadcrumbs.tsx` and `Breadcrumbs/HiddenBreadcrumbs.tsx` to drop the dead `size={16}` on the separator chevrons, since the theme owns the size.

Everything else is theme styles (`theme-rui`), the `(examples)` shell, and Storybook stories.

## Accessibility

| Text | Token | Contrast on sidebar bg |
| --- | --- | --- |
| Active / titles | `foreground` | 15.6:1 ✓ |
| Idle nav / section labels | `secondary` | 4.97:1 ✓ |
| (was) section label | ~~`charcoal-500`~~ | ~~3.3:1 ✗~~ |

## Verification

- Contrast verified numerically (section label 3.3:1 to 4.97:1).
- Story tests pass: `Styles/RUI` (back-action alignment and weight), `Sidebar → WithBadges`, and `AppShell → ScrollingSidebar` (nav overflow, footer seam, page-scroll precondition). Firefox exercises the always-on fallback. The Chromium scroll-reveal was verified visually in `docs/(examples)`.

## Out of scope / follow-up

- Pre-existing dangling `text-secondary-muted` in the Table cell-alignment demo (undefined on `main` too).
- VRT / Chromatic not yet triggered.

[DST-1586]: https://reservix.atlassian.net/browse/DST-1586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


themes/theme-rui/src/components/Sidebar.styles.ts
```diff
@@ -14,38 +32,71 @@ export const Sidebar: ThemeComponent<'Sidebar'> = {
...
+  // Two things on every grid row:
+  // 1. `border-…-0`: the seamless shell separates its regions with whitespace,
+  //    so the `border-b`/`border-t` that `ui-panel-*` carries for Panel is
+  //    switched off here.
+  // 2. `min-w-0`: keep the sidebar's inner grid column from auto-sizing to its
+  //    widest item (e.g. a long footer link) and overflowing the fixed `w-64`
+  //    aside — otherwise rows (and the inset pill) get pushed past the edge,
+  //    where the aside's `overflow-hidden` clips them.
+  header: cva({
+    base: 'ui-panel-header h-14 min-w-0 border-b-0',
```
sarahgm: ⚠️ **No seam on a scrolling sidebar.** The sidebar is `grid-rows-[auto_1fr_auto]` with `Sidebar.Nav` as the scroll container between a sticky header and footer. Dropping the header `border-b` / footer `border-t` (here and at L74) reads fine for short lists, but on a long/scrolling nav the `py-1` buffer scrolls away and mid-list items butt against the sticky regions with no divider — and all three share `bg-background`, so the seam blends.

This is a theme-level change affecting all consumers, and the docs demos never scroll so the new story test / VRT won't catch it. It's also inconsistent with the `TopNavigation` fix in this same PR (`ui-scroll-edge`).

Note `ui-scroll-edge` can't be reused as-is: `scroll()` binds to the nearest *ancestor* scroller, but the nav is a grid *sibling* of the header/footer. A scroll-revealed hairline here would need a **named** scroll timeline (`scroll-timeline-name` on the nav, `animation-timeline: --name` on header/footer) — or fall back to a faint always-on hairline / scroll-revealed shadow.

themes/theme-rui/src/components/Breadcrumbs.styles.ts
```diff
@@ -21,9 +21,24 @@ export const Breadcrumbs: ThemeComponent<'Breadcrumbs'> = {
...
-  item: cva({ base: 'inline-flex items-center gap-1.5 sm:gap-2.5' }),
+  item: cva({
+    base: 'inline-flex items-center',
+    variants: {
+      size: {
+        // The chevron scales with the text so the separator stays a quiet
+        // mark between crumbs instead of competing with them (its rendered
+        // 16px default is chunky next to text-sm).
+        small: 'gap-1 sm:gap-1.5 [&_svg]:size-3',
+        default: 'gap-1.5 sm:gap-2.5 [&_svg]:size-3.5',
```
sarahgm, line 32: 💡 These `[&_svg]:size-*` classes now override the hardcoded `size={16}` on `<ChevronRight>` in `Breadcrumbs.tsx` / `HiddenBreadcrumbs.tsx` (author CSS wins over the SVG `width`/`height` attributes), so `default` renders at 14px, not 16px. The render is correct, but those `size={16}` props are now misleading — the source of truth for separator size lives here. Worth a follow-up to drop/comment them.

docs/app/(examples)/_shared/shell-layout.tsx
```diff
@@ -124,71 +137,90 @@ export const ShellLayout = ({
...
+                  // On a drill-in the leaf becomes a real link back to the list.
+                  <Breadcrumbs.Item href={leafHref}>
+                    {leaf.label}
+                  </Breadcrumbs.Item>
+                )}
+                {trailing.map((segment, index) => (
+                  <Breadcrumbs.Item
+                    key={segment}
+                    // The last trailing segment is the page we are on.
+                    href={index === trailing.length - 1 ? pathname : '#'}
```
sarahgm, line 210: 💡 This PR upgraded ancestor crumbs to real links (the "no dead `#` links" goal) and points the last trailing segment at `pathname`, but intermediate trailing segments still get `href="#"`. Minor inconsistency with that goal — acceptable if dynamic segments have no real page, just noting it.

sarahgm (commented):
## Automated Code Review

Reviewed against Marigold standards (TS, component patterns, a11y, React). No correctness bugs found and CI is green; the AA contrast claim was verified numerically (section label 3.29:1 → 4.99:1). The one substantive item is a discuss-level warning about the sidebar losing its header/footer seam once the nav scrolls (theme-level, not exercised by the docs demos). Remaining notes are optional polish. Inline comments below.

---

docs/app/(examples)/_shared/shell-layout.tsx
```diff
@@ -124,71 +137,90 @@ export const ShellLayout = ({
...
+              </Stack>
+            </Sidebar.Footer>
+          </Sidebar>
+          <TopNavigation>
+            <TopNavigation.Start>
+              <Sidebar.Toggle />
+            </TopNavigation.Start>
+            <TopNavigation.Middle>
+              <Breadcrumbs>
+                {ancestors
```
sarahgm, line 186: 💡 With the hard-coded "Home" crumb removed, a path that matches no nav leaf now renders an empty `Breadcrumbs` (`findActive` falls back to `leaf: undefined` with empty ancestors/trailing). Docs-only edge, low severity.

.changeset/sidebar-nav-hierarchy.md
```diff
@@ -0,0 +1,32 @@
+---
+'@marigold/theme-rui': minor
+---
+
+style(Sidebar): quieter navigation hierarchy and a unified app-shell hairline
```
sarahgm: 💡 theme-rui's CHANGELOG convention is ticket-scoped (e.g. `fix([DST-925]): …`); these changesets use component scope (`style(Sidebar):`). Minor consistency point — align if the ticket-scope convention is intended.

themes/theme-rui/src/components/TopNavigation.styles.ts
```diff
@@ -4,7 +4,10 @@ export const TopNavigation: ThemeComponent<'TopNavigation'> = {
   container: cva({
     base: [
       'w-full min-h-14',
-      'border-b border-border',
+      // Seamless shell: borderless at rest; once page content scrolls under
+      // the sticky bar, `ui-scroll-edge` fades in a bottom hairline so the
+      // seam stays legible (progressive enhancement, see ui.css).
+      'ui-scroll-edge',
```
sarahgm, line 10: 💡 Swapping the static `border-b` for `ui-scroll-edge` means browsers without `animation-timeline: scroll()` never show the seam. Disclosed and acceptable as progressive enhancement — noting the trade-off.

themes/theme-rui/src/components/Sidebar.styles.ts
sebald: Fixed in `d617c6e05`. The sticky header and footer now reveal a seam driven by the nav's scroll position: the header's bottom hairline fades in as content scrolls under it, and the footer's top hairline shows while content remains below, fading out as the list bottoms out — seamless at rest, matching the `TopNavigation` scroll-reveal model.

You're right that `ui-scroll-edge` couldn't be reused: the nav is a grid sibling *between* the header and footer, so an anonymous `scroll()` timeline can't reach it. The nav declares a **named** timeline (`--sidebar-nav`), hoisted with `timeline-scope` on the grid wrapper so the header — a preceding sibling — can reference it too. Without scroll-driven-animation support, an `@supports not` fallback pins an always-on hairline, so no engine is left without a divider. The utilities are `ui-sidebar-seam-*`, marked non-reusable since they're specific to this layout.

And the "docs demos never scroll" gap: added an `AppShell → ScrollingSidebar` story with a long nav so the scrolling case is actually exercised (its play-test asserts the nav overflows and the footer seam is present).

themes/theme-rui/src/components/Breadcrumbs.styles.ts
sebald: Good catch — fixed in `4c8ec3330`. Dropped the `size={16}` prop from all four `<ChevronRight>` sites (visible + hidden-measurement copies). The theme's `Breadcrumbs.item` slot now solely owns the separator size via `[&_svg]:size-*`, so there's no dead number fighting the CSS. Themes without a separator-size rule fall back to the icon default.

docs/app/(examples)/_shared/shell-layout.tsx
sebald: Intentional: only the *last* trailing segment is a real page (the one we're on → `pathname`). Intermediate dynamic segments (e.g. a `users/:id/…` middle) have no standalone route in the examples, so `#` is the honest placeholder rather than a link that 404s. If a segment ever gains a real page, `config.resolveLabel` is the natural place to also resolve its href. Leaving as-is.

docs/app/(examples)/_shared/shell-layout.tsx
sebald: Acknowledged — docs-only, and not reachable by the shipped examples (every example route is in the nav config, so `findActive` always resolves a leaf or trailing segment). An empty `<Breadcrumbs>` renders a harmless empty `<ol>`. Not adding a guard to example code for a path that can't occur here; if we later drive the shell from dynamic routes, a `leaf ?? trailing.length` guard around `TopNavigation.Middle` is the fix.

.changeset/sidebar-nav-hierarchy.md
sebald: Aligned in `4c8ec3330` — all three changesets now use the ticket scope `style(DST-1586): …` to match the theme-rui CHANGELOG convention (`feat(DST-…)`, `fix(DST-…)`).

themes/theme-rui/src/components/TopNavigation.styles.ts
sebald: Yep, deliberate trade-off: without `animation-timeline: scroll()` the bar stays borderless, which is the intended at-rest look anyway, and the sticky bar's opaque background still separates it from passing content. The sidebar seams (the other thread) now use the same reveal model for consistency, with an always-on hairline fallback there since a sidebar divider carries more structural weight than the top bar's.

packages/components/src/Sidebar/SidebarToggleIcon.tsx
```diff
@@ -5,7 +5,9 @@ import type { IconProps } from '../icons/Icons.types';
 
 const CHEVRON_LEFT = 'M 16 15 L 13 12 L 16 9';
 const CHEVRON_RIGHT = 'M 13 15 L 16 12 L 13 9';
-const SPRING_BEZIER = 'cubic-bezier(0.34, 1.56, 0.64, 1)';
+// ease-out-quint: fast start, smooth settle — no overshoot, matching the
+// theme's motion register (real objects decelerate, they don't bounce).
+const SETTLE_BEZIER = 'cubic-bezier(0.22, 1, 0.36, 1)';
```
sarahgm, line 10: ⚠️ The PR body's "Constraint" section says *"No changes to the Sidebar component implementation — this is theme styles + docs demo only."* This bezier change is Sidebar component source, and the PR also changes `Breadcrumbs.tsx` and `HiddenBreadcrumbs.tsx`, shipping a `@marigold/components: patch` release. Please correct the body so reviewers don't under-test the component surface — three shipped component files change, not just theme + docs.

packages/components/src/Styles.stories.tsx
```diff
@@ -222,3 +226,145 @@ export const Surface = meta.story({
...
+      <Stack space="group">
+        <Headline level="3">Sidebar navigation</Headline>
+        <p className="text-secondary max-w-prose text-sm">
+          Hierarchy comes from two token-driven charcoal tiers plus the inset
+          active pill — never from raw colours. The section label stays a clear
+          tier through treatment (uppercase, smaller, heavier, tracked) rather
+          than a lighter colour, so it holds the WCAG AA 4.5:1 contrast floor.
+        </p>
+        <Inline space="section" alignY="top">
+          <div className="border-surface-border/60 w-fit overflow-hidden rounded-lg border">
```
sarahgm: 💡 This specimen frame uses `border-surface-border/60`, but the shipped sidebar divider is full-strength `border-surface-border` (see `Sidebar.styles.ts`). The ticket scope and PR "Accessibility" prose still describe the `/60` unified hairline — worth reconciling the narrative with what actually shipped (the changeset already explains the full-strength decision).

packages/components/src/AppShell/AppShell.stories.tsx
```diff
@@ -274,3 +321,60 @@ export const WideContentDoesNotOverflow = meta.story({
...
+    await expect(nav.scrollHeight).toBeGreaterThan(600);
+
+    // The footer shows its seam whenever nav content sits below it — which it
+    // does at load (nav rests at the top of an overflowing list). That holds
+    // both where scroll-driven animations run (the timeline pins the "content
+    // below" keyframe) and in the always-on fallback, so the box-shadow
+    // hairline is present here regardless of engine.
+    const footer = canvas
+      .getByRole('complementary')
+      .querySelector<HTMLElement>('[class*="ui-panel-actions"]');
```
sarahgm: 💡 `querySelector('[class*="ui-panel-actions"]')` couples the test to a compiled theme class name, so it'll break if the slot class is renamed. Defensible here since there's no accessible handle for a decorative seam — just noting the brittleness.

sarahgm (commented):
## Automated Code Review

Reviewed against Marigold standards, dependency integrity, and accessibility. **CI is fully green** and I found **no correctness or a11y regressions** — verified the `--color-surface-border` stacking dependency (#5566) has landed on `beta-release`, and that all custom tokens/utilities (`size-control-small`, `rounded-surface`, `bg-selected`/`bg-hover`, Badge `admin`/`master`) resolve. The `HiddenBreadcrumbs` chevrons that dropped `size={16}` still get sized via the theme's `[&_svg]:size-*` on the `item` slot, so measurement stays correct.

One warning-level doc fix and a couple of nice-to-haves inline. VRT is not enforced here since this targets `beta-release` (it runs when the release flows to `main`).

**Recommendation: Comment** — approve pending the PR-body correction below.

📋 Body-level cleanup (can't anchor to a line):
- **Stale callout:** the `[!IMPORTANT] Draft — stacked on #5566 / retarget to beta-release` block is obsolete — base is already `beta-release` and the token landed. Remove it and mark ready.

---

packages/components/src/Sidebar/SidebarToggleIcon.tsx
sebald: You're right, the body was stale. It described the original "quieter hierarchy" scope and never caught up to the seamless-shell work. I dropped the "theme + docs only" constraint and added a "Component surface touched" section that names all three shipped component files (`SidebarToggleIcon.tsx`, `Breadcrumbs.tsx`, `HiddenBreadcrumbs.tsx`) and the `@marigold/components: patch`, so the component surface is called out for review rather than hidden.

packages/components/src/Styles.stories.tsx
sebald: Reconciled both ways. I bumped the two specimen frames from `border-surface-border/60` to full-strength `border-surface-border` in `481ba1344`, so the demo matches the shipped divider and this story's own legend row, which already labels full-strength as "the shell's one line". I also rewrote the PR body. The stale "/60 unified hairline" bullet is gone, replaced with the full-strength divider and borderless-elsewhere seamless-shell description that matches the changeset.

packages/components/src/AppShell/AppShell.stories.tsx
sebald: Agreed it's a class-coupling. I kept it as-is on purpose. A decorative scroll seam has no accessible handle to target, and the test already fails loudly (`expect(footer).not.toBeNull()`) if `ui-panel-actions` is renamed, so it degrades to a clear signal rather than a silent false-pass. If the seam ever needs a stable test hook I'd add a `data-*` marker on `Sidebar.Footer` rather than assert on compiled classes.

packages/components/src/AppShell/AppShell.stories.tsx
```diff
@@ -274,3 +321,63 @@ export const WideContentDoesNotOverflow = meta.story({
...
+ *
+ * Both are scroll-linked (not autonomous), so intentionally not gated by
+ * reduced-motion. In browsers without scroll-driven animations the sidebar
+ * seams fall back to an always-on hairline; the top bar stays borderless (its
+ * opaque background still separates it from passing content).
+ */
+export const ScrollingSidebar = meta.story({
+  tags: ['component-test'],
+  render: () => <ShellFrame longNav longPage />,
+  play: async ({ canvasElement }) => {
```
aromko: Please turn this into a .test function in the stories, e.g. ScrollingSidebar.test() to stay consistent. Also think about if this needs to be snapshot for chromatic or not.

sebald: Done in `54bc16681`. Converted the inline play to `ScrollingSidebar.test()` to match the `.test()` convention. On Chromatic, I disabled the snapshot for this one. It is a behavior test whose seam reveals are scroll-driven, so a static frame cannot capture them, and its 40-item plus 10-panel layout would only add a large, noisy snapshot over what the Basic story already covers. The `.test()` assertions guard the behavior (nav overflow, footer seam present, page scrolls under the top bar). I am also converting the sibling AppShell stories (`Basic`, `ControlledSidebar`, `WideContentDoesNotOverflow`) to `.test()` in this PR so the whole file lands on one convention.

sarahgm (approved):

PR #5593, opened 2026-07-01
feat(DST-1587): share wheel-to-horizontal-scroll between Tabs and SegmentedControl

sarahgm:
# Description

Extracts the wheel-to-horizontal-scroll handler that DST-1282 built for `Tabs` into a shared `useWheelScrollX` hook, and adopts it in `SegmentedControl` too. Previously, wheeling vertically over an overflowing `SegmentedControl` track did nothing; now it scrolls the track horizontally, matching `Tabs`. The hook only hijacks the wheel when the container overflows and the gesture is vertical-dominant, leaves ctrl+wheel zoom alone, normalizes Firefox line-mode deltas, and doesn't trap the wheel at the scroll ends. No public API change on either component.

Stacked on #5578 (DST-1282) — merge that first.

Closes DST-1587

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Open Storybook → `Components/SegmentedControl/Overflow` on a narrow viewport.
2. With a physical mouse (no trackpad), hover the track and scroll the wheel vertically — the segments should scroll horizontally.
3. Confirm ctrl+wheel still zooms the page, and that a trackpad's horizontal swipe still scrolls natively.
4. Repeat over `Components/Tabs/Mobile` to confirm no regression.
5. `pnpm test:unit` — covers the wheel matrix (`useWheelScrollX.test.tsx`) and the component wiring.

## Breaking Changes

No

## Checklist


packages/components/src/SegmentedControl/SegmentedControl.test.tsx
```diff
@@ -132,3 +133,64 @@ test('forwards a ref to the group element', () => {
 
   expect(ref.current).toBeInstanceOf(HTMLElement);
 });
+
+describe('translates a vertical mouse wheel into horizontal scroll', () => {
```
sebald: These two wiring tests are a strict subset of `useWheelScrollX.test.tsx`, and the `Overflow` play test already proves the real-CSS integration. Meanwhile `Tabs.test.tsx` drops *all* of its wheel tests in this PR — so the two adopters end up tested to different standards for identical wiring.

I'd lean toward deleting these here (the hook owns the logic, the play test owns the wiring) so the two components stay symmetric. If you'd rather keep unit-level wiring coverage, add one equivalent test back to Tabs instead.

packages/components/src/SegmentedControl/SegmentedControl.tsx
```diff
@@ -109,6 +110,10 @@ function SegmentedControlBase({
   const stretch = width === 'full';
   const scrollRef = useRef<HTMLDivElement | null>(null);
 
+  // Let pointer users without a trackpad reach overflowing segments by wheeling
```
sebald: This comment now restates the hook's JSDoc first paragraph, and `TabList.tsx` carries the same duplicated sentence. The `useWheelScrollX` name + its JSDoc are self-documenting — I'd drop the call-site comments in both places.

packages/components/src/SegmentedControl/SegmentedControl.tsx
```diff
@@ -177,7 +182,11 @@ function SegmentedControlBase({
           stretch ? 'flex w-full' : 'inline-flex w-fit max-w-full'
         )}
       >
-        <div ref={attachScrollContainer} className={classNames.list}>
+        <div
+          ref={attachScrollContainer}
+          className={classNames.list}
+          data-testid="segmented-control-scroll"
```
sebald: Minor: this ships a `data-testid` into production DOM. It matches the existing `tabs-list-scroll` precedent so it's consistent, but it's test-coupling in shipped markup and cuts against the repo's "prefer accessible queries" guidance. Fine to keep for consistency — flagging so it's a conscious choice rather than drive-by.

packages/components/src/utils/useWheelScrollX.ts
```diff
@@ -0,0 +1,68 @@
...
+        return;
+      }
+      const delta =
+        event.deltaMode === 1 ? event.deltaY * LINE_HEIGHT : event.deltaY;
+      // Don't trap the wheel at the ends: when there's no room left in the
+      // wheel's direction, let the event bubble so the page can scroll past the
+      // container instead of swallowing the gesture.
+      const max = el.scrollWidth - el.clientWidth;
+      if (
+        (delta < 0 && el.scrollLeft <= 0) ||
```
sebald, line 53: This end-trap logic assumes LTR (`delta > 0` ⇒ right, positive `scrollLeft`). That was safe to inherit for `Tabs`, but adopting the hook in `SegmentedControl` is a new surface where it runs — and `scrollLeft` sign conventions differ by browser in RTL.

Is RTL in scope for either component? If so this check may need normalizing; if not, no change needed — just confirming it's intentional.

sebald (commented):
Clean extraction — the hook is a verbatim lift of the Tabs logic (−169 / +68 shared), and coverage is preserved by moving the full wheel matrix into `useWheelScrollX.test.tsx`. No correctness issues or regressions found. A few minor/opinion notes inline; nothing blocking.

packages/components/src/SegmentedControl/SegmentedControl.test.tsx
sarahgm: Went with deleting these for symmetry — the hook test owns the logic and Tabs already relies on that plus the `Overflow` play test. Both adopters now stay on the same footing. Done in 08f729dbb.

packages/components/src/SegmentedControl/SegmentedControl.tsx
sarahgm: Dropped the call-site comment in both `SegmentedControl` and `TabList` — the hook name + JSDoc carry it. Done in 08f729dbb.

packages/components/src/SegmentedControl/SegmentedControl.tsx
sarahgm: Good flag. Since the wiring tests above are gone, `segmented-control-scroll` was the only consumer, so I removed it — and took out the already-unused `tabs-list-scroll` too. No test-coupling ships in the DOM now. Done in 08f729dbb.

packages/components/src/utils/useWheelScrollX.ts
sarahgm: Confirmed intentional — RTL isn't in scope for either component, so the LTR `scrollLeft` end-trap assumption is fine to inherit. Left as-is; we can normalize the sign convention if/when RTL support lands.

sebald (approved):

PR #5594, opened 2026-07-01
test(DST-1589): pin DateRangePicker story calendars to a fixed month

sarahgm:
# Description

The `Basic`, `UnavailableDate` and `MultiMonth` DateRangePicker stories rendered with no `value`/`defaultValue`. Their open-calendar Chromatic snapshots therefore defaulted to the *current* month and drifted every month, producing spurious visual diffs with no code change.

This pins each of the three stories to `placeholderValue={new CalendarDate(2026, 7, 1)}` — fixing the opening month without pre-selecting a range. July 2026 matches the current live baseline layout, so the one-time approval diff is near-zero. The already-pinned stories (`Controlled`, `MinMax`, `Mobile`) are untouched.

Closes DST-1589

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm test:sb --run DateRangePicker` — all 5 story tests pass.
2. In Storybook (`pnpm sb`), open Components/DateRangePicker → Basic, open the calendar, confirm it lands on July 2026.

## Breaking Changes

No

## Checklist

- [ ] Changeset added (`pnpm changeset`) — N/A, stories file is not published

### DIFF
--- a/packages/components/src/DateRangePicker/DateRangePicker.stories.tsx
+++ b/packages/components/src/DateRangePicker/DateRangePicker.stories.tsx
@@ -109,6 +109,7 @@ export const Basic: any = meta.story({
         label="Date Range Picker"
         description="This is a description"
         errorMessage="This is an error"
+        placeholderValue={new CalendarDate(2026, 7, 1)}
         {...args}
       />
     </I18nProvider>
@@ -168,6 +169,7 @@ export const UnavailableDate: any = meta.story({
       <DateRangePicker
         label="Date Range Picker"
         dateUnavailable={date => date.toDate('Europe/Berlin').getDay() === 0}
+        placeholderValue={new CalendarDate(2026, 7, 1)}
         {...args}
       />
     </I18nProvider>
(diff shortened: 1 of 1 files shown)


sebald:
Hmmm, why is this called placeholderValue. shouldn't it be `defaultValue` 👀 I am confused

sarahgm:
Good question @sebald  — `defaultValue` would actually change what these stories demonstrate. 👀

The difference:

- **`defaultValue`** sets an *uncontrolled selected value* — it fills the segments and highlights a selected range in the calendar. The `Basic` story is meant to show the **empty** picker, so pre-selecting a range would misrepresent its default/empty state (and `Controlled`/`Mobile` already cover the has-a-value case).
- **`placeholderValue`** selects nothing. It only controls which month the calendar opens on (and the placeholder shown in empty segments) when there's no value. The picker stays empty — exactly what `Basic` should show.

The whole bug is that with no value at all, react-aria defaults the opening month to *today*, so the open-calendar Chromatic snapshot drifted every month. `placeholderValue` pins that opening month to a fixed date **without** turning the empty picker into a filled one — so it fixes the drift while keeping each story's intent intact.

TL;DR: `defaultValue` = pre-selected range (wrong for an "empty state" story); `placeholderValue` = empty picker, fixed opening month.

sebald (approved):

PR #5595, opened 2026-07-01
feat(DST-1381): add master/admin access variants to Link, MenuItem, Badge

aromko:
# Description

Adds `master` / `admin` **access variants** to `Link` and `MenuItem`, and moves `Badge`'s master/admin glyph from the `Lock` icon component to a themed CSS `::before` mask (lock for `master`, key for `admin` — so admin finally gets its own glyph). The glyph treatment lives in shared `ui-access-master` / `ui-access-admin` utilities in the theme, so the same rendering works in React **and** in the generated BEM/Twig CSS.

**Accessibility:** the utilities expose the restriction via CSS alternative text (`content: '' / 'Master'`), so restricted links, menu items, and badges get an accessible name prefix — no deferral.

**API decision (from the review discussion):** access marks stay on the **`variant` axis** — no dedicated `access` prop. `variant` is a single axis, so an access variant cannot combine with e.g. `destructive`; for restricted destructive actions the **access variant takes precedence**, with the destructive nature conveyed by the label and confirmation flow. This rule is documented in the pattern docs and shown in the demos/stories.

The branch is rebased onto `beta-release` (adopts the new `tokens.css` token home, the tinted Badge chip design, and the renamed Menu tokens).

Closes DST-1381

## Where to test

**Storybook** ([branch preview](https://marigold-storybook-git-mk-dst-1381-access-mark-marigold.vercel.app) or `pnpm sb`):

- **Components/Link → Access Variants** — `master` (lock) and `admin` (key) links
- **Components/Menu → Access Sections** — `ActionMenu` with `Master-Aktionen` / `Admin-Aktionen` sections, including the master-only destructive `Löschen` item using `variant="master"`
- **Components/Badge → Basic** — master/admin chips now render the mask glyph (key for admin) instead of the `Lock` icon
- **Patterns/Master- & Admin-Mark** — end-to-end showcase: Badge Variants, Table Rows, Form, Inline Table Actions, Menu With Sections

**Docs** ([branch preview](https://marigold-docs-git-mk-dst-1381-access-mark-marigold.vercel.app) or `pnpm start`):

- **Patterns → Layout → Admin- and Mastermark** — new *Inline actions* and *Menus* sections with demos, plus *Combining access marks with other variants* (the single-axis rule)
- **Foundations → Token Overview → Access control** — the `--access-mask-lock` / `--access-mask-key` variables and `ui-access-*` utilities

## Test Instructions

1. Open the Storybook preview and check the stories listed above; in *Access Sections*, open the menu and confirm restricted items carry the glyph (`ui-access-master` / `ui-access-admin`) and `Löschen` renders as a master item, not destructive.
2. Open the docs preview pages listed above and confirm the demos render.
3. Locally: `pnpm typecheck:only`, `pnpm test:unit` (Link/Badge), `pnpm test:sb` (Menu access-section story tests).

## Breaking Changes

No — additive `minor` changes to `@marigold/components` and `@marigold/theme-rui` (new variants, theme mask vars, shared utilities).

## Checklist

- [x] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) (for new/changed interactive components) — restriction exposed via CSS alternative text, verified in the browser accessibility tree
- [ ] Visual regression tests updated (for UI changes) — `/run-chromatic` will be triggered once the open review threads are resolved


themes/theme-rui/src/components/Badge.styles.ts
```diff
@@ -15,9 +15,9 @@ export const Badge: ThemeComponent<'Badge'> = cva({
       info: 'bg-info-muted text-info-muted-foreground',
       error: 'bg-destructive-muted text-destructive-muted-foreground',
       admin:
-        'text-access-admin-foreground border border-access-admin-foreground bg-background',
+        'text-access-admin-foreground border border-access-admin-foreground bg-background before:content-[""] before:inline-block before:size-4 before:mr-1 before:bg-current before:[mask-size:contain] before:[mask-repeat:no-repeat] before:[mask-position:center] before:[mask-image:var(--access-mask-key)]',
```
sarahgm: 🔴 **Visual Regression Tests — not run.** This PR changes theme styles, `theme.css`, and stories — a purely visual glyph change — but the `Visual-Regression-Tests` workflow has never run for this branch. Please trigger it by commenting `/run-chromatic` on this PR.

themes/theme-rui/src/components/Link.styles.ts
```diff
@@ -6,6 +6,10 @@ export const Link: ThemeComponent<'Link'> = cva({
     variant: {
       default: 'text-link font-normal',
       secondary: 'font-medium text-foreground underline hover:no-underline',
+      master:
+        'text-link font-normal inline-flex items-center before:content-[""] before:size-4 before:shrink-0 before:mr-1 before:bg-access-master-foreground before:[mask-size:contain] before:[mask-repeat:no-repeat] before:[mask-position:center] before:[mask-image:var(--access-mask-lock)]',
```
sarahgm: 🔴 The master/admin restriction here (and in `Badge.styles.ts`) is conveyed **only** through a decorative `::before` mask glyph + color. `content-[""]` adds nothing to the accessibility tree, so a screen-reader user cannot distinguish a `master` link from a normal one — fails WCAG 1.1.1 (Non-text Content). The `Menu` path is fine (textual `Master-Aktionen` / `Admin-Aktionen` section titles), but `Link`/`Badge` have no accessible name.

The PR defers this "tracked in DST-1381" — but that's the ticket this PR closes, so the deferral has no home. Please either add a visually-hidden label now or open a dedicated a11y follow-up and reference it in the changeset.

packages/components/src/Link/Link.test.tsx
```diff
@@ -2,6 +2,9 @@ import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { createRef } from 'react';
 import { MockInstance, vi } from 'vitest';
+import { theme } from '@marigold/theme-rui';
```
sarahgm: 🟡 These tests hand-roll a theme fixture (`MarigoldProvider theme={theme}`). CLAUDE.md: *"Don't create test fixtures/themes — import stories instead."* You now have an `AccessVariants` story — import it and render `<AccessVariants.Component />` (the `Badge` test proves the story decorator already supplies the theme, so the manual wrap is redundant). Also add a blank line between arrange/act/assert to match the file's existing tests.

packages/components/src/Badge/Badge.test.tsx
```diff
@@ -30,11 +30,14 @@ test('renders the info variant correctly', () => {
   `);
 });
 
-test('shows the lock svg', () => {
+test('master variant applies the lock mask glyph', () => {
```
sarahgm: 🟡 Two nits: (1) add a blank line between arrange/act/assert (the `renders the info variant` test above does this). (2) Asserting on the raw Tailwind class string is brittle — it couples the test to the theme utility. It's the only observable surface **because** the glyph has no accessible name (see the `Link.styles.ts` comment); fixing that would let this become a behavior assertion.

packages/components/src/patterns/admin-master-mark.stories.tsx
```diff
@@ -87,3 +90,71 @@ export const Form = meta.story({
...
+      </Table.Body>
+    </Table>
+  ),
+});
+
+export const MenuWithSections = meta.story({
+  parameters: { chromatic: { disableSnapshot: true } },
+  decorators: [
+    Story => (
+      <div id="storybook-root">
```
sarahgm: 💡 The hardcoded `<div id="storybook-root">` decorator reads as a portal-mount workaround. If it's load-bearing, a one-line comment would help; if not, drop it.

sarahgm (changes requested):
Automated review via review-pr skill. 2 blockers (missing VRT, a11y gap), plus test-quality and a nit. See inline comments.

themes/theme-rui/src/components/Link.styles.ts
```diff
@@ -6,6 +6,10 @@ export const Link: ThemeComponent<'Link'> = cva({
...
+      admin:
+        'text-link font-normal inline-flex items-center before:content-[""] before:size-4 before:shrink-0 before:mr-1 before:bg-access-admin-foreground before:[mask-size:contain] before:[mask-repeat:no-repeat] before:[mask-position:center] before:[mask-image:var(--access-mask-key)]',
```
sebald: **Consolidation (applies to Badge/Link/Menu styles):** this `before:*` glyph chain is repeated six times across `Badge.styles.ts`, `Link.styles.ts`, and `Menu.styles.ts` (master + admin each). `ui.css` already has the mechanism for exactly this kind of shared treatment (`@utility ui-state-focus`, `ui-surface`, …). Extracting the boilerplate:

```css
@utility ui-access-glyph {
  &::before {
    content: '';
    display: inline-block;
    width: 1rem;
    height: 1rem;
    margin-right: 0.25rem;
    mask-size: contain;
    mask-repeat: no-repeat;
    mask-position: center;
  }
}
```

collapses each variant to something like:

```ts
master:
  'text-link font-normal inline-flex items-center ui-access-glyph before:bg-access-master-foreground before:[mask-image:var(--access-mask-lock)]',
```

That gives the glyph treatment the single source of truth the PR description promises ("the glyph lives entirely in the theme") and keeps the six copies from drifting apart — the Menu variants have already drifted slightly (see comment there). Since `ui-*` utilities are consumed from the theme styles today, this should stay compatible with the generated BEM/Twig CSS as well.

packages/components/src/Badge/Badge.tsx
```diff
@@ -1,6 +1,5 @@
 import type { ReactNode } from 'react';
 import { useClassNames } from '@marigold/system';
-import { Lock } from '../icons/Lock';
```
sebald, line 3: With this import gone, `packages/components/src/icons/Lock.tsx` has no consumers left anywhere in the repo — it isn't exported from the package index, and docs/themes don't import it either. Worth deleting the icon file in this PR so it doesn't linger as dead code.

sarahgm:
🟡 **Test style: missing Arrange/Act/Assert blank lines**

All four new tests run their statements together with no AAA separation, unlike the existing tests in both files. Convention is a blank line before the assert block.

**`Badge.test.tsx`** (`master variant applies...`, `admin variant applies...`):

```tsx
test('master variant applies the lock mask glyph', () => {
  render(<Basic.Component />);
  const master = screen.getAllByText('Status')[VARIANT_ORDER.indexOf('master')];

  expect(master).toHaveClass('before:[mask-image:var(--access-mask-lock)]');
});
```

The `renders the info variant correctly` test just above already does this.

**`Link.test.tsx`** (`master variant...`, `admin variant...`):

```tsx
  const link = screen.getByRole('link');

  expect(link).toHaveClass('before:[mask-image:var(--access-mask-lock)]');
  expect(link).toHaveClass('before:bg-access-master-foreground');
  expect(link).toHaveTextContent('verschieben');
```

The existing `supports "onPress"` test in the same file already separates its sections this way.

None of these have an Act step (static render assertions, no user interaction), so it is just a single blank line between Arrange and Assert in each of the four tests.

packages/components/src/Link/Link.test.tsx
```diff
@@ -60,3 +63,31 @@ test('supports "onPress"', async () => {
...
+      <Link variant="master" href="#">
+        verschieben
+      </Link>
+    </MarigoldProvider>
+  );
+  const link = screen.getByRole('link');
+  expect(link).toHaveClass('before:[mask-image:var(--access-mask-lock)]');
+  expect(link).toHaveClass('before:bg-access-master-foreground');
+  expect(link).toHaveTextContent('verschieben');
+});
```
sebald: These two tests hand-roll the same markup the new `AccessVariants` story already renders, plus `MarigoldProvider`/`@marigold/theme-rui` boilerplate. Our convention ("import stories instead of creating fixtures/inline themes") — and `Badge.test.tsx` in this very PR — show the leaner pattern:

```tsx
import { AccessVariants } from './Link.stories';

test('access variants apply the mask glyph in the access color', () => {
  render(<AccessVariants.Component />);
  const [master, admin] = screen.getAllByRole('link');
  expect(master).toHaveClass('before:[mask-image:var(--access-mask-lock)]');
  expect(master).toHaveClass('before:bg-access-master-foreground');
  expect(admin).toHaveClass('before:[mask-image:var(--access-mask-key)]');
  expect(admin).toHaveClass('before:bg-access-admin-foreground');
});
```

Single canonical markup for the access links, and the provider setup + extra imports go away.

sarahgm:
I think it makes more sense targeting this in `beta-release` since you changed styles, what do you think? @aromko

packages/components/src/Menu/MenuItem.tsx
```diff
@@ -6,7 +6,7 @@ import { useClassNames } from '@marigold/system';
 // ---------------
 type RemovedProps = 'style' | 'className';
 export interface MenuItemProps extends Omit<RAC.MenuItemProps, RemovedProps> {
-  variant?: 'destructive' | 'default' | (string & {});
+  variant?: 'destructive' | 'default' | 'master' | 'admin' | (string & {});
```
sebald, line 15: **API design question:** `variant` now multiplexes two orthogonal axes — style (`destructive`) and access (`master`/`admin`). A master-only *destructive* action becomes unrepresentable (`variant` can only hold one value), and DST-1381 explicitly names master-only + destructive (e.g. "löschen") as a real co-occurrence — the ticket's own sketch used a separate prop for this reason (`<ActionMenu.Item access="admin">`). Same on `Link`: `secondary` + `admin` can't combine.

Since this ships as an additive minor, switching later means deprecating variants, so it's worth deciding deliberately now. If we stick with `variant` (simpler, consistent with `Badge`/`Table.Row`), let's document the constraint; otherwise a dedicated `access?: 'master' | 'admin'` prop as a second cva axis keeps the axes composable.

themes/theme-rui/src/components/Menu.styles.ts
```diff
@@ -20,6 +20,10 @@ export const Menu: ThemeComponent<'Menu'> = {
         default:
           'text-foreground focus:bg-focus [&_svg]:text-muted-foreground [&_svg]:opacity-60',
         destructive: 'text-destructive focus:bg-destructive/10',
+        master:
+          'text-foreground focus:bg-focus before:content-[""] before:inline-block before:size-4 before:mr-1 before:bg-access-master-foreground before:[mask-size:contain] before:[mask-repeat:no-repeat] before:[mask-position:center] before:[mask-image:var(--access-mask-lock)]',
+        admin:
+          'text-foreground focus:bg-focus before:content-[""] before:inline-block before:size-4 before:mr-1 before:bg-access-admin-foreground before:[mask-size:contain] before:[mask-repeat:no-repeat] before:[mask-position:center] before:[mask-image:var(--access-mask-key)]',
```
sebald: Small drift risk: `master`/`admin` restate `default`'s `text-foreground focus:bg-focus`, but silently drop its `[&_svg]:text-muted-foreground [&_svg]:opacity-60` — an icon inside a master/admin item renders full-strength, unlike every other non-destructive item. Is that intended? (Same restating pattern in `Link.styles.ts`, where `text-link font-normal` is repeated from `default`.)

If the shared glyph utility suggested in `Link.styles.ts` is adopted, composing base + glyph removes this class of drift altogether.

themes/theme-rui/src/theme.css
```diff
@@ -108,7 +108,11 @@
   --color-access-admin: var(--color-purple-100);
   --color-access-admin-foreground: var(--color-purple-800);
 
-  /* 
+  /* mask-image shapes for access-control glyphs (lock = master, key = admin) */
+  --access-mask-lock: url('data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAyNCAyNCIgZmlsbD0ibm9uZSIgc3Ryb2tlPSIjMDAwIiBzdHJva2Utd2lkdGg9IjIiIHN0cm9rZS1saW5lY2FwPSJyb3VuZCIgc3Ryb2tlLWxpbmVqb2luPSJyb3VuZCI+PGNpcmNsZSBjeD0iMTIiIGN5PSIxNiIgcj0iMSIvPjxyZWN0IHg9IjMiIHk9IjEwIiB3aWR0aD0iMTgiIGhlaWdodD0iMTIiIHJ4PSIyIi8+PHBhdGggZD0iTTcgMTBWN2E1IDUgMCAwIDEgMTAgMHYzIi8+PC9zdmc+');
+  --access-mask-key: url('data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAyNCAyNCIgZmlsbD0ibm9uZSIgc3Ryb2tlPSIjMDAwIiBzdHJva2Utd2lkdGg9IjIiIHN0cm9rZS1saW5lY2FwPSJyb3VuZCIgc3Ryb2tlLWxpbmVqb2luPSJyb3VuZCI+PHBhdGggZD0ibTE1LjUgNy41IDIuMyAyLjNhMSAxIDAgMCAwIDEuNCAwbDIuMS0yLjFhMSAxIDAgMCAwIDAtMS40TDE5IDQiLz48cGF0aCBkPSJtMjEgMi05LjYgOS42Ii8+PGNpcmNsZSBjeD0iNy41IiBjeT0iMTUuNSIgcj0iNS41Ii8+PC9zdmc+');
```
sebald: The base64 payloads are write-only — nobody can review or tweak the shapes without decoding them first. URL-encoded SVG is equally valid in `mask-image` and stays readable/diffable:

```css
--access-mask-lock: url('data:image/svg+xml,%3Csvg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="%23000" ...%3E%3Ccircle cx="12" cy="16" r="1"/%3E...%3C/svg%3E');
```

Also worth a short comment naming the source glyphs (these look like lucide `lock` and `key`), so future edits regenerate from the source icon instead of hand-editing a data URI.

sebald (commented):
Simplification-focused review (reuse/consolidation/API altitude — not a bug hunt). Six inline notes, roughly in priority order:

1. Extract the 6× repeated `before:*` glyph chain into a `ui-*` utility (Link.styles.ts, applies to Badge/Menu too)
2. Delete the now-unused `icons/Lock.tsx` (Badge.tsx)
3. Link tests can import the new `AccessVariants` story instead of hand-rolling provider + markup (Link.test.tsx)
4. `variant` conflates style and access axes — master+destructive becomes unrepresentable (MenuItem.tsx)
5. Menu master/admin variants drop `default`'s `[&_svg]` styling — drift or intended? (Menu.styles.ts)
6. Prefer URL-encoded SVG over base64 for the mask vars (theme.css)

Also noting here rather than inline: the PR says "Closes DST-1381" while deferring a11y "tracked in DST-1381" — merging closes the ticket that tracks the deferral, so a dedicated follow-up ticket would keep it from getting lost.

sarahgm:
would also add it in the documenation about "Admin & Mastermark" https://marigold-docs-oql5ardj1-marigold.vercel.app/foundations/token-overview#form-controls

aromko:
@sebald , @sarahgm thx for the review but my initial request was if we want to go this way if the admin/master mark for inline and menu stuff. Based on the comments I would say yes so I also will adjust the docs.

themes/theme-rui/src/components/Badge.styles.ts
aromko: VRT stays deliberately un-run for the moment — we will trigger `/run-chromatic` once all review comments are resolved, so Chromatic snapshots the final state exactly once.

themes/theme-rui/src/components/Link.styles.ts
aromko: Fixed via CSS alternative text on the glyph itself: the new shared `ui-access-*` utilities set `content: '' / 'Master'` (resp. `'Admin'`), so the restriction lands in the accessible name. Verified in the browser accessibility tree: the link announces as “Master verschieben”, menu items as “Master Verschieben”. Since it lives in the CSS, it also covers the generated BEM/Twig output, not just React.

Badge uses this mechanism as of 8c6cf3bd8. Link/Menu switch to the same utilities together with the `access`-prop change (see https://github.com/marigold-ui/marigold/pull/5595#discussion_r3513131103) and then inherit the alt text as well. No follow-up ticket needed — solved in-PR; the changeset's deferral sentence is removed in that pending commit.

packages/components/src/Link/Link.test.tsx
aromko: Rewritten locally on the `AccessVariants` story (provider fixture and extra imports gone, blank line between arrange/assert). It lands together with the `access`-prop change, because the story then renders `access="master"` instead of `variant="master"` — see https://github.com/marigold-ui/marigold/pull/5595#discussion_r3513131103.

packages/components/src/Badge/Badge.test.tsx
aromko: Fixed in 8c6cf3bd8 — blank line added, and the assertions now target the semantic `ui-access-master`/`ui-access-admin` utility instead of the raw Tailwind chain. That decouples the test from the styling internals; it stays class-based because jsdom/testing-library cannot see pseudo-element alt text — the accessible-name behavior itself was verified in the browser accessibility tree (see the `Link.styles.ts` a11y thread).

packages/components/src/patterns/admin-master-mark.stories.tsx
aromko: It is load-bearing: `.storybook/decorators.tsx` wraps every story in `OverlayContainerProvider container="storybook-root"`, and when a story renders outside the Storybook iframe (vitest browser tests, unit tests importing stories) no `#storybook-root` exists, so overlays need this element as their portal target. Same uncommented pattern as `Menu.stories.tsx` and ~20 other story files — keeping it consistent with those rather than commenting only this one.

themes/theme-rui/src/components/Link.styles.ts
aromko: Extracted in 8c6cf3bd8, one step further than sketched: `ui-access-master` / `ui-access-admin` bundle box + mask + color + CSS alternative text (`content: '' / 'Master'`), since those always travel together — that resolves the a11y blocker and the Menu `[&_svg]` drift through the same single source of truth. Badge consumes the utilities already; Link/Menu switch over together with the `access`-prop change (https://github.com/marigold-ui/marigold/pull/5595#discussion_r3513131103) so their variant groups only change once.

packages/components/src/Badge/Badge.tsx
aromko: Deleted in 6e2a1b06c.

packages/components/src/Link/Link.test.tsx
aromko: Done locally, essentially as sketched (two tests, one per access level, story import instead of the provider fixture). Lands with the `access`-prop change, since `AccessVariants` then uses `access=` instead of `variant=`.

themes/theme-rui/src/components/Menu.styles.ts
aromko: Not intended — and it disappears structurally with the `access` axis: master/admin become an `access` variant group that composes with `variant: default`, so the `[&_svg]` styling is inherited instead of restated. Lands with the access-prop commit (see the `MenuItem.tsx` thread).

themes/theme-rui/src/theme.css
aromko: Fixed in 6e2a1b06c — URL-encoded, with a comment naming the sources. They are lucide `lock-keyhole` and `key` (decoded the base64 payloads to confirm the replacement is byte-identical in shape).

aromko:
Status after this review round:

**Pushed** (8c6cf3bd8, 6e2a1b06c):
- Shared `ui-access-master` / `ui-access-admin` utilities in `ui.css`, including CSS alternative text so the restriction lands in the accessible name (a11y blocker)
- `Badge` styles on the utilities; `Badge.test` with AAA blank lines and semantic class assertions
- URL-encoded, commented mask SVGs in `theme.css` (sources: lucide `lock-keyhole` / `key`)
- Unused `icons/Lock.tsx` removed

**Ready locally, pushed once we align on the API thread** (https://github.com/marigold-ui/marigold/pull/5595#discussion_r3513131103):
- Dedicated `access?: 'master' | 'admin'` prop on `Link` / `MenuItem` (+ `useClassNames` forwarding in `@marigold/system`), so access composes with `variant` (destructive master actions become representable)
- `Link`/`Menu` styles switched to the shared utilities (also fixes the `[&_svg]` drift)
- `Link` tests rewritten on the `AccessVariants` story with AAA blank lines
- Docs: new "Inline actions and menus" section + demo on the Admin- and Mastermark pattern page
- Changeset rewrite (adds `@marigold/system` minor, drops the deferred-a11y sentence)

**Base branch**: retargeted to `beta-release` as suggested.
**VRT**: `/run-chromatic` will be triggered once all comments are resolved, so Chromatic snapshots the final state exactly once.

packages/components/src/Menu/MenuItem.tsx
sarahgm: i think it makes sense to have it as own property, because you may need destructive and master, but the concern is the other components keep then variant? Or do we also want to switch for consistency, or we need to document this stricter.

aromko: @sebald @sarahgm In my grill-me session this was also topic with access prop. I decided against to keep the API consistent and easy for now. Right now the only component who would use it is Menu right? I did it locally and here the screenshot how it would looks like right now with the minimal changes. Not sure if we want it this way.

Bildschirmfoto 2026-07-06 um 11 32 37

So what do you think?

sarahgm: theoretically we have also Panel which has destructive and master/admin variants (where you can also just use a master badge then within). But I'm also not sure how often this combination will appear. 

So yes, guess it would be also enough to have stricter documentation (that its intended to not have master + destructive in menu, and a demo to show how to solve this)

aromko:
> theoretically we have also Panel which has destructive and master/admin variants (where you can also just use a master badge then within). But I'm also not sure how often this combination will appear.
> 
> So yes, guess it would be also enough to have stricter documentation (that its intended to not have master + destructive in menu, and a demo to show how to solve this)

@sarahgm @sebald So I updated it now in this way and also added some docs. So you can review the changes again.

themes/theme-rui/src/ui.css
```diff
@@ -364,3 +364,51 @@
...
+ * BEM/Twig CSS.
+ *
+ * The `content` alternative text exposes the restriction to assistive
+ * technology — without it, the glyph does not appear in the accessibility
+ * tree and a restricted element is indistinguishable from a regular one
+ * (WCAG 1.1.1).
+ */
+@utility ui-access-master {
+  &::before {
+    content: '' / 'Master';
```
sarahgm: ⚠️ The entire access affordance (visible glyph + AT name) rides on `content: '' / 'Master'` here, and the test suite only runs in Firefox.

Two things worth verifying before this reaches consumers:
- On engines that don't support the `content` alternative-text syntax, the whole `::before` declaration is dropped — so the **glyph disappears**, not just the alt text. Please confirm Safari + Chrome render the lock/key.
- In **forced-colors / Windows High Contrast**, `background-color`+`mask` glyphs and CSS generated-content text are both commonly stripped, which would leave a restricted action with neither signal. A quick NVDA/VoiceOver + forced-colors check would validate the "exposed to assistive technology" claim.

.changeset/access-variants.md
```diff
@@ -0,0 +1,10 @@
+---
+'@marigold/components': minor
+'@marigold/theme-rui': minor
+---
+
+Add `master`/`admin` access variants to `Link` and `MenuItem` that mark actions requiring elevated access rights with a themed glyph (lock = master, key = admin). `Badge` renders its `master`/`admin` variants through the same shared `ui-access-*` utilities.
```
sarahgm: ⚠️ Changeset bodies for ticketed work start with the Conventional-Commits prefix in this repo. Please prepend `feat(DST-1381):`.

```suggestion
feat(DST-1381): add `master`/`admin` access variants to `Link` and `MenuItem` that mark actions requiring elevated access rights with a themed glyph (lock = master, key = admin). `Badge` renders its `master`/`admin` variants through the same shared `ui-access-*` utilities.
```

packages/components/src/Link/Link.stories.tsx
```diff
@@ -75,3 +75,16 @@ export const Inline = meta.story({
     </Text>
   ),
 });
+
+export const AccessVariants = meta.story({
```
sarahgm, line 80: 💡 The `variant` `argTypes` control (near the top of this file) still lists only `['default', 'secondary']`, while `Badge.stories` includes `master`/`admin`. Consider adding them for parity so the control exercises the new variants too. This `AccessVariants` story covers the visual regardless.

themes/theme-rui/src/components/Menu.styles.ts
```diff
@@ -25,6 +25,10 @@ export const Menu: ThemeComponent<'Menu'> = {
         default:
           'text-foreground focus:bg-focus-highlight [&_svg]:text-secondary [&_svg]:opacity-60',
         destructive: 'text-destructive-accent focus:bg-destructive-accent/10',
+        master:
+          'text-foreground focus:bg-focus-highlight [&_svg]:text-secondary [&_svg]:opacity-60 ui-access-master',
```
sarahgm: 💡 The item is a `grid grid-cols-[auto_1fr]`, so the `::before` glyph auto-places into the icon column (works). But real icon items get `gap-x-2` via `[&:has(>svg)]:gap-x-2`, whereas the access glyph relies only on its own `margin-right: 0.25rem` — so a master/admin glyph sits ~4px tighter to the label than an icon would. Minor visual inconsistency.

Also a guard-rail note: a master/admin item that *also* carried an `<svg>` would have both the `::before` and the icon competing for grid cells (no such case in the demos today).

sarahgm (commented):
## Automated Code Review

**PR #5595 — feat(DST-1381): add master/admin access variants to Link, MenuItem, Badge**

### Overview
| Field | Value |
|-------|-------|
| Branch | `mk/dst-1381-access-mark` → `beta-release` |
| Files Changed | 19 (+302 / −34) |
| CI | ✅ All checks pass |
| Visual Regression | Not required (base is `beta-release`) |

### Linked ticket
**DST-1381** (In Review). Implements Phase 2 candidate #1 (variant on `Link`/`MenuItem`) with the prefix-glyph treatment, grouping restricted menu items into sections. Destructive-vs-access resolved via documented single-axis precedence, consistent with the ticket scoping destructive as orthogonal. Good alignment.

### What's good
- **Shape distinction, not just color** — lock (master) vs key (admin) survives for colorblind users; admin finally gets its own glyph.
- **Theme-only glyph** genuinely satisfies the "works in React *and* generated BEM/Twig" constraint.
- Tests updated coherently (`VARIANT_ORDER` indexing preserved; menu `.test` asserts `Löschen` is `ui-access-master` and *not* `text-destructive-accent`).
- Demos follow the existing pattern-demo convention; single-axis rule documented in the MDX and changeset.

### Issues

**Critical:** none — CI green, decisions documented.

**Warnings (should fix):**
1. The accessibility story rests on `content: '' / 'Master'` + a `mask-image` glyph but is verified only in Firefox. On engines lacking `content` alt-text support the whole `::before` declaration drops (glyph disappears, not just the alt text); in forced-colors mode both the glyph and generated-content text are commonly stripped, leaving restricted actions indistinguishable. Please confirm in Safari/Chrome + a real screen reader + forced-colors. (see inline)
2. Changeset body should start with the `feat(DST-1381):` conventional prefix. (see inline)

**Suggestions:**
3. `Link.stories` `argTypes` still lists only `default`/`secondary` (Badge lists master/admin). (see inline)
4. Menu glyph spacing/placement vs icon items in the grid. (see inline)

### Recommendation
**Comment** — no blockers. Main thing to confirm before merge is warning #1 (cross-browser + forced-colors AT behavior). Changeset prefix is a trivial fix.

---

packages/components/src/Menu/Menu.stories.tsx
```diff
@@ -353,6 +353,61 @@ export const MultiSelection = meta.story({
...
+});
+
+AccessSections.test(
+  'renders master and admin access sections',
+  { parameters: { chromatic: { disableSnapshot: false } } },
+  async ({ canvas }) => {
+    await userEvent.click(
+      canvas.getByRole('button', { name: 'Filial-Aktionen' })
+    );
+    expect(await canvas.findByText('Master-Aktionen')).toBeVisible();
```
sarahgm, line 392: blank lines missing for AAA

packages/components/src/patterns/admin-master-mark.stories.tsx
```diff
@@ -91,3 +94,74 @@ export const Form = meta.story({
...
```
sarahgm: ⚠️ **Should fix — duplicate reserved `storybook-root` id + dead wrapper.** The global preview decorator already wraps every story in `OverlayContainerProvider container="storybook-root"`, and Storybook's own canvas root carries that id. So this wrapper is a *second* element with `id="storybook-root"` (invalid HTML), and `getElementById` returns the real ancestor root — the ActionMenu overlay portals there, not into this div. No test imports this story either, so the wrapper is inert. Recommend deleting it:
```suggestion
  render: () => (
```
(dropping the `decorators` block entirely).

themes/theme-rui/src/components/Badge.styles.ts
```diff
@@ -15,9 +15,9 @@ export const Badge: ThemeComponent<'Badge'> = cva({
       info: 'bg-info text-info-foreground',
       error: 'bg-destructive text-destructive-foreground',
       admin:
-        'text-access-admin-foreground border border-access-admin-accent bg-access-admin',
+        'text-access-admin-foreground border border-access-admin-accent bg-access-admin ui-access-admin',
       master:
-        'text-access-master-foreground bg-access-master border border-access-master-accent ',
+        'text-access-master-foreground bg-access-master border border-access-master-accent ui-access-master',
```
sarahgm: ⚠️ **Accessibility — Badge double-announces its label** (a facet of the `content` alt-text approach flagged on `ui.css`). `ui-access-*` injects an unconditional `content: '' / 'Master'` alt text, but a Badge's visible text is often the access level itself — e.g. `<Badge variant="master">Master</Badge>` in the `BadgeVariants` pattern demo. *If* the alt text is exposed to AT (the open question on `ui.css`), a screen reader reads **"Master Master"**; the old decorative `<svg>` had no accessible name, so this is a regression. The prefix is meaningful on Link/MenuItem (label ≠ access level) but redundant on Badge. Also note the alt text is hardcoded English while the demos are German — no per-instance/localization hook once it lives in the theme utility. Consider not applying the AT alt text for the Badge surface.

themes/theme-rui/src/components/Menu.styles.ts
```diff
@@ -25,6 +25,10 @@ export const Menu: ThemeComponent<'Menu'> = {
         default:
           'text-foreground focus:bg-focus-highlight [&_svg]:text-secondary [&_svg]:opacity-60',
         destructive: 'text-destructive-accent focus:bg-destructive-accent/10',
+        master:
```
sarahgm: 💡 **Minor cleanup + latent alignment nit.**
1. `master`/`admin` copy the entire `default` variant string (`text-foreground focus:bg-focus-highlight [&_svg]:text-secondary [&_svg]:opacity-60`) verbatim just to append one utility — it now appears three times and will drift if the base changes. Consider composing a shared base fragment + the access utility.
2. Latent: the `::before` glyph auto-places into the icon column but (unlike `<svg>`, which gets `[&_svg]:row-span-full [&_svg]:self-center`) is centered on the *label row* only. A master/admin item that also has a `slot="description"` would show the glyph aligned to the label rather than centered across the two-row block. No current demo hits this.

packages/components/src/Badge/Badge.test.tsx
```diff
@@ -30,11 +30,16 @@ test('renders the info variant correctly', () => {
...
   render(<Basic.Component />);
+  const master = screen.getAllByText('Status')[VARIANT_ORDER.indexOf('master')];
 
-  // the master and admin variants both render a lock icon
-  const svgs = screen.getAllByTestId('lock-icon');
+  expect(master).toHaveClass('ui-access-master');
```
sarahgm: 💡 **Test coverage nit (judgment call).** The replaced tests (here + Link + Menu) only assert `toHaveClass('ui-access-master')`, so the actual a11y mechanism — the mask glyph rendering and the `content` alt-text exposure the ticket is built around — has no enforcement. If `content: '' / 'Master'` or `--access-mask-lock` were later dropped, these would stay green. Since these run in real Firefox, a `getComputedStyle(el, '::before')` check could close the gap, though asserting the class is a defensible proxy.

docs/content/patterns/layout/admin-master-mark/index.mdx
```diff
@@ -58,6 +58,24 @@ Use this approach when:
 
 <ComponentDemo name="admin-master-mark-table" />
 
+### Inline actions
+
+For actions rendered as inline text links, badges are too heavy. Use the "master" or "admin" variant on `<Link>` instead. It prepends a small glyph in the corresponding access color — a lock for `master`, a key for `admin` — and exposes the restriction to assistive technology, so screen-reader users can distinguish a restricted action from a regular one.
```
sarahgm: 💡 **Low / style (unenforced).** This new prose reintroduces em dashes (also L69, L75, `token-overview/index.mdx` L667, and the changeset), which the recent `docs-punctuation-scrub` changeset removed across the docs. It's a precedent rather than a linted rule (CI passed), so purely a consistency nit — worth a quick rewrite into plain sentences to avoid the next scrub pass undoing it.

sarahgm (commented):
## Automated Code Review — follow-up pass (new findings)

A second pass focused on findings beyond my earlier review. Each was critically re-verified; severity is noted honestly per comment. The headline is the duplicate `storybook-root` wrapper (should fix); the rest sharpen the existing accessibility thread or are minor cleanup/nits.

---

themes/theme-rui/src/ui.css
aromko: Fixed in e63b03f61. Both risks were real: the Playwright Firefox our CI tests in turns out to lack the alternative-text syntax entirely, so before this fix our own test browser dropped the glyph. Each utility now declares a plain `content: ''` first as fallback (the glyph always renders, the alt text is progressive enhancement), and a `forced-colors: active` rule paints the glyph in `CanvasText` so it survives Windows High Contrast. Verified the accessible name in the Chromium accessibility tree against the preview: menu items announce as "Master Verschieben" / "Admin Freigeben". Safari supports the syntax from 17.4; older engines fall back to glyph-only through the same mechanism.

.changeset/access-variants.md
aromko: Applied in e63b03f61.

packages/components/src/Link/Link.stories.tsx
aromko: Added in e63b03f61.

packages/components/src/Menu/Menu.stories.tsx
aromko: Added in e63b03f61.

themes/theme-rui/src/components/Menu.styles.ts
aromko: Fixed in e63b03f61 at the root: the shared utilities no longer own spacing. Link and Badge add `before:mr-1`, and the Menu access variants use the grid's `gap-x-2`, so the glyph gets exactly the same label spacing as an `<svg>` icon. The svg co-occurrence is now a documented rule on the pattern page (restricted items should not carry an additional icon, since the glyph occupies the icon slot).

packages/components/src/patterns/admin-master-mark.stories.tsx
aromko: Deleted in e63b03f61, decorators block dropped entirely as suggested. Your analysis checks out: these pattern stories only render inside Storybook where the real `#storybook-root` is an ancestor, so the wrapper was a duplicate id and never the actual portal target.

themes/theme-rui/src/components/Badge.styles.ts
aromko: Fixed in e63b03f61: a new `ui-access-decorative` utility (declared after the access utilities, so it wins the cascade at equal specificity) resets the alternative text, and Badge's master/admin variants apply it. `<Badge variant="master">Master</Badge>` announces once again. On localization I would push back: "Master" and "Admin" are invariant product terms, the German demos use exactly these words, so the hardcoded alt text on Link/MenuItem is intentional (noted in the ui.css comment).

themes/theme-rui/src/components/Menu.styles.ts
aromko: Both fixed in e63b03f61. (1) `default`/`master`/`admin` now compose a shared `itemBase` const, so the icon treatment cannot drift. (2) The access variants add `before:row-span-full before:self-center`, mirroring the `[&_svg]` placement, so the glyph centers across the label+description block.

packages/components/src/Badge/Badge.test.tsx
aromko: Adopted in e63b03f61, with one adjustment found while implementing: the story tests now assert the rendered mechanism via `getComputedStyle(el, '::before')` (mask image present, content set), but the alt-text string itself is not assertable because the Playwright Firefox lacks the `content` alternative-text syntax altogether. That finding validated the browser-support concern in the ui.css thread and is covered by the new fallback there. The accessible-name prefix was verified in the Chromium accessibility tree ("Master Verschieben"), and a new Badge story test enforces glyph-without-alt for the decorative case.

docs/content/patterns/layout/admin-master-mark/index.mdx
aromko: Rewritten in f14894880: the pattern page sections, the token-overview addition, and the changeset are plain sentences without em dashes now.

themes/theme-rui/src/components/Badge.styles.ts
sarahgm: Thanks, the `ui-access-decorative` reset does fix the double-announce and I confirmed the Badge announces once again.

My worry is the direction, not the fix. This utility only exists to *undo* an accessible name that the base utility adds, and it wins just by being declared last. That works now, but the ordering is load-bearing and no CI test can catch a regression, since our Firefox can't read the `content` alt-text.

The real cause is that `content: '' / 'Master'` does two jobs at once: it paints the glyph and it carries the accessible name. Every workaround here (the fallback, the forced-colors rule, this decorative reset) comes from that coupling.

Cleaner split:

- Theme utility = glyph only. `ui-access-master`/`-admin` render `content: ''` plus mask and color. No alt text, and `ui-access-decorative` goes away.
- Accessible name = the component's job. Link and MenuItem render a `<VisuallyHidden>` label (we already export it), Badge renders nothing extra because its text already says "Master".

That makes double-announce impossible by design, testable in any browser, and localizable.

One question before we go this way: your reason for keeping the name in CSS was Twig parity. Is the generated-output name actually in scope for DST-1381, or is that a separate track? If it is in scope, the Twig macro can emit its own hidden label instead.

Happy to show it as a diff.

aromko: Implemented your split in bb0fef4d9 — the scope check settled it: DST-1381's deliverables are all Marigold-side (components, theme, pattern story, Figma, changeset); the ticket explicitly puts legacy adoption on the migrators' track ("Migrators move off them by adopting the new pattern"). So the generated-output name is a separate track, and when it lands the Twig macro can emit its own hidden label, matching this split.

What changed:

- `ui-access-master`/`-admin` are glyph-only now (`content: ''`, no alternative text); `ui-access-decorative` and the dual-declaration fallback are gone.
- `Link` and `MenuItem` render a visually hidden "Master"/"Admin" label via a shared internal `AccessLabel` (`packages/components/src/utils/AccessLabel.tsx`). It renders *after* the visible text so an item's accessible name — and menu typeahead — still starts with the action itself.
- `Badge` renders nothing extra; its tests now pin `textContent` to exactly the visible label as the double-announce regression guard.
- The accessible name is now asserted in CI: `getAllByRole('link', { name: 'verschieben Master' })` in the unit tests and `getByRole('menuitem', { name: 'Verschieben Master' })` in the Firefox story test — the assertion that wasn't possible with the `content` alt text.

Changeset and docs prose (pattern page, token overview) updated to describe the decorative-glyph + hidden-label mechanism.

aromko: Follow-up in b0c07065e: your Twig question made me re-check the transform script on the consumer side, and I'm dropping that rationale entirely. The generated-output project builds its CSS through a transform that renders its own access icons and never consumed `ui-access-*` — keeping the glyph baked into the theme's variant classes would even have doubled the icon over there on a future theme bump, since the transform `@apply`s the variant classes verbatim.

So the split now goes all the way: `Link`, `MenuItem`, and `Badge` render a decorative Lock/Key `<svg>` (shared `AccessIcon`, colored by the theme via the access foreground tokens — same pattern as `SectionMessage`'s variant icons), and `ui-access-master`/`-admin`, the `--access-mask-*` tokens, the forced-colors override, and the `::before` spacing workarounds are all gone. The hidden `AccessLabel` announcement from the previous commit is unchanged, and icon rendering is now asserted in story tests.

sarahgm (approved):

PR #5596, opened 2026-07-02
refactor(DST-1534): build Calendar year dropdown on RAC CalendarYearPicker

sarahgm:
# Description

The `Calendar` year dropdown (`YearListBox`) now consumes react-aria's `CalendarYearPicker` render-prop, mirroring how the month dropdown already uses `CalendarMonthPicker`. This replaces the hand-rolled year list, the local `useDateFormatter`, and the inline `Intl.DisplayNames` `aria-label` workaround.

The custom code existed only because RAC treated `maxValue` as **exclusive**, dropping the max year so it was unreachable from the dropdown. react-aria's June 2026 fix ([#10149](https://github.com/adobe/react-spectrum/pull/10149)) made it inclusive, and the repo already resolves the fixed versions (`react-aria@3.50.0` / `react-aria-components@1.19.0`), so the workaround's reason to exist is gone.

`CalendarYearPicker` centers a fixed `visibleYears` window on the focused year and clamps it to the bounds, so `visibleYears` is sized to reach whichever bound is farther. Behavior:
- **Unbounded** — keeps the focused-year ±20 window (41 years, unchanged).
- **Both bounds** — the full in-range list stays reachable at both ends (validated for a wide 1900–2100 range).
- **One-sided** — the open side now widens to keep the single bound reachable, instead of staying at a fixed ±20 years (the one intentional behavior change).

Closes DST-1534

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm sb` and open `Components/Calendar`.
2. Open the year dropdown on an unbounded calendar — confirm the focused year is centered with ~±20 years around it.
3. On a calendar with `minValue={1900}` / `maxValue={2100}`, open the year dropdown and scroll — confirm both `1900` and `2100` are present and selectable.
4. Select a year and confirm the calendar navigates to it (incl. Feb 29 → non-leap year clamps to Feb 28).
5. `pnpm test:sb --run Calendar.stories` and `pnpm test:unit --run Calendar.test` both pass.

## Breaking Changes

No

## Checklist

### DIFF
--- a/packages/components/src/Calendar/YearListBox.tsx
+++ b/packages/components/src/Calendar/YearListBox.tsx
@@ -1,5 +1,5 @@
 import { Dispatch, SetStateAction } from 'react';
-import { CalendarYearPicker } from 'react-aria-components/Calendar';
+import { CalendarYearPicker } from 'react-aria-components';
 import { useCalendarOrRangeState } from './Context';
 import { ListBox } from './ListBox';
 
--- a/packages/components/src/Calendar/YearListBox.tsx
+++ b/packages/components/src/Calendar/YearListBox.tsx
@@ -1,5 +1,5 @@
 import { Dispatch, SetStateAction } from 'react';
-import { useDateFormatter, useLocale } from '@react-aria/i18n';
+import { CalendarYearPicker } from 'react-aria-components';
 import { useCalendarOrRangeState } from './Context';
 import { ListBox } from './ListBox';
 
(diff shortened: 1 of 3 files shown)


packages/components/src/Calendar/YearListBox.tsx
```diff
@@ -1,5 +1,5 @@
 import { Dispatch, SetStateAction } from 'react';
-import { useDateFormatter, useLocale } from '@react-aria/i18n';
+import { CalendarYearPicker } from 'react-aria-components';
```
sebald: Nit: import source is inconsistent with `MonthListBox.tsx`, which imports `CalendarMonthPicker` from the subpath `'react-aria-components/Calendar'`. Both resolve, but the sibling files should match — prefer the subpath here too for consistency and tree-shaking:

```suggestion
import { CalendarYearPicker } from 'react-aria-components/Calendar';
```

packages/components/src/Calendar/YearListBox.tsx
```diff
@@ -9,53 +9,32 @@ interface YearDropdownProps {
 
 const YearListBox = ({ setSelectedDropdown }: YearDropdownProps) => {
   const state = useCalendarOrRangeState();
-  const focusedYear = state.focusedDate.year;
+  const focused = state.focusedDate.year;
```
sebald: Nit: `focused` is a *year number*, but the name reads like a date (or a boolean). The old `focusedYear` was clearer, and `MonthListBox` keeps the analogous `currentYear`. Suggest keeping `focusedYear`.

packages/components/src/Calendar/YearListBox.tsx
```diff
@@ -9,53 +9,32 @@ interface YearDropdownProps {
...
-    new Intl.DisplayNames(locale, { type: 'dateTimeField' }).of('year') ??
-    'year';
+  // `CalendarYearPicker` centers a fixed `visibleYears` window on the focused
+  // year and clamps it to the bounds, so size the window to reach whichever
+  // bound is farther. Unbounded sides keep the focused-year ±20 window. A wide
+  // both-bounds range still renders every in-range year, so virtualizing the
+  // listbox is the path forward if the DOM size ever becomes a concern.
+  const min = state.minValue ? state.minValue.year : focused - 20;
+  const max = state.maxValue ? state.maxValue.year : focused + 20;
+  const visibleYears = 2 * Math.max(focused - min, max - focused) + 1;
```
sebald: Confirming this is intended (I believe it is): a one-sided bound far from the focused year now renders a much larger symmetric list — e.g. `minValue=1900` with no max and focus in 2026 yields ~250 options vs. the old ~146. This is unavoidable with `CalendarYearPicker`'s centered window (widening symmetrically is the only way to keep a far bound reachable), and the comment already flags virtualization as the follow-up, so no change requested. Semver `patch` is fine since there's no API change.

sebald (commented):
Reviewed and verified the `visibleYears` clamping against the react-aria `useCalendarYearPicker` source (v3.50.0) — the reachability logic is correct for unbounded, both-bounds, and far one-sided-bound configs, and the `id`-is-array-index contract lines up with `ListBox`. No new effects; render-prop consumption is idiomatic. Approving in spirit — just a couple of consistency nits and one behavior note inline.

packages/components/src/Calendar/YearListBox.tsx
sarahgm: Checked this — in `src/` every Calendar file imports from `'react-aria-components'`, including `MonthListBox.tsx:2`. The `react-aria-components/Calendar` subpath only appears in `dist/`, where the bundler rewrites imports for tree-shaking. So `YearListBox` already matches `MonthListBox` here; switching to the subpath would make it the only source file that diverges. Leaving as-is — let me know if you'd rather migrate all of them separately.

packages/components/src/Calendar/YearListBox.tsx
sarahgm: Confirmed intended. The centered window has to widen symmetrically to keep a far one-sided bound reachable, and virtualization is the tracked follow-up. Thanks for verifying the clamping against the v3.50.0 source.

packages/components/src/Calendar/YearListBox.tsx
sarahgm: Done in 53ebbf94d — renamed `focused` back to `focusedYear`.

packages/components/src/Calendar/YearListBox.tsx
sebald: Yes, lets migrate all, otherwise treeshaking is not working correctly.

sarahgm: Heads up: applying the subpath suggestion broke the Unit Tests in CI (Storybook Tests and Typecheck stayed green). Reverted to the barrel import here.

Root cause: `react-aria-components/Calendar` never appears anywhere else in main's `src/` — on main `MonthListBox.tsx:2` imports `CalendarMonthPicker` from the barrel too. So YearListBox introducing the subpath makes it a brand-new specifier that Vite's up-front dep scan (`browserScanEntries` in `vitest.config.shared.ts`) never pre-bundles. In the full sharded CI run it gets discovered and optimized mid-run:

```
new dependencies optimized: react-aria-components/Calendar
Vite unexpectedly reloaded a test.
TypeError: can't access property "useContext", dispatcher is null
Vitest caught 76 unhandled errors  (ActionBar.test.tsx, etc.)
```

The reload pulls a second React copy, nulls the dispatcher, and cascades into 76 unhandled errors across unrelated test files. It passed locally only because filtering to `Calendar` let the optimizer catch it in the initial scan.

On main, the barrel import is also the one that matches its sibling `MonthListBox`, and `CalendarYearPicker` is exported from the top-level barrel, so nothing is lost for tree-shaking (the build rewrites `src/` barrel imports to subpaths in `dist/`).

Note for beta-release: there the subpath IS the right form — beta's `MonthListBox.tsx` already imports from `react-aria-components/Calendar`, so (a) the subpath matches its sibling there, and (b) that exact specifier is already pre-bundled via MonthListBox, so it won't trigger the mid-run re-optimization. Same RAC version on both branches (`^1.18.0`), so this is purely an import-convention difference, not a version one. This change hasn't landed on beta yet (it still has the hand-rolled YearListBox), so retargeting to beta means using the subpath there.

sebald (approved):
Verified the import-source point resolves correctly: tsdown (`unbundle: true`) rewrites the source barrel import to the `react-aria-components/Calendar` subpath in dist (confirmed in the built `MonthListBox.mjs`), so tree-shaking in the published package is preserved with the barrel import in source — and it matches the sibling `MonthListBox` while avoiding the CI re-optimization breakage. `focusedYear` rename and the behavior note are addressed. LGTM 👍

PR #5598, opened 2026-07-02
feat(DST-1568): add error and errorMessage support to Switch and Checkbox

sarahgm:
# Description

`Switch` had no help-text support and `Checkbox` had `error` styling but no way to surface an error message. This PR adds `error` and `errorMessage` to both. When `error` is set, the field is marked invalid and `errorMessage` is shown in place of the `description`, wired to the input via `aria-describedby`. When both are unset, rendering is unchanged.

Internally, `Switch` and `Checkbox` now render their `HelpText` inside the RAC field (`SwitchField` / `CheckboxField`) via a generalized `BooleanField as={...}`, relying on RAC's native `aria-describedby` + `FieldErrorContext` wiring instead of the previous hand-rolled `Provider` + `useId` plumbing. This adopts the shared-wrapper alternative noted in the ticket rather than duplicating Checkbox's bespoke field, keeping both primitives on one path. No other public API change.

Closes DST-1568

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Open Storybook, go to `Components/Switch` and `Components/Checkbox`, and open the `WithError` story for each.
2. Confirm the error message renders in place of the description and shows the alert icon.
3. Inspect the input and confirm `aria-describedby` points at the error message's element.
4. Toggle `error` off (or clear it) and confirm the description reappears and no `aria-describedby` is set when neither description nor error is present.
5. `pnpm test:unit --run packages/components/src/Switch packages/components/src/Checkbox` and `pnpm typecheck:only` both pass.

## Breaking Changes

No

## Checklist


packages/components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -138,3 +144,27 @@ export const WithDescription = meta.story({
...
+    const errorMessage = canvas.getByText('This selection is required');
+
+    // The error message replaces the description when `error` is set.
+    expect(canvas.queryByText('This is a description')).not.toBeInTheDocument();
+
+    const helpTextId = errorMessage.closest('[id]')?.getAttribute('id');
+    const checkboxDescribedBy = checkbox.getAttribute('aria-describedby');
+
+    expect(errorMessage).toBeInTheDocument();
+    expect(checkboxDescribedBy).toBe(helpTextId);
```
sebald: The id-plumbing here (`closest('[id]')` + `toBe(helpTextId)`) is brittle — `aria-describedby` may legitimately hold multiple ids — and asserts id equality rather than the accessible outcome. The `toHaveAccessibleDescription` matcher (already used by this PR's own unit tests) does the same in one line:

```ts
const checkbox = await canvas.findByRole('checkbox');

// The error message replaces the description when `error` is set.
expect(canvas.queryByText('This is a description')).not.toBeInTheDocument();
await expect(checkbox).toHaveAccessibleDescription('This selection is required');
```

This also drops the redundant `expect(errorMessage).toBeInTheDocument()` — `getByText` already throws when the element is absent.

Same pattern in `Switch.stories.tsx` → `WithError`. Caveat: the existing `WithDescription` plays use the `closest('[id]')` approach, so either update all four for consistency or leave as-is — just flagging the simpler option.

packages/components/src/FieldBase/BooleanField.tsx
```diff
@@ -1,61 +1,84 @@
...
   description,
+  errorMessage,
+  error,
   variant,
-  context,
+  size,
   children,
-}: PropsWithChildren<BooleanFieldProps>) => {
+  ...rest
+}: BooleanFieldProps<T> & DistributiveOmit<ComponentPropsWithRef<T>, 'as'>) => {
```
sebald: The rest-prop type only omits `'as'`, but the component also owns `isInvalid` and `className` — a caller passing either gets no type error and no effect (both are overwritten in `fieldProps` below). Omitting `'as' | 'isInvalid' | 'className'` makes that contract explicit.

Note `FieldBase.tsx` has the same looseness with its `'as'`-only omit, so the alternative is keeping this consistent for now and aligning both in a small follow-up — either is fine, just worth a deliberate choice.

packages/components/src/FieldBase/BooleanField.tsx
```diff
@@ -1,61 +1,84 @@
...
+  // field's native `aria-describedby` + `FieldErrorContext` wiring. Without it,
+  // the field stays transparent (`display: contents`) and the toggle keeps its
+  // own layout.
+  const fieldProps = {
+    ...rest,
+    isInvalid: error,
+    className: hasHelpText
+      ? cn('group/field group/booleanfield', classNames.container)
+      : 'contents',
+    ...(hasHelpText ? { 'data-booleanfield': true } : {}),
```
sebald: Micro-nit: the conditional spread can fold into the object literal —

```ts
'data-booleanfield': hasHelpText || undefined,
```

`undefined` keeps the attribute off the DOM, same as omitting the key.

packages/components/src/Switch/Switch.tsx
```diff
@@ -78,35 +92,35 @@ const _Switch = ({
     ...rest,
   } satisfies RAC.SwitchFieldProps;
   return (
+    // `SwitchField` (rendered via `BooleanField as`) natively wires the label,
+    // `aria-describedby`, and validation state; the `SwitchButton` (the rendered
+    // `label`) carries the width, matching the standalone layout.
```
sebald: This comment (and its twin in `Checkbox.tsx`) restates what `BooleanField`'s `as` prop JSDoc already documents — "the field natively provides the `aria-describedby`, `FieldErrorContext`, and description/error text slots". One source of truth in `BooleanField` is enough; the consumer-side copies will drift when the wrapper changes. The width note in the second half is the only part that's Switch-specific and worth keeping.

.changeset/dst-1568-switch-error-support.md
```diff
@@ -0,0 +1,9 @@
+---
+'@marigold/components': minor
+---
+
+feat(DST-1568): add `error` and `errorMessage` support to `Switch` and `Checkbox`
```
sebald, line 5: Changeset bodies end up as changelog text, and the other changesets in the repo start with plain prose — the `feat(DST-1568):` conventional-commit prefix belongs to the PR title, not here. Suggest starting directly with "Add `error` and `errorMessage` support…".

sebald (commented):
Simplification-focused review (reuse/consolidation/altitude — not a bug hunt). Overall this is a nice net simplification: the `BooleanField` rework deletes the hand-rolled `Provider`/`useId` plumbing and mirrors `FieldBase`'s established `as` + `DistributiveOmit` pattern. Five small inline notes, in priority order:

1. `WithError` play functions: `toHaveAccessibleDescription` replaces the `closest('[id]')` id-plumbing (Checkbox.stories.tsx, same in Switch.stories.tsx)
2. `BooleanField`'s rest-prop omit could also exclude `isInvalid`/`className`, which the component overwrites (BooleanField.tsx)
3. Fold the `data-booleanfield` conditional spread into the object literal (BooleanField.tsx)
4. Consumer-side comments in Switch/Checkbox restate `BooleanField`'s `as` JSDoc (Switch.tsx)
5. Drop the `feat(DST-1568):` prefix from the changeset body (changelog-facing text)

Notes without a diff location:

- **Scope:** DST-1568 explicitly listed "refactoring Checkbox onto the shared wrapper" as *not included* and recommended mirroring Checkbox's bespoke pattern — this PR did the harmonization anyway. The result is genuinely simpler and the PR body is transparent about it, but it deserves explicit sign-off, and if a separate harmonization task exists it should be closed/updated so the work isn't double-planned.
- **Checklist:** the a11y box is unchecked although this PR is precisely an `aria-describedby` feature (tests cover the wiring — probably just needs ticking), and VRT is unchecked while the ticket scope includes "VRT for the new states".

packages/components/src/Checkbox/Checkbox.stories.tsx
sarahgm: Done in 941546084. Switched both `WithError` plays to `toHaveAccessibleDescription`, and updated the `WithDescription` plays too so all four are consistent. Dropped the redundant `toBeInTheDocument` on the error message.

packages/components/src/FieldBase/BooleanField.tsx
sarahgm: Done in 941546084. `DistributiveOmit` now excludes `'as' | 'isInvalid' | 'className'`. Left `FieldBase.tsx`'s parallel `'as'`-only omit alone as a separate small follow-up.

packages/components/src/FieldBase/BooleanField.tsx
sarahgm: Done in 941546084 — folded into `'data-booleanfield': hasHelpText || undefined`.

packages/components/src/Switch/Switch.tsx
sarahgm: Done in 941546084. Trimmed the restated `as` JSDoc from both consumers — kept only the Switch-specific width note here and removed the Checkbox comment entirely.

.changeset/dst-1568-switch-error-support.md
sarahgm: Leaving this one as-is: the repo convention is actually to start changeset bodies with the conventional-commit + ticket prefix. Recent examples: `refactor(DST-1585): drive SegmentedControl's reduced-motion scroll from CSS`, `fix(DST-1559): remove dead and mis-named props...`, `chore(DST-1503): Migrate Checkbox, Switch, and Radio...`. So `feat(DST-1568):` here matches existing practice rather than diverging from it.

sebald: Fair enough — you're right that the recent changesets lead with the `type(DST-XXXX):` prefix, so this matches convention rather than diverging from it. Thanks for pulling the examples. Resolving.

sebald (approved):
LGTM — all review threads addressed, CI green. Note: the branch currently has merge conflicts with the base, so it'll need a rebase/merge from main before it can be merged.

sebald:
@sarahgm approved and CI is green ✅ — just one thing: the branch has merge conflicts with `main` right now, so it can't be merged yet. Could you rebase / merge `main` in? Once it's back to mergeable, feel free to merge it yourself. 🙂

PR #5599, opened 2026-07-03
refactor(DST-1548): rename Card.Body to Card.Content

sarahgm:
# Description

Renames the Card body sub-component from `Card.Body`/`CardBody` to `Card.Content`/`CardContent` so all three container primitives (`Card`, `Panel`, `Page`) expose the main body region under one name. `Panel.Content` and `Page.Content` already use `.Content`; Card was the lone outlier. Folded into the `18.0.0` major since Card already ships breaking changes there.

The `bleed` prop and padding behavior are unchanged. The internal `data-card-body` attribute is removed to match `Panel.Content`, and the `theme-rui` Card theme slot key is renamed `body` → `content`.

Closes DST-1548

## Test Instructions

1. Open Storybook → `Components/Card`, exercise each story (Basic, WithBleedContent, etc.).
2. Run `pnpm test:unit --run packages/components/src/Card/Card.test.tsx` — 27 tests pass.
3. Run `pnpm typecheck:only` to confirm no dangling `Card.Body` references.

## Breaking Changes

Yes. `Card.Body` (`CardBody`) is removed — rename usages to `Card.Content`:

```diff
- <Card.Body>...</Card.Body>
+ <Card.Content>...</Card.Content>
```

Theme authors overriding the Card `body` slot must rename the key to `content`.

## Checklist

- [ ] Visual regression tests updated (for UI changes) — N/A, PR targets `beta-release` and no visual change


packages/components/src/Card/Card.test.tsx
```diff
@@ -164,24 +164,22 @@ describe('Card', () => {
...
+    test('Card.Content applies horizontal padding by default', () => {
       render(<Basic.Component />);
 
-      const body = screen
+      const content = screen
         .getByText(/was an English/)
-        .closest('[data-card-body]');
-      expect(body).not.toBeNull();
-      expect(body!.className).toContain('px-(--card-px)');
+        .closest('div[class*="px-(--card-px)"]');
```
OsamaAbdellateef: ⚠️ **Test selector is circular**

This selects by the exact class token injected into the component — if the token is ever renamed, the selector quietly returns `null` and `not.toBeNull()` is the only safety net. Selecting by class internals is fragile.

Consider adding `data-testid="card-content"` to the `CardContent` div for a stable test anchor, then using `screen.getByTestId('card-content')` here.

packages/components/src/Card/Card.test.tsx
```diff
@@ -164,24 +164,22 @@ describe('Card', () => {
...
-      render(<WithBleedBody.Component />);
+    test('Card.Content with bleed opts out of horizontal padding', () => {
+      render(<WithBleedContent.Component />);
 
-      const body = screen
-        .getByText(/Edge-to-edge banner/)
-        .closest('[data-card-body]');
-      expect(body).not.toBeNull();
-      expect(body!.className).not.toContain('px-(--card-px)');
+      const content = screen.getByText(/Edge-to-edge banner/).parentElement;
```
OsamaAbdellateef, line 167: ⚠️ **Bleed test uses `.parentElement` — fragile DOM traversal**

`.parentElement` will silently target the wrong element if `CardContent`'s children gain a wrapper. The old `[data-card-body]` selector was precise; removing the attribute made this test harder to anchor.

If `data-*` attributes are intentionally dropped to match `Panel.Content`, consider using `.closest('div')` or a `data-testid` as a stable anchor.

```suggestion
      const content = screen.getByText(/Edge-to-edge banner/).closest('div');
```

OsamaAbdellateef (changes requested):

OsamaAbdellateef:
💡 ** — verify symmetry with `Panel.Content`**

`data-card-body` was removed here to align with `Panel.Content` (which has no equivalent `data-*` attribute). Worth a quick check that `Panel.Content` is also attribute-free — if it does have one, adding `data-card-content` here would be consistent rather than divergent. If it doesn't, the removal is correct as-is.

packages/components/src/Card/Card.test.tsx
sarahgm: Kept `.parentElement` here on purpose. The `.closest('div')` suggestion would actually target the wrong element: the bleed banner in the story is itself a `<div>` (`Card.stories.tsx:263`), and `closest()` starts at the element itself, so `getByText(/Edge-to-edge banner/).closest('div')` returns the banner, not the `CardContent` wrapper. The assertion `not.toContain('px-(--card-px)')` would then pass trivially against the inner banner (which never had the padding) instead of verifying the wrapper opts out.

`.parentElement` correctly reaches the `CardContent` wrapper and mirrors the sibling bleed test in `Panel.test.tsx:333` (`table.parentElement!`), which is the established pattern for this component family.

sarahgm:
Re: symmetry with `Panel.Content` — confirmed and correct as-is. `PanelContent.tsx:24` renders `<div className={cn(!bleed && 'px-(--panel-px)', classNames.content)}>` with no `data-*` attribute. Only `Panel.Header`/`Panel.Footer` carry `data-panel-header`/`data-panel-footer`.

Card now matches that shape exactly: `Card.Header` → `data-card-header`, `Card.Footer` → `data-card-footer`, `Card.Content` → attribute-free. So dropping `data-card-body` is the symmetric choice, not a divergent one.

packages/components/src/Card/Card.test.tsx
sarahgm: Fixed in the latest commit. Switched to a structural anchor plus a direct className assertion, so it no longer selects by the class it is verifying:

```js
const content = screen.getByText(/was an English/).parentElement;

expect(content!.className).toContain('px-(--card-px)');
```

Kept it attribute-free deliberately: `<Text>` renders a `div` by default, and it is the direct child of `CardContent`, so `.parentElement` lands on the content wrapper. Adding `data-testid` would also cut against CLAUDE.md ("Avoid `getByTestId`, prefer accessible queries") and against the `Panel.Content` symmetry this PR is aligning to (`Panel.Content` has no `data-*`/testid anchor either).

.changeset/dst-1548-card-content-rename.md
```diff
@@ -0,0 +1,19 @@
+---
+'@marigold/components': major
+'@marigold/theme-rui': patch
```
aromko, line 4: ⚠️ **Missing `@marigold/system` in changeset**

This PR changes `@marigold/system`'s public `Theme` type (`packages/system/src/types/theme.ts` — Card slot key `body` → `content`), but the changeset doesn't declare the package. It currently ships correctly only because other changesets in the beta train already major-bump system — and the rename won't show up in `@marigold/system`'s changelog, where theme authors would look for it.

```suggestion
'@marigold/system': major
'@marigold/theme-rui': patch
```

docs/content/releases/blog/release-2026-06-30.mdx
```diff
@@ -345,9 +345,9 @@ The discriminated union shape is unchanged: `p` is mutually exclusive with `px`
 +     <Title>Payment method</Title>
 +     <Description>Visa ending in 4242</Description>
 +   </Card.Header>
-+   <Card.Body>
++   <Card.Content>
```
aromko, line 348: 💡 **Retroactive edit of a dated release post**

Rewriting this (and the `docs/CHANGELOG.md` entry) to `Card.Content` is the right call while 18.0.0 is in beta pre-mode — `Card.Body` will never exist in a stable release. Just double-check this post isn't already live on production docs; if it is, readers who saw `Card.Body` get a silent history edit rather than a migration note.

.changeset/dst-1548-card-content-rename.md
sarahgm: Fixed in 74e23f69f. Added `'@marigold/system': major` to the changeset and reworded the body to name the `@marigold/system` `Theme` type change explicitly, so the `body` → `content` slot rename now surfaces in that package's own changelog rather than only riding along on the beta train's other system bumps. Good catch.

docs/content/releases/blog/release-2026-06-30.mdx
sarahgm: Checked: `release-2026-06-30.mdx` is not on `main` (`git ls-tree main` returns empty), so it only exists on the beta-release train and is not live on production docs. No reader has seen `Card.Body`, so the retroactive edit is safe rather than a silent history change. Keeping it as-is, in line with your read of the pre-mode situation.

aromko (approved):

OsamaAbdellateef (approved):

PR #5615, opened 2026-07-03
feat(DST-1545): replace ActionBar.Button with Button via ButtonContext cascade

sarahgm:
# Description

`ActionBar.Button` is removed in favor of a plain `<Button>`. `ActionBar` now wraps its `Toolbar` in a `Provider` that cascades a `ghost`/`default` `ButtonContext`, so authors place a standard `<Button>` inside the bar and it adapts to the toolbar look automatically, with the full Button API available (`disabled`, `loading`, `slot`, `size="icon"`). A local `variant`/`size` on a button still wins. This mirrors the pattern already used by `Panel.Header` and `ButtonGroup`, and removes the duplicated `actionButton` theme style that hand-mirrored Button's `ghost` variant minus its press/loading affordances.

It also fixes an accessibility defect: the old wrapper only accepted `onPress`/`children` and silently dropped `aria-label`, shipping unlabeled icon-only buttons. Icon-only actions now use `<Button size="icon" aria-label="…">`.

Closes DST-1545

## Screenshots / Preview

Static appearance is unchanged. The `Basic` story renders byte-for-byte identically before and after (captured from the `beta-release` and this branch's deployed Storybook). This PR swaps the hand-mirrored `actionButton` style for Button's real `ghost` variant, and the only visible additions are press and loading affordances, which a static shot cannot show.

| Before (beta-release) | After (this branch) |
| --------------------- | ------------------- |
| image     | image    |

Given the confirmed static parity, VRT is not needed to gate this PR.

## Test Instructions

1. Open Storybook → `Components/ActionBar` and exercise the `Basic`, `WithoutClearButton`, `IntegratedWithTable`, and `NoSelection` stories.
2. Confirm actions render with the ghost/default toolbar look and that icons are sized correctly.
3. On `NoSelection`, verify each icon-only button exposes its `aria-label` (`Edit`, `Copy`, `Delete`) to the accessibility tree.
4. Run `pnpm typecheck:only` and `pnpm test:unit` — the new `keeps aria-label on icon-only action buttons` test should pass.

## Breaking Changes

Yes. `ActionBar.Button` is removed. Replace `<ActionBar.Button>…</ActionBar.Button>` with `<Button>…</Button>`; for icon-only actions use `<Button size="icon" aria-label="…">`. The `actionButton` theme style key is also removed from `theme-rui` and the `Theme` type. Changeset bumps `@marigold/components` and `@marigold/theme-rui` as `major`.

## Checklist


packages/components/src/ActionBar/ActionBar.test.tsx
```diff
@@ -81,6 +81,14 @@ test('has proper accessibility attributes', () => {
   expect(toolbar).toBeInTheDocument();
 });
 
+test('keeps aria-label on icon-only action buttons', () => {
```
aromko, line 84: 💡 **Suggestion:** Nice test for the `aria-label` fix! Consider one more test asserting that a local `variant`/`size` on a `<Button>` inside the bar wins over the cascade. That override is now a documented contract (the ActionBar docs and the changeset both promise it), so a test — e.g. a story with a `variant="primary"` action, asserted via its rendered classes — would guard it against future `ButtonContext` changes.

packages/components/src/ActionBar/ActionBar.tsx
```diff
@@ -12,16 +13,20 @@ import {
...
 import { IconButton } from '../IconButton/IconButton';
 import { X } from '../icons/X';
 import { intlMessages } from '../intl/messages';
 import { MotionFeatures } from '../lazyMotion';
-import { ActionBarButton } from './ActionBarButton';
 import { useActionBarContext } from './ActionBarContext';
 
+// Cascade a ghost/default look onto plain `<Button>` children; a local
+// `variant`/`size` still wins, so an icon-only action opts into `size="icon"`.
+const actionButtonContext = { variant: 'ghost', size: 'default' } as const;
```
aromko, line 25: 💡 **Suggestion:** Since this PR targets `beta-release`, Chromatic won't gate it here, and this change swaps the hand-mirrored `actionButton` style for Button's real `ghost` variant (which adds press/loading affordances). Consider attaching a before/after screenshot to the PR description, or running VRT once `beta-release` heads toward `main`, to document the visual-parity claim.

packages/components/src/ActionBar/ActionBar.test.tsx
sarahgm: Added in d39b249: a `WithButtonVariantOverride` story (ghost + `variant="primary"`) plus a unit test `local variant on a Button wins over the toolbar cascade`. The assertion is token-free. It compares the primary action's rendered classes to a ghost sibling, so if the `ButtonContext` cascade ever overrode a local `variant` both would collapse to ghost and the test fails. Kept it as a single unit test (mirroring the aria-label test), with the story demoted to a plain visual fixture, to avoid duplicating the same assertion in a `play`.

packages/components/src/ActionBar/ActionBar.tsx
sarahgm: Good call. I'm attaching before/after screenshots to the PR description to document the visual-parity claim, so we don't need to gate on VRT for this beta-release PR.

aromko (approved):

PR #5616, opened 2026-07-04
test(DST-1592): convert remaining beta-branch stories to .test() syntax

aromko:
# Description

Adjusts the component stories on `beta-release` that still used the old `play`-function style so they use the `.test()` API, matching the DST-1562 story cleanup already on `main`. This resolves the divergence called out in the ticket and gets the beta branch's typecheck and test suites fully green.

Along the way it also fixes the pre-existing failures those suites were already carrying (unrelated to the conversion but blocking a clean run):

- **Converted `play` → `.test()`**: ButtonGroup, Description, Dialog, Drawer, Inline, Inset, ListBox, SectionMessage, SegmentedControl, Select, SelectList, Tabs, TagGroup, TextValue, Title, ToggleButton, Tooltip, Tray (plus Card, Icons, LinkButton, TagField, ToggleButtonGroup, TooltipTrigger).
- **Restored missing story fixtures** imported by unit tests: `Switch.Settings`, `ContextualHelp.WithDescription`.
- **Fixed stale/contradictory tests**: the ContextualHelp title assertion, the Button "add icon" test, and a contradictory Card `aria-label` unit test.
- **Relocated one Select unit test**: `default trigger render hides description slot` imported the `Sections` story fixture, which no longer exists as an export on `beta-release` (it was folded into a `Basic.test()` before this PR). The assertion now lives in the `Opens the list grouped into sections` story test, which selects a default value and verifies the trigger shows the `TextValue` while its description slot stays hidden.
- **Guarded intentional invalid-prop tests** (`width="fit"`) with `@ts-expect-error`.
- **Removed dead imports** (Button, Menu, Stack) and the **orphan empty `Multiselect.stories.tsx`** stub (the component was removed in DST-1283).

Result: `pnpm typecheck:only` → 0 errors · full unit suite → 1350 passed · full Storybook suite → 378 passed.

Closes DST-1592

## Visual regression / Chromatic

No component or theme source changed — this is a test/story refactor, not a visual redesign. Chromatic will show snapshot-count deltas because a few purely-visual stories were collapsed into `.test()` cases and some interaction tests toggle `disableSnapshot`. These are expected, not visual regressions. The Select `Opens the list grouped into sections` snapshot additionally changes because the trigger now renders a default selection ("Harry Potter") instead of the placeholder.

## Test Instructions

1. `pnpm install`
2. `pnpm typecheck:only` — expect 0 errors
3. `pnpm test:unit --run` — expect all passing
4. `pnpm test:sb --run` — expect all passing

## Breaking Changes

No

## Checklist

- [ ] Changeset added (`pnpm changeset`) — N/A, story/test-only (nothing published changes)

### DIFF
--- a/packages/components/src/Icons.stories.tsx
+++ b/packages/components/src/Icons.stories.tsx
@@ -41,7 +41,7 @@ const meta = preview.meta({
   title: 'Icons/Icon',
   component: DesignTicket,
   parameters: {
-    parameters: { chromatic: { disableSnapshot: true } },
+    chromatic: { disableSnapshot: true },
     surface: false,
   },
   argTypes: {
--- a/packages/components/src/Icons.stories.tsx
+++ b/packages/components/src/Icons.stories.tsx
@@ -41,6 +41,7 @@ const meta = preview.meta({
   title: 'Icons/Icon',
   component: DesignTicket,
   parameters: {
+    parameters: { chromatic: { disableSnapshot: true } },
     surface: false,
   },
   argTypes: {
--- a/packages/components/src/SectionMessage/SectionMessage.stories.tsx
+++ b/packages/components/src/SectionMessage/SectionMessage.stories.tsx
@@ -1,6 +1,7 @@
 import { expect } from 'storybook/test';
 import preview from '.storybook/preview';
-import { Button, Stack } from '@marigold/components';
+import { Button } from '../Button/Button';
+import { Stack } from '../Stack/Stack';
 import { Text } from '../Text/Text';
 import { SectionMessage } from './SectionMessage';
 
--- a/packages/components/src/SectionMessage/SectionMessage.stories.tsx
+++ b/packages/components/src/SectionMessage/SectionMessage.stories.tsx
@@ -1,5 +1,6 @@
 import { expect } from 'storybook/test';
 import preview from '.storybook/preview';
+import { Button, Stack } from '@marigold/components';
 import { Text } from '../Text/Text';
 import { SectionMessage } from './SectionMessage';
 
@@ -37,6 +38,11 @@ const meta = preview.meta({
 });
 
 export const Basic = meta.story({
+  tags: ['component-test'],
+  parameters: {
+    chromatic: { disableSnapshot: true },
+    surface: false,
+  },
   render: args => (
     <SectionMessage closeButton {...args}>
       <SectionMessage.Title>Danger Zone!</SectionMessage.Title>
@@ -47,7 +53,21 @@ export const Basic = meta.story({
   ),
 });
 
+Basic.test(
+  'renders the Title as a real heading element',
+  {
+    parameters: { chromatic: { disableSnapshot: true } },
+  },
+  async ({ canvas }) => {
+    const heading = canvas.getByRole('heading', { name: 'Danger Zone!' });
+    await expect(heading).toBeInTheDocument();
+    // Default `headingLevel` is 3.
+    await expect(heading.tagName).toBe('H3');
+  }
+);
+
 export const WithDescription = meta.story({
+  tags: ['component-test'],
   render: args => (
     <SectionMessage {...args}>
       <SectionMessage.Title>Backup completed</SectionMessage.Title>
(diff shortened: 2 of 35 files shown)


packages/components/src/Icons.stories.tsx
```diff
@@ -41,6 +41,7 @@ const meta = preview.meta({
   title: 'Icons/Icon',
   component: DesignTicket,
   parameters: {
+    parameters: { chromatic: { disableSnapshot: true } },
```
sarahgm: 🔴 Nested `parameters.parameters` — this lands at `parameters.parameters.chromatic`, which Storybook never reads, so `disableSnapshot: true` has no effect and this story is still snapshotted. Hoist `chromatic` up one level:

```suggestion
    chromatic: { disableSnapshot: true },
```

packages/components/src/ToggleButton/ToggleButtonGroup.stories.tsx
```diff
@@ -120,6 +120,7 @@ Basic.test(
 
 export const BothSurfaces = meta.story({
   parameters: {
```
sarahgm, line 122: 🔴 Same nested-`parameters` bug as in `Icons.stories.tsx` — `chromatic` ends up at `parameters.parameters.chromatic` and is ignored. Fix:

```suggestion
    chromatic: { disableSnapshot: true },
```

packages/components/src/SectionMessage/SectionMessage.stories.tsx
```diff
@@ -1,5 +1,6 @@
 import { expect } from 'storybook/test';
 import preview from '.storybook/preview';
+import { Button, Stack } from '@marigold/components';
```
sarahgm: ⚠️ Barrel import from `@marigold/components` inside the components package itself. Every other import here is relative (e.g. `../Text/Text`), and pulling the package barrel risks a circular dependency. Prefer:

```suggestion
import { Button } from '../Button/Button';
import { Stack } from '../Stack/Stack';
```

packages/components/src/Select/Select.test.tsx
```diff
@@ -2,12 +2,7 @@ import { render, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import type { RefObject } from 'react';
 import { mockMatchMedia, renderWithOverlay } from '../test.utils';
-import {
-  Basic,
-  MultiSelectSummary,
-  Sections,
-  WithRenderValue,
-} from './Select.stories';
+import { Basic, MultiSelectSummary, WithRenderValue } from './Select.stories';
```
sarahgm, line 5: ⚠️ This conversion also drops the `default trigger render hides description slot` test (and the `Sections` import), which isn't called out in the PR body unlike the other removals. Was this intentional, and is the "description slot hidden in the default trigger" behavior still covered elsewhere? Otherwise it's a silent coverage loss.

sarahgm (commented):
## Automated Code Review

Solid, mechanical `play` → `.test()` conversion that matches the DST-1562 pattern on `main`, and the `disableSnapshot` toggling (interaction tests off, one visual-state test back on) is well judged. CI is fully green.

Four inline notes below. The two `parameters.parameters` nestings are unambiguous copy-paste bugs — they typecheck and pass CI only because Storybook types `parameters` as an open record, so the `disableSnapshot: true` is silently ignored. Non-blocking on a `beta-release`, test-only PR, but worth fixing since they defeat their own intent.

---

packages/components/src/Icons.stories.tsx
aromko: Good catch — fixed in 4961fb17d. `chromatic` is hoisted up to the correct level, so the icon showcase stories are now actually excluded from Chromatic snapshots.

packages/components/src/ToggleButton/ToggleButtonGroup.stories.tsx
aromko: Fixed in 4961fb17d with the same hoist — `BothSurfaces` is now really excluded from snapshots.

packages/components/src/SectionMessage/SectionMessage.stories.tsx
aromko: Agreed — swapped for relative imports (`../Button/Button`, `../Stack/Stack`) in 4961fb17d.

packages/components/src/Select/Select.test.tsx
aromko: Thanks for flagging — the drop was forced, but staying silent about it wasn't right. Context: the `Sections` story export no longer exists on `beta-release`; it was folded into `Basic.test('Opens the list grouped into sections')` before this PR, so this unit test was importing a fixture that no longer exists (one of the pre-existing typecheck failures this branch carried). Removing the test fixed the broken import, but you're right that the behavior — the default trigger hiding `[slot=description]` content — was then covered nowhere (the remaining assertion only checks the custom `renderValue` path).

Restored in 4961fb17d: the grouped-sections story test now sets `defaultValue: 'harry-potter'` and asserts (before opening the list) that the trigger shows the `TextValue` while the description slot stays hidden. Since the fixture only exists as a `.test()` child now, the story test (real browser, CSS applied) is the natural home — a unit test can't import the removed `Sections` export, and recreating an inline fixture is against our testing guidelines. Also called this out in the PR description now.

Heads-up for Chromatic: that story's snapshot stays enabled and will show a diff (trigger now renders "Harry Potter" instead of the placeholder) — expected.

sarahgm (approved):

PR #5618, opened 2026-07-06
fix(DST-1556): rename SectionMessage's controlled close prop to open

OsamaAbdellateef:
# Description

`close={true}` showed the message and `close={false}` hid it — the opposite of what the prop name and docs implied. Renamed the controlled prop to `open`/`onOpenChange` (plus a new `defaultOpen` prop for uncontrolled use) to match the positive polarity already used by `Dialog`/`Drawer`/`Tray`/`Sidebar`.

This is a cherry-pick of the same fix from `main` (#dst-1556_fix-section-message-close-prop) onto `beta-release`, since `SectionMessage` has diverged there (slot-configuration pattern).

Closes DST-1556

## Screenshots / Preview

Not applicable — prop rename only, no visual change.

## Test Instructions

1. Open the `SectionMessage` Storybook stories and confirm `open`/`onOpenChange`/`defaultOpen` control visibility as expected.
2. Run `pnpm test:unit` for `SectionMessage.test.tsx`.

## Breaking Changes

Yes — `close` prop renamed to `open` (inverted polarity). Documented in the included changeset.

## Checklist

- [x] Stories added/updated
- [ ] Accessibility reviewed against ARIA APG
- [ ] Visual regression tests updated


packages/components/src/SectionMessage/SectionMessage.tsx
```diff
@@ -125,31 +131,31 @@ export const SectionMessage = ({
...
     if (!shouldAnnounce) return;
     const text = containerRef.current?.textContent?.trim();
     if (!text) return;
     const priority = variant === 'error' ? 'assertive' : 'polite';
     announce(text, priority);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isCurrentlyVisible]);
+  }, [isOpen]);
 
   const handleClose = () => {
```
sarahgm, line 150: 🔴 **Critical** — `handleClose` branches on `onOpenChange` presence, but visibility is derived from `open` (`isOpen = open ?? internalOpen`). These use different signals, so a legit combination breaks:

**Uncontrolled + `onOpenChange` listener** (`open` undefined, `onOpenChange` provided — e.g. to move focus on dismiss, which the docs recommend): clicking close fires `onOpenChange(false)` but never sets `internalOpen`, so `isOpen` stays `true` and **the message never hides**.

DST-1556's suggested solution branches on controlled-ness. The idiomatic fix (fires the callback in both modes, updates internal state only when uncontrolled) also passes the existing controlled test:

```suggestion
  const handleClose = () => {
    if (open === undefined) {
      setInternalOpen(false);
    }
    onOpenChange?.(false);
  };
```

packages/components/src/SectionMessage/SectionMessage.test.tsx
```diff
@@ -154,14 +154,14 @@ test('allow to close message with button in message', async () => {
...
       <Basic.Component
         data-testid="messages"
         closeButton
-        close={!close}
-        onCloseChange={setClose}
+        open={open}
+        onOpenChange={setOpen}
       />
     );
   };
```
sarahgm, line 167: 🟡 **Warning** — This only covers the controlled + handler path (which works). Add a test for **uncontrolled** dismiss with an `onOpenChange` spy, asserting both that the callback fires with `false` and that the message unmounts. That path is currently broken — see the comment on `SectionMessage.tsx`.

packages/components/src/SectionMessage/SectionMessage.stories.tsx
```diff
@@ -150,3 +151,30 @@ export const MultiLineTitle = meta.story({
...
+          <SectionMessage.Title>Item deleted</SectionMessage.Title>
+          <SectionMessage.Content>
+            The item was removed successfully. Dismiss this message or delete
+            again to bring it back.
+          </SectionMessage.Content>
+        </SectionMessage>
+      </Stack>
+    );
+  },
+});
```
sarahgm, line 209: 💡 **Suggestion** — Consider adding `tags: ['component-test']` and a `play` function to this story to lock in the controlled show/dismiss round-trip in the browser test runner.

sarahgm (changes requested):
## Automated Code Review

The rename, polarity, and API alignment are correct and match DST-1556. One functional regression needs fixing before merge.

### What the PR does well
- Correct positive polarity (`open={true}` shows, `open={false}` hides), aligned with `Dialog`/`Drawer`/`Tray`/`Sidebar`.
- Adds `defaultOpen` (default `true`), preserving uncontrolled behavior.
- `onOpenChange` now fires `false` on dismiss instead of the stale current value.
- Docs, changeset (`major` + migration table), story, and test updated. No stale `close`/`onCloseChange` usages remain (release blog correctly left untouched).

### 🔴 Critical
**`handleClose` branches on `onOpenChange` presence, not `open` presence.** Visibility is derived from `open` (`isOpen = open ?? internalOpen`), but the dismiss handler decides controlled-vs-uncontrolled from whether `onOpenChange` was passed. In uncontrolled mode with an `onOpenChange` listener (a documented, a11y-relevant case — the docs recommend moving focus on dismiss), clicking close fires the callback but never sets `internalOpen`, so the message never hides. The ticket's suggested solution branches on controlled-ness. See inline comment for the fix.

### 🟡 Warning
**Test coverage gap.** The controlled dismiss test only exercises the controlled + handler path (which works). The broken uncontrolled-with-listener path and the pure uncontrolled `defaultOpen` dismiss path aren't tested. See inline comment.

### 💡 Suggestions
- New `ControlledSectionMessage` story could add `tags: ['component-test']` + a `play` function.
- Minor changeset prose polish (stray space / comma-splice in the intro line).

### Recommendation: **Request Changes**
One-line change to `handleClose` plus a test resolves the Critical finding.

---

packages/components/src/SectionMessage/SectionMessage.tsx
OsamaAbdellateef: Fixed — `handleClose` now sets internal state only when uncontrolled (`open === undefined`) and always calls `onOpenChange?.(false)`, matching your suggestion exactly. The uncontrolled + `onOpenChange`-listener path now correctly hides the message.

packages/components/src/SectionMessage/SectionMessage.test.tsx
OsamaAbdellateef: Added — new test `support uncontrolled dismiss message with onOpenChange listener` in SectionMessage.test.tsx covers this exact path: renders uncontrolled with only an `onOpenChange` spy, clicks close, and asserts both that the callback is called with `false` and that the message unmounts.

packages/components/src/SectionMessage/SectionMessage.stories.tsx
OsamaAbdellateef: Done — added `tags: ['component-test']` and a `play` function to `ControlledSectionMessage` that clicks "Delete item" to open the message, asserts it renders, clicks the close button, and asserts it unmounts. Runs via `pnpm test:sb`.

sarahgm (approved):
resolve conflicts

PR #5620, opened 2026-07-06
feat(DST-1533): add Table.Footer component

OsamaAbdellateef:
# Description

Adds `<Table.Footer>` for rendering a semantic `<tfoot>` summary row (totals, counts, averages) after `<Table.Body>`, composed from `<Table.Row>` and `<Table.Cell>` just like the body. Unlike faking a totals row inside the body, a real `<tfoot>` stays out of the body's scroll/sticky-header semantics and is announced correctly by assistive technology. A `sticky` prop pins the footer to the bottom of the viewport while scrolling, mirroring how sticky headers keep column labels in view.

Closes DST-1533

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Open Storybook → `Components/Table`, exercise the new footer stories (including the sticky variant).
2. Run `pnpm start`, navigate to the Table docs page, and check the "Footer totals" and sticky-footer demos.
3. Run `pnpm test:unit` to confirm the new `Table.test.tsx` assertions for `Table.Footer` pass.

## Breaking Changes

No

## Checklist


packages/components/src/Table/Table.stories.tsx
```diff
@@ -1831,3 +1831,137 @@ ColumnAlignment.test(
...
+              />
+            </Table.Cell>
+          </Table.Row>
+        </Table.Footer>
+      </Table>
+    );
+  },
+});
+
+FooterTotals.test(
```
jim761: not a component test, should be a normal test

packages/components/src/Table/TableFooter.tsx
```diff
@@ -0,0 +1,36 @@
+import type RAC from 'react-aria-components';
+import { TableFooter as RACTableFooter } from 'react-aria-components/Table';
```
sebald, line 2: **Please import from the package barrel, not the `/Table` subpath.**

```ts
import { TableFooter as RACTableFooter } from 'react-aria-components';
```

`TableFooter` (and `TableFooterProps`) are exported from the main `react-aria-components` entry, and all five sibling components (`TableHeader`, `TableBody`, `TableRow`, `TableCell`) import from that barrel. This lone subpath import is inconsistent — and the ticket explicitly asked to "mirror the `TableHeader` implementation."

Beyond consistency, there's a subtle correctness concern: this pulls a RAC collection component from a *different module path* than the parent `RACTable` (imported from the barrel in `Table.tsx`), so both paths must resolve the same internal collection-context singleton. It works today, but it's fragile.

(Note: Vercel's `bundle-barrel-imports` guidance actually favors subpath imports for bundle size — so if the team wants that optimization, do it as a separate, *consistent* sweep across all Table files, not just here.)

themes/theme-rui/src/components/Table.styles.ts
```diff
@@ -183,4 +184,11 @@ export const Table: ThemeComponent<'Table'> = {
       '[&_svg]:pointer-events-none [&_svg]:shrink-0 [&_svg]:size-4.5',
     ],
   }),
+  // <tfoot>
+  footer: cva({
+    base: [
+      'bg-surface/90 border-border border-t',
```
sebald: **Nit — use `bg-background/90` to match the sticky-header convention.**

The header's sticky background is `bg-background/90` (the `head` slot, commented "for sticky header"), but the footer uses `bg-surface/90`. Both tokens resolve to `--color-white` today, so there's no visual difference right now — but if a theme ever sets `surface ≠ background`, the sticky header and footer backgrounds would silently diverge. Aligning to `bg-background/90` keeps the two sticky tiers consistent.

packages/components/src/Table/Table.test.tsx
```diff
@@ -165,6 +167,45 @@ describe('Sticky Header', () => {
   });
 });
 
+describe('Footer', () => {
```
sebald, line 169: **Optional — redundant with the story `.test()` play functions.**

Three of these four assertions are already covered by the `component-test` play functions in `Table.stories.tsx`: `FooterTotals.test` checks the `<tfoot>` + `colspan=4`, and `StickyFooter.test` checks the sticky class. Only the negative case here ("no `sticky` prop → no sticky class") is unique.

Not blocking — I see the repo intentionally keeps both a jsdom and a browser layer — but you could trim the duplicated positive cases here if you'd like a leaner diff.

packages/components/src/Table/Table.stories.tsx
```diff
@@ -1831,3 +1831,137 @@ ColumnAlignment.test(
     );
   }
 );
+
+export const FooterTotals = meta.story({
```
sebald, line 1835: **Noted, not actionable — leave as-is.** `FooterTotals` and `StickyFooter` (and the two `.demo.tsx` files) are ~90% identical, differing only by the `Scrollable` wrapper and the `sticky` prop. That's idiomatic for this file — stories are meant to be explicit, copy-pasteable examples — so I'd keep the duplication rather than extract a shared render helper. Flagging only for completeness.

sebald (changes requested):
Thanks for this — scope matches DST-1533 exactly (nothing missing, nothing over-built), coverage is 100% on both changed source files, and the `sticky`/`colSpan`/semantic-`tfoot` behavior all look correct. No `forwardRef` is the right call for a RAC collection component, and it mirrors `TableHeader` well.

Requesting one change before merge:

- **`TableFooter.tsx:2`** — import `TableFooter` from the `react-aria-components` barrel instead of the `/Table` subpath, to match all five sibling components (and the ticket's "mirror `TableHeader`" note). Details inline.

Plus two optional cleanups from a `/simplify` pass (both inline, non-blocking):
- `Table.styles.ts` footer slot — `bg-surface/90` → `bg-background/90` to match the sticky-header token convention.
- `Table.test.tsx` `describe('Footer')` — a few assertions duplicate the story `.test()` play functions.

Once the import is switched I'm happy to approve.

themes/theme-rui/src/components/Table.styles.ts
```diff
@@ -183,4 +184,11 @@ export const Table: ThemeComponent<'Table'> = {
       '[&_svg]:pointer-events-none [&_svg]:shrink-0 [&_svg]:size-4.5',
     ],
   }),
+  // <tfoot>
+  footer: cva({
```
jim761: 💡 The `footer` style is appended at the very end (after `dragHandle`), while `body`/`head` sit near the top of the object. Consider grouping `footer` next to `body` so the `<tbody>`/`<tfoot>` styles read together. Non-blocking — the `// <tfoot>` comment already helps discoverability.

themes/theme-rui/src/components/Table.styles.ts
```diff
@@ -99,6 +99,7 @@ export const Table: ThemeComponent<'Table'> = {
 
   // <tbody>
   body: cva({}),
+
```
jim761, line 102: 💡 Stray blank line inserted between `body: cva({})` and `cell:` — looks unrelated to this change. Harmless, but worth dropping to keep the diff tight.

packages/components/src/Table/TableFooter.tsx
OsamaAbdellateef: Fixed in 05d47d1: `TableFooter.tsx` now imports `TableFooter` from the `react-aria-components` barrel, matching `TableHeader`/`TableBody`/`TableRow`/`TableCell`.

themes/theme-rui/src/components/Table.styles.ts
OsamaAbdellateef: Fixed in 05d47d1: the footer's sticky background now uses `bg-background/90`, matching the sticky-header token convention.

packages/components/src/Table/Table.test.tsx
OsamaAbdellateef: Fixed in 05d47d1: trimmed the duplicated positive cases here — removed `applies sticky class to footer` (already covered by `StickyFooter.test`'s play function, which also checks the actual scroll geometry). Kept the two `tfoot`/colSpan checks and the unique negative case (see reply on the `FooterTotals.test` thread for why those two are no longer duplicated with the story play function).

packages/components/src/Table/Table.stories.tsx
OsamaAbdellateef: Noted, left as-is per your call — kept the duplication between `FooterTotals`/`StickyFooter` (and the two `.demo.tsx` files) since explicit, copy-pasteable stories are the convention here.

themes/theme-rui/src/components/Table.styles.ts
OsamaAbdellateef: Fixed in 05d47d1: moved the `footer` slot to sit right below `body`, so `<tbody>`/`<tfoot>` styles read together.

themes/theme-rui/src/components/Table.styles.ts
OsamaAbdellateef: Fixed in 05d47d1: removed the stray blank line.

packages/components/src/Table/Table.stories.tsx
OsamaAbdellateef: Fixed in 05d47d1: removed the `FooterTotals.test` play function — it only asserted on static markup (`<tfoot>` presence + `colspan`), no real interaction, so it didn't need a browser-based component test. That coverage now lives solely in `Table.test.tsx`'s `describe('Footer')` block as plain jsdom tests (`renders a semantic footer with the row content`, `respects colSpan on footer cells`), which is exactly the 'normal test' you were pointing at.

jim761 (commented):
## Automated Code Review

2 warnings and 2 suggestions inline below. The component logic is clean and idiomatic — recommend addressing the two warnings (footer background token + anatomy a11y name) before merge.

---

themes/theme-rui/src/components/Table.styles.ts
```diff
@@ -99,6 +99,15 @@ export const Table: ThemeComponent<'Table'> = {
...
+  // <tfoot>
+  footer: cva({
+    base: [
+      'bg-background/90 border-border border-t',
```
jim761: 🟡 **Footer background diverges from the sticky header.**

The sticky header uses `bg-surface/90` (white, `--color-surface`) but this footer uses `bg-background/90` (gray, `--color-charcoal-100`). Since the table base is `bg-surface`, the sticky footer renders a gray strip that matches neither the table surface nor the sticky header.

If the shading is a deliberate totals treatment, please confirm it. Otherwise, match the header:

```suggestion
      'bg-surface/90 border-border border-t',
```

docs/content/components/collection/table/table-anatomy.tsx
```diff
@@ -0,0 +1,198 @@
+export const TableAnatomy = () => (
+  <svg
```
jim761, line 2: 🟡 **Anatomy SVG has no accessible name.**

This `<svg>` has no `role="img"`, `aria-label`, or `<title>`, so screen readers announce nothing. It regresses the previous `<ImageZoom alt="Anatomy of Table component" />` and is inconsistent with `card-anatomy.tsx`, which sets `role="img"` + `aria-label`.

```suggestion
  <svg
    role="img"
    aria-label="Anatomy of a Table: sticky Header with selection checkbox and columns, Body with selectable rows and cells, and a sticky Footer summary row"
```

packages/components/src/Table/TableFooter.tsx
```diff
@@ -0,0 +1,36 @@
+import type RAC from 'react-aria-components';
+import { TableFooter as RACTableFooter } from 'react-aria-components';
```
jim761: 💡 **Import path inconsistency.**

Sibling files (`TableHeader.tsx`, `TableBody.tsx`) import table values from the `react-aria-components/Table` subpath; this imports from the package root. Align for folder consistency and tree-shaking. (The type import on line 1 can stay at the root — that matches the siblings.)

```suggestion
import { TableFooter as RACTableFooter } from 'react-aria-components/Table';
```

packages/components/src/Table/Table.stories.tsx
```diff
@@ -1831,3 +1831,125 @@ ColumnAlignment.test(
...
+  tags: ['component-test'],
```
jim761: 💡 **Remove the `component-test` tag from `FooterTotals`.**

The tag pulls this story into the Storybook test runner, but it has no `play` function so nothing is asserted. Its render is already covered by the unit test (`Table.test.tsx` imports `<FooterTotals.Component />`), and interaction coverage lives in `StickyFooter.test(...)`.

```suggestion
```

packages/components/src/Table/TableFooter.tsx
OsamaAbdellateef: https://github.com/marigold-ui/marigold/pull/5620#discussion_r3536165040
interfering with this conversation  @jim761

jim761: TableHeader and the other components import from /Table or /(component), so it should be the correct way

sebald (approved):

PR #5622, opened 2026-07-07
feat(DST-1439): give SectionMessage a neutral surface

sarahgm:
# Description

Aligns `<SectionMessage>` with a neutral, Toast-adjacent surface instead of the per-variant tinted background.

**Before:** each variant filled its background with a solid tint (`bg-info` / `bg-success` / `bg-warning` / `bg-destructive`) and used variant-colored text. A default `<Button>` or `<Link>` placed inside floated on the colored fill and read as a foreign element.

**After:** the message sits on the shared `ui-surface` primitive (neutral fill, subtle border, `rounded-surface`) with neutral text (`text-foreground` title/content, `text-secondary` description). The variant is conveyed by the colored icon plus a muted per-variant colored border (the accent mixed halfway into the neutral border). Because the surface is neutral, standard `<Button>` / `<Link>` actions read correctly with no new primitive and no variant-aware button work. The bordered, in-flow treatment keeps it visually distinct from the floating, shadowed `<Toast>`.

Also fixed as part of the visual pass: the leading icon and the close button now align on the title's first line, and the close button keeps a 24×24 touch target.

## Why a muted border (and not icon-only / tint / glow)

The variant cue was chosen after building and reviewing the alternatives side by side:

- **Icon-only (like Toast) was too quiet.** Toast is transient, singular, and animates in, so its colored icon suffices. SectionMessage is persistent and in-flow and is used for at-a-glance cases like a form-level error banner, so it needs a surface-level cue, not just a 20px icon.
- **A tinted fill reintroduces the original bug.** Any fill puts the actions on a colored surface. Marigold's default `<Button>` is `variant: 'secondary'`, which is `ui-surface` with its own opaque surface fill, so on a tint it renders as a foreign chip and floats. That is the exact problem this ticket set out to remove, so a fill of any strength is out.
- **A full accent border reads like an error or invalid state**, which the system reserves colored borders for. Mixing the accent 50% into the neutral border keeps the message legible as a variant without that connotation.
- **A soft inset glow read as "not visible enough" in review.** A border is the more visible, system-native cue, and it sits at the container edge, away from the actions.

The border color is set through `ui-surface`'s `--ui-border-color` hook using tokens (`color-mix(... var(--color-*-accent) 50%, var(--color-border))`). That property is registered as a non-inheriting custom property, so the variant color stays scoped to the container and does not leak into the borders of nested actions or inputs.

Closes DST-1439

## Screenshots / Preview

| Before | After |
| ------ | ----- |
| Per-variant tinted fill, colored text | Neutral surface, colored icon + muted variant border |

## Test Instructions

1. Storybook, `Components/SectionMessage`. Check all four variants render on a neutral surface with the icon and a muted variant-colored border carrying the variant.
2. Toggle `closeButton` on and confirm the close icon and the leading icon both line up with the title.
3. Place the message on a Card and on the page (the `surface: both` stories) and confirm actions read correctly and the box reads distinct from a Toast.
4. Verify light and dark themes.

## Breaking Changes

No. No API changes. Visual change only.

## Checklist

- [x] Unit tests pass (existing suite, description-slot assertion updated)
- [x] Accessibility reviewed (axe: 0 violations, contrast measured AA-compliant)
- [x] Visual regression tests: N/A (PR targets `beta-release`; VRT is enforced only on PRs into `main`)
- [x] Changeset added (`@marigold/theme-rui` minor, `@marigold/components` patch)


themes/theme-rui/src/components/SectionMessage.styles.ts
```diff
@@ -3,18 +3,23 @@ import { ThemeComponent, cva } from '@marigold/system';
 export const SectionMessage: ThemeComponent<'SectionMessage'> = {
   container: cva({
     base: [
-      "grid-cols-[min-content_auto_min-content] gap-x-4 [grid-template-areas:'icon_title_close''icon_description_description''icon_content_content']",
-      'bg-surface rounded-md border px-3 py-4 [--section-message-description:currentColor]',
+      "grid-cols-[min-content_auto_min-content] gap-x-4 [grid-template-areas:'icon_title_close''icon_description_close''icon_content_close']",
```
aromko: ⚠️ **Phantom 16px right gap when `closeButton` is off (the default)**

The `close` area now spans all three rows, so description/content no longer span into column 3. That empty `min-content` column still costs its `gap-x-4`: measured in a browser, the text's right inset goes **12px → 28px** without a close button (and 12px → 48px with one), making the horizontal insets visibly asymmetric for every default-configured message. Since this PR targets `beta-release`, Chromatic won't catch it until the merge-to-`main` PR.

A fix that keeps the alignment win: drop `gap-x-4` and use margins instead (`mr-4` on the icon area, `ml-4` on the close button — `Toast.styles.ts` uses this pattern with `ml-2`), so the empty column contributes exactly 0.

packages/components/src/SectionMessage/SectionMessage.stories.tsx
```diff
@@ -150,3 +138,45 @@ export const MultiLineTitle = meta.story({
     </div>
   ),
 });
+
+export const WithoutTitle = meta.story({
```
aromko, line 162: 💡 **No snapshot-enabled story covers the non-default variants**

All Chromatic-enabled stories use the default `info` args, but this redesign restyles all four variants and DST-1439 explicitly asks for a VRT pass over every variant. An `AllVariants` story (snapshot enabled) would make the eventual `main` Chromatic run cover the success/warning/error borders.

packages/components/src/SectionMessage/SectionMessage.stories.tsx
```diff
@@ -150,3 +138,45 @@ export const MultiLineTitle = meta.story({
...
+  'aligns the leading icon and close button with the first content line',
+  {
+    parameters: { chromatic: { disableSnapshot: true }, surface: false },
+  },
+  async ({ canvas, canvasElement }) => {
+    const content = canvas.getByText(
+      'Hello, I am a simple message without a title.'
+    );
+    const closeButton = canvas.getByRole('button');
+    const icon = canvasElement.querySelector('[class*="grid-area:icon"]')!;
```
aromko, line 184: 💡 **Brittle selector**

`querySelector('[class*="grid-area:icon"]')!` is a class-substring selector with a non-null assertion. Acceptable since the icon is decorative (no role to query), but a short comment explaining why no accessible query is possible would help the next reader.

aromko (commented):
## Automated Code Review

### Overview
| Field | Value |
|-------|-------|
| Branch | `dst-1439-sectionmessage-neutral-surface` → `beta-release` |
| Files Changed | 6 (+95 / −51) |
| CI | All 18 checks pass (typecheck, lint, unit, storybook tests) |

### Linked Ticket
**[DST-1439](https://reservix.atlassian.net/browse/DST-1439)**: 💄 Align SectionMessage with Toast design (neutral surface background) — In Review

The PR matches the ticket scope precisely: neutral `ui-surface` background, neutral text, variant carried by the colored icon, docs + action demo updated to use a real `<Button>`, changeset added. The ticket explicitly left the edge treatment open — the muted 50% accent-mix border is within that scope and the PR body documents why the alternatives were rejected.

### What checks out
- **TypeScript / patterns**: no `any`, theme-driven styling, no public `className`/`style` props (the `className` on `CloseButton` is internal usage).
- **Border scoping**: `--ui-border-color` is set through the registered **non-inheriting** `@property` in `ui.css`, so the variant border verifiably can't leak into nested `ui-surface` buttons.
- **Testing**: `component-test` tags present, no `fireEvent`/Jest APIs; the `WithDescription` assertion was updated meaningfully (`text-secondary` instead of the removed CSS-var plumbing); the new `WithoutTitle` alignment test is a real regression test and passes in CI.
- **Accessibility**: variant not conveyed by color alone (icon shape + color + title); close button keeps a 24×24 touch target via `ui-touch-hitbox` while shrinking to a 20px visual box.
- **React performance**: no findings — the runtime change is a single className string.

### Issues

**⚠️ Warning (should fix)**
- `themes/theme-rui/src/components/SectionMessage.styles.ts:6` — the new grid template creates a 16px dead zone on the right when `closeButton` is off (the default). Details in the inline comment.

**💡 Suggestions**
- No snapshot-enabled story renders the non-default variants (details inline).
- Brittle `querySelector` in the `WithoutTitle` test (details inline).

### Recommendation
A well-executed, well-documented change that lands exactly the ticket scope, with improved tests and correct changeset bumps (theme-rui minor, components patch). The one thing worth addressing before merge is the phantom right-gap in the default no-close-button layout — a visible regression that VRT won't flag on this PR (base is `beta-release`), with a low-risk margin-based fix.

---

packages/components/src/SectionMessage/SectionMessage.stories.tsx
```diff
@@ -150,3 +138,45 @@ export const MultiLineTitle = meta.story({
...
+    <SectionMessage {...args}>
+      <SectionMessage.Content>
+        <Text>Hello, I am a simple message without a title.</Text>
+      </SectionMessage.Content>
+    </SectionMessage>
+  ),
+});
+
+WithoutTitle.test(
+  'aligns the leading icon and close button with the first content line',
```
aromko, line 175: A tests which tests the CloseButton here would also be nice with disabledSnapshot.

themes/theme-rui/src/components/SectionMessage.styles.ts
sarahgm: Fixed in 1e24c7cbb — dropped `gap-x-4`, added `mr-4` on the icon and `ml-4` on the close button (the `Toast.styles.ts` pattern). The empty close column now contributes 0.

packages/components/src/SectionMessage/SectionMessage.stories.tsx
sarahgm: Added an `AllVariants` snapshot-enabled story in 1e24c7cbb rendering all four variants with the default (no close button) layout, so the merge-to-`main` Chromatic run covers the variant borders and the fixed inset.

packages/components/src/SectionMessage/SectionMessage.stories.tsx
sarahgm: Leaving the selector as-is without an added comment (deliberate call to keep the test terse). The icon stays decorative with no accessible query.

packages/components/src/SectionMessage/SectionMessage.stories.tsx
sarahgm: Added a `component-test` on `WithoutTitle` in 1e24c7cbb that clicks the close button and asserts the message is dismissed (covers the uncontrolled visibility path).

aromko (approved):

sebald (approved):

PR #5625, opened 2026-07-09
feat(DST-1461): align Accordion like Table inside a bled Panel

sebald:
# Description

`Panel.Content` already lets a `<Table>` bleed edge-to-edge so row backgrounds/borders reach the Panel border while cell content stays inset and aligned with the Panel title. There was no equivalent for `<Accordion>`, a common "section-owning" content block inside a Panel.

The `default` Accordion now gains internal horizontal padding via a new `--accordion-x-padding` custom property, which resolves to the Panel's `--panel-px` when nested inside a Panel and falls back to `0px` standalone. Dropping an `<Accordion>` into `<Panel.Content bleed>` now gives full-width item dividers **and** header/content aligned with the Panel title — mirroring how Table behaves, with **no new Accordion prop or variant**. Standalone Accordions are unchanged (`--panel-px` is only set inside a Panel).

Also adds Panel docs for the pattern — including guidance on choosing `<Accordion>` (peer, equally-weighted expandable sections) vs `Panel.Collapsible` (a single optional/advanced fold) — a canonical demo, and a Panel story.

Closes DST-1461

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm sb`, open **Components/Panel/AccordionInside**. Confirm the accordion's item dividers span the full Panel width while the header text and expanded content align with the Panel title; the standalone Accordion below has no horizontal indent.
2. Run `pnpm start`, open **/components/layout/panel** → "Full-width content" and check the new Accordion demo and the Accordion-vs-Collapsible guidance.
3. `pnpm typecheck:only` and `pnpm test:unit Accordion Panel` (60 tests pass locally).

## Breaking Changes

No

## Checklist


themes/theme-rui/src/components/Accordion.styles.ts
```diff
@@ -2,7 +2,12 @@ import { ThemeComponent, cva } from '@marigold/system';
 
 export const Accordion: ThemeComponent<'Accordion'> = {
   container: cva({
-    base: 'flex-col',
+    // Horizontal padding for header/content. Resolves to the Panel's
+    // `--panel-px` when nested inside a Panel (so a bled Panel keeps dividers
+    // edge-to-edge while header/content align with the Panel title, mirroring
+    // Table), otherwise falls back to `0px` — standalone Accordions are
+    // unchanged.
+    base: 'flex-col [--accordion-x-padding:var(--panel-px,0px)]',
```
jim761: ⚠️ **This is a regression for a non-bled `Panel.Content`, not just a thing to verify.**

`--panel-px` is inherited by every Panel descendant, so this resolves to the Panel's padding even when the surrounding `Panel.Content` is **not** bled — and a non-bled `Panel.Content` already applies `px-(--panel-px)` itself. Net result:

- item divider (`border-b`, no padding): `1 × --panel-px` from the panel edge
- header/content text: `2 × --panel-px` (Panel.Content padding **+** this new Accordion padding)

So the header text no longer lines up with the divider's left edge. Before this PR it did (default header/content had no `px`). Any existing `Accordion` inside a plain, non-bled Panel picks up this extra indent silently.

It happens to mirror `Table`'s current behavior, but that doesn't make it correct — it's a real misalignment in a supported composition. Options:

1. Scope the padding to the bleed case only (e.g. `Panel.Content` neutralizes `--panel-px` for its subtree when not bleeding — this would fix `Table` too, but is a larger change and best split into its own ticket), or
2. At minimum, document that Accordion-in-Panel alignment is intended for **bled** content and guard against the non-bled case in the demo/story.

Worth deciding before merge rather than shipping the silent regression.

jim761 (commented):
## Automated Code Review — #5625

The bleed case is implemented cleanly: it reuses the `Table` `--cell-edge-padding` pattern, adds no new prop/variant, and standalone Accordions stay visually unchanged (`px-0px`). CI is fully green. One genuine regression to fix (below), plus notes. Specifics are in the inline comments.

**Well done**
- Correctly adds `defaultVariants: { variant: 'default' }` to `content` — required now that the default carries padding (previously the empty default made it moot).
- Right fallback choice: `0px` for Accordion vs Table's `--cell-x-padding` (header/content need no intrinsic padding).
- Good docs + Accordion-vs-`Panel.Collapsible` guidance; clear explanatory comment in the styles file.

**Problem — non-bled `Panel.Content` regresses**
- ⚠️ Because `--panel-px` is inherited unconditionally, an Accordion inside a **non-bled** `Panel.Content` now gets `2 × --panel-px` on header/content (the Panel.Content padding *plus* the new Accordion padding), while the item divider stays at `1 × --panel-px`. Header text no longer aligns with the divider — it did before this PR. This is a real visual regression for a supported composition, not just something to eyeball. See inline on `Accordion.styles.ts` for the mechanism and fix options.

**Should also address**
- 🧪 No unit test for the new wiring (checklist left unchecked). A lightweight `Accordion.test.tsx` assertion — container contains `[--accordion-x-padding:var(--panel-px,0px)]`, and default `header`/`content` contain `px-(--accordion-x-padding)` — would guard the plumbing (mirrors `Panel.test.tsx`'s `toContain('px-(--panel-px)')`).

**Minor**
- Naming: `--accordion-x-padding` vs Table's `--cell-edge-padding` — both fine, noting only for cross-component discoverability.

---

docs/content/components/layout/panel/panel-bleed-accordion.demo.tsx
```diff
@@ -0,0 +1,75 @@
...
+export default () => (
+  <Panel>
+    <Panel.Header>
+      <Title>Prop values</Title>
+      <Description>
+        Usage across all Button instances, last 30 days.
+      </Description>
+    </Panel.Header>
+    <Panel.Content bleed>
+      <Accordion>
```
jim761: 💡 The demo renders every item collapsed, so the header/content alignment the surrounding prose describes isn't visible until the reader clicks. Consider `<Accordion defaultExpandedKeys={['variant']}>` (as the `AccordionInside` story already does) to showcase the aligned expanded content out of the box.

.changeset/accordion-bleed-panel.md
```diff
@@ -0,0 +1,18 @@
...
+
+- The `default` Accordion `header` and `content` now pick up horizontal padding from a new `--accordion-x-padding` custom property, which resolves to the Panel's `--panel-px` when nested inside a `Panel` and falls back to `0px` otherwise.
+
+**Why:**
+
+Dropping an `<Accordion>` into `<Panel.Content bleed>` now gives full-width item dividers _and_ header/content aligned with the Panel title — the same behavior `Table` already had, with no new Accordion prop or variant.
+
+**Impact:**
+
+- Standalone Accordions are unchanged (`--panel-px` is only set inside a Panel, so the padding resolves to `0px`).
```
aromko: ⚠️ **Impact section misses a composition that changes: Accordion in a non-bled `Panel.Content`**

The wrapper already applies `px-(--panel-px)`, and since `--panel-px` is inherited from the Panel root regardless of `bleed`, the header/content now add another `--panel-px` on top: a double indent that didn't exist before this PR. This matches Table's existing behavior, but it's a visible change for consumers with an Accordion in a non-bled Panel today (e.g. Insights on beta), so the release notes should name it.

```suggestion
- Standalone Accordions are unchanged (`--panel-px` is only set inside a Panel, so the padding resolves to `0px`).
- Inside a **non-bled** `Panel.Content`, the Accordion now adds `--panel-px` on top of the wrapper's own padding (a double indent), same as Table. Use `bleed` when the Accordion owns the Panel body.
```

themes/theme-rui/src/components/Accordion.styles.ts
```diff
@@ -37,7 +42,8 @@ export const Accordion: ThemeComponent<'Accordion'> = {
     ],
     variants: {
       variant: {
-        default: 'focus-visible:ui-state-focus outline-none',
+        default:
+          'focus-visible:ui-state-focus outline-none px-(--accordion-x-padding)',
```
aromko: 💡 **Focus ring now sits flush against the Panel border**

In a bled Panel the header spans the full Panel width, so its `rounded-md` focus ring touches (and its corners overlap) the Panel border. `Panel.Collapsible` insets its ring for exactly this case via `m-1` + `px-[calc(var(--panel-px)-var(--spacing))]` (`Panel.styles.ts:40-41`). Worth a design pass on whether the Accordion header should do the same when nested in a Panel.

docs/content/components/layout/panel/index.mdx
```diff
@@ -147,6 +147,12 @@ Tables, media, charts, and status bars are the exceptions. They look cramped whe
 
 Tables get a small bonus: when bled, their cell edges line up with the Panel's header and footer, so what's inside a row still reads in rhythm with the title above it.
 
+An `<Accordion>` behaves the same way. Drop one into a bled content area and its item dividers span edge-to-edge while the header and content stay inset, aligned with the Panel title. No extra prop is needed: the Accordion picks up the Panel's horizontal padding automatically, exactly like Table cells do.
+
+Reach for this only when the Panel's body **is** a set of peer, equally-weighted sections that each expand — a per-field breakdown, a list of records each with details. For a single optional or advanced fold _alongside_ visible primary content, use [`Panel.Collapsible`](#collapsible) instead. The two aren't interchangeable: a stack of Collapsibles is exactly the anti-pattern Accordion exists to solve (see [One Collapsible per Panel](#one-collapsible-per-panel)).
```
aromko: 💡 **Docs style: avoid em dashes, rephrase with commas/colons**

```suggestion
Reach for this only when the Panel's body **is** a set of peer, equally-weighted sections that each expand: a per-field breakdown, a list of records each with details. For a single optional or advanced fold _alongside_ visible primary content, use [`Panel.Collapsible`](#collapsible) instead. The two aren't interchangeable: a stack of Collapsibles is exactly the anti-pattern Accordion exists to solve (see [One Collapsible per Panel](#one-collapsible-per-panel)).
```

docs/content/components/layout/panel/index.mdx
```diff
@@ -147,6 +147,12 @@ Tables, media, charts, and status bars are the exceptions. They look cramped whe
 
 Tables get a small bonus: when bled, their cell edges line up with the Panel's header and footer, so what's inside a row still reads in rhythm with the title above it.
 
+An `<Accordion>` behaves the same way. Drop one into a bled content area and its item dividers span edge-to-edge while the header and content stay inset, aligned with the Panel title. No extra prop is needed: the Accordion picks up the Panel's horizontal padding automatically, exactly like Table cells do.
```
aromko, line 150: 💡 **Consider noting the ticket's two open visual questions**

DST-1461 scopes "leading divider between `Panel.Header` and the first item" and "rounding of the last item's bottom corners" as open questions, documented only. A one-line note here would keep consumers from filing them as bugs.

themes/theme-rui/src/components/Accordion.styles.ts
sebald: Good catch — this was a real regression, fixed in 3a9d267 by scoping the inset to the bled case instead of reading `--panel-px` unconditionally:

- A bled `Panel.Content` / `Panel.CollapsibleContent` now publishes `--bleed-px` (set to `--panel-px`); non-bled publishes nothing.
- The Accordion sources `--accordion-x-padding` from `var(--bleed-px, 0px)`, so it is `0px` in a non-bled `Panel.Content` (and standalone). Header/content align with the dividers exactly as before; only a bled Panel adds the inset.

Verified in a browser: non-bled header text is back to a single `--panel-px` inset (aligned with the divider), bled unchanged.

Also covering the two summary notes here: added unit guards for the wiring (`Accordion.test.tsx` + `Panel.test.tsx`), and on naming — `--bleed-px` is now the shared bridge, with Table still reading `--panel-px` directly (left untouched). Leaving this open so you can re-check the approach.

.changeset/accordion-bleed-panel.md
sebald: Rewritten in 3a9d267. Because the inset is now scoped to the bled case (via `--bleed-px`), the non-bled double-indent no longer occurs, so there is nothing to warn about. The Impact section now states that both standalone **and** non-bled `Panel.Content` Accordions are unchanged, and only bled ones gain the inset. Leaving open in case you want to sanity-check the wording against the new mechanism.

themes/theme-rui/src/components/Accordion.styles.ts
sebald: Did the design pass and implemented the inset (3a9d267). In a bled Panel the default header now pulls itself (and its focus ring) off the border by one spacing step, like `Panel.Collapsible`:

`--accordion-ring-inset: min(var(--accordion-x-padding), var(--spacing))`, applied as `mx` + `w-[calc(100% - inset - inset)]` + reduced `px`. It resolves to `0` when not bled, so standalone / non-bled headers are untouched.

Confirmed in the browser: the ring sits one spacing step off the Panel border with no overflow, text still aligned with the title. Leaving open for your review of the result.

docs/content/components/layout/panel/panel-bleed-accordion.demo.tsx
sebald: Done in c3affe8 — the demo now uses `defaultExpandedKeys={['variant']}` so the aligned content is visible on load.

docs/content/components/layout/panel/index.mdx
sebald: Done in c3affe8 — replaced the em dash with a colon. (This was also the open Vercel feedback on the same paragraph.)

docs/content/components/layout/panel/index.mdx
sebald: Done in c3affe8 — added a line after the demo noting the two intentionally-open details (no divider before the first item, and the last item's bottom corners not following the Panel radius) so they are not filed as bugs.

jim761 (approved):

aromko (approved):

PR #5630, opened 2026-07-10
feat(DST-1265): add `marigold doctor` setup diagnostics command

jim761:
A read-only CLI command that inspects a consumer project's Marigold setup and reports a checklist with actionable fixes. `--format json` for AI agents; exits 1 only on deterministic errors, so it is safe to gate CI on.

Checks: package presence, @marigold/components === @marigold/system version match, best-effort latest-version freshness, MarigoldProvider wrapper, theme prop, Tailwind config, and React peer deps. Following the `validate` command's policy, only deterministic + broken findings (missing core package, components/system mismatch) are errors; everything heuristic is a warning.

- Extract the pure AST helpers out of edit-tsx.ts into lib/tsx-ast.ts (edit-tsx re-imports them; behaviour unchanged) for read-only reuse.
- Freshness reads the docs manifest's new `packages` map. `runDoctor` warm-fetches the manifest on a short, best-effort budget (2.5s) unless `--offline` is passed; the freshness check itself is cache-only. So the checks are filesystem-only, doctor never blocks on the network, and `--offline` skips the fetch entirely.
- doctor is lazy-loaded in the bin to keep @babel/parser off the docs/list hot path.
- Telemetry records only `{ format }` (per the pending DST-1600 GDPR/ePrivacy review) and inherits the shared consent path.
- Wire the telemetry command enum on both the CLI and docs route (+ test) so doctor events aren't silently dropped.
- Document on all four surfaces: --help, CLI README, CLAUDE.md, and the docs CLI page.

# Description

Closes DST-XXXX

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1.
2.

## Breaking Changes

No

## Checklist


docs/content/getting-started/cli/index.mdx
```diff
@@ -202,6 +208,37 @@ It detects Next.js or Vite, then:
   CLI prints a manual fallback for that step instead of guessing.
 </Callout>
 
+### marigold doctor
+
+Diagnose why Marigold isn't working in a project. Reads the filesystem only (no network) and prints a checklist with actionable fixes, ideal when components render unstyled or throw, and for AI agents to self-diagnose and repair a setup.
```
aromko: ⚠️ **Docs vs. behavior: doctor does hit the network by default**

`runDoctor` performs a best-effort manifest warm-fetch (2.5s budget) unless `--offline` is set, so "Reads the filesystem only (no network)" isn't accurate. The checks themselves are cache-only, but the command fetches.

Suggest rewording to something like: "All checks read the filesystem only; the freshness check additionally makes a short, best-effort fetch of the docs manifest (skipped with `--offline`)." The flag table below should also list `--offline` (it's already in `--help` and the completion spec).

Same wording issue in `packages/cli/README.md`, and the PR description's "never hits the network" bullet.

packages/cli/README.md
```diff
@@ -191,6 +191,31 @@ Detects Next.js or Vite, then:
 
 Edits are idempotent — re-running leaves files untouched. If a file shape can't be recognized, the CLI prints a manual fallback for that step instead of guessing.
 
+### `marigold doctor`
+
+Diagnose why Marigold isn't working in a project. Reads the filesystem only (no network) and prints a checklist with actionable fixes.
```
aromko: ⚠️ Same "no network" wording issue as in `docs/content/getting-started/cli/index.mdx`, and the Flags list here is also missing `--offline`.

packages/cli/src/bin/marigold.ts
```diff
@@ -370,6 +390,28 @@ export const main = async (
         yes: values.yes,
         skipInstall: values['skip-install'],
       });
+    } else if (command === 'doctor') {
+      const { values } = parseDoctorCommand(rest);
+      // Record only { format }: the pending DST-1600 GDPR/ePrivacy review scopes
+      // doctor telemetry to the output format alone, so --offline is not tracked.
+      telemetryArgs = {
+        format: values.format ?? 'text',
```
aromko: ⚠️ **Telemetry can record an unvalidated `--format` string**

`telemetryArgs.format` is assigned before `isDoctorFormat` validates it, and `fail()` throws a `CliError` that `main()` catches without bypassing the telemetry emit. So `marigold doctor --format banana` records the raw string `banana`, while the README promises "values redacted — only flag presence/enum value is kept" and the comment above scopes doctor telemetry to the output format per the pending DST-1600 review. Consider clamping:

```suggestion
        format: isDoctorFormat(values.format ?? 'text')
          ? (values.format ?? 'text')
          : 'invalid',
```

(The same pattern pre-exists in `docs`/`search`/`examples`, so a pattern-wide follow-up would also work.)

packages/cli/src/lib/doctor/version.ts
```diff
@@ -0,0 +1,52 @@
...
+  /** true when the version carries a prerelease tag (e.g. -beta.3). */
+  prerelease: boolean;
+}
+
+const parse = (version: string): Parsed | null => {
+  const m = /^(\d+)\.(\d+)\.(\d+)(?:[-+].*)?$/.exec(version.trim());
+  if (!m) return null;
+  return {
+    parts: [Number(m[1]), Number(m[2]), Number(m[3])],
+    prerelease: /[-]/.test(version),
```
aromko: 💡 **Build metadata is misclassified as a prerelease**

`/[-]/.test(version)` looks at the whole string, so `1.0.0+build-5` (hyphen inside build metadata) parses as a prerelease and compares below `1.0.0`, which could produce a false "outdated" warning. Semver ignores `+` metadata for precedence.

```suggestion
    prerelease: /-/.test(version.split('+')[0]),
```

packages/cli/src/commands/doctor.test.ts
```diff
@@ -0,0 +1,557 @@
...
+  test('warns when the theme value is passed but not imported', async () => {
+    seedNext(dir);
+    write(
+      dir,
+      'app/providers.tsx',
+      `'use client';\nimport { MarigoldProvider } from '@marigold/components';\nexport function Providers({ children }) {\n  return <MarigoldProvider theme={theme}>{children}</MarigoldProvider>;\n}\n`
+    );
+
+    const { errorIds, warningIds } = checks(await report());
+    const tp = (await report()).warnings.find(w => w.check === 'theme-passed');
```
aromko: 💡 A few tests run the full doctor twice (`checks(await report())` followed by another `await report()`; also in the `@source` test below). Reusing a single report keeps one Act per test and halves the fixture work:

```ts
const r = await report();
const { errorIds, warningIds } = checks(r);
const tp = r.warnings.find(w => w.check === 'theme-passed');
```

packages/cli/src/bin/marigold.ts
```diff
@@ -179,6 +189,16 @@ const parseInitCommand = (argv: string[]) =>
     },
   });
 
+const parseDoctorCommand = (argv: string[]) =>
+  parseArgs({
+    args: argv,
+    allowPositionals: true,
```
aromko, line 198: 💡 `doctor` takes no positionals, but `allowPositionals: true` means `marigold doctor foo` silently ignores the argument (`examples` validates this). Consider failing on unexpected positionals.

packages/cli/src/lib/doctor/index.ts
```diff
@@ -0,0 +1,65 @@
...
+import { type RenderRow, renderText } from './format.js';
+import { readPackageJson } from './package-json.js';
+import { detectProvider } from './provider.js';
+import type { DoctorContext, DoctorIssue, DoctorReport } from './types.js';
+
+export type { DoctorReport, DoctorIssue } from './types.js';
+
+/** Thrown when doctor can't run at all (e.g. no package.json at cwd). Distinct
+ * from a health-check error so the bin layer surfaces it as a usage failure. */
+export class DoctorUsageError extends Error {}
```
aromko, line 15: 💡 The comment says the bin layer surfaces `DoctorUsageError` as a usage failure, but `bin/marigold.ts` never imports or `instanceof`-checks it; it rides the same generic catch as `CliError` (message + exit 1, which is fine behaviorally). Either consume the class distinctly or simplify the class/comment.

docs/content/getting-started/cli/index.mdx
jim761: Good catch — reworded to "All checks read the filesystem only; the freshness check additionally makes a short, best-effort fetch of the docs manifest (skipped with `--offline`)." Added an `--offline` row to the flag table and an example line. Same fix applied to the README and the PR description bullet. Done in 30f69b724.

packages/cli/README.md
jim761: Fixed here too — same reworded sentence, and added `--offline` to the Flags list. 30f69b724.

packages/cli/src/bin/marigold.ts
jim761: Right — the raw string leaked because `telemetryArgs` is set before validation and `fail()` doesn't bypass the emit. Clamped to your suggestion:

```ts
format: isDoctorFormat(values.format ?? 'text')
  ? (values.format ?? 'text')
  : 'invalid',
```

Added a bin test asserting `doctor --format banana` records `format: 'invalid'`. Kept this scoped to `doctor` for now — happy to open a follow-up for the same pattern in `docs`/`search`/`examples`. 30f69b724.

packages/cli/src/lib/doctor/version.ts
jim761: Fixed — `prerelease: /-/.test(version.split('+')[0])`, so build metadata (`1.0.0+build-5`) no longer parses as a prerelease. Added test cases covering `1.0.0+build-5` vs `1.0.0` and vs `1.0.0-beta.3`. 30f69b724.

packages/cli/src/commands/doctor.test.ts
jim761: Done — both the `theme-passed` and `@source` tests now reuse a single `report()` and derive both `checks(r)` and the specific warning from it. 30f69b724.

packages/cli/src/bin/marigold.ts
jim761: Added a positional guard mirroring `examples`: `marigold doctor foo` now fails with `Usage: marigold doctor (takes no arguments)`. Covered by a new bin test. 30f69b724.

packages/cli/src/lib/doctor/index.ts
jim761: Reworded the comment to match reality — the bin treats it like any other thrown error (prints the message, exits 1); the distinct type stays as a semantic marker for callers embedding doctor, not something the bin `instanceof`-checks. 30f69b724.

aromko:
@jim761 How this can be tested locally?

jim761:
@aromko First uninstall the cli if you already have it installed, then go to the cli folder and type in console "pnpm link -g". to also test the telemetry make a .env.local in the cli folder and add: MARIGOLD_DOCS_URL="https://marigold-docs-git-dst-1265-marigold-doctor-command-marigold.vercel.app/" in there. when youre done type pnpm uninstall @marigold/cli to remove it.

aromko:
> @aromko First uninstall the cli if you already have it installed, then go to the cli folder and type in console "pnpm link -g". to also test the telemetry make a .env.local in the cli folder and add: MARIGOLD_DOCS_URL="[marigold-docs-git-dst-1265-marigold-doctor-command-marigold.vercel.app](https://marigold-docs-git-dst-1265-marigold-doctor-command-marigold.vercel.app/)" in there. when youre done type pnpm uninstall @marigold/cli to remove it.

yeah well seems not to work with the description on my machine. Maybe someone else will try.

jim761:
@aromko one thing i forgot to explain that you need to build the cli first, also to remove it type pnpm uninstall -g @marigold/cli. It's in the readme so claude should be able to tell you how to test

packages/cli/src/lib/doctor/format.ts
```diff
@@ -0,0 +1,117 @@
...
+// and only interactive terminals gain color. picocolors is TTY-aware, so every
+// helper below degrades to plain text automatically when color is unsupported.
+const highlightCode = (text: string): string =>
+  text.replace(/`[^`]+`/g, match => pc.cyan(match));
+
+// A single check can pack several findings into one message joined by "; "
+// (e.g. the Tailwind and React-peer checks). For humans we break those into a
+// bulleted list with an optional "headline:" lead-in. The raw, unsplit message
+// is still preserved verbatim in the JSON report, so this is presentation only.
+const splitFindings = (
```
sarahgm: ⚠️ **`splitFindings` re-parses a string the checks just serialized — and the round-trip has a latent presentation bug.**

The checks build multi-finding messages with `problems.join('; ')` (plus a `'headline: '` prefix) and *also* keep the raw array in `details`. `splitFindings` then reverses that exact transform by splitting on `'; '` / `': '` — an array → string → array round-trip coupled by an implicit string grammar.

Concrete cost: `checkVersionFreshness` joins with `', '` (checks.ts:128-130), **not** `'; '`, so it hits the early return on line 32 and renders multiple out-of-date packages as one comma-run instead of bullets — inconsistent with the tailwind/react-peer checks. The `': '` headline split is also positional, so it mis-parses the first finding that ever contains a colon.

Deeper fix: carry the findings structurally onto `RenderRow` (`findings: string[]` + optional `headline`) and render bullets from that; keep the joined string only for the JSON `message`. (Touches `checks.ts` too, since the tailwind headline currently lives in message prose.)

sarahgm (commented):
## Automated cleanup review (`/simplify`)

Four independent passes — reuse, simplification, efficiency, altitude — over the doctor diff. The architecture is solid: the `tsx-ast.ts` extraction, the freshness cache-only layering, and the provider file-scope approximation are all at the right depth. These are quality cleanups, **none blocking**. Two are worth doing before merge — the `splitFindings` round-trip (which also fixes a freshness-bullets inconsistency) and collapsing the triplicated `text|json` enum; the rest are polish.

---
*Generated with Claude Code review-pr + simplify skills*

packages/cli/src/bin/marigold.ts
```diff
@@ -107,6 +112,11 @@ See https://www.marigold-ui.io for component documentation.
 const isOutputFormat = (v: string): v is OutputFormat =>
   v === 'markdown' || v === 'json' || v === 'plain';
 
+// Defined locally (not imported from ../commands/doctor.js) so the doctor module
+// — and its @babel/parser dependency — stays lazily loaded.
+const isDoctorFormat = (v: string): v is 'text' | 'json' =>
```
sarahgm: ⚠️ **The `text | json` enum has three sources of truth.**

`doctorFormatValues` (commands-spec.ts), this local `isDoctorFormat`, and a second `isDoctorFormat` in commands/doctor.ts. The comment here notes the bin can't import from `doctor.ts` without eagerly loading `@babel/parser` — true, but `commands-spec.ts` is babel-free, so exporting `doctorFormatValues` from there and deriving the guard via `.includes()` (same pattern as `(EXAMPLES_SUBCOMMANDS as readonly string[]).includes(v)`) collapses all three to one source without touching the lazy-load boundary.

packages/cli/src/bin/marigold.ts
```diff
@@ -370,6 +390,35 @@ export const main = async (
         yes: values.yes,
         skipInstall: values['skip-install'],
       });
+    } else if (command === 'doctor') {
+      const { positionals, values } = parseDoctorCommand(rest);
+      // Record only { format }: the pending DST-1600 GDPR/ePrivacy review scopes
+      // doctor telemetry to the output format alone, so --offline is not tracked.
+      // Clamp to a known enum value so an invalid `--format` never leaks the raw
+      // string into telemetry (validation below runs after telemetryArgs is set).
+      telemetryArgs = {
```
sarahgm: 💡 **The telemetry format-clamp is a doctor-only special case bolted onto shared dispatch.**

Doctor clamps an invalid `--format` to `'invalid'` before recording telemetry (DST-1600 rationale), but none of `docs`/`list`/`search`/`examples` do — `--format banana`, `--section`, `--category` all record the raw string straight into telemetry. If clamping unknown enum values is the intended privacy property, it belongs in the shared `telemetryArgs` construction for every enum flag (the enums already live in `commands-spec.ts`), not in one command.

Minor: `values.format ?? 'text'` is recomputed three times here — hoist to one `const format`.

packages/cli/src/lib/doctor/checks.ts
```diff
@@ -0,0 +1,376 @@
...
+    const file = firstExisting(cwd, candidates);
+    if (!file) return false;
+    try {
+      return fs.readFileSync(file, 'utf8').includes(needle);
+    } catch {
+      return false;
+    }
+  };
+  if (project.framework === 'nextjs') {
+    return read(
```
sarahgm: 💡 **Duplicates `init.ts`'s postcss detection.**

This candidate list + `.includes('@tailwindcss/postcss')` is copied verbatim from `commands/init.ts` (~l.160-168). Consider a shared constant next to `CSS_ENTRY_CANDIDATES` in `fs-utils.ts` (or a `hasPostcssTailwindPlugin(cwd)` helper) so the config filenames and plugin name live in one place.

packages/cli/src/lib/doctor/checks.ts
```diff
@@ -0,0 +1,376 @@
...
+  if (!cssEntry) {
+    problems.push('could not find a CSS entry file with Tailwind imports');
+  } else {
+    let css = '';
+    try {
+      css = fs.readFileSync(cssEntry, 'utf8');
+    } catch {
+      /* treat as empty */
+    }
+    if (!/@import\s+["']tailwindcss["']/.test(css)) {
```
sarahgm, line 289: 💡 **These CSS-detection regexes reimplement `edit-css.ts`, and have already diverged.**

`edit-css.ts` already exports `hasImport(contents, target)` / `hasSource(contents, sourcePath)`; this rewrites all three checks as looser inline regexes, so the writer (`init`) and the diagnoser (`doctor`) can now disagree about whether an import is present. Reuse `hasImport`/`hasSource` (generalizing the theme check to the `@marigold/theme-*` prefix).

Also, the `@import 'tailwindcss'` test on this line re-derives what `detectProject` already computed — `ctx.project.tailwindVersion === 4` is set by scanning the same CSS entry for that exact import.

packages/cli/src/lib/doctor/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  for (const check of ALL_CHECKS) {
+    const outcome = check(ctx);
+    if (outcome.status === 'skip') continue;
+    if (outcome.status === 'ok') {
+      passed.push(outcome.title);
+      rows.push({ title: outcome.title, status: 'ok' });
+      continue;
+    }
+    // status === 'issue'
+    const issue: DoctorIssue = {
```
sarahgm, line 49: 💡 **Hand-copies fields out of a value that is already a `DoctorIssue`; and partitions by severity twice.**

The `'issue'` variant of `CheckOutcome` is `{ status; title } & DoctorIssue`, so `outcome` already carries every `DoctorIssue` field:

```ts
const { status, title, ...issue } = outcome; // issue is exactly DoctorIssue
(issue.severity === 'error' ? errors : warnings).push(issue);
```

This also won't silently drop a field (e.g. `details`) when `DoctorIssue` grows. Separately, the loop builds `errors/warnings/passed` *and* a parallel `rows[]`, then `renderText` re-filters `rows` back into the same three groups — the partition happens twice. Building `rows` inside `renderText` from the already-partitioned arrays removes the second pass.

packages/cli/src/commands/doctor.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  // Best-effort: warm the manifest cache so the freshness check has published
+  // versions to compare against on a first run, without the user having to
+  // prime it via another command. The check itself stays cache-only; this only
+  // populates the cache. Never fatal — offline, a slow/failed fetch, or a
+  // manifest without a `packages` field all degrade to the cache-only path,
+  // where freshness simply skips.
+  if (!options.offline) {
+    try {
+      await loadManifest({ timeoutMs: MANIFEST_WARM_TIMEOUT_MS });
```
sarahgm, line 41: 💡 **The docs manifest is fully transformed twice, and the warm-fetch result is discarded.** (low priority — doctor is one-shot)

`loadManifest` here runs the full `transformManifest` (iterates every page, builds + sorts the category map) and the return value is thrown away; then `runDoctorChecks` → `latestVersions()` → `loadManifestSync()` re-reads the same cache file and transforms it a second time — all to read the tiny `manifest.packages` map. Reading just `.packages` from the cached JSON (skipping `transformManifest`) would avoid two O(pages) passes.

Related: `installedVersion`/`readInstalled` re-read+parse the same `node_modules/*/package.json` 2–3× across `checkVersionMatch`/`checkVersionFreshness`/`checkReactPeer`; memoizing behind the `DoctorContext` collapses each to one read.

packages/cli/src/lib/doctor/types.ts
```diff
@@ -0,0 +1,62 @@
+// The `marigold doctor` diagnostic model. The report shape mirrors the sibling
+// `validate` command's ValidationReport (errors/warnings/passed/text) so the two
```
sarahgm: 💡 **Comment references a `validate` command / `ValidationReport` that doesn't exist.**

Grep finds no `validate` command or `ValidationReport` anywhere in the repo — this comment (and the PR description's "Following the `validate` command's policy") is the only mention. Reword to describe the `errors/warnings/passed/text` convention without implying a sibling command to keep in sync.

sarahgm:
@aromko @jim761 

I tested it like this in the current branch: 

> build the current branch, then run with plain node:
> pnpm --filter @marigold/cli build
> node packages/cli/dist/bin/marigold.mjs doctor --format json

looks good to me

packages/cli/src/lib/doctor/format.ts
jim761: Real bug — and it bit twice. Freshness joined with `', '` while the splitter looked for `'; '`, so multiple outdated packages collapsed into one comma-run; and the Tailwind v3 finding contains `'; '` itself (`...requires Tailwind v4; you appear to be on v3`), so a single finding got torn across two bullets.

Fixed by carrying findings structurally onto `RenderRow` (`findings: string[]` + optional `headline`) and rendering bullets from that — `splitFindings` is gone, so there's no array→string→array round-trip anymore. The JSON `message` is unchanged (still the joined form). Added regression tests for both the multi-package and embedded-`; ` cases. cd1c970d6

packages/cli/src/bin/marigold.ts
jim761: Done — `doctorFormatValues` (and a derived `DoctorFormat` type) now live in `commands-spec.ts` (babel-free), exactly as you suggested. Both the bin and `commands/doctor.ts` derive their guard via `(doctorFormatValues as readonly string[]).includes(v)`, so the three declarations collapse to one source and the lazy-load boundary that keeps `@babel/parser` off the hot path stays intact. cd1c970d6

packages/cli/src/bin/marigold.ts
jim761: Hoisted `values.format ?? 'text'` to a single `const format` (was computed three times).

Left the clamp scoped to `doctor` for now. Extending it to `docs`/`list`/`search`/`examples` is the right idea, but it changes the shared `telemetryArgs` construction and needs one thing verified first: that the docs telemetry route accepts a clamped `'invalid'` value for `format` rather than rejecting/dropping the event on its enum validation — otherwise we'd trade a raw-string leak for silent telemetry loss. Tracking it as a separate follow-up so it gets its own review. cd1c970d6

packages/cli/src/lib/doctor/checks.ts
jim761: Done — extracted `POSTCSS_CONFIG_CANDIDATES` and `TAILWIND_POSTCSS_PLUGIN` into `fs-utils.ts` (next to `CSS_ENTRY_CANDIDATES`). `doctor/checks.ts` (`buildWiringOk`) and `commands/init.ts` — including the `POSTCSS_CONFIG` template's plugin name — now reference them, so the config filenames and plugin name live in one place. cd1c970d6

packages/cli/src/lib/doctor/checks.ts
jim761: Holding this one — and I think the divergence is intentional rather than accidental drift. doctor's looseness is protective: matching any `@marigold/theme-*` (not just `theme-rui`) and doing a presence-only `@source … @marigold` check keeps the diagnoser robust to however `init` writes the file. Reusing the exact-literal `hasImport`/`hasSource` would tighten it and risk false positives on valid setups the writer didn't produce verbatim.

On the sub-point: the `@import 'tailwindcss'` check isn't equivalent to `ctx.project.tailwindVersion === 4` either — a project with a `tailwind.config.js` **and** the v4 `@import` reads as v3, because `detectTailwind` short-circuits to v3 on the JS config before it ever scans the CSS. So swapping in the derived flag would misfire. Left as-is.

packages/cli/src/lib/doctor/index.ts
jim761: Leaving as-is. After the freshness fix above, `CheckOutcome`'s `issue` variant carries text-only render hints (`findings`/`headline`), so a `const { status, title, ...issue }` rest-spread would leak those into the JSON `DoctorIssue` unless they're stripped — and stripping the `status` discriminant trips `no-unused-vars`. The second half isn't feasible either: building `rows` inside `renderText` from the partitioned arrays needs a `title`, which `DoctorIssue` doesn't carry (title lives on the outcome, not the issue). The explicit copy stays the clearest option here.

packages/cli/src/commands/doctor.ts
jim761: Leaving as non-blocking, agreed on the low priority. This is per-run cost on a one-shot command that exits in tens of ms, and both parts add caching state (a lighter packages-only manifest read, plus memoized `package.json` reads on `DoctorContext`) for no user-perceptible gain. Happy to revisit if doctor ever moves onto a hot path.

packages/cli/src/lib/doctor/types.ts
jim761: Correct — grep confirms no `validate` command or `ValidationReport` exists anywhere in the repo. Reworded the comment to describe the `errors/warnings/passed/text` convention on its own terms, without implying a sibling command to keep in sync. cd1c970d6

sarahgm (approved):

PR #5631, opened 2026-07-10
feat(DST-1511): add relative date range presets

aromko:
# Description

Adds a `presets` prop to `Calendar`, `RangeCalendar`, `DatePicker`, and `DateRangePicker` that renders quick-select date presets as a listbox rail beside the calendar, per the design handoff. Built-in presets (`today`, `yesterday`, `tomorrow`, `this-week`, `next-7-days`, `next-30-days`, `last-7-days`, `last-30-days`, `this-month`, `this-quarter`) ship with localized labels (de-DE/en-US) and resolve at selection time so relative values never go stale; custom presets take a `label` plus a value or resolver function. The active preset is derived by value equality (no stored state), presets outside `minValue`/`maxValue` or covering unavailable dates render disabled, and selecting a preset keeps the picker overlay open. `useDatePresets`/`useDateRangePresets` are exported for userland compositions.

On small screens the grid renders first with a "Quick selection" row: inline calendars open the preset list in a bottom sheet (tray), while the pickers switch their existing sheet to the list in place with a Back row.

The preset UI is imported statically. An earlier revision lazy-loaded it via `React.lazy`, but the Suspense swap made the preset rail visibly pop in and reflow the popover on first open, so the split was reverted. The Calendar size-limit budget is raised from 72 kB to 120 kB to cover the Tray overlay stack and listbox machinery the presets pull in (the entry measures 113.53 kB). The branch is rebased onto `beta-release` and typechecks with zero errors.

Closes DST-1511

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm sb` and open the `Presets` stories under Components/Calendar, RangeCalendar, DatePicker, and DateRangePicker: verify the rail beside the grid, the check indicator on the active preset, and that selecting a preset jumps the visible month (pickers keep the popover open).
2. Open `PresetsWithMinValue` (Calendar/RangeCalendar): out-of-bounds presets render disabled.
3. Switch to a narrow viewport (e.g. the "Extra Small Screen" toolbar entry) and open the `PresetsMobile` stories: inline calendars open the quick-selection tray (tap applies and closes); in the pickers, the "Quick selection" row switches the sheet to the preset list in place, and Back returns without selecting.
4. Docs: run `pnpm start` and check the "Quick select presets" sections on the four component pages.
5. Bundle: run `pnpm --filter @marigold/components size` — all four entries stay under budget.

## Breaking Changes

No

## Checklist


packages/components/.size-limit.json
```diff
@@ -18,7 +18,7 @@
     "path": "dist/index.mjs",
     "import": "{ Calendar }",
     "ignore": ["react", "react-dom", "react/jsx-runtime"],
-    "limit": "72 kB"
+    "limit": "120 kB"
```
sebald, line 21: 🟡 **Bundle: isolated `Calendar` entry +48 kB (72 → 120 kB), but scoped — corrected after verifying.**

I re-checked this against tree-shaking. The `Calendar` size-limit entry measures `import { Calendar }` from `dist/index.mjs` with react excluded, so this number is *already* tree-shaking-aware — 113.53 kB is the tree-shaken cost of importing just Calendar.

Tree-shaking / per-component output files can't reclaim the preset code here: `Calendar.tsx` statically imports `CalendarPresets` → `Tray` + ListBox, and the `hasPresets ? … : …` is a **runtime** branch, not a static one. A consumer importing `Calendar` without ever passing `presets` still ships it. (The revert of `React.lazy` for the pop-in is what made this static — so it's a real UX-vs-isolated-bundle tradeoff, not an oversight.)

**That said, the blast radius is narrow, so this is not a blocker:**
- Other entries unchanged (Button 27 kB, leaf 13 kB) — non-date-component consumers pay nothing.
- The **Full barrel budget was left at 325 kB** (not raised), i.e. the Tray/ListBox machinery is already counted at barrel scale. In an app already using those components the *marginal* cost of presets is close to zero; most of the 113 kB is shared code counted once.
- The near-full +48 kB only bites someone importing `Calendar` (or the three date components that embed it) in relative isolation.

So: worth a conscious ack of the isolated-Calendar delta, but no action needed unless you specifically care about that isolated-import case (only code-splitting would help, and that was already tried).

packages/components/src/Calendar/CalendarPresets.tsx
```diff
@@ -0,0 +1,508 @@
...
+                    {/* RAC wires the slots to `aria-labelledby`/
+                        `aria-describedby`, so the option's accessible name
+                        stays the label alone (not "Kickoff Aug 1") while
+                        screen readers still announce the resolved date as a
+                        description. */}
+                    <Text slot="label">{item.label}</Text>
+                    <Text
+                      slot="description"
+                      className={cn(
+                        'text-secondary',
```
sebald: 🎨 **Minor — state-driven color belongs in the theme, not the component.**

The values here are correct semantic tokens (`text-secondary` / `text-foreground` / `text-disabled`), but per the repo convention (`CLAUDE.md`: "Don't hardcode styles — use `useClassNames`"; state styling via data-attributes in `*.styles.ts`) this conditional would ideally live in the ListBox/Calendar theme keyed off `data-selected` / `data-disabled`, the way the sibling `.selection-indicator` check already is.

I get that a runtime-conditional `cn()` on a nested `<Text>` is awkward to express in `cva`, so I'd treat this as a nit — but it's the one spot that drifts from "styling lives in the theme."

sebald (commented):
## Automated Code Review

Strong PR overall — idiomatic React 19, clean React Aria usage, **zero `useEffect`** (selection is derived from calendar state rather than mirrored + reconciled, which is exactly right), and unusually thorough inline docs & tests. Nothing blocks correctness.

Most of my feedback is attached **inline** at the relevant lines. Summary of themes:

- ⚠️ **Bundle size** — the `Calendar` entry budget jumps 72 kB → 120 kB (+66%) and the preset UI is now statically imported, so *every* Calendar consumer pays for it. Worth a conscious team sign-off (see inline on `.size-limit.json`).
- 🎨 **Design tokens** — one spot places state-driven color logic in the component instead of the theme (inline on `CalendarPresets.tsx`). Values are correct semantic tokens, so this is a nit. No z-index rules violated (Tray owns its layer).
- 🔁 **/simplify** — the diff is already close to minimal. The one real candidate is the `content` layout block now duplicated verbatim across `Calendar.tsx` and `RangeCalendar.tsx` (inline). This predates the PR — good follow-up, not a blocker.
- 🧩 **New `presets` API** — well-judged. The `['today', 'tomorrow']` common case is excellent and consistent across all four components; the object escape hatch reads well. Two subtleties users will hit are noted inline (single-date vs range key split; when to pass a `value` resolver function).

Process note: the checklist item **"Visual regression tests updated"** is unchecked on a UI-heavy change — that's the one CI gap I'd close before merge (I have not triggered Chromatic).

Recommendation: **Comment** (approve-leaning once bundle-size tradeoff is acknowledged and VRT is run).

---

packages/components/src/Calendar/Calendar.tsx
```diff
@@ -101,46 +118,73 @@ const _Calendar = ({
...
-    return (
-      <CalendarContext
-        value={{
-          classNames,
-          visibleMonths,
-          minValue,
-          maxValue,
-          disabled,
-        }}
+  const content = isMultiMonth ? (
```
sebald, line 132: 🔁 **/simplify (follow-up, not blocking): this `content` block is duplicated verbatim in `RangeCalendar.tsx`.**

The multi-month + single-month layout JSX is now identical (~40 lines) in both files. Hoisting it into a `content` variable is already an improvement; the natural next step is a shared `<CalendarBody>` next to `CalendarPresetsShell` so the two callers just render it.

The duplication predates this PR, so I'd file it as a follow-up rather than expand scope here.

packages/components/src/RangeCalendar/RangeCalendar.tsx
```diff
@@ -175,6 +205,64 @@ const _RangeCalendar = <T extends DateValue>({
     [error]
   );
 
+  const content = isMultiMonth ? (
```
sebald, line 208: 🔁 Same duplicated `content` block as `Calendar.tsx:132` — see the note there. A shared `<CalendarBody>` would let both files drop this.

packages/components/src/Calendar/presets.ts
```diff
@@ -0,0 +1,167 @@
...
+// ---------------
+// Re-exported so consumers can type custom range presets without also
+// depending on react-aria-components directly.
+export type { DateRange };
+
+export type BuiltInDatePresetKey = 'today' | 'yesterday' | 'tomorrow';
+
+// Day presets are also valid range presets: they resolve to a single-day
+// range (start === end === the day), see `builtInDateRangePresets` below.
+export type BuiltInDateRangePresetKey =
```
sebald, line 23: 🧩 **API — the single-date vs range key split is invisible at the call site.**

`Calendar`/`DatePicker` accept only `today | yesterday | tomorrow`, while the reporting-style keys (`this-week`, `last-30-days`, …) exist only here on the range components. This is logically correct (a week isn't a single day) and TS enforces it — but the error a user hits (`presets={['this-week']}` rejected on a `Calendar`) reads as "why can't my date picker do this-week?"

Suggestion: call this asymmetry out explicitly in the four MDX pages so it's discoverable before the type error.

packages/components/src/Calendar/presets.ts
```diff
@@ -0,0 +1,167 @@
...
+  id?: string;
+  label: string;
+  /**
+   * The value the preset selects. Pass a function to resolve the value at
+   * selection time, so relative presets stay correct in long-lived views.
+   * Resolvers also run on every render of the preset list (to derive the
+   * active and disabled options and the small-screen sublabels), so they
+   * must be pure and cheap.
+   */
+  value: T | (() => T);
```
sebald, line 48: 🧩 **API — the `T | (() => T)` resolver is powerful but subtle; and `id` defaulting to `label` is a quiet foot-gun.**

The "pass a function so relative presets resolve at selection time" concept is great, but it's easy to miss when to reach for it. A one-liner in the MDX ("pass a function if the value is relative to *now*; a plain value otherwise") would save a support question.

Separately, `id ?? label` (line 39) means two custom presets sharing a `label` collide silently and make the active/selected state ambiguous. It's documented in the JSDoc — fine — just flagging it as the kind of thing worth a sentence in the docs too.

docs/content/components/form/datepicker/index.mdx
```diff
@@ -75,6 +75,25 @@ Invalid dates or unrecognized formats are ignored, ensuring data integrity.
 
 <ComponentDemo name="datepicker-paste" />
 
+### Quick select presets
+
+Use the `presets` prop to offer common dates as a one-click list beside the calendar in the popover, such as "Today" or "Tomorrow" for a due date. Built-in keys ship with localized labels and correct date math, and custom presets take a `label` plus a date value or a resolver function. Presets that fall outside `minValue`/`maxValue` or land on an unavailable date are disabled.
```
sebald, line 80: 💡 **Docs suggestion.** Two clarifications would smooth adoption of the new prop (applies to the equivalent section on the other three component pages too):

1. Note that `DatePicker`/`Calendar` only take the single-date built-ins (`today`/`yesterday`/`tomorrow`); the range keys like `this-week` are `DateRangePicker`/`RangeCalendar` only.
2. For custom presets, mention *when* to pass `value` as a function (relative-to-now dates) vs a plain value (fixed dates).

packages/components/.size-limit.json
aromko: Acknowledged, and intentional: the `React.lazy` split was tried first and reverted (6f5e843) because the preset rail popping in on first open caused a visible reflow. The 120 kB isolated-Calendar budget was set with that tradeoff in mind; the barrel budget stays at 325 kB.

packages/components/src/Calendar/Calendar.tsx
aromko: Agreed, keeping it out of this PR. Filed [DST-1627](https://reservix.atlassian.net/browse/DST-1627) to extract a shared `<CalendarBody>` next to `CalendarPresetsShell`.

packages/components/src/RangeCalendar/RangeCalendar.tsx
aromko: Agreed, keeping it out of this PR. Filed [DST-1627](https://reservix.atlassian.net/browse/DST-1627) to extract a shared `<CalendarBody>` next to `CalendarPresetsShell`.

packages/components/src/Calendar/CalendarPresets.tsx
aromko: Moved to the theme: the `calendarPresets` slot in `Calendar.styles.ts` now styles `[slot=description]` via descendant selectors keyed on `data-selected`/`data-disabled`, same pattern as the `.selection-indicator` rules on `item`. The conditional `cn()` is gone from `CalendarPresets.tsx` and the ListBoxItem child simplified from a render function to plain children (the render props were only used for the colors). Lands with the next push.

packages/components/src/Calendar/presets.ts
aromko: Called out in the docs now: the preset sections on all four component pages state which built-ins apply where (single-date keys on `Calendar`/`DatePicker`; range keys plus the one-day built-ins on `DateRangePicker`/`RangeCalendar`) and cross-link each other, so the split is discoverable before the type error. Lands with the next push.

packages/components/src/Calendar/presets.ts
aromko: Both documented now on all four pages: when to pass `value` as a function (relative-to-now dates, resolved at selection time) vs a plain value (fixed dates), and that `id` defaults to `label`, so custom presets sharing a label collide and need explicit unique ids. Lands with the next push.

docs/content/components/form/datepicker/index.mdx
aromko: Added both clarifications to the preset sections of all four component pages (datepicker, calendar, daterangepicker, rangecalendar): the single-date vs range built-in split with cross-links, and the function-vs-plain `value` guidance including the unique-`id` note. Lands with the next push.

sebald (approved):

PR #5636, opened 2026-07-13
fix(DST-1621): keep Accordion stickyHeader pinned when the header has actions

sarahgm:
# Description

`Accordion`'s `stickyHeader` did not keep the header pinned when the header carried action buttons. Actions were composed by wrapping `Accordion.Header` in a layout component, which collapsed the sticky wrapper's containing block down to the short header row, so the header scrolled away with the content.

`Accordion.Header` now accepts an `actions` prop, so the header row (title plus actions) is owned by the component and the sticky wrapper stays a direct child of the item. The header now stays pinned together with its actions while the item content scrolls. The sticky fill also changes from `bg-background/90` with a backdrop blur to `bg-surface/90`, matching the Table sticky header so the pinned header no longer shows a grey tint. The docs "Long-form content" section is updated and a new sticky demo is added.

Closes DST-1621

## Screenshots / Preview

| Before | After |
| ------ | ----- |
| Before: the header scrolls away when the item content scrolls | After: the header stays pinned together with its Delete/Edit actions |

<sub>`Components/Accordion/StickyHeader`, viewport 1000×720, scrolled to `scrollY=400` inside the expanded item. Before: sticky wrapper sits at `top: -331px` (scrolled off, `bg-background/90` + blur). After: `top: 0px` (pinned, `bg-surface/90`).</sub>

## Test Instructions

1. Open Storybook and navigate to `Components/Accordion/StickyHeader`.
2. Shrink the viewport so the content scrolls, then scroll through an expanded item. The header, with its Delete and Edit actions, stays pinned to the top.
3. Confirm the pinned header uses the surface colour (no grey tint) and does not scroll horizontally.
4. Run `pnpm start`, open the Accordion docs page, and check the "Long-form content" sticky demo.

## Breaking Changes

No

## Checklist

### DIFF
--- a/packages/components/src/Accordion/AccordionHeader.tsx
+++ b/packages/components/src/Accordion/AccordionHeader.tsx
@@ -8,39 +8,55 @@ import { useAccordionContext } from './AccordionContext';
 
 export interface AccordionHeaderProps {
   children?: ReactNode;
+  /**
+   * Content rendered next to the title, inside the same header row (e.g. action
+   * buttons). Use this instead of wrapping `Accordion.Header` in a layout
+   * component, otherwise `stickyHeader` cannot keep the header pinned.
+   */
+  actions?: ReactNode;
 }
 
-export const AccordionHeader = ({ children }: AccordionHeaderProps) => {
+export const AccordionHeader = ({
+  children,
+  actions,
+}: AccordionHeaderProps) => {
   const { classNames, stickyHeader, iconPosition } = useAccordionContext();
   const { isExpanded } = use(DisclosureStateContext)!;
 
+  const heading = (
+    <Heading slot={noSlot} className={actions ? 'min-w-0 flex-1' : undefined}>
+      <Button slot="trigger" className={classNames.header}>
+        {iconPosition === 'left' && (
+          <MorphCaret
+            size="16"
+            expanded={isExpanded}
+            className={classNames.icon}
+          />
+        )}
+        <div className="flex-1 items-center">{children}</div>
+        {iconPosition === 'right' && (
+          <MorphCaret
+            size="16"
+            expanded={isExpanded}
+            className={classNames.icon}
+          />
+        )}
+      </Button>
+    </Heading>
+  );
+
   return (
     <div
-      className={
-        stickyHeader
-          ? 'bg-background/90 sticky top-0 z-1 backdrop-blur-xs'
-          : undefined
-      }
+      className={stickyHeader ? 'bg-surface/90 sticky top-0 z-1' : undefined}
     >
-      <Heading slot={noSlot}>
-        <Button slot="trigger" className={classNames.header}>
-          {iconPosition === 'left' && (
-            <MorphCaret
-              size="16"
-              expanded={isExpanded}
-              className={classNames.icon}
-            />
-          )}
-          <div className="flex-1 items-center">{children}</div>
-          {iconPosition === 'right' && (
-            <MorphCaret
-              size="16"
-              expanded={isExpanded}
-              className={classNames.icon}
-            />
-          )}
-        </Button>
-      </Heading>
+      {actions ? (
+        <div className="flex items-center">
+          {heading}
+          <div className="shrink-0">{actions}</div>
+        </div>
+      ) : (
+        heading
+      )}
     </div>
   );
 };
--- a/packages/components/src/Accordion/AccordionHeader.tsx
+++ b/packages/components/src/Accordion/AccordionHeader.tsx
@@ -2,6 +2,9 @@ import { ReactNode, use } from 'react';
 import { Button } from 'react-aria-components/Button';
 import { DisclosureStateContext } from 'react-aria-components/Disclosure';
 import { Heading } from 'react-aria-components/Heading';
+import { Provider } from 'react-aria-components/slots';
+import { cn } from '@marigold/system';
+import { ButtonContext } from '../Button/Context';
 import { MorphCaret } from '../icons/MorphCaret';
 import { noSlot } from '../utils/noSlot';
 import { useAccordionContext } from './AccordionContext';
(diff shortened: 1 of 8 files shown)


sebald:
### Consider making `Accordion.Header` actions slot-aware (match `Panel.Header`)

The `actions` **prop** is the right mechanism here — the whole header row is a single `<button slot="trigger">`, so actions have to be injected as a sibling of the trigger rather than composed as children the way `Panel.Header` does. 👍

What's missing is the **context cascade**. `Panel.Header` wraps its actions region in a `ButtonContext` provider so a bare `<Button>` automatically adopts the header treatment (ghost variant, small size, positional className) — see `PanelHeader.tsx` (`actionProps`, ~L59-67). Here `{actions}` is rendered raw, so every call site has to style the buttons by hand — and that inconsistency already shows up in this PR:

- `accordion-sticky.demo.tsx` uses `<Button size="small">`
- `Accordion.stories.tsx` uses default-size `<Button>`

Same UX role, two different looks. A cascade removes that decision from the consumer and makes `<ButtonGroup>` / `<ActionMenu>` behave uniformly — consistent with the **Familiarity** design principle (align with existing patterns).

Proposed change — `Button` already consumes Marigold's `ButtonContext` via `useContextProps`, so it's low-effort. Keep the `classNames.actions` wrapper for layout and add a provider around the content:

```tsx
import { Provider } from 'react-aria-components';
import { ButtonContext } from '../Button/Context';

// …the actions cell in the render:
<div className={classNames.actions}>
  <Provider values={[[ButtonContext, { variant: 'ghost', size: 'small' }]]}>
    {actions}
  </Provider>
</div>
```

…and drop the manual `size="small"` from the demo/story.

Two caveats worth calling out:

1. It's a **design** decision as much as a refactor — ghost/low-emphasis changes the current full-emphasis look in the story. Probably worth a quick confirm with design; consistency with Panel argues for ghost.
2. If actions can contain overlays (e.g. `ActionMenu`), reset the cascade at the portal the way Panel relies on (`ResetButtonContext`) so the ghost styling doesn't leak into the menu.

Not a blocker on the sticky fix itself — could land here or as a fast follow-up.

packages/components/src/Accordion/AccordionHeader.tsx
```diff
@@ -8,39 +8,55 @@ import { useAccordionContext } from './AccordionContext';
...
-      </Heading>
+      {actions ? (
+        <div className="flex items-center gap-2">
+          {heading}
+          <div className={classNames.actions}>{actions}</div>
+        </div>
+      ) : (
+        heading
+      )}
     </div>
```
sebald, line 77: **Optional cleanup:** when `actions` exist, the sticky `<div>` wraps a second flex `<div>`. You can fold the flex onto the sticky div with `cn()` and drop one DOM node plus the wrapper-returning ternary:

```tsx
import { cn } from '@marigold/system';

return (
  <div
    className={cn(
      stickyHeader && 'bg-surface/90 sticky top-0 z-1',
      actions && 'flex items-center gap-2'
    )}
  >
    {actions ? (
      <>
        {heading}
        <div className={classNames.actions}>{actions}</div>
      </>
    ) : (
      heading
    )}
  </div>
);
```

packages/components/src/Accordion/AccordionHeader.tsx
```diff
@@ -8,39 +8,55 @@ import { useAccordionContext } from './AccordionContext';
...
-export const AccordionHeader = ({ children }: AccordionHeaderProps) => {
+export const AccordionHeader = ({
+  children,
+  actions,
+}: AccordionHeaderProps) => {
   const { classNames, stickyHeader, iconPosition } = useAccordionContext();
   const { isExpanded } = use(DisclosureStateContext)!;
 
+  const heading = (
+    <Heading slot={noSlot} className={actions ? 'min-w-0 flex-1' : undefined}>
```
sebald: **Minor:** this ternary can be unconditional. When there are no `actions` the `<Heading>` isn't a flex child, so `flex-1`/`min-w-0` are inert — `className="min-w-0 flex-1"` behaves identically and drops the conditional.

sebald (commented):
Two small, non-blocking cleanups on `AccordionHeader.tsx` — the sticky fix itself looks correct and is well-tested. (See also the top-level comment about making the header actions slot-aware like `Panel.Header`.)

packages/components/src/Accordion/AccordionHeader.tsx
sarahgm: Done in e221b3fa9. The `Heading` className is now the unconditional `min-w-0 flex-1`.

packages/components/src/Accordion/AccordionHeader.tsx
sarahgm: Done in e221b3fa9. Folded the flex onto the sticky `<div>` with `cn()` and dropped the wrapper `<div>` plus the ternary.

sarahgm:
Addressed the slot-aware suggestion in e221b3fa9. `Accordion.Header` now cascades a `ButtonContext` (`variant: 'ghost'`, `size: 'small'`) around the actions cell, matching `Panel.Header`, so a bare `<Button>` adopts the header treatment. I dropped the manual `size="small"` from the sticky demo, so the demo and story now render identically.

On the two caveats you flagged:

- **Overlays:** no extra work needed. Every overlay (`Popover`, `Modal`, `Tray`, `Drawer`) already wraps its content in `ResetButtonContext`, and `ActionMenu` renders through a `Popover`, so the cascade cannot leak into a menu's buttons.
- **Design:** the treatment matches `Panel.Header` (Familiarity), so I am happy to have design react to the Storybook preview if the ghost emphasis needs a tweak.

One deliberate divergence from Panel: I kept the `classNames.actions` wrapper `<div>` instead of folding a positional className into the context. The Accordion title is itself the `<button slot="trigger">`, actions are often wrapped in a non-consuming layout like `<Inline>`, and the row is flex not grid, so the layout class belongs on the cell rather than the individual buttons.

sebald (approved):

PR #5640, opened 2026-07-14
docs(DST-1383, DST-1384): add the Bulk Actions pattern and working example

sebald:
# Description

Adds the **Bulk Actions pattern** — the docs page (DST-1383) and its capstone **working example** (DST-1384) — in one PR. Marigold ships the primitives (`ActionBar`, `Table` selection, `Drawer`, `Dialog`/`useConfirmation`, `Toast`) but the end-to-end pattern of selecting many records and acting on all of them was undocumented and lived as tribal knowledge. This PR composes those primitives into one source of truth and bakes in the design decisions Reservix has settled on. (#5641 was folded into this PR.)

## Pattern page (`/patterns/user-input/bulk-actions`)

- **Current:** no bulk-actions pattern doc; teams reinvented selection scope, destructive-confirm, bulk-edit, and partial-failure handling per feature.
- **New:** a single pattern page walking the flow in build order — when (not) to use, structure, choosing actions (including how bulk and per-row actions coexist by taking turns), selection and its visible-scope boundary, direct actions, confirming destructive actions, multi-field bulk edit, progress feedback, and result/partial-failure handling — with an anatomy diagram, seven inline demos, Do/Don't guidance, an accessibility section, and an implementation section covering the load-bearing practices.

## Working example (`/examples/bulk-actions`)

The page the pattern doc's `<DemoLinks>` points to. It wires the whole flow together as one real product surface, following the filter example's architecture (stateless API route + TanStack Query + nuqs URL state + client-owned session mutations).

- A paginated, searchable events table (42 fixture records, content-derived column widths) with page-bounded selection and a floating `ActionBar` (Publish, Edit, Export, overflow menu with Archive/Send reminders, separated Delete).
- **Partial-failure retry loop:** publish fails deterministically for events without a venue — failed records stay selected as the retry list, and assigning a venue via the bulk-edit `Drawer` (mixed values, explicit price opt-in, re-keyed per selection) makes the retry succeed, closing the loop the doc describes.
- **Per-row action menus** (Publish, Archive, Send reminder, Delete) so single-record and bulk operations can be compared side by side; row menus disable while a selection is active, demonstrating the "take turns" guidance.
- Delete confirms with exact count and reservation impact. Export shows button loading; reminders show a running count in a live region.
- **Pagination per the Pagination docs:** results indicator left, pager centered, "Results per page" quantity selector (10/20/30) right, with the default text strings.
- Search/filter/page/page-size persist in the URL and clear the selection through wrapped setters.
- **Data layer:** `/api/events` (list) and `/api/events/bulk` (per-record authorizer) stay stateless; the client resends its session overlay (overrides + deletions) with every request and folds it into the query key, so filters and refetches always see the visitor's world — no manual cache invalidation anywhere.

Docs-only change: no component, system, or theme code is touched.

Closes DST-1383
Closes DST-1384

## Screenshots / Preview

Docs preview (Vercel) once the deploy is ready:

- **Patterns → User Input → Bulk Actions** (`/patterns/user-input/bulk-actions`)
- **Examples → User Input → Bulk Actions** (`/examples/bulk-actions`)

## Test Instructions

**Pattern page**

1. Navigate to **Patterns → User Input → Bulk Actions** and confirm the page renders: anatomy diagram, the seven inline demos, Do/Don't tiles, and the code blocks with their highlighted lines.
2. Exercise the inline demos: select rows to reveal the ActionBar; run a destructive delete and confirm the dialog names the count/impact; open the bulk-edit drawer and check mixed-value handling; trigger the partial-failure demo and confirm failed rows stay selected.
3. Verify cross-links resolve: ActionBar `#scope`, Drawer `#editing-multiple-fields`, and Table Records inbound anchors still land correctly.

**Working example**

4. Open `/examples/bulk-actions`. Select the two "No venue" drafts plus one regular draft → **Publish** → warning toast "1 of 3 events published", the two failures stay selected.
5. **Edit** → venue shows the mixed-values handling, price shows the explicit opt-in → assign a venue → **Apply to 2 events** → re-select both → **Publish** → success.
6. Select rows with reservations → **Delete** → dialog names the count and summed reservations; confirm removes them.
7. Select all on the page → overflow menu → **Send reminders** → running count "Sending 4 of 10…" replaces the bar actions.
8. Open a row's **⋮ menu** → Publish/Archive/Send reminder/Delete act on that one named event (a venue-less publish explains itself, delete confirms with that row's reservations). Select any row → all row menus disable while the bar is up.
9. Change page / page size / search / status filter with rows selected → the selection clears each time. Filter "On sale" after publishing → freshly published events appear.
10. Change **Results per page** to 20 → resets to page 1, indicator reads "Showing 1 - 20 of 42", URL carries `pageSize=20`.
11. Search gibberish → empty state with "Clear all filters". Press **Reset demo** → the original 42 records return.
12. Reload with `?status=Draft&page=2` → the view restores from the URL.
13. `pnpm --filter @marigold/docs test` → events-query, events routes, and navigation tests pass.

## Breaking Changes

No

## Checklist

- [ ] Stories added/updated (with `component-test` tag where applicable) — N/A, docs-only
- [x] Unit tests added/updated (query pipeline, both routes, navigation contract)
- [ ] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) (for new/changed interactive components) — N/A, composes existing components; live regions and labels verified in-browser
- [ ] Visual regression tests updated (for UI changes) — N/A, no component changes
- [x] Changesets added (`pnpm changeset`, one per ticket)

## Notes for reviewers

Two component-level observations surfaced while verifying (out of scope here, no component changes in this PR):

1. `useConfirmation`'s `autoFocusButton: 'cancel'` leaves focus on the dialog section rather than the Cancel button (RAC FocusScope appears to win over the button's `autoFocus`). Enter is a no-op in that state, so the safe path holds, but it doesn't match the documented intent.
2. A persistent warning toast (bottom-right, `z-80`) can overlap a Drawer's action buttons and intercept clicks — the stacking convention working as specified, but worth a design look for the toast-over-drawer case.


aromko (commented):
## Automated Code Review

### Overview

| Field | Value |
|-------|-------|
| Branch | `dst-1383-bulk-actions-pattern` → `beta-release` |
| Files Changed | 43 (+4,177 / −6) |
| CI | All green ("Vercel Preview Comments" failure is the bot integration, not a code check) |
| Visual Regression | Not applicable (base: `beta-release`; docs-only change) |

### Linked Tickets

**DST-1383** — pattern doc (In Review) · **DST-1384** — working example (Done)

Both tickets' scope is fully covered; all seven locked anchor decisions from DST-1383 (page-bounded selection, confirm-with-impact, sync-with-progress-fallback, honest mixed values, explicit opt-in for high-stakes fields, count-labeled save, failed-records-stay-selected) are implemented and documented. Two deliberate, defensible deviations from the ticket text: the anatomy ships as a theme-aware TSX component (matching `actionbar-anatomy.tsx`) instead of a design-owned SVG, and the example upgrades DST-1384's stub suggestion to a real stateless API + session-overlay architecture mirroring the filter example.

### Checklist

- **TypeScript** ✅ — no `any`, consistent `import type`, `unknown` at the JSON/CSV boundaries, zod on both routes
- **React** ✅ — React 19 idioms (`use(Context)`, context-as-provider), memoized context values, lazy state init, hoisted formatters; no performance findings
- **Testing** ✅ — Vitest, no `fireEvent`/`getByTestId`, clean AAA, navigation contract test updated, query pipeline + both routes covered
- **Accessibility** ✅ — live regions for progress and the drawer's change summary, `VisuallyHidden` summaries, per-row `aria-label`s, cancel-focused destructive dialogs
- **API hardening** ✅ — session size cap + zod, `delay` clamped, ids capped, malformed input degrades gracefully

### Findings (details in inline comments)

- ⚠️ `BulkEditDrawer` runs the edit mutation on a second `useBulkActions()` instance, so the bar's `busy` flag never covers it — the "one operation at a time" guarantee doesn't hold for bulk edit
- 💡 "Clear all filters" leaves stale text in the uncontrolled `SearchField` (same behavior exists in the filter example)
- 💡 Only the single-verb demo mounts the `ToastProvider` all seven pattern-page demos depend on — worth a comment marking the hidden dependency
- 💡 `parseSession` is a non-handler export from a route file; it belongs in `events-query.ts`
- 💡 `escapeCsv` could guard CSV formula injection for a copy-paste reference
- 💡 `pageSize` from the URL is unclamped and sizes the skeleton
- 💡 The demos and the example use different `Text` APIs for the same muted helper text
- 💡 One sentence in the pattern MDX uses em dashes; existing pattern docs contain none
- 💡 `badge: alpha` — DST-1383 says the badge promotes once the example lands, and it lands in this PR; intentional bake-in?

### Recommendation

Non-blocking. High-quality PR: green CI, well-tested data layer, hardened demo API, strong accessibility composition, docs faithful to every locked design decision. The one finding genuinely worth addressing is the `busy` flag gap, since the example exists to be copied and its serialization claim doesn't currently hold.

---

docs/app/(examples)/examples/bulk-actions/BulkEditDrawer.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+export const BulkEditDrawer = ({
+  affected,
+  disabled,
+}: {
+  affected: Event[];
+  /** Disabled while a sibling bulk operation runs (see BulkActionBar). */
+  disabled?: boolean;
+}) => {
+  const actions = useBulkActions();
```
aromko, line 146: ⚠️ **`busy` doesn't cover the bulk-edit mutation**

This is a second `useBulkActions()` instance. TanStack Query's `useMutation` pending state is per-hook-instance, so while this drawer's `updateMutation` is in flight, `BulkActionBar`'s `busy` flag stays `false` and Publish/Export/Delete remain enabled — contradicting the "One operation at a time … so two operations cannot race on the same selection" comment in `useBulkActions.ts`. The window is tiny against the local API, but against a real backend the race is real, and this example is explicitly a copy-paste reference.

Simplest fix: pass the bar's `actions.applyEdit` (and `busy`) in as props instead of re-invoking the hook here.

docs/app/(examples)/examples/bulk-actions/Toolbar.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <Inline space="related" alignX="between" alignY="center">
+      <Inline alignY="input" space="related">
+        <SearchField
+          aria-label="Search events"
+          description="Search by name or venue"
+          width={64}
+          autoComplete="off"
+          defaultValue={search}
```
aromko: 💡 **Stale search text after "Clear all filters"**

The empty state's "Clear all filters" resets the URL state, but this field is uncontrolled (`defaultValue`), so the typed query stays visible while the table shows unfiltered rows. Re-keying with `key={search}` syncs it (with the tradeoff that the field remounts and drops focus on submit). Note the filter example has the identical behavior — if this is worth fixing, both places should change together.

docs/content/patterns/user-input/bulk-actions/bulk-actions-single-verb.demo.tsx
```diff
@@ -0,0 +1,102 @@
...
+    addToast({
+      title: `${scope} ${status === 'On sale' ? 'put on sale' : 'paused'}`,
+      variant: 'success',
+    });
+    setSelected(new Set()); // [!code highlight]
+  };
+
+  return (
+    <>
+      <ToastProvider position="bottom-right" />
```
aromko, line 64: 💡 **This is the only `ToastProvider` on the pattern page**

The confirm, edit-drawer, progress, and partial-failure demos all fire toasts through this region via the global queue. It works because all seven demos mount together, but removing or reordering this demo would silently kill toasts for the rest. A short comment would prevent that trap:

```suggestion
      {/* The one toast region for this page: the other bulk-actions demos
          enqueue into the same global queue and render through it. */}
      <ToastProvider position="bottom-right" />
```

docs/app/api/events/route.ts
```diff
@@ -0,0 +1,70 @@
...
+
+const status = (value: string | null): EventStatus | undefined =>
+  eventStatuses.find(candidate => candidate === value);
+
+// Session payloads are small (a demo visitor touches at most the 42 fixture
+// rows); the size cap just keeps the parser from chewing on garbage. Anything
+// invalid degrades to "no session" instead of failing the request.
+const MAX_SESSION_LENGTH = 8192;
+
+export const parseSession = (
```
aromko: 💡 **Non-handler export from a route file**

Next.js reserves route-file exports for handlers and segment config — this builds today, but it's fragile across Next upgrades. `parseSession` fits naturally next to `eventsSessionSchema` in `events-query.ts`, where it would also gain direct unit coverage.

docs/app/(examples)/examples/bulk-actions/csv.ts
```diff
@@ -0,0 +1,46 @@
+import type { Event } from '@/lib/data/events';
+
+const escapeCsv = (value: unknown) => {
```
aromko, line 3: 💡 **Consider guarding CSV formula injection**

Quoting and comma handling look right. For a reference implementation teams copy, consider also prefixing values that start with `=`, `+`, `-`, or `@` (e.g. with `'`) — real apps will export user-entered event names into Excel, where those execute as formulas.

docs/app/(examples)/examples/bulk-actions/hooks/useEventsParams.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  const setPage = (next: number | null) => {
+    clearSelection();
+    return setPageState(next);
+  };
+
+  return [page, setPage] as const;
+};
+
+export const usePageSize = () => {
```
aromko, line 33: 💡 **`pageSize` from the URL is unclamped**

`?pageSize=5000` renders a 5,000-row skeleton on initial load (`EventsTableSkeleton rows={pageSize}`). Clamping the parsed value to the quantity-selector set (10/20/30) — or a max — in this hook would close it.

docs/content/patterns/user-input/bulk-actions/bulk-actions-edit-drawer.demo.tsx
```diff
@@ -0,0 +1,229 @@
...
+              onChange={setPrice}
+              description={
+                sharedPrice === undefined
+                  ? 'Selected events currently have different prices.'
+                  : undefined
+              }
+              formatOptions={{ style: 'currency', currency: 'EUR' }}
+            />
+          </Stack>
+          <Text size="sm" color="secondary" aria-live="polite">
```
aromko: 💡 **Two `Text` APIs for the same muted helper text**

The demos use `size="sm" color="secondary"` while the working example's `BulkEditDrawer` uses `fontSize="sm" variant="muted"` for the same element. Both are valid, but aligning the two near-identical teaching artifacts on one vocabulary (at least `variant="muted"` over a raw `color`) would avoid readers wondering which to copy. Same applies to `bulk-actions-partial-failure.demo.tsx`.

docs/content/patterns/user-input/bulk-actions/index.mdx
```diff
@@ -0,0 +1,303 @@
...
+
+Name every action with a verb, or a verb plus noun where the target is ambiguous, so its effect is obvious before pressing: "Publish", "Assign category", "Delete". Avoid vague labels such as "Apply" or "OK" on their own. Order the actions from safest to most consequential and place a destructive action such as delete last, visually separated from the routine ones. When a selection carries more actions than fit comfortably, move the long tail into an [`<ActionMenu>`](/components/overlay/action-menu) rather than crowding the bar.
+
+While the bar is visible, the selection is the object the user works on. Avoid competing per-row actions during that time: a row-level delete next to a bulk delete makes it unclear which scope a press affects. Bulk and per-row actions can still share a table — one-off operations belong in a row menu (see [Table Records](/patterns/user-input/table-record-management)), repetitive ones in the bar — as long as they take turns: while rows are selected, disable the row actions so the bar holds the only pressable scope.
```
aromko: 💡 **Em dashes — the other pattern docs contain none**

```suggestion
While the bar is visible, the selection is the object the user works on. Avoid competing per-row actions during that time: a row-level delete next to a bulk delete makes it unclear which scope a press affects. Bulk and per-row actions can still share a table when they take turns: one-off operations belong in a row menu (see [Table Records](/patterns/user-input/table-record-management)), repetitive ones in the bar, and while rows are selected the row actions disable so the bar holds the only pressable scope.
```

docs/content/patterns/user-input/bulk-actions/index.mdx
```diff
@@ -0,0 +1,303 @@
+---
+title: Bulk Actions
+description: Let users select many records and act on all of them at once.
+badge: alpha
```
aromko: 💡 **`badge: alpha` — still intentional?**

DST-1383 notes the badge "promotes to stable once the working example lands," and the example lands in this same PR. Keeping alpha for bake-in time is reasonable — just confirming it's a choice, not an oversight.

docs/app/(examples)/examples/bulk-actions/BulkEditDrawer.tsx
sebald: Fixed in d4f909d4. Rather than prop-drilling `applyEdit`/`busy` into the drawer (that trips Next's client-entry "props must be serializable" rule on the exported `BulkEditDrawer`), `useBulkActions` is now a `BulkActionsProvider` mounted once on the page. The bar and the drawer read that single instance, so the edit mutation is covered by `busy` and the "one operation at a time" guarantee actually holds. It mirrors the existing `useSession`/`useSelection` context pattern.

docs/app/(examples)/examples/bulk-actions/Toolbar.tsx
sebald: Fixed in **both** places, as you suggested. The field is now controlled with a render-time sync (no `useEffect`, no re-key), so "Clear all filters" resets it and focus survives submit. Applied to the bulk-actions `Toolbar` and the filter `Toolbar` together so the two examples stay aligned.

docs/content/patterns/user-input/bulk-actions/bulk-actions-single-verb.demo.tsx
sebald: Done — added the comment above the `ToastProvider`.

docs/app/api/events/route.ts
sebald: Moved `parseSession` (and its size cap) into `events-query.ts` beside `eventsSessionSchema`; the route now imports it and exports only its `GET` handler. Added direct unit tests too (valid / null / malformed JSON / bad shape / oversized).

docs/app/(examples)/examples/bulk-actions/csv.ts
sebald: Added the guard — values starting with `=`, `+`, `-`, `@` (plus tab/CR) are now prefixed with a single quote so a user-entered event name stays literal text in Excel/Sheets.

docs/app/(examples)/examples/bulk-actions/hooks/useEventsParams.ts
sebald: Fixed — introduced `PAGE_SIZES` as the single source of the offered sizes and clamp the parsed URL value to it, so `?pageSize=5000` falls back to the default instead of rendering a giant skeleton. The quantity selector now maps over `PAGE_SIZES` as well.

docs/content/patterns/user-input/bulk-actions/bulk-actions-edit-drawer.demo.tsx
sebald: Aligned — the demos now use `variant="muted"` (with `fontSize="sm"`) to match the working example, including `bulk-actions-partial-failure.demo.tsx`.

docs/content/patterns/user-input/bulk-actions/index.mdx
sebald: Applied your reword.

docs/content/patterns/user-input/bulk-actions/index.mdx
sebald: Good catch — it was a choice (bake-in time), but since the example lands in this PR I've promoted it to `beta` here.

jim761 (approved):

docs/app/(examples)/examples/bulk-actions/page.tsx
```diff
@@ -0,0 +1,94 @@
...
+      {/* One bulk-actions instance for the whole page (see useBulkActions):
+          the bar and the edit drawer share its `busy` state. It sits inside
+          the session and selection providers because it reads both. */}
+      <BulkActionsProvider>
+        <Page>
+          <Page.Header>
+            <Title>Events</Title>
+            <Description>
+              Select events in the table and run one operation across all of
+              them — publish, edit, export, or delete.
```
sebald:
```suggestion
              them.
```

aromko (approved):

PR #5643, opened 2026-07-15
feat(DST-990): add Menu selection visuals, shortcuts, and dividers

sarahgm:
# Description

Revisits `<Menu>` (DST-990) with richer building blocks for advanced menus, plus a new standalone `<Keyboard>` primitive.

- **Selection visuals.** In `selectionMode="single"`/`"multiple"`, items show a leading checkmark and a highlighted row, aligned like `<ListBox>`. Command menus (no `selectionMode`) render exactly as before.
- **Keyboard-shortcut hints** via a new shared `<Keyboard>` component (a sibling to `<TextValue>` and `<Description>`). It renders a `<kbd>` key-cap on its own, and inside a `Menu.Item` it becomes a muted, right-aligned hint wired to react-aria's `aria-describedby`.
- **Dividers.** Drop the shared `<Divider>` between `<Menu.Item>`s to separate groups with a `role="separator"` line, rather than a `Menu.Divider` alias.
- **Docs.** Use-case-driven Menu documentation, SVG anatomy, and a new Keyboard page.

Closes DST-990

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm sb` and open `Components/Menu` → the `Advanced`, `MultiSelection`, and `AccessSections` stories. Confirm icons, right-aligned `⌘` shortcut hints, a `role="separator"` divider, and (in multi-select) a leading checkmark plus highlighted row on selected items.
2. Open `Components/Keyboard` → `Basic`. Confirm it renders a standalone `<kbd>` key-cap.
3. Run `pnpm start` and visit `/components/overlay/menu` and `/components/content/keyboard` to review the new docs, demos, and anatomy diagrams.
4. Run `pnpm typecheck:only && pnpm test:unit` (Keyboard and Menu unit tests pass locally).

## Breaking Changes

Yes (`@marigold/system`, major)

Custom themes that implement the `Menu` theme record must add a `keyboard` key, since the `@marigold/system` `Theme` type now requires it. This is a compile-time break for custom themes, so `@marigold/system` ships as a major bump. Usage of `@marigold/components` is backward compatible.

## Checklist


sarahgm:
### Why the shared `<Divider>` instead of a `Menu.Divider` compound component

DST-990's scope suggested adding `Menu.Divider`. An earlier commit did exactly that, then this PR removed it in favor of dropping the shared `<Divider>` straight into a menu. The reasoning:

1. **`Menu.Divider` would only be an alias.** It carried no menu-specific behavior. It just re-exported the shared `<Divider>`, so it added a second name for one component without adding capability.

2. **The separator role and spacing are already handled without it.** The shared `<Divider>` renders `role="separator"` on its own, and the menu-specific spacing lives in the theme (`[&_[role=separator]]:my-1` in `Menu.styles.ts`). The Menu adapts the divider through the theme layer rather than through a wrapper component, which is where the rest of the Menu's slot styling already lives (label, description, keyboard).

3. **One separator, one API.** Using the same `<Divider>` everywhere means one thing to learn, document, and keep in sync. A `Menu.Divider` alias would need its own doc entry and export, and would invite the "which divider do I use?" question for no gain.

4. **Consistent with how Marigold composes collections.** Menus already compose shared primitives (icons, `<Keyboard>`, text slots). Dropping in the shared `<Divider>` keeps that mental model uniform instead of introducing a menu-only variant of a primitive that already exists.

5. **No door closed.** If a genuinely menu-specific divider behavior ever appears, we can introduce `Menu.Divider` then without breaking the shared usage. Until there is such a behavior, the alias is maintenance weight.

The docs and changeset reflect this: the divider is documented as "drop the shared `<Divider>` between `<Menu.Item>`s," and the `Advanced` story asserts the resulting `role="separator"`.

cc: worth updating the DST-990 *Included* scope so the ticket matches this decision.

packages/system/src/types/theme.ts
```diff
@@ -158,7 +159,13 @@ export type Theme = {
...
     >;
     Menu?: Record<
-      'container' | 'section' | 'item' | 'button' | 'label' | 'description',
+      | 'container'
+      | 'section'
+      | 'item'
+      | 'button'
+      | 'label'
+      | 'description'
+      | 'keyboard',
```
aromko, line 168: ⚠️ **Semver: required `keyboard` key is a compile-time break for custom themes**

Adding `'keyboard'` to the `Menu` record means any custom theme implementing `Menu` stops compiling until it adds the new key, but the PR declares "Breaking Changes: No" and the changeset ships `@marigold/system` as `minor`. The migration note in the changeset is great; just confirm the `minor` bump (vs `major`) is the intended convention for required theme-slot additions before merging.

packages/components/src/Keyboard/Keyboard.test.tsx
```diff
@@ -0,0 +1,23 @@
+import { render, screen } from '@testing-library/react';
+import { KeyboardContext } from 'react-aria-components/Keyboard';
+import { Provider } from 'react-aria-components/slots';
+import { Basic } from './Keyboard.stories';
+
+test('renders as a kbd element', () => {
+  render(<Basic.Component data-testid="keyboard">⌘K</Basic.Component>);
```
aromko: ⚠️ **Avoid `getByTestId`**: the project guidelines prefer accessible queries. `screen.getByText('⌘K')` returns the `<kbd>` element itself, so the testid isn't needed:

```tsx
test('renders as a kbd element', () => {
  render(<Basic.Component>⌘K</Basic.Component>);

  const el = screen.getByText('⌘K');

  expect(el.tagName).toBe('KBD');
});
```

Same applies to the `KeyboardContext` test below.

themes/theme-rui/src/components/Menu.styles.ts
```diff
@@ -13,19 +13,20 @@ export const Menu: ThemeComponent<'Menu'> = {
       // The surrounding Popover (or Tray) paints the overlay surface; the menu
       // renders flat inside it.
       'w-full',
-      'text-foreground overflow-x-hidden p-1 outline-none overflow-y-auto',
+      'text-foreground overflow-x-hidden p-1 outline-none overflow-y-auto space-y-px',
+      // Breathing room around dividers (Menu.Divider is the shared <Divider>).
```
aromko: 💡 The comment implies a `Menu.Divider` alias exists, but this PR deliberately went with the shared `<Divider>` instead (as stated in the PR description).

```suggestion
      // Breathing room around dividers (the shared <Divider> rendered between items).
```

packages/components/src/Menu/Menu.stories.tsx
```diff
@@ -524,3 +543,51 @@ Mobile.test('Mobile Menu close with Escape', async ({ canvas, step }) => {
     });
   });
 });
+
+export const Advanced = meta.story({
```
aromko, line 591: 💡 **Chromatic snapshots**: the base `Advanced` story snapshots only the closed trigger (a near-duplicate of `Basic`), while `Advanced.test` captures the open menu, which is the visual actually worth guarding. If you want only the open-menu snapshot, disable it on the base story and re-enable it on the test, since `.test()` stories inherit the base's `disableSnapshot`:

```tsx
export const Advanced = meta.story({
  tags: ['component-test'],
  // ...
});

Advanced.test(
  'Renders icons and a divider, and wires the shortcut to the item description',
  async ({ canvas }) => {
    // ...
  }
);
```

packages/components/src/Menu/Menu.stories.tsx
```diff
@@ -353,6 +356,22 @@ export const MultiSelection = meta.story({
   },
 });
 
+MultiSelection.test(
```
aromko, line 361: 💡 **AAA structure**: this test asserts `aria-checked="false"` between two user actions. Per the strict Arrange-Act-Assert convention (one act phase per test, no expects between actions), consider splitting it: one test for the initial unchecked `menuitemcheckbox` state, one for the toggle. Both stay cheap since they share the story.

packages/system/src/types/theme.ts
sarahgm: We went with the honest option: bumped `@marigold/system` to **major** and flipped the PR's Breaking Changes section to Yes. The changeset now calls out the required `keyboard` key as a compile-time break for custom themes. Leaving this open for you to confirm the major bump is the convention you'd like (vs. a minor + migration note). Note it cascades the fixed changeset group to major for the next release.

packages/components/src/Keyboard/Keyboard.test.tsx
sarahgm: Done in 762db34f9 — both tests now use `screen.getByText('⌘K')` and assert the `<kbd>` tag (and the context class on that element); dropped the test id.

themes/theme-rui/src/components/Menu.styles.ts
sarahgm: Done in 762db34f9 — comment now reads "the shared <Divider> rendered between items."

packages/components/src/Menu/Menu.stories.tsx
sarahgm: Done in 762db34f9 — `Advanced` now sets `chromatic.disableSnapshot: true` and `Advanced.test` re-enables it, so only the open menu is snapshotted.

packages/components/src/Menu/Menu.stories.tsx
sarahgm: Done in 762db34f9 — split into 'Exposes the checkbox role in an unchecked state' and 'Toggles selection on click', each with a single act phase.

aromko (approved):
Idk I just let the thread open for discussion like you proposed.

PR #5644, opened 2026-07-15
fix(DST-1622): respect consumer aria-label on ProgressCircle and Loader

sebald:
# Description

`<ProgressCircle>` set its localized "loading" `aria-label` **after** spreading `{...props}`, so any consumer-provided `aria-label` / `aria-labelledby` was silently overwritten. Every spinner announced the same generic string regardless of context.

This makes the localized string a **fallback**: it is applied only when the caller provides no accessible name. Placing it before `{...props}` lets the consumer's label win.

- **`ProgressCircle`** — respects consumer `aria-label` / `aria-labelledby`; the localized message fills in only when neither is present.
- **`Loader` (`BaseLoader`)** — same fallback pattern, now also keyed off `aria-labelledby`, and it no longer emits a redundant `aria-label="Loading..."` next to a consumer's `aria-labelledby`.

**Current vs. new behavior:** before, `<ProgressCircle aria-label="Sending reminders" />` still announced "Loading…". Now it announces "Sending reminders". Spinners that pass no label (inside `Button`, `ComboBox`, `SearchInput`) are unchanged — they still get the localized fallback.

Closes DST-1622

## Screenshots / Preview

No visual change — accessible-name (a11y) fix only.

## Test Instructions

1. Render `<ProgressCircle aria-label="Sending reminders" />` and inspect the `progressbar` node — its `aria-label` is now `"Sending reminders"` (previously `"Loading..."`).
2. Render `<ProgressCircle />` with no label — it still falls back to the localized `"Loading..."`.
3. Render `<ProgressCircle aria-labelledby="…" />` — the `progressbar` carries `aria-labelledby` and no redundant `aria-label`.
4. `pnpm test:unit packages/components/src/ProgressCircle packages/components/src/Loader` — all pass (new assertions cover the above for both `ProgressCircle` and `Loader`).

## Breaking Changes

No

## Checklist

### DIFF
--- a/packages/components/src/Loader/BaseLoader.tsx
+++ b/packages/components/src/Loader/BaseLoader.tsx
@@ -195,10 +195,7 @@ export const BaseLoader = ({
   const stringFormatter = useLocalizedStringFormatter(intlMessages, 'marigold');
   const className = useClassNames({ component: 'Loader', variant, size });
 
-  // The localized "loading" message is only a fallback: apply it solely when the
-  // consumer supplies no accessible name of any kind (`aria-label`,
-  // `aria-labelledby`, or a visible `children` label). This keeps the consumer's
-  // label winning and avoids emitting a redundant `aria-label` next to it.
+  // Localized label is only a fallback; let a consumer's label win.
   const hasLabel = ariaLabel || ariaLabelledby || children;
 
   return (
--- a/packages/components/src/Loader/BaseLoader.tsx
+++ b/packages/components/src/Loader/BaseLoader.tsx
@@ -188,21 +188,27 @@ export const BaseLoader = ({
   size,
   children,
   'aria-label': ariaLabel,
+  'aria-labelledby': ariaLabelledby,
   loaderType = 'circle',
   ...props
 }: BaseLoaderProps) => {
   const stringFormatter = useLocalizedStringFormatter(intlMessages, 'marigold');
   const className = useClassNames({ component: 'Loader', variant, size });
 
+  // The localized "loading" message is only a fallback: apply it solely when the
+  // consumer supplies no accessible name of any kind (`aria-label`,
+  // `aria-labelledby`, or a visible `children` label). This keeps the consumer's
+  // label winning and avoids emitting a redundant `aria-label` next to it.
+  const hasLabel = ariaLabel || ariaLabelledby || children;
+
   return (
     <ProgressBar
       className={className.container}
       isIndeterminate
       aria-label={
-        ariaLabel || children
-          ? ariaLabel
-          : stringFormatter.format('loadingMessage')
+        hasLabel ? ariaLabel : stringFormatter.format('loadingMessage')
       }
+      aria-labelledby={ariaLabelledby}
       {...props}
     >
       {loaderType === 'xloader' ? (
--- a/packages/components/src/ProgressCircle/ProgressCircle.test.tsx
+++ b/packages/components/src/ProgressCircle/ProgressCircle.test.tsx
@@ -92,6 +92,17 @@ test('respects a consumer-provided aria-label', () => {
   expect(progressCircle).toHaveAttribute('aria-label', 'Sending reminders');
 });
 
+test('falls back to the localized label when aria-label is explicitly undefined', () => {
+  // A conditional label (`aria-label={cond ? '…' : undefined}`) puts the key
+  // `aria-label: undefined` into props. The fallback must still apply so the
+  // progressbar always has an accessible name.
+  render(<Basic.Component aria-label={undefined} />);
+
+  const progressCircle = screen.getByRole('progressbar');
+
+  expect(progressCircle).toHaveAttribute('aria-label', 'Loading...');
+});
+
 test('respects a consumer-provided aria-labelledby without a redundant aria-label', () => {
   render(
     <>
--- a/packages/components/src/ProgressCircle/ProgressCircle.test.tsx
+++ b/packages/components/src/ProgressCircle/ProgressCircle.test.tsx
@@ -83,3 +83,25 @@ test('has aria label', () => {
 
   expect(progressCircle).toHaveAttribute('aria-label', 'Loading...');
 });
+
+test('respects a consumer-provided aria-label', () => {
+  render(<Basic.Component aria-label="Sending reminders" />);
+
+  const progressCircle = screen.getByRole('progressbar');
+
+  expect(progressCircle).toHaveAttribute('aria-label', 'Sending reminders');
+});
+
+test('respects a consumer-provided aria-labelledby without a redundant aria-label', () => {
+  render(
+    <>
+      <span id="progress-label">Sending reminders</span>
+      <Basic.Component aria-labelledby="progress-label" />
+    </>
+  );
+
+  const progressCircle = screen.getByRole('progressbar');
+
+  expect(progressCircle).toHaveAttribute('aria-labelledby', 'progress-label');
+  expect(progressCircle).not.toHaveAttribute('aria-label');
+});
--- a/packages/components/src/ProgressCircle/ProgressCircle.tsx
+++ b/packages/components/src/ProgressCircle/ProgressCircle.tsx
@@ -90,10 +90,16 @@ export const ProgressCircle = ({
   ...props
 }: ProgressCircleProps) => {
   const stringFormatter = useLocalizedStringFormatter(intlMessages);
+  // The localized "loading" message is only a fallback: it fills the accessible
+  // name when the consumer provides none. Placing it before `{...props}` lets a
+  // caller's `aria-label`/`aria-labelledby` win instead of being overwritten.
+  const hasLabel = props['aria-label'] || props['aria-labelledby'];
   return (
     <ProgressBar
+      aria-label={
+        hasLabel ? undefined : stringFormatter.format('loadingMessage')
+      }
       {...props}
-      aria-label={stringFormatter.format('loadingMessage')}
       isIndeterminate
     >
       <ProgressCircleSvg size={size} />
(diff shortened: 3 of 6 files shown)


sarahgm (commented):
## Automated Code Review

### Overview
| Field | Value |
|-------|-------|
| Author | @sebald |
| Branch | `dst-1622-progresscircle-aria-label` → `beta-release` |
| Files Changed | 5 |
| Additions | +66 |
| Deletions | -4 |
| Visual Regression | Not needed (a11y-attribute-only fix; PR targets `beta-release`) |
| CI | ✅ All green — Unit, Typecheck, Lint, Storybook, Coverage, Size Limit |

### Linked Ticket
**DST-1622**: 🐛 ProgressCircle ignores a consumer-provided aria-label — Status: Bereit (To Do), Unplanned.
The PR implements the ticket's suggested solution verbatim for `ProgressCircle`, and extends the same fallback pattern to `Loader` as the ticket asked. Acceptance expectations (custom `aria-label` respected + unit test) are met.

### Checklist
- **TypeScript** — ✅ No `any`, `import type` used, prop keys typed via `Pick<ProgressBarProps, ...>`.
- **Component patterns** — ✅ Correct ARIA precedence (`aria-labelledby` > `aria-label` > localized fallback); avoids redundant `aria-label` next to `aria-labelledby`. Terse one-line comments.
- **React** — ✅ Functional, no new hooks, `hasLabel` is a trivial derived const (no `useMemo` warranted). No performance concerns.
- **Testing** — ✅ Vitest, `screen.getByRole('progressbar')` (accessible query), imports the `Basic` story (no fixtures), blank-line AAA separation, no `fireEvent`.
- **Accessibility** — ✅ This is the fix and it's correct; naming precedence and no-redundant-name behavior both asserted by tests.
- **Changeset** — ✅ Present, `patch` (correct for a bug fix), body starts with `fix(DST-1622):`, documents what/why/impact.

### Issues Found

#### Critical (Must Fix)
None.

#### Warnings (Should Fix)
None.

**1. `packages/components/src/ProgressCircle/ProgressCircle.tsx` — inconsistent implementation with `BaseLoader`, with a minor edge case.**
`BaseLoader` destructures `aria-label` / `aria-labelledby` *out* of props and sets them explicitly, so the spread can never re-apply them. `ProgressCircle` instead leaves them in `{...props}` and relies on spread ordering:

```tsx
aria-label={hasLabel ? undefined : stringFormatter.format('loadingMessage')}
{...props}   // re-applies whatever aria-label prop was passed, even undefined/""
```

Consequence: `<ProgressCircle aria-label={undefined} />` or `aria-label=""` computes `hasLabel = false`, applies the localized fallback, then the spread overwrites it back to `undefined`/`""` — the spinner ends up with no accessible name. This matches the ticket's suggested snippet, so it's deliberate and the edge case is unlikely, but aligning `ProgressCircle` with `BaseLoader`'s destructure approach would remove it and make the two components consistent. Low priority.

**2. Stories not added** (checklist item unchecked). The ticket suggested "add a story/unit test." Only unit tests were added. Since this is a non-visual a11y-attribute fix, unit tests are arguably sufficient and a story adds little — noting only for completeness, not a blocker.

### Recommendation
**Approve.** Correct, focused, well-tested a11y fix that matches the ticket's prescribed solution and passes CI. The one suggestion (align `ProgressCircle` with `BaseLoader`'s destructure pattern) is a minor consistency/edge-case cleanup, not a blocker.

---

packages/components/src/ProgressCircle/ProgressCircle.tsx
```diff
@@ -90,10 +90,14 @@ export const ProgressCircle = ({
...
 }: ProgressCircleProps) => {
   const stringFormatter = useLocalizedStringFormatter(intlMessages);
+  // Localized label is only a fallback; let a consumer's label win.
+  const hasLabel = props['aria-label'] || props['aria-labelledby'];
   return (
     <ProgressBar
+      aria-label={
+        hasLabel ? undefined : stringFormatter.format('loadingMessage')
+      }
       {...props}
```
aromko, line 103: ⚠️ **Explicit `aria-label={undefined}` clobbers the fallback**

Because `hasLabel` is read from `props` while `aria-label` stays inside the spread, a conditional label like `<ProgressCircle aria-label={isBulk ? 'Sending reminders' : undefined} />` puts the key `aria-label: undefined` into `{...props}`, which overrides the fallback set before the spread — the progressbar ends up with no accessible name at all (RAC dev-warning, axe violation). `BaseLoader` is immune because it destructures the label props out of the spread. Destructuring here the same way keeps all three tested behaviors and additionally restores the fallback for that case:

```suggestion
  'aria-label': ariaLabel,
  'aria-labelledby': ariaLabelledby,
  ...props
}: ProgressCircleProps) => {
  const stringFormatter = useLocalizedStringFormatter(intlMessages);
  // Localized label is only a fallback; let a consumer's label win.
  const hasLabel = ariaLabel || ariaLabelledby;
  return (
    <ProgressBar
      aria-label={
        hasLabel ? ariaLabel : stringFormatter.format('loadingMessage')
      }
      aria-labelledby={ariaLabelledby}
      {...props}
```

packages/components/src/ProgressCircle/ProgressCircle.test.tsx
```diff
@@ -83,3 +83,25 @@ test('has aria label', () => {
...
+      <span id="progress-label">Sending reminders</span>
+      <Basic.Component aria-labelledby="progress-label" />
+    </>
+  );
+
+  const progressCircle = screen.getByRole('progressbar');
+
+  expect(progressCircle).toHaveAttribute('aria-labelledby', 'progress-label');
+  expect(progressCircle).not.toHaveAttribute('aria-label');
+});
```
aromko, line 118: 💡 **Test idea:** add the conditional-label case — `render(<Basic.Component aria-label={undefined} />)` should still fall back to `aria-label="Loading..."`. Cheap regression guard for the fallback-vs-spread ordering (see comment in `ProgressCircle.tsx`).

packages/components/src/Loader/BaseLoader.tsx
```diff
@@ -188,21 +188,24 @@ export const BaseLoader = ({
...
     <ProgressBar
       className={className.container}
       isIndeterminate
       aria-label={
-        ariaLabel || children
-          ? ariaLabel
-          : stringFormatter.format('loadingMessage')
+        hasLabel ? ariaLabel : stringFormatter.format('loadingMessage')
       }
+      aria-labelledby={ariaLabelledby}
```
aromko, line 208: 💡 **Pre-existing edge case, surfaced by this change (non-blocking):** in `fullscreen` mode the `Dialog` labels itself via `aria-labelledby={id}` pointing at this ProgressBar node. Now that a consumer `aria-labelledby` suppresses the ProgressBar's own `aria-label`, a fullscreen loader with `aria-labelledby` and no `children` leaves the Dialog with an essentially empty accessible name (the SVG is `aria-hidden`). Not for this PR — just flagging for a possible follow-up.

aromko (commented):
## Automated Code Review

### Overview

| Field | Value |
|-------|-------|
| Branch | `dst-1622-progresscircle-aria-label` → `beta-release` |
| Files Changed | 5 (+66 / −4) |
| CI | All 18 checks passing |

### Linked Ticket

**[DST-1622](https://reservix.atlassian.net/browse/DST-1622)**: 🐛 ProgressCircle ignores a consumer-provided aria-label (In Review)

The PR implements exactly the ticket's suggested solution, including the `Loader`/`BaseLoader` follow-up, and adds unit tests for both components. ✅

### Review Summary

The fix is correct and minimal: placing the localized `aria-label` before `{...props}` makes it a true fallback, both components now share the same semantics, and the internal spinner usages (`Button`, `ComboBox`, `SearchInput` — all render a bare `<ProgressCircle />`) keep the localized fallback. New tests use accessible queries, import stories, and cover all three behaviors from the PR description. Changeset is a well-scoped patch bump. No React performance concerns (the localized `format()` call is now only evaluated in the fallback branch).

### Issues Found

**Warnings (Should Fix)**

1. `ProgressCircle.tsx` — an explicitly passed `aria-label={undefined}` (e.g. from a conditional) clobbers the fallback, leaving the progressbar unnamed. Destructure the label props like `BaseLoader` does — see inline comment with suggestion.

**Suggestions (Nice to Have)**

2. `ProgressCircle.test.tsx` — add a regression test for the conditional-label case (see inline comment).
3. `BaseLoader.tsx` — pre-existing fullscreen `Dialog` naming edge case surfaced by this change; follow-up material, not for this PR (see inline comment).

### Recommendation

**Comment** (approve-leaning) — the single warning is an edge-case hardening one step beyond the ticket's own suggested snippet; worth fixing while in here, but not a blocker.

---

packages/components/src/ProgressCircle/ProgressCircle.tsx
sebald: Good catch — fixed in 8f47c3f9d. `ProgressCircle` now destructures `aria-label` / `aria-labelledby` out of the spread (same as `BaseLoader`), so an explicit `aria-label={undefined}` no longer re-applies over the fallback. The localized message is a true fallback again.

packages/components/src/ProgressCircle/ProgressCircle.test.tsx
sebald: Added in 8f47c3f9d: `falls back to the localized label when aria-label is explicitly undefined` asserts the progressbar still gets `aria-label="Loading..."` for the conditional-label case.

packages/components/src/Loader/BaseLoader.tsx
sebald: Went ahead and fixed this here too (8f47c3f9d) since we were in the area. `LoaderFullScreen` now points the Dialog's `aria-labelledby` at the consumer's element directly (`aria-labelledby={ariaLabelledby ?? id}`) instead of the intermediate loader node — the accname spec won't follow a second `labelledby` hop, which is what left the modal unnamed. Added two tests covering the fallback-name and consumer-`aria-labelledby` cases via `toHaveAccessibleName`.

aromko (approved):

PR #5647, opened 2026-07-16
fix(DST-1629): apply SelectList action slot placement to trailing actions

sarahgm:
# Description

A trailing action inside a `SelectList` item (icon `Button` or `ActionMenu`) was not receiving the `action` slot's grid-placement classes, so it auto-placed into the title row and stretched it, opening a gap above the description.

Root cause: the action slot className was provided only on **RAC's** `ButtonContext`, but a Marigold `Button`/`LinkButton`/`ActionMenu` reads the **Marigold** `ButtonContext`. The fix routes the action className through the Marigold `ButtonContext` (the same cascade `Panel.Header` and `Page.Header` use), so the action now spans both rows and stays centered. The now-dead RAC `ButtonContext` branch was removed.

The ButtonGroup half was an example, not a component bug: the `WithActionMenu` story set `size="large"` on the ActionMenu, so its trigger rendered larger than the small sibling buttons. Dropped the override so the trigger inherits the group's size, and split the test into inherit-vs-local-wins coverage (with a snapshot-disabled fixture story for the override case).

Verified in a real browser: title→description gap is 0 and the action spans rows 1-2 for both the icon-action and ActionMenu stories.

Closes DST-1629

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm sb` and open **Components/SelectList → With Icon Action** and **With Action Menu**: the title and description sit tightly together and the action stays vertically centered. Action-less items are unchanged.
2. Open **Components/ButtonGroup → With Action Menu**: the "More actions" trigger matches the small Edit / Duplicate buttons.

## Breaking Changes

No

## Checklist

### DIFF
--- a/packages/components/src/ButtonGroup/ButtonGroup.stories.tsx
+++ b/packages/components/src/ButtonGroup/ButtonGroup.stories.tsx
@@ -112,6 +112,20 @@ export const WithActionMenu = meta.story({
     <ButtonGroup aria-label="With ActionMenu" size="small" variant="ghost">
       <Button aria-label="Edit">Edit</Button>
       <Button aria-label="Duplicate">Duplicate</Button>
+      <ActionMenu aria-label="More actions">
+        <ActionMenu.Item id="archive">Archive</ActionMenu.Item>
+        <ActionMenu.Item id="delete">Delete</ActionMenu.Item>
+      </ActionMenu>
+    </ButtonGroup>
+  ),
+});
+
+// Not snapshotted: deliberately oversizes the trigger to exercise local-wins precedence (the default inheriting look lives in `WithActionMenu`).
+export const ActionMenuLocalSize = meta.story({
+  parameters: { chromatic: { disableSnapshot: true } },
+  render: () => (
+    <ButtonGroup aria-label="Item actions" size="small" variant="ghost">
+      <Button aria-label="Edit">Edit</Button>
       <ActionMenu aria-label="More actions" size="large">
         <ActionMenu.Item id="archive">Archive</ActionMenu.Item>
         <ActionMenu.Item id="delete">Delete</ActionMenu.Item>
--- a/packages/components/src/ButtonGroup/ButtonGroup.stories.tsx
+++ b/packages/components/src/ButtonGroup/ButtonGroup.stories.tsx
@@ -120,7 +120,9 @@ export const WithActionMenu = meta.story({
   ),
 });
 
-// Not snapshotted: deliberately oversizes the trigger to exercise local-wins precedence (the default inheriting look lives in `WithActionMenu`).
+// Fixture for the "local size wins" unit test, so not snapshotted. The ActionMenu
+// sets an explicit `size="large"` that must beat the group's `small`, which
+// `WithActionMenu` (its trigger inherits the group size) can't assert.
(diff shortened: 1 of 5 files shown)


packages/components/src/SelectList/SelectList.test.tsx
```diff
@@ -327,6 +327,17 @@ describe('SelectList', () => {
...
+
+    test('applies the action slot placement to a trailing action button', () => {
+      render(<WithIconAction.Component />);
+
+      const action = screen.getByRole('button', {
+        name: /learn more about credit card/i,
+      });
+
+      // Action spans both rows so a tall control doesn't stretch the title row and gap the description.
+      expect(action).toHaveClass('col-start-3', 'row-span-2');
```
aromko, line 340: 💡 **Suggestion (non-blocking):** This locks the placement for the icon `Button` path (`WithIconAction`), but the ticket's Chromatic evidence covered the ActionMenu path as well. The ActionMenu trigger consumes the option's `ButtonContext` through a different component chain (`ActionMenu` → `Button`), so a matching assertion on the `WithActionMenu` trigger would lock both consumers cheaply:

```tsx
test('applies the action slot placement to a trailing ActionMenu trigger', () => {
  render(<WithActionMenu.Component />);

  const trigger = screen.getAllByRole('button', { name: /manage/i })[0];

  expect(trigger).toHaveClass('col-start-3', 'row-span-2');
});
```

packages/components/src/ButtonGroup/ButtonGroup.stories.tsx
```diff
@@ -112,6 +112,20 @@ export const WithActionMenu = meta.story({
...
+      <ActionMenu aria-label="More actions">
+        <ActionMenu.Item id="archive">Archive</ActionMenu.Item>
+        <ActionMenu.Item id="delete">Delete</ActionMenu.Item>
+      </ActionMenu>
+    </ButtonGroup>
+  ),
+});
+
+// Not snapshotted: deliberately oversizes the trigger to exercise local-wins precedence (the default inheriting look lives in `WithActionMenu`).
+export const ActionMenuLocalSize = meta.story({
```
aromko, line 126: Don't get the point of this story? What is this good for?

packages/components/src/SelectList/SelectList.test.tsx
sarahgm: Good call, added it in 297d95ff2. The assertion renders `WithActionMenu` and checks the first `Manage …` trigger keeps `col-start-3 row-span-2`, so both consumers (plain `Button` via `WithIconAction` and the `ActionMenu → Button` chain) are now locked.

packages/components/src/ButtonGroup/ButtonGroup.stories.tsx
sarahgm: Fair, the intent wasn't clear. It's the fixture for the `ButtonGroup` "local size wins" unit test, so it's not snapshotted. The `ActionMenu` sets an explicit `size="large"` that must beat the group's `small`, which `WithActionMenu` can't assert because its trigger deliberately inherits the group size. I rewrote the comment to say exactly that in 297d95ff2.

aromko (approved):

PR #5649, opened 2026-07-16
feat(DST-1635): add bleed prop to Drawer.Content for full-width children

sebald:
# Description

`Drawer.Content` applied a fixed horizontal padding (the shared `ui-panel-content` utility → `px-6`, 24px) with no way to opt out. A full-width child like `Accordion` or `Table` was trapped inside that 24px gutter: its item dividers, hover/selection backgrounds and trailing icons could not reach the drawer edges.

This adds a `bleed` prop to `Drawer.Content`, mirroring `Panel.Content`'s existing bleed contract. When set, it drops the horizontal padding and publishes a `--bleed-px` custom property (matching the drawer's `px-6`), so edge-aware children inset their own text to stay aligned with the drawer title while their dividers/backgrounds span the full width. The default (unbled) path keeps `ui-panel-content` and does not set `--bleed-px`, so existing drawers are unchanged.

Closes DST-1635

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. In Storybook, open `Components/Drawer/BleedAccordion` and open the drawer: the Accordion item dividers reach the drawer edges while the header text stays aligned with the drawer title.
2. Run the tests: `pnpm test:unit --run packages/components/src/Drawer/Drawer.test.tsx` and `pnpm test:sb --run packages/components/src/Drawer/Drawer.stories.tsx`.
3. Docs: `pnpm start`, then Components → Drawer → "Full-width content".

## Breaking Changes

No

## Checklist


docs/content/components/overlay/drawer/index.mdx
```diff
@@ -156,6 +156,18 @@ The [Bulk Actions pattern](/patterns/user-input/bulk-actions#editing-multiple-fi
...
 
+### Full-width content
+
+By default, drawer content sits inside the drawer's horizontal padding, so its left and right edges line up with the title and actions. That shared edge is what keeps the drawer reading as one column.
+
+Some content reads best edge to edge: a grouped [Accordion](/components/navigation/accordion) of filter facets or settings sections, or a [Table](/components/collection/table) of records. Opt in with the `bleed` prop on `Drawer.Content`. It drops the horizontal padding so backgrounds and dividers reach the drawer edges, while edge-aware children inset their own text to stay aligned with the drawer title.
+
+<ComponentDemo name="drawer-bleed-accordion" />
+
+Drop an `<Accordion>` into a bled content area and its item dividers span the full width while the header and content stay aligned with the title. No extra prop is needed on the Accordion: it picks up the drawer's horizontal padding automatically, the same way `Table` cells do. This mirrors [`Panel.Content bleed`](/components/layout/panel#full-width-content).
```
aromko, line 167: ⚠️ **`Table` cells do not pick up the drawer's padding.** The only consumer of `--bleed-px` on `beta-release` is `Accordion.styles.ts`. `Table` sources its edge padding from `[--cell-edge-padding:var(--panel-px,var(--cell-x-padding))]` (`Table.styles.ts:9`), so inside a bled drawer (no `--panel-px`) its cell text falls back to the small `--cell-x-padding` and will *not* align with the drawer title, which is the exact misalignment this feature is meant to fix. Either extend Table's fallback chain to `var(--panel-px, var(--bleed-px, var(--cell-x-padding)))` in `Table.styles.ts`, or scope this sentence (and the Table mentions in the changeset and TSDoc) down to Accordion.

.changeset/drawer-content-bleed.md
```diff
@@ -0,0 +1,18 @@
+---
+'@marigold/components': minor
+---
+
+feat(DST-1635): add a `bleed` prop to `Drawer.Content` so edge-aware children can span the full Drawer width.
+
+**What changed:**
+
+- `<Drawer.Content bleed>` drops the Drawer's horizontal content padding and publishes a `--bleed-px` custom property (matching the Drawer's `px-6`), mirroring `Panel.Content`'s `bleed`.
+- Edge-aware children that read `--bleed-px` (e.g. `Accordion` via `--accordion-x-padding`, `Table`) inset their own content to stay aligned with the Drawer title while their dividers/backgrounds reach the Drawer edges.
```
aromko: ⚠️ Same issue as in `index.mdx`: `Table` does not read `--bleed-px` (it reads `--panel-px`), so listing it here as an edge-aware consumer is inaccurate unless `Table.styles.ts` is extended.

packages/components/src/Drawer/DrawerContent.tsx
```diff
@@ -11,6 +11,15 @@ export interface DrawerContentProps {
    * Children of the component.
    */
   children?: ReactNode;
+  /**
+   * Render the content edge-to-edge horizontally, skipping the Drawer's
+   * horizontal padding. Publishes `--bleed-px` so edge-aware children
+   * (Accordion, Table) inset their own content to stay aligned with the
```
aromko, line 17: ⚠️ Same issue as in `index.mdx`: of the two named children, only `Accordion` actually consumes `--bleed-px` today. Suggest dropping `Table` from this TSDoc (or wiring it up in `Table.styles.ts`).

packages/components/src/Drawer/Drawer.stories.tsx
```diff
@@ -534,3 +535,57 @@ TitleOnlyWithoutHeader.test(
...
+
+/**
+ * `<Drawer.Content bleed>` drops the horizontal padding and publishes
+ * `--bleed-px`, letting edge-aware children like `Accordion` span the full
+ * width: item dividers reach the Drawer edges while the header text stays
+ * aligned with the Drawer title.
+ */
+export const BleedAccordion = meta.story({
+  tags: ['component-test'],
+  parameters: { chromatic: { disableSnapshot: false } },
```
aromko: 💡 This snapshots the *closed* drawer (just the trigger button), where the bleed isn't visible. The file convention is base `disableSnapshot: true` with the `.test()` re-enabling via `false`, so Chromatic captures the opened drawer after `play` (see `Basic.test`).

```suggestion
  parameters: { chromatic: { disableSnapshot: true } },
```

…and re-enable on the test:

```ts
BleedAccordion.test(
  'Opens the drawer with a full-width Accordion',
  async ({ canvas, userEvent }) => { ... }
);
```

packages/components/src/Drawer/DrawerContent.tsx
```diff
@@ -29,7 +39,18 @@ export const DrawerContent = ({
...
+      className={cn(
+        '[grid-area:content]',
+        // The unbled path uses the shared `ui-panel-content` utility
+        // (`overflow-y-auto px-6 py-4 outline-none`). When bled we keep the
+        // scroll/vertical rhythm but drop the horizontal padding and publish
+        // `--bleed-px` matching that `px-6`, so edge-aware children can inset
+        // themselves to align with the Drawer title (also `px-6`, via
+        // `ui-panel-header`) while their dividers/backgrounds reach the edges.
+        bleed
+          ? 'overflow-y-auto py-4 outline-none [--bleed-px:calc(var(--spacing)*6)]'
```
aromko: 💡 Two small notes on this line, both fine within the ticket's scope:
1. The hardcoded `calc(var(--spacing)*6)` duplicates the theme's `px-6` from `ui-panel-content`; if the theme padding ever changes, the bled inset silently drifts. A follow-up where the Drawer publishes a `--drawer-px` variable (like Panel's `--panel-px`) would give component and theme one source of truth.
2. Since the bled branch replaces `classNames.content` entirely, `variant`/`size` on `Drawer.Content` are silently ignored when `bleed` is set. Harmless today (theme-rui's Drawer content has no variants), but worth a sentence in the TSDoc.

aromko (commented):
## Automated Code Review

### Overview

| Field | Value |
|-------|-------|
| Branch | `dst-1635-drawer-bleed` → `beta-release` |
| Files Changed | 6 (+219 / -2) |
| CI | All 18 checks passing |
| Linked Ticket | [DST-1635](https://reservix.atlassian.net/browse/DST-1635) (In Review) |

### Summary

Clean, well-scoped implementation that matches both the ticket and the existing `Panel.Content` bleed contract: the bled path keeps scroll and vertical rhythm, publishes `--bleed-px` matching the drawer's `px-6`, and the default path is untouched (verified: `Accordion.styles.ts` on `beta-release` consumes `--bleed-px` via `--accordion-x-padding`, so the mechanism works). Both paths are unit-tested, the story is tagged `component-test`, and the docs explain *when* to reach for the prop. No TypeScript, accessibility, or React performance findings.

### Issues

**Warnings (should fix)**
1. **The `Table` claim doesn't match reality** (docs `index.mdx`, changeset, `DrawerContent.tsx` TSDoc). The only consumer of `--bleed-px` on `beta-release` is `Accordion.styles.ts`. `Table`'s edge padding reads `[--cell-edge-padding:var(--panel-px,var(--cell-x-padding))]` (`Table.styles.ts:9`), wired to the Panel-specific `--panel-px`, not `--bleed-px`. A `Table` inside `<Drawer.Content bleed>` gets full-width backgrounds/dividers, but its cell text falls back to the small `--cell-x-padding` instead of aligning with the drawer title, which is exactly the misalignment this ticket set out to fix. Either extend Table's fallback chain (`var(--panel-px, var(--bleed-px, var(--cell-x-padding)))`) or scope the claims down to Accordion. See inline comments.

**Suggestions (nice to have)**
2. `Drawer.stories.tsx`: `BleedAccordion` enables the Chromatic snapshot on the base story (closed drawer, just the trigger button). The file convention is base `disableSnapshot: true` with the `.test()` re-enabling via `false`, so the snapshot captures the opened drawer after `play` (like `Basic.test`).
3. `DrawerContent.tsx`: the hardcoded `calc(var(--spacing)*6)` duplicates the theme's `px-6` from `ui-panel-content` (drift risk if the theme padding changes), and the bled path replaces `classNames.content` entirely, so `variant`/`size` on `Drawer.Content` are silently ignored when bled. Both are acceptable within the ticket's scope; a follow-up publishing a `--drawer-px` variable (like Panel's `--panel-px`) would give component and theme one source of truth.

### Recommendation

Non-blocking. Worth fixing the Table claim (Warning 1) before merge since it appears in three places and will mislead consumers; the fix is either one line in `Table.styles.ts` or a docs trim.

---

docs/content/components/overlay/drawer/index.mdx
sebald: Good catch — confirmed, `Table` read `--panel-px` and got nothing inside a bled Drawer. Fixed in 2a5539104 by extending Table's fallback chain to `var(--panel-px, var(--bleed-px, var(--cell-x-padding)))` (`Table.styles.ts`): it prefers the always-present Panel token, falls back to the generic `--bleed-px` published by other bled containers, then the plain cell padding. So Table now aligns in a bled Drawer and this sentence is accurate. Added a `BleedTable` story + test so it's covered.

Kept `--panel-px` first on purpose: `--bleed-px` is only published when bled, but Table must also align in *un*bled Panels where only `--panel-px` exists — collapsing onto `--bleed-px` alone would regress that.

.changeset/drawer-content-bleed.md
sebald: Fixed in 2a5539104 — `Table` is now genuinely wired (reads `--bleed-px` via its fallback chain), so listing it here is accurate. Updated the changeset wording to describe the actual wiring and added a `@marigold/theme-rui` patch since the fix touches `Table.styles.ts`, `ui.css`, and `tokens.css`.

packages/components/src/Drawer/DrawerContent.tsx
sebald: Kept the `Table` mention — it's true now (2a5539104): `Table`'s `--cell-edge-padding` falls back to `--bleed-px`, so a `Table` in a bled `Drawer.Content` aligns to the title like `Accordion` does.

packages/components/src/Drawer/Drawer.stories.tsx
sebald: Adopted the convention in 2a5539104: base story is `chromatic: { disableSnapshot: true }`, and the `.test()` re-enables with `disableSnapshot: false` so Chromatic captures the opened drawer after `play`. Applied the same to the new `BleedTable` story/test.

packages/components/src/Drawer/DrawerContent.tsx
sebald: Both addressed in 2a5539104:

1. No more hardcoded `calc(var(--spacing)*6)`. The horizontal padding shared by `ui-panel-header/content/actions` now comes from a single `--ui-panel-px` token (`tokens.css`), and the bled branch publishes `[--bleed-px:var(--ui-panel-px)]`. Component and theme reference the one token, so they can't drift. Went with `--ui-panel-px` rather than a per-component `--drawer-px` because the same padding is shared by Dialog/Tray too — and `--bleed-px` already is the generic child-facing contract.
2. Added a TSDoc note that `variant`/`size` on `Drawer.Content` are ignored when `bleed` is set (the bled branch replaces `classNames.content`).

themes/theme-rui/src/components/Table.styles.ts
```diff
@@ -4,9 +4,9 @@ export const Table: ThemeComponent<'Table'> = {
   table: cva({
     base: [
       'text-sm bg-surface',
-      // Edge padding for Panel integration — resolves to --panel-px
-      // when inside a Panel with bleed, otherwise falls back to --cell-x-padding
-      '[--cell-edge-padding:var(--panel-px,var(--cell-x-padding))]',
+      // Edge padding: Panel's `--panel-px`, then a bled container's
+      // `--bleed-px`, then the standalone cell default.
+      '[--cell-edge-padding:var(--panel-px,var(--bleed-px,var(--cell-x-padding)))]',
```
aromko, line 9: 🔴 **This change breaks the existing Panel bleed test — it's why Unit Tests (2/4) is red.**

`Panel.test.tsx:335` locates the bled content wrapper via `table.closest('[class*="--bleed-px"]')`. Now that the table's own base class contains `var(--bleed-px, …)` inside the `--cell-edge-padding` chain, `closest()` matches the **table itself** (it starts at the element it's called on), so the follow-up assertion `toContain('[--bleed-px:var(--panel-px)]')` fails.

Fix in `packages/components/src/Panel/Panel.test.tsx` (untouched by this PR, which is why it slipped): tighten the selector to match where the var is *set*, the same trick the new `Drawer.test.tsx` already uses:

```ts
const contentWrapper = table.closest('[class*="[--bleed-px:"]')!;
```

packages/components/src/Drawer/DrawerContent.tsx
```diff
@@ -29,7 +42,14 @@ export const DrawerContent = ({
 
   return (
     <div
-      className={cn('[grid-area:content]', classNames.content)}
+      className={cn(
+        '[grid-area:content]',
+        // Bled: drop the padding and re-publish `--ui-panel-px` as `--bleed-px`
+        // so edge-aware children inset to align with the title.
+        bleed
+          ? 'overflow-y-auto py-4 outline-none [--bleed-px:var(--ui-panel-px)]'
```
aromko, line 50: 💡 **Drift risk:** `'overflow-y-auto py-4 outline-none'` hand-copies the non-horizontal parts of the `ui-panel-content` utility. If that utility's vertical padding or scroll behavior ever changes, the bled branch silently diverges (the `--ui-panel-px` token only protects the horizontal axis). Consider splitting the utility (layout parts + separate px part), or at least adding a pointer comment on `ui-panel-content` in `ui.css` so future edits know about this copy.

themes/theme-rui/src/ui.css
```diff
@@ -289,15 +289,15 @@
  * these.
  */
 @utility ui-panel-header {
-  @apply border-border border-b px-6 py-4;
+  @apply border-border border-b px-(--ui-panel-px) py-4;
 }
 
 @utility ui-panel-content {
-  @apply overflow-y-auto px-6 py-4 outline-none;
+  @apply overflow-y-auto px-(--ui-panel-px) py-4 outline-none;
```
aromko, line 303: 💡 **Scope deviation, documented but worth an explicit ack:** DST-1635 explicitly excluded changing the shared `ui-panel-content` utility ("to avoid affecting Dialog/Tray/Sidebar"), yet this PR converts all three `ui-panel-*` utilities to `px-(--ui-panel-px)`. It's behavior-preserving (the token resolves to the same `24px`, and the changeset explains the single-source-of-truth rationale), but Dialog/Tray now depend on the new token — accepting the wider blast radius should be a conscious reviewer decision, not a later discovery.

themes/theme-rui/src/components/Table.styles.ts
sebald: Fixed in 0a97fb6ab — exactly as suggested. `Panel.test.tsx:337` now walks up via `table.closest('[class*="[--bleed-px:"]')`, matching the *declaration* rather than the `var()` reference the Table now carries. `pnpm test:unit` Panel suite is green again (36/36).

packages/components/src/Drawer/DrawerContent.tsx
sebald: Good call on the drift risk. Added a pointer comment on `ui-panel-content` in `ui.css` (fcbb733ef) noting that the Drawer `bleed` branch hand-copies the non-horizontal parts, so a future edit to `py-4`/scroll/outline knows to update the branch too. Kept it a comment rather than splitting the utility to stay within the ticket's scope.

themes/theme-rui/src/ui.css
sebald: Acknowledged and accepted. The conversion is behavior-preserving — `--ui-panel-px: --spacing(6)` resolves to the same 24px, so Dialog/Tray/Sidebar render identically; they just now resolve the padding through the token instead of a literal `px-6`. Taking on the wider blast radius on purpose: a single source of truth for the shared panel padding is worth more than keeping the utility literal, and it's what lets the `bleed` contract stay in sync.

sebald:
@aromko ready for another look 🙏

Addressed all three of the latest comments:

- 🔴 **Panel bleed test** — fixed in \`0a97fb6ab\`, exactly as suggested (`table.closest('[class*="[--bleed-px:"]')`). Panel unit suite green again (36/36).
- 💡 **`ui-panel-content` drift risk** — added a pointer comment on the utility (\`fcbb733ef\`) so a future edit to the vertical padding/scroll/outline knows the Drawer `bleed` branch copies it.
- 💡 **Scope deviation** — acknowledged; the `px-6` → `px-(--ui-panel-px)` conversion is behavior-preserving (token resolves to the same 24px), taken on intentionally for a single source of truth.

All eight review threads are resolved.

aromko (approved):

PR #5654, opened 2026-07-17
feat(DST-1609): two-level sidebar navigation (rail + panel)

sebald:
# Description

Adds a two-level navigation mode to the sidebar: `Sidebar.Rail` renders a persistent rail of icon-first top-level destinations next to a panel showing the active section's sub-navigation. A `Sidebar.RailItem` wrapping a `Sidebar.Nav` is a section; one with only an `href` is a direct link; one inside `Sidebar.Footer` pins to the bottom of the rail. Its `active` prop overrides href matching for pages the URL can't identify (mirroring `Sidebar.Item`). Collapsing (toggle or Cmd/Ctrl+B) hides the panel while the rail narrows to an icon-only strip; on small screens the drawer shows rail and panel side by side (section taps swap the panel, links close the drawer).

The shell adapts without configuration: `AppShell` switches to a full-width top bar when it contains a rail — detected in pure CSS via `:has([data-rail])` — so the brand never moves when the panel collapses. `Sidebar.Toggle` gains `variant="rail"`. Rail dimensions and the shell's shared datum live in the theme as new tokens (`--spacing-topbar`, `--spacing-rail` / `--spacing-rail-collapsed` / `--spacing-rail-panel`). The `TopNavigation` bottom edge is now an always-on border, replacing the non-reusable `ui-scroll-edge`/`ui-sidebar-seam-header` scroll-driven utilities. Idle single-column nav labels also darken a step (new `--color-secondary-bold` token) so they out-rank the quiet group-label captions.

Closes DST-1609

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm sb`, open `Components/Sidebar` → **Rail**: switch sections (panel swaps, focus moves to its heading), re-click the active section (panel toggles), collapse via the top-bar toggle or Cmd/Ctrl+B (rail narrows to icons, tooltips appear on hover), click "Berichte" (direct link — no panel, toggle stays live). Keyboard: every rail tile is a `Tab` stop, and Arrow keys / Home / End also move between them (including the pinned footer). In the section panel, rove focus down a few items, then change the current page — `Tab` should re-enter the panel at the new current item, not the roved one.
2. Open **RailMobile** (or resize to <600px): the drawer shows rail + panel side by side; section taps swap the panel in place; leaf/direct links close the drawer; the exposed backdrop strip dismisses on tap.
3. Compare with `Components/AppShell` → **Basic**: the single-column sidebar keeps its full-height, inset-header layout — the header-first switch applies only when a rail is present.
4. Docs: `pnpm start` → `/components/navigation/sidebar` ("Two-level navigation" section, rail anatomy, `Sidebar.Rail`/`Sidebar.RailItem` prop tables), plus the notes on `/components/layout/app-shell` and `/patterns/layout/app-frame`.

## Breaking Changes

No. All new API surface is additive. Two visual changes reach the existing single-column sidebar: the top bar's bottom hairline is now always on (was scroll-revealed), and idle nav labels darken one step (charcoal-600 → charcoal-700). The removed `ui-scroll-edge`/`ui-sidebar-seam-header` utilities were documented as non-reusable, layout-specific helpers.

## Checklist


packages/components/src/Sidebar/SidebarRail.tsx
```diff
@@ -0,0 +1,304 @@
...
+    </div>
+  );
+
+  const panelBody =
+    selectedNode?.isSection && selectedNode.nav ? (
+      <>
+        <h2 ref={titleRef} tabIndex={-1} className={classNames.panelTitle}>
+          {selectedNode.panelTitle}
+        </h2>
+        {cloneElement(selectedNode.nav, { current })}
```
aromko: 💡 **Panel landmark name falls back to the generic default**

If a section's `Sidebar.Nav` omits `aria-label`, the panel landmark is announced as the localized "App Navigation" while the visible heading shows the section name. `panelTitle` is already computed (nav `aria-label` → item text), so it could be threaded through:

```suggestion
        {cloneElement(selectedNode.nav, {
          current,
          'aria-label': selectedNode.panelTitle,
        })}
```

Keeps the landmark name and the visible heading in sync for un-labelled navs.

packages/components/src/Sidebar/SidebarRail.test.tsx
```diff
@@ -0,0 +1,469 @@
...
+    await user.keyboard('{/Control}');
+    stop();
+
+    // Re-clicking the active section normally toggles the panel and prevents
+    // navigation — a browser-owned click must do neither.
+    expect(result.prevented).toBe(false);
+    expect(toggle).toHaveAttribute('aria-expanded', 'true');
+  });
+
+  test('an onPress-only rail item activates with Enter and Space', async () => {
```
aromko, line 472: 💡 A few of these unit tests bundle two act→assert rounds (Enter then Space here; click then Cmd+B in "the toggle stays live on a direct-link page"). Splitting them into one Arrange–Act–Assert per test would match the convention used in the rest of the suite — same coverage, easier failure diagnosis.

packages/components/src/AppShell/AppShell.tsx
```diff
@@ -54,7 +55,16 @@ export const AppShell = ({
...
     <div
       ref={ref}
-      className="grid min-h-dvh grid-cols-[auto_1fr] grid-rows-[3.5rem_1fr] [grid-template-areas:'sidebar_header'_'sidebar_main']"
+      // The header row is `auto` — its height comes from `<TopNavigation>`.
+      // Single-column sidebar → sidebar-first (sidebar full height, top bar to
+      // its right). Two-level rail (detected via the aside's `data-rail`) →
+      // header-first: the top bar spans full width so its brand never moves
+      // when the panel collapses, and the rail hangs below it.
+      className={cn(
+        'grid min-h-dvh grid-cols-[auto_1fr] grid-rows-[auto_1fr]',
```
aromko: 💡 Observation: the header row going `3.5rem` → `auto` means an `AppShell` without a `TopNavigation` no longer reserves the empty band — a third (likely welcome) visual change to the existing shell alongside the two listed under Breaking Changes. Worth a line in the changeset/PR body?

packages/components/src/Sidebar/SidebarRail.stories.tsx
```diff
@@ -0,0 +1,475 @@
...
+              </Panel>
+            </Page>
+          </AppShell>
+        </I18nProvider>
+      </RouterProvider>
+    </div>
+  );
+};
+
+export const Rail = meta.story({
```
aromko, line 343: Except this all other stories and tests can be disabled for chromatic?

packages/components/src/Sidebar/SidebarRailItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * The rail label (plain text) plus, for a section, a nested `Sidebar.Nav`
+   * whose items become the panel. Omit the nav to make the item a direct link.
+   */
+  children?: ReactNode;
+}
+
+const _SidebarRailItem = (_props: SidebarRailItemProps): ReactElement | null =>
+  null;
+_SidebarRailItem.__SIDEBAR_RAIL_ITEM__ = true as const;
```
aromko, line 43: maybe blank line before and after

packages/components/src/Sidebar/SidebarRail.tsx
sebald: Good catch — done in ac2bdba16. `panelTitle` is already "nav aria-label → item text", so threading it into the cloned nav as the default `aria-label` is a no-op when the consumer labelled their nav and fills in the section name when they didn't. Landmark name and visible heading now stay in sync.

packages/components/src/AppShell/AppShell.tsx
sebald: Agreed it's a third change. Documented it in both the changeset and the PR body: the `AppShell` header row is now `auto` (was fixed `3.5rem`), so a shell without a `TopNavigation` no longer reserves an empty header band.

packages/components/src/Sidebar/SidebarRail.stories.tsx
sebald: `RailControlled` was already `chromatic: { disableSnapshot: true }`, and I've now disabled the snapshot on `RailMobile` too — `Rail` stays as the single visual baseline. The `.test` play functions still run regardless; only the Chromatic snapshots are trimmed.

packages/components/src/Sidebar/SidebarRailItem.tsx
sebald: Done in ac2bdba16.

packages/components/src/Sidebar/SidebarRail.test.tsx
sebald: Left as-is for now: each bundles the two activation paths of a single behavior (click+Cmd/Ctrl+B for the toggle; Enter+Space for onPress), so the pairing is intentional rather than two unrelated scenarios. Happy to split them if you'd prefer strict one-Act-per-test consistency.

docs/content/foundations/elevation/elevation-layers.demo.tsx
```diff
@@ -1,19 +1,26 @@
 import { Text } from '@marigold/components';
 
 export default () => (
-  <div className="bg-background rounded-xl p-8">
```
sarahgm, line 4: Why do you change this demo? It should not change

packages/components/src/Sidebar/SidebarRail.tsx
```diff
@@ -0,0 +1,317 @@
...
+const RailItemLink = ({
+  node,
+  selected,
+  matched,
+  onActivate,
+  collapsed,
+  className,
+}: RailItemLinkProps) => {
+  const router = useRouter();
+  const routerLinkProps = useLinkProps({ href: node.href });
```
sarahgm: 💡 `useLinkProps` here is dead. With only `href` passed in it returns just `{ href: router.useHref(node.href) }`, and the `{...routerLinkProps}` spread on line 83 is immediately overridden by `href={node.href}` on the next line — so it contributes nothing. This also matches the sibling `SidebarLink`, which renders a raw `href` and routes via `handleLinkClick` in `onClick` without `useLinkProps`. Safe to drop the import (line 6), the `routerLinkProps` variable, and the spread.

packages/components/src/Sidebar/railCollection.ts
```diff
@@ -0,0 +1,265 @@
...
+ *   `leaves` drive active-ancestor resolution and `href` is the landing page.
+ * - A **direct link** has an `href` and no `nav`; it navigates and shows no panel.
+ */
+export interface SidebarRailNode {
+  type: 'railItem';
+  key: string;
+  /** Full accessible name (tooltip + `aria-label`). */
+  textValue: string;
+  /** Rail label content (the text shown under the icon). */
+  label: ReactNode[];
```
sarahgm: 💡 The `label` field is set on every node (line 144) but never read back off it — the tile renders `node.textValue` (`SidebarRail.tsx:105`) and nothing reads `.label`. It's only used locally in `buildNode` to derive `textValue` via `extractTextValue(label)`. Keep `label` as a local and drop it from the `SidebarRailNode` interface + the returned object. The doc comment ("the text shown under the icon") is also inaccurate since nothing renders it.

packages/components/src/Sidebar/railCollection.ts
```diff
@@ -0,0 +1,265 @@
...
+    } else if (node.href) {
+      const tag = `leaf-${index++}`;
+      tagged.push({ key: tag, href: node.href });
+      owner.set(tag, node.key);
+    }
+  }
+
+  // `resolveCurrent`'s per-nav warning never sees cross-section duplicates
+  // (only one nav renders), so check the whole rail here.
+  if (process.env.NODE_ENV !== 'production') {
```
sarahgm: 💡 This dev-only duplicate-href warning is a near-copy of the one in `collection.ts` `resolveCurrent` (line 235) — same normalize→`Map`→`console.error` loop, differing only in wording and how the owner is named. Consider a shared `warnDuplicateHrefs(leaves, format)` helper in `collection.ts`, next to `matchLeaves` which both already share.

packages/components/src/Sidebar/SidebarRail.tsx
```diff
@@ -0,0 +1,317 @@
...
+        ref={railNavRef}
+        aria-label={ariaLabel || stringFormatter.format('railNavigation')}
+        className="flex min-h-0 flex-1 flex-col overflow-hidden"
+        onKeyDown={onRailKeyDown}
+      >
+        <div className={classNames.rail}>
+          {nodes
+            .filter(node => !node.inFooter)
+            .map(node => (
+              <RailItemLink
```
sarahgm: 💡 This `<RailItemLink … />` block is identical to the pinned-footer one at line 217, differing only in the source array. A local `renderRailItem = node => <RailItemLink … />` mapped over both lists removes the duplication.

packages/components/src/Sidebar/Context.tsx
```diff
@@ -57,8 +57,26 @@ export const SidebarProvider = ({
...
+        toggleSidebar();
+      }
+    };
+    window.addEventListener('keydown', handler);
+    return () => window.removeEventListener('keydown', handler);
+  }, [toggleSidebar]);
+
   const value = useMemo(
-    () => ({ state, toggleSidebar, isMobile, classNames }),
+    () => ({
```
sarahgm: 💡 Nit: this `value` object was expanded from one line to six with identical contents and deps — pure churn (it fit on one line before, so Prettier didn't force it). Reverting to `() => ({ state, toggleSidebar, isMobile, classNames })` keeps the diff focused on the Cmd/Ctrl+B effect just above.

sarahgm (commented):
Automated simplification pass (review-pr skill) — 5 non-blocking cleanups not already covered by the existing threads. All logic looks sound and well-tested; these are dead-code removals and minor DRY/churn nits.

packages/components/src/Sidebar/collection.test.tsx
```diff
@@ -441,3 +441,45 @@ describe('applyCurrent', () => {
...
+        Branch
+        <>
+          <SidebarItem id="nested" href="/nested">
+            Nested
+          </SidebarItem>
+        </>
+      </SidebarItem>
+    );
+
+    const branch = collection.getItem('branch');
```
sarahgm, line 481: blank line between AAA blocks

themes/theme-rui/src/tokens.css
```diff
@@ -68,6 +69,10 @@
   --color-background: var(--color-charcoal-100);
   --color-foreground: var(--color-charcoal-900);
   --color-secondary: var(--color-charcoal-600);
+  /* Emphatic sibling of `secondary` (cf. `selected`/`selected-bold`): text that
+     must out-rank quiet captions without reaching full-strength `foreground` —
+     e.g. idle, interactive sidebar nav labels sitting above `groupLabel`s. */
+  --color-secondary-bold: var(--color-charcoal-700);
```
sarahgm, line 75: Is this new token really needed? If so it must be mentioned in the Token Overview page

docs/content/foundations/elevation/elevation-layers.demo.tsx
sebald: You're right — this crept in from the DST-1602 surface work and doesn't belong here. Reverted to the beta-release version in 507a47b8a.

packages/components/src/Sidebar/SidebarRail.tsx
sebald: Confirmed dead and removed in 507a47b8a — with only `href` passed, `useLinkProps` returns just `{ href: router.useHref(href) }`, which the raw `href={node.href}` on the next line overrode anyway.

One aside: `SidebarLink` actually has the same spread-then-override pattern (its `{...routerLinkProps}` is overridden by `href={href}` too) — pre-existing from DST-979/#5188, so I left it out of this PR. Happy to clean it up in a follow-up.

packages/components/src/Sidebar/railCollection.ts
sebald: Done in 507a47b8a — `label` is gone from `SidebarRailNode` and the returned node; it stays a local in `buildNode` where `extractTextValue(label)` derives the accessible name. The `textValue` doc comment now notes it's also the visible tile label.

packages/components/src/Sidebar/SidebarRail.tsx
sebald: Done in 507a47b8a — extracted a local `renderRailItem` inside `railColumn` (it closes over `collapsed`) and mapped it over both the list and the pinned footer.

packages/components/src/Sidebar/Context.tsx
sebald: Reverted to the one-liner in 507a47b8a — the diff there is now just the Cmd/Ctrl+B effect.

packages/components/src/Sidebar/collection.test.tsx
sebald: Done in 507a47b8a.

packages/components/src/Sidebar/railCollection.ts
sebald: Done in 7452717ff — extracted `warnDuplicateHrefs(leaves, format)` into `collection.ts` next to `matchLeaves`. The `format` callback keeps each caller's exact message and vocabulary: `resolveCurrent` reports leaf keys (and keeps its string-mode-only guard), `resolveActiveRail` reports owner-mapped rail items and warns in both modes, as before. Also let `railCollection.ts` drop its now-unused `normalizePath` import.

themes/theme-rui/src/tokens.css
sebald: Yes, it's needed: the idle rail/nav labels must out-rank the quiet `groupLabel` captions without jumping to full-strength `foreground` — `secondary` (charcoal-600) is too quiet for interactive rows, `foreground` (charcoal-900) would flatten the hierarchy. It follows the system's existing `-bold` emphasis modifier (`selected-bold`, `destructive-bold`, `focus-highlight-bold`).

Documented it in 7452717ff on the Token Overview page: table entry under *Page background & text* plus the "No pairing needed" callout now covers it as a standalone text color.

packages/components/src/Sidebar/SidebarRail.tsx
sarahgm: OK so clean it up in a Follow-Up ?

sebald: Yes — tracked as **DST-1647** (🧹🏚️ Remove dead `useLinkProps` spread in `SidebarLink`). Same treatment as the `RailItemLink` cleanup here: drop the `useLinkProps` import, the `routerLinkProps` variable, and the spread, since the raw `href` overrides it anyway. Kept out of this PR to keep the diff focused on the rail.

aromko (approved):

sarahgm (approved):

sebald:
@sarahgm @aromko — merged the latest `beta-release` into this branch and resolved the two conflicts, so the PR is green-path again: **mergeable**, both approvals still standing, CI re-running on the merge commit (`0b299d45b`).

**Conflict resolutions:**
- **`themes/theme-rui/src/components/Menu.styles.ts`** — kept `beta-release`'s 3-column grid (`auto_1fr_auto`, for the new shortcut-hint column) *and* our semantic `max-sm:min-h-touch-target` token (same 44px as the old `min-h-11`). Non-overlapping, so both apply.
- **`docs/app/(examples)/examples/navigation.tsx`** — kept our two-level rail model (`RailTile` / `ShellConfig`); `beta-release`'s `userInput: NavSection` shape no longer exists in the merged `_shared`. Registered `beta-release`'s new **Pick** example as a rail link tile so `/examples/pick` stays linked and searchable — the merged `navigation.test.ts` expectation for it passes.

Verified locally after the merge: examples navigation test 5/5, root + docs typecheck, and lint all clean.

Repo is squash-only — **feel free to merge whenever you're happy with the changes.** No further action needed from my side.

PR #5657, opened 2026-07-17
fix(Autocomplete): correct onSubmit argument order to match documented signature

zigzagdev:
# Description

Autocomplete's public `onSubmit` type is documented as `(value, key) => void`, but the implementation had it backwards: selecting an option called it with `(key, null)`, and typing a custom value called it with `(null, text)`. 
Consumers following the documented contract would treat selections as free-text searches and free text as selected keys. So, this fixes both call sites to match the documented signature. 

No UI/visual change — internal argument-order fix only.

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

```bash

$ vitest --config vite.config.ts --project=unit-tests packages/components/src/Autocomplete/Autocomplete.test.tsx

 RUN  v4.1.9 /app

stderr | packages/components/src/Autocomplete/Autocomplete.test.tsx > mobile view > renders mobile trigger with placeholder

A PressResponder was rendered without a pressable child. Either call the usePress hook, or wrap your DOM node with <Pressable> component.               

stderr | packages/components/src/Autocomplete/Autocomplete.test.tsx > mobile view > opens tray and shows empty state

A PressResponder was rendered without a pressable child. Either call the usePress hook, or wrap your DOM node with <Pressable> component.               

 ✓  unit-tests (firefox)  packages/components/src/Autocomplete/Autocomplete.test.tsx (23 tests) 1590ms

 Test Files  1 passed (1)

      Tests  23 passed (23)

   Start at  11:23:36

   Duration  30.63s (transform 0ms, setup 100ms, import 24.79s, tests 1.59s, environment 0ms)

```

## Test Instructions

Run `pnpm test:unit packages/components/src/Autocomplete/Autocomplete.test.tsx` — updated assertions verify `onSubmit` receives `(value, key)` in the correct order for both selection and free-text submission.

## Breaking Changes

No

## Checklist

### DIFF
--- a/.changeset/autocomplete-onsubmit-argument-order.md
+++ b/.changeset/autocomplete-onsubmit-argument-order.md
@@ -1,0 +1,5 @@
+---
+'@marigold/components': patch
+---
+
+Fix `Autocomplete`'s `onSubmit` argument order to match the documented `(value, key)` signature.
--- a/.changeset/autocomplete-onsubmit-argument-order.md
+++ b/.changeset/autocomplete-onsubmit-argument-order.md
@@ -2,4 +2,4 @@
 '@marigold/components': patch
 ---
 
-Fix `Autocomplete`'s `onSubmit` argument order to match the documented `(value, key)` signature.
+Fix `Autocomplete`'s public `onSubmit` type, which declared `(value, key)` while the implementation, JSDoc, and both docs demos all used `(key, value)`. The type now matches the actual `(key, value)` signature — no runtime behavior changes.
--- a/packages/components/src/Autocomplete/Autocomplete.test.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.test.tsx
@@ -228,7 +228,7 @@ test('calls onSubmit with custom value on Enter when no option is focused', asyn
   const input = screen.getByRole('combobox');
   await user.type(input, 'custom value{Enter}');
 
-  expect(onSubmit).toHaveBeenCalledWith('custom value', null);
+  expect(onSubmit).toHaveBeenCalledWith(null, 'custom value');
 });
 
 test('calls onSubmit with the selected key when an option is chosen', async () => {
@@ -241,7 +241,7 @@ test('calls onSubmit with the selected key when an option is chosen', async () =
   const option = await screen.findByRole('option', { name: 'Harry Potter' });
   await user.click(option);
 
-  expect(onSubmit).toHaveBeenCalledWith(null, 'Harry Potter');
+  expect(onSubmit).toHaveBeenCalledWith('Harry Potter', null);
 });
 
 describe('mobile view', () => {
--- a/packages/components/src/Autocomplete/Autocomplete.test.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.test.tsx
@@ -228,7 +228,20 @@ test('calls onSubmit with custom value on Enter when no option is focused', asyn
   const input = screen.getByRole('combobox');
   await user.type(input, 'custom value{Enter}');
 
-  expect(onSubmit).toHaveBeenCalled();
+  expect(onSubmit).toHaveBeenCalledWith('custom value', null);
+});
+
+test('calls onSubmit with the selected key when an option is chosen', async () => {
+  const onSubmit = vi.fn();
+  renderWithOverlay(<Basic.Component label="Label" onSubmit={onSubmit} />);
+
+  const input = screen.getByRole('combobox');
+  await user.type(input, 'ha');
+
+  const option = await screen.findByRole('option', { name: 'Harry Potter' });
+  await user.click(option);
+
+  expect(onSubmit).toHaveBeenCalledWith(null, 'Harry Potter');
 });
 
 describe('mobile view', () => {
--- a/packages/components/src/Autocomplete/Autocomplete.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.tsx
@@ -229,7 +229,7 @@ const _Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>(
     ref
   ) => {
     const props: RAC.ComboBoxProps<object> = {
-      onSelectionChange: key => key !== null && onSubmit?.(null, key),
+      onSelectionChange: key => key !== null && onSubmit?.(key, null),
       defaultInputValue: defaultValue,
       inputValue: value,
       onInputChange: onChange,
--- a/packages/components/src/Autocomplete/Autocomplete.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.tsx
@@ -29,7 +29,7 @@ interface AutocompleteInputProps {
    * A `value` will be passed if the submission is a custom value (e.g. a user types then presses enter).
    * If the input is a selected item, `value` will be `null`.
    */
-  onSubmit?: (key: Key | null, value: string | null) => void;
+  onSubmit?: (value: string | null, key: Key | null) => void;
 
   /**
    * Called when the clear button is pressed.
@@ -77,7 +77,7 @@ const AutocompleteInput = ({
         }
         if (e.key === 'Enter') {
           if (state?.selectionManager.focusedKey === null) {
-            onSubmit?.(null, state?.inputValue);
+            onSubmit?.(state?.inputValue, null);
           }
         }
       }}
@@ -229,7 +229,7 @@ const _Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>(
     ref
   ) => {
     const props: RAC.ComboBoxProps<object> = {
-      onSelectionChange: key => key !== null && onSubmit?.(key, null),
+      onSelectionChange: key => key !== null && onSubmit?.(null, key),
       defaultInputValue: defaultValue,
       inputValue: value,
       onInputChange: onChange,
--- a/packages/components/src/Autocomplete/Autocomplete.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.tsx
@@ -29,7 +29,7 @@ interface AutocompleteInputProps {
    * A `value` will be passed if the submission is a custom value (e.g. a user types then presses enter).
    * If the input is a selected item, `value` will be `null`.
    */
-  onSubmit?: (value: string | null, key: Key | null) => void;
+  onSubmit?: (key: Key | null, value: string | null) => void;
 
   /**
    * Called when the clear button is pressed.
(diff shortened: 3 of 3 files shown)


packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -229,7 +229,7 @@ const _Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>(
     ref
   ) => {
     const props: RAC.ComboBoxProps<object> = {
-      onSelectionChange: key => key !== null && onSubmit?.(key, null),
+      onSelectionChange: key => key !== null && onSubmit?.(null, key),
```
sebald: 🔴 This is the highest-impact line. Passing `(null, key)` on selection cascades into breakage outside this PR:

- **`Autocomplete.stories.tsx` → `Controlled.test`** (a `component-test`) selects an option and asserts the *first* `onSubmit` arg is `'harry-potter'`. I ran `pnpm test:sb` with this change applied — it fails (`Expected: harry-potter, Received: (empty)`). Please update it here.
- **`autocomplete-async.demo.tsx`** interprets args as `(key, value)`; after this change, selecting an option hits its `if (value)` branch and shows *all* results instead of the selected item.
- **`autocomplete-appearance.demo.tsx`** displays the two values in swapped slots.

All three need updating in this PR (or the direction reconsidered — see the type comment).

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -29,7 +29,7 @@ interface AutocompleteInputProps {
    * A `value` will be passed if the submission is a custom value (e.g. a user types then presses enter).
    * If the input is a selected item, `value` will be `null`.
    */
-  onSubmit?: (key: Key | null, value: string | null) => void;
+  onSubmit?: (value: string | null, key: Key | null) => void;
```
sebald: 💡 This public type was the **only** signal pointing at `(value, key)`. The internal type, the JSDoc prose just above, the implementation, and both docs demos all used `(key, value)`. So there are two valid fixes:

1. Flip the runtime to match this type (what this PR does) — breaks the demos + story test.
2. Flip this type to `(key, value)` — zero runtime change, demos keep working.

Worth a maintainer decision before landing. Whichever way it goes, update the JSDoc prose (lines above) so it leads with the same parameter as the signature.

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -77,7 +77,7 @@ const AutocompleteInput = ({
         }
         if (e.key === 'Enter') {
           if (state?.selectionManager.focusedKey === null) {
```
sebald, line 79: ⚠️ Same flip as the selection path — consistent with the chosen direction. Just note the JSDoc for this handler (above) still describes `key` before `value`; update it to match.

.changeset/autocomplete-onsubmit-argument-order.md
```diff
@@ -0,0 +1,5 @@
+---
+'@marigold/components': patch
```
sebald, line 2: ⚠️ `patch` under-states this. Reordering a public callback's arguments is a behavior change for any consumer who followed the prose/demos. Consider `minor` with an explicit before/after migration note. (Also: PR body says "Breaking Changes: No", which contradicts this.)

packages/components/src/Autocomplete/Autocomplete.test.tsx
```diff
@@ -228,7 +228,20 @@ test('calls onSubmit with custom value on Enter when no option is focused', asyn
   const input = screen.getByRole('combobox');
   await user.type(input, 'custom value{Enter}');
 
-  expect(onSubmit).toHaveBeenCalled();
+  expect(onSubmit).toHaveBeenCalledWith('custom value', null);
+});
+
+test('calls onSubmit with the selected key when an option is chosen', async () => {
```
sebald, line 234: 💡 Good addition and correctly asserts `(null, key)` on selection. Please mirror this contract in the `Controlled` story test and the two docs demos so the API is asserted consistently in every place it's documented.

sebald (changes requested):
## Automated Code Review

Thanks for spotting this — there **is** a real inconsistency in the `onSubmit` API. But the fix as written is incomplete and its framing needs a second look, so I'm requesting changes.

### Is this actually a bugfix? Partly — the "documented signature" is self-contradictory

The PR says the implementation "had it backwards" vs the documented `(value, key)`. In fact only **one** signal says `(value, key)` — the public type at `AutocompleteProps.onSubmit`. **Six** other signals all say `(key, value)`:

| Signal | Order |
|--------|-------|
| Public type signature (`AutocompleteProps.onSubmit`) | `(value, key)` |
| Implementation — selection (`onSubmit(key, null)`) | `(key, value)` |
| Implementation — custom value (`onSubmit(null, inputValue)`) | `(key, value)` |
| Internal type (`AutocompleteInputProps`) | `(key, value)` |
| Public JSDoc prose (describes `key` first) | `(key, value)` |
| Docs demo `autocomplete-async.demo.tsx` (`handleSubmit(key, value)`) | `(key, value)` |
| Docs demo `autocomplete-appearance.demo.tsx` (`(id, val)`) | `(key, value)` |

So the docs page itself is inconsistent: the generated props table shows `(value, key)` while the prose right above it and both live demos use `(key, value)`. The real defect is the **type declaration disagreeing with everything else**. Flipping the runtime to match the type is a defensible choice, but it's the *more disruptive* direction and needs maintainer sign-off — flipping the type to `(key, value)` instead would be a zero-runtime-change fix that keeps the demos working.

### 🔴 Critical — this PR breaks things it doesn't touch

**1. Breaks the `Controlled` story test (`component-test`) — verified.** `Autocomplete.stories.tsx` → `Controlled.test` selects the *Harry Potter* option and asserts the **first** `onSubmit` arg is `'harry-potter'`. I applied this PR locally and ran `pnpm test:sb`:

```
× Reflects the controlled value and submitted key
  Expected element to have text content: harry-potter
  Received: submitted value:            (empty)
Tests  1 failed | 13 passed (14)
```

This story isn't in the diff, so please update it here.

**2. Silently breaks `docs/.../autocomplete-async.demo.tsx`** (the featured async example). `handleSubmit = (key, value) => { if (key) findItem(); if (value) showAll() }`. After the flip, selecting an option calls `onSubmit(null, key)` → `handleSubmit(key=null, value=<the key>)` → the `if (value)` branch runs and it shows **all** results instead of the selected character. The demo's behavior inverts.

**3. Breaks `docs/.../autocomplete-appearance.demo.tsx`** — `onSubmit={(id, val) => setSubmitted([id, val])}` displays the two values in swapped slots after the flip.

### 🟡 Should fix

**4. Mislabeled as non-breaking.** "Breaking Changes: No" + `patch`. Flipping the argument order of a public callback breaks every consumer who followed the prose/demos (all `(key, value)`). This warrants a `minor` + explicit migration note, not a silent `patch`.

**5. JSDoc prose not updated** — both blocks still introduce `key` before `value`, perpetuating the ambiguity that caused this.

**6. Direction needs sign-off**, and the title/branch lack the `DST-XXXX` scope required by the repo's Conventional-Commits convention.

### Note on VRT

No visual surface changes here, so I would **not** block on Chromatic despite the component-src touch.

---

packages/components/src/Autocomplete/Autocomplete.tsx
zigzagdev: @sebald 

[fix it](https://github.com/marigold-ui/marigold/pull/5657/commits/6da60f4c3d2743bd27664af6c156edbc03825f2f).

Keep as `(key, value)`.
Verified with `pnpm test:sb`: Controlled story's 14 tests, including the selection assertion and all tests passed.

packages/components/src/Autocomplete/Autocomplete.tsx
zigzagdev: @sebald 

[fixed it](https://github.com/marigold-ui/marigold/pull/5657/commits/c0009d2c4109adf45f0fd50726b63e3f4dc9d5af)

Flipped both type declarations to `(key, value)` in `[c0009d2c](https://github.com/marigold-ui/marigold/pull/5657/commits/c0009d2c4109adf45f0fd50726b63e3f4dc9d5af)`, zero runtime change. 
JSDoc already led with `key` first, so no prose update needed at here.

packages/components/src/Autocomplete/Autocomplete.tsx
zigzagdev: @sebald 

[fixed it](https://github.com/marigold-ui/marigold/pull/5657/commits/24c0523db5a870608cc7e091d6fd670b2df5fa02).

Reeerted in `[24c0523d](https://github.com/marigold-ui/marigold/pull/5657/commits/24c0523db5a870608cc7e091d6fd670b2df5fa02)`, consistent with the chosen direction. 
JSDoc above already led with `key`, so it already matches.

.changeset/autocomplete-onsubmit-argument-order.md
zigzagdev: @sebald 

[fixed it](https://github.com/marigold-ui/marigold/pull/5657/commits/262a2299a50a4b5509dfacfdeab20efff1de7f4b).

Reworded in `[262a2299](https://github.com/marigold-ui/marigold/pull/5657/commits/262a2299a50a4b5509dfacfdeab20efff1de7f4b)`. 
Since we flipped the type instead of the runtime, there's no behavior change — `patch` is accurate now and matches the PR body's "Breaking Changes: No".

packages/components/src/Autocomplete/Autocomplete.test.tsx
zigzagdev: @sebald 

[fixed it](https://github.com/marigold-ui/marigold/pull/5657/commits/9db6b926ed94132e1e41a3b407ea4b7921cd465c).

Assertion flipped to `(key, value)` in `[9db6b926](https://github.com/marigold-ui/marigold/pull/5657/commits/9db6b926ed94132e1e41a3b407ea4b7921cd465c)`. 
No changes needed in the Controlled story or the two docs demos — they already used that order.

sebald (approved):
## Approving — all review threads resolved

Thanks @zigzagdev. This supersedes my earlier `CHANGES_REQUESTED`. We landed on the zero-runtime-change direction I preferred: flip the public `onSubmit` **type** to `(key, value)` so it matches the implementation, internal type, JSDoc prose, and both docs demos — rather than reordering the runtime and breaking the `Controlled` story test and the two demos.

Verified against the final tree (`262a2299`):

- ✅ Public `onSubmit` type is now `(key: Key | null, value: string | number | null)`
- ✅ JSDoc already leads with `key`, so prose and signature agree
- ✅ Test asserts `(key, value)` for both selection (`('Harry Potter', null)`) and custom value (`(null, 'custom value')`)
- ✅ Changeset is `patch` with an accurate "no runtime behavior changes" note — correct now that only the type changed
- ✅ No demo/story changes needed — they already used `(key, value)`

Type-only runtime impact; no visual surface change. LGTM. 🚀

zigzagdev:
@sebald 

Thank you for the thorough review and for verifying each point carefully! 
Thank you for appreciate the detailed feedback throughout the process.

PR #5311, opened 2026-04-13
refa([DST-1283]): remove Multiselect component and related files

OsamaAbdellateef:
### ‼️‼️‼️ BREAKING CHANGES ‼️‼️‼️

# Description

- [ ] Remove MultiSelect component source (packages/components/src/Multiselect/)
- [ ] Remove MultiSelect export from packages/components/src/index.ts
- [ ] Remove MultiSelect theme styles (themes/theme-rui/src/components/Multiselect.styles.ts) and its export
- [ ] Remove MultiSelect type from theme definition (packages/system/src/types/theme.ts)
- [ ] Remove MultiSelect documentation (docs/content/components/form/multiselect/)
- [ ] Update multiple selection pattern docs (docs/content/patterns/multiple-selection/index.mdx)
- [ ] Remove react-select from root package.json and Storybook optimizeDeps config
- [ ] Remove anatomy image (docs/public/multiselect/)
- [ ] Clean up demo registry entries in docs/.registry/demos.tsx

# Reviewers:

@marigold-ui/developer

# Pull Request Checklist:

- [x] Marigold docs and Storybook Preview is available
- [ ] Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [ ] Added/Updated documentation (if it already exists for this component).
- [ ] Updated visual regression tests (only necessary when ui changes in the PR)

PR #5659, opened 2026-07-18
fix(Radio): apply width via CSS variable instead of raw class name

zigzagdev:
# Description

`Radio` and `Multiselect` built their `width` class name from the raw token directly (e.g. a literal `"1/2"` used as a class, or a template-literal `` `w-${width}` ``), which Tailwind can't statically detect at build time, so no CSS was ever generated — setting `width` had no visible effect. 
Both now follow the same pattern already used by `FieldBase`/`TextField`/`NumberField`: a static `w-(--field-width)` class with the actual value injected via `createWidthVar`. 
Individual `Radio` items inside a sized `Radio.Group` now inherit the group's already-computed width instead of recomputing it, which previously caused a double-shrink (e.g. `width="1/2"` rendering at 1/4 instead of 1/2).

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Storybook → `Components/Radio` → `Basic`: set the `width` control (e.g. `1/2`, `full`) — the whole group and each individual radio item resize together, matching the set width.                                              
2. Storybook → `Components/Multiselect` → any story: set the `width` control — the field resizes accordingly.    
3. `pnpm test:unit packages/components/src/Radio/Radio.test.tsx packages/components/src/Multiselect/Multiselect.test.tsx` — updated assertions cover the new class/style output.

## Breaking Changes

No

## Checklist

### DIFF
--- a/packages/components/src/Multiselect/Multiselect.tsx
+++ b/packages/components/src/Multiselect/Multiselect.tsx
@@ -19,7 +19,12 @@ import Select, {
 import { useField } from '@react-aria/label';
 import { useId } from '@react-aria/utils';
 import type { WidthProp } from '@marigold/system';
-import { ComponentClassNames, cn, useClassNames } from '@marigold/system';
+import {
+  ComponentClassNames,
+  cn,
+  createWidthVar,
+  useClassNames,
+} from '@marigold/system';
 import { FieldBaseProps } from '../FieldBase/FieldBase';
 import { HelpText } from '../HelpText/HelpText';
 import { Label } from '../Label/Label';
@@ -282,7 +287,12 @@ export const Multiselect = ({
       ]}
     >
       <div
-        className={cn(classNames.field, 'group/field', `w-${width}`)}
+        className={cn(
+          classNames.field,
+          'group/field',
+          width ? 'w-(--field-width) max-w-full min-w-0' : 'w-full'
+        )}
+        style={width ? createWidthVar('field-width', `${width}`) : undefined}
         data-required={props.required}
         data-invalid={error}
         data-readonly={readOnly}
--- a/packages/components/src/Multiselect/Multiselect.tsx
+++ b/packages/components/src/Multiselect/Multiselect.tsx
@@ -19,12 +19,7 @@ import Select, {
 import { useField } from '@react-aria/label';
 import { useId } from '@react-aria/utils';
 import type { WidthProp } from '@marigold/system';
-import {
-  ComponentClassNames,
-  cn,
-  createWidthVar,
-  useClassNames,
-} from '@marigold/system';
+import { ComponentClassNames, cn, useClassNames } from '@marigold/system';
 import { FieldBaseProps } from '../FieldBase/FieldBase';
 import { HelpText } from '../HelpText/HelpText';
 import { Label } from '../Label/Label';
@@ -287,12 +282,7 @@ export const Multiselect = ({
       ]}
     >
       <div
-        className={cn(
-          classNames.field,
-          'group/field',
-          width ? 'w-(--field-width) max-w-full min-w-0' : 'w-full'
-        )}
-        style={width ? createWidthVar('field-width', `${width}`) : undefined}
+        className={cn(classNames.field, 'group/field', `w-${width}`)}
         data-required={props.required}
         data-invalid={error}
         data-readonly={readOnly}
--- a/packages/components/src/Radio/Radio.tsx
+++ b/packages/components/src/Radio/Radio.tsx
@@ -6,7 +6,7 @@ import {
 } from 'react';
 import type RAC from 'react-aria-components';
 import { RadioButton, RadioField } from 'react-aria-components';
-import { cn, useClassNames } from '@marigold/system';
+import { cn, createWidthVar, useClassNames } from '@marigold/system';
 import { useRadioGroupContext } from './Context';
 import { RadioGroup } from './RadioGroup';
 
@@ -62,7 +62,7 @@ const _Radio = forwardRef<HTMLLabelElement, RadioProps>(
     },
     ref
   ) => {
-    const { variant, size, width: groupWidth } = useRadioGroupContext();
+    const { variant, size } = useRadioGroupContext();
 
     const classNames = useClassNames({
       component: 'Radio',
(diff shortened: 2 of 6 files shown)


packages/components/src/Multiselect/Multiselect.tsx
```diff
@@ -282,7 +287,12 @@ export const Multiselect = ({
       ]}
     >
       <div
-        className={cn(classNames.field, 'group/field', `w-${width}`)}
+        className={cn(
```
sebald: 🔴 Please **drop the Multiselect changes** from this PR. `Multiselect` is `@deprecated` on `main` ("Use `TagField` instead. Will be removed in a future major version") and has **already been deleted on `beta-release`** (commit `f6293199d`, DST-1283 / #5311).

Two consequences:
- Low value — it's a cosmetic fix on a component being removed.
- Modifying this file on `main` while `beta-release` deleted it will produce a **modify/delete merge conflict** on the next `main`→`beta-release` merge, which someone has to resolve by re-confirming the deletion.

Suggest making this PR Radio-only. (The change itself is fine in isolation and even fixes a latent `w-undefined` class — but that doesn't outweigh the above.)

packages/components/src/Radio/Radio.tsx
```diff
@@ -62,17 +62,24 @@ const _Radio = forwardRef<HTMLLabelElement, RadioProps>(
...
     });
 
+    // `groupWidth` is already applied to the group's FieldBase container, so an
+    // individual Radio only needs to compute its own CSS variable when it has
+    // its own `width` (i.e. used standalone, outside a Radio.Group). Reusing
+    // `groupWidth` here would resize against an already-resized container.
     return (
       <RadioField
-        className={cn(width || groupWidth || 'w-full')}
+        className={cn(
```
sebald, line 79: ✅ This is the right fix — the old `cn(width || groupWidth || 'w-full')` used the raw token as a class name, so it never generated CSS. Switching to the static `w-(--field-width)` + `createWidthVar` pattern matches `FieldBase`/`TextField`, and dropping per-item `groupWidth` correctly avoids resizing against the already-sized group container.

⚠️ But `Radio.test.tsx` is **not** updated in this PR, despite the description claiming "updated assertions cover the new class/style output." Please add coverage: standalone `<Radio width="1/2">` → `w-(--field-width)` + `--field-width` style var, and a `Radio` inside a sized `Radio.Group` → no re-applied width.

sebald (changes requested):
## Automated Code Review

The underlying bug is real and the **Radio** fix is correct — but the PR should be reshaped before landing.

### The Radio fix is good; the Multiselect fix isn't justifiable

| Component | Status | Verdict |
|-----------|--------|---------|
| **Radio** | Not deprecated. Same width bug on both `main` and `beta-release` — `cn(width \|\| groupWidth \|\| 'w-full')` uses the raw token (e.g. `"1/2"`) as a class name → invalid → no effect. | ✅ Needed & correctly implemented (matches the `FieldBase` `createWidthVar` + `w-(--field-width)` pattern; dropping per-item `groupWidth` correctly avoids double-shrink against the already-sized `FieldBase` container). |
| **Multiselect** | `@deprecated` on `main` ("Use `TagField` instead. Will be removed in a future major version") and **already deleted on `beta-release`** (commit `f6293199d`, DST-1283 / #5311). | ⚠️ Low value + will cause a **modify/delete merge conflict** on the next `main`→`beta-release` merge. |

### 🔴 Critical

**1. Make this PR Radio-only.** Fixing a cosmetic `width` bug on a component that's deprecated in stable and already removed in the next major is throwaway work, and modifying `Multiselect.tsx` here guarantees a modify/delete conflict when `main` merges into `beta-release`. If there's a concrete reason to patch it for current-stable users, please state it explicitly.

**2. No changeset.** This is a user-facing fix (`width` on `Radio` genuinely didn't render). Add one via `pnpm changeset` (`patch`).

### 🟡 Should fix

**3. Radio has no test coverage for the fix.** The diff touches `Radio.tsx` but not `Radio.test.tsx`, yet the PR body claims "updated assertions cover the new class/style output" and the test instructions reference `Radio.test.tsx`. Please add a test: standalone `<Radio width="1/2">` emits `w-(--field-width)` + the `--field-width` style var, and a `Radio` inside a sized `Radio.Group` does **not** re-apply width.

**4. Visual regression not run.** This changes actual rendered width (real visual delta), so Chromatic is warranted here — unlike a pure-logic change.

### 💡 Suggestions

**5.** Title/branch lack the `DST-XXXX` scope required by the repo's Conventional-Commits convention.

---

packages/components/src/Radio/Radio.tsx
zigzagdev: @sebald 

[fixed it ](https://github.com/marigold-ui/marigold/pull/5659/commits/8ec61ac3210a6958924efb81a83d922bb9497299)

packages/components/src/Multiselect/Multiselect.tsx
zigzagdev: @sebald 

[fixed it](https://github.com/marigold-ui/marigold/pull/5659/commits/41a586becdb25514f981f2f2e1289c519e5eb412). Multiselect reverted to main, this PR is now Radio-only.

sebald:
Thanks a lot for this, @zigzagdev — this was a genuinely good catch, and the review iteration (making it Radio-only after we spotted the deprecated-`Multiselect` conflict, and adding the width test coverage) landed it in solid shape. 🙏

We're going to close this PR, but **your fix isn't going away** — we're redirecting it to target the `beta-release` branch instead, where it will actually ship. Two reasons:

1. **`beta-release` is the active integration line** for the upcoming release, and it's where `Radio` currently lives in its refactored form (React 19 ref-as-prop + deep subpath imports from DST-1432). We verified the exact same bug still reproduces there — `<Radio width="1/2">` emits a literal `1/2` class that Tailwind never generates — so the fix is needed on that branch.
2. Because of that refactor, your commits don't cherry-pick cleanly onto `beta-release`, so we hand-applied the **identical logical change** (same `w-(--field-width)` + `createWidthVar` pattern, same story, same test assertions) adapted to the new component shape.

That work is now up as **#5674**, which supersedes this PR and credits you directly. Please do keep them coming — this was exactly the kind of fix we want. 🌼

Closing in favor of #5674.

zigzagdev:
@sebald 

Thank you so much for the kind words and for taking a lot of time to redirect the fix to the right branch.

I'm very appreciated that you hand-applied the change and credited me directly. 
Looking forward to contributing more! 🌼

PR #5662, opened 2026-07-20
fix(DST-1642): inset Slider track so the thumb isn't clipped at the ends

sebald:
# Description

The `Slider` thumb is centered on its track position, so at the **min/max values it overhangs the track ends by half its width** (thumb is `size-5` → 10px each side). Inside a scroll container such as `Drawer.Content` — whose `overflow-y: auto` promotes `overflow-x` to a clipping value per the CSS overflow spec — that overhang gets sliced off, so the end thumbs render as flat half-circles.

It is currently intermittent (the drawer's `px-6` padding usually leaves enough clearance), but becomes **guaranteed** once `Drawer.Content` gains a `bleed` prop with no horizontal padding (DST-1635 / #5649).

**Fix:** inset the track by half the thumb width (`mx-2.5`) and let the grid item stretch to the column minus the margin (`w-auto`). Because the margin is applied equally on **both** sides, the start and end positions get the same 10px of room and cannot diverge — the thumb always stays within the Slider's own box, so no ancestor scroll container can clip it, independent of any padding. Current vs. new: the visible track bar is 10px shorter on each side (standard slider behavior); fill and thumb geometry are unchanged.

Closes DST-1642

## Screenshots / Preview

Verified in Storybook (`Components/Drawer/WithForms`) with both range thumbs pushed to the extremes:

| Before | After |
| ------ | ----- |
| End thumbs sliced to half-circles at the drawer edges | Full round thumbs, flush but never clipped |

Measured with Playwright at **both** the start (min / left) **and** end (max / right) positions:

- **Before** — start thumb clipped 10px on the left, end thumb clipped 10px on the right (`scrollWidth 394 > clientWidth 384`).
- **After** — `scrollWidth == clientWidth`, and clipping is `0/0` on **both** thumbs (`clipL` and `clipR`) in both the padded and the zero-horizontal-padding (bleed) cases.

Covered across single-thumb (same thumb driven to min **and** max), range (start thumb → min, end thumb → max), and a narrow 160px fixed width. The fill edges align exactly to both thumb centers.

## Test Instructions

1. `pnpm sb`, open `Components/Slider/MultipleThumbs`, drag the **start** thumb fully left and the **end** thumb fully right — both stay fully round.
2. Open `Components/Drawer/WithForms`, open the drawer, set the price range to `€10 – €140` (both extremes) — neither the left nor the right thumb is clipped by the drawer edge.
3. `pnpm test:unit --run packages/components/src/Slider/Slider.test.tsx packages/components/src/Drawer/Drawer.test.tsx` and `pnpm test:sb --run packages/components/src/Slider/Slider.stories.tsx packages/components/src/Drawer/Drawer.stories.tsx` — all green.

## Breaking Changes

No

## Checklist

- [x] Accessibility reviewed against ARIA APG 

---

Note: this is a purely visual change, so a Chromatic VRT run is the real regression guard. I have **not** triggered VRT — ping me and I'll dispatch it (or add a dedicated extreme-value story) if you'd like.


sarahgm:
I could not reproduce this problem in beta-release and main. Is this really an issue?

sebald:
@sarahgm you're right, and the reason it doesn't reproduce on `beta-release`/`main` is worth writing down because it also confirms the fix is needed.

**Short version:** the bug is real, but the *default* `Drawer.Content` hides it. `Drawer.Content` uses `ui-panel-content` = `overflow-y-auto px-6 py-4`. The `overflow-y: auto` promotes `overflow-x` to a clipping value, and the `size-5` thumb overhangs each track end by exactly **10px**. But `px-6` = 24px of horizontal padding, so the 10px overhang stays inside the padding box and nothing is clipped. That's why you can't see it in a normal Drawer. It only becomes visible once the horizontal padding drops below ~10px, i.e. the `bleed` prop coming in #5649 (`px-0`).

**Verified empirically.** I built (`pnpm build`) and measured a Slider inside a 320px `overflow-y: auto` box in Firefox (via the storybook-test runner) on both `beta-release` and `feat/DST-1609-two-level-sidebar` — identical results on both branches:

| Case (320px scroll box) | scrollWidth vs clientWidth | thumb clipped |
| --- | --- | --- |
| Default padding (`px-6`), min value | 320 = 320 | none |
| Default padding (`px-6`), max value | 320 = 320 | none |
| Zero padding (`px-0`), min value | 320 = 320 | **10px off the left** |
| Zero padding (`px-0`), max value | **330 > 320** | **10px off the right** |

So: with `px-6` there's no clipping (matches what you saw), and with `px-0` the thumb is clipped by 10px at each end.

**One subtlety about the `scrollWidth 394 > 384` number in the PR description:** that signal only fires for the **max**-value thumb overhanging the *right* edge. A thumb overhanging the *left* edge (min value) is clipped too, but it does **not** grow `scrollWidth` — an LTR scroll box has no leftward scroll region, it just clips. So the right way to detect the min-value case is geometric (thumb rect vs the container's padding box), not `scrollWidth`.

**Bottom line:** there's no behavioral difference between the branches here, and the default Drawer genuinely doesn't clip today. The fix (`mx-2.5 w-auto`) is worth keeping because it makes the Slider immune to any zero/low horizontal padding scroll container — which is exactly the situation `bleed` (#5649) introduces.

Happy to push the small repro story somewhere if it's useful for a regression test.

sarahgm (approved):

sebald:
@sarahgm I've resolved the conflicts and rebased onto the current `beta-release`. The only conflict was in `Slider.tsx` (the branch predated the `forwardRef` → ref-as-prop refactor); it's now the modern structure with just the `mx-2.5 w-auto` inset applied, so the PR diff is back to `Slider.tsx` + the changeset only.

I also re-verified after the merge, building each branch and measuring a Slider inside a 320px `overflow-y: auto` box in Firefox:

- `beta-release` (no fix), zero-padding container: thumb clipped 10px at min (left) and 10px at max (right, `scrollWidth 330 > clientWidth 320`).
- This branch (with fix): `0 / 0` clipping and `scrollWidth == clientWidth` in every case — padded and zero-padding.

So the bug is confirmed on the base and gone with the fix. Conflicts cleared (state is green now) and ready for another look.

PR #5664, opened 2026-07-21
feat(DST-1643): add a fullscreen size to Dialog

sarahgm:
# Description

Adds a `fullscreen` size to `<Dialog>`. `<Dialog size="fullscreen">` fills the viewport minus a small margin at every breakpoint, giving content-heavy picks room for search, filters, and a long scrollable list while the title and actions stay fixed. The existing `xsmall`/`small`/`medium`/`large` sizes are unchanged.

Implementation: the Modal theme gains a `fullscreen` size (parametrized `--dialog-max-h` + a dedicated `--dialog-spacing-y` so it can exceed the 80vh cap), and the Dialog surface becomes a clipped grid for fullscreen so only the content region scrolls, never the dialog itself.

Discovered while implementing the Pick pattern (DST-1532). Targets `beta-release`; Chromatic/VRT is not enforced on beta PRs. One Chromatic snapshot captures the open fullscreen dialog (the `fills the viewport` test); the other interaction test stays snapshot-disabled. Dialog docs + the Pick "Sizing a heavy pick" callout are a follow-up.

Closes DST-1643

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Open Storybook → `Components/Dialog/Fullscreen`.
2. Click "Select venues" — the dialog fills the viewport with a small margin.
3. Scroll the table: only the list scrolls; the title, search/filter, staged tags, and footer stay fixed, and the dialog surface itself does not scroll.
4. Toggle the `size` control to `small`/`medium`/`large` and confirm those still cap and hug their content as before.

## Breaking Changes

No

## Checklist

### DIFF
--- a/packages/components/src/Dialog/Dialog.stories.tsx
+++ b/packages/components/src/Dialog/Dialog.stories.tsx
@@ -397,6 +397,7 @@ const FullscreenPick = ({ size, ...args }: FullscreenPickProps) => {
               </Tag.Group>
             )}
 
+            {/* Plain scroll container with no accessible role, so tests assert its scroll via this testid rather than getByRole. */}
             <div
               className="min-h-0 flex-1 overflow-auto"
               data-testid="venue-list"
--- a/packages/components/src/Dialog/Dialog.stories.tsx
+++ b/packages/components/src/Dialog/Dialog.stories.tsx
@@ -1,11 +1,19 @@
+import { useMemo, useState } from 'react';
 import { expect, waitFor } from 'storybook/test';
 import preview from '.storybook/preview';
+import type { Key, Selection } from '@react-types/shared';
 import { Copy, Download, Pencil, Trash2 } from '@marigold/icons';
 import { Button } from '../Button/Button';
 import { ButtonGroup } from '../ButtonGroup/ButtonGroup';
 import { Description } from '../Description/Description';
+import { EmptyState } from '../EmptyState/EmptyState';
 import { Form } from '../Form/Form';
+import { Inline } from '../Inline/Inline';
 import { ActionMenu } from '../Menu/ActionMenu';
+import { SearchField } from '../SearchField/SearchField';
+import { Select } from '../Select/Select';
+import { Table } from '../Table/Table';
+import { Tag } from '../TagGroup/Tag';
 import { Text } from '../Text/Text';
 import { TextField } from '../TextField/TextField';
 import { Title } from '../Title/Title';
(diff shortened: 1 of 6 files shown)


packages/components/src/Dialog/Dialog.tsx
```diff
@@ -89,9 +90,16 @@ const InnerDialog = ({
...
+        // Fullscreen fills the Modal box, other sizes hug their content.
+        isFullscreen
+          ? 'size-full max-h-full max-w-full'
+          : 'max-h-[80vh] max-w-[90vw]',
         "grid [grid-template-areas:'title'_'content'_'actions']",
-        classNames.container
+        classNames.container,
+        // Grid with pinned rows and a clipped surface, so only the inner content scrolls, not the dialog.
+        isFullscreen &&
+          'grid grid-rows-[auto_minmax(0,1fr)_auto] overflow-x-hidden overflow-y-hidden'
```
jim761: 💡 `overflow-x-hidden overflow-y-hidden` collapses to `overflow-hidden`.

```suggestion
          'grid grid-rows-[auto_minmax(0,1fr)_auto] overflow-hidden'
```

Also worth a one-line note here: the `grid` in this conditional is load-bearing — `classNames.container` sets `flex`, and this trailing `grid` relies on twMerge's last-wins to flip fullscreen back to grid layout. It's correct (tests confirm it), just subtle for the next reader.

packages/components/src/Dialog/Dialog.stories.tsx
```diff
@@ -124,6 +132,370 @@ export const WithFormValidation = meta.story({
...
+                  <Tag key={venue.id} id={venue.id}>
+                    {venue.name}
+                  </Tag>
+                ))}
+              </Tag.Group>
+            )}
+
+            <div
+              className="min-h-0 flex-1 overflow-auto"
+              data-testid="venue-list"
```
jim761, line 403: ⚠️ CLAUDE.md discourages `getByTestId` in favor of accessible queries (used here and at L487–488). This is a defensible exception — scroll containers have no accessible role — but a short comment explaining *why* the testid escape hatch is needed would keep it from being cargo-culted into tests that could use `getByRole`.

jim761:
Should the docu also be included on when to use fullscreen or is that for later?

sarahgm:
@jim761 good call. I added a short `### Sizes` section to the Dialog docs covering when to reach for `fullscreen` (content-heavy tasks like multi-column pickers or dense tables) versus a smaller boxed size for focused, single-task dialogs. Pushed in 795b8c1a9.

Kept it prose-only rather than a full sizes matrix, since the other sizes aren't documented individually yet either. Happy to expand into a proper sizes overview as a follow-up if you think it's worth it.

jim761 (approved):

PR #5665, opened 2026-07-21
fix(DST-1645): align preset quick-selection rows with the tray nav row

sarahgm:
# Description

The Calendar / DatePicker / DateRangePicker **"Quick selection" preset tray** reused the shared `ListBox` list style, whose `p-1` focus-ring gutter insets each row by 4px per side. The nav row ("Back" / "Quick selection") sits *outside* the listbox at full width, so on small screens the preset rows (and the selected-row fill) rendered narrower than the nav row and didn't line up with it or the tray edges.

This negates the gutter on small screens with `-mx-1` (plus `w-auto`, so the list widens instead of staying pinned to `w-full`): rows now span the tray edge-to-edge and align with the nav row, while the `p-1` gutter stays in place so each row's focus outline still isn't clipped by the list's `overflow-y-auto`. Same cancel trick as Accordion's `panel` (`px-1 -mx-1`). The desktop rail keeps its inset, and the shared `ListBox` style is untouched (Select / ComboBox / Autocomplete / Menu unaffected).

Found during the Chromatic VRT review of `beta-release` (build 151, under DST-1584).

Closes DST-1645

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm sb` and pick the **Extra Small Screen (320px)** viewport (or resize below the `sm` breakpoint).
2. Open a `DatePicker` (or `DateRangePicker`) story that has presets, open the picker, and tap **"Quick selection"** to switch the tray to the preset view.
3. Confirm the preset rows — and the selected-row fill — align with the full-width **"Back"** nav row and the tray edges (no 4px inset mismatch).
4. Tab through the preset rows and confirm each focus outline renders fully, unclipped.
5. On desktop (≥ `sm`), confirm the presets rail is unchanged (rows keep their inset inside the `bg-muted` rail).

## Breaking Changes

No

## Checklist

### DIFF
--- a/.changeset/dst-1645-preset-tray-alignment.md
+++ b/.changeset/dst-1645-preset-tray-alignment.md
@@ -1,0 +1,5 @@
+---
+'@marigold/components': patch
+---
+
+fix(DST-1645): align preset quick-selection rows with the tray nav row on small screens. The preset listbox rows reuse the shared `ListBox` `p-1` focus-ring gutter, which insets them narrower than the full-width nav row. On small screens the list now negates that gutter (`-mx-1`) so rows span the tray edge-to-edge and line up with the nav row, while the gutter still keeps each row's focus outline from being clipped by the list's overflow.
--- a/packages/components/src/Calendar/CalendarPresets.tsx
+++ b/packages/components/src/Calendar/CalendarPresets.tsx
@@ -169,7 +169,11 @@ const PresetListBox = <T,>({
         disabledKeys={disabledKeys}
         onSelectionChange={handleSelectionChange}
         autoFocus={autoFocus}
-        className={listBoxClassNames.list}
+        // Small screens: negate the list's `p-1` (`-mx-1`) so rows span the tray
+        // edge-to-edge and align with the full-width nav row; the gutter stays,
+        // still keeping each row's focus outline from clipping. Desktop rail keeps
+        // the inset.
+        className={cn(listBoxClassNames.list, isSmallScreen && '-mx-1 w-auto')}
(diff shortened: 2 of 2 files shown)


sebald:
image

sebald (approved):
Reviewed the change — LGTM. ✅

**The fix is correct and elegant.** Because the list sets `overflow-y-auto`, `overflow-x` computes to `auto` too, so the `p-1` gutter is genuinely load-bearing: it's the buffer that keeps each row's 3px `outline-offset-0` focus ring from being clipped horizontally. `-mx-1` shifts the whole list out by 4px per side while the padding stays, so rows land flush with the full-width nav row *and* the focus-ring buffer survives. Nicely reasoned.

- `w-auto` is necessary, not incidental — with `w-full` retained, the negative margin would just shift the fixed-width box left and misalign the other edge; dropping to `w-auto` lets it grow to `100% + 8px`. Good that the description calls this out.
- Gating on `isSmallScreen` is right: the full-width nav row / in-place preset view only exist below `sm`; the desktop rail (inside `bg-muted`) keeps its inset and is untouched.
- Shared `ListBox` style is unchanged, so Select / ComboBox / Autocomplete / Menu are unaffected.

**Non-blocking notes:**
- No unit/story test added. Defensible for a pure alignment tweak; the real guard here is VRT.
- This was found *during* Chromatic VRT review, and it's a visual change — the Chromatic baseline should be updated (checklist item still unticked). Normal post-merge flow, just flagging.

PR #5668, opened 2026-07-22
feat(DST-1492): add usePageFocus to move focus to page heading on route change

OsamaAbdellateef:
# Description

Adds a new `usePageFocus` hook that moves focus to the page `<h1>` on client-side route changes.

**Current behavior:** after SPA navigation, focus lingers on the clicked link or falls back to `<body>`, so screen-reader and keyboard users get no signal that the screen changed.

**New behavior:** given the current route key (typically the pathname), `usePageFocus` makes the page `<h1>` programmatically focusable (`tabIndex={-1}`) and focuses it on each change. The first render is skipped so the initial load never steals focus, and it is a no-op on a page with no `<h1>` (e.g. an `aria-label`-only `<Page>`). `<Page>` stays router-agnostic — the route signal comes from the app router — so the hook is called from a component that persists across navigations (layout/shell level).

Also exports `usePageFocus` and `UsePageFocusOptions` from `@marigold/components`.

Closes DST-1492

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Open Storybook → `Components/Page` and exercise the new focus stories.
2. Trigger a simulated route change and confirm focus lands on the page `<h1>` (visible focus ring on the heading).
3. Confirm the initial render does **not** steal focus.
4. Confirm a `<Page>` with no `<h1>` (aria-label only) is a no-op — focus is not moved.
5. Verify types/tests: `pnpm typecheck:only` and `pnpm test:unit -- usePageFocus`.

## Breaking Changes

No — purely additive (new hook + exports).

## Checklist


packages/components/src/Page/Page.stories.tsx
```diff
@@ -357,3 +359,120 @@ export const ExternalLabel = meta.story({
...
+/**
+ * `usePageFocus` moves focus to the page `<h1>` on a route change, the standard
+ * SPA technique for announcing navigation to screen-reader and keyboard users.
+ * The router is the source of the route signal — pass the current pathname to
+ * the hook from a component that persists across navigations (here, a small
+ * harness swaps the "route" via the nav buttons). The initial mount is skipped,
+ * so first paint never steals focus.
+ */
+export const FocusOnRouteChange = meta.story({
+  tags: ['component-test'],
```
jim761, line 419: 🟡 This is a behavioral (focus) story with no static visual diff, and its `play` clicks *before* Chromatic captures — so a snapshot would baseline a flaky, focus-ringed heading in an arbitrary post-click state. Opt it out of snapshots, matching the repo convention for interaction-only stories (TagGroup, Tooltip, Drawer, Form). Do this before triggering `/run-chromatic`.

```suggestion
  tags: ['component-test'],
  parameters: { chromatic: { disableSnapshot: true } },
```

packages/components/src/Page/Page.stories.tsx
```diff
@@ -357,3 +359,120 @@ export const ExternalLabel = meta.story({
...
+            <Text>Route: {route}</Text>
+          </Panel.Content>
+        </Panel>
+      </Page>
+    </>
+  );
+};
+
+export const FocusWithoutHeading = meta.story({
+  tags: ['component-test'],
```
jim761, line 468: 🟡 Same as `FocusOnRouteChange` — exclude this behavioral story from Chromatic snapshots.

```suggestion
  tags: ['component-test'],
  parameters: { chromatic: { disableSnapshot: true } },
```

packages/components/src/Page/usePageFocus.ts
```diff
@@ -0,0 +1,81 @@
...
+ *   </Page>
+ * </RouterProvider>
+ * ```
+ */
+export const usePageFocus = (
+  routeKey: string,
+  { enabled = true }: UsePageFocusOptions = {}
+) => {
+  const { titleId } = usePageContext();
+  const isInitialMountRef = useRef(true);
```
jim761: 🟡 The `isInitialMountRef` boolean isn't resilient to React StrictMode. Dev StrictMode runs setup → cleanup → setup on mount **while preserving refs**, so the second setup sees `current === false` and focuses the `<h1>` on initial load — defeating the "don't steal focus on first paint" guarantee. Production is unaffected (no double-invoke), and the unit tests miss it because RTL `render` doesn't wrap in StrictMode.

Track the previous `routeKey` instead — resilient to re-runs, and it also removes the spurious re-focus when `enabled` toggles on an unchanged route:

```ts
const previousRouteKeyRef = useRef(routeKey);
useEffect(() => {
  const changed = previousRouteKeyRef.current !== routeKey;
  previousRouteKeyRef.current = routeKey;
  if (!changed || !enabled) return;
  const heading = document.getElementById(titleId);
  if (!heading) return;
  heading.tabIndex = -1;
  heading.focus();
}, [routeKey, titleId, enabled]);
```

packages/components/src/Page/usePageFocus.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  useEffect(() => {
+    // Skip the first run so the initial page load is left untouched; only a
+    // subsequent route change should move focus.
+    if (isInitialMountRef.current) {
+      isInitialMountRef.current = false;
+      return;
+    }
+
+    if (!enabled) {
```
jim761: 🟡 `enabled` is untested and can move focus without a route change. No story/test exercises `enabled: false`, leaving this new-file branch uncovered against the 85% branch gate. And because `enabled` is an effect dep, toggling it `false → true` on an unchanged route re-focuses the heading. The previous-routeKey fix (see the comment above) removes the spurious re-focus; still add an `enabled: false` test, or drop the option — note it's a public export (`UsePageFocusOptions`), so removing it later would be breaking.

packages/components/src/Page/usePageFocus.test.tsx
```diff
@@ -0,0 +1,54 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { usePageFocus } from './usePageFocus';
+import { FocusOnRouteChange, FocusWithoutHeading } from './Page.stories';
+
+describe('usePageFocus', () => {
```
jim761, line 7: 💡 Three of the four tests here re-assert what the `FocusOnRouteChange` / `FocusWithoutHeading` `play` functions already cover in-browser; only "throws outside a Page" is unique. Consider trimming to the unique case — the jsdom-speed coverage is a fair reason to keep some, so this is optional.

jim761 (changes requested):
## Automated Code Review

### PR Review: #5668 — feat(DST-1492): add `usePageFocus` to move focus to page heading on route change

#### Overview
| Field | Value |
|-------|-------|
| Author | @OsamaAbdellateef |
| Branch | `dst-1492_manage-focus-on-page` → `beta-release` |
| Files Changed | 6 (+326 / -1) |
| CI | ✅ Green (Typecheck, Lint, Unit ×4, Storybook ×4, Coverage) |
| Visual Regression | Change has no static visual diff; the two new behavioral stories should be opted out of snapshots before any Chromatic run (see inline) |

#### Linked Ticket
**DST-1492** — ✨ Page: move focus to the page heading on route change · *In Review*

Scope match is excellent: exported `usePageFocus`, skip-on-initial-mount, no-op when no `<h1>`, tests for all three cases, `RouterProvider` docs guidance, and a changeset. One benign deviation from the ticket's suggested solution — it targets the heading via `document.getElementById(titleId)` rather than the proposed `titleSlotRef` (a reasonable simplification).

#### Checklist
- **TypeScript** ✅ no `any`; value/type imports correct; clean public `UsePageFocusOptions`.
- **React / Performance** ✅ effect responds to an external router signal (correct `useEffect` use); primitive deps `[routeKey, titleId, enabled]`; story helpers and `ROUTES` hoisted to module scope; `useRef` for the transient mount flag.
- **Testing** ✅ Vitest, `userEvent` (not `fireEvent`), `getByRole` (no `getByTestId`), `component-test` tags, imports stories instead of fixtures.
- **Accessibility** ✅ correct APG route-change pattern — `<h1>` made programmatically focusable (`tabIndex={-1}`) and focused; no-op path preserves prior focus.

#### Issues Found

**🟡 Warnings (Should Fix)**

1. **Opt the two new behavioral stories out of Chromatic** (`chromatic: { disableSnapshot: true }`). `FocusOnRouteChange` / `FocusWithoutHeading` have no static visual diff and their `play` clicks before capture, so a snapshot baselines a flaky focus-ringed heading. Matches the repo convention for interaction-only stories (TagGroup, Tooltip, Drawer, Form). — `Page.stories.tsx`
2. **`isInitialMountRef` guard isn't resilient to React StrictMode.** Dev StrictMode runs setup → cleanup → setup on mount while preserving refs, so the second setup sees `current === false` and focuses the `<h1>` on initial load — defeating the "don't steal focus on first paint" guarantee. Prod is unaffected; tests miss it because RTL `render` doesn't wrap in StrictMode. Track the previous `routeKey` instead. — `usePageFocus.ts:55`
3. **`enabled` is untested and can move focus without a route change.** No test/story exercises `enabled: false` (uncovered branch vs. the 85% gate), and because it's an effect dep, toggling `false → true` on an unchanged route re-focuses the heading. The previous-routeKey fix removes the spurious re-focus; still add a test, or drop the option (public export → removing later is breaking). — `usePageFocus.ts:65`

**💡 Suggestions (Nice to Have)**

4. **`usePageFocus.test.tsx` largely duplicates the story `play` functions** — only "throws outside a Page" is unique. Consider trimming (jsdom speed is a fair reason to keep some). — `usePageFocus.test.tsx:6`

#### Recommendation
**Request Changes** — no production-facing runtime bugs, and the design/docs/tests are strong. Two items before merge: opt the behavioral stories out of Chromatic (#1) and harden the initial-mount guard against StrictMode (#2). #3 and #4 are quick follow-ups.

---

packages/components/src/Page/usePageFocus.ts
```diff
@@ -0,0 +1,81 @@
...
+  routeKey: string,
+  { enabled = true }: UsePageFocusOptions = {}
+) => {
+  const { titleId } = usePageContext();
+  const isInitialMountRef = useRef(true);
+
+  useEffect(() => {
+    // Skip the first run so the initial page load is left untouched; only a
+    // subsequent route change should move focus.
+    if (isInitialMountRef.current) {
```
aromko: 🔴 **First-render guard breaks under StrictMode and when `enabled` flips**

Two ways this effect moves focus without a route change:

1. **React StrictMode (dev):** effects run setup → cleanup → setup on mount, and refs persist across the simulated remount. The second setup sees `isInitialMountRef.current === false` and focuses the heading on the initial load, which is exactly what the skip is documented to prevent. StrictMode is the default in Next.js, so consumers will see focus stolen on first paint in dev.
2. **`enabled` is in the dep array:** toggling it from `false` to `true` re-runs the effect and moves focus with no navigation involved.

Tracking the last-seen key fixes both, because the effect then only acts when `routeKey` actually changed:

```ts
const lastRouteKeyRef = useRef(routeKey);

useEffect(() => {
  if (lastRouteKeyRef.current === routeKey) {
    return;
  }
  lastRouteKeyRef.current = routeKey;

  if (!enabled) {
    return;
  }
  // … focus logic unchanged
}, [routeKey, titleId, enabled]);
```

Consider adding a `<StrictMode>` unit test to pin the "no focus theft on first mount" guarantee (the current tests pass because they render without StrictMode).

docs/content/components/layout/page/index.mdx
```diff
@@ -119,6 +119,49 @@ Put `<Panel>`s directly inside `<Page>` for the common case. Reach for the optio
...
 
+### Focus on route change
+
+On client-side navigation the browser does not reset focus the way a full page load does: it lingers on the link that was clicked, or falls back to `<body>`. Screen-reader and keyboard users then get no signal that the screen changed and are stranded at the top of the tab order. The standard fix is to move focus to the new screen's heading, so the page is announced and the next `Tab` continues from the top of the content.
+
+`usePageFocus` does this. Pass it the current route key (usually the pathname); on each change it makes the page `<h1>` programmatically focusable (`tabIndex={-1}`) and focuses it. The **first render is skipped**, so the initial load is left untouched, and it is a **no-op on a page with no `<h1>`** (an `aria-label`-only `<Page>`).
+
+`<Page>` is router-agnostic and cannot observe navigation on its own, so the route signal comes from your app router — the same [`RouterProvider`](/components/application/routerprovider) you already wire up for links. Call the hook from a component that persists across navigations (the layout / shell level), since the skip-first-render guard is per mount:
```
aromko: ⚠️ **Docs style: avoid em dashes**

Our docs convention is to avoid em dashes; rephrase with commas, colons, or parentheses. In this section: "your app router — the same `RouterProvider`" (this line) and "on every navigation — otherwise every route reads as a first render" (closing paragraph). The changeset text has one too ("router-agnostic — the route signal").

packages/components/src/Page/usePageFocus.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export interface UsePageFocusOptions {
+  /**
+   * When `false`, the hook is inert: route changes do not move focus. Lets a
+   * consumer keep the hook mounted while gating the behaviour (e.g. respecting a
+   * "reduce motion / reduce focus jumps" preference) without conditionally
+   * calling the hook.
+   * @default true
+   */
+  enabled?: boolean;
```
aromko: 💡 **`enabled` is speculative API**

No consumer needs this yet and it is outside the ticket's scope. Since it is exported public API, adding it later is cheap while removing it is breaking. Consider shipping without options and adding `enabled` once a real use case shows up.

packages/components/src/Page/Page.stories.tsx
```diff
@@ -357,3 +359,120 @@ export const ExternalLabel = meta.story({
...
+  '/billing': 'Billing',
+  '/team': 'Team members',
+};
+
+const RouteFocusHarness = () => {
+  const [route, setRoute] = useState('/billing');
+
+  return (
+    <>
+      <nav aria-label="Demo navigation" style={{ display: 'flex', gap: 8 }}>
```
aromko: 💡 **Use `<Inline>` instead of inline styles**

`<Inline space={2}>` replaces `style={{ display: 'flex', gap: 8 }}` and dogfoods the system (the repo convention is to avoid inline styles). Same for the nav in `NoHeadingFocusHarness`.

aromko (changes requested):
## Automated Code Review

### Summary

Solid, well-scoped implementation of DST-1492: the hook uses Page's existing `titleId` plumbing, skips the first paint, no-ops without an `<h1>`, and ships tests, stories, docs, and a changeset that all match the ticket's scope. One blocker before merge, detailed as an inline comment:

1. **`usePageFocus.ts` first-render guard** breaks under React StrictMode (focus is stolen on initial load in dev, the exact thing the hook documents it never does) and fires when `enabled` flips without a navigation. Tracking the last-seen `routeKey` in a ref fixes both.

Non-blocking: em dashes in the new docs section (docs style: rephrase with commas/colons), the speculative `enabled` option, and inline styles in the story harness navs where `<Inline space={2}>` would do.

### Checklist

- TypeScript: no `any`, strict-safe, named exports ✓
- Testing: Vitest, `userEvent`, accessible queries, stories imported in unit tests, `component-test` tags ✓
- Accessibility: standard SPA route-announcement pattern (`tabIndex={-1}` + focus on the `<h1>`), initial load untouched ✓ (once the StrictMode guard is fixed)
- React performance: primitive effect deps, no inline component definitions ✓
- Docs + changeset present ✓

**Recommendation: Request Changes** (first-render guard bug).

---

packages/components/src/Page/Page.stories.tsx
OsamaAbdellateef: Done in 1ddff388a: added `parameters: { chromatic: { disableSnapshot: true } }` to `FocusOnRouteChange`.

packages/components/src/Page/Page.stories.tsx
OsamaAbdellateef: Done in 1ddff388a: same opt-out added to `FocusWithoutHeading`.

packages/components/src/Page/usePageFocus.ts
OsamaAbdellateef: Fixed in 1ddff388a: replaced `isInitialMountRef` with a `previousRouteKeyRef` that tracks the last-seen `routeKey`, so the effect only acts when the key actually changes. Resilient to StrictMode's setup → cleanup → setup remount since it no longer depends on a boolean flipped exactly once.

packages/components/src/Page/usePageFocus.ts
OsamaAbdellateef: Dropped `enabled` in 1ddff388a rather than adding a test for it — aromko flagged the same speculative-API concern below, and with no consumer yet it's cheaper to add back later (with a test) than to carry untested surface now. The effect's dep array is just `[routeKey, titleId]`, so there's no more toggle-without-navigation case either.

packages/components/src/Page/usePageFocus.test.tsx
OsamaAbdellateef: Trimmed in 1ddff388a: removed the three tests already covered by the `FocusOnRouteChange` / `FocusWithoutHeading` story plays, kept "throws when used outside a Page" (the unique case), and added a StrictMode test that pins the initial-mount regression from the thread above — RTL's default render doesn't reproduce that double-invoke, so it wasn't caught by the existing coverage.

packages/components/src/Page/usePageFocus.ts
OsamaAbdellateef: Fixed in 1ddff388a, same approach: a `previousRouteKeyRef` replaces the boolean, so the effect only fires on an actual `routeKey` change. Also added a `<StrictMode>` unit test in `usePageFocus.test.tsx` to pin the "no focus theft on first mount" guarantee, as suggested.

docs/content/components/layout/page/index.mdx
OsamaAbdellateef: Fixed in 1ddff388a — removed both em dashes in this section and the one in the changeset text.

packages/components/src/Page/usePageFocus.ts
OsamaAbdellateef: Agreed — dropped `enabled` and `UsePageFocusOptions` entirely in 1ddff388a. `usePageFocus` now just takes `routeKey`. Happy to add the option back with a real use case and tests if one comes up.

packages/components/src/Page/Page.stories.tsx
OsamaAbdellateef: Done in 1ddff388a: both `nav` elements now wrap their buttons in `<Inline space={2}>` instead of inline `style`.

packages/components/src/Page/usePageFocus.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  useEffect(() => {
+    // Only act once the route key actually changes. Tracking the previous key
+    // (rather than an "is this the first run" flag) keeps the skip resilient to
+    // React StrictMode's dev-only setup → cleanup → setup remount: refs survive
+    // that remount, so a boolean flipped to `false` on the first setup would
+    // read as "not the first run" on the second setup and steal focus on the
+    // initial page load.
+    const changed = previousRouteKeyRef.current !== routeKey;
+    previousRouteKeyRef.current = routeKey;
```
jim761, line 60: ⚠️ **Silent no-op under lazy / Suspense routes.** This effect advances `previousRouteKeyRef.current` to the new key *before* the heading lookup, then bails if `getElementById(titleId)` is `null`. If a route's `<Title>` mounts *after* the effect fires — the common case for `React.lazy` + `<Suspense>` route content inside `<Outlet>` — focus silently never moves: the ref has already advanced, and the effect only re-runs on `[routeKey, titleId]` change, neither of which changes when the heading later appears.

The docs example keeps `<Title>` at the shell level (always mounted), which sidesteps this, but per-route titles are the common case. Suggest a doc caveat: the page `<h1>` must be present synchronously on the route change, so lazy route content needs the title (or the route signal) hoisted above the Suspense boundary.

packages/components/src/Page/Page.stories.tsx
```diff
@@ -357,3 +360,142 @@ export const ExternalLabel = meta.story({
...
+ * The router is the source of the route signal — pass the current pathname to
+ * the hook from a component that persists across navigations (here, a small
+ * harness swaps the "route" via the nav buttons). The initial mount is skipped,
+ * so first paint never steals focus.
+ */
+export const FocusOnRouteChange = meta.story({
+  tags: ['component-test'],
+  parameters: { chromatic: { disableSnapshot: true } },
+  render: () => <RouteFocusHarness />,
+  play: async ({ canvasElement }) => {
```
aromko: Please rewrite as .test - no play functions anymore

packages/components/src/Page/Page.stories.tsx
```diff
@@ -357,3 +360,142 @@ export const ExternalLabel = meta.story({
...
+      </Page>
+    </>
+  );
+};
+
+export const FocusWithoutHeading = meta.story({
+  tags: ['component-test'],
+  parameters: { chromatic: { disableSnapshot: true } },
+  render: () => <NoHeadingFocusHarness />,
+  play: async ({ canvasElement }) => {
```
aromko: Sam here - use .test syntax

packages/components/src/Page/Page.stories.tsx
```diff
@@ -357,3 +360,142 @@ export const ExternalLabel = meta.story({
...
+      name: 'Team members',
+    });
+    await expect(teamHeading).toHaveFocus();
+    // The heading is made programmatically focusable, not tabbable.
+    await expect(teamHeading).toHaveAttribute('tabindex', '-1');
+  },
+});
+
+/**
+ * When the page has no `<Title>` (named instead by an `aria-label`), there is no
```
aromko: ⚠️ This JSDoc block describes `FocusWithoutHeading`, but it sits directly above `const NoHeadingFocusHarness`, so Storybook attaches it to the harness instead of the story. The description won't render on the docs page.

Move the block down so it immediately precedes `export const FocusWithoutHeading`.

packages/components/src/Page/Page.stories.tsx
```diff
@@ -357,3 +360,142 @@ export const ExternalLabel = meta.story({
...
+  render: () => <RouteFocusHarness />,
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+
+    // On first paint the initial heading must not be focused.
+    const billingHeading = canvas.getByRole('heading', {
+      level: 1,
+      name: 'Billing',
+    });
+    await expect(billingHeading).not.toHaveFocus();
```
aromko: ⚠️ This play function is assert → act → assert, which is two AAA blocks in one test. Per our one-Arrange-Act-Assert-per-test rule, split it:

- one story asserting the initial heading is not focused on first paint
- one story clicking through and asserting focus lands on the new heading with `tabindex="-1"`

Same coverage, each failure points at exactly one behaviour.

docs/content/components/layout/page/index.mdx
```diff
@@ -119,6 +119,49 @@ Put `<Panel>`s directly inside `<Page>` for the common case. Reach for the optio
...
 - **Heading outline:** `<Page>` sets the base heading level to `1`, so the page title renders as the `<h1>` and each `<Panel>` below it as an `<h2>` (Panel collapsibles as `<h3>`). The outline a screen-reader user skims to jump between sections falls out of these defaults, with no heading levels to set by hand.
 
+### Focus on route change
+
+On client-side navigation the browser does not reset focus the way a full page load does: it lingers on the link that was clicked, or falls back to `<body>`. Screen-reader and keyboard users then get no signal that the screen changed and are stranded at the top of the tab order. The standard fix is to move focus to the new screen's heading, so the page is announced and the next `Tab` continues from the top of the content.
+
+`usePageFocus` does this. Pass it the current route key (usually the pathname) on each change it makes the page `<h1>` programmatically focusable (`tabIndex={-1}`) and focuses it. The **first render is skipped**, so the initial load is left untouched, and it is a **no-op on a page with no `<h1>`** (an `aria-label`-only `<Page>`).
```
aromko: ⚠️ Two sentences are fused here: "Pass it the current route key (usually the pathname) on each change it makes the page `<h1>` programmatically focusable".

```suggestion
`usePageFocus` does this. Pass it the current route key (usually the pathname). On each change it makes the page `<h1>` programmatically focusable (`tabIndex={-1}`) and focuses it. The **first render is skipped**, so the initial load is left untouched, and it is a **no-op on a page with no `<h1>`** (an `aria-label`-only `<Page>`).
```

docs/content/components/layout/page/index.mdx
```diff
@@ -119,6 +119,49 @@ Put `<Panel>`s directly inside `<Page>` for the common case. Reach for the optio
...
+```tsx
+import { Outlet, useLocation, useNavigate } from 'react-router';
+import {
+  Page,
+  RouterProvider,
+  Title,
+  usePageFocus,
+} from '@marigold/components';
+
+// Renders nothing it just connects the route signal to the page heading.
```
aromko: 💡 Missing punctuation.

```suggestion
// Renders nothing; it just connects the route signal to the page heading.
```

packages/components/src/Page/usePageFocus.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (!changed) {
+      return;
+    }
+
+    // The heading is looked up by the id `<Page>` assigns to it. When the page
+    // has no `<Title>` there is nothing to focus, so this is a no-op.
+    const heading = document.getElementById(titleId);
+    if (!heading) {
```
aromko, line 68: 💡 When there is no `<h1>`, focus is left wherever it was, which in a real app is often the link inside the screen that just unmounted. The browser then drops focus to `<body>` and the keyboard user lands at the top of the document, the exact problem this hook exists to solve.

`<Page>` always renders a `<main>` and always has an accessible name (it dev-errors otherwise), so it is a safe fallback target.

DST-1492 scopes the no-op deliberately and the test asserts it, so this is follow-up material rather than a change request here. Worth a ticket?

packages/components/src/Page/Page.stories.tsx
```diff
@@ -357,3 +360,142 @@ export const ExternalLabel = meta.story({
...
+          {Object.entries(ROUTES).map(([path, label]) => (
+            <Button key={path} onPress={() => setRoute(path)}>
+              {`Open ${label}`}
+            </Button>
+          ))}
+        </Inline>
+      </nav>
+      <Page aria-label={ROUTES[route]}>
+        <RouteFocus routeKey={route} />
+        <Panel aria-label="Summary">
```
aromko: 💡 `<Panel aria-label="Summary">` gives the region the accessible name "Summary" while it visibly renders "Team members". In an accessibility demo story that mismatch is a bit confusing.

Moving the `<Title>` into `<Panel.Header>` lets `useSlot` pick it up, so the `aria-label` can go and the region is named by its visible heading.

packages/components/src/Page/usePageFocus.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * <RouterProvider navigate={navigate}>
+ *   <Page>
+ *     <PageFocus pathname={location.pathname} />
+ *     <Outlet />
+ *   </Page>
+ * </RouterProvider>
+ * ```
+ */
+export const usePageFocus = (routeKey: string) => {
```
aromko, line 48: 💡 Design note, not a change request. Every consumer has to write the same null-rendering wrapper:

```tsx
const PageFocus = ({ pathname }) => { usePageFocus(pathname); return null; };
```

Because the hook must be inside `<Page>` *and* `<Page>` must persist across routes anyway, a `routeKey` prop on `<Page>` would do the same job with no wrapper:

```tsx
<Page routeKey={pathname}>
```

That stays opt-in and router-agnostic, so it does not conflict with the ticket's "no automatic, always-on behavior inside `Page`" exclusion. DST-1492 asked for a hook, so shipping this as-is is correct. Flagging it for the team as a possible follow-up ergonomics win.

packages/components/src/Page/Page.stories.tsx
```diff
@@ -357,3 +360,142 @@ export const ExternalLabel = meta.story({
...
+
+/**
+ * `usePageFocus` moves focus to the page `<h1>` on a route change, the standard
+ * SPA technique for announcing navigation to screen-reader and keyboard users.
+ * The router is the source of the route signal — pass the current pathname to
+ * the hook from a component that persists across navigations (here, a small
+ * harness swaps the "route" via the nav buttons). The initial mount is skipped,
+ * so first paint never steals focus.
+ */
+export const FocusOnRouteChange = meta.story({
```
aromko, line 418: 🔴 **Visual Regression Tests — not run**

This PR touches `packages/components/src/**` and a `*.stories.tsx` file but the `Visual-Regression-Tests` workflow has not run for this branch.

Caveat: both new stories set `chromatic: { disableSnapshot: true }`, no existing story's render output changes, and `usePageFocus` only mutates `tabIndex` on navigation, so a run would add no coverage here. Either trigger it with `/run-chromatic` or tick the checklist box with that rationale.

packages/components/src/Page/usePageFocus.ts
OsamaAbdellateef: Added a doc caveat in f498fdd99 rather than a code fix — the heading lookup is inherently synchronous (DOM `getElementById` at effect-run time), so there's no way to wait for a heading that mounts later without changing the hook's contract. Documented in both the `usePageFocus` JSDoc and the mdx docs: a route's `<Title>` behind `React.lazy` + `<Suspense>` needs to be hoisted above the boundary (or the route signal moved above it) for focus to move reliably.

packages/components/src/Page/Page.stories.tsx
OsamaAbdellateef: Done in f498fdd99: removed the `play` function from `FocusOnRouteChange`. The route-change focus assertions now live as separate tests in `usePageFocus.test.tsx`, rendering this story's `.Component`.

packages/components/src/Page/Page.stories.tsx
OsamaAbdellateef: Same treatment in f498fdd99: removed the `play` function from `FocusWithoutHeading`. The no-op assertions moved to `usePageFocus.test.tsx`.

packages/components/src/Page/Page.stories.tsx
OsamaAbdellateef: Fixed in f498fdd99: moved the JSDoc block down so it directly precedes `export const FocusWithoutHeading`.

packages/components/src/Page/Page.stories.tsx
OsamaAbdellateef: Addressed in f498fdd99 by dropping the `play` function entirely (see the '.test' rewrite thread above) — the two behaviours are now separate `test()` blocks in `usePageFocus.test.tsx`: one asserting no focus theft on initial paint, one asserting focus lands on the new heading with `tabindex="-1"` after a route change.

packages/components/src/Page/usePageFocus.ts
OsamaAbdellateef: Agreed this is follow-up material rather than a change for this PR. Left a `TODO(follow-up)` comment at the no-op branch in `usePageFocus.ts` (f498fdd99) so it's easy to pick up

packages/components/src/Page/Page.stories.tsx
OsamaAbdellateef: Fixed in f498fdd99: moved `<Title>` into `<Panel.Header>` and dropped the `aria-label`, so the panel is now named by its own visible heading instead of a mismatched "Summary".

packages/components/src/Page/usePageFocus.ts
OsamaAbdellateef: Agreed, shipping as-is per the ticket's scope. Left a `TODO(follow-up)` comment above `usePageFocus` in f498fdd99 noting the `routeKey` prop-on-`<Page>` idea for a future ergonomics pass.

packages/components/src/Page/Page.stories.tsx
OsamaAbdellateef: Not triggering `/run-chromatic` for the reason you gave: both new stories set `chromatic: { disableSnapshot: true }` and no existing story's render output changed in f498fdd99, so a run would add no coverage here. Flagging in case you'd still like it run for the record.

packages/components/src/Page/usePageFocus.ts
aromko: I would rather write a ticket. A comment will no one read afterwards 😅

packages/components/src/Page/usePageFocus.ts
aromko: Same write a ticket pls.

packages/components/src/Page/usePageFocus.ts
OsamaAbdellateef: ticket created: [DST-1668](https://reservix.atlassian.net/browse/DST-1668) (Unplanned, under DST-1319, linked to DST-1492).

It covers the `<main>` fallback, flipping the current "no-op" test to assert the fallback, and updating the JSDoc + Page docs that document the no-op as intended. Removing the TODO comment is part of that ticket's scope.

packages/components/src/Page/usePageFocus.ts
OsamaAbdellateef: Ticket created: [DST-1669](https://reservix.atlassian.net/browse/DST-1669) (Unplanned, under DST-1319, linked to DST-1492).

Scoped as an optional `routeKey` prop on `<Page>` that calls `usePageFocus` internally, staying opt-in and router-agnostic so it doesn't collide with DST-1492's "no always-on behavior inside `Page`" exclusion, with the hook kept exported as the escape hatch. Removing the TODO comment is part of that ticket's scope.

jim761 (approved):

aromko (approved):

PR #5669, opened 2026-07-22
feat(DST-1646): add OverflowRegion for measured item demotion

sebald:
# Description

Adds `OverflowRegion`, a single-row region that hides its trailing items instead of wrapping when horizontal space runs out, and restores them as space returns. Items stay mounted while hidden (state preserved) but are removed from paint, tab order, and the accessibility tree via `inert` + `visibility`, so consumers must pair the region with an `indicator` (a "More" trigger rendered while items are hidden) and/or an `onOverflowChange` callback so hidden items remain reachable through another surface, such as a menu or a filter panel.

Measurement happens in place inside a clipped (`overflow-clip`) region: nothing is ever `display: none`, so `getBoundingClientRect` always reports real widths without a hidden measurement clone or a transient "show all" render pass. A `ResizeObserver` (region + items, so an item resizing itself is caught too) re-measures on demand; the first client-side measurement runs pre-paint in a layout effect so server-rendered (all-visible) markup is corrected without a flash.

Included in this branch:
- `OverflowRegion` + `useOverflowRegion` implementation. Like the other layout primitives (`Inline`, `Stack`), it has no theme layer: it styles itself with utilities via `cn()`, takes a `space` prop (inheriting `--space` from a parent layout when unset), and forwards `AriaRegionProps` (`role`, `aria-label`, ...). No `theme-rui` styles and no `@marigold/system` `Theme` type slot are added.
- Stories covering demotion order, priority navigation, an external-surface recipe (`onOverflowChange` badging a pinned trigger), and a filter-bar recipe (quick filters demoting into an "All filters" drawer).
- Standalone docs page (`docs/content/components/layout/overflow-region`) with interactive resizable demos via a new `DemoResizer` helper.

No public API or measurement-mechanism changes from what's documented above; this is the initial introduction of the component.

### Scope note (vs. DST-1646 as originally written)

DST-1646 originally scoped a dedicated `FilterBar` compound component plus updates to `/examples/filter` and the `filter-bar-layout` / `filter-promoted` demos. During implementation the direction shifted to shipping a **generic, reusable primitive** (`OverflowRegion`) with a filter-bar *recipe* in the stories and docs, rather than a bespoke `FilterBar`. This is more broadly useful (toolbars, priority-nav, and filter bars all reuse it) and keeps the filter surface composed from existing components.

Two consequences, to be tracked separately:
- The `FilterBar` compound component and the `/examples/filter` rewire onto `OverflowRegion` are **deferred to a follow-up** (overlaps the pending PR #5522 rewire), not dropped.
- The counter-badge semantics were **deliberately changed** from the ticket's "count active-but-demoted" to "count *applied* filters". A number on a filter trigger conventionally reads as "n filters are active", so a count that changes purely on resize would mislead; counting applied filters also covers the active-but-demoted case. The ticket will be updated to reflect this.

Closes DST-1646

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. `pnpm sb` -> **Components/OverflowRegion** -- exercise Demotion, PriorityNavigation, Resize, and FilterBar stories (resize the viewport / use the "Toggle width" button to trigger demotion and restoration).
2. `pnpm start` -> **Components -> Layout -> Overflow Region** -- try the interactive resizable demos (drag the handle or focus + arrow keys).
3. `pnpm typecheck:only`, `pnpm test:unit -- OverflowRegion`, `pnpm test:sb -- OverflowRegion` -- all pass.

## Breaking Changes

No -- purely additive (a single new component export, `OverflowRegion`). No theme slot and no `Theme` type changes.

## Checklist


aromko (commented):
## Automated Code Review

Solid first cut of `OverflowRegion` — the measure-in-place design (clipped region, `visibility` + `inert`, no hidden clone or two-pass render) is careful and well-documented, tests are thorough (tab-order coverage 👏), and the component follows the layout-primitive conventions of `Inline`/`Stack`.

**Should fix:**

- ⚠️ PR description claims `theme-rui` styles and a `@marigold/system` `Theme` type extension that aren't in the diff — stale text or missing files?
- ⚠️ DST-1646's scope is a `FilterBar` compound component plus updates to `/examples/filter` and the `filter-bar-layout`/`filter-promoted` demos; this PR ships the generic primitive with a filter-bar recipe and closes the ticket. If the direction changed (reasonably!), please update the ticket or file a follow-up so the FilterBar/example work isn't dropped. Same for the badge semantics (ticket: count active-but-demoted; PR: count applied — the PR's reasoning is convincing, but the ticket should reflect it).

**Suggestions:** see inline comments (observer re-attach on children swap, `useLayoutEffect` comment, `AriaRegionProps` consistency, `DemoResizer` `onPointerCancel`, link semantics in the nav demo).

---

packages/components/src/OverflowRegion/useOverflowRegion.ts
```diff
@@ -0,0 +1,138 @@
...
+      setVisibleCount(count);
+    };
+
+    const observer = new ResizeObserver(calculate);
+    observer.observe(region);
+    for (const child of region.children) {
+      observer.observe(child);
+    }
+    return () => observer.disconnect();
+  }, [region, itemCount]);
```
aromko: 💡 The effect re-observes only when `region` or `itemCount` changes. If children are replaced with the *same* count but different keys, the new wrapper nodes are never observed — an item growing on its own (the case the comment above cites, e.g. a select with a longer value) goes undetected until the region itself resizes. Consider keying the effect by the items' keys as well, or observing wrappers via callback refs.

packages/components/src/OverflowRegion/OverflowRegion.tsx
```diff
@@ -0,0 +1,125 @@
...
+    getIndicatorProps();
+
+  const state: OverflowRegionState = {
+    count: items.length,
+    visibleCount,
+    hiddenCount: items.length - visibleCount,
+  };
+
+  // Notify subscribers after paint, only when the visible count settles on
+  // a new value.
```
aromko: 💡 The comment says "after paint", but `useLayoutEffect` runs *before* paint. If pre-paint is intentional (to let a consumer badge update without flicker), just fix the comment; otherwise use `useEffect`.

packages/components/src/OverflowRegion/OverflowRegion.tsx
```diff
@@ -0,0 +1,125 @@
...
+export interface OverflowRegionState {
+  /** Total number of items in the region. */
+  count: number;
+  /** Number of items that currently fit. */
+  visibleCount: number;
+  /** Number of items that are currently hidden. */
+  hiddenCount: number;
+}
+
+export interface OverflowRegionProps {
```
aromko: 💡 `Inline` and `Stack` extend `AriaRegionProps`, so they can take `aria-label`/role. `OverflowRegionProps` can't — worth adding for consistency, e.g. when the region itself should be a labelled toolbar.

docs/ui/DemoResizer.tsx
```diff
@@ -0,0 +1,111 @@
...
+        role="separator"
+        aria-orientation="vertical"
+        aria-label="Resize demo area"
+        aria-valuenow={width}
+        aria-valuemin={minWidth}
+        aria-valuemax={Number.isFinite(maxWidth) ? maxWidth : undefined}
+        tabIndex={0}
+        onPointerDown={onPointerDown}
+        onPointerMove={onPointerMove}
+        onPointerUp={onPointerUp}
```
aromko, line 105: 💡 Without `onPointerCancel`, a cancelled drag (touch interruption) leaves `dragStartRef` set and later hover moves resize without pressing.
```suggestion
        onPointerUp={onPointerUp}
        onPointerCancel={onPointerUp}
```

docs/content/components/layout/overflow-region/overflow-region-priority-nav.demo.tsx
```diff
@@ -0,0 +1,31 @@
...
+              <Menu.Item key={link} id={link}>
+                {link}
+              </Menu.Item>
+            ))}
+          </Menu>
+        )}
+      >
+        {links.map(link => (
+          <Button key={link} variant="ghost">
+            {link}
```
aromko, line 25: 💡 A navigation demo teaching Priority+ would be semantically stronger with links instead of `Button`s inside `<nav>` — readers copy demos verbatim.

.changeset/olive-wombats-overflow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@marigold/components': minor
+---
+
+feat(DST-1646): add `OverflowRegion`, a single-row layout primitive that hides trailing items instead of wrapping when horizontal space runs out, and restores them as space returns. Hidden items stay mounted (state preserved) but are removed from paint, tab order, and the accessibility tree. Space its items with the `space` prop (like `Inline`), and pair it with the `indicator` render prop or the `onOverflowChange` callback so hidden items stay reachable through another surface, e.g. quick filters demoting into a filter panel.
```
aromko: ⚠️ The PR description says this branch includes theme styles (`theme-rui`) and a `Theme` type extension in `@marigold/system`, but neither is in the diff (and the component styles itself like `Inline`, so none seem needed). Please reconcile the description — or push the missing files if they were intended.

packages/components/src/OverflowRegion/useOverflowRegion.ts
sebald: Good catch, fixed in 1dbf8fb. Kept the single unified observer, but the hook now takes an `itemsKey` (a signature of the children's keys) and includes it in the effect deps, so replacing or reordering items at the same count re-runs the effect and re-attaches the observer to the fresh wrapper nodes. The same-key case (e.g. a select growing with a longer value) was already covered, since React keeps the same wrapper `<div>` and it stays observed.

packages/components/src/OverflowRegion/OverflowRegion.tsx
sebald: Pre-paint is intentional (so a consumer badging an external trigger updates in the same frame the row reflows, without a flicker). Fixed the misleading comment in 1dbf8fb to say "before paint" and explain why.

packages/components/src/OverflowRegion/OverflowRegion.tsx
sebald: Agreed. `OverflowRegionProps` now extends `AriaRegionProps` and the region forwards them, matching `Inline`/`Stack` (so the region can be a labelled toolbar). Done in 1dbf8fb.

docs/ui/DemoResizer.tsx
sebald: Applied the suggestion in 1dbf8fb (`onPointerCancel={onPointerUp}`).

docs/content/components/layout/overflow-region/overflow-region-priority-nav.demo.tsx
sebald: Agreed, a Priority+ nav demo should use real links. Swapped the `Button`s for `Link`s inside the `<nav>` in 1dbf8fb.

.changeset/olive-wombats-overflow.md
sebald: You're right, the description was stale. Reconciled it in 1dbf8fb: the component styles itself like `Inline` (no `theme-rui` styles, no `@marigold/system` `Theme` slot), so the description and Breaking Changes section now say so, and I added a scope note explaining the shift to a generic primitive (the `FilterBar` compound + `/examples/filter` rewire are deferred to a follow-up, and the badge semantics change is deliberate). The changeset frontmatter was already correctly scoped to `@marigold/components` only; its text now also mentions the forwarded region props.

packages/components/src/OverflowRegion/OverflowRegion.stories.tsx
```diff
@@ -0,0 +1,336 @@
...
+);
+
+export const Basic = meta.story({
+  render: () => <DemoRow />,
+});
+
+export const Demotion = meta.story({
+  tags: ['component-test'],
+  render: () => (
+    <div className="w-[560px]">
```
aromko: Maybe use just another viewport for this story?

packages/components/src/OverflowRegion/OverflowRegion.stories.tsx
```diff
@@ -0,0 +1,336 @@
...
+  'restores hidden items when space returns',
+  { parameters: { chromatic: { disableSnapshot: true } } },
+  async ({ canvas, userEvent }) => {
+    // Narrow: the trailing item is hidden; `onOverflowChange` reports it.
+    await waitFor(() =>
+      expect(canvas.queryByRole('button', { name: /price/i })).toBeNull()
+    );
+    expect(canvas.getByText('1 of 3 visible')).toBeVisible();
+
+    // Grow: every item fits again, the indicator disappears.
```
aromko: instead of these comments you could use steps - so the description is visible when running the test

packages/components/src/OverflowRegion/OverflowRegion.stories.tsx
```diff
@@ -0,0 +1,336 @@
...
+      {overflow ? (
+        <span>
+          {overflow.visibleCount} of {overflow.count} visible
+        </span>
+      ) : null}
+    </Stack>
+  );
+};
+
+export const Resize = meta.story({
```
aromko, line 167: I think in this story the snapshots can be disabled. no real visual differences to the basic and demotion stories.

packages/components/src/OverflowRegion/OverflowRegion.stories.tsx
```diff
@@ -0,0 +1,336 @@
...
+  'Orders',
+  'Reports',
+  'Settings',
+  'Team',
+];
+
+export const PriorityNavigation = meta.story({
+  tags: ['component-test'],
+  render: () => (
+    <nav aria-label="Main" className="w-[480px]">
```
aromko: Same here maybe just using another viewport and disable snapshot?

packages/components/src/OverflowRegion/OverflowRegion.stories.tsx
```diff
@@ -0,0 +1,336 @@
...
+        </Drawer>
+      </Drawer.Trigger>
+    </Inline>
+  );
+};
+
+export const FilterBar = meta.story({
+  tags: ['component-test'],
+  render: () => (
+    <div className="w-[640px]">
```
aromko: Use viewport from storybook. And in general not sure if this story works well. Shows one applied filter but no filter is applied when opening the drawer.

packages/components/src/OverflowRegion/OverflowRegion.stories.tsx
sebald: Done in f9a608a3. `Demotion` now uses `globals: { viewport: { value: 'smallScreen' } }` (639px) instead of the `w-[560px]` wrapper, with an `expect(window.innerWidth).toBeLessThan(640)` guard so the test fails loudly if the viewport doesn't apply (same approach as `Tabs.stories.tsx`). The snapshot is pinned with `chromatic: { viewports: [639] }` so Chromatic still captures the demoted state at that width.

packages/components/src/OverflowRegion/OverflowRegion.stories.tsx
sebald: Good call, done in f9a608a3. The play functions now wrap their phases in `step(...)`, so the descriptions surface in the test runner instead of living in comments.

packages/components/src/OverflowRegion/OverflowRegion.stories.tsx
sebald: Agreed, disabled in f9a608a3. `Resize` is interaction-only (the toggle drives the width), so there's no static visual delta from Basic/Demotion. Only Basic (all fit) and Demotion (demoted) still snapshot now.

packages/components/src/OverflowRegion/OverflowRegion.stories.tsx
sebald: Done in f9a608a3. `PriorityNavigation` now uses the `extraSmallScreen` (320px) viewport to force the overflow, and its snapshot is disabled (interaction-only; the "More" menu is exercised by the play test).

packages/components/src/OverflowRegion/OverflowRegion.stories.tsx
sebald: Both addressed in f9a608a3:

- Switched to the `smallScreen` viewport (no wrapper div).
- You were right that the story didn't hold up — the badge showed an applied filter but the drawer didn't reflect it. I reworked the recipe to follow the Filter pattern's *instant bar + batched panel* model (like `examples/filter/Toolbar.tsx`): the bar quick filters apply instantly, and the drawer is now a `<Form>` with `Drawer.Actions` (Close + primary **Apply**) that commits an uncontrolled draft via `parseFormData`. The applied Price is preselected in the panel, editing a draft doesn't move the badge until Apply, and both surfaces write the same applied state so they stay in sync.

Leaving this thread open so you can confirm the reworked behavior. One open question: Apply commits but doesn't auto-close the drawer (matching `Toolbar.tsx`) — happy to make Apply also close if you'd prefer.

aromko (approved):

PR #5670, opened 2026-07-22
fix(DST-1649): run CLI main() when invoked via a symlinked bin

aromko:
# Description

`marigold <command>` invoked through a symlinked bin exited silently with code 0. The entry-point guard compared `fileURLToPath(import.meta.url)` (always the realpath) with `path.resolve(process.argv[1])` (the symlink path), so `main()` never ran. Since global bins are symlinks, this affects published `npm install -g @marigold/cli` installs, not just local dev links. `argv[1]` is now resolved through `realpathSync` (with a `path.resolve` fallback), so the CLI behaves identically via its real path, a symlink, or `node dist/bin/marigold.mjs`.

Also rewrites the README local-development instructions: `pnpm link --global` was removed in pnpm 10 — on pnpm 11 it instead mutates `package.json`/the lockfile and drops a stray `pnpm-workspace.yaml`. The docs now symlink the built entry point into `$(pnpm bin -g)`.

**Versions:**

- **Affected:** every published `@marigold/cli` release up to and including `0.4.0`; the changeset ships this fix as the next patch.
- **pnpm:** the new README link instructions target pnpm ≥ 10 (this repo pins `pnpm@11`); `pnpm link --global` still works on pnpm 9 and earlier.
- **Node:** unchanged, ≥ 22 per the package `engines` field.

No UI changes — stories, screenshots, and VRT are N/A.

Closes DST-1649

## Test Instructions

1. `pnpm --filter @marigold/cli build`
2. `ln -sf "$(pwd)/packages/cli/dist/bin/marigold.mjs" "$(pnpm bin -g)/marigold"`
3. `marigold --version` — prints the version. Before this fix: no output, exit code 0.
4. `pnpm --filter @marigold/cli test` — all 280 tests pass (includes a new subprocess regression test that runs the built CLI through a temp symlink).

## Breaking Changes

No

## Checklist

### DIFF
--- a/.changeset/cli-symlink-entrypoint.md
+++ b/.changeset/cli-symlink-entrypoint.md
@@ -1,0 +1,5 @@
+---
+'@marigold/cli': patch
+---
+
+fix: run the CLI when invoked via a symlinked bin. The entry-point guard compared `import.meta.url` (always the realpath) against `path.resolve(process.argv[1])`, so symlinked global bins (`npm install -g`, manual links) exited silently with code 0. `argv[1]` is now resolved through `realpathSync`.
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -26,9 +26,9 @@ deployment:
 # 1. Build the CLI
 pnpm --filter @marigold/cli build
 
-# 2. Link it globally so `marigold ...` resolves to the local build
-cd packages/cli
-pnpm link --global
+# 2. Symlink the built entry point into your global bin so `marigold ...`
+#    resolves to the local build (pnpm ≥10 removed `pnpm link --global`)
+ln -sf "$(pwd)/packages/cli/dist/bin/marigold.mjs" "$(pnpm bin -g)/marigold"
 
 # 3. Point at a docs origin (preview or local). Create
 #    packages/cli/.env.local with:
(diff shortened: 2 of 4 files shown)


jim761 (commented):
## Automated Code Review

**⚠️ Missing regression test for the symlink fix**

The fix itself looks correct. The one thing worth adding: there's no test guarding the exact regression this PR fixes.

`packages/cli/src/bin/marigold.test.ts` imports `main()` directly and deliberately relies on the entry-point guard being *false* under vitest, so `resolveEntry`/`isEntryPoint` are never exercised. The silent-exit-via-symlink failure mode wouldn't be caught by the suite.

A subprocess-based regression test would close the gap: create a temp symlink to the built `marigold.mjs`, `spawn` `node <symlink> --version`, and assert non-empty stdout / exit 0 with output.

_(Caveat: the `@marigold/cli` suite is excluded from CI vitest, so this guards local/pre-publish runs rather than CI — but given the bug was a silent failure that shipped in published releases, it's worth pinning down.)_

---

aromko:
Addressed in ff43e06d5: added `packages/cli/src/bin/entrypoint.test.ts`, a subprocess-based regression test that symlinks the built `marigold.mjs` into a temp dir and asserts `node <symlink> --version` prints the package version (plus a direct-path control). Verified it catches the regression: against a build of the pre-fix code, the symlink case fails with empty stdout — the exact shipped failure mode.

sarahgm (approved):
Worked after I updated my pnpm global settings with `pnpm setup --force`

jim761 (approved):

PR #5671, opened 2026-07-23
docs(DST-1651): sharpen the Pick pattern (commit button, selection limits, on-page picking, example filters)

sarahgm:
# Description

First-round feedback on the just-shipped Pick pattern (DST-1532) surfaced how to name a pick's commit button and how to keep its example from dead-ending. This PR addresses that feedback and the follow-ups it opened, all docs-only.

**Commit button.** Promotes the commit-button guidance into its own "Naming the commit button" section. The button always names its outcome with the host task's own verb ("Add venues", "Assign users", "Save venues"), never a navigation word ("Next", "Continue", "Done"), the wrong-pattern word ("Filter"), or an over-promising "Save" for a discardable result. The staged count is treated as status, and its home is the removable-tag rail that already shows the set. The label may echo the count ("Add 3 venues") when it stays short and mirrors the trigger, and stays a bare verb when the label would run long or a limit applies. The reframe is backed by research into how real products place the count, which splits roughly evenly between putting it in the button and beside it, so count-in-label reads as a blessed option rather than a rule.

**Selection limits.** Lifts the terse "bound the selection" bullet into its own "Selection limits" subsection. State the rule up front in helper text and keep a running count on the staged rail ("Staged (3 of 5)"). A minimum keeps the commit active and refuses a too-small set on press, revealing a `SectionMessage` at the top of the surface that says what is still needed and announces itself to assistive technology, rather than disabling the button. A maximum makes the ceiling visible, disables the unchecked rows once reached with a reason, and keeps deselection free so the user can swap one choice for another.

**On-page picking.** Reframes the on-page tier of "Choose the surface by collection size" as two shapes, chosen by how the options are best shown: a searchable field (`TagField`) for a roster the user knows by name, and an inline list when the user needs to see every option and its state at once. Adds a second demo for the inline-list shape, an on-page pick of concert dates for a season subscription (an Abonnement) with sold-out dates visibly disabled, so the reader sees the case where inline beats a search.

**Faceted example filters.** Makes the `/examples/pick` venue picker's Type, Region, and Status filters reference one another. Each option carries its result count under the other active filters and the search, and options that would return nothing are disabled, so a user is steered away from empty combinations (picking "Theater" then a region with no theaters was the reported case). Counts derive from the same predicate that filters the table so they never disagree with the results, and the current value is never disabled so a filter can always be changed back.

Closes DST-1651

## Test Instructions

1. Run `pnpm start` and open `/patterns/user-input/pick`.
2. Read "Naming the commit button" with its Do/Don't tiles, and the "Selection limits" subsection.
3. In "Choose the surface by collection size", try the two on-page demos: the performers `TagField` (searchable) and the concert-date subscription (inline list). Confirm the sold-out dates are visibly disabled, the minimum is stated up front, and that "Add to subscription" stays active: pressing it with nothing picked reveals an error message (which re-announces on each empty press) instead of committing, and staging a date clears it.
4. Open `/examples/pick`. Narrow the filters and confirm each option shows a count, options at zero are disabled, the currently selected value is never disabled, and an all-empty combination falls back to the table empty state.

## Breaking Changes

No

## Checklist

- [ ] Visual regression tests updated (for UI changes). Not applicable, docs-only and targets `beta-release`.


sebald:
Reviewed the diff — the reframe reads well and the faceted counting logic is right (excluding each facet's own constraint is the correct shape, and deriving the counts from the same predicate as the table is the detail that keeps them honest). Component APIs all check out too: `SectionMessage` `open`/`variant="error"` announces assertively by default, `Select` `renderValue`/`disabledKeys`, `SelectList` with `TextValue`/`Description`, `Text variant="muted"`, `Stack alignX`.

Four things worth addressing:

**1. A second empty press does nothing.** Once `attemptedEmpty` is `true`, pressing the commit again produces no visual change, and `SectionMessage` only announces on a hidden → visible transition — so there is no feedback at all for the repeat press. The component's own docs note that a same-content re-announce needs a `key` remount. A press counter used as `key` (or reset-then-set) would make each refused press announce.

**2. The message can reappear without a press.** `attemptedEmpty` is never reset, and visibility is `attemptedEmpty && count === 0`. So staging a record (message hides) and then clearing the set re-shows the error unprompted, before the user has tried to commit again. Resetting it in `onChange`/`onSelectionChange` would match the guidance's "clear it the moment the set becomes valid".

**3. The demos don't implement the helper text the guidance asks for.** "Selection limits" says to state the rule up front and pair the error with helper text so the requirement is known before the user reaches the footer — but none of the four surfaces (fullscreen example, `pick-dialog`, `pick-select-list`, `pick-abonnement-dates`) states the minimum before the footer. The abonnement demo's `description` mentions sold-out dates instead. Also unillustrated: the maximum behavior and the `Staged (3 of 5)` rail count the section prescribes. Fine to leave the max as text-only, but the minimum is cheap to show and the guidance points at `/examples/pick` as the reference implementation.

**4. The PR description is stale.** Its "Selection limits" paragraph ("A minimum holds the commit disabled until it is met") and test step 3 ("`Add to subscription` is disabled until at least one date is picked") describe the pre-`7cab05c` behavior — the opposite of what the branch now does. The changeset is correct; only the PR body needs updating before merge.

Minor: deriving `types` from the data changes the option order (Theater moves from third to last). Region grouping survives. Consider sorting, or keeping the curated order the way `statuses` does.

docs/content/patterns/user-input/pick/index.mdx
```diff
@@ -83,14 +89,47 @@ The example below adds venues from a searchable, filterable, multi-select table.
...
 - **Never clear staged selections while filtering.** Typing in the search narrows the visible rows, but records staged earlier stay staged even when they scroll out of view. Clearing them on every keystroke would make the pick impossible. This is the same tension as the filter paradox in [Table Records](/patterns/user-input/table-record-management).
 - **Show records the way the user recognizes them.** When a record carries detail worth scanning, such as a city or a capacity, show each attribute in its own [`<Table>`](/components/collection/table) column instead of a bare name. A plain list is enough when the label stands on its own.
 - **Offer select all for bulk picks.** A header control that selects every row, shown in a mixed state while the selection is partial, saves the user from working through a long list by hand. Keep it paired with search and filters so "all" applies to a sensibly narrowed set.
 - **Cancel discards, commit persists.** Closing or cancelling throws the staged set away with no consequence. Only the primary action carries the selection back to the task.
 
-### A list when the label is enough
+### Naming the commit button
```
sebald, line 97: Does this need to be that long? :D

sarahgm: you mean the section or the headline?

sarahgm:
Thanks, all five addressed in 72eb41a89.

**1. Second empty press now re-announces.** Swapped the sticky `attemptedEmpty` boolean for an `emptyAttempts` press counter used as `key` on the `SectionMessage`, so each refused press remounts it and announces again. This is the same key-remount idiom the component's own docs and `section-message-aria-live-error` demo prescribe for a same-content re-announce.

**2. Message no longer reappears without a press.** The counter resets to `0` on any selection change (`onChange` / `onSelectionChange`), so staging then clearing the set never re-shows the error unprompted. Clearing it the moment the set changes matches the guidance's wording.

**3. Minimum stated up front on all four surfaces.** Added a muted "Pick at least one …" line above the filters in the fullscreen example, `pick-dialog`, and `pick-select-list`, and changed the abonnement `SelectList` `description` to lead with the rule ("Choose at least one concert for your season pass. Sold-out dates can't be added."). Left the maximum and the "Staged (3 of 5)" count as text-only, as you suggested.

**4. PR description updated.** The "Selection limits" paragraph and test step 3 now describe the active-button, refuse-on-press behavior instead of the pre-`7cab05c` disabled-until-met wording.

**Minor: filter option order.** Sorted the derived Type and Region options with `localeCompare`, leaving `statuses` explicit for its lifecycle order.

docs/app/(examples)/examples/pick/PickVenuesDialog.tsx
```diff
@@ -100,6 +199,26 @@ const PickBody = ({
...
+          >
+            <SectionMessage.Title>Nothing staged yet</SectionMessage.Title>
+            <SectionMessage.Content>
+              Tick at least one venue to add it to the report.
+            </SectionMessage.Content>
+          </SectionMessage>
+
+          {/* State the one-venue minimum up front, so the rule is known before
+              the footer rather than only on a refused press. */}
+          <Text variant="muted" fontSize="sm">
```
sebald: The up-front rule is visual and reading-order only here — a bare `<Text>` inside `Dialog.Content` isn't associated with the table or the commit button, so a screen reader user who tabs straight to either never hears it. That's the gap the line was added to close, and the abonnement demo already does it properly via `SelectList`'s `description`.

`Dialog.Description` would fix it: it renders `<Description slot="description">`, i.e. the dialog's `aria-describedby`. One caveat from `DialogDescription.tsx:14-27` — outside `<Dialog.Header>` it lands in the same grid area as `<Dialog.Content>` and the two overlap, so it means wrapping the title and the description together:

```tsx
<Dialog.Header>
  <Dialog.Title>{title}</Dialog.Title>
  <Dialog.Description>
    Pick at least one venue to add it to the report.
  </Dialog.Description>
</Dialog.Header>
```

Worth doing since the docs point at this example as the reference implementation. Same applies to `pick-dialog` and `pick-select-list`.

docs/content/patterns/user-input/pick/pick-dialog.demo.tsx
```diff
@@ -139,6 +145,26 @@ const PickVenuesBody = ({ initial, onConfirm }: PickBodyProps) => {
...
+          >
+            <SectionMessage.Title>Nothing staged yet</SectionMessage.Title>
+            <SectionMessage.Content>
+              Tick at least one venue to add it.
+            </SectionMessage.Content>
+          </SectionMessage>
+
+          {/* State the one-venue minimum up front, so the rule is known before
+              the footer rather than only on a refused press. */}
+          <Text variant="muted" fontSize="sm">
```
sebald: Same as in the fullscreen example: this line isn't programmatically associated with anything, so it's invisible to a screen reader user who tabs to the table or the footer. `<Dialog.Header>` + `<Dialog.Description>` would put it on the dialog's `aria-describedby`.

docs/content/patterns/user-input/pick/pick-select-list.demo.tsx
```diff
@@ -151,6 +157,26 @@ const PickPeopleBody = ({ initial, onConfirm }: PickBodyProps) => {
...
+          >
+            <SectionMessage.Title>No one staged yet</SectionMessage.Title>
+            <SectionMessage.Content>
+              Tick at least one person to grant access.
+            </SectionMessage.Content>
+          </SectionMessage>
+
+          {/* State the one-person minimum up front, so the rule is known before
+              the footer rather than only on a refused press. */}
+          <Text variant="muted" fontSize="sm">
```
sebald: Same here — and this one has a second option, since the collection is a `<SelectList>`: its `description` prop is wired to the field, which is how `pick-abonnement-dates` states its minimum. Either that or `<Dialog.Description>`.

docs/app/(examples)/examples/pick/PickVenuesDialog.tsx
```diff
@@ -11,34 +11,108 @@ import {
...
-  'Tyrol',
-  'Zurich',
-  'Basel-Stadt',
-];
+// Type and region options are derived from the data so they never drift from
+// it: every option maps to real venues, and every venue value is selectable.
+// Sorted so the option order is stable and predictable.
+const unique = (key: 'type' | 'region') =>
+  [...new Set(venues.map(venue => venue[key]))].sort((a, b) =>
+    a.localeCompare(b)
```
sebald, line 28: Flagging the trade rather than asking for a change: alphabetical sorting drops the geographic grouping the hardcoded list had (DE → AT → CH). Regions now read Baden-Württemberg, Basel-Stadt, Bavaria, Berlin…, so Swiss and German regions interleave. For a 15-item filter alphabetical is the conventional and more scannable choice, so I'd keep it — just noting it isn't a neutral reordering, unlike `types`.

sebald (commented):
Follow-up on the four findings — all resolved, and the mechanisms are right:

- **Re-announce:** the counter-as-`key` is exactly the mechanism `SectionMessage` needs — `announce()` sits in a `useEffect` keyed on `[isOpen]` (`SectionMessage.tsx:138-148`), so a remount is the only way to re-announce identical text.
- **Unprompted reappearance:** resetting in every selection handler closes it. The tag-rail removal path can only reach an empty set from a non-empty one, and staging always goes through the reset, so no path re-shows the error without a press.
- **Up-front minimum:** now on all four surfaces.
- **PR body:** matches the shipped behavior.

Two optional observations inline, neither blocking.

docs/app/(examples)/examples/pick/PickVenuesDialog.tsx
sarahgm: Fixed in 4e2a70f69. The up-front rule now lives in `Dialog.Header` + `Dialog.Description`, so it lands on the dialog's `aria-describedby` and is announced when tabbing to the table or the footer. The header wrap keeps it out of the `Dialog.Content` grid area, as you flagged. Same change applied to `pick-dialog` and `pick-select-list`.

docs/content/patterns/user-input/pick/pick-dialog.demo.tsx
sarahgm: Done in 4e2a70f69 via `Dialog.Header` + `Dialog.Description`.

docs/content/patterns/user-input/pick/pick-select-list.demo.tsx
sarahgm: Done in 4e2a70f69. I used `Dialog.Description` here as well, to keep this parallel with the `pick-dialog` sibling rather than routing it through `SelectList`'s `description`.

docs/app/(examples)/examples/pick/PickVenuesDialog.tsx
sarahgm: Agreed, keeping the alphabetical sort. Thanks for putting the DE -> AT -> CH trade on the record so the choice is explicit.

sebald (approved):
All four review findings are resolved, and the follow-up moving the minimum into `Dialog.Header` + `Dialog.Description` is the right structure. Approving.

One correction to my earlier comment, for the record: `Dialog.Description` does **not** land on the dialog's `aria-describedby` — `Dialog.tsx` wires only `aria-labelledby` from the title slot, and RAC's Dialog does the same, so the description is header content rather than an accessible description. The move is still an improvement (canonical header, one place for the rule, first thing after the title), just not the programmatic association I claimed. Teaching `Dialog` to wire a description slot to `aria-describedby` is a component-level gap and gets its own ticket — nothing to change here.

PR #5674, opened 2026-07-24
fix(Radio): apply width via CSS variable instead of raw class name

sebald:
# Description

> **Supersedes #5659** (by @zigzagdev). That PR targeted `main`; this one carries the same fix forward onto `beta-release`, where `Radio` has been refactored (React 19 ref-as-prop + deep subpath imports), so the change had to be hand-applied rather than cherry-picked. Full credit to @zigzagdev for the original fix and review iteration — this PR preserves their work. #5659 is being closed in favor of this one.

Forward-port of #5659 to `beta-release`.

`Radio` built its `width` class name from the raw token directly (`cn(width || groupWidth || 'w-full')`, where `width` could be a literal `"1/2"`), which Tailwind can't statically detect at build time — so no CSS was ever generated and setting `width` had no visible effect. `Radio` now follows the same pattern already used by `FieldBase`/`TextField`/`NumberField`: a static `w-(--field-width)` class with the actual value injected via `createWidthVar`. Individual `Radio` items inside a sized `Radio.Group` now inherit the group's already-computed width instead of recomputing it, which previously caused a double-shrink (e.g. `width="1/2"` rendering at 1/4).

### Why a hand-port rather than a cherry-pick

`beta-release` diverged from `main` on `Radio.tsx` (DST-1432 deep subpath imports + the React 19 ref-as-prop refactor), so #5659's commits don't apply cleanly. The **logical** change is identical to what was reviewed on #5659; only the surrounding component structure differs.

### Verified still needed on `beta-release`

Reproduced the bug against pristine `beta-release`: `<Radio width="1/2">` renders the field with a literal `1/2` class (which Tailwind never generates), no `w-(--field-width)`, and an empty `--field-width` var. No safelist rescues the raw token. So the fix is genuinely required here.

## Test Instructions

1. Storybook → `Components/Radio` → `WithOwnWidth`: an individual radio with `width="1/2"` renders at half width.
2. Storybook → `Components/Radio` → `Basic`: set the group `width` control — the whole group resizes without individual items double-shrinking.
3. `pnpm test:unit packages/components/src/Radio/Radio.test.tsx` — **8 passed** locally (6 existing + 2 new width assertions).

## Breaking Changes

No

## Checklist

### DIFF
--- a/packages/components/src/Radio/Radio.stories.tsx
+++ b/packages/components/src/Radio/Radio.stories.tsx
@@ -88,6 +88,17 @@ export const Basic = meta.story({
   ),
 });
 
+export const WithOwnWidth = meta.story({
+  render: () => (
+    <Radio.Group label="Label">
+      <Radio value="1" width="1/2">
+        Option 1
+      </Radio>
+      <Radio value="2">Option 2</Radio>
+    </Radio.Group>
+  ),
+});
+
 export const Error = meta.story({
   render: args => (
     <Radio.Group errorMessage="Das ist ein Error" error {...args}>
--- a/packages/components/src/Radio/Radio.stories.tsx
+++ b/packages/components/src/Radio/Radio.stories.tsx
@@ -89,13 +89,27 @@ export const Basic = meta.story({
 });
 
 export const WithOwnWidth = meta.story({
+  parameters: {
+    docs: {
+      description: {
+        story:
+          'An individual `Radio` can set its own `width`. The tinted boxes mark ' +
+          'each radio field: the first spans half the group, the second the full width.',
+      },
+    },
+  },
   render: () => (
-    <Radio.Group label="Label">
-      <Radio value="1" width="1/2">
-        Option 1
-      </Radio>
-      <Radio value="2">Option 2</Radio>
-    </Radio.Group>
+    // The tint marks the radio field — the element `width` actually sizes.
+    // Without it the demo looks unchanged, since a radio dot and its label
+    // never fill their container on their own.
+    <div className="w-96 [&_[data-rac]:has(>label)]:bg-slate-200">
+      <Radio.Group label="Label">
+        <Radio value="1" width="1/2">
+          Option 1
+        </Radio>
+        <Radio value="2">Option 2</Radio>
+      </Radio.Group>
+    </div>
   ),
 });
 
--- a/packages/components/src/Radio/Radio.stories.tsx
+++ b/packages/components/src/Radio/Radio.stories.tsx
@@ -98,12 +98,12 @@ export const WithOwnWidth = meta.story({
       },
     },
   },
-  render: () => (
+  render: args => (
     // The tint marks the radio field — the element `width` actually sizes.
     // Without it the demo looks unchanged, since a radio dot and its label
     // never fill their container on their own.
     <div className="w-96 [&_[data-rac]:has(>label)]:bg-slate-200">
-      <Radio.Group label="Label">
+      <Radio.Group {...args}>
         <Radio value="1" width="1/2">
           Option 1
         </Radio>
--- a/packages/components/src/Radio/Radio.test.tsx
+++ b/packages/components/src/Radio/Radio.test.tsx
@@ -1,5 +1,5 @@
 import { render, screen } from '@testing-library/react';
-import { Basic, Error } from './Radio.stories';
+import { Basic, Error, WithOwnWidth } from './Radio.stories';
 
 // Tests
 // ---------------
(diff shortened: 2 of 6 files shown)


packages/components/src/Radio/Radio.test.tsx
```diff
@@ -52,3 +52,25 @@ test('radio group can be horizontal', () => {
   const group = screen.getByTestId('group');
   expect(group).toHaveClass('flex-row');
 });
+
+test('applies width variable when set directly on an individual radio', () => {
+  render(<WithOwnWidth.Component />);
+
+  const radio = screen.getByRole('radio', { name: 'Option 1' });
+  // eslint-disable-next-line testing-library/no-node-access
+  const field = radio.closest('div[data-rac]') as HTMLElement;
```
sarahgm, line 61: missing blank lines in tests between AAA blocks

packages/components/src/Radio/Radio.tsx
```diff
@@ -55,17 +55,22 @@ const _Radio = ({
   ref,
   ...props
 }: RadioProps) => {
-  const { variant, size, width: groupWidth } = useRadioGroupContext();
+  const { variant, size } = useRadioGroupContext();
```
sarahgm, line 58: 💡 With `width: groupWidth` removed here, nothing reads `useRadioGroupContext().width` anymore, so the `width` value still provided at `RadioGroup.tsx:185` (and typed in `RadioGroupContextProps`) is now dead. Optional cleanup: drop `width` from the context value + interface. Harmless to keep if it's intended for future use.

packages/components/src/Radio/Radio.tsx
```diff
@@ -55,17 +55,22 @@ const _Radio = ({
...
+  // individual Radio only needs to compute its own CSS variable when it has
+  // its own `width` (i.e. used standalone, outside a Radio.Group). Reusing
+  // `groupWidth` here would resize against an already-resized container.
   return (
     <RadioField
-      className={cn(width || groupWidth || 'w-full')}
+      className={cn(width ? 'w-(--field-width) max-w-full min-w-0' : 'w-full')}
+      style={width ? createWidthVar('field-width', `${width}`) : undefined}
       value={value}
       isDisabled={disabled}
```
sarahgm, line 75: 💡 Behavior note (no change needed): an invalid `width` token on an individual `<Radio>` will now throw via `createWidthVar` (fail-fast) instead of silently emitting a broken class. This matches `FieldBase` and is the better behavior — just flagging it as a conscious trade-off.

sarahgm (commented):
## Automated Code Review

Recommend **approve**. Clean forward-port of #5659 that adopts the existing `w-(--field-width)` + `createWidthVar` pattern (matching `FieldBase`/`NumberField`), backed by correct tests and a proper changeset. CI is green; VRT isn't required for `beta-release`. Two optional cleanup notes inline — neither is a blocker.

---

packages/components/src/Radio/Radio.stories.tsx
```diff
@@ -88,6 +88,17 @@ export const Basic = meta.story({
   ),
 });
 
+export const WithOwnWidth = meta.story({
```
sarahgm, line 91: would it make sense to set `parameters: { chromatic: { disableSnapshot: true } }` ? 

Also this story doesn'T support `args`

packages/components/src/Radio/Radio.test.tsx
sebald: Done in 4b5a9ba — blank line before the assert block in both new tests. Left the pre-existing `takes full width by default` test as-is to keep the diff scoped.

packages/components/src/Radio/Radio.tsx
sebald: Good catch — confirmed dead and removed in 4b5a9ba.

Checked the blast radius first: there is no `Radio/index.ts`, `RadioGroupContext`/`RadioGroupContextProps` are not referenced anywhere outside `src/Radio/`, and the package index only exports `Radio`/`RadioGroup` and their prop types. So the context is internal and this is not an API change — no changeset needed.

Removed the value from the provider, the field from the interface, and the then-unused `WidthProp` import. `width` is still passed to `FieldBase` at `RadioGroup.tsx:167`, which is what actually sizes the group.

packages/components/src/Radio/Radio.tsx
sebald: Agreed — intended, and it matches `FieldBase`. Failing fast on an invalid token beats silently emitting a class Tailwind never generates, which is the bug this PR fixes in the first place. No change made.

packages/components/src/Radio/Radio.stories.tsx
sarahgm: since you updated the story at least the `args` should be overgiven

sebald: Fair — sorry, I missed this thread in the first round.

**`args`**: fixed in f10c8b8. The story now does `render: args => ... <Radio.Group {...args}>` instead of hardcoding `label="Label"`. `label` comes from the meta `args` default, so the rendering is byte-identical, but the controls (label, orientation, width, …) actually do something now.

**`chromatic: { disableSnapshot: true }`**: I'd rather keep the snapshot on. This story is now the only visual proof that a per-`Radio` `width` sizes the field — which is exactly the bug this PR fixes — so a baseline there earns its keep. The cost is one accept in Chromatic, since the tint is new. Happy to disable it if you'd prefer to avoid the churn, just say so.

sarahgm (approved):

PR #5666, opened 2026-07-22
feat(DST-1641): add ErrorState component and error states pattern

aromko:
# Description

Adds `ErrorState`, the error sibling of `EmptyState` from the DST-1496 convention: same anatomy (`title`, `description`, `action`, `headingLevel`), plus DOM passthrough (`role`, `tabIndex`, `ref`) so it works as an error boundary fallback. The boundary itself stays consumer-side.

Ships the docs to adopt it without reading the spike: a new ErrorState component page and an Error Boundaries pattern page covering the two tiers, the replaced-vs-annotated rule, retry microcopy, and the a11y contract (landmark, heading outline, focus vs `role="alert"`), with working react-error-boundary demos for both tiers. Cross-linked from Page, Panel, EmptyState, and SectionMessage.

The final error illustration is still with design, so the component ships with a placeholder (noted in code).

Closes DST-1641

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Storybook: open `Components/ErrorState`, story **Page Tier**. Press "Break this page", confirm the content is replaced by a focused ErrorState below the intact page header, then "Try again" restores it.
2. Docs (`pnpm start`): open Patterns → Feedback → Error Boundaries and exercise the page-tier and region-tier demos the same way.
3. Unit and story tests run in CI (`pnpm test:unit ErrorState`, `pnpm test:sb`).

## Breaking Changes

No

## Checklist

- [ ] Storybook preview and Marigold docs preview are available
- [x] Stories added/updated (with `component-test` tag where applicable)
- [x] Unit tests added/updated
- [x] Component documentation added/updated (if it exists)
- [ ] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) (for new/changed interactive components)
- [ ] Visual regression tests updated (for UI changes)
- [x] Changeset added (`pnpm changeset`)

PR #5675, opened 2026-07-24
feat(DST-1543): add `marigold migrate` codemods for breaking releases

aromko:
# Description

Adds `marigold migrate [version]` to `@marigold/cli`: data-driven codemods for breaking style and HTML-structure changes in consumer themes and application code.

The core design goal: **per-version work is data, not code**. The transforms in `lib/codemod/primitives/` are version-agnostic; everything v18-specific lives in `manifests/v18.ts` (slot sets, swap baselines, icon renames, token changes), so a future v19 migration is a new manifest, not new transform code. Codegen for those manifests plus the CI drift check is the follow-up ticket DST-1650.

What the v18 migration does:

- **Theme files** (anchored on `ThemeComponent<'X'>` with a verified `@marigold/system` import, file layout does not matter): restructures single-cva components to the new slot shapes (consumer classes move verbatim, never overridden), swaps baseline styles only on a byte-exact match with the old baseline (proof the slot was never customized, with a token diff report), stubs new slots as `cva({})`, and scaffolds required new components (`BooleanField`) next to the theme files that need them, including the barrel export.
- **Application code** (anchored on the `@marigold/components` / `@marigold/icons` import): the official icon renames (direct rename of import plus usages when provably safe, alias fallback otherwise), `Tabs.TabPanel` to `Tabs.Panel`, `Inset` spacing props, `acceptedFileType` with array wrap, `TextField` `min`/`max` removal.
- **Design tokens** (report-only, since token values are consumer property): renamed or removed tokens still referenced (`bg-brand`: use `bg-primary`), new tokens that component internals hardcode but the consumer CSS does not define (`SelectList` needs `selected-bold`/`disabled-surface`), and repurposed tokens that kept their name but changed meaning (`disabled` flipped background to text) with a remap recipe at the definition site.
- **Everything undecidable is a warning, never a guess**: spreads, customized slots, structural JSX moves, DOM changes that break consumer CSS selectors. Warnings link to pinned GitHub permalinks with line numbers.
- **UX**: the version parameter is optional (installed `@marigold/components` is detected and the migration proposed for confirmation), and interactive runs pre-analyze the target and offer the fired changes as a multiselect, with `--only <names>` as the non-interactive equivalent.

Validated against the portal repo: the full dry run edits 123 files, scaffolds `BooleanField`, and raises 50 warnings, all pointing at real definition sites (for example the `--color-disabled` role flip at their `tailwind.css:110`). Re-running is a no-op by design.

See `packages/cli/src/lib/codemod/README.md` for the architecture, invariants, and how to add a v19 manifest.

Closes DST-1543

## Test Instructions

All commands run from the repo root unless noted. The CLI runs from source via `tsx`, no build needed.

1. **Unit tests** (356, includes 14 token and 3 selection tests):

   ```sh
   pnpm --filter @marigold/cli exec vitest run
   ```
2. Download a project which is using marigold e.g. ACC, Reporting, Core etc.
3. Run node packages/cli/dist/bin/marigold.mjs migrate v18 <local path to project> --dry-run
4. Run it again without param --dry-run (changes will be made review them)

or what claude suggests...

2. **Create a small demo consumer** that triggers every codemod class:

   ```sh
   DEMO=$(mktemp -d)
   mkdir -p $DEMO/theme
   cat > $DEMO/theme/Card.styles.ts <<'EOF'
   import { cva, ThemeComponent } from '@marigold/system';
   export const Card: ThemeComponent<'Card'> = cva({ base: ['bg-white'] });
   EOF
   cat > $DEMO/theme/Switch.styles.ts <<'EOF'
   import { cva, ThemeComponent } from '@marigold/system';
   export const Switch: ThemeComponent<'Switch'> = {
     container: cva({ base: 'disabled:cursor-not-allowed disabled:text-disabled-foreground' }),
     track: cva({ base: 'flex h-6 w-10' }),
     thumb: cva({ base: 'block size-5' })
   };
   EOF
   printf "export * from './Card.styles';\nexport * from './Switch.styles';\n" > $DEMO/theme/index.ts
   cat > $DEMO/App.tsx <<'EOF'
   import { Inset, SelectList, Tooltip } from '@marigold/components';
   import { Pickup } from '@marigold/icons';
   export const App = () => (
     <Inset space={4}>
       <Pickup />
       <Tooltip open>hint</Tooltip>
       <SelectList aria-label="x" />
       <div className="bg-brand text-muted-foreground" />
     </Inset>
   );
   EOF
   printf ":root { --color-disabled: #eee; --color-disabled-foreground: #999; }\n" > $DEMO/tokens.css
   ```

5. **Dry run** and read the report:

   ```sh
   cd packages/cli
   npx tsx src/bin/marigold.ts migrate v18 $DEMO --dry-run
   ```

   Expect: Card restructure, Switch.container baseline swap, `Inset space` to `p` rename, `Pickup` to `Store` icon rename, warnings for `Tooltip open`, the `bg-brand`/`text-muted-foreground` token renames, the `SelectList` token dependency, the `--color-disabled` repurpose recipe, and 1 scaffolded file (`BooleanField.styles.ts`).

6. **Pre-analysis and selection**: run without `--dry-run` in your terminal. A multiselect lists the changes that fire, with counts. Everything is preselected, Enter applies all; deselect entries with space to skip them. Warnings always run.

   ```sh
   npx tsx src/bin/marigold.ts migrate v18 $DEMO
   ```

7. **Subset via flag** (non-interactive equivalent, also try a typo to see the error):

   ```sh
   npx tsx src/bin/marigold.ts migrate v18 $DEMO --only rename-imports,scaffold-components
   ```

8. **Idempotence**: run step 4 again and confirm `Edited 0 file(s)`.

9. **Version auto-detection** (needs a target with `@marigold/components` installed or declared, for example a real consumer checkout):

   ```sh
   npx tsx src/bin/marigold.ts migrate /path/to/consumer --dry-run
   ```

   Expect a confirm prompt naming the detected version and the proposed v18 migration. Piped runs fail with instructions to pass the version explicitly.

## Breaking Changes

No. New CLI subcommand only; existing commands are untouched (`doctor`'s `highlightCode` moved to `lib/format.ts`, same behavior).

## Checklist

- [ ] Storybook preview and Marigold docs preview are available (n/a, CLI only)
- [ ] Stories added/updated (n/a, CLI only)
- [x] Component documentation added/updated (`packages/cli/src/lib/codemod/README.md`)
- [ ] Accessibility reviewed (n/a, CLI only)
- [ ] Visual regression tests updated (n/a, CLI only)
- [x] Changeset added (`.changeset/migrate-codemods.md`, minor for `@marigold/cli`)


.changeset/iconography-docs.md
```diff
@@ -68,6 +68,7 @@ Old names from the legacy `@marigold/icons` set and their Lucide replacement. Na
 | Notification    | `MessageSquareWarning` |
 | Parking         | `CircleParking`        |
 | PDF             | `PDF` _(custom)_       |
+| Print           | `Printer`              |
```
sarahgm, line 71: why is this added?

packages/cli/src/lib/codemod/primitives/tokens.ts
```diff
@@ -0,0 +1,235 @@
...
+
+// Utility prefixes that take a color token (`bg-brand`). A curated list
+// keeps text scanning honest: matching bare `-brand` suffixes would also
+// hit variants and unrelated identifiers.
+// ponytail: covers the color utilities Marigold themes actually use; extend
+// the list when a consumer surfaces one we missed.
+const COLOR_PREFIXES =
+  'bg|text|border|ring|inset-ring|outline|fill|stroke|decoration|divide|accent|caret|shadow|from|via|to|placeholder';
+
+const escapeRegExp = (name: string): string =>
```
sarahgm: 🪣 **reuse** — This `escapeRegExp` is byte-for-byte identical to the private `escapeRegex` in `lib/edit-css.ts:37` (same char-class, same body). This PR already set the precedent of hoisting shared helpers (`highlightCode` → `format.ts`, exported `insertImport`); do the same here — export one helper from a shared module and import it in both `tokens.ts` and `edit-css.ts`.

packages/cli/src/bin/marigold.ts
```diff
@@ -419,6 +440,111 @@ export const main = async (
...
+        ?.split(',')
+        .map(s => s.trim())
+        .filter(Boolean);
+
+      for (const version of versions) {
+        // pre-analysis: dry-run in memory, offer the fired changes for
+        // selection. Skipped for explicit --only / --dry-run / non-TTY runs.
+        let selected = only;
+        if (!selected && !values['dry-run'] && process.stdout.isTTY) {
+          const analysis = runMigrate({ version, targetPath, dryRun: true });
```
sarahgm: 🪣 **efficiency** — On the default interactive path (TTY, no `--only`/`--dry-run`) the whole migration runs twice: this dry-run pre-analysis, then the real `runMigrate` below. `collectSourceFiles`, every `readFileSync`, the `definedTokens` CSS scan, and the pass-1 inventory parse are identical across both calls — only the edit set differs by the user's selection. Consider computing `texts`/`definedTokens`/inventory once and threading them into both phases (or have the analysis return the per-file outputs so the apply step just writes the selected subset).

packages/cli/src/lib/codemod/primitives/jsx.ts
```diff
@@ -0,0 +1,529 @@
...
+      if (imported && local && wanted.has(imported)) {
+        locals.set(local, imported);
+      }
+    }
+  }
+  return locals;
+};
+
+/** Renames props on Marigold components, e.g. Inset `space` to `p`. */
+export const renameJsxProps = (manifest: MigrationManifest): Codemod => ({
```
sarahgm: 💡 **simplification/reuse** — `renameJsxProps`, `removeJsxProps` (l.449) and `reportJsxUsage` (l.488) repeat the same skeleton: `localsFor` guard → `for el of jsxElements` → resolve component → match entry → match attr → edited/unchanged epilogue. The theme side abstracts exactly this with the `themeCodemod` factory; the JSX side has no equivalent. A `jsxAttrCodemod`/`forEachMarigoldAttr` factory would remove the copy-paste, let `localsFor` (l.36) build on `marigoldLocalName` instead of re-implementing its import resolution, and naturally add the `if (!source.includes('@marigold/components')) return unchanged` short-circuit these four skip today (they full-parse every theme-only/icons-only file just to bail on `locals.size === 0`).

packages/cli/src/lib/codemod/primitives/report.ts
```diff
@@ -0,0 +1,52 @@
+import { findThemeComponents } from '../anchor.js';
+import { parseOr } from '../engine.js';
+import type { Codemod, CodemodOutcome, MigrationManifest } from '../types.js';
+
+const analyze = (
```
sarahgm: 💡 **reuse/altitude** — `analyze` hand-rolls the exact frame `engine.ts`'s `themeCodemod` already owns: the `includes('ThemeComponent')` pre-check, `parseOr`, the `findThemeComponents` iteration, and the uniform `unchanged`+warnings outcome. Since `themeCodemod` already returns `unchanged` (carrying warnings) when no changes are pushed, `reportDeadKeys`/`reportStructure` are just warning-only visits — giving `themeCodemod` a report-only path would delete `analyze` and keep report and edit primitives on one shared frame.

packages/cli/src/commands/migrate.ts
```diff
@@ -0,0 +1,482 @@
...
+    }
+    if (edited(report) && !options.dryRun) {
+      writeFileSync(file, report.output);
+    }
+  }
+
+  // pass 2.5: token references live anywhere, not only in @marigold
+  // importers (own token CSS, generated CSS, plain components) — text-scan
+  // the files the pipeline did not see.
+  for (const [file, text] of texts) {
```
sarahgm, line 311: 💡 **altitude** — `reportTokenUsage` genuinely applies to *every* file, but the main pipeline's file set is filtered to `@marigold/`-importers, so its true scope is bolted back on here as a one-off "pass 2.5" over the leftovers. Coverage now depends on two code paths staying jointly exhaustive — any future tightening of the `sources` filter silently narrows token coverage. Cleaner: partition primitives by scope (`@marigold`-anchored vs whole-tree text-scan) and drive each set over its proper file collection once, dropping `reportTokenUsage` from the anchored pipeline entirely.

packages/cli/src/commands/migrate.ts
```diff
@@ -0,0 +1,482 @@
...
+  /** changes that fired on this target, in pipeline order — the
+   *  pre-analysis a caller can offer for selection via `only` */
+  summary: CodemodSummary[];
+}
+
+const SCAFFOLD = 'scaffold-components';
+
+// user-facing one-liners for the selectable changes (report-only passes are
+// not selectable: they are the safety net and always run)
+const DESCRIPTIONS: Record<string, string> = {
```
sarahgm: 💡 **altitude** — A codemod's human description lives in this runner-side lookup keyed by its `name` string — a third touchpoint alongside the primitive's `name` and the pipeline wiring. A rename or typo degrades silently to `''` via `DESCRIPTIONS[name] ?? ''` (l.409) with no error. Consider a `description` field on the `Codemod` interface (populated by each factory) so name/description/apply travel together and can't drift.

packages/cli/src/commands/migrate.ts
```diff
@@ -0,0 +1,482 @@
...
+    changes: [],
+    warnings: [],
+    skips: [],
+    output: source,
+    perCodemod: {},
+  };
+  for (const codemod of codemods) {
+    // each transform re-parses its input when chained; babel parse is fast
+    // and shared-AST plumbing is not worth it
+    const outcome = codemod.apply(report.output);
```
sarahgm, line 186: 💡 **efficiency (low priority / deferrable)** — Each of the ~7–13 pipeline codemods independently `parseTsx`-es the file even though the text is unchanged between all but the 0–1 codemods that edit it (plus the pass-1 inventory parse at l.256 that's discarded). Parsing once per file and re-parsing only after an `edited` outcome would collapse ~7–13 parses to 1–2 across the whole consumer tree. The inline comment calls this "not worth it" — reasonable for a one-shot CLI, but the multiplier is larger than it reads. The `includes` short-circuit (see the jsx.ts comment) and the double-`runMigrate` fix are the cheaper subsets worth doing first.

packages/cli/src/lib/codemod/engine.ts
```diff
@@ -0,0 +1,171 @@
...
+ * Class-string literals of a cva() argument, in source order — skipping the
+ * `defaultVariants` subtree, whose values are variant names (`'default'`),
+ * not classes.
+ */
+export const classStringsIn = (node: AnyNode): string[] => {
+  const out: string[] = [];
+  walk(node, n => {
+    if (
+      n.type === 'ObjectProperty' &&
+      (n.key as AnyNode | undefined as { name?: string } | undefined)?.name ===
```
sarahgm: 💡 **simplification (minor)** — `(x as AnyNode | undefined as { name?: string } | undefined)?.name` — the intermediate `as AnyNode | undefined` is dead weight, since `AnyNode`'s index signature already yields `unknown`. `(x as { name?: string } | undefined)?.name` compiles identically. Same idiom in `primitives/jsx.ts:32` (`attrName`).

packages/cli/src/lib/codemod/primitives/jsx.ts
```diff
@@ -0,0 +1,529 @@
...
+      return ((parent.params as AnyNode[]) ?? []).includes(n);
+    case 'CatchClause':
+      return parent.param === n;
+    // destructuring targets and function params inside patterns
+    case 'ObjectPattern':
+    case 'ArrayPattern':
+    case 'RestElement':
+    case 'AssignmentPattern':
+      return true;
+    default:
```
sarahgm, line 203: 💡 **dead code (behaviour-neutral)** — This `default` arm re-tests `FunctionDeclaration`/`FunctionExpression`, but both are already handled by the `case` labels at l.157–158 that `return parent.id === n`, so this `if` is unreachable. `default: return false;` is equivalent. Verified this is *not* a correctness gap: a plain function-param shadow of a renamed import produces a correct uniform alpha-rename, and target collisions are caught by `namesInFile`. (If you *did* want function params handled distinctly — e.g. to prefer the alias fallback — they'd need adding to the `ArrowFunctionExpression`-style `params.includes(n)` case, not the dead default.)

sarahgm (commented):
## Automated review — `/simplify` + `/review-pr` (DST-1543)

Solid, well-documented PR with an excellent data-driven architecture (generic engine, all component specificity as manifest data) and strong safety discipline. Nothing here blocks merge — all findings are quality/reuse refinements.

One note on verification: my first pass suspected a latent rename bug from the unreachable `default` branch in `isBindingPosition`, but I tested it (plain function-param shadow of a renamed icon) and the output is a correct uniform alpha-rename, with target-name collisions caught independently by `namesInFile`. So that item is demoted to a behaviour-neutral dead-code cleanup, not a bug.

_Generated with Claude Code (review-pr + simplify skills)._

packages/cli/src/commands/migrate.test.ts
```diff
@@ -0,0 +1,403 @@
...
+      );
+    }
+    // the migration target usually sits far below the repo root
+    const target = path.join(root, 'src', 'theme');
+    mkdirSync(target, { recursive: true });
+    return target;
+  };
+
+  test('walks up to node_modules and proposes the applicable migration', () => {
+    const detected = detectMigration(setupRepo('17.9.1'));
```
sarahgm, line 82: missing blank lines between the blocks (in all tests)

packages/cli/src/lib/codemod/manifests/v18.ts
```diff
@@ -0,0 +1,687 @@
...
+      'dropZoneLabel',
+      'item',
+      'itemLabel',
+      'itemDescription',
+      'itemRemove',
+    ],
+    EmptyState: ['container', 'title', 'description', 'action'],
+    ToggleButton: ['group', 'button'],
+    SegmentedControl: ['group', 'list', 'field', 'option', 'indicator'],
+    Sidebar: [
```
sebald, line 239: 🔴 **Manifest drift: `Sidebar` is missing the 9 rail slots** added in #5654 (DST-1609): `railRoot`, `railLayout`, `railColumn`, `railToggle`, `rail`, `railItem`, `railFooter`, `panel`, `panelTitle`.

Verified programmatically against `packages/system/src/types/theme.ts` on this branch, and reproduced in a live run: `stub-missing-slots` stubs only the 12 missing slots from this list, so a migrated consumer still fails the exhaustive-Record typecheck on the rail slots — with no codemod assist and no warning.

The branch merged `beta-release` after this file was written, so the code saw the change but the hand-written data didn't — exactly the drift DST-1650's codegen will prevent. Until then, suggest re-syncing this entry and treating every `beta-release` merge as a manifest re-check trigger.

packages/cli/src/lib/codemod/manifests/v18.ts
```diff
@@ -0,0 +1,687 @@
...
+      'container',
+      'dropZone',
+      'dropZoneContent',
+      'dropZoneLabel',
+      'item',
+      'itemLabel',
+      'itemDescription',
+      'itemRemove',
+    ],
+    EmptyState: ['container', 'title', 'description', 'action'],
```
sebald, line 235: 🔴 **Manifest drift: `ErrorState` is missing from `slots`** — added on `beta-release` in #5666 (DST-1641) with slots `container | title | description | action`.

Because it's absent here, `report-dead-keys` emits a **false warning** for a themed ErrorState (reproduced locally):

```
ErrorState: not a themeable component in v18 — these styles are silently unused
```

…which is actively misleading, since ErrorState *is* themeable in v18. Same root cause as the Sidebar comment above.

```suggestion
    EmptyState: ['container', 'title', 'description', 'action'],
    ErrorState: ['container', 'title', 'description', 'action'],
```

packages/cli/src/lib/codemod/primitives/tokens.ts
```diff
@@ -0,0 +1,235 @@
...
+// meaning is exactly the problem.
+
+/** token names defined (`--color-x:`) in a chunk of CSS */
+export const definedTokensIn = (css: string): string[] =>
+  [...css.matchAll(/--color-([\w-]+)\s*:/g)].map(m => m[1]);
+
+// Utility prefixes that take a color token (`bg-brand`). A curated list
+// keeps text scanning honest: matching bare `-brand` suffixes would also
+// hit variants and unrelated identifiers.
+// ponytail: covers the color utilities Marigold themes actually use; extend
```
sebald: ⚠️ What does the `ponytail:` marker mean? It reads like a personal TODO convention — it also appears in `swap-exact-classes.ts:100`. Suggest `TODO:`/`Note:` (greppable, self-explanatory) or dropping the prefix.

packages/cli/src/lib/codemod/primitives/swap-exact-classes.ts
```diff
@@ -0,0 +1,116 @@
...
+          arg.end as number,
+          reindent(entry.newSource, unit, base)
+        );
+        changes.push(
+          `${component}.${entry.slot}: swapped baseline styles to the ${manifest.version} baseline`
+        );
+
+        // Render the change as a token diff (the report colorizes -/+ lines)
+        // so renamed tokens and new layout utilities are visible at a glance.
+        // ponytail: added utilities are flagged for manual verification; the
```
sebald: ⚠️ Same `ponytail:` marker as in `tokens.ts:27` — see the comment there.

packages/cli/src/lib/codemod/engine.ts
```diff
@@ -0,0 +1,171 @@
...
+  );
+  return 'cva';
+};
+
+/** render a list of names as `code` spans for the report */
+export const codeList = (names: string[]): string =>
+  names.map(name => `\`${name}\``).join(', ');
+
+/** the empty-stub property line shared by restructure and stubbing */
+export const stubSlotLine = (slot: string, indent: string): string =>
```
sebald: ⚠️ **`stubSlotLine` hardcodes `cva`, but the rest of the system is alias-aware.** `ensureCvaImport` correctly resolves `import { cva as c } from '@marigold/system'` and returns the local name — but `themeCodemod` ignores the return value, and stubs always emit `cva({})`.

An aliasing consumer therefore gets stubs referencing an undefined `cva` *and* no import inserted (`ensureCvaImport` sees the aliased import and bails). Edge case — the consumer's typecheck catches it — but the fix is small: resolve the local name once per file and thread it into `stubSlotLine`/`generateScaffold`'s stub rendering.

packages/cli/src/bin/marigold.ts
```diff
@@ -419,6 +440,111 @@ export const main = async (
...
+      const only = values.only
+        ?.split(',')
+        .map(s => s.trim())
+        .filter(Boolean);
+
+      for (const version of versions) {
+        // pre-analysis: dry-run in memory, offer the fired changes for
+        // selection. Skipped for explicit --only / --dry-run / non-TTY runs.
+        let selected = only;
+        if (!selected && !values['dry-run'] && process.stdout.isTTY) {
```
sebald: ⚠️ **Interactivity gate is weaker than `init`'s.** This (and the confirm gate at L490) checks only `process.stdout.isTTY`; `init.ts:283` checks `process.stdout.isTTY && process.stdin.isTTY`. With stdin piped but stdout a TTY, the clack prompt can hang waiting for input that never arrives. Suggest matching init's check.

packages/cli/src/lib/codemod/manifests/v18.ts
```diff
@@ -0,0 +1,687 @@
...
+        to: 'Angry',
+      },
+      { package: '@marigold/icons', from: 'SmilieVerySatisfied', to: 'Laugh' },
+      { package: '@marigold/icons', from: 'User', to: 'UserRound' },
+      { package: '@marigold/icons', from: 'Share', to: 'Share2' },
+      {
+        package: '@marigold/icons',
+        from: 'Print',
+        to: 'Printer',
+        note: 'missing from the official mapping table — Printer is the Lucide equivalent',
```
sebald: ⚠️ This note is now self-contradicting: this same PR adds `Print → Printer` to the official mapping table in `.changeset/iconography-docs.md`, so it's no longer "missing from the official mapping table". Drop the note (or reword to "added to the table in this PR").

packages/cli/src/bin/marigold.ts
```diff
@@ -419,6 +440,111 @@ export const main = async (
 
       writeOutput(result.output);
       if (result.hasErrors) exitCode = 1;
+    } else if (command === 'migrate') {
```
sebald, line 443: 💡 **Consider moving the interactive flow into `commands/migrate.ts`.** This block is ~110 lines including detection, confirm, and multiselect orchestration; `init` keeps its prompt flow in `commands/init.ts` and the router thin. Relocating would match that convention and make the interactive path unit-testable — right now only `detectMigration`/`runMigrate` have tests; the confirm/multiselect flow has none.

packages/cli/src/commands/migrate.ts
```diff
@@ -0,0 +1,482 @@
...
+  })();
+  if (!found) return null;
+
+  // Propose migrations targeting the installed major too (>=, not >): the
+  // documented procedure is "upgrade the packages, then migrate", so someone
+  // on 18.x is exactly who needs the v18 migration. Re-running is a no-op.
+  const major = Number.parseInt(found.installed, 10);
+  const versions = Object.keys(MANIFESTS)
+    .map(v => ({ v, target: Number.parseInt(v.replace(/^v/, ''), 10) }))
+    .filter(({ target }) => Number.isFinite(target) && target >= major)
```
sebald, line 90: 💡 **`target >= major` proposes the v18 migration to consumers still on v17.** The comment justifies `>=` for the same-major case ("upgrade first, then migrate" → someone on 18.x needs v18), but the filter also fires for `major = 17` — i.e. someone who has *not* upgraded yet. Running the theme migration while still on v17 leaves the theme mismatched until they upgrade.

Worth a deliberate decision: either detect the not-yet-upgraded case and say "upgrade `@marigold/components` first", or keep the behavior and state it in the confirm prompt.

packages/cli/src/bin/marigold.ts
```diff
@@ -419,6 +440,111 @@ export const main = async (
...
+      const { positionals, values } = parseMigrateCommand(rest);
+      // the version positional is optional: `migrate ./src` treats the first
+      // positional as a path, `migrate v18 ./src` as version + path
+      const looksLikeVersion = (p: string | undefined): p is string =>
```
sebald: 💡 `marigold migrate 18.1 .` silently treats `18.1` as a *path* (the regex only matches `v?\d+`) and dies with a raw filesystem error. A cheap nudge: if the first positional looks version-ish (`/^v?\d+[.\d]*$/`) but isn't an exact match, fail with `did you mean v18?`.

packages/cli/src/lib/codemod/primitives/restructure-to-slots.ts
```diff
@@ -0,0 +1,55 @@
...
+      }
+
+      const base = lineIndentAt(source, start);
+      const inner = base + unit;
+      const stubs = slots
+        .filter(slot => slot !== primary)
+        .map(slot => stubSlotLine(slot, inner))
+        .join('\n');
+      const original = source.slice(start, end);
+      s.overwrite(
```
sebald, line 44: 💡 Cosmetic, observed in the live run: the moved cva body keeps its original indentation, so after wrapping it sits under-indented one level relative to its new nesting (`container: cva({` at depth 1, but `base:` still at the old depth). "Moved verbatim" is by design and the consumer's Prettier fixes it — but consider running the moved span through `reindent`, or noting the behavior in the README so nobody files it as a bug.

packages/cli/src/lib/codemod/README.md
```diff
@@ -0,0 +1,254 @@
...
+4. **Idempotent.** Re-running a migration is a no-op; already-swapped slots
+   are recognized as being on the target baseline.
+5. **The Theme contract is a stable API of this system.** The primitives and
+   the manifest format assume optional component keys, exhaustive slot
+   Records, and cva-based style functions. Changing that contract in
+   `@marigold/system` is itself a breaking change: bump the manifest
+   `schemaVersion`, keep the old interpreter so chained migrations
+   (v17 to v19) still work.
+
+## Adding a new migration (e.g. v19)
```
sebald, line 183: 💡 **Suggestion: turn this section into an agent skill that authors `manifests/v*.ts`.** The provenance rules documented here (slots from the `Theme` type, `oldClasses` from the previous major's published theme-rui, tokens from the release-notes tables / CSS diff, permalinks pinned to a tag) are already a mechanical, step-by-step recipe — exactly the shape of a good skill.

It would bridge the gap until DST-1650's codegen lands, and this PR contains the motivating example: the `Sidebar`/`ErrorState` drift (see the manifest comments) came from `beta-release` merges after the manifest was hand-written — a skill-driven re-check on merge would have caught it.

sebald (changes requested):
## Code Review — local deep-dive + hands-on testing

Reviewed in depth at `064046d` and re-checked against `ee83609` (merge-only, no `packages/cli` changes). Findings are inline; summary here.

**Overall: this is excellent work.** The data-vs-code split (version-agnostic `primitives/`, per-version `manifests/`), the warn-never-guess discipline, the byte-preserving edits, and the `codemod/README.md` are all above bar. The CLI fit is exemplary — existing helpers (`tsx-ast` walk, `edit-tsx` insertImport, doctor's package-json/version utils, `highlightCode`) are reused and extended instead of duplicated, and every integration point (help text, `commands-spec`, telemetry, lazy imports) is updated coherently. Test quality is high: fixture-based, byte-preservation and idempotency asserted, and the hairy `rename-imports` edge cases (shadowing, shorthand props, re-exports, alias collisions) are covered.

**Hands-on verification** (demo consumer exercising every codemod class, CLI built and run from this branch):

- ✅ 363 unit tests pass; typecheck + lint clean
- ✅ Dry-run report: every documented change/warning class fired correctly (restructure, byte-exact swap with `-/+` token diff, prop/member/import renames, removals, structure/token/namespace warnings, repurpose recipe at the definition site, scaffold announcement)
- ✅ Applied run: consumer class strings survive byte-for-byte, `newSource` correctly re-indented to the consumer's 4-space style, all outputs re-parse, `BooleanField.styles.ts` scaffold + barrel export correct
- ✅ Icon renames: direct rename (import + usages), alias fallback with stated reason (`Trash2 as Delete` because re-exported), re-export keeps the public name (`export { Store as Pickup }`)
- ✅ Idempotent (`Edited 0 file(s)` on re-run; the swap's token-diff warning correctly disappears once applied)
- ✅ `--only` subset applies exactly the named changes; typo → helpful error listing valid names, exit 1
- ✅ Non-TTY: auto-detection fails with actionable instructions; interactive confirm + multiselect verified via pty

**Why "request changes" despite all green:** the two inline 🔴 findings — the hand-written v18 manifest has drifted behind `beta-release` (`Sidebar` rail slots from #5654, `ErrorState` from #5666). Both confirmed empirically: the rail slots are not stubbed, and a themed `ErrorState` gets a false *"not a themeable component in v18"* warning. The product here *is* the manifest data, so this is the one thing worth blocking on. Everything else is quick fixes / follow-up material.

packages/cli/src/lib/codemod/manifests/v18.ts
aromko: Fixed in afec44809: all 9 rail slots added.

I also checked whether the drift was wider than the two you found. It was not, exactly `ErrorState` plus these 9 (64 components in the `Theme` type, 63 in the manifest).

To make it not depend on someone remembering: `manifests/theme-drift.test.ts` now parses `packages/system/src/types/theme.ts` and asserts slot parity with the manifest. I verified it bites by removing `ErrorState` and one rail slot again, which fails with `expected [ 'ErrorState' ]` and `Sidebar: missing [railFooter]`. The codemod README names it as a step when authoring a new manifest, so DST-1650's codegen inherits the check rather than replacing it.

packages/cli/src/lib/codemod/manifests/v18.ts
aromko: Fixed in afec44809, applied your suggestion. Confirmed in a live dry run: a themed `ErrorState` now produces no warning at all instead of the false "not a themeable component in v18".

Same drift test as on the Sidebar thread covers this one, so the next component merged from `beta-release` fails CI instead of silently mis-warning.

packages/cli/src/lib/codemod/primitives/tokens.ts
aromko: Fair, that is a personal convention leaking into the repo. Both markers are `Note:` now (afec44809).

packages/cli/src/lib/codemod/primitives/swap-exact-classes.ts
aromko: Same, changed to `Note:` in afec44809.

packages/cli/src/lib/codemod/engine.ts
aromko: Fixed in afec44809, smaller than you sized it: `themeCodemod` resolves the local name once per file into the visit context, and `stubSlotLine` takes it. `generateScaffold` writes its own import header into a brand-new file, so it never sees a consumer alias and did not need threading.

Verified against a consumer importing `cva as c`:

```ts
export const Card: ThemeComponent<'Card'> = {
  container: c({ base: ['bg-white'] }),
  header: c({}),
  ...
};
```

No duplicate import inserted, and re-running is still a no-op.

packages/cli/src/bin/marigold.ts
aromko: Fixed in afec44809. Both gates now check stdout AND stdin, same as `init.ts:283`. The check moved with the flow into `commands/migrate.ts` (see your other comment) and is overridable in tests.

packages/cli/src/lib/codemod/manifests/v18.ts
aromko: Reworded in afec44809 to "added to the official mapping table in this PR". See also sarahgm's question on the changeset itself, which is the other half of this.

packages/cli/src/bin/marigold.ts
aromko: Done in afec44809. `runMigrateCommand` now lives in `commands/migrate.ts` and owns detection, the confirm, the pre-analysis and the multiselect; the router keeps only argument parsing, telemetry and the lazy import. It takes a `write` callback and an `interactive` override so the whole flow is drivable from a test.

Coverage that did not exist before: explicit version runs without prompting, the non-TTY refusal, the up-to-date report, the missing-Marigold failure, a declined confirm returning 130, and a multiselect subset leaving the deselected files byte-identical.

packages/cli/src/bin/marigold.ts
aromko: Added in afec44809: `18.1`, `v18.2` and `18.0.0` all fail with `Unknown migration '18.1' — migrations are named by major version. Did you mean v18?`.

Worth flagging that my first attempt at this did not work: the positional-count check ran first, so `migrate 18.1 ./src` still died with the generic usage error. Caught it in manual testing, moved the hint above the count check, and pinned the ordering with a test in `bin/marigold.test.ts` so it cannot silently regress.

packages/cli/src/commands/migrate.ts
aromko: Deliberately unchanged for now, and I would like your call on it.

The reason I lean toward keeping `>=`: migrating the theme before bumping the package is a legitimate order, since the theme edits are what makes the app compile after the bump. Someone on 17.x who runs this gets a theme shaped for v18 and a typecheck that tells them exactly what is left, which is the same end state as migrating after the bump. Blocking them would force the sequence "bump, break everything, then migrate".

What is genuinely missing is that we do not say any of this. My proposal is to keep the behavior and add a line to the confirm prompt when the installed major is below the target, along the lines of "you are on 17.9.1; this migrates your theme to the v18 shape, bump `@marigold/components` to v18 to complete the upgrade". One string, no behavior change.

Say the word and I will add it, or flip it to a hard "upgrade first" if you would rather not have a half-migrated state exist at all.

packages/cli/src/lib/codemod/primitives/restructure-to-slots.ts
aromko: Not changed, and I would rather document it than fix it.

Running the moved span through `reindent` would mean rewriting bytes inside code we promised to move verbatim. That span can contain template literals and multi-line strings where leading whitespace is significant, so the transform would have to start reasoning about what is safe to re-indent. That is a real risk for a cosmetic gain the consumer's Prettier already covers.

Proposing a line in the codemod README's invariants section stating that the moved expression keeps its original indentation by design. Not in afec44809 yet; say if you would rather have the `reindent` and I will do it instead.

packages/cli/src/lib/codemod/README.md
aromko: Agreed, and worth its own ticket rather than growing this PR. I will file one linked to DST-1650 so the skill and the codegen land as one story.

One correction on the motivating example though: the `Sidebar`/`ErrorState` drift is now caught mechanically rather than by a skill-driven re-check. `manifests/theme-drift.test.ts` parses the live `Theme` type and fails CI on any slot mismatch, and the README section you are commenting on names it as step 3 when authoring a manifest. That covers the one part of the recipe a machine can verify; the parts the skill would own (oldClasses from the previous major's published theme, token tables from the release notes, pinned permalinks) stay unguarded, which is exactly where it earns its keep.

.changeset/iconography-docs.md
aromko: The changeset itself is not new, it came from `beta-release`. This branch adds one row to it in 09e60574c: `| Print | Printer |`.

v17 shipped a custom `Print` icon (`info/Print`) that the migration table had no row for. I found the gap while generating the icon-rename codemod, because that table is the source of truth the manifest's `importRenames` are derived from: no row means no rename, and consumers keep a dead import.

Happy to pull it out into its own PR against `beta-release` if you would rather keep this one CLI-only. It is one line either way, so it is purely a question of where you want it reviewed.

packages/cli/src/lib/codemod/primitives/tokens.ts
aromko: Done in afec44809. Hoisted to `lib/regex.ts`, imported by both `edit-css.ts` and `codemod/primitives/tokens.ts`. One-function module, same shape as the existing `lib/strip-ansi.ts`, so it follows a precedent rather than inventing a home.

packages/cli/src/lib/codemod/primitives/report.ts
aromko: Done in afec44809, and it needed no engine change at all: `themeCodemod` already returns `unchanged` carrying warnings when a visit pushes no changes, so both report primitives are now plain `themeCodemod` calls and `analyze` is deleted. Documented that property on `themeCodemod` so the next report-only pass does not hand-roll the frame again.

packages/cli/src/lib/codemod/engine.ts
aromko: Both dropped in afec44809.

packages/cli/src/lib/codemod/primitives/jsx.ts
aromko: Done in afec44809, `default: return false`. Kept your finding as the comment on the arm, since "plain function params are deliberately absent" is exactly the thing the next reader will want to re-litigate:

```ts
// A plain function param (`function f(Pickup) {}`) is deliberately not
// listed: it produces a correct uniform alpha-rename, and a collision
// with the target name is caught by `namesInFile` independently.
default:
  return false;
```

Thanks for testing it before reporting rather than after.

packages/cli/src/commands/migrate.ts
aromko: Done in afec44809, and I went further than optional: `description` is a **required** field on `Codemod`, so the map is gone entirely and a typo or rename is a compile error rather than an empty hint. Every codemod carries one, including the report-only passes that are not selectable, because a required field cannot be forgotten and the descriptions document the pipeline anyway.

`scaffold-components` is the one exception, since it creates files rather than editing one and is not a `Codemod`. Its name and description now sit adjacent as two constants, with no lookup between them.

packages/cli/src/commands/migrate.test.ts
aromko: Done in afec44809, across `migrate.test.ts`, `codemod.test.ts`, `jsx.test.ts` and `tokens.test.ts`. The new tests follow it too.

packages/cli/src/lib/codemod/primitives/jsx.ts
aromko: Took the short-circuit and a narrower frame in afec44809.

`jsxCodemod` now wraps `renameJsxProps`, `renameJsxMembers`, `removeJsxProps` and `reportJsxUsage`: text check for `@marigold/components` before any parse, then `parseOr`, then `localsFor`, then skip when nothing is imported. `renameImports`, `reportNamespaceImports` and `reportTokenDependencies` cannot use it (they anchor on `@marigold/icons` and friends) so they got the equivalent guard against their own package list. So every primitive in the pipeline now skips theme-only and unrelated files before parsing rather than after.

What I did not do is push the attribute loop itself into the factory. `renameJsxMembers` walks member expressions rather than attributes and would not fit, and for the other three the loops iterate manifest entries then attributes; inverting that to attributes-then-entries changes the order changes land in the report for an element with two affected props. Real risk, no real payoff, so the shared part stops at the frame.

packages/cli/src/bin/marigold.ts
aromko: Real, and I measured it before deciding: a full dry run over this repo's entire `packages/` tree is ~0.6s, so the pre-analysis costs about that much on the default interactive path.

Threading the computed per-file outputs through selection is a genuine restructure of `runMigrate`'s return contract, and the payoff is half a second in an interactive command the user is already reading a prompt in. Deferring, and taking the cheap subset you named instead: every primitive now short-circuits before parsing on files it cannot match (see the jsx.ts thread), which shrinks both passes.

If a real consumer turns out to be much slower than 0.6s I would rather revisit it with that number than guess now.

packages/cli/src/commands/migrate.ts
aromko: Deferring, as you suggested. The `includes` short-circuits from your jsx.ts comment are in (afec44809) and they take a decent bite out of this: theme-only files no longer get parsed by the four JSX primitives, the two import-anchored ones or the token-dependency pass, and `themeCodemod` already skipped non-theme files. What is left is genuinely the chained-transform re-parse, which needs the shared-AST plumbing to fix.

packages/cli/src/commands/migrate.ts
aromko: Pushing back on this one, though the readability point stands.

The failure mode you describe cannot happen as written: pass 2.5 iterates `texts` and skips exactly `sources.has(file)`, so the two sets are complements by construction. Tightening the `sources` filter moves files from pass 2 into pass 2.5 and `reportTokenUsage` still sees them; it cannot narrow coverage, only shift which pass provides it.

The placement is also load-bearing rather than incidental. Inside the pipeline the token scan runs on the post-edit text, so it sees classes `swap-exact-classes` just introduced. Partitioning primitives by scope and driving the whole-tree set over the original `texts` would lose that, and swapped-in baselines referencing tokens a standalone theme lacks is precisely the case the scan exists for.

So: keeping it, with a comment stating both properties. If the two-loop shape still reads wrong to you after that, I would rather solve it by making the complement explicit in one loop than by splitting the primitive sets.

aromko:
## Review feedback addressed (afec44809, 4446028f7)

Thanks both, the two 🔴 findings were real and the reuse comments were worth taking. Every thread has an individual reply; summary here.

### Blocking: manifest drift

`ErrorState` and `Sidebar`'s 9 rail slots are in. I also checked whether the drift went further than the two you found: it did not, exactly those (64 components in the `Theme` type, 63 in the manifest).

More importantly it is no longer a "remember to re-check on merge" problem. `manifests/theme-drift.test.ts` parses the live `Theme` type in `@marigold/system` and asserts slot parity, verified to fail on both drift kinds. The codemod README names it as a step when authoring a manifest, so DST-1650's codegen inherits the guard instead of replacing it.

### Fixed

- **Alias-aware `cva` stubs** (sebald): `themeCodemod` resolves the local name once per file, `stubSlotLine` uses it. A consumer importing `cva as c` gets `c({})`, verified end to end.
- **Interactivity gate** matches `init`'s (stdout AND stdin).
- **`migrate 18.1`** now suggests `v18`. My first cut of this was unreachable behind the positional-count check; caught it in manual testing and pinned the ordering with a test.
- **Interactive flow moved** into `commands/migrate.ts`, router stays thin. It had zero coverage and now has 7 tests, including the declined confirm returning 130 and a multiselect subset leaving deselected files byte-identical.
- **`description` is a required field on `Codemod`** (sarahgm): the runner-side lookup and its silent `?? ''` are gone, so a rename is a compile error.
- **`report.ts`'s `analyze` folded into `themeCodemod`**, which already returns `unchanged` carrying warnings. No engine change needed.
- **`jsxCodemod` frame + package guards**: every primitive now skips files it cannot match before parsing rather than after.
- **`escapeRegex` hoisted** to `lib/regex.ts`, dead `default` arm dropped, redundant casts dropped, `ponytail:` markers renamed to `Note:`, AAA blank lines in the codemod tests.
- **Docs**: `marigold migrate` documented on `/getting-started/cli` (sarahgm's Vercel comment), with a `@marigold/docs` changeset.

### Deliberately not changed, reasons on the threads

- **Double `runMigrate`** on the interactive path: measured at ~0.6s for a full dry run over this repo's `packages/` tree. Took the cheap subset (short-circuits) instead.
- **Per-file re-parse**: same, largely mitigated by the short-circuits.
- **Token scan in two passes**: the sets are exact complements by construction, and the in-pipeline placement is load-bearing (it runs on post-edit text, so it sees classes the swap introduced). Now stated in a comment.
- **Restructure indentation**: proposing a README note over re-indenting a span we promised to move verbatim.
- **Manifest-authoring skill**: filing a follow-up linked to DST-1650.

### Open question for @sebald

The `target >= major` behavior for v17 consumers. I lean toward keeping it and adding one line to the confirm prompt when the installed major is below the target, rather than blocking. Reasoning on that thread, happy to go either way.

Verification: 377 CLI tests pass, repo `tsc` and `eslint` clean, and I re-ran the full dry run + apply + re-run against a demo consumer covering the aliased `cva`, the rail stubs and the `ErrorState` case.

sarahgm (approved):

packages/cli/src/commands/migrate.ts
sebald: Agreed — keep `>=`, your reasoning holds: the theme edits are what make the app compile after the bump, so forcing "bump, break everything, then migrate" is the worse sequence. Take the confirm-prompt line as proposed (naming the installed version and that the bump completes the upgrade).

Not blocking the merge on that one string — it can land as a follow-up on `beta-release`.

packages/cli/src/lib/codemod/primitives/restructure-to-slots.ts
sebald: Agreed — document it, do not `reindent`. Rewriting bytes inside a span we promised to move verbatim is the wrong trade for cosmetics the consumer's Prettier already fixes, and the significant-whitespace argument (template literals) settles it.

The README invariant line is not in the branch yet — follow-up is fine.

packages/cli/src/lib/codemod/README.md
sebald: Agreed on both counts — own ticket rather than growing this PR, and your correction stands: `theme-drift.test.ts` covers the mechanical half, so the skill earns its keep on the parts a test cannot check (`oldClasses` from the previous major's published theme, token tables from the release notes, pinned permalinks). Resolving; the ticket is still to be filed.

sebald (approved):
Approved. Every blocking finding is addressed and verified at 4446028:

- the `Sidebar` rail slots and `ErrorState` are back in sync with the `Theme` type, and `manifests/theme-drift.test.ts` now fails CI on the next drift instead of relying on someone remembering
- stubs render under the file's local `cva` name, so an aliasing consumer gets valid output
- the interactive gate matches `init`'s (stdout **and** stdin), and the whole confirm/multiselect flow moved into `commands/migrate.ts` with real coverage
- `migrate 18.1 .` now says "did you mean v18?", with the check ordered above the positional count and pinned by a test

Two agreed items are deliberately left as follow-ups on `beta-release`: the confirm-prompt line for consumers still on the previous major, and the README invariant note about the moved cva keeping its indentation. Neither changes behaviour.

PR #5679, opened 2026-07-27
test(DST-1590): cover ghost Button disabled/loading on every ground

sebald:
# Description

`ui-state-disabled` paints the opaque, white-calibrated `--color-disabled-surface`. `ActionBar` cascades `variant="ghost"` onto its plain `<Button>` children, so a loading or disabled bulk action renders as a **near-white pill on the dark bar**, with a dark spinner on top. On the charcoal-100 page the same fill matches the ground exactly, so the button disappears entirely. `text-disabled` on that fill measures **2.08:1**.

Neither state had a story, so VRT has never seen it. **This PR adds the coverage only** — no behavior, token or style changes. The fix itself is DST-1590.

- `Components/ActionBar → DisabledAndLoading` — rest / disabled / loading on the bar's dark `ui-contrast` ground
- `Components/Button → GhostStatesOnBackground` — the same three states across surface / page / contrast

Added as a *new* Button story rather than extending the baselined `GhostOnBackground`, so existing Chromatic snapshots stay put. Both stories carry doc comments explaining the cause and pointing at DST-1590, so the snapshots don't get "fixed" by accident.

Related: DST-1590 (this PR does not close it — it is the regression coverage landing ahead of the token work)

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. `pnpm sb`, open `Components/ActionBar → DisabledAndLoading`. The "Copy" (disabled) and loading actions render as light blocks on the dark bar; the disabled label sits at ~2:1 against its own fill.
2. Open `Components/Button → GhostStatesOnBackground`. Top row (white surface) is correct. Middle row (page) — the disabled and loading buttons have **no visible pill at all**, just a floating spinner. Bottom row (contrast) — white blocks.
3. `pnpm vitest run --config vite.config.ts --project=storybook-tests ActionBar Button`

## Breaking Changes

No

## Checklist

- [ ] Unit tests added/updated — N/A, no source changes
- [ ] Component documentation added/updated (if it exists) — N/A
- [ ] Visual regression tests updated (for UI changes) — adds two new snapshots; not triggered
- [ ] Changeset added (`pnpm changeset`) — N/A, `*.stories.tsx` only, no published behavior change

---

**Note for the reviewer:** while writing the ActionBar play function I found a separate defect — a pending `<Button>` has **no accessible name**. `loading` swaps the label for a `visibility: hidden` copy (`Button.tsx:93-99`), which removes it from the accessibility tree. Confirmed in both testing-library's and Chromium's name computation. The play function therefore queries the pending button by `data-pending` rather than by name, so the test doesn't lock the bug in. Unrelated to the colour work; needs its own ticket.

### DIFF
--- a/packages/components/src/ActionBar/ActionBar.stories.tsx
+++ b/packages/components/src/ActionBar/ActionBar.stories.tsx
@@ -110,6 +110,65 @@ export const WithoutClearButton = meta.story({
   ),
 });
 
+/**
+ * Inactive states on the ActionBar's dark `ui-contrast` ground.
+ *
+ * The bar cascades `variant="ghost"` onto its plain `<Button>` children, and a
+ * ghost Button takes its disabled/pending fill from `ui-state-disabled` — which
+ * paints the opaque, white-calibrated `--color-disabled-surface`. On this ground
+ * that reads as a light block rather than a receding one. Kept as a snapshot so
+ * the regression is visible in VRT; see DST-1590.
+ */
+export const DisabledAndLoading = meta.story({
+  tags: ['component-test'],
+  args: {
+    selectedItemCount: 2,
+  },
+  render: args => (
+    <ActionBar {...args}>
+      <Button onPress={fn()}>
+        <Pencil />
+        Edit
+      </Button>
+      <Button disabled>
+        <Copy />
+        Copy
+      </Button>
+      <Button loading>
+        <Trash2 />
+        Delete
+      </Button>
+    </ActionBar>
+  ),
+  play: async ({ canvas, step }) => {
+    await step('the enabled action stays operable', async () => {
+      await expect(canvas.getByRole('button', { name: 'Edit' })).toBeEnabled();
+    });
+
+    await step('the disabled action is not operable', async () => {
+      await expect(canvas.getByRole('button', { name: 'Copy' })).toBeDisabled();
+    });
+
+    await step(
+      'the loading action is marked pending, not disabled',
+      async () => {
+        // Queried by state rather than by name: `loading` swaps the label for a
+        // `visibility: hidden` copy, so a pending Button currently has *no*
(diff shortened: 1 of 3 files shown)


packages/components/src/ActionBar/ActionBar.stories.tsx
```diff
@@ -110,6 +110,65 @@ export const WithoutClearButton = meta.story({
...
+        <Copy />
+        Copy
+      </Button>
+      <Button loading>
+        <Trash2 />
+        Delete
+      </Button>
+    </ActionBar>
+  ),
+  play: async ({ canvas, step }) => {
```
aromko: rewrite as .test pls.

sebald: Done in 10f5abe — moved all three assertions into `ActionBar.test.tsx` (importing `DisabledAndLoading` rather than building a fixture) and dropped the `play` function here.

You're right that none of them needed a real browser: they only read ARIA state off the rendered output, so jsdom is enough and it's ~4s faster per run.

`DisabledAndLoading` stays as a story because the actual point of DST-1590 is visual — the ghost Button's opaque `--color-disabled-surface` reading as a light block on the bar's dark `ui-contrast` ground. That only shows up in VRT. Kept the `component-test` tag so it still gets a smoke render, same as `WithoutClearButton` above.

One thing worth flagging: the pending button is still queried via `data-pending` instead of by name, because `loading` swaps the label for a `visibility: hidden` copy and the button ends up with no accessible name (WCAG 4.1.2). That's fixed on `DST-1660-loading-button-accessible-name`, which isn't in `main` yet — once it lands, this test can query by name and the comment above it can go.

aromko (approved):

PR #5681, opened 2026-07-27
feat(DST-1590): add the alpha wash ramp and a tri-ground Storybook playground

sebald:
# Description

Phase 1 of DST-1590: the two-polarity alpha ramp, plus the Storybook playground that makes it reviewable.

**Deliberately additive.** `muted`, `focus-highlight`, `hover` and `selected` still point at the opaque charcoal rungs, so nothing changes visually. Repointing them onto the ramp is the next step — and the reason for shipping in this order is that `Styles/Alpha` gets baselined in Chromatic while it still renders today's colours. The diff on that one story then *is* the review artifact for the change that follows, instead of hunting it across 190 component stories.

Relates to DST-1590 (does not close it).

## Tokens

Two ramps rather than one, because an alpha fill only adapts *within* a polarity: a dark-base wash rides light grounds and vanishes on dark ones, and a light-base wash does the reverse. The alphas are solved per polarity to a shared fill-vs-ground contrast target rather than shared as numbers.

| step | job | ramp A | ramp B | target |
| `50` | `muted` | 0.02 | 0.02 | 1.04:1 |
| `100` | `focus-highlight` | 0.05 | 0.045 | 1.11:1 |
| `200` | `hover` | 0.11 | 0.085 | 1.27:1 |
| `300` | `selected` | 0.19 | 0.14 | 1.52:1 |

Four steps, not the "5 or so" the ticket guessed at — there are exactly four wash jobs in the system, and a fifth rung would be inventing a role.

`--color-fill-*` is the seam between "which ramp" and "which job": a dark ground repoints those four once and every wash beneath it flips polarity with no component-level change. **They must inherit**, which is the opposite of the `--ui-*` vars' deliberate `inherits: false` — called out in a comment where someone would otherwise pattern-match it wrong.

## Playground

`Styles/Alpha`, a sibling of `Styles/RUI`. Five stories, each rendered on all three production grounds:

| Story | Shows |
| --- | --- |
| `Ramp` | both ramps, each swatch measured live against its ground |
| `Adaptive` | derived (`bg-current/10`) vs. pinned polarity |
| `ComponentStates` | the regression grid |
| `Stacking` | R4 — one wash per element, never two |
| `DisabledVeilVsReset` | the open design decision, rendered both ways |

The grounds harness uses the real surface utilities rather than hand-rolled backgrounds. That matters more than it looks: `ui-contrast` also sets `text-primary-foreground`, and polarity derivation keys off `currentColor` — a harness that set only the background would render the adaptive specimens as if they worked while testing nothing at all.

`DisabledVeilVsReset` exists to settle the one question in DST-1590 that measurement cannot: whether a disabled control should reset to a flat neutral or keep its identity under a veil. Both are on screen on all three grounds, so it can be decided by looking.

## Screenshots / Preview

`Ramp` is the whole thesis in one image — ramp A on white measures 1.04 / 1.11 / 1.27 / 1.52 (dead on target) while ramp B there measures 1.00:1, and the relationship inverts on the contrast ground.

`ComponentStates` shows the actual defect: on the contrast ground the ghost Button's disabled and loading states render as near-white pills, and `muted` / `focus-highlight` / `hover` / `selected` paint near-white blocks with white text on them.

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. `pnpm sb`, then open **Styles → Alpha**.
2. On `Ramp`, check the measured ratios: the ramp native to each ground hits its target; the other collapses to ~1.0:1.
3. On `ComponentStates`, compare the three grounds — the contrast column is the defect DST-1590 was filed for.
4. `pnpm test:sb` — 6 tests in `AlphaPalette.stories.tsx`.

## Breaking Changes

No. Purely additive; no existing token changes value.

## Notes for the reviewer

**The regression coverage is two layers and they are not equally strong.** `disabled` and `loading` are real props, but hover cannot be forced from a class on a real component and there is no pseudo-states addon installed — so the hover specimens apply the state utility to a wrapper. That is the same shortcut `Styles/RUI` already takes with `ListSurface`, and it carries the same caveat: a forced specimen shows what the utility paints, not that the component reaches for it. The play function closes that gap by hovering the real button and asserting the computed fill matches the forced specimen beside it.

**The measurement tests are verified to guard.** Ratios come from 1x1 canvas readback plus WCAG relative luminance, so a future token tweak that breaks a target fails CI rather than shipping. Confirmed by perturbing `a-200` from 0.11 to 0.14 and watching it fail with `a-200 on surface: 1.369 vs target 1.27`.

**Two things I deliberately left out.** The ActionBar is not rendered in `ComponentStates` — it is `position: fixed`, so it floats over the story rather than sitting in it, and it already has coverage in `ActionBar → DisabledAndLoading` (#5679). And the pending-Button assertion is a `data-pending` query rather than a name lookup, because the DST-1660 fix is still unmerged in #5680; there is a comment saying to swap it once that lands.

**Ratios on the contrast ground are measured against `--color-primary`**, not the rendered `ui-contrast` gradient. The gradient varies by a couple of hundredths across the surface — inside the tolerance, but it is an approximation and the code says so.

## Checklist

### DIFF
--- a/.changeset/alpha-wash-ramp.md
+++ b/.changeset/alpha-wash-ramp.md
@@ -1,0 +1,9 @@
+---
+'@marigold/theme-rui': minor
+---
+
+Add the two-polarity alpha wash ramp (`--color-charcoal-a-*` / `--color-charcoal-b-*`) and the `--color-fill-*` indirection layer they feed.
+
+An alpha fill only adapts within one polarity — a dark-base wash rides light grounds and vanishes on dark ones, and a light-base wash does the reverse — so there are two ramps rather than one, with alphas solved per polarity to a shared contrast target instead of shared as numbers.
+
+Purely additive: `muted`, `focus-highlight`, `hover` and `selected` still point at the opaque charcoal rungs, so nothing changes visually. Repointing them is the next step of DST-1590.
--- a/packages/components/src/AlphaPalette.stories.tsx
+++ b/packages/components/src/AlphaPalette.stories.tsx
@@ -1,0 +1,712 @@
+import {
+  type PropsWithChildren,
+  type ReactNode,
+  useLayoutEffect,
+  useRef,
+} from 'react';
+import { expect, userEvent, waitFor, within } from 'storybook/test';
+import preview from '.storybook/preview';
+import { cn } from '@marigold/system';
+import { Button } from './Button/Button';
+import { Headline } from './Headline/Headline';
+import { Inline } from './Inline/Inline';
+import { SegmentedControl } from './SegmentedControl/SegmentedControl';
+import { Stack } from './Stack/Stack';
+
+const meta = preview.meta({
+  title: 'Styles/Alpha',
+  parameters: { surface: false },
+});
+
+/* ------------------------------------------------------------------ *
+ * Grounds
+ * ------------------------------------------------------------------ */
+
+/**
+ * The three grounds Marigold ships onto. Every specimen below renders on all
+ * three, because a wash that is only ever reviewed on white is a wash whose
+ * behaviour on the other two is a guess.
+ *
+ * `solid` is the flat colour a ground reduces to for measurement: `ui-contrast`
+ * paints a gradient and a glow over `--color-primary`, so ratios quoted against
+ * it are measured against that base rather than against the rendered gradient.
+ * The gradient varies by a couple of hundredths across the surface, which is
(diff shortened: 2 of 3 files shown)


sebald:
Closing for now — not needed yet. The work is preserved on the `DST-1590-alpha-palette-playground` branch and can be reopened when Phase 1 of DST-1590 is picked up.

PR #5663, opened 2026-07-21
docs(DST-1639): write and finalize v18.0.0 release notes

sebald:
# Description

Fills in the v18.0.0 release blog against the full changeset set. The previous draft was written in early June and had drifted from the final shape of the release.

- Renames the file to `release-2026-08-17.mdx` and updates the release `date` to the mid-August target.
- Reworks the frontmatter `changed:` list to match the real docs slugs (e.g. `app-shell` + `page` replacing the removed `app-layout`, plus the new component/pattern pages).
- Corrects sections that later changesets superseded: the `AppLayout` → `AppShell` + `Page` rename, and the final surface/elevation model (hairline `ui-surface`/`ui-control` rings, single `shadow-elevation-overlay` tier, removed shadow tiers).
- Adds a "New components" section (`SegmentedControl`, `DateRangePicker`, `RangeCalendar`, date presets, `Table.Footer`, `Page`), the remaining breaking changes (Tabs rename, `ActionBar.Button` removal, `SectionMessage` `open`/`announce`, `TagGroup` validation, `Drawer` single-open, `Breakout` removal, dead-prop cleanup), the visual-refresh design story, and the AI tooling story (`marigold search`, docs MCP server, static `.md` endpoints).
- Prose follows the docs punctuation convention: no em/en-dashes or semicolons outside code blocks.

Closes DST-1639

## Test Instructions

1. Run `pnpm start` and open the docs site.
2. Navigate to the Releases section and open the v18.0.0 post (`/releases/blog/release-2026-08-17`).
3. Confirm the page renders, the date reads 2026-08-17, the `changed:` links resolve, and the code samples/tables display correctly.

## Breaking Changes

No — documentation only.

## Checklist

- [ ] Storybook preview and Marigold docs preview are available
- [ ] Stories added/updated (with `component-test` tag where applicable)
- [ ] Unit tests added/updated
- [x] Component documentation added/updated (if it exists)
- [ ] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) (for new/changed interactive components)
- [ ] Visual regression tests updated (for UI changes) — N/A, no UI changes
- [ ] Changeset added (`pnpm changeset`) — N/A, docs-only on the release branch

PR #5682, opened 2026-07-27
docs(DST-1570): add v18 migration guide (MIGRATION-v18.md)

sarahgm:
# Description

Adds `MIGRATION-v18.md`, an exhaustive, agent-consumable guide for upgrading a consuming app from Marigold 17 to 18. It is a plain Markdown file at the repo root, per DST-1570 scope (not published to the docs site yet).

Every breaking change lists what changed, a before→after, why, and gotchas, split into Section A (app developers) and Section B (custom theme authors). Each entry is tagged `[auto]` / `[flagged]` / `[manual]` / `[adopt]` for whether `marigold migrate v18` handles it, so the guide and the codemod compose. The codemod does the mechanical edits and the guide covers everything it will not guess. The guide opens with a `marigold migrate v18` step, an app-structure section (AppShell to Page to Panel, the two-level `Sidebar.Rail`, the background token), and closes with a verification checklist.

Written as if the sibling PRs are already merged: the migrate CLI (#5675, DST-1543) and the v18.0.0 release notes (#5663, DST-1639). Sources of truth are those two plus `marigold docs <Component>`.

Closes DST-1570

## Test Instructions

1. Read `MIGRATION-v18.md` top to bottom for accuracy and completeness.
2. Coverage was cross-checked against all 21 breaking-change headings in the v18.0.0 release notes (#5663) and the ticket scope.
3. Self-tested. A demo consumer exercising the codemod classes was run through `marigold migrate v18 --dry-run`, and every warning and edit the codemod emits maps to a section in the guide.

## Breaking Changes

No. Documentation only (new root file, no code or package changes).

## Checklist

- [ ] Storybook preview and Marigold docs preview are available (N/A, root file, not on the docs site)
- [ ] Stories added/updated (N/A, docs only)
- [ ] Unit tests added/updated (N/A, docs only)
- [ ] Component documentation added/updated (N/A, migration guide rather than component docs)
- [ ] Accessibility reviewed (N/A, docs only)
- [ ] Visual regression tests updated (N/A, docs only, and VRT is not required for beta-release PRs)
- [x] Changeset added (`@marigold/docs: patch`, DST-1570)

## Follow-ups / open items

- [x] **After the release-notes PR #5663 (DST-1639) merges**, link `MIGRATION-v18.md` from the v18 release notes (`release-2026-08-17.mdx`). This is the primary discoverability hook and the main reason to hold until #5663 lands. A README pointer is a smaller in-scope option. A `marigold migrate` "see the full guide" line belongs to DST-1543 / DST-1650. Docs-site publishing was deferred by the ticket.
- [x] **Changeset:** added a `@marigold/docs: patch` changeset (`.changeset/dst-1570-v18-migration-guide.md`). The guide is a root-level file that ships in no published package, but the docs-only precedent (Bulk Actions) applies, so it gets a `@marigold/docs` entry for the changelog.
- [x] **Panel header slot migration:** added as **A3.3**, a beta-only `[manual]` note covering the removed `Panel.Title` / `Panel.Description` / `Panel.HeaderActions` and the collapsible equivalents. Placed in A3 next to the Card→Panel content (`Inset` bumped to A3.4). Verified the before→after against the shipped `Panel` source.


aromko (commented):
## Automated Code Review

Docs-only PR, so I verified the guide's claims against source rather than reading for prose quality. Summary: **unusually accurate**, with four fixable exceptions noted inline.

### What I verified

| Check | Result |
|-------|--------|
| Breaking-change coverage vs. the v18 release notes | 20/20 headings plus the theme checklist, all mapped |
| `[auto]`/`[flagged]`/`[manual]` tags vs. `manifests/v18.ts` | All correct against `renames`, `memberRenames`, `removals`, `warnings`, `scaffolds`, `swaps` |
| Icon rename table vs. manifest `importRenames` | 71/71 exact match, no mismatched targets |
| Icon rename targets vs. `lucide-react@1.17.0` | All 71 resolve to real exports |
| Token rename table vs. `tokens.renamed` | Exact match, and more complete than the release notes |
| Repurposed tokens + B2 recipes vs. `tokens.repurposed` | Exact match |
| `Sidebar.Toggle variant="rail"`, Cmd/Ctrl+B, 5 shell spacing tokens | Verified in source |
| `SelectList` `onChange` signature | Verified (`SelectList.tsx:143-145`) |
| `Panel` compound members (A3.3) | Verified, no `Title`/`Description`/`HeaderActions` |
| `--color-background` = `charcoal-100`, B9 preflight rules | Verified in `tokens.css:69`, `preflight.css:31,35` |
| Changeset | Correct, `@marigold/docs` is in the `fixed` group with precedent |

One place the guide is **more accurate than the release notes**: B1 states the missing-`BooleanField` error unconditionally. That is right. `Switch` always wraps in `BooleanField`, which calls `useClassNames` unconditionally, so a theme without it throws whether or not a `description` is passed. The release notes' "with a `description` prop" qualifier is the wrong one.

### Should fix

1. **`A4.1` Switch before/after is not valid v17 code** (line 476). `children` is in `RemovedProps` in v17 too, so the "before" never compiled. Migrating devs will not recognize their own call sites.
2. **The `Close`/`Print` sentence contradicts its own table** (line 721) and leaves `PauseAlt`, `PlayAlt`, `Whatsapp` with no migration path. The catch-all rule does not cover them.
3. **Two docs-site-relative links 404 on GitHub** (lines 503, 930), one of them the stated source of truth.

### Nice to have

- `<Page>` `console.error`s when its `<main>` has no accessible name (`Page.tsx:127-140`). Apps swapping `AppLayout.Main` to `<Page>` without a page title hit this. The release notes cover it (line 226), the guide does not.
- `HelpText` to `Description` is in the DST-1570 Section A scope but absent. It is non-breaking (`HelpText` is still exported), so `[adopt]` guidance rather than a breaking-change entry, but currently it is silently dropped.
- The release-notes callout links `blob/main/MIGRATION-v18.md`, which 404s until `beta-release` reaches `main`.

No VRT needed: no `packages/*/src`, `themes/`, `.storybook/`, `*.stories.tsx`, or `*.styles.ts` changes. All 18 checks pass.

---

MIGRATION-v18.md
```diff
@@ -0,0 +1,931 @@
...
+### A4.1 `<Switch>` layout and sizing rework `[flagged]` `[manual]`
+
+**What changed.** `<Switch>` now sits on the left of its label, matching `<Checkbox>`
+and `<Radio>`. The default track is smaller. `size="large"` is gone. For the
+settings-page layout (label and description on the left, toggle on the far right),
+use the new `variant="settings"`. `<Switch>` also accepts a `description` prop and
+passes `name` through to the input.
+
+```diff
+- <Switch size="large">Email notifications</Switch>
```
aromko: ⚠️ This "before" is not valid v17 code. `children` is in `RemovedProps` for `<Switch>` in **v17 as well as v18** (`@marigold/components@17.9.1:packages/components/src/Switch/Switch.tsx:8`), so `<Switch size="large">Email notifications</Switch>` never compiled. A real v17 call site used the `label` prop, which means nobody migrating will recognize their own code here.

```suggestion
- <Switch size="large" label="Email notifications" />
```

Worth a sentence noting that only two things actually change for a v17 app: `size="large"` is gone, and `variant="settings"` plus `description` are the new opt-ins. Right now the diff reads as if `description` were required.

MIGRATION-v18.md
```diff
@@ -0,0 +1,931 @@
...
+| `Required`        | `Asterisk`             | `SortDown`               | `ChevronDown`           |
+| `ResaleLogbook`   | `BookOpenText`         | `SortUp`                 | `ChevronUp`             |
+| `Spinner`         | `Loader`               | `Stop`                   | `CircleStop`            |
+| `Thumb`           | `ThumbsUp`             | `Underlined`             | `Underline`             |
+| `Deal`            | `BadgePercent`         | `Zoom`                   | `ZoomIn`                |
+| `Print`           | `Printer`              |                          |                         |
+
+`Close` and `Print` are not in the official mapping table but map to their Lucide
+equivalents, `X` and `Printer`. If you hit an old icon name that is not listed
+above, use its Lucide name. The import still comes from `@marigold/icons`.
```
aromko: ⚠️ Two problems in this sentence.

**`Print` is in the official mapping table.** It is row-listed two lines above (line 719) and appears in `.changeset/iconography-docs.md:71`. The codemod carries it as an `importRename` with the note "added to the official mapping table in this PR", so this sentence is stale. Only `Close` is genuinely unlisted.

**Three legacy names break with no path.** Checked against `lucide-react@1.17.0` (the resolved version): `PauseAlt`, `PlayAlt`, and `Whatsapp` are not in the rename table, are not valid Lucide exports, and are not among v18's 13 retained brand icons. The catch-all "use its Lucide name" fails for all three. `Whatsapp` is the worst case: Lucide dropped brand icons, so there is no equivalent and a consumer using it has no documented path at all. From the v17 sources, `PauseAlt` and `PlayAlt` are circle-outline pause/play, so `CirclePause` / `CirclePlay` are the faithful mappings.

```suggestion
`Close` is not in the official mapping table; it maps to Lucide's `X`. Three legacy
names have no Lucide equivalent and are not retained brand icons, so each needs a
decision: `PauseAlt` → `CirclePause`, `PlayAlt` → `CirclePlay`, and `Whatsapp` is
gone with no replacement (vendor the SVG if you still need it). For any other old
icon name not listed above, use its Lucide name. The import still comes from
`@marigold/icons`.
```

MIGRATION-v18.md
```diff
@@ -0,0 +1,931 @@
...
+```diff
+- <Multiselect label="Genres" options={options} onChange={setValues} />
++ <TagField label="Genres">
++   <TagField.Option id="rock">Rock</TagField.Option>
++   <TagField.Option id="jazz">Jazz</TagField.Option>
++ </TagField>
+```
+
+**Gotchas.** The API differs enough that the codemod will not attempt it. See the
+[TagField docs](/components/form/tag-field) for controlled selection, disabled
```
aromko: ⚠️ Docs-site-relative link in a root-level GitHub file. This resolves to `https://github.com/components/form/tag-field` and 404s. The release-notes callout points readers at the GitHub blob URL, so this file is read on GitHub, not on the docs site.

```suggestion
[TagField docs](https://www.marigold-ui.io/components/form/tag-field) for controlled selection, disabled
```

MIGRATION-v18.md
```diff
@@ -0,0 +1,931 @@
...
+      using `getByRole('alert')` updated
+- [ ] `width="fit"` replaced with an explicit width
+- [ ] `Tooltip open` moved onto `Tooltip.Trigger`
+- [ ] `pnpm typecheck:only` clean (the slot Records are exhaustive, so typecheck is
+      the completeness check for anything the codemod could not fix)
+- [ ] App builds and renders on v18
+
+---
+
+_Source of truth for this guide: the [v18.0.0 release notes](/releases/blog/release-2026-08-17),
```
aromko: ⚠️ Same as the `TagField` link above: this 404s on GitHub, and it is the guide's stated source of truth, so it is the one link most likely to be followed.

```suggestion
_Source of truth for this guide: the [v18.0.0 release notes](https://www.marigold-ui.io/releases/blog/release-2026-08-17),
```

sebald (commented):
I didn't want to re-run @aromko's pass, so I checked the guide a different way — and it holds up. Verifying their four findings first, then what my angle turned up.

### On aromko's four

All correct, and the `Print` one is worth confirming from a second source: `packages/cli/src/lib/codemod/manifests/v18.ts:476-480` carries it as an `importRename` (`Print` → `Printer`) with the note *"added to the official mapping table in this PR"*. So the sentence at line 723 is definitely stale — the codemod already handles it, and `Close` is the only genuinely unlisted name.

On the two 404 links: I swept the whole file for site-relative paths rather than spot-checking, and there are **exactly two** — lines 503 and 930, both already flagged. Nothing else to chase.

### The check aromko's method couldn't make

Their coverage table verifies the guide against the v18 release notes. That's the right reference for *prose* consistency, but both documents are artifacts of this PR, so "20/20 headings mapped" can't catch a breaking change that's missing from the release notes too. So I mapped the guide against the **21 changesets on `beta-release` carrying a `major` bump** instead — the actual source of truth for what breaks.

**All 21 land in a section.** Including the three I specifically expected to be gaps, all of which turned out to be handled correctly:

- **DST-990 (`Menu` revisit)** — I went looking for a missing `A`-section and there isn't one needed. It's `components: minor` / `system: major`, and the changeset states outright that all component additions are backward compatible; the only break is the `Theme` type requiring a `keyboard` key. Covering it in **B4** and nowhere in Section A is exactly right. Easy one to get wrong in either direction.
- **DST-1164** removes `Container`'s `align` alongside `<Breakout>`, which the A8.1 heading doesn't advertise — but lines 730-732 cover it, including *why* it was dead (`[data-breakout]`).
- **DST-1548** `Card.Body` → `Card.Content` is at 399, plus the theme slot key and the verification checklist at 916.

I also spot-checked the `[auto]` tags against the codemod manifest rather than the release notes, since a wrong `[auto]` is the costliest kind of error here — a reader skips a fix the codemod never makes. `Inset` `space`/`spaceX`/`spaceY` → `p`/`px`/`py`, the `Tabs.TabPanel`→`Panel` and `SelectList.Item`→`Option` `memberRenames`, the `TextField` `min`/`max` `removals`, and the `BooleanField` scaffold are all there. The `FileTrigger` example at 588 is exact down to the array literal — the manifest rule sets `wrapInArray: true`, so `acceptedFileTypes={["image/png"]}` is genuinely what you get, not a plain key rename. `Breakout` has zero manifest entries, consistent with its `[manual]` tag.

### One small gap

`.changeset/dst-1559-fix-dead-mis-named-props.md` has three bullets; A4.6 covers two. The third: `Loader`'s `loaderType` JSDoc documented a `cycle` value that never existed — the accepted values are `xloader` and `circle`.

Strictly it isn't a breaking change, so I can see the argument for leaving it out. But someone who wrote `loaderType="cycle"` in v17 because the JSDoc told them to has code that silently never worked and still won't after the upgrade, and an upgrade audit is exactly when you'd want to hear that. A line in A4.6's list would cover it. Your call — the section is already the "props that silently did nothing" bucket, so it'd sit naturally there.

Worth noting the `Loader` in the icon rename table at 716 is `Spinner` → `Loader`, a different thing entirely. If you do add this, it's worth disambiguating so nobody reads them as related.

Nothing else from me. Fix aromko's four and I'm happy — the accuracy here is genuinely unusual for a document this size.

MIGRATION-v18.md
aromko: Fixed in 64521ee. Confirmed against `main` before changing it: `children` sits in
`RemovedProps` in v17 too, so the "before" never typechecked.

Took the `label` suggestion and added a line after the diff, since the diff on its
own still read as if `description` were required:

> Only `size="large"` actually breaks. `variant="settings"` and `description` are
> new opt-ins, not requirements: `<Switch label="Email notifications" />` keeps working.

MIGRATION-v18.md
aromko: Fixed in 64521ee. Both halves check out.

`Print` → `Printer` is an `importRename` in `packages/cli/src/lib/codemod/manifests/v18.ts:477`
(now on this branch via the `beta-release` merge, not just #5675), so the sentence
contradicted both the table and the codemod. `Close` is the only genuinely unlisted name.

Re-resolved the other three against `lucide-react@1.17.0` as installed rather than
taking the version on trust: `PauseAlt`, `PlayAlt` and `Whatsapp` are not exports,
`CirclePause` and `CirclePlay` are, and the v17 SVG paths for `PauseAlt` / `PlayAlt`
are circle-outline pause/play, so those two mappings are faithful.

One wording change to the suggestion: only `Whatsapp` is a brand icon, so "are not
retained brand icons" shouldn't cover all three. Landed as:

> `Close` is not in the official mapping table; it maps to Lucide's `X`. Three legacy
> names have no Lucide equivalent and each needs a decision: `PauseAlt` →
> `CirclePause`, `PlayAlt` → `CirclePlay`, and `Whatsapp` is gone with no replacement
> (Lucide dropped brand icons and it is not among the 13 retained ones, so vendor the
> SVG if you still need it). For any other old icon name not listed above, use its
> Lucide name. The import still comes from `@marigold/icons`.

MIGRATION-v18.md
aromko: Fixed in 64521ee, now `https://www.marigold-ui.io/components/form/tag-field`.

MIGRATION-v18.md
aromko: Fixed in 64521ee, now `https://www.marigold-ui.io/releases/blog/release-2026-08-17`.

Swept the file independently rather than spot-checking (`grep -nE '\]\(/'`): these two
were the only site-relative links, and there are none left.

aromko:
All four inline findings are fixed in 64521ee, replies in each thread. Rebased onto
the `beta-release` merge first, so the codemod manifest is now on this branch and I
re-verified the `[auto]` claims against the merged `manifests/v18.ts` rather than #5675.

Also landed from the nice-to-haves, plus two I deliberately did not take.

### Added

**`<Page>` accessible name** (@aromko). Real, and the guide's own A3.1 example hides
it: the example includes a `<Title>`, so the failure mode is invisible to anyone
pattern-matching off it. `Page.tsx:130-144` `console.error`s in development when there
is no `<Title>`, `aria-label`, or `aria-labelledby`. Two lines added to A3.1.

**`Loader` `loaderType`** (@sebald). Added to A4.6 with one correction to the framing.
`LoaderVisualType` was `'xloader' | 'circle'` in v17 as well, so `loaderType="cycle"`
was a type error, not a silent no-op. Only the JSDoc was wrong. The other A4.6 entries
are genuinely "compiled fine, did nothing", so this one is worded to not claim that:

> - **`Loader` `loaderType`**: the v17 JSDoc documented a `cycle` value that never
>   existed. The type only ever accepted `xloader` and `circle` (default `circle`),
>   so this is a doc fix, not a code change. Unrelated to the `Spinner` → `Loader`
>   icon rename in A7.1.

Disambiguation from the icon rename included, as you suggested.

### Not taken

**`HelpText` → `Description`.** It is in the DST-1570 Section A scope list, so the
observation is correct, but the underlying change is
`.changeset/dst-1517-helptext-uses-description.md`, `@marigold/components: patch`,
which states "No names, output, or behaviour change". `HelpText` is still exported
(`packages/components/src/index.ts:160`) with an unchanged API, and the release notes
omit it too. There is nothing for an app developer to do, so a Section A entry would
document a breaking change that does not exist. The ticket's scope line looks like it
came from the changeset digest before DST-1517 landed as a no-op refactor.

**The `blob/main/MIGRATION-v18.md` callout.** It 404s today, but the page carrying it
is the v18 release notes, which publish when v18 ships, and by then `beta-release` has
reached `main` and the link resolves. Pointing it at `blob/beta-release/` would fix it
today and break it permanently after the merge. Leaving it on `main` is the self-healing
option.

Happy to add either if you disagree.

### Verification

- `grep -nE '\]\(/'` returns nothing, both site-relative links gone and no others.
- `prettier --check MIGRATION-v18.md` passes.
- Docs-only, no `packages/*/src`, `themes/`, `.storybook/`, `*.stories.tsx` or
  `*.styles.ts` changes, so still no VRT.

sebald (approved):
Verified all four inline fixes plus the two nice-to-haves in `64521ee`, and both "not taken" calls are the right ones. Thanks for picking this up while @sarahgm is out.

### The four

1. **A4.1 Switch** — now `label="Email notifications"` (:482), and the follow-up line at :491 does the job the diff couldn't: `variant="settings"` and `description` read as opt-ins, not requirements.
2. **Icon sentence** (:734-737) — `Close` as the only unlisted name, `PauseAlt` → `CirclePause`, `PlayAlt` → `CirclePlay`, `Whatsapp` gone. Your narrowing of the brand-icon clause to `Whatsapp` alone is more accurate than my suggestion was.
3. **Both site-relative links** — `grep -nE '\]\(/'` on the file at head returns nothing. Independently confirmed.
4. **`<Page>` accessible name** — good catch that A3.1's own example masked the failure mode by including a `<Title>`. :361-365 covers it.

### On the Loader entry

Your correction stands and mine was wrong: `LoaderVisualType` is `'xloader' | 'circle'` on `main` (`Loader.tsx:8`), so `loaderType="cycle"` was a type error, not the silent no-op I described. Only the JSDoc — including its `@default cycle` — was wrong. The A4.6 wording at :597-600 is right to frame it as a doc fix and to keep it out of the "compiled fine, did nothing" bucket, and the A7.1 disambiguation is there.

### On the two you didn't take

- **`HelpText` → `Description`**: agreed, drop it. Verified `HelpText` is still exported (`index.ts:160`) and `dst-1517` is a `patch` ending in "No names, output, or behaviour change". A Section A entry would document a break that doesn't exist; the ticket's scope line predates DST-1517 landing as a no-op.
- **`blob/main/MIGRATION-v18.md`**: agreed. `blob/beta-release/` trades a temporary 404 for a permanent one. Self-healing is correct.

All 17 checks green. Good to merge into `beta-release`.

PR #5683, opened 2026-07-28
fix(DST-1661): give menu items and the file drop zone a real focus ring

sebald:
# Description

`Menu` items and the `FileField` drop zone signalled keyboard focus with a background wash and nothing else — **1.11:1** for a menu item, **~1.05:1** for the drop zone. Both are far under the 3:1 that WCAG 2.1 SC 1.4.11 asks of a state indicator. In a menu the wash is doubly unreliable: React Aria moves `focusedKey` on hover, so `data-focused` (and the wash) fires on hover too, making keyboard focus and mouse hover visually identical.

**Copying `ListBox` would not have fixed it.** `ui-state-focus` is a *two-part* indicator:

```css
@utility ui-state-focus {
  @apply outline-ring/50 outline-3 outline-offset-0 outline-solid;
  --ui-border-color: var(--color-ring);   /* the load-bearing half */
}
```

On a framed control the **border** flips to full-opacity `--color-ring` (5.50:1) and the outline is a soft halo around it. On an element with no border to flip, only the halo lands — measured at **2.08:1** on white. Confirmed on the real component, not inferred: a keyboard-focused `ListBox` row renders `outline: solid 3px oklab(… / 0.5)` and no border flip. **ListBox's own focus indicator is 2.08:1 today.**

So this adds the missing counterpart, `ui-state-focus-item`, for borderless rows:

```css
@utility ui-state-focus-item {
  @apply inset-ring-2 inset-ring-ring;
}
```

Full opacity because this ring *is* the whole indicator — 4.97:1 over `focus-highlight`, 3.59:1 over `selected`, 3.14:1 on the contrast ground. Inset rather than outline because menu items sit only 4px (`p-1`) inside an `overflow-hidden`, rounded `Popover`; I confirmed that clip boundary in the DOM (item spans 32–142.7px inside a container clipping at 28–146.7px), so an outward 3px outline would land right on it. An inset ring cannot be clipped and needs no outline reset from the caller.

The ring goes on Menu's `item` base rather than `itemBase` — `destructive` does not share `itemBase` and every variant needs it.

**Only keyboard focus changes.** Hover, selection and pointer interaction are untouched.

Closes DST-1661
Follow-up: DST-1662 — migrate `ListBox`, `SelectList`, `Table` rows and `Sidebar` off the 2.08:1 halo. Deliberately not in this PR: it is a visual change on every focusable row in the system and wants its own Chromatic review rather than riding along on a bug fix.

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Storybook → `Components/Menu` → `Basic`. Open the menu with the mouse, then press <kbd>ArrowDown</kbd>. The focused item should carry a solid 2px inset ring in addition to the background wash.
2. Move the mouse over a different item. The hovered row should get the wash **but no ring** — the ring is `focus-visible` only.
3. `Components/Menu` → `Advanced`, <kbd>ArrowUp</kbd> to "Delete". The `destructive` variant should also get the ring (it does not share `itemBase`).
4. `Components/Menu` → `Mobile` (small viewport, renders in a Tray). Same behaviour, ring not clipped.
5. `Components/FileField` → `Basic`. <kbd>Tab</kbd> to the drop zone — inset ring inside the dashed border.
6. Automated: `pnpm test:sb --run packages/components/src/Menu/Menu.stories.tsx`

## Breaking Changes

No

## Checklist

---

### Note on the test

The new assertion reads the rendered `box-shadow`, walks the ancestor chain to find what is actually painted underneath (menu items and the Menu container are both transparent — `Popover` paints the surface), and composites via canvas against both adjacent colors. It fails on a token or utility regression, not merely on a class name being present.

**Verified by perturbation** rather than assumed to work: dropping the ring to `/50` reports `focus ring vs inside (row fill) is 2.01:1, needs 3:1`; removing it reports the empty box-shadow. Both reverted.

### Known, deliberately left out

After keyboard navigation, a single synthetic mouse-move leaves the hovered row with `data-hovered` but no background — React Aria suppresses focus-follow-hover while focus-visible is active. With a real mouse the next move corrects it. Closable with a `hover:` rule on the same lines, but it is a transient I could only produce synthetically, so I did not widen scope for it.

### DIFF
--- a/packages/components/src/Menu/Menu.stories.tsx
+++ b/packages/components/src/Menu/Menu.stories.tsx
@@ -332,6 +332,113 @@ Basic.test(
   }
 );
 
+// WCAG 1.4.11: visual information required to identify a state needs 3:1
+// against adjacent colors. A focus indicator is such a state.
+const WCAG_NON_TEXT = 3;
+
+/**
+ * Every painted background from the page down to `element`, bottom layer first.
+ * Walked rather than assumed: menu items are transparent, the Menu container is
+ * transparent, and the Popover is what actually paints the surface.
+ */
+const paintedGround = (element: HTMLElement | null) => {
+  const layers: string[] = [];
+  for (let node = element; node; node = node.parentElement) {
+    const background = getComputedStyle(node).backgroundColor;
+    if (background !== 'rgba(0, 0, 0, 0)' && background !== 'transparent') {
+      layers.unshift(background);
+    }
+  }
+  return layers;
+};
+
+/**
+ * Composite the layers onto white and read the pixel back, so alpha, oklch and
+ * color-space conversion are resolved by the browser instead of by us.
+ */
+const flatten = (layers: string[]) => {
+  const canvas = document.createElement('canvas');
+  canvas.width = 1;
+  canvas.height = 1;
+  const context = canvas.getContext('2d', { willReadFrequently: true })!;
+  context.fillStyle = '#fff';
+  context.fillRect(0, 0, 1, 1);
+  for (const layer of layers) {
+    context.fillStyle = layer;
+    context.fillRect(0, 0, 1, 1);
+  }
+  const [r, g, b] = context.getImageData(0, 0, 1, 1).data;
+  return [r!, g!, b!] as const;
+};
+
+const contrast = (a: readonly number[], b: readonly number[]) => {
+  const luminance = (rgb: readonly number[]) => {
+    const [r, g, b] = rgb.map(value => {
+      const channel = value! / 255;
+      return channel <= 0.03928
+        ? channel / 12.92
+        : ((channel + 0.055) / 1.055) ** 2.4;
+    });
+    return 0.2126 * r! + 0.7152 * g! + 0.0722 * b!;
+  };
+  const [high, low] = [luminance(a), luminance(b)].sort((x, y) => y - x);
+  return (high! + 0.05) / (low! + 0.05);
+};
+
+Basic.test(
+  'Keyboard focus draws a ring that clears 3:1 against both adjacent colors',
+  { parameters: { chromatic: { disableSnapshot: true } } },
+  async ({ canvas }) => {
+    await userEvent.click(
+      canvas.getByRole('button', { name: 'Hogwarts Houses' })
+    );
+    await userEvent.keyboard('{ArrowDown}');
+
+    const item = await waitFor(() => {
+      const focused = canvas
+        .getByRole('menu')
+        .querySelector<HTMLElement>('[data-focus-visible]');
+      expect(focused).not.toBeNull();
+      return focused!;
+    });
+
+    // The background wash measures ~1.11:1 on its own and, because focus
+    // follows the mouse in a menu, it also fires on hover. It cannot be the
+    // indicator. Assert the ring is a separate, real layer.
+    const { boxShadow } = getComputedStyle(item);
+    const ring = boxShadow
+      .split(/,(?![^(]*\))/)
+      .map(shadow => shadow.trim())
+      .find(shadow => shadow.includes('inset'));
+    expect(
+      ring,
+      `no inset focus ring in box-shadow: ${boxShadow}`
+    ).toBeTruthy();
+
+    const ringColor = ring!.match(
+      /(?:oklch|oklab|rgba?|color)\([^)]*\)|#[0-9a-f]{3,8}/i
+    )?.[0];
+    expect(ringColor, `no color found in shadow: ${ring}`).toBeTruthy();
+
+    // WCAG's "adjacent colors": the row's own fill inside the ring, and the
+    // menu surface outside it.
+    const inside = paintedGround(item);
+    const outside = paintedGround(item.parentElement);
+    expect(inside.length).toBeGreaterThan(outside.length);
+
+    for (const [side, ground] of [
+      ['inside (row fill)', inside],
+      ['outside (menu surface)', outside],
+    ] as const) {
+      const ratio = contrast(flatten([...ground, ringColor!]), flatten(ground));
+      expect(
+        ratio,
+        `focus ring vs ${side} is ${ratio.toFixed(2)}:1, needs ${WCAG_NON_TEXT}:1`
+      ).toBeGreaterThanOrEqual(WCAG_NON_TEXT);
+    }
+  }
+);
+
 export const MultiSelection = meta.story({
   tags: ['component-test'],
   parameters: { chromatic: { disableSnapshot: true } },
--- a/packages/components/src/Menu/Menu.stories.tsx
+++ b/packages/components/src/Menu/Menu.stories.tsx
@@ -7,6 +7,13 @@ import { Description } from '../Description/Description';
 import { Divider } from '../Divider/Divider';
 import { Keyboard } from '../Keyboard/Keyboard';
 import { TextValue } from '../TextValue/TextValue';
+import {
+  WCAG_NON_TEXT,
+  contrast,
+  flatten,
+  insetFocusRing,
+  paintedGround,
+} from '../contrast.utils';
 import { ActionMenu } from './ActionMenu';
 import { Menu } from './Menu';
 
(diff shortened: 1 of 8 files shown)


themes/theme-rui/src/ui.css
```diff
@@ -183,12 +183,39 @@
...
+ * Inset rather than outline: rows live inside overflow-hidden, rounded overlays
+ * (Popover paints the rim, the Menu inside it is only p-1 off the edge), so an
+ * outward outline gets clipped at the corners. An inset ring cannot be, and it
+ * needs no outline reset from the caller.
+ *
+ * Pairs with, rather than replaces, a focus background -- the wash tracks the
+ * roving pointer/keyboard cursor, this proves where keyboard focus actually is.
+ */
+@utility ui-state-focus-item {
+  @apply inset-ring-ring inset-ring-2;
```
aromko, line 216: 🔴 **This utility needs to suppress the UA focus ring itself.**

`ui-state-focus` draws an outline, so its callers add `outline-none` to the base and the utility's outline replaces the UA one. All 23 call sites in the theme do this. `ui-state-focus-item` draws only a box-shadow, so the UA ring is left standing and the caller has to kill it.

Menu's `item` base already carries `outline-hidden`, so Menu is fine. `FileField.styles.ts:13` has no outline reset, and I could not find a global one: nothing in `preflight.css`, `styles.css`, the `ui.css` base layer, and `packages/system` ships no CSS. `DropZone` receives only `classNames.dropZone`.

Since `@apply` inside the utility only lands when the utility does (under `focus-visible:`), which is exactly when the UA ring appears, fixing it here covers both call sites and every future one:

```suggestion
  @apply inset-ring-ring inset-ring-2 outline-hidden;
```

That also makes line 210's "needs no outline reset from the caller" true, which it currently is not.

Static analysis only, I did not render it. Worth a 10 second check in Storybook: `Components/FileField → Basic`, <kbd>Tab</kbd> to the drop zone, and read `outline` off the focused element. If something does reset it that I missed, point me at it and I will drop this.

themes/theme-rui/src/components/FileField.styles.ts
```diff
@@ -6,7 +6,11 @@ export const FileField: ThemeComponent<'FileField'> = {
     base: [
       'relative overflow-hidden transition-[color,background]',
       'data-[drop-target=true]:bg-muted',
-      'focus-visible:bg-focus-highlight/50',
+      // The tint alone composites to ~1.05:1 against the surface -- it reads as
+      // a warmth, not as focus. The ring is the indicator; the tint stays as its
+      // accompaniment. The dashed border is decorative and never flips, so
+      // ui-state-focus would have nothing to recolor here.
+      'focus-visible:bg-focus-highlight/50 focus-visible:ui-state-focus-item',
```
aromko, line 13: 🔴 **Same root cause as my comment on `ui.css:216`, this is the call site it affects.**

The drop zone is the one `ui-state-focus-item` consumer with no `outline-none` / `outline-hidden` in its base, so it should paint the UA focus ring alongside the new inset ring.

This also bears on the comment above the line. If the UA ring was already there, the drop zone was not "a background wash and nothing else" at ~1.05:1, and this change adds a second indicator rather than supplying a missing one. Worth re-measuring before the comment lands as documentation.

Preferred fix is in the utility (see the other comment). If you would rather keep the utility purely additive, then here:

```suggestion
      'focus-visible:bg-focus-highlight/50 focus-visible:ui-state-focus-item outline-hidden',
```

Note this is also the only change in the PR with neither a story test nor a Chromatic snapshot behind it.

packages/components/src/Menu/Menu.stories.tsx
```diff
@@ -332,6 +332,113 @@ Basic.test(
...
+    });
+    return 0.2126 * r! + 0.7152 * g! + 0.0722 * b!;
+  };
+  const [high, low] = [luminance(a), luminance(b)].sort((x, y) => y - x);
+  return (high! + 0.05) / (low! + 0.05);
+};
+
+Basic.test(
+  'Keyboard focus draws a ring that clears 3:1 against both adjacent colors',
+  { parameters: { chromatic: { disableSnapshot: true } } },
```
aromko: ⚠️ **This disables the only snapshot that would show the new ring.**

`Basic` does not set `disableSnapshot`, so this test inherits snapshots-on and then opts out. Elsewhere in this file the convention runs the other way: `AccessSections.test`, `Mobile.test` and `Advanced.test` all set `disableSnapshot: false` to capture the interaction state.

I checked all four snapshotted tests in the file. Every one drives the UI with `userEvent.click` only, so none produces `:focus-visible` and none renders the ring. Net result: the indicator this PR exists to add has no visual baseline, which matches the Chromatic check reporting "39 changes accepted as baselines" with no focus-ring diff among them.

```suggestion
  { parameters: { chromatic: { disableSnapshot: false } } },
```

The box-shadow assertion catches token and utility regressions, which is genuinely more than a snapshot does. But it cannot catch the ring being clipped, doubled or misaligned, and DST-1662 expects "real Chromatic diffs on keyboard-focus snapshots" to compare against.

packages/components/src/Menu/Menu.stories.tsx
```diff
@@ -332,6 +332,113 @@ Basic.test(
...
+    const ringColor = ring!.match(
+      /(?:oklch|oklab|rgba?|color)\([^)]*\)|#[0-9a-f]{3,8}/i
+    )?.[0];
+    expect(ringColor, `no color found in shadow: ${ring}`).toBeTruthy();
+
+    // WCAG's "adjacent colors": the row's own fill inside the ring, and the
+    // menu surface outside it.
+    const inside = paintedGround(item);
+    const outside = paintedGround(item.parentElement);
+    expect(inside.length).toBeGreaterThan(outside.length);
```
aromko: 💡 This asserts the focus wash still exists, inside a test whose subject is the ring. The PR argues the wash is not the indicator and could legitimately go; if it ever does, this line fails for a reason unrelated to the ring.

If it is here as a guard against `paintedGround` returning something vacuous (fair, given the DST-1590 test that passed against nothing), consider making that explicit instead, for example `expect(inside.at(-1)).not.toBe(outside.at(-1))`, so the intent reads as "the row paints its own layer" rather than "the wash is present".

themes/theme-rui/src/ui.css
sebald: Good catch to chase this, but I rendered it and the premise doesn't hold — no `outline-hidden` needed, and adding it here wouldn't do anything. Three things, in order of how much they matter:

**1. `focus-visible:` in this theme isn't `:focus-visible`.** `styles.css:20` loads `@plugin 'tailwindcss-react-aria-components'`, which remaps the variant onto RAC's data attribute. Measured on the preview deployment, drop zone focused:

```
dz.matches(':focus-visible')        → false
getComputedStyle(dz).boxShadow      → "… oklch(0.52 0.008 54) 0px 0px 0px 2px inset …"
dz.hasAttribute('data-focus-visible') → true
```

The ring lands via `[data-focus-visible]`, not `:focus-visible`. So "the `@apply` only lands under `focus-visible:`, which is exactly when the UA ring appears" isn't the relationship — the two are decoupled here.

**2. The drop zone is never the focused element.** It's a `<div>` with no `tabindex`. RAC puts DOM focus on its own **0×0 `<button aria-label="DropZone">`** child, and mirrors the state onto the div as `data-focus-visible`. So `outline-hidden` on the utility (or on `dropZone`) would be a no-op twice over: the div never focuses, and `outline` isn't inherited — a parent can't reset a child's outline.

**3. Nothing is painted.** I cropped the screenshot to the proxy's exact coordinates (727.5, 103.5 within the zone) — a zero-area box paints no outline ring. Chromium via Playwright; the new story test also passes in Firefox, which is where `test:sb` runs.

Menu lines up with this: `outline-hidden` on `item` is inert for the same reason (virtual focus, the DOM-focused node is the menu container), so it isn't what's saving Menu — there was nothing to save it from.

Net: line 210's "needs no outline reset from the caller" is true, just for a different reason than the comment implies. I've left the utility as-is.

themes/theme-rui/src/components/FileField.styles.ts
sebald: Three separate points here — rejecting the first, and you're right on the other two.

**The `outline-hidden`:** not needed, see [my reply on `ui.css:216`](https://github.com/marigold-ui/marigold/pull/5683#discussion_r3666798895). Short version: this `<div>` never takes DOM focus (RAC uses a 0×0 proxy `<button>` inside it), the theme's `focus-visible:` variant is remapped onto `[data-focus-visible]` by `tailwindcss-react-aria-components`, and nothing is painted. `outline-hidden` here can't reset a child's outline anyway.

**Re-measuring:** did it, on the preview deployment, compositing through a canvas so the browser resolves the alpha and oklch:

| | ratio |
| --- | --- |
| tint vs surface (what was there before) | **1.05:1** |
| new ring vs tint, inside | 5.23:1 |
| new ring vs surface, outside | 5.50:1 |

So the `~1.05:1` in the changeset is exact, and since no UA ring was ever there, "a background wash and nothing else" holds — this supplies a missing indicator rather than adding a second one. No re-wording needed.

**"the only change with neither a story test nor a Chromatic snapshot":** entirely fair, and the one I should have caught. Fixed in 8b5ead2 — added `Basic.test('Keyboard focus draws a ring that clears 3:1 against both adjacent colors')` with `disableSnapshot: false`.

I checked it isn't vacuous, since that was the DST-1590 lesson. With `focus-visible:ui-state-focus-item` removed it fails on `no inset focus ring in box-shadow: none: expected undefined to be truthy`, then passes again restored.

Worth noting for anyone reading this later: the test asserts on the styled `<div>` via `data-focus-visible`, *not* on `document.activeElement` — that's the 0×0 proxy and carries none of the theme's styling. There's a comment in the test saying so.

Rather than copy ~50 lines of color math into a second stories file, the helpers moved to `packages/components/src/contrast.utils.ts` (not exported from `index.ts`). Menu now imports from there.

packages/components/src/Menu/Menu.stories.tsx
sebald: Agreed on the outcome — fixed in 8b5ead2, now `disableSnapshot: false`.

I verified the part that actually matters: all four snapshotted tests in this file (`AccessSections` 519, `DisabledWithDescription` 568, `Mobile` 604, `Advanced` 728) drive the UI with `userEvent.click` alone, so none produces `:focus-visible` and none renders the ring. The two `Mobile` tests that *do* use `keyboard()` (655, 680) inherit the story-level `disableSnapshot: true`. So the net effect was exactly as you describe: no baseline for the indicator this PR exists to add.

One correction on the convention, since it'll come up again: `disableSnapshot: true` on a `Basic.test` *is* this file's convention — 147, 281 and 313 all do it. The split isn't one convention running the other way, it's two groups. `Basic` is snapshot-**on** at story level, so its tests opt out to avoid piling up near-identical baselines; `AccessSections` / `Mobile` / `Advanced` are snapshot-**off** at story level, so their one visual test opts in. Same intent, opposite default.

Which is why this test is the right exception rather than a correction: it's the one `Basic` interaction that produces a genuinely distinct visual state. I've put that reasoning in a comment above the parameter so the next person doesn't "fix" it back.

Also took your framing on the split of duties — assertions catch token and utility regressions, the snapshot catches clipping, doubling and misalignment. Both, not either.

packages/components/src/Menu/Menu.stories.tsx
sebald: You read the intent correctly — it was a guard against `paintedGround` coming back vacuous, and you're right that as written it asserted the wash. Changed in 8b5ead2.

I went with `expect(outside.length).toBeGreaterThan(0)` rather than your `inside.at(-1) !== outside.at(-1)`, because that one keeps the coupling you were objecting to: if the wash is ever dropped, the row stops painting its own layer and the assertion fails — again for a reason unrelated to the ring. Guarding that the walk found *any* real ground covers the actual failure mode (an empty walk makes `flatten` fall back to bare white, so the ratios get measured against something not on screen — passing against nothing, the DST-1590 shape) while staying indifferent to whether the wash is there at all.

Comment now says that explicitly, so it reads as "the walk found real ground" instead of either "the wash is present" or "the row paints its own layer". Shout if you'd still rather have the stricter version.

aromko (approved):

sebald:
/run-chromatic

Dispatching VRT for the two new keyboard-focus baselines added by this PR — the `Menu` item focus ring and the `FileField` drop-zone ring. Both tests flip `chromatic.disableSnapshot` to `false`, and they are the only snapshotted interactions in either file that drive the UI with the keyboard rather than `click`, so these are the first baselines that show a focus ring at all.

Expect two new snapshots to come up for approval rather than diffs against an existing baseline.

PR #158, opened 2020-03-27
chore(setup): use tsdx (#138)

viktoria-schwarz:
closes #138 
idk which dependencies (where) to drop?

PR #5687, opened 2026-07-30
fix(DST-1666): apply DateFormat's documented tabular default

sebald:
# Description

`DateFormat` documented `tabular` as defaulting to `true` but never applied that default. The prop was destructured without one, so it stayed `undefined` unless a consumer passed it explicitly, and `tabular-nums` was never applied.

Three places already documented the opposite:

- the `@default true` JSDoc, which is what feeds the generated props table on the docs site
- the prose docs — "To disable it, set `tabular={false}`", which only parses if the default is on
- `NumericFormat`, the sibling formatter, which has always destructured `tabular = true` and behaved correctly

So the docs site actively told consumers that dates come out with tabular digits, and they did not.

**Current vs. new behavior:** `<DateFormat value={date} />` previously rendered proportional digits; it now renders with `tabular-nums`. `tabular={false}` still opts out, and now does so without leaving an empty `class=""` attribute behind (matching `NumericFormat`'s `undefined` falsy branch).

**Why it went unnoticed:** the only place exercising the prop is `dateformat-tabular-format.demo.tsx`, which passes `tabular={tabular}` explicitly from a `useState(true)` toggle. The demo never relied on the default, so it always looked correct. Neither formatter had any test for `tabular`.

Found while reviewing the `suppressHydrationWarning` change to these same two files in #5522 (DST-1434). No behavior from that change is touched here.

Closes DST-1666

## Screenshots / Preview

Not captured locally — this is a digit-width change that only reads meaningfully in a table, and it lands wherever `DateFormat` is used. The `dateformat-tabular-format` and `numericformat-tabular-format` demos in the docs preview are the clearest place to see it: the Switch in each demo toggles `tabular` explicitly, so both states are directly comparable.

~~Chromatic is the real check here — see the note under Breaking Changes.~~ **Correction:** Chromatic does not cover this change. `.storybook/main.ts` globs stories only from `packages/components/src` and `packages/system/src`, and no story renders `<DateFormat>` — every call site is under `docs/`. The docs preview is the real check. See [this thread](https://github.com/marigold-ui/marigold/pull/5687#discussion_r3681429209) for the detail.

## Test Instructions

1. `pnpm vitest run --config vite.config.ts --project=unit-tests packages/system/src/components/Formatters` — 22 tests pass. The four new ones cover the default and the opt-out on both formatters.
2. To see the bug rather than the fix: revert `tabular = true` back to `tabular` in `DateFormat.tsx` and re-run. `uses tabular digits by default` fails with an empty received class list, and `allows opting out of tabular digits` fails with `class=""`. Both were confirmed failing before the fix was applied.
3. In the docs preview, open **Components → Formatters → DateFormat → Tabular display** and toggle "Use tabular digits". Compare against the same demo on **NumericFormat**; the two now behave identically.
4. Anywhere `<DateFormat>` renders in a `Table` column (e.g. the billing and users examples), digits should now be uniform width and line up between rows.

## Breaking Changes

No — in the semver sense. `patch` on `@marigold/system`, since this restores the documented contract rather than changing it.

Worth flagging for the reviewer anyway: this is a **visible rendering change at every `DateFormat` call site that omits the prop, which is all 38 of them in this repo** — none passed `tabular` explicitly. Consumers who want the old look pass `tabular={false}`.

~~The Chromatic diff will therefore be wide, and every date in it is expected to change.~~ It turned out not to be: all 38 call sites are in `docs/`, which Storybook does not glob, so the expected snapshot delta is zero and Chromatic [build 159](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=159) showed no change attributable to this diff. Verified on the docs preview instead. The flip side is that this change has **no VRT coverage** — nothing would catch a future regression of it.

## Checklist

- [ ] Stories added/updated (with `component-test` tag where applicable) — n/a, these formatters have no stories; covered by unit tests instead
- [x] Component documentation added/updated (if it exists) — added "(the default)" to the `DateFormat` tabular section so it matches `NumericFormat`'s wording
- [ ] Accessibility reviewed against ARIA APG — n/a, no interactive or semantic change; `tabular-nums` is purely a font-feature setting
- [x] Visual regression tests updated (for UI changes) — Chromatic [build 159](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=159) run on `a31a2d916` and reviewed; the 4 pending changes are pre-existing baseline drift, unrelated to this diff. Note that no story renders `DateFormat`, so VRT does not actually cover this change — see Screenshots / Preview

### DIFF
--- a/packages/system/src/components/Formatters/DateFormat.test.tsx
+++ b/packages/system/src/components/Formatters/DateFormat.test.tsx
@@ -33,7 +33,7 @@ test('allows opting out of tabular digits', () => {
 
   expect(date).not.toHaveClass('tabular-nums');
   // No empty `class=""` left behind when the only class is opted out of.
-  expect(date).not.toHaveAttribute('class');
+  expect(date).not.toHaveAttribute('class', '');
 });
 
 test('formats a date range', () => {
--- a/packages/system/src/components/Formatters/DateFormat.test.tsx
+++ b/packages/system/src/components/Formatters/DateFormat.test.tsx
@@ -12,6 +12,30 @@ test('supports formatting date based on specific locale', () => {
   expect(date).toBeInTheDocument();
 });
 
+test('uses tabular digits by default', () => {
+  render(
+    <I18nProvider locale="ru-RU">
+      <DateFormat value={new Date('2021-11-07T07:45:00Z')} />
+    </I18nProvider>
+  );
+
+  expect(screen.getByText('07.11.2021')).toHaveClass('tabular-nums');
+});
+
+test('allows opting out of tabular digits', () => {
+  render(
+    <I18nProvider locale="ru-RU">
+      <DateFormat value={new Date('2021-11-07T07:45:00Z')} tabular={false} />
+    </I18nProvider>
+  );
+
+  const date = screen.getByText('07.11.2021');
+
+  expect(date).not.toHaveClass('tabular-nums');
+  // No empty `class=""` left behind when the only class is opted out of.
+  expect(date).not.toHaveAttribute('class');
+});
+
 test('formats a date range', () => {
   render(
     <I18nProvider locale="de-De">
--- a/packages/system/src/components/Formatters/DateFormat.tsx
+++ b/packages/system/src/components/Formatters/DateFormat.tsx
@@ -12,7 +12,11 @@ export interface DateFormatProps extends DateFormatterOptions {
   tabular?: boolean;
 }
 
-export const DateFormat = ({ value, tabular, ...props }: DateFormatProps) => {
+export const DateFormat = ({
+  value,
+  tabular = true,
+  ...props
+}: DateFormatProps) => {
   const formatter = useDateFormatter({
     ...props,
   });
@@ -31,7 +35,10 @@ export const DateFormat = ({ value, tabular, ...props }: DateFormatProps) => {
     // composed inside it. That child is entirely formatter output, which is
     // exactly what varies. Keep it that way — wrapping the value in another
     // element would move the real output out from under the suppression.
-    <span suppressHydrationWarning className={tabular ? 'tabular-nums' : ''}>
+    <span
+      suppressHydrationWarning
+      className={tabular ? 'tabular-nums' : undefined}
+    >
       {Array.isArray(value)
         ? formatter.formatRange(value[0], value[1])
         : formatter.format(value)}
--- a/.changeset/dateformat-tabular-default.md
+++ b/.changeset/dateformat-tabular-default.md
@@ -1,0 +1,18 @@
+---
+'@marigold/system': patch
+---
+
+fix(DST-1666): `DateFormat` applies its documented `tabular` default.
+
+**What changed:**
+
+- `tabular` now defaults to `true`, so `<DateFormat value={date} />` renders with `tabular-nums`. Previously the prop was destructured without a default, so it was `undefined` unless passed explicitly and the class was never applied.
+- Opting out with `tabular={false}` no longer leaves an empty `class=""` attribute behind, matching `NumericFormat`.
+
+**Why:**
+
+Three places already documented the default as `true` — the `@default true` JSDoc that feeds the docs site props table, the prose in the `DateFormat` docs ("To disable it, set `tabular={false}`"), and `NumericFormat`, which has always defaulted it correctly. Only the implementation disagreed, so consumers reading the docs and relying on the default got misaligned digits in table columns with no indication why.
+
+**Impact:**
+
+This is a visible rendering change anywhere `DateFormat` is used without an explicit `tabular` prop, which in practice is everywhere — no call site in this repo passed it. Digits shift to uniform width, which is what the documentation always promised. Pass `tabular={false}` to keep proportional digits.
--- a/docs/content/components/formatters/dateformat/index.mdx
+++ b/docs/content/components/formatters/dateformat/index.mdx
@@ -42,7 +42,7 @@ For accessibility, screen readers will treat the formatted range as one continuo
 
 ### Tabular display
 
-Use the `tabular` prop to align dates in columns or tables. When enabled, each digit has the same width, keeping rows straight. This is ideal for data tables, reports, or any layout where vertical alignment helps. To disable it, set `tabular={false}`.
+Use the `tabular` prop to align dates in columns or tables. When enabled (the default), each digit has the same width, keeping rows straight. This is ideal for data tables, reports, or any layout where vertical alignment helps. To disable it, set `tabular={false}`.
 
 <ComponentDemo name="dateformat-tabular-format" />
 
--- a/packages/system/src/components/Formatters/NumericFormat.test.tsx
+++ b/packages/system/src/components/Formatters/NumericFormat.test.tsx
@@ -12,6 +12,34 @@ test('renders Currency', () => {
   expect(currency).toBeInTheDocument();
 });
 
+test('uses tabular digits by default', () => {
+  render(
+    <I18nProvider locale="en-US">
+      <NumericFormat style="currency" currency="USD" value={20} />
+    </I18nProvider>
+  );
+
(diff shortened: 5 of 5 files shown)


packages/system/src/components/Formatters/DateFormat.tsx
```diff
@@ -31,7 +35,10 @@ export const DateFormat = ({ value, tabular, ...props }: DateFormatProps) => {
     // composed inside it. That child is entirely formatter output, which is
     // exactly what varies. Keep it that way — wrapping the value in another
     // element would move the real output out from under the suppression.
-    <span suppressHydrationWarning className={tabular ? 'tabular-nums' : ''}>
+    <span
+      suppressHydrationWarning
+      className={tabular ? 'tabular-nums' : undefined}
```
aromko, line 40: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files but the `Visual-Regression-Tests` workflow has not run for this branch (`gh run list --workflow=visual-regression-tests.yml --branch=dst-1666-dateformat-tabular-default` returns no runs).

Blocking on this rather than nitpicking, since per the PR description this changes rendering at ~38 call sites and Chromatic is the only thing that meaningfully reviews it. Please trigger it by commenting `/run-chromatic` on this PR, or by manually dispatching the workflow.

packages/system/src/components/Formatters/DateFormat.test.tsx
```diff
@@ -12,6 +12,30 @@ test('supports formatting date based on specific locale', () => {
...
+    <I18nProvider locale="ru-RU">
+      <DateFormat value={new Date('2021-11-07T07:45:00Z')} tabular={false} />
+    </I18nProvider>
+  );
+
+  const date = screen.getByText('07.11.2021');
+
+  expect(date).not.toHaveClass('tabular-nums');
+  // No empty `class=""` left behind when the only class is opted out of.
+  expect(date).not.toHaveAttribute('class');
```
aromko: 💡 `not.toHaveAttribute('class')` asserts the absence of the attribute entirely, so this test also fails if any unrelated class is ever added to that span. The `not.toHaveClass('tabular-nums')` above already covers the `tabular` contract.

Keeping it is defensible, dropping the empty `class=""` is explicitly in DST-1666's scope, so there's a reason to pin it. Just flagging that the coupling is intentional. Same applies to `NumericFormat.test.tsx:40`.

aromko (commented):
## Automated Code Review

Verified the premise and the blast radius rather than taking the description on faith:

- `@default true` JSDoc, the prose docs, and `NumericFormat`'s `tabular = true` all agreed; only the destructure disagreed. `NumericFormat.tsx` at head already had the `undefined` falsy branch, so `DateFormat` now genuinely matches its sibling.
- The impact claim checks out: 40 `<DateFormat` occurrences at `30f1d6c`, exactly two involving `tabular` (the demo and the new test). So ~38 call sites change rendering.
- `docs/.../dateformat/index.mdx:45` now matches `numericformat/index.mdx:45` word for word.
- `patch` on `@marigold/system` is fine, `.changeset/pre.json` is in beta pre-mode and system is already `18.0.0-beta.4`, so the bump level is moot.
- Tests follow the repo conventions (Vitest, no `fireEvent`, no `getByTestId`, one AAA per test). CI fully green.

One blocker: **Chromatic has not run on this branch.** You flagged this yourself, and it's the right thing to gate on, a wide but intended date diff is exactly the case that needs a human looking at it rather than a green check. Comment `/run-chromatic` and this is an approve.

One optional note left inline on the `not.toHaveAttribute('class')` assertions.

Separately, and unrelated to this diff: I'll add a line for this change to the v18 migration guide on `DST-1570_v18-migration-guide`, since it lands as a visible change for every consumer who omits the prop. Nothing needed from you.

---

sebald:
/run-chromatic

Triggering VRT as requested. Dispatched against `a31a2d916` so it covers the test tweak from the second comment too.

Reminder for whoever reviews the diff: it is expected to be wide. Every `DateFormat` call site in the repo omits `tabular`, so ~38 of them shift from proportional to `tabular-nums` digits. Date digit widths changing is the intended result; anything *else* moving is not.

packages/system/src/components/Formatters/DateFormat.test.tsx
sebald: Fair point — tightened it rather than keeping it as-is, in `a31a2d916`.

`not.toHaveAttribute('class', '')` passes when the attribute is absent *or* has a non-empty value, so it pins exactly what DST-1666 covers (no empty `class=""` left behind) without also pinning "this span has no classes at all". Confirmed it still catches the regression: reverting `DateFormat.tsx` to `: ''` fails the assertion, and adding an unrelated class would not.

Applied to `NumericFormat.test.tsx` as well.

packages/system/src/components/Formatters/DateFormat.tsx
sebald: ✅ **VRT ran** — [run 30534579242](https://github.com/marigold-ui/marigold/actions/runs/30534579242), Chromatic [build 159](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=159). Dispatched against `a31a2d916` so it also covers the test tweak from your other comment. Workflow green; 395 snapshots, **4 changes pending review**.

But the result invalidates the premise we were both gating on — mine in the PR description, yours in this comment. **Chromatic does not cover this change at all, and those 4 changes are not from this diff.**

`.storybook/main.ts` only globs stories from `packages/components/src` and `packages/system/src`. Every one of the ~38 `DateFormat` call sites is under `docs/` — examples, demos, MDX. Not one `*.stories.tsx` renders `<DateFormat>`:

- the five story files that mention the formatters (`Table`, `legacy/Table`, `Panel`, `ActionBar`, `Calendar`) use `NumericFormat` exclusively — 16 usages, zero `DateFormat`
- `Calendar.stories.tsx`'s hit is `DateFormatter`, the Intl wrapper from `@react-aria/i18n`, unrelated to our component
- `NumericFormat` already had `tabular = true` and the `undefined` falsy branch at head, so this PR changes its rendering not at all — only its test file

So the expected snapshot delta from this PR is zero, and the 4 pending changes are pre-existing baseline drift. Consistent with what the CLI reported: it couldn't resolve the commits for builds #157 and #158, found 4 parent builds, and `.storybook/main.ts` forced a full rebuild. The PR also targets `beta-release` rather than `main`. Someone still needs to look at those 4 and confirm they're unrelated, but they shouldn't be attributed here.

**The real check is the docs preview, which is already deployed.** [dateformat-tabular-format](https://marigold-docs-git-dst-1666-dateformat-tabular-default-marigold.vercel.app/components/formatters/dateformat) for the toggle, and the [billing](https://marigold-docs-git-dst-1666-dateformat-tabular-default-marigold.vercel.app/examples/billing) / [users](https://marigold-docs-git-dst-1666-dateformat-tabular-default-marigold.vercel.app/examples/users) examples for dates in real table columns — that's where the ~38 call sites actually live and where the digit-width change is visible.

My description overstated Chromatic's reach ("Chromatic is the real check here"), which is what sent you here. Worth noting for the v18 migration guide too: a consumer-visible change with no Storybook coverage means no VRT safety net on it, now or in future regressions.

aromko (approved):

packages/system/src/components/Formatters/DateFormat.tsx
sebald: Reviewed [build 159](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=159) — the 4 pending changes are fine and unrelated to this diff, as expected from the coverage analysis above. Nothing here is caused by the `tabular` default.

Also corrected the PR description, since its "Chromatic is the real check here" framing is what put you on this thread in the first place: the Screenshots / Preview and Breaking Changes sections now state that VRT does not reach `DateFormat`, and the checklist item is ticked with the build link and that caveat.

This is unblocked from my side. @aromko over to you.

PR #5688, opened 2026-07-30
fix(DST-1656): resolve literals in JSX expression containers

aromko:
# Description

`reportJsxUsage` handles value-conditional warnings — entries like `Select[width]` that should only fire for one specific value. Its value check only read a bare `StringLiteral`:

```ts
const literal = value?.type === 'StringLiteral' ? value.value ?? null : null;
```

Anything wrapped in braces produced `null`, which the surrounding logic treats as "cannot be ruled out statically" and warns on deliberately. So every `width={20}` looked indistinguishable from `width={someVar}`.

Running `migrate v18` over a real consumer app surfaced how noisy that is: **16 `width="fit"` warnings across `Select`, `ComboBox` and `Autocomplete`, and not one of the sites used `fit`** — they were `width={20}`, `width={48}`, `width={96}`, `width={'1/2'}` and similar. The real `width="fit"` usages in that app were all on `NumberField`, `DatePicker`, `TextField`, `Switch` and `Scrollable`, where v18 still accepts it, and were correctly left alone.

16 false positives out of 21 total warnings is enough noise to make people stop reading the report, which defeats the point of the manual-decision warnings.

`staticAttrValue` unwraps `JSXExpressionContainer` and resolves `StringLiteral` and `NumericLiteral`, so a literal in braces is as knowable as a bare one. The deliberate warn-when-unknowable behaviour is untouched.

Relates to DST-1656 — found while migrating the core apps onto the v18 beta. The false positives have no dedicated ticket; happy to file one if you'd prefer the scope changed.

## Test Instructions

1. `pnpm vitest run src/lib/codemod/` in `packages/cli` — 72 pass.
2. New case `stays quiet for literals wrapped in an expression container` covers `width={20}`, `width={'1/2'}` (silent) and `width={'fit'}` (warns).
3. The pre-existing `warns when a conditional value cannot be verified statically` case (`<Switch size={size} />`, an `Identifier`) still passes — that is the regression guard for the warn-on-unknowable path.

## Breaking Changes

No. Report-only codemod; it can only remove spurious warnings, never suppress a real one, since a value that cannot be resolved still warns.

## Checklist

- [ ] Storybook preview and Marigold docs preview are available — n/a, CLI only
- [ ] Stories added/updated — n/a, no component change
- [ ] Component documentation added/updated — n/a
- [ ] Accessibility reviewed against ARIA APG — n/a
- [ ] Visual regression tests updated — n/a, no UI change

### DIFF
--- a/packages/cli/src/lib/codemod/jsx.test.ts
+++ b/packages/cli/src/lib/codemod/jsx.test.ts
@@ -327,6 +327,17 @@ export const B = () => <Select width="full" />;
     expect(warnings[0]).toContain('width="fit"');
   });
 
+  test('stays quiet for literals wrapped in an expression container', () => {
+    const source = `import { Select } from '@marigold/components';
+export const A = () => <Select width={20} />;
+export const B = () => <Select width={'1/2'} />;
+export const C = () => <Select width={'fit'} />;
+`;
+    const warnings = warningsOf(reportJsxUsage(v18).apply(source));
+    expect(warnings).toHaveLength(1);
+    expect(warnings[0]).toContain('width="fit"');
+  });
+
   test('warns when a conditional value cannot be verified statically', () => {
     const source = `import { Switch } from '@marigold/components';
 export const App = ({ size }: { size: string }) => <Switch size={size} />;
--- a/packages/cli/src/lib/codemod/jsx.test.ts
+++ b/packages/cli/src/lib/codemod/jsx.test.ts
@@ -327,14 +327,15 @@ export const B = () => <Select width="full" />;
     expect(warnings[0]).toContain('width="fit"');
   });
 
-  test('stays quiet for literals wrapped in an expression container', () => {
+  test('resolves literals in an expression container, warns on dynamic ones', () => {
     const source = `import { Select } from '@marigold/components';
 export const A = () => <Select width={20} />;
 export const B = () => <Select width={'1/2'} />;
 export const C = () => <Select width={'fit'} />;
+export const D = ({ w }: { w: string }) => <Select width={w} />;
 `;
     const warnings = warningsOf(reportJsxUsage(v18).apply(source));
-    expect(warnings).toHaveLength(1);
+    expect(warnings).toHaveLength(2);
     expect(warnings[0]).toContain('width="fit"');
   });
 
--- a/packages/cli/src/lib/codemod/primitives/jsx.ts
+++ b/packages/cli/src/lib/codemod/primitives/jsx.ts
@@ -31,6 +31,21 @@ const jsxAttributes = (opening: AnyNode | undefined): AnyNode[] =>
 const attrName = (attr: AnyNode): string | null =>
   (attr.name as { name?: string } | undefined)?.name ?? null;
 
+/**
+ * The statically known value of a JSX attribute, or null when it cannot be
+ * resolved. A literal wrapped in an expression container (`width={20}`,
+ * `width={'1/2'}`) is just as knowable as a bare `width="fit"`.
+ */
+const staticAttrValue = (value: AnyNode | undefined): string | null => {
+  const node =
+    value?.type === 'JSXExpressionContainer'
+      ? (value.expression as AnyNode | undefined)
+      : value;
+  return node?.type === 'StringLiteral' || node?.type === 'NumericLiteral'
+    ? String((node as { value: string | number }).value)
+    : null;
+};
+
 /** local names for the manifest's component names, only when imported */
 export const localsFor = (file: AnyNode, components: Iterable<string>) => {
   const wanted = new Set(components);
@@ -543,12 +558,10 @@ export const reportJsxUsage = (manifest: MigrationManifest): Codemod =>
           if (entry.component !== component || !entry.prop) continue;
           for (const attr of jsxAttributes(opening)) {
             if (attrName(attr) !== entry.prop) continue;
-            const value = attr.value as AnyNode | undefined;
             if (entry.value !== undefined) {
-              const literal =
-                value?.type === 'StringLiteral'
-                  ? ((value as { value?: string }).value ?? null)
-                  : null;
+              const literal = staticAttrValue(
+                attr.value as AnyNode | undefined
+              );
               // a non-literal value cannot be ruled out statically: warn too
               if (literal !== null && literal !== entry.value) continue;
             }
--- a/packages/cli/src/lib/codemod/primitives/jsx.ts
+++ b/packages/cli/src/lib/codemod/primitives/jsx.ts
@@ -42,7 +42,7 @@ const staticAttrValue = (value: AnyNode | undefined): string | null => {
       ? (value.expression as AnyNode | undefined)
       : value;
   return node?.type === 'StringLiteral' || node?.type === 'NumericLiteral'
-    ? String((node as { value: string | number }).value)
+    ? String(node.value)
     : null;
 };
 
--- a/packages/cli/src/lib/codemod/primitives/jsx.ts
+++ b/packages/cli/src/lib/codemod/primitives/jsx.ts
@@ -33,8 +33,9 @@ const attrName = (attr: AnyNode): string | null =>
 
 /**
  * The statically known value of a JSX attribute, or null when it cannot be
- * resolved. A literal wrapped in an expression container (`width={20}`,
- * `width={'1/2'}`) is just as knowable as a bare `width="fit"`.
+ * resolved. A string or number literal wrapped in an expression container
+ * (`width={20}`, `width={'1/2'}`) is just as knowable as a bare `width="fit"`.
+ * Everything else, template literals included, stays unresolved and warns.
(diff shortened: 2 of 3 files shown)


packages/cli/src/lib/codemod/primitives/jsx.ts
```diff
@@ -31,6 +31,21 @@ const jsxAttributes = (opening: AnyNode | undefined): AnyNode[] =>
...
+ */
+const staticAttrValue = (value: AnyNode | undefined): string | null => {
+  const node =
+    value?.type === 'JSXExpressionContainer'
+      ? (value.expression as AnyNode | undefined)
+      : value;
+  return node?.type === 'StringLiteral' || node?.type === 'NumericLiteral'
+    ? String((node as { value: string | number }).value)
+    : null;
+};
```
sebald, line 48: 🔴 **Typecheck fails on this cast** — CI is red on exactly line 45:

```
error TS2352: Conversion of type 'AnyNode' to type '{ value: string | number; }' may be a mistake
because neither type sufficiently overlaps with the other.
  Property 'value' is missing in type 'AnyNode' but required in type '{ value: string | number; }'.
```

`AnyNode` (`packages/cli/src/lib/tsx-ast.ts:8`) declares `value` nowhere, so a cast to a type with a **required** `value` is not a permitted narrowing. Every other cast in this file uses an optional property for precisely this reason (`{ name?: string }` at :32, `{ value?: string }` at :39, :292, :358, :421) — including the line this PR replaces.

Making it optional fixes the error, and guarding `undefined` avoids `String(undefined) === "undefined"` sneaking through as a bogus non-null literal. Returning `null` on a malformed node is also the safe direction here: unknowable means warn.

```suggestion
const staticAttrValue = (value: AnyNode | undefined): string | null => {
  const node =
    value?.type === 'JSXExpressionContainer'
      ? (value.expression as AnyNode | undefined)
      : value;
  if (node?.type !== 'StringLiteral' && node?.type !== 'NumericLiteral') {
    return null;
  }
  const literal = (node as { value?: string | number }).value;
  return literal === undefined ? null : String(literal);
};
```

packages/cli/src/lib/codemod/primitives/jsx.ts
```diff
@@ -31,6 +31,21 @@ const jsxAttributes = (opening: AnyNode | undefined): AnyNode[] =>
 const attrName = (attr: AnyNode): string | null =>
   (attr.name as { name?: string } | undefined)?.name ?? null;
 
+/**
+ * The statically known value of a JSX attribute, or null when it cannot be
+ * resolved. A literal wrapped in an expression container (`width={20}`,
```
sebald: 💡 A no-expression `TemplateLiteral` stays unresolved: ``width={`fit`}`` and ``width={`1/2`}`` still fall into the warn branch. Rare in real code, and erring toward a warning is the safe default — fine to leave.

But the doc comment reads as though literals-in-braces are handled in general, so either narrow the wording or handle it: `node.type === 'TemplateLiteral' && node.expressions.length === 0`, reading `quasis[0].value.cooked`.

packages/cli/src/lib/codemod/jsx.test.ts
```diff
@@ -327,6 +327,17 @@ export const B = () => <Select width="full" />;
     expect(warnings[0]).toContain('width="fit"');
   });
 
+  test('stays quiet for literals wrapped in an expression container', () => {
+    const source = `import { Select } from '@marigold/components';
+export const A = () => <Select width={20} />;
+export const B = () => <Select width={'1/2'} />;
+export const C = () => <Select width={'fit'} />;
```
sebald, line 334: 💡 Optional: the regression guard for "still warns when unknowable *through* a container" lives in a separate test three cases down (`<Switch size={size} />`). Adding the dynamic counterpart here would keep both halves of the contract visible in one place:

```tsx
export const D = () => <Select width={w} />;
```

…with `toHaveLength(2)`. Coverage is already there either way.

sebald (changes requested):
## Automated Code Review

Diagnosis and fix shape are right: `staticAttrValue` unwrapping `JSXExpressionContainer` before resolving `StringLiteral`/`NumericLiteral` is the missing step, `String()` coercion is safe against the manifest (every `JsxUsageWarning.value` in `v18.ts` is a string: `'fit'`, `'large'`), and the node type names match the parser (`@babel/parser`, `tsx-ast.ts:16`). It also silently improves `Switch size={'large'}`.

The warn-on-unknowable path is genuinely preserved: the pre-existing `<Switch size={size} />` case now flows *through* the new unwrapping and lands on an `Identifier`, so it is a real regression guard for the new code path, not just the old one.

Base branch `beta-release` is correct — the `marigold migrate` codemods landed there in `c3fde8ea6` (DST-1543) and are not on `main` yet.

**Linked ticket** — DST-1656 "Update core apps to beta.4" (In Review). The ticket is the rollout, not this bug; the PR description says so. Legitimate spin-off of a report-only false positive in the very tool the rollout uses — no separate ticket needed in my view.

**Visual regression** — not needed. CLI-only change; no `packages/components`, `packages/system`, `themes`, `.storybook`, stories or styles touched.

### Recommendation

**Request Changes**, solely for the typecheck failure at `jsx.ts:45` (one-line fix, see inline). Everything else about the change is sound: correct root cause, right abstraction, honest changeset matching repo convention, and tests covering both the silent and warning paths. Happy to approve once CI is green.

---

packages/cli/src/lib/codemod/primitives/jsx.ts
aromko: This one does not match the diff. There is no cast at `jsx.ts:45` — the line is `String(node.value)`:

```ts
return node?.type === 'StringLiteral' || node?.type === 'NumericLiteral'
  ? String(node.value)
  : null;
```

`AnyNode` (`packages/cli/src/lib/tsx-ast.ts:8`) is intersected with `{ [key: string]: unknown }`, so `node.value` is `unknown` and `String(unknown)` is legal — no narrowing involved, so TS2352 cannot apply.

Typecheck is green on `4b7e633` ([run 30522828108](https://github.com/marigold-ui/marigold/actions/runs/30522828108)), and `tsconfig.check.json` includes `**/*.ts` with no `packages/cli` exclusion, so this file is in that run. Reproduced locally as well: `pnpm typecheck:only` exits clean on the branch.

`String(undefined)` cannot reach it either — the type guard runs first, and Babel's `StringLiteral`/`NumericLiteral` always carry `.value`.

Leaving the code as is. Happy to revisit if you can point at a red run.

packages/cli/src/lib/codemod/primitives/jsx.ts
aromko: Good catch on the wording. Narrowed the comment rather than handling the case — 26e710be0:

```
 * The statically known value of a JSX attribute, or null when it cannot be
 * resolved. A string or number literal wrapped in an expression container
 * (`width={20}`, `width={'1/2'}`) is just as knowable as a bare `width="fit"`.
 * Everything else, template literals included, stays unresolved and warns.
```

Resolving ``width={`fit`}`` needs an extra guard plus casts for `quasis`/`expressions`, for a form our Prettier config rewrites to `'fit'` anyway, and whose failure mode is one extra warning — the safe direction. Not worth the code until a real app reports one.

packages/cli/src/lib/codemod/jsx.test.ts
aromko: Added in 26e710be0, and renamed the test since it now asserts both halves:

```tsx
test('resolves literals in an expression container, warns on dynamic ones', () => {
  const source = `import { Select } from '@marigold/components';
export const A = () => <Select width={20} />;
export const B = () => <Select width={'1/2'} />;
export const C = () => <Select width={'fit'} />;
export const D = ({ w }: { w: string }) => <Select width={w} />;
`;
  const warnings = warningsOf(reportJsxUsage(v18).apply(source));
  expect(warnings).toHaveLength(2);
  expect(warnings[0]).toContain('width="fit"');
});
```

`pnpm vitest run src/lib/codemod/` — 72 pass.

packages/cli/src/lib/codemod/primitives/jsx.ts
sebald: You're right, and my finding was wrong — apologies for the noise. I read a cast that isn't there; `String(node.value)` on an `AnyNode` whose index signature yields `unknown` needs no narrowing, so TS2352 can't apply. Verified Typecheck green on `26e710be`. Nothing to change here.

sebald (approved):
All three comments handled — thanks for the pushback on the first one, which was my error.

- `staticAttrValue` cast concern: retracted, no cast exists and Typecheck is green.
- Template-literal wording: comment narrowed in `26e710be`, which is the outcome I said was fine.
- Dynamic-counterpart test: added in `26e710be`, both halves of the contract now visible in one case.

CI green across all 17 checks. Good to merge.

PR #5690, opened 2026-07-30
feat(DST-1600): align the CLI with clig.dev and make telemetry identifier-free

jim761:
# Description

Two related pieces of CLI work. The first is the clig.dev alignment DST-1600 asks for. The second came out of it: writing the first-run telemetry notice made it clear the notice would have been lying, because the event carried `args` and a persistent per-machine UUID while the notice claimed neither. Fixing that honestly meant removing the identifier rather than rewording the notice.

**clig.dev alignment**

- **First-run telemetry notice.** Telemetry is opt-out, so the CLI now prints a one-time disclosure explaining what is collected and how to opt out (`marigold telemetry disable` or `DO_NOT_TRACK=1`). It goes to `stderr`, so `--format json` and stdout stay clean. Mirrors the .NET SDK model.
- The "shown" flag is persisted only after the notice actually prints on an interactive terminal, so agent and pipe runs never consume it before a human has seen it.
- The invocation that shows the notice is itself **not** tracked. No data leaves the machine before the user has seen the disclosure and had a chance to opt out. Tracking begins on the next run.
- **"Did you mean?" suggestions.** Unknown commands now suggest the nearest valid one by edit distance (`marigold serach` → *Did you mean "search"?*), matching the error-recovery pattern in `git`, `cargo` and `npm`.

**Identifier-free telemetry**

- **Removed `anonymousId`.** The persistent per-machine UUID is gone from the event payload, from `UserConfig` and from the server schema. A stale value written by an older CLI is stripped on the next config read and erased from disk.
- **Corrected the first-run notice**, which had claimed no arguments and no personal data were collected.
- **Clamped `args`.** Enum flags are recorded as their validated value or as `invalid`, so a mistyped `--format=jsonn` is no longer echoed back verbatim. Positionals are forwarded only when identifier-shaped. Free-text search terms stay recorded as `used`.
- **Hour-granular server timestamps.** `receivedAt` is rounded to the hour for CLI events, so they cannot be stitched into per-session sequences by timing once the identifier is gone. MCP events keep full precision: they are already tied to a hashed caller.
- **Strict event schema.** The endpoint rejects unknown keys instead of stripping them, so an identifying field can never be accepted quietly.
- **Full disclosure in the docs.** The `marigold telemetry` section of the CLI page now documents every field sent, what is never sent, why the default is opt-out, and every way to turn it off. It lives there rather than on a standalone privacy page nobody would navigate to. `marigold --help` links the anchor, so non-interactive users have a discoverable disclosure too.

Why this makes the opt-out default defensible: with no identifier there is no personal data to establish a lawful basis for under GDPR, and the CLI no longer writes an identifier onto the user's device, which would otherwise need consent under ePrivacy Art. 5(3) / § 25 TDDDG. The trade is deliberate. We can count invocations, not people, and unique-user numbers come from public npm download counts instead.

Closes DST-1600

## Reviewer notes

Two things worth a deliberate look rather than a skim.

**The endpoint now 400s on unknown keys instead of stripping them.** `POST /api/telemetry` is public and unauthenticated, so this is a visible behaviour change for anything posting to it. Stripping would have been enough to keep a stray field out of the store, but it is invisible: nobody learns the sender is wrong, and the field returns the moment someone relaxes the schema. A stale CLI still sending `anonymousId` now gets a 400.

**The CLI lost its per-caller rate limit.** `telemetry:rl:cli:{anonymousId}:{date}` is gone, because there is no identifier left to key it on. The endpoint-wide 50000/day ceiling is now the CLI's only bound, which means one looping CLI can spend the whole fleet's daily budget. That is a real loosening, accepted rather than worked around: the alternatives are re-introducing a per-machine identifier, which is the thing being removed, or keying on IP, which puts personal data in Redis. Bounding a single misbehaving caller belongs in WAF rate limiting in front of the route, which needs no identifier in the body. The reasoning is recorded in `docs/app/api/telemetry/README.md`.

This branch also merges `main`, which had refactored telemetry into a shared `record.ts`/`schema.ts` used by `/mcp` and keyed on a **required** `anonymousId`. Taken as-is, every CLI event would have failed the server schema and returned 400, stopping telemetry silently. The merge commit resolves that by porting the identifier-free design into main's structure, and explains each decision.

## Test Instructions

Build the CLI first: `pnpm --filter @marigold/cli build`.

1. `pnpm --filter @marigold/cli test` (916 pass, 12 skipped). This suite has its own `cli-tests` CI job.
2. `pnpm --filter @marigold/docs exec vitest run app/api/telemetry app/mcp` (67 pass), covering the strict schema, the endpoint-wide ceiling and `receivedAt` granularity.
3. First-run notice, on a clean config and a real terminal:

   ```sh
   D=$(mktemp -d)
   MARIGOLD_CONFIG_DIR=$D marigold telemetry status   # prints the notice to stderr
   MARIGOLD_CONFIG_DIR=$D marigold telemetry status   # silent the second time
   cat $D/config.json                                 # only telemetryNoticeShown, no identifier
   ```

   The notice only prints when stderr is a TTY, and only from a command that emits telemetry. `--version` and `--help` never trigger it.

4. Confirm it stays off stdout: append `2>/dev/null` to the first command in step 3 and only the status line remains.
5. Suggestions: `marigold serach` prints `Did you mean "search"?`.
6. Opt-out paths: `marigold telemetry disable`, then `marigold telemetry status`. Separately, `DO_NOT_TRACK=1 marigold telemetry status`.
7. Docs preview: the `marigold telemetry` section of `/getting-started/cli`, which `marigold --help` links under its Telemetry heading.

## Breaking Changes

No, for consumers of `@marigold/cli`. The commands, flags and output shapes are unchanged, and a stale `anonymousId` in an existing config is cleaned up silently rather than erroring.

Two internal behaviour changes, both called out under Reviewer notes: `POST /api/telemetry` rejects unknown keys with 400 instead of stripping them, and the CLI's per-caller rate limit is replaced by the endpoint-wide one.

## Checklist

- [ ] ~~Stories added/updated~~ — not applicable, no component changes
- [x] Component documentation added/updated (if it exists) — the CLI docs page gained a full `marigold telemetry` disclosure section
- [ ] ~~Accessibility reviewed against ARIA APG~~ — not applicable, no interactive components
- [ ] ~~Visual regression tests updated~~ — not applicable, no UI changes
- [x] Changeset added (`pnpm changeset`) — two, one per piece of work


sarahgm:
@jim761 can you make the PR ready, so we can review it while your still here?

aromko (commented):
Review notes as inline comments. Nothing blocking; the three privacy-disclosure ones are the substantive ones.

docs/content/getting-started/cli/index.mdx
```diff
@@ -311,13 +311,60 @@ marigold telemetry disable
...
+| `exitCode`       | `0`                       | Whether the command succeeded                       |
+| `durationBucket` | `100-500`                 | A coarse bucket in ms, never an exact timing        |
+| `cacheHit`       | `true`                    | Whether the local docs cache was used               |
+| `isTTY`          | `false`                   | Whether output went to a terminal or a pipe         |
+| `isAIAgent`      | `true`                    | Whether an agent runtime env var was set            |
+| `receivedAt`     | `2026-07-30T14:00:00Z`    | Added server-side, truncated to the hour            |
+
+`args` keeps only flag presence and values from a closed set. Enum flags (`--format`, `--section`, `--limit`) are recorded as their validated value, or as `invalid` if unrecognised, so a typo is never echoed back. Identifier-shaped positionals (a component name, an example slug, a `--category`) are recorded only when they look like an identifier. **Free-text is never sent**: a `marigold search` query or a `list --search` term is recorded as `used`, so we learn that search was used but not what you searched for.
```
aromko: ⚠️ `--limit` is not clamped like `--format` and `--section`. `marigold.ts:374` sends it through `slugArg`, which forwards anything identifier-shaped, so `marigold search foo --limit abc123` records `limit: "abc123"` verbatim — a typo *is* echoed back.

Either clamp it at the call site:

```ts
...(values.limit
  ? { limit: /^\d+$/.test(values.limit) ? values.limit : 'invalid' }
  : {}),
```

or drop `--limit` from this sentence. Given this PR exists because the old notice contradicted the code, the clamp seems more in keeping.

docs/content/getting-started/cli/index.mdx
```diff
@@ -311,13 +311,60 @@ marigold telemetry disable
...
+
+`receivedAt` is deliberately hour-granular: exact timestamps would let separate events be stitched back into a per-session sequence by timing alone, undoing the point of having no identifier.
+
+#### What is never sent
+
+- Your source code, file contents, file paths, or project names
```
aromko, line 340: ⚠️ `SLUG_PATTERN` (`packages/cli/src/lib/telemetry.ts:51`) permits `/` and `.`, which it needs for slugs like `getting-started/cli` and `form.field`. That also means a relative path matches it: `marigold docs packages/components/src/Button.tsx` records `component: "packages/components/src/Button.tsx"`.

Only paths starting with `.` or `/` collapse to `invalid`. Same issue in the `slugArg` comment on line 49, which claims "a path" is excluded. Either narrow this bullet, or reject extension-shaped values in the pattern.

docs/content/getting-started/cli/index.mdx
```diff
@@ -311,13 +311,60 @@ marigold telemetry disable
...
+- Search queries and any other free-form text you type
+- Your name, email, username, hostname, or any account data
+- Any persistent identifier, of any kind
+- Your IP address is not stored. It's necessarily visible to our host in transit, as with any HTTP request, but it is never written into a telemetry record.
```
aromko: ⚠️ The second half of this bullet is exactly right and provable from the schema. The first sentence is broader: Vercel retains client IPs in its own request logs, so "not stored" is a claim about the platform, not about the telemetry record.

DST-1600 lists *"Server-side review of whether the telemetry endpoint logs/retains IP addresses"* under **Not included**, so this is asserted rather than verified.

```suggestion
- Your IP address is never written into a telemetry record. It's necessarily visible to our host in transit, as with any HTTP request.
```

docs/content/getting-started/cli/index.mdx
```diff
@@ -311,13 +311,60 @@ marigold telemetry disable
...
+
+Events are kept indefinitely, to support long-run usage trends. That is a decision about volume, not about you: with no identifier in a record and timestamps rounded to the hour, a five-year-old event says no more about you than a five-minute-old one.
+
+#### Why it's opt-out
+
+Because there's nothing to consent to storing. We removed the per-machine identifier the CLI used to send, which means we can count **invocations, not people**. Unique-user numbers come from public npm download counts instead. In exchange, the data stops being personal data, the CLI stops writing an identifier onto your device, and no per-user history exists to secure or hand over. An opt-in prompt would also be close to meaningless here: most runs are non-interactive, driven by AI agents in pipelines where no human is at a prompt to ask.
+
+<Callout variant="warning" title="If this ever changes">
```
aromko: 💡 `Callout` takes `type`, not `variant` — `CalloutType = 'info' | 'warn' | 'error' | 'success' | 'warning' | 'idea'`, and `docs/lib/markdown/plugins/simplify-jsx.ts:93` reads `type` too, defaulting to `info`. So this renders as a neutral info box on the site and as ℹ️ in `marigold docs` / MCP output. `SectionMessage` is the one that takes `variant`.

```suggestion
<Callout type="warning" title="If this ever changes">
```

Not yours: `variant="warning"` is already on `main` in four places (`extendTheme`, `tabs`, `usage-with-ai`, and line 299 of this file), all rendering as info. Probably worth its own cleanup ticket.

docs/app/api/telemetry/schema.ts
```diff
@@ -14,7 +19,6 @@ export const CliCommandEventSchema = z.object({
   exitCode: z.number().int().min(-1).max(255),
   cacheHit: z.boolean().optional(),
   args: z.record(z.string(), z.string().max(64)).optional(),
```
aromko: ⚠️ `z.strictObject` guards top-level keys only. `args` is a `z.record` with an unconstrained key type, so `args: { anonymousId: "00000000-0000-4000-8000-000000000000" }` parses cleanly and gets written to the stream.

The README, the changeset and the PR body all say "an identifying field can never be accepted quietly" — true for the top level, not for `args`. Worth bounding the key and the cardinality:

```suggestion
  args: z
    .record(z.string().max(32), z.string().max(64))
    .refine(a => Object.keys(a).length <= 16)
    .optional(),
```

A determined sender can still hide a UUID in a legitimate key (`component` accepts slug-shaped values, and a UUID is slug-shaped), so the real guarantee is client-side. Probably worth scoping the claim to unknown top-level keys wherever it appears.

packages/cli/src/lib/suggest.ts
```diff
@@ -0,0 +1,39 @@
...
+
+// Return the closest candidate to `input`, or undefined when nothing is near
+// enough to be a helpful suggestion. The threshold scales with input length so
+// short words (`lst` → `list`) still match without suggesting wild guesses for
+// long typos.
+export const nearest = (
+  input: string,
+  candidates: readonly string[]
+): string | undefined => {
+  const threshold = Math.max(2, Math.floor(input.length / 2));
```
aromko, line 29: 💡 Two small things on the threshold.

The comment above says it "scales with input length so short words … still match without suggesting wild guesses for long typos", but `Math.max(2, floor(len/2))` grows with length — longer input gets a *more* permissive threshold. The `max(2, …)` floor is what actually makes `lst` → `list` work.

Second, the comparison is case-sensitive: `marigold DOCS` is distance 4 from `docs`, over the threshold of 2, so it gets no suggestion at all. `levenshtein(input.toLowerCase(), candidate)` would catch it, and every candidate is already lowercase.

.changeset/cli-telemetry-identifier-free.md
```diff
@@ -0,0 +1,17 @@
...
+- **Corrected the first-run notice.** It previously claimed that no arguments and no personal data were collected, while the event did carry `args` and a persistent UUID. The notice now states exactly what is sent and links to the new telemetry page.
+- **Clamped `args`.** Enum flags are recorded as their validated value or as `invalid`, so a mistyped `--format=jsonn` is no longer echoed back verbatim. Positionals (component name, example slug, `--category`) are only forwarded when they are identifier-shaped. Free-text search terms remain recorded as `used`.
+- **Hour-granular server timestamps.** `receivedAt` is truncated to the hour, so events cannot be stitched back into per-session sequences by timing once the identifier is gone.
+- **Hour-granular timestamps, not expiry.** Events are kept for long-run usage trends rather than expiring. Rounding `receivedAt` to the hour is what bounds the disclosure instead: with no identifier in a record, an old event says no more about you than a new one. The reasoning is in [`docs/app/api/telemetry/README.md`](../docs/app/api/telemetry/README.md).
```
aromko: 💡 This relative link resolves from `.changeset/`, but changesets moves the text into `packages/cli/CHANGELOG.md` on release, where `../docs/app/…` points at `packages/docs/app/…`. It ships to npm as a dead link — worth an absolute GitHub URL.

Also, this bullet and the one above it are both titled "Hour-granular …" and largely say the same thing. Reads better merged.

docs/content/getting-started/cli/index.mdx
jim761: Clamped it in 9a114359e. It goes through a new `intArg` helper, so anything but a positive integer is recorded as `invalid`.

docs/content/getting-started/cli/index.mdx
jim761: Fixed in 9a114359e: values ending in a known source-file extension now collapse to `invalid`. I used a fixed list rather than any extension so dotted slugs like `form.field` still pass. The comment is updated too.

docs/content/getting-started/cli/index.mdx
jim761: Took your wording, thanks.

docs/app/api/telemetry/schema.ts
jim761: Bounded the keys (32 characters, 16 keys) and scoped the claim to top-level keys in the changeset, the README and the PR body, pointing at the CLI as the real guarantee.

docs/content/getting-started/cli/index.mdx
jim761: Fixed here and on the PowerShell one in the same file. The other three on `main` are worth a separate cleanup ticket.

packages/cli/src/lib/suggest.ts
jim761: Rewrote the comment and lowercased the input, with a test for `DOCS` → `docs`.

.changeset/cli-telemetry-identifier-free.md
jim761: Absolute URL now, and the two bullets are merged.

PR #5691, opened 2026-07-30
fix(DSTSUP-269): wrap long unbreakable option labels in ListBox items

OsamaAbdellateef:
# Description

A listbox option is a two-column grid (`grid-cols-[auto_1fr]`: selection indicator, then label). The `1fr` track's automatic minimum is its content's `min-content` width, so a label with no break opportunity — a long word without spaces — sized that track to the whole word and pushed the option out of a narrow `Select`, `ComboBox` or `ListBox`. Labels containing spaces were never affected: their `min-content` is just the longest word, so they wrapped.

The item now carries `wrap-anywhere` (`overflow-wrap: anywhere`), which both makes the word breakable and lowers its `min-content` contribution, so the track no longer inflates. Measured on a `width={40}` Select: the option went from 431px of content in a 152px list to wrapping cleanly at 150px, and the virtualizer measures the taller row correctly. The mobile tray was checked separately (390px option, no collapse) since `overflow-wrap: anywhere` also affects intrinsic sizing.

Closes DSTSUP-269

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. `pnpm sb` → **Components/Select** → story *Wraps an unbreakable option label inside a narrow list*.
2. Confirm the long label breaks across lines and stays inside the popover, with no horizontal overflow.
3. Narrow the viewport below 640px and reopen: the same label wraps inside the tray.
4. Check **Opens the list with long option labels** is unchanged — labels with spaces still wrap as before.

## Breaking Changes

No

## Checklist


packages/components/src/Select/Select.test.tsx
```diff
@@ -151,3 +151,22 @@ test('updates the controlled `value` when selecting in the tray (small screen)',
...
+// its `min-content` width. A label without a break opportunity — a long word
+// without spaces — therefore sized that track to the whole word and pushed the
+// option out of a narrow Select. The theme's `wrap-anywhere` both makes the word
+// breakable and keeps it from inflating the track. Unit tests load no CSS, so
+// the class is the assertable invariant here; the rendered wrapping was verified
+// in the browser (option content 431px -> 150px inside a 152px list).
+test('applies wrap-anywhere to options so unbreakable labels cannot widen them', async () => {
+  renderWithOverlay(<Basic.Component label="Label" />);
+
+  await user.click(screen.getByRole('button'));
```
sarahgm: can this be a component test?

.changeset/listbox-item-wrap-anywhere.md
```diff
@@ -0,0 +1,7 @@
+---
+'@marigold/theme-rui': patch
+---
+
+fix: wrap long unbreakable option labels inside the list instead of letting them widen the item
```
sarahgm: not our convention how we write changesets. should start with `fix(DSTSUP-269)`

themes/theme-rui/src/components/ListBox.styles.ts
```diff
@@ -19,7 +19,10 @@ export const ListBox: ThemeComponent<'ListBox'> = {
   }),
   item: cva({
     base: [
-      'relative grid grid-cols-[auto_1fr] items-center gap-x-2 rounded-md px-2 py-1.5 text-sm text-foreground max-sm:min-h-11',
+      // `wrap-anywhere`: a label without a break opportunity would otherwise set
+      // the `1fr` track's automatic minimum to its own width and push the item
+      // out of a narrow popover — breaking anywhere keeps it inside the list.
+      'relative grid grid-cols-[auto_1fr] items-center gap-x-2 rounded-md px-2 py-1.5 text-sm text-foreground wrap-anywhere max-sm:min-h-11',
```
sarahgm: 🔴 **The only visual guard for this fix was deleted in the last commit.**

`d3c7ef7` added `Basic.test('Wraps an unbreakable option label inside a narrow list')` with `parameters: { chromatic: { disableSnapshot: false } }`, which is the one mechanism in this repo that can catch this regression. `fd73ab4b` removed all 37 lines of it.

The replacement test's own comment says it: unit tests load no CSS, so nothing in CI renders the wrapped label. A CSS-only fix would ship with zero rendering coverage. Please restore the story.

themes/theme-rui/src/components/ListBox.styles.ts
```diff
@@ -19,7 +19,10 @@ export const ListBox: ThemeComponent<'ListBox'> = {
   }),
   item: cva({
```
sarahgm, line 18: 🔴 **Visual Regression Tests — not run**

`gh run list --workflow=visual-regression-tests.yml --branch=dstsup-269_fix-listbox` returns `[]`, and this PR changes `themes/**` into `main`.

Running it as-is proves little, though, since no remaining story exercises a narrow list with an unbreakable label. Restore the story from the other comment first, then comment `/run-chromatic` on this PR.

themes/theme-rui/src/components/ListBox.styles.ts
```diff
@@ -19,7 +19,10 @@ export const ListBox: ThemeComponent<'ListBox'> = {
...
     base: [
-      'relative grid grid-cols-[auto_1fr] items-center gap-x-2 rounded-md px-2 py-1.5 text-sm text-foreground max-sm:min-h-11',
+      // `wrap-anywhere`: a label without a break opportunity would otherwise set
```
sarahgm: 💡 Three comment lines for one class. House style is a terse one-liner, and the measurement detail already lives in the PR description.

```suggestion
      // `wrap-anywhere`: an unbreakable label would otherwise set the `1fr` track's automatic minimum to its own width.
```

packages/components/src/Select/Select.test.tsx
```diff
@@ -151,3 +151,22 @@ test('updates the controlled `value` when selecting in the tray (small screen)',
...
+
+  const listbox = await screen.findByRole('listbox');
+  const option = within(listbox).getAllByRole('option')[0];
+
+  expect(option).toHaveClass('wrap-anywhere');
```
sarahgm: ⚠️ This couples a test in `@marigold/components` to a string in `@marigold/theme-rui`. It passes if someone later changes the grid tracks while keeping the class, and it fails if the fix legitimately moves to the `label` slot.

Fine as a cheap smoke check if you want to keep it, but the real guard should be the Chromatic story (see the comment on `ListBox.styles.ts`).

sarahgm (commented):
**Automated review — items without a line anchor**

- **PR description is out of sync with the branch.** Test Instructions steps 1-4 point at the story `Wraps an unbreakable option label inside a narrow list`, which commit `fd73ab4b` deleted. The checklist ticks "Stories added/updated" (no story in the diff) and leaves "Unit tests added/updated" unticked (a unit test was added). The before/after screenshot table is empty, which for a visual-only fix is the main thing a reviewer needs.
- **The v18 line still reproduces.** On `beta-release` the item base was rewritten to `'group/option relative grid grid-cols-[auto_1fr] … max-sm:min-h-touch-target'` with no `wrap-anywhere`. `main` is currently an ancestor of `beta-release`, so the next merge conflicts on exactly this line. Please make sure the resolution keeps the class instead of taking beta's side.
- **Worth stating that the width behavior is unchanged.** The ticket title is about the dropdown width adapting to the selected entry. This PR makes long labels readable but leaves the popover capped at the trigger width, and a 431px word in a `width={40}` Select becomes a tall row. Saying so explicitly (and that the virtualizer's absolutely positioned rows rule out growing the popover) should save a round trip with the reporter before DSTSUP-269 leaves Review.

On the fix itself: the diagnosis is right and the utility choice is the non-obvious correct one. `break-words` would not have worked, since `overflow-wrap: break-word` does not lower the min-content contribution and only `anywhere` does. All three components named in the ticket inherit it through the shared `ListBox.styles.item`, and both the virtualized desktop popover and the non-virtualized mobile tray were checked.

---

packages/components/src/Select/Select.test.tsx
```diff
@@ -151,3 +151,22 @@ test('updates the controlled `value` when selecting in the tray (small screen)',
   );
   expect(button).toHaveTextContent('Peach');
 });
+
+// Regression guard for DSTSUP-269: a listbox option is a two-column grid
```
sarahgm: 💡 An 8-line comment block above a 9-line test. Worth trimming to one line and letting the PR description carry the mechanism.

.changeset/listbox-item-wrap-anywhere.md
sarahgm: ⚠️ Changeset bodies carry the ticket scope here. The previous support fix in this same package reads `fix(DSTSUP-266): render the required-field asterisk…` (theme-rui 5.5.1).

```suggestion
fix(DSTSUP-269): wrap long unbreakable option labels inside the list instead of letting them widen the item
```

.changeset/listbox-item-wrap-anywhere.md
```diff
@@ -0,0 +1,7 @@
...
+
+A listbox item is a grid, so a label without any break opportunity (a long word with no spaces) set the automatic minimum of the label track to its own width and pushed the item out of a narrow `Select`, `ComboBox` or `ListBox`. Items now break anywhere, matching how labels with spaces already wrapped.
```
sarahgm: 💡 The ticket names `Select`, `ComboBox` **and `Autocomplete`**. `Autocomplete` does get the fix (it renders the same `ListBox` items), so this is only wording, but the reporter will scan the changelog for their component.

packages/components/src/Select/Select.test.tsx
OsamaAbdellateef: Converted -- removed the standalone unit test and restored the Chromatic story test (`Wraps an unbreakable option label inside a narrow list`) so this is now a component test rather than a unit test. See 8ddbbe133.

.changeset/listbox-item-wrap-anywhere.md
OsamaAbdellateef: Fixed -- the changeset now starts with `fix(DSTSUP-269):`.

themes/theme-rui/src/components/ListBox.styles.ts
OsamaAbdellateef: Restored the story in 8ddbbe133 (`Basic.test('Wraps an unbreakable option label inside a narrow list')`), and dropped the unit test that had replaced it -- the story is once again the only regression guard.

themes/theme-rui/src/components/ListBox.styles.ts
OsamaAbdellateef: Story is restored (see the comment above) and I've triggered the Visual-Regression-Tests workflow on this branch: https://github.com/marigold-ui/marigold/actions/runs/30814563840

themes/theme-rui/src/components/ListBox.styles.ts
OsamaAbdellateef: Trimmed to the suggested one-liner in 8ddbbe133.

packages/components/src/Select/Select.test.tsx
OsamaAbdellateef: Removed this test entirely -- the restored Chromatic story is now the sole guard, so there's no coupled string assertion left.

packages/components/src/Select/Select.test.tsx
OsamaAbdellateef: Moot now -- the test and its comment block were removed in favor of the restored Chromatic story.

.changeset/listbox-item-wrap-anywhere.md
OsamaAbdellateef: Applied -- title now reads `fix(DSTSUP-269): wrap long unbreakable option labels inside the list instead of letting them widen the item`.

.changeset/listbox-item-wrap-anywhere.md
OsamaAbdellateef: Added -- the changeset body now lists `Select`, `ComboBox`, `Autocomplete` and `ListBox`.

sarahgm (approved):
changed the target  - one conflict, otherwise approved

pnpm-workspace.yaml
```diff
@@ -9,3 +9,10 @@ allowBuilds:
   esbuild: true
   sharp: true
   unrs-resolver: true
+overrides:
```
sarahgm: This change looks not related to the actual fix

sarahgm: also the lock file should not update

sarahgm:
@OsamaAbdellateef now conflicts are back

packages/components/src/Select/Select.stories.tsx
```diff
@@ -112,6 +112,24 @@ export const Basic = meta.story({
...
+/**
+ * Clicks an option inside the (virtualized) popover list.
+ *
+ * While the virtualizer scrolls — which it does when it measures its rows and
+ * brings the focused option into view after opening — react-aria sets
+ * `pointer-events: none` on its content wrapper, and `userEvent.click()` refuses
+ * to interact with such an element. Wait for the list to settle first, the way
+ * a user would, instead of clicking into an unsettled list.
+ */
+const clickOption = async (listbox: HTMLElement, name: string | RegExp) => {
```
sarahgm: 🔴 **This wait is a symptom of a user-facing regression, not a test-env quirk.**

The mechanism checks out. `useScrollView` sets `pointerEvents: isScrolling ? 'none' : 'auto'` on the virtualizer content (`react-aria/dist/private/virtualizer/ScrollView.mjs:264`), and `isScrolling` clears only 300 ms after the last scroll (lines 110-118). `pointer-events` is inherited, so every option inherits `none`.

What worries me is *why the list scrolls at all now*. `ListLayout.buildItem` sizes each row from `virtualizer.size.width` (`react-stately/.../ListLayout.mjs:337`), which is 0 on the first layout pass, because layout effects run child-first and the ScrollView measures before RAC's Popover positioning effect has set `--trigger-width`. `useVirtualizerItem.getSize` then blanks the row height and reads `scrollHeight` (`useVirtualizerItem.mjs:38-45`). Before `wrap-anywhere`, a zero-width row still measured about one line, because `min-content` floored at the longest word. Now it measures as a one-character column dozens of lines tall, five rows blow past `maxHeight: 24rem` (`ListBox.tsx:55`), and RAC scrolls the focused option into view.

That is why two tests with short labels (`Star Wars`, `Bob Smith`) broke on a CSS-only change. The same 300 ms window applies to real users: a click landing within ~300 ms of opening a Select, ComboBox, Autocomplete or TagField would be swallowed. Could you check in a real Chromium with CPU throttled 4-6x whether the list flashes or jumps on open, and whether an immediate click registers? If it reproduces, the fix belongs in the measurement path (give the virtualizer's first layout pass a real width rather than `w-full` of a not-yet-positioned popover) instead of in the tests.

packages/components/src/Select/Select.stories.tsx
```diff
@@ -112,6 +112,24 @@ export const Basic = meta.story({
...
+  await waitFor(() => {
+    const option = within(listbox).getByRole('option', { name });
+    expect(getComputedStyle(option).pointerEvents).not.toBe('none');
+  });
+
+  await userEvent.click(within(listbox).getByRole('option', { name }));
```
sarahgm: ⚠️ Residual race: pointer-events is checked, then the click happens in a separate statement. Any re-measure that fires another scroll in between makes this throw again. Retrying the click itself is race-free, since `userEvent.click` throws before dispatching anything when pointer-events is `none`:

```suggestion
  await waitFor(() =>
    userEvent.click(within(listbox).getByRole('option', { name }))
  );
```

packages/components/src/Select/Select.stories.tsx
```diff
@@ -232,6 +247,43 @@ Basic.test(
...
+
+    const listbox = await canvas.findByRole('listbox');
+
+    // The label has no break opportunity, so it has to break mid-word and stay
+    // inside the narrow list instead of widening the option. That is a purely
+    // visual invariant here, guarded by the Chromatic snapshot of the open
+    // list: the virtualizer does not size its rows under the browser test
+    // runner, so per-option widths cannot be asserted in this environment.
+    expect(
+      within(listbox).getByRole('option', { name: /IchliebeDeutschland/ })
```
sarahgm: ⚠️ This assertion passes without the fix too. The option was always visible, it just overflowed the list. So the Chromatic snapshot is the only real guard, and the accepted baseline (build 167, commit `8ddbbe1`) was captured while the story still had `<Inset space={24}>`, which is not an `Inset` prop (`Inset.tsx:10-34` accepts only `p`/`px`/`py`) and rendered with no padding. The follow-up commit changed it to `p={24}`, so that baseline no longer matches what this story renders. Worth a `/run-chromatic`.

packages/components/src/Select/Select.stories.tsx
```diff
@@ -232,6 +247,43 @@ Basic.test(
...
+      </Inset>
+    ),
+  },
+  async ({ canvas }) => {
+    await userEvent.click(canvas.getByLabelText(/Favorite country/i));
+
+    const listbox = await canvas.findByRole('listbox');
+
+    // The label has no break opportunity, so it has to break mid-word and stay
+    // inside the narrow list instead of widening the option. That is a purely
```
sarahgm: 💡 If the virtualizer genuinely does not size rows under the test runner, consider saying outright that this story is snapshot-only, rather than shipping an assertion that reads as a behavioural guard. Or assert the invariant that survives a bad row width, which is the one the bug actually violated: `expect(list.scrollWidth).toBeLessThanOrEqual(list.clientWidth)`.

packages/components/src/Select/Select.stories.tsx
```diff
@@ -232,6 +247,43 @@ Basic.test(
   }
 );
 
+Basic.test(
+  'Wraps an unbreakable option label inside a narrow list',
```
sarahgm, line 238: ⚠️ Worth confirming the resolution with the reporter. DSTSUP-269 frames the problem as the popover never getting wider than the field, whose width depends on the selected entry, with longer entries cut off mid-word. This PR makes labels wrap instead of letting the list grow, and the description states labels with spaces were never affected. Those two framings do not fully overlap. If the reporter's repro was an unbreakable token, this closes it. If it was ordinary long labels, the "dropdown never gets wider than the field" half stays open.

themes/theme-rui/src/components/ListBox.styles.ts
```diff
@@ -17,7 +17,8 @@ export const ListBox: ThemeComponent<'ListBox'> = {
   }),
   item: cva({
     base: [
-      'group/option relative grid grid-cols-[auto_1fr] items-center gap-x-2 rounded-md px-2 py-1.5 text-sm text-foreground max-sm:min-h-touch-target',
+      // `wrap-anywhere`: an unbreakable label would otherwise set the `1fr` track's automatic minimum to its own width.
+      'group/option relative grid grid-cols-[auto_1fr] items-center gap-x-2 rounded-md px-2 py-1.5 text-sm text-foreground wrap-anywhere max-sm:min-h-touch-target',
```
sarahgm, line 21: ⚠️ The snapshot run covered very little of this class's blast radius. Chromatic build 167's log shows TurboSnap reduced it to *"2 story files affected by recent changes"*, capturing 18 snapshots and copying 278. `classNames.item` is also consumed by ComboBox, Autocomplete, TagField (`TagField.tsx:188`) and the Calendar preset rows (`CalendarPresets.tsx:183`, `:249`). Since `wrap-anywhere` lowers the `min-content` of all of those lists, they are exactly the snapshots worth having, and they were copied rather than re-captured.

The placement itself is right, for the record: `SelectionIndicator` renders `display: contents`, so plain-string children are anonymous grid items that no class can target, and only inheritance reaches them. `wrap-anywhere` over `break-words` is right too, since only `overflow-wrap: anywhere` lowers `min-content`.

.changeset/listbox-item-wrap-anywhere.md
```diff
@@ -0,0 +1,7 @@
+---
+'@marigold/theme-rui': patch
+---
+
+fix(DSTSUP-269): wrap long unbreakable option labels inside the list instead of letting them widen the item
+
+A listbox item is a grid, so a label without any break opportunity (a long word with no spaces) set the automatic minimum of the label track to its own width and pushed the item out of a narrow `Select`, `ComboBox`, `Autocomplete` or `ListBox`. Items now break anywhere, matching how labels with spaces already wrapped.
```
sarahgm: 💡 The same `ListBox.item` class is also used by `TagField` and the Calendar preset rows, so consumers of those will see the change too. Worth naming them alongside Select, ComboBox, Autocomplete and ListBox.

sarahgm (changes requested):
Inline findings below. The CSS fix itself is well-reasoned and correctly placed. What holds this up is the virtualizer wait standing in for a real-browser check of the new scroll-on-open behaviour, plus one question about whether wrapping is the resolution DSTSUP-269 asked for.

pnpm-workspace.yaml
OsamaAbdellateef: Already gone — this change isn't part of the branch anymore (not in the diff against `main`). Must have been dropped in a later commit before this review round.

OsamaAbdellateef: Same here — `pnpm-lock.yaml` isn't touched by this branch either.

packages/components/src/Select/Select.stories.tsx
OsamaAbdellateef: Checked this directly — it's real, not a test-env quirk, and it's deterministic in `pnpm test:sb` (Firefox headless via Playwright), no CPU throttling needed.

I added a temporary probe to `Basic`: open the Select, click "Star Wars" immediately with no wait, 15 times in a row. With `wrap-anywhere` present: the very first open succeeds, **every reopen after that fails** — `pointer-events: none`, click throws, listbox stuck open. Removed `wrap-anywhere` (all else identical) and reran the same probe: 15/15 pass, zero flakiness. So the causal link you flagged is confirmed, not speculative.

Your mechanism read is right. `ListLayout.buildItem` (react-stately) sizes each row from `virtualizer.size.width`, which starts at 0 until `ScrollView`'s `ResizeObserver` reports the real `clientWidth` — that's async, so the first commit can paint with a 0-width row. Before this PR, `min-width: auto` on the grid item floored that at the longest word's `min-content`, so a stray 0px assigned width was silently ignored. After `wrap-anywhere`, there's no floor: the row actually renders near-0px, wraps character-by-character, balloons in height, and RAC scrolls the focused option into view — a real DOM `scroll` event, which is what flips `pointer-events: none` for 300ms (react-aria's own cooldown, not new).

I didn't attempt a fix here. The mount-timing issue lives inside react-aria-components' `ScrollView` (ResizeObserver racing layout effects), and closing it properly means changing how the virtualizer gets its first real width across `Select`, `ComboBox`, `Autocomplete` and `TagField` — real scope beyond a `ListBox` CSS bugfix. Filed [DST-1684](https://reservix.atlassian.net/browse/DST-1684) to track it as a follow-up with the full repro and root-cause writeup.

packages/components/src/Select/Select.stories.tsx
OsamaAbdellateef: Applied — extracted into a shared `clickOption` in `.storybook/test-utils.ts` that retries the click itself inside `waitFor`, exactly as suggested. Used by Select, ComboBox, Autocomplete and TagField's `LargeDataset` tests now (972e351f3).

packages/components/src/Select/Select.stories.tsx
OsamaAbdellateef: Fixed — swapped the visibility assertion for `expect(listbox.scrollWidth).toBeLessThanOrEqual(listbox.clientWidth)`, which does fail without `wrap-anywhere`. The Chromatic story stays as the guard for the wrap actually looking right (972e351f3).

.changeset/listbox-item-wrap-anywhere.md
OsamaAbdellateef: Added — the changeset now lists TagField and the Calendar's preset list alongside Select, ComboBox, Autocomplete and ListBox (972e351f3).

packages/components/src/Select/Select.stories.tsx
OsamaAbdellateef: Pulled up DSTSUP-269 directly. The reporter's own description (translated from German): "the expanded dropdown of a Select never gets wider than the selection field itself... longer entries get cut off — not truncated with '…', but in the middle of the word, so they're no longer readable." That's the unbreakable-token symptom this PR fixes, not an ordinary long label — labels with spaces already wrapped instead of getting cut mid-word (per the PR description). Given this is a German product and German is full of long unbroken compound nouns, that tracks as the real-world trigger. The ticket also names exactly `Select`, `ComboBox`, `Autocomplete` as affected, matching this fix's scope.

I'd call the framing confirmed without needing to loop back to the reporter, but flag if you read the ticket differently.

themes/theme-rui/src/components/ListBox.styles.ts
OsamaAbdellateef: Added the same Chromatic-backed "Wraps an unbreakable option label inside a narrow list" story to ComboBox, Autocomplete and TagField (972e351f3), so the next run captures real coverage for all four consumers instead of copying stale snapshots. Left the Calendar preset rows alone — they render through React Aria's raw `ListBox` directly, not Marigold's virtualized wrapper, so there's no measurement/scroll path to regress there, and preset labels are developer-authored rather than user data. Happy to add a story there too if you'd still like the coverage.

packages/components/src/Select/Select.stories.tsx
OsamaAbdellateef: Agreed. Pushed 972e351f3 on top of the Inset prop fix, and posting /run-chromatic on this PR to get a fresh baseline for this story against the current render.

OsamaAbdellateef:
C:/Program Files/Git/run-chromatic

themes/theme-rui/src/components/ListBox.styles.ts
OsamaAbdellateef: Ran `/run-chromatic` on 972e351f3 (Build 173): https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=173 — 29 visual changes to review. Most of that is expected: it's the first real capture of the new ComboBox/Autocomplete/TagField stories (no baseline before), plus every other consumer of `ListBox.item` (TagField, Calendar presets, etc.) reflowing under `wrap-anywhere`. Worth a look before merge to confirm nothing unexpected snuck in.

sarahgm:
> C:/Program Files/Git/run-chromatic

what is this? :D

OsamaAbdellateef:
C:/Program Files/Git/run-chromatic

sarahgm (approved):

PR #4613, opened 2025-05-21
docs([DST-815]): Make RUI the default 

OsamaAbdellateef:
# Description

The documentation site needs to be updated to reflect the current visual direction and design system standards defined by the RUI theme. The transition involves replacing the default styling and structure currently based on the "core" theme with the modernized RUI visual language. This ensures visual consistency across the design and development ecosystem, and enables easier adoption of new patterns and tokens introduced by RUI.

- [ ] This change requires a UI-Kit update - inform @tirado-rx @benjirsvx

# What should be tested?

## Are they some special informations required to test something?

# Reviewers:

Which persons should review this PR? Add person with @mentions

Examples:
@marigold-ui/developer
@marigold-ui/designer

PR #5693, opened 2026-07-31
fix(DST-1672): restructure RUI tokens and fix three token defects

sebald:
# Description

Audit of `themes/theme-rui/src/tokens.css`. The driving question was whether the token API is a contract a *second* theme could implement — specifically a white-label theme that doesn't share RUI's card-and-surface look — or whether it has hardened around the current styling. Along the way the audit turned up three defects.

The public token surface is unchanged: every token a consumer can reference still resolves to the same computed value.

Closes DST-1672

## Defects found and fixed

**1. The pre-compiled bundle force-generated the wrong palette.** `styles.css` declared `@source inline('{bg}-{stone}-…')`, but the theme's palette is `charcoal`. So consumers writing `bg-stone-500` got Tailwind's *default* stone — never a theme color — and `bg-charcoal-*` was absent from the bundle entirely. The bundle now ships the 11 `bg-charcoal-*` utilities.

The line wasn't wrong when it landed: the theme genuinely used Tailwind's default stone via utilities back then (`border-stone-800` survived until DST-878). It went stale at DST-878, when `charcoal` was introduced and the theme's own usages migrated without updating it. There was never a declared `stone` palette to rename — thanks to @sarahgm for catching that my original wording said otherwise.

**2. `<Panel p="collapsed">` referenced an undefined variable.** `collapsed` is a member of `InsetSpacingTokens`, and `resolveInsetAxes` appends an axis suffix to every non-numeric token — so it resolved to `var(--spacing-collapsed-x)`, which was never declared. The declaration was invalid at computed-value time, `padding` is not inherited, so it fell to its initial value and landed on `0`. Right answer, wrong reason, and only by luck. Adds `--spacing-collapsed-x` / `-y`.

This patches the symptom in the theme so the bug is gone now. The real fix belongs in `resolveInsetAxes` and is filed as **DST-1675** — including the warning that `SelectList.tsx` hand-rolls the same logic *deliberately*, with three documented divergences, so the two must not simply be merged.

**3. Docs drift.** `token-overview` listed `disabled-surface` as `charcoal-200`; it is `charcoal-100`.

## Rebrand-safety

`--color-border`, `--color-soft` and all six `--shadow-elevation-overlay` layers hardcoded hue `54` — the palette's hue, copied by hand. All now derive it via relative color syntax (`oklch(from var(--color-charcoal-950) … h)`), so a palette swap carries the divider color, the soft cap and the shadow's warm cast along instead of leaving them behind at the old hue. Computed values today are identical, verified in-browser rather than by inspection. Lightness and chroma stay literal, since those are tuned by eye and deliberately independent of the palette rungs.

Two honest limits on how far this gets you, both raised in review:

- On the **pre-compiled path** `tokens.css` is imported with `theme(inline)`, so palette *utilities* like `bg-charcoal-500` compile to inlined literals rather than `var(--color-charcoal-500)`. Overriding the token at runtime does not retint them. This change makes the *token definitions* palette-derived; it does not make the shipped bundle live-retintable.
- The `@supports` fallback branch that Tailwind emits for `border-border/60` still carries a hardcoded literal and interpolates in `srgb`. Low impact — the guarded branch wins in any browser that supports relative color, which was already required at HEAD — but it is not itself rebrand-safe.

No new browser-support floor: relative color syntax was already required at HEAD (3 uses in `tokens.css`, 9 in `ui.css`).

## Scope reduction

`--contrast-*` (7 vars) sat on the public `:root` surface but is private implementation detail — not in a Tailwind namespace, so it never generated a utility, and nothing outside its own file read it. The contrast deltas move *into* the `ui-contrast` utility that consumes them.

`--ui-panel-px` was also demoted in the first pass, and that was a mistake — see below. It stays in `@theme`.

## Structure and comments

Boundary tokens are grouped so all five edge tokens sit together instead of being spread across three sections; radius is split out of the surface section; `TYPOGRAPHY` → `FONT FAMILIES` (it holds no type scale). Comments were trimmed against a simple test — does this explain something the declaration can't say, or warn about a real trap? Restatements of values went; the load-bearing bits stayed (the deliberate WCAG 1.4.11 deviation on `control-border`, the `collapsed` axis trap). One 13-line "palette step mapping" comment was deleted outright: it was factually wrong about which rung carried which role.

## Review round (20526129d)

**Fixed a regression I introduced.** Moving `--ui-panel-px` out of `@theme` onto a bare `:root` in `ui.css` dropped `:host` and dropped `@layer theme` — and the comment I wrote claiming parity with the `@theme` tokens was false. Reverted to `@theme`.

Worth recording *how* this was verified, because the obvious test hides it: custom properties inherit through the shadow boundary, so any mount where the outer document also loads the theme CSS resolves the token anyway. In a fully encapsulated mount (theme CSS loaded **only** inside the shadow root):

| stylesheet | `ui-panel-content` padding-left | `--ui-panel-px` |
| --- | --- | --- |
| before | `0px` | **unresolved** |
| after | `24px` | `calc(0.25rem * 6)` |

The unlayered half mattered too: an unlayered rule outranks every layered origin, so the token had become unoverridable from a consumer's own `@theme`.

**Rejected one suggestion, with a measurement.** Colocating `@source inline` next to the palette in `tokens.css` would be nicer, but `theme.css` imports `tokens.css` too and Tailwind collects `@source` across the whole import graph — so it forces 11 palette utilities into the tree-shaking path (+812 B) for a consumer who may use none of them. It stays in `styles.css`, now with a note at both ends so a palette rename can't silently miss it.

**Deferred with numbers.** Only `bg-` is force-generated; `text-`/`border-charcoal-*` don't exist in the bundle. Notably the design system itself writes *zero* charcoal utilities — they exist purely as a consumer escape hatch, which makes covering background but not text or border arbitrary. Widening costs +1,861 B (+0.65%), but how wide that hatch should be is an API decision, so it's recorded on DST-1674 rather than expanded here.

## Test Instructions

The interesting verification is that this *doesn't* change rendered output.

1. `pnpm --filter @marigold/theme-rui build`
2. Diff the compiled `dist/styles.css` against a build from the parent commit. A naive `diff` is useless — the whole `:root` block is one enormous line — so normalize: strip comments, collapse whitespace, split on `;`, then diff sorted multisets of declarations *and* selectors separately. Every difference is one of the intended changes above. Bundle 287,175 → 288,471 B (+0.45%).
3. `pnpm typecheck:only` — exit 0.
4. Chromatic is the real confirmation for the derived-color changes.

## Breaking Changes

No, with one caveat worth a reviewer's eye.

Fixing defect #1 means the bundle no longer ships the 9 `bg-stone-*` utilities (`50/100/200/400/500/600/700/900/950`). A consumer who wrote `bg-stone-500` was getting Tailwind's default stone rather than a theme color, so they were already not doing what they thought — but they *were* getting *a* color, and now they'd get none. This is the only consumer-visible removal. Shipping as `patch` because we're in the `beta` pre-release cycle and `tokens.css` doesn't exist on `main` at all.

Follow-ups, none of which this PR depends on:

- **DST-1675** — the `resolveInsetAxes` fix for defect #2, one layer up.
- **DST-1674** — the token-API questions that need breaking changes: the overloaded semantic-color renames, the duplicate `padding-*` / `square-*` spacing families, and the raw-palette escape-hatch decision. Splitting `tokens.css` by tier was evaluated there and **rejected**; the unused easings, orange rungs and charcoal rungs are deliberately **kept**.

## Checklist

- [ ] Stories added/updated (with `component-test` tag where applicable) — n/a, no component behavior changes
- [ ] Unit tests added/updated — n/a; verification is the compiled-CSS diff described above
- [x] Component documentation added/updated (if it exists) — `token-overview` value corrections
- [ ] Accessibility reviewed against ARIA APG — n/a, no interactive components touched
- [x] Visual regression tests updated (for UI changes) — Chromatic triggered on this PR
- [x] Changeset added


sebald:
/run-chromatic

Triggering VRT for the token restructure. The compiled-CSS diff says computed values are unchanged, so the expected result is **zero visual changes** — this run is there to catch anything the static diff can't see, particularly the two colors that moved to relative color syntax (`--color-border` and the six `--shadow-elevation-overlay` layers).

sebald:
### VRT result — 1 change, attribution unclear

Chromatic [build 161](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=161) reports **1 change must be accepted as baseline**. 0 errors, 0 interaction-test failures, 713 stories / 395 tests.

I can't attribute it to this PR, and I don't want to imply it's clean:

- TurboSnap **bailed out** — `bailReason: { changedStorybookFiles: ['.storybook/main.ts'] }` — so it snapshotted everything instead of only what this diff touches.
- The baseline Chromatic chose is `199e06658`, **33 commits behind this PR's parent**. Its `baselineCommits` are `a31a2d9`, `bb4b29e`, `b6fec7ac`, `199e0665` — none of them is `32475579d`, the commit this branch forked from.
- Several of those 33 commits are themselves intentional visual changes: DST-1659 (Checkbox drawn from theme tokens), DST-1661 (menu focus ring), DST-1666 (DateFormat tabular default), the Radio width fix.

So the flagged story may well predate this branch. Whoever reviews: if the changed story is a divider, a border or an overlay shadow, trace it to the relative-color-syntax conversion of `--color-border` / `--shadow-elevation-overlay` in this PR. Anything else is almost certainly pre-existing.

A Chromatic build on `beta-release` would give a clean baseline and settle this automatically — happy to trigger it on request.

themes/theme-rui/src/ui.css
```diff
@@ -314,7 +327,18 @@
...
  * these.
+ *
+ * The shared horizontal padding lives here rather than in tokens.css: `--ui-*`
+ * is not a Tailwind namespace, so it never generated a utility, and the only
+ * readers are the three utilities below plus DrawerContent's `bleed` branch
+ * (which republishes it as `--bleed-px`). `:root` keeps it globally resolvable —
+ * postcss.config.js excludes `:root` from the `[data-theme="rui"]` prefixing,
+ * exactly as it does for the @theme tokens.
  */
+:root {
```
sarahgm: 🔴 **The relocation lost `:host` and lost `@layer theme`.**

Verified by diffing the compiled `dist/styles.css` on both commits, not inferred:

| | base | branch |
|---|---|---|
| selector | `:root, :host` | `:root` |
| layer | inside `@layer theme` | unlayered, line 8299 |

Two consequences:

1. **`:host` is gone.** In a shadow-DOM mount `:root` does not match, so the three `ui-panel-*` utilities and `DrawerContent`'s `--bleed-px` lose their padding, while every `@theme` token still resolves through `:host`. `:host` is in the `postcss.config.js` exclude list deliberately, which suggests that mount is meant to work.
2. **`@layer theme` is gone.** An unlayered rule outranks every layered origin, so `--ui-panel-px` stopped being overridable from a consumer's own `@theme`. It is also now the last rule in the file.

The comment above claims parity: *"`:root` keeps it globally resolvable ... exactly as it does for the @theme tokens."* The compiled output shows that is not the case for either property.

Cheapest fix is to leave the token in `@theme` where it was. The namespace argument is about utility generation, and `--ui-*` generates none either way, so keeping it costs nothing. If you want it to live next to its readers:

```suggestion
@layer theme {
  :root,
  :host {
    --ui-panel-px: --spacing(6);
  }
}
```

sarahgm (commented):
## Automated Code Review

Reviewed against Marigold's conventions plus a `/simplify` pass (reuse, simplification, altitude). Linked ticket: **DST-1672**.

### I reproduced the verification

I rebuilt `dist/styles.css` on `32475579d` (merge base) and on `5f0b8603e` in throwaway worktrees, then diffed both normalized to one declaration per line.

**The central claim holds.** Byte counts reproduce exactly (287,175 → 288,449, matching the PR body). After filtering the intended changes the residual is two lines, both of them `--color-disabled-border` and `--color-surface-border` changing position inside `:root`. Every converted value is computed-identical. That claim is now independently confirmed rather than asserted.

The build also surfaced three things the verification record does not name. They are inline.

### Chromatic attribution, resolved

The changed story is **Calendar → "selecting a preset jumps the visible month to its date"**.

It *does* render `border-border`: `themes/theme-rui/src/components/Calendar.styles.ts:66` puts `sm:border-r sm:border-border` on the preset rail as a full-height divider. So by the criterion in your comment this lands in "trace it", not "almost certainly pre-existing".

Tracing it exonerates the PR. Plain `border-border` compiles to a computed-identical value. The only site whose *output shape* changed is the `/60` variant, and Calendar does not use it. Six commits in the 33-commit baseline gap touch `themes/theme-rui/src` or Calendar, including DST-1661 and DST-990. A Chromatic run on `beta-release` would close this out.

### Should fix before merge

1. **`--ui-panel-px` lost `:host` and lost `@layer theme`** in the relocation. Verified in compiled output, not inferred. See `ui.css`.
2. **`--color-border` no longer guarantees its own opacity.** See `tokens.css`.

Both are one-line fixes.

### Two more the build found

**The new `@supports` wrapper is `border-border/60`.** The ticket dismisses it as "an existing Tailwind pattern". The pattern is, but the reason a *new* one appeared is that `--color-border` stopped being statically resolvable, so one folded declaration became two:

```css
/* fallback  */ color-mix(in srgb, oklch(from oklch(0.15 0.008 54) 0.9 0.004 h) 60%, transparent)
/* @supports */ color-mix(in oklab, oklch(from var(--color-charcoal-950) 0.9 0.004 h) 60%, transparent)
```

The fallback hardcodes the palette literal, so that branch is not rebrand-safe, and it interpolates in `srgb` rather than `oklab`. Harmless where the guarded branch wins. Worth a line in the record since it is the one output-shape change in the diff.

**`bg-charcoal-*` emits inlined literals, not `var()`** (`background-color: oklch(0.62 0.007 54)`), because `styles.css` imports tokens with `theme(inline)`. The removed `bg-stone-*` used `var(--color-stone-500)`. So on the pre-compiled path a runtime palette override does not retint palette utilities. The rebrand win is real but lands only on the derived tokens, which is narrower than the changeset implies.

### /simplify pass

**Altitude**

- The `--spacing-collapsed-x/-y` fix sits one layer below the defect. Detail inline.
- The `@source inline` line was never a stale rename. `git log -S stone -- themes/theme-rui/src/` shows the theme never declared a `stone` palette, so it has been force-generating Tailwind's default stone since DST-815. Detail inline.
- **The hue derivation trades 7 hardcoded `54`s for 10 hardcoded `charcoal`s.** Within-theme robustness improves and a hue tweak now propagates. Coupling to the palette *name* goes up, so a second theme must still name its palette `charcoal` or rewrite all 34 references.
- **The `--contrast-*` move is at the right altitude.** Seven single-use arguments, in a namespace that generated no utilities, moved into their only reader, and it makes `ui-contrast` consistent with `ui-soft`. No change needed.

**Simplification**

- **Comment volume went up, not down.** `tokens.css` 153 → 147, `ui.css` 250 → **262**, net **+6** across the two files. The 13-line palette map deletion is genuine trim, which means the rest of the file grew by roughly the same amount. Worth another pass against the PR's own stated test.
- The two back-references (`"grouped with the boundaries below"` / `"above"`) encode file order and rot silently on any reorder. One replaced a comment stating design intent with a pointer stating nothing.
- The new boundaries preamble restates the `--color-border` declaration comment 20 lines below it. The use-site list and the opacity rule each appear twice. `"the state edges are solid rungs"` is plural but only one state edge is in the section.
- The `--spacing-topbar` / `--spacing-rail*` dependency inventory is already stale on this commit. `AppShell.tsx` never references `--spacing-topbar`, and `Sidebar/SidebarModal.tsx:52` does but is not listed. Keep the intent sentence, drop the inventory.
- Moving `--ui-panel-px` orphaned the "Shared anatomy for sectioned overlay surfaces" comment from `@utility ui-panel-header`.

**Reuse**

- `ui-contrast` names its deltas while `ui-soft` keeps the equivalents inline, and they share literals (`0.2`, and the byte-identical `0 1px 1.5px -1px` geometry). The PR documents the asymmetry rather than removing it. Name both or inline both.
- The six shadow layers repeat `from var(--color-charcoal-950) 0.14 0.035 h` to inherit exactly one channel. A clean hoist is blocked by this PR's own rule against non-namespaced vars in `@theme`, which is worth stating as a decision rather than leaving implicit.
- Hue `54` is still hand-copied in `docs/content/foundations/token-overview/contrast-matrix.demo.tsx:57-74`, which recomputes the **published** contrast numbers, and in `index.mdx:83` prose. A rebrand would silently leave the documented matrix wrong.

### Nits

- The changeset body drops the `fix(DST-1672):` prefix that the rest of `.changeset/` uses.
- The stale locked worktree note in the PR description is a local-environment artifact. It reads oddly in the permanent record and would sit better in a comment.

### Recommendation

Unusually well-evidenced PR, and the non-breaking claim survives independent reproduction. Items 1 and 2 first. Everything under `/simplify` is optional polish except the `resolveInsetAxes` altitude point, which is worth a follow-up on DST-1672 even if the token fix stays as the shipping form.

---

themes/theme-rui/src/tokens.css
```diff
@@ -92,36 +84,44 @@
...
      focus ring stays the perceivable affordance. State overrides stay opaque. */
   --color-control-border: oklch(from var(--color-charcoal-950) l c h / 0.26);
+
+  /* Dimmed rather than absent — a borderless input reads as broken. */
+  --color-disabled-border: var(--color-charcoal-300);
+
+  /* Structural lines: dividers, grid lines, section rules, table grids. Quiet
+     (~1.4:1 on white) but opaque, so it doesn't dissolve into a white field.
+     Lightness and chroma are tuned by eye; only the hue follows the palette. */
+  --color-border: oklch(from var(--color-charcoal-950) 0.9 0.004 h);
```
sarahgm: ⚠️ **`border` no longer guarantees its own opacity.**

In relative color syntax an omitted alpha defaults to **the origin color's alpha**, not to `1`. This resolves to alpha 1 today only because `charcoal-950` happens to be opaque.

The comment three lines above says *"`border` must STAY opaque"*. This change removes the structural guarantee in the same breath as documenting the invariant, and every sibling token states alpha explicitly (`/ 0.13`, `/ 0.26`, `/ 45%`).

```suggestion
  --color-border: oklch(from var(--color-charcoal-950) 0.9 0.004 h / 1);
```

Related, one line below the soft cap at :125: `--color-soft: oklch(0.99 0.002 54)` still hardcodes hue 54. It is the one remaining literal-hue token after this PR, so a rebrand now carries the divider and the shadow but leaves the soft cap behind. Same defect class the PR just fixed twice.

themes/theme-rui/src/styles.css
```diff
@@ -33,8 +33,11 @@
   }
 }
 
+/* The theme's own palette. Force-generated because this bundle is pre-compiled:
+   a consumer's markup is never scanned, so palette utilities they write would
+   otherwise not exist. Must track the palette name in tokens.css. */
 /* prettier-ignore */
-@source inline('{bg}-{stone}-{50,{100..900..100},950}');
+@source inline('{bg}-{charcoal}-{50,{100..900..100},950}');
```
sarahgm, line 47: 💡 **Three things about this line.**

**It was never a stale rename.** `git log -S stone -- themes/theme-rui/src/` shows the theme never declared a `stone` palette. This has been force-generating Tailwind's *default* stone since DST-815 (#4613), so it was wrong from the start rather than drifting out of sync.

**Only `bg-` is force-generated.** Post-fix a consumer writing `text-charcoal-700` or `border-charcoal-300` against the pre-compiled bundle still gets nothing, with no comment or test that would surface it. Either widen the group or say in the comment that only backgrounds are supported.

**A comment cannot hold the tracking requirement.** There is no test infrastructure under `themes/` at all, so *"Must track the palette name in tokens.css"* is the only guard against the exact failure this PR is fixing. Colocating the `@source inline` in `tokens.css` directly under the palette declarations removes the cross-file requirement entirely, since Tailwind collects `@source` from imported files.

Worth noting for the rebrand thesis: these compile to **inlined literals** (`background-color: oklch(0.62 0.007 54)`), because `styles.css` imports tokens with `theme(inline)`. The `bg-stone-*` they replace used `var(--color-stone-500)`. So a runtime `--color-charcoal-*` override does not retint them.

themes/theme-rui/src/tokens.css
```diff
@@ -168,95 +168,95 @@
...
 
-  /* Universal: no spacing (works for gap, padding, and inset recipes) */
+  /* Universal: no spacing (works for gap, padding, and inset recipes).
+   *
+   * The `-x` / `-y` variants are required, not redundant: `collapsed` is a
+   * member of `InsetSpacingTokens`, and `resolveInsetAxes` appends the axis
+   * suffix to every non-numeric token — so `<Panel p="collapsed">` resolves to
+   * `var(--spacing-collapsed-x)` and would otherwise be undefined. */
   --spacing-collapsed: --spacing(0);
+  --spacing-collapsed-x: --spacing(0);
```
sarahgm, line 275: 💡 **Altitude: this fixes the symptom one layer below the defect.**

The bug is in `resolveInsetAxes` (`packages/system/src/utils/css-variables.utils.ts`), which appends an axis suffix to every non-numeric token including the one token where an axis split is meaningless.

Three reasons the theme-layer fix is one layer too low:

- **The burden moves to every future theme.** `collapsed` is documented as the universal no-spacing token and four components accept it (`Panel`, `Card`, `Page`, `SelectList`). Any theme implementing this API must now ship three tokens that all mean `0` to make one documented prop value work, and nothing tells it to. That is the "hardened around the current styling" failure this audit set out to find.
- **The root cause is duplicated.** `packages/components/src/SelectList/SelectList.tsx:254` hand-rolls the same `scale ? inset : \`${inset}-x\`` logic. Fixing at the theme masks both copies at once, which makes the resolver less likely to get fixed.
- **It is untestable where the defect lives.** `css-variables.utils.test.ts` has five `resolveInsetAxes` cases and no `collapsed` case. A CSS declaration cannot be asserted by that suite, so the defect returns silently the moment someone touches the resolver.

Suggested deeper fix: treat the no-spacing token as axis-less in the resolver plus SelectList's copy. One line, one regression test, zero theme obligation. Not via `isScale`, which `createWidthVar` / `createHeightVar` / `createSpacingVar` also use and where `collapsed` must stay non-numeric.

Fine to ship the token fix now and take this as a follow-up on DST-1672, but the follow-up is worth filing.

docs/content/foundations/token-overview/index.mdx
```diff
@@ -579,7 +579,7 @@ Every boundary token in one place. Marigold draws element edges with five tokens
     },
     {
       name: 'border',
-      value: 'oklch(0.9 0.004 54)',
+      value: 'oklch(0.9 0.004 54) — hue from palette',
```
sarahgm: 💡 **Prose in a value cell.**

`TokenTable` renders `value` inside a `<code>` element (`docs/ui/TokenTable.tsx:43`), so the note and its dash come out monospace. This is the only row in roughly 60 whose Value is not a pure value, and two rows above the same table writes palette-derived values as `charcoal-950 / 13%`.

It also restates the `54` literal this PR just removed from `tokens.css`, so it is a second source of truth for the hue.

```suggestion
      value: 'oklch(0.9 0.004 54)',
```

Hue provenance reads better in the token file, or in the `useFor` prose column if it belongs in the docs at all.

themes/theme-rui/src/ui.css
sebald: Confirmed and fixed in 20526129d — this was a real regression I introduced, and my comment claiming parity with the `@theme` tokens was simply false.

Reverted to declaring it inside `@theme static` rather than wrapping it in `@layer theme { :root, :host }`, so it gets the same placement as every other token with no special-casing.

I verified the shadow-DOM consequence rather than reasoning about it, because custom properties *do* inherit through the shadow boundary — so a mount where the outer document also loads the theme CSS would have masked it. In a fully encapsulated mount (theme CSS loaded **only** inside the shadow root, no outer stylesheet):

| stylesheet | `ui-panel-content` padding-left | `--ui-panel-px` |
| --- | --- | --- |
| before | `0px` | **unresolved** |
| after | `24px` | `calc(0.25rem * 6)` |

Compiled output now matches the base commit: selector `:root, :host`, layer `theme`.

docs/content/foundations/token-overview/index.mdx
sebald: Fixed in 20526129d. Moved the provenance into `useFor`, where prose belongs, and left `value` as the bare computed value.

Also applied the same treatment to the `soft` row, since this round derives that token’s hue too — otherwise the table would have had the same problem in a second place.

themes/theme-rui/src/styles.css
sebald: Split verdict — you are right on the history, and I have corrected the changeset. But the colocation suggestion is wrong, and I have the measurement.

**History: you are right, and my wording was the inaccurate one.** No file at `d35022b83^` declares `--color-stone-N`, and `d35022b83` is where `--color-charcoal-*` was *added* — so there was never a `stone` palette to rename. I have dropped "leftover from before the rename" from the changeset.

One refinement though: "wrong from the start" is a touch strong. When the line landed in #4613 the theme genuinely did use Tailwind’s default stone via utilities (`border-stone-800` survived until DST-878), so force-generating `bg-stone-*` matched reality *then*. It went stale at DST-878, when charcoal arrived and the usages migrated but this line did not. The changeset now says exactly that.

**Colocation: this would break `theme.css`.** `theme.css` imports `tokens.css` too, and Tailwind collects `@source` across the whole import graph — so moving the directive forces the palette utilities into the tree-shaking path, which is that entry point’s entire purpose. Measured on a probe build that imports `theme.css` with a single-class content file:

| `@source inline` location | forced `bg-charcoal-*` on the native path | size |
| --- | --- | --- |
| `styles.css` (today) | **0** | 176,671 B |
| `tokens.css` (colocated) | **11** | 177,483 B (+812) |

So it stays in `styles.css`. I have instead made the coupling explicit in both directions: a "renaming this palette? update `@source inline`" note at the palette declaration, and an expanded note here recording why it cannot move.

**Only `bg-`: real, and left deliberately.** Worth noting the design system itself writes *zero* charcoal utilities — every component goes through the semantic tokens, so all 11 exist purely as a consumer escape hatch. That makes a hatch covering background but not text or border genuinely arbitrary, and I would rather widen it than not. Widening to `{bg,text,border}` costs 1,861 B (+0.65%), which is cheap — but how wide the raw-palette escape hatch should be is a token-API decision rather than a defect fix, so I have recorded it with that number on DST-1674 instead of expanding scope here.

themes/theme-rui/src/tokens.css
sebald: Agreed on the diagnosis, and I have filed it — DST-1675, linked to DST-1672. Keeping the token fix in this PR so `<Panel p="collapsed">` stops depending on invalid-at-computed-value-time luck now, with the resolver fix tracked separately since it lands in `packages/system`, not the theme.

Your counts check out. Five `resolveInsetAxes` cases and zero for `collapsed` — I had said seven earlier, but two of those were the import and the `describe`. Four components accept it (`Page`, `Card`, `Panel`, `SelectList`), as you say.

One correction to the estimate, though: **"one line, one regression test" understates the SelectList side.** That copy is not accidental duplication — `SelectList.tsx:245-252` carries an explicit comment listing three deliberate divergences from `resolveInsetAxes` (`p` short-circuits instead of losing to explicit `px`/`py`; no theme-default fallback; partial-axis overrides valid), ending with "Don’t *consolidate* this with resolveInsetAxes without accounting for all three."

That does not weaken your point — the axis-suffix bug is present in both and both need the same guard. It just means the fix is two independent guarded changes plus tests on both paths, not a shared one-liner, and whoever picks it up should not start by merging the two functions. I have written that constraint into the ticket so it is not rediscovered the hard way.

sebald:
Thanks @sarahgm — this was a good review. One genuine regression of mine, and two places where my own verification record overclaimed. Pushed as 20526129d.

**Fixed**

| # | Finding | Outcome |
| --- | --- | --- |
| 🔴 | `--ui-panel-px` lost `:host` and `@layer theme` | Reverted to `@theme`. Verified `0px` → `24px` on `ui-panel-content` in a fully encapsulated shadow-DOM mount |
| 💡 | `--color-soft` still hardcoded hue `54` | Derived. It was the last literal-hue token; browser-verified identical |
| 💡 | Prose in the docs value cell | Moved to `useFor`, applied to the `soft` row too |
| 💡 | Stone history wording | Corrected in the changeset and PR body — you were right |
| 💡 | `theme(inline)` means palette utilities inline literals | Noted as an explicit limit on the rebrand claim. This was visible in my own diff and I missed it |
| 💡 | `border-border/60` `@supports` fallback isn't rebrand-safe | Recorded. I'd dismissed it too quickly |

**Not taken — with a measurement**

Colocating `@source inline` in `tokens.css` breaks `theme.css`. It imports `tokens.css` too, and Tailwind collects `@source` across the whole import graph, so the directive would force palette utilities into the tree-shaking path:

| `@source inline` location | forced `bg-charcoal-*` on the native path | size |
| --- | --- | --- |
| `styles.css` (today) | 0 | 176,671 B |
| `tokens.css` (colocated) | 11 | 177,483 B (+812) |

Your underlying point stands, so the coupling is now documented at both ends instead.

**Deferred**

- **DST-1675** — the `resolveInsetAxes` fix. Your diagnosis is right and your counts check out (five cases, zero for `collapsed`; I'd said seven, but two were the import and the `describe`). One correction: `SelectList.tsx:245-252` diverges *deliberately*, with three documented reasons and an explicit "don't consolidate" warning — so it's two guarded changes plus tests on both paths, not a shared one-liner. That constraint is in the ticket.
- **DST-1674** — the `bg-`-only escape hatch. Worth noting the design system writes *zero* charcoal utilities itself, so all 11 exist purely for consumers, which makes covering background but not text or border arbitrary. Widening costs +1,861 B (+0.65%). Recorded there since it's an API decision rather than a defect.

**Still open:** the `border` opacity question. You're right that the alpha matters, and it turns out that invariant is the thing to re-examine rather than harden — translucent borders blend across grounds, which is behaviour earlier work put there on purpose. I'm looking at which borders had alpha before this branch and will follow up separately rather than bolt `/ 1` onto it here.

Also: the stale locked worktree the old test-instructions note warned about has been deleted, so that note is gone from the description.

sarahgm (commented):
## Review follow-up — traced against `aa0f1a8ec`

Thanks for the round, and for the two measurements. I re-checked every finding against the branch head rather than the reply table, including the `/simplify` items that weren't inline. **9 addressed, 1 open by your own account, 7 untouched.**

### Addressed, verified at head

| Finding | Where it landed |
| --- | --- |
| 🔴 `--ui-panel-px` lost `:host` + `@layer theme` | Back in `@theme` at `tokens.css:284`. Your shadow-DOM verification is the right one, and it's a better record than my original claim |
| 💡 `--color-soft` hardcoded hue `54` | Derived at `tokens.css:135` |
| 💡 Prose in the docs value cell | Moved to `useFor`, and to the `soft` row too |
| 💡 Stone history | Corrected in changeset and PR body |
| 💡 Colocate `@source inline` | Rejected with the +812 B measurement. Coupling now noted at `styles.css:41-45` and `tokens.css:43` |
| 💡 Only `bg-` is force-generated | DST-1674, with the +1,861 B number |
| 💡 `resolveInsetAxes` altitude | DST-1675, including the SelectList "don't consolidate" constraint |
| 💡 Orphaned "Shared anatomy" comment | Resolved by the revert. `ui.css:323` intact, `:331` cross-references |
| 💡 `@supports` fallback, `theme(inline)` literals | Recorded in the PR body |

### Open

**`--color-border` alpha** (`tokens.css:124`). `aa0f1a8ec` removed the crossing-strokes justification and replaced it with a measured debunk, which is the right correction to make. It does not touch the point the thread was about: with an omitted alpha, relative color syntax inherits the origin's, so opacity is incidental rather than declared. Fine to carry that as a separate follow-up as you describe.

One thing that may change the shape of that follow-up. `--color-soft` at `:135` now has the same construction, so this round added a second token whose alpha is inherited rather than stated. Whatever the invariant turns out to be, it applies to both.

### Not addressed

These were all in the `/simplify` section rather than inline, which is probably why they were skipped. Every one is still true at head.

1. **Comment volume went from +6 to +18.** `tokens.css` 155 → 165, `ui.css` 250 → 258. `20526129d` added 10 comment lines and `aa0f1a8ec` added 6. The PR states a test for comments and it hasn't been re-run against the round.
2. **Back-references still present.** `tokens.css:90` "grouped with the boundaries below", `:182` "grouped with the boundaries above". Both still rot silently on a reorder.
3. **Boundaries preamble still duplicates the declaration comment, and grew.** `:94-105` and `:121-123` both state the opacity rule. "the state edges are solid rungs" is still plural with one state edge in the section.
4. **`--spacing-topbar` inventory is still stale.** `tokens.css:255-256` credits `AppShell`, which has no reference anywhere on this branch, and omits `Sidebar/SidebarModal.tsx:52`. Actual consumers: `SidebarRail.tsx:290` and `SidebarModal.tsx:52`. This is a new comment in this PR, so it shipped stale.
5. **`ui-contrast` / `ui-soft` asymmetry.** Documented at `ui.css:117` rather than resolved. The ask was to name both or inline both.
6. **Six shadow layers still repeat** `from var(--color-charcoal-950) 0.14 0.035 h` (`tokens.css:217-222`), with no note recording that the hoist was rejected and why.
7. **Hue `54` still hand-copied in docs.** `contrast-matrix.demo.tsx:57` and `:75` recompute the *published* contrast numbers from a literal `54`, and `index.mdx:83` states it in prose. This is the one worth doing before merge: it's the same defect class the PR just fixed in the theme, and a rebrand would leave the documented matrix wrong with nothing to catch it.

### Chromatic

Build 161 ran on `5f0b8603e`, two commits back, so it's stale relative to head. The `beta-release` baseline run that would have settled the Calendar attribution wasn't triggered — the last one there is 2026-07-16, before this branch's merge base. Not blocking, since VRT isn't enforced for PRs into `beta-release`.

---

sebald:
`/run-chromatic`

Triggering VRT for the `--color-border` change (0.14 alpha), per @sebald's go-ahead.

**Expected diffs** — every component that draws a divider, but only on tinted grounds:

- On `surface` (white): should be visually unchanged. 1.352:1 → 1.359:1, a delta of 0.007.
- On `muted` (charcoal-50): slightly crisper. 1.297 → 1.362.
- On `background` (charcoal-100): most visible change. 1.222 → 1.359.

So expect movement in Table (`muted` variant and grid rules), LegacyTable, Sidebar (charcoal-100 ground, `border-r`), Calendar's `sm:bg-muted sm:border-r`, TopNavigation, Tabs, Accordion, Panel and the `ui-panel-*` surfaces.

**One deliberate outlier:** `Styles.stories.tsx` uses `border-border/60` in two places. The modifier now multiplies the token's own alpha, so those two swatches go to 0.084 effective alpha and will read noticeably fainter. That's the only usage of an opacity modifier on a border token in the repo, and it's story-only — no shipped component does this.

**Expected NOT to move:** Drawer and SectionMessage. Both read the divider's channels rather than its appearance, so they now use the private `--ui-divider-solid` and render byte-identical — Drawer back to `oklch(0.8 0.004 54)`, SectionMessage's status edges back to opaque `rgb(126,164,241)`. If either shows a diff, that fix didn't take and I want to know.

themes/theme-rui/src/tokens.css
sebald: Resolved, though not the way this thread proposed — @sebald's call was that the alpha should be *declared*, not that the token should be opaque.

`--color-border` is now `oklch(from var(--color-charcoal-950) l c h / 0.14)` as of cbbb75f96. Alpha stated explicitly, so the ambiguity this thread was about is gone, and the token holds one contrast rather than one lightness.

The reason it changed rather than gaining `/ 1`: once the crossing-strokes rationale fell, nothing was left defending opacity, and measuring showed a fixed lightness was actively causing a problem. The same token measured **1.352:1** on `surface`, **1.297:1** on `muted`, **1.222:1** on `background` — a divider in the Sidebar reading 11% weaker than the same divider in a Table. At 0.14 the three land within 0.003 of each other and within 0.007 of the old value on `surface`, so white is unchanged to the eye and the tinted grounds tighten up. Alpha picked by sweeping 0.04–0.30 and sampling rendered pixels on all three grounds rather than computing them. Chromatic accepted.

**Your `--color-soft` point was the sharper one, and it caught a live bug two tokens over.** Following it up: `--color-soft` now states `/ 1` (it is a fill, opaque is intended). But checking who *else* derives from these tokens found two recipes that read `--color-border`'s **channels** rather than its rendered appearance, and both broke silently on the switch:

| consumer | intent | against the translucent token | measured |
| --- | --- | --- | --- |
| `Drawer.styles.ts:16` | `calc(l - 0.1)`, a step darker | operates on the rung's `0.15`, not the old `0.9` → near-black at 14% | 1.87:1 → **1.38:1** on white |
| `SectionMessage.styles.ts` ×4 | accent mixed 50% with the divider | `color-mix` propagates alpha | opaque → **0.57** |

Both now read a private `--ui-divider-solid` and render byte-identical to before — Drawer back to `oklch(0.8 0.004 54)`, the status edges back to opaque `rgb(126,164,241)`. Verified in-browser, not by inspection.

Generalising your observation: switching any token from a lightness to an alpha silently breaks anything doing `oklch(from … calc(l …))` or `color-mix()` on it. Nothing fails loudly. `grep` found both consumers, but that is the check to run, and it is worth remembering for the Tier 3 renames on DST-1674.

sebald:
## Round 2 — traced against `10e7546ce`

Thanks, this was the more useful of the two reviews. **Chromatic is accepted** — the run on the border change came back clean apart from the diffs I predicted, so that closes your stale-build point too.

Of the 7 `/simplify` items: **4 fixed, 1 filed, 1 declined, 1 incorrect.** Plus your open item, which turned out to be the most valuable thing in the review.

### Your open item found a live bug

`--color-border` is now `charcoal-950 / 14%` — alpha declared, so the ambiguity is gone. Details in the [thread](https://github.com/marigold-ui/marigold/pull/5693#discussion_r3690766844).

But your `--color-soft` observation was the sharper half, and chasing it turned up two recipes that read the divider's **channels** rather than its appearance. Both broke silently:

| consumer | against the translucent token | measured |
| --- | --- | --- |
| `Drawer` `calc(l - 0.1)` | operates on the rung's `0.15`, not the old `0.9` → near-black at 14% | 1.87:1 → **1.38:1** |
| `SectionMessage` ×4 `color-mix` | alpha propagates | opaque → **0.57** |

Both now read a private `--ui-divider-solid`, verified byte-identical to before. Nothing would have failed loudly. Worth carrying into DST-1674.

### Fixed

**5 — `ui-contrast` / `ui-soft` asymmetry.** You were right and the resolution was the cheaper direction: all seven `--contrast-*` vars had **exactly one reference each**, and nothing outside `ui.css` overrode them, so naming bought nothing. Inlined. That removes the block, its 5-line justification, and **14** emitted declarations — they were emitted twice, once for `ui-contrast` and once for `ui-contrast-destructive`, which I hadn't noticed. Now symmetric with `ui-soft`.

**3 — boundaries preamble.** Cut 13 lines → 3. Once `--color-border` went translucent the crossing-strokes debunk became vestigial: there's no opaque-border decision left to justify. Also fixed "the state edges" → `disabled-border` is the only one in the section, as you said.

**2 — back-references.** Both "grouped with the boundaries above/below" gone.

**1 — comment volume.** Fair hit, and the above is the response rather than an argument. By my count (`^\s*(/\*|\*|//)|\*/`) `tokens.css` was 105 → 128 before this pass and is 126 now, `ui.css` 248 → 251. Our counting methods differ — you have 155 → 165 — but the direction was the same and the trend was the wrong way. Most of the remaining delta is measurement records (the alpha rationale, the `--ui-divider-solid` warning), which I'd argue are the load-bearing kind, but I'm not going to claim the test was re-run when it plainly wasn't.

### Filed rather than fixed

**7 — hue `54` in docs.** Agreed this is the same defect class, and it's **worse than you described**: `contrast-matrix.demo.tsx:57` duplicates not just the hue but all 22 lightness/chroma values, and recomputes the *published* contrast numbers from that copy. Change a rung and the documented matrix silently lies.

Filed as **DST-1676** with both fix approaches, because the real fix is making the demo read computed values rather than swapping one literal for another — that means restructuring it from module scope to a post-mount read, which is a docs-tooling refactor rather than a token change. It's also pre-existing rather than introduced here. Ticket includes the `theme(inline)` gotcha: `bg-charcoal-*` compiles to inlined literals on the pre-compiled path, so reading the utility's computed `background-color` is the robust route.

### Declined

**6 — the six repeated shadow layers.** Not adding a note recording why the hoist was rejected. A comment whose only job is to explain why a refactor *wasn't* done is exactly what the test in item 1 should reject, and it would trade a real cost (a line that rots when the shadow changes) for no reader benefit. The six explicit layers are legible and greppable; if someone wants to hoist them later, the code is the record. Happy to be overruled, but not while also agreeing comment volume is a problem.

### Incorrect

**4 — `--spacing-topbar` inventory.** `AppShell` does exist — `packages/components/src/AppShell/AppShell.tsx`, exported from `index.ts:20`, and referenced at `SidebarRail.tsx:283`. The comment also doesn't credit a *component*; it says "the app shell's shared vertical datum" descriptively, then names three roles: the top bar's height (`TopNavigation.styles.ts:6`), the sidebar's brand row (`Sidebar.styles.ts:56`), and the rail's sticky offset (`SidebarRail.tsx:290`). All three are real.

You're right that `SidebarModal.tsx:52` is a fourth consumer, but I'm deliberately not adding it — turning that comment into an inventory of call sites is precisely the failure mode of item 2, which you and I just agreed to delete two of. It describes roles, not references.

---

Compiled output this round: **3,159 → 3,145** declarations, every delta accounted for. Typecheck clean.

sarahgm (approved):

PR #5698, opened 2026-08-04
chore(DST-1682): switch prerelease channel to rc, drop local scripts

sebald:
# Description

Switches the v18 prerelease channel from `beta` to `rc`, and retires the local release scripts now that CI owns the release.

**The channel switch is a one-line change.** `changeset publish` derives the npm dist-tag from `.changeset/pre.json`, so flipping `"tag": "beta"` → `"tag": "rc"` moves published packages onto the `rc` dist-tag. `release.yml` needs no change: it runs `changesets/action` on every push to `main`, which reads `pre.json` itself and — while pre mode is active — opens a version PR and publishes to the pre tag. The 160 already-consumed changeset IDs are kept, so nothing is double-bumped.

The prerelease counter continues rather than resetting, because changesets derives it from the current `package.json` version. The next release is therefore `18.0.0-rc.5`, not `rc.0` — still semver-greater than `18.0.0-beta.4`, so npm ordering is unaffected:

```
@marigold/components  18.0.0-beta.4 -> 18.0.0-rc.5
@marigold/system      18.0.0-beta.4 -> 18.0.0-rc.5
@marigold/theme-rui    6.0.0-beta.4 ->  6.0.0-rc.5
@marigold/icons        2.0.0-beta.4 ->  2.0.0-rc.5
@marigold/cli          0.5.0-beta.0 ->  0.5.0-rc.1
```

**The local scripts are removed.** `scripts/beta-{enter,exit,release}.mjs` and the `release:beta*` package.json entries are gone, with no `rc-*` replacement. They assumed a dedicated `beta-release` branch that no longer exists on the remote — pre mode has lived on `main` since b7122c0b7 — and CI has been doing the publishing anyway. Entering and exiting pre mode is now plain `pnpm changeset pre enter rc` / `pnpm changeset pre exit`.

One thing goes with `beta-release.mjs`: its notes on publishing locally with a granular npm token (bypass-2FA, plus `CI=true` to skip the `npm profile get` pre-flight that 403s on such tokens). CI authenticates via `setup-node`, so that only ever mattered for manual publishes, and the history has it if we need it back.

**Two review follow-ups are folded in** (2e00758f3): the dead `pnpm slack` script is dropped — it pointed at `scripts/slack-release-notification.mjs`, deleted in 99eb6fcae (#2641) and never re-added — and `README.md` gains a "Releasing" section, since the removed scripts were the only in-repo record of the prerelease flow.

## Rollout

⚠️ **The `beta` dist-tag needs a manual cleanup after the first rc publish.** It is not part of this diff — `changeset publish` only ever moves the tag it is publishing to, so `beta` stays pinned at `beta.4` (`@marigold/cli`: `beta.0`) forever. Anyone who installed with `@beta` keeps resolving a stale version with no error telling them the channel moved.

Once `18.0.0-rc.5` is out, per package either repoint so existing consumers keep moving:

```sh
npm dist-tag add @marigold/components@18.0.0-rc.5 beta
```

…or remove so `@beta` fails loudly instead of resolving stale:

```sh
npm dist-tag rm @marigold/components beta
```

Repointing is the gentler option. Affects `@marigold/{components,system,theme-rui,icons,cli}`. This belongs in the DST-1682 rollout communication.

N/A — no UI changes, so no stories, VRT, a11y review or changeset apply here (release tooling and root `package.json` scripts only).

Closes DST-1682

## Test Instructions

1. `head -3 .changeset/pre.json` — confirm `"tag": "rc"`.
2. `grep -rn "release:beta\|release:rc" package.json` — no hits; nothing else in the repo references the removed scripts.
3. Real verification is the next push to `main`: the changesets action should open a version PR bumping to `18.0.0-rc.5` and publish to the `rc` dist-tag rather than `beta`.

## Breaking Changes

No — for consumers; the published packages only change dist-tag and prerelease identifier.

For maintainers: `pnpm release:beta`, `release:beta:enter` and `release:beta:exit` are removed with no direct replacement. Use the changesets CLI (`pnpm changeset pre enter rc` / `pnpm changeset pre exit`) and let `release.yml` publish.

## Checklist


.changeset/pre.json
```diff
@@ -1,6 +1,6 @@
 {
   "mode": "pre",
-  "tag": "beta",
+  "tag": "rc",
```
aromko, line 3: ⚠️ The `beta` dist-tag is left stranded by this switch.

Current npm state:

```
@marigold/components  beta: 18.0.0-beta.4   latest: 17.9.1
@marigold/system      beta: 18.0.0-beta.4   latest: 17.9.1
@marigold/theme-rui   beta:  6.0.0-beta.4   latest:  5.5.1
@marigold/icons       beta:  2.0.0-beta.4   latest: 1.3.45
@marigold/cli         beta:  0.5.0-beta.0   latest:  0.4.0
```

Once publishing moves to `rc`, `beta` stays pinned at `beta.4` indefinitely. Anyone who installed with `@beta` keeps resolving a stale version and gets no error to tell them the channel moved.

Nothing to change in this diff, it is a one-off npm operation after the first rc publish. Two options:

- **Repoint** so existing consumers keep moving: `npm dist-tag add @marigold/components@18.0.0-rc.5 beta` (per package).
- **Remove** so `@beta` fails loudly instead of resolving stale: `npm dist-tag rm @marigold/components beta` (per package).

Repointing is the gentler one. Either way this is worth a line in the DST-1682 rollout communication.

package.json
```diff
@@ -83,9 +83,6 @@
     "typecheck:only": "tsc --noEmit --resolveJsonModule --project tsconfig.check.json",
     "typecheck:docs": "pnpm --filter @marigold/docs types:check",
     "changeset": "changeset",
-    "release:beta": "zx scripts/beta-release.mjs",
-    "release:beta:enter": "zx scripts/beta-enter.mjs",
-    "release:beta:exit": "zx scripts/beta-exit.mjs",
     "release:notification": "zx scripts/create-slack-notification.mjs",
     "slack": "zx scripts/slack-release-notification.mjs --quiet",
```
aromko: 💡 While retiring dead release scripts: `"slack"` points at `scripts/slack-release-notification.mjs`, which was deleted in 99eb6fcae ("chore: cleanup scripts", #2641) and never re-added, so `pnpm slack` has been broken ever since. `release:notification` → `create-slack-notification.mjs` is the live one and is fine.

```suggestion
    "release:notification": "zx scripts/create-slack-notification.mjs",
```

aromko (commented):
## Automated Code Review

**Recommendation: Approve.** No defects in the diff. Removing 205 lines of local tooling that duplicated what CI already does is the right trade.

I checked the release mechanics against the installed changesets source and against npm rather than taking the description on trust. Every claim in the PR body holds:

| Claim | Result |
|---|---|
| `changeset publish` derives the dist-tag from `pre.json` | Confirmed: `@changesets/cli` returns `preState.tag` and passes it as `--tag` |
| Next version is `18.0.0-rc.5`, not `rc.0` | Confirmed: `getPreVersion()` reads `prerelease[1]` from the current `package.json` (4), increments to 5, and `incrementVersion` appends `-${state.tag}.${preVersion}` |
| `@marigold/cli` lands on `0.5.0-rc.1` | Confirmed: same path from `beta.0` |
| `release.yml` needs no change | Confirmed: `changesets/action` with `publish: pnpm changeset publish` reads `pre.json` itself |
| No `rc`/`latest` mix-up | Confirmed: changesets diverts to `latest` only for packages with no regular release; all five have one (`@marigold/cli` latest is `0.4.0`) |
| Nothing else references the removed scripts | Confirmed: no hits outside the deleted files |
| 160 consumed changeset IDs kept | Confirmed: 160 IDs recorded, 182 changeset files present, so 22 feed the rc.5 changelog |

Visual regression tests are not needed: nothing under `packages/*/src`, `themes/`, or `.storybook/` is touched, and there are no story or style files in the diff. All 23 CI checks pass.

Two follow-ups, both left as inline comments and neither blocking:

1. The `beta` dist-tag stays stranded at `beta.4` after the switch, so `@beta` consumers go quiet without an error. One npm command per package once rc.5 is out, plus a note in the rollout communication.
2. `"slack"` in `package.json` points at a script deleted back in #2641. Cheap to drop in the same sweep.

One thing I would suggest beyond the diff: there is no `CONTRIBUTING.md` and no prose anywhere in the repo describing pre mode, dist-tags, or the prerelease flow. The removed scripts were the only in-repo record. Your description covers it thoroughly, but descriptions are not discoverable six months from now. A short "Releasing" section in `README.md` with `pnpm changeset pre enter rc` / `pnpm changeset pre exit` and a line on how `release.yml` picks it up would preserve it.

---

.changeset/pre.json
sebald: Good catch — agreed on repointing, and you're right that it can't live in this diff.

Captured it in two places so it doesn't get lost between merge and the first rc publish:

- A **Rollout** section in the PR description with the exact `npm dist-tag add` / `rm` commands and the affected packages, which is what feeds the DST-1682 rollout communication.
- A permanent line in the new README "Releasing" section, so the next channel switch doesn't rediscover this: *"Switching channel leaves the old dist-tag pinned at its last version … repoint or remove it per package after the first publish on the new channel."*

Going with repoint over remove: `@beta` installs in consumer projects keep moving instead of hard-failing, and `@rc` becomes the tag we actually advertise.

package.json
sebald: Confirmed and fixed in 2e00758f3 — `scripts/slack-release-notification.mjs` went away in 99eb6fcae (#2641) and was never re-added, so `pnpm slack` has been dead for ~3.5 years. Nothing referenced it (no hits in workflows, docs or scripts), so I dropped the entry rather than restoring the script.

Kept `release:notification` as-is — it's the live one, and `release.yml` calls it directly after a successful publish.

Small note on the suggestion block: applying it as written would have duplicated the `release:notification` key instead of removing `slack`, so I made the deletion by hand. Intent was clear though.

sebald:
@aromko thanks for the review — all three points are addressed, ready for another look. 2e00758f3 + a description update.

**1. Dead `pnpm slack` script** — dropped. Confirmed your archaeology: `scripts/slack-release-notification.mjs` was deleted in 99eb6fcae (#2641) and never re-added, so it's been broken for ~3.5 years. Nothing referenced it, so removing the entry beat restoring the script. `release:notification` untouched.

**2. Stranded `beta` dist-tag** — agreed, and agreed it can't live in this diff. Now recorded twice so it survives the gap between merge and the first rc publish: a **Rollout** section in the description with the exact commands and affected packages (this is what goes into the DST-1682 rollout communication), plus a permanent line in the README so the *next* channel switch doesn't rediscover it. Going with repoint over remove, for the reason you gave.

**3. No prose record of the prerelease flow** — this was the best catch, and you're right that a PR description isn't discoverable in six months. `README.md` now has a **Releasing** section covering: CI owns publishing (`release.yml` → `changesets/action` on push to `main`, version PR then publish), pre mode lives on `main` with no separate release branch, `pnpm changeset pre enter rc` / `pre exit`, and the two non-obvious bits — the prerelease counter comes from `package.json` rather than the tag (hence `beta.4` → `rc.5`, not `rc.0`), and the old-dist-tag cleanup.

I put it in the README rather than a new `CONTRIBUTING.md`: the README already owns "Local Development" and the command table, and one findable place beats two. Happy to split it out if you'd rather have a `CONTRIBUTING.md`.

Still no changeset and no VRT here — release tooling, root `package.json` and README only, nothing under `packages/*/src` or `themes/`.

aromko (approved):

PR #5699, opened 2026-08-04
chore(DST-1680): update react-aria to 1.20.0

sebald:
# Description

Updates React Aria to the 1.20.0 line (published 2026-07-31). No API or behaviour change on our side — the work here is verification.

| Package                 | From   | To         |
| ----------------------- | ------ | ---------- |
| `react-aria-components` | 1.19.0 | **1.20.0** |
| `react-aria`            | 3.50.0 | 3.51.0     |
| `react-stately`         | 3.48.0 | 3.49.0     |
| `@internationalized/date` | 3.12.2 | 3.12.3   |
| `@react-types/shared`   | 3.36.0 | 3.36.1     |

Closes DST-1680

### The two companion bumps are load-bearing, not cosmetic

RAC 1.20.0 declares `@internationalized/date: ^3.12.3` and `@react-types/shared: ^3.36.1`. pnpm will not move an in-range dependency unless the specifier changes, so leaving the old floors (`^3.12.2` / `^3.35.0`) keeps the lockfile at 3.12.2 / 3.36.0 while RAC pulls the newer ones — **two copies of `@internationalized/date`, which carries runtime state** (calendar objects), not just types. The exact pin in `docs/package.json` has to move in lockstep for the same reason.

### `pnpm dedupe` is required, and that's new

`pnpm install` alone leaves **two `react-aria` copies** (3.50.0 + 3.51.0): the frozen `@react-aria/*` shims float `react-aria: ^3.48.0`, so pnpm keeps the already-locked 3.50.0 alongside the exact 3.51.0 that RAC 1.20.0 pins. `scripts/check-react-aria-dedupe.mjs` fails; `pnpm dedupe` collapses it.

This is a **second, previously undocumented** mechanism that can duplicate the tree — the guard docs only described the `@react-types/*` inversion from #5485. Renovate automerges the `react-aria` group on minor/patch, but `react-aria-dedupe.yml` runs the guard on every PR to `main` and automerge waits for green checks, so a duplicated lockfile **cannot land silently — the PR just goes red and stalls.** Worth a follow-up to run `pnpm dedupe` in that flow so those PRs stop needing manual unblocking. (Corrected from an earlier claim here that it could land; thanks @aromko.)

### Remaining floors brought up to latest

`@react-aria/form` → `^3.2.1`, `@react-aria/live-announcer` → `^3.5.1`, `@react-stately/form` → `^3.3.1`. These lagged the versions already installed. Every other `@react-aria/*` (17) and `@react-stately/*` (8) entry was already at its latest published release, so the whole family is now current.

### `@react-types/{button,checkbox,grid,table}` deliberately stay held

All four latest minors pull `@react-spectrum/provider`, which drags in `@adobe/react-spectrum` and splits the i18n + overlay contexts. Three declare it as a direct dependency; **`grid` 3.4.0 declares it as a `peerDependency`**, so `npm view @react-types/grid dependencies` looks clean while pnpm still auto-installs the peer. Bumping `grid` also poisons the held `@react-types/table`, whose transitive `@react-types/grid: ^3.3.8` floats up to 3.4.0. I hit this by trying it — the dedupe guard caught it. Noted in the hold rule so the next person doesn't repeat it.

### `TableEditableCell.ssr.test.tsx`

RAC 1.20.0 re-creates the Table subtree during hydration instead of hydrating in place. The emitted markup is byte-identical (verified by comparing server and client structural signatures), but the DOM nodes are new, so the reference captured from the server markup goes stale.

Node identity is a RAC implementation detail. The DST-1507 behaviour this test guards — that the editor opens once the client takes over — still holds, so the test re-queries the trigger after hydration and keeps asserting that. Confirmed against a bare RAC table with no Marigold code: node replacement happens there too on 1.20.0 and not on 1.19.0, so it's upstream behaviour, not our editable-cell pattern.

### `vitest.config.shared.ts`

Excludes `**/.claude/**` from test collection. Agent worktrees under `.claude/worktrees/` are full checkouts of other branches on their own (often stale) dependency state; collecting them makes a local run report hundreds of failures that have nothing to do with the working tree. No effect in CI (fresh checkout), but it currently breaks local verification on `main`.

## Mobile tray a11y regression in `Autocomplete` **and** `ComboBox` — fixed

This started as a real, user-facing a11y regression, not a test artifact: the whole mobile tray subtree (dialog + input + listbox) picked up `aria-hidden="true"`, so **a screen reader could not reach the tray contents at all**. `getByRole` respects that, which is why 6 tests could not find `role="combobox"` or `role="dialog"` (1 unit, 5 story, plus `Coverage Report` downstream of the unit failure).

### Root cause

`useComboBox` keeps the combobox surface reachable with

```js
ariaHideOutside([inputRef.current, popoverRef.current].filter(el => el != null));
```

`ariaHideOutside` hides everything *outside* the elements it is handed, so an **empty list hides the whole document** — including the overlay itself. On the mobile path both refs are null:

- **`popoverRef`** — we render a `Tray`, not RAC's `Popover`, so nothing ever claims `PopoverContext`.
- **`inputRef`** — RAC's `Popover` mounts in the same commit that flips `isOpen`. The tray instead portals in a *later* commit, so the input inside it has not mounted yet when the effect runs.

The observable result is `aria-hidden="true"` on the portal root, which the tray then renders inside.

Two corrections to my earlier read of this, both from instrumenting the installed packages rather than reasoning about them:

- The effect body is **byte-identical** in `react-aria` 3.50.0 and 3.51.0, so nothing about the hide logic changed. RAC 1.20.0's new `FieldInputContext` is the only ComboBox delta, but it is not what empties the list — the commit-ordering above is, and it is why the pass fires with nothing to keep.
- Confirming this needed `rm -rf node_modules/.vite`: Vitest runs Vite's **pre-bundled** copy of `react-aria`, so patches to `node_modules` silently no-op and the call looks like it never happens.

### The fix

`popoverRef` now tracks whatever stands in for the combobox surface: the **dialog element** while the tray is open, and the **container the tray portals into** while it is closed — so the pass keeps the tray visible instead of hiding it.

Crucially this does **not** weaken background isolation, which never came from this pass: the tray builds on RAC's `Modal`, and `useModalOverlay` runs its own `ariaHideOutside(..., { shouldUseInert: true })`. That is why the ComboBox pass is redundant on this path rather than load-bearing. It also repairs `useComboBox`'s blur handling, which asks whether focus moved into `popoverRef.current` before closing and was always false with a null ref.

Pointing `popoverRef` at the tray is also what React Spectrum's mobile ComboBox did, so this restores the contract rather than working around it.

Three small changes: a `useComboBoxTrayRef` hook holding the reasoning, a `ref` passthrough on `Tray`, and one line in each of `MobileAutocomplete` / `MobileCombobox`. One scoped `eslint-disable` for `react-hooks/immutability`, which cannot model writing to a ref owned by RAC.

## Merges with `main`

Merged twice while this was in flight:

- **#5700 / #5701 / #5702** — `pnpm-lock.yaml` only.
- **#5703** (safe bumps for 42 packages + play-to-`.test()` story migration) — also conflicted in `docs/package.json` and `packages/components/package.json`. Took main's bumps throughout (`motion` 12.43.0, `@aws-sdk/client-bedrock-runtime` 3.1101.0, `@shikijs/transformers` ^4.4.1, `@tanstack/react-query` ^5.101.4, `@upstash/redis` 1.38.1) and kept this branch's react-aria floors, which main still has on the older line.

Both times the lockfile was regenerated from main's version with `pnpm install` + `pnpm dedupe` rather than hand-merged.

Note for anyone verifying locally: #5703 bumps Playwright to 1.62.1, so `pnpm exec playwright install firefox` is needed before the browser-mode tests will run.

## Test Instructions

Verified locally (Firefox, browser mode) on the merged branch:

1. `pnpm install && pnpm dedupe` — then `node scripts/check-react-aria-dedupe.mjs` passes (single `react-aria` 3.51.0 / `react-stately` 3.49.0, no `@react-spectrum`)
2. `pnpm typecheck:only` — 0 errors
3. `pnpm lint` — 0 errors (236 pre-existing warnings)
4. `pnpm test:unit` — **1512/1512 pass across 131 files** (was 1511/1512)
5. `pnpm test:sb` — **507/507 pass**, including all 5 previously-failing mobile story tests
6. `pnpm install --frozen-lockfile` — lockfile in sync

`main`'s CI is green on 1.19.0, so the failure was upgrade-caused rather than pre-existing.

## Breaking Changes

No.

## Checklist

- [x] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) 
- [x] Visual regression tests updated (for UI changes) 

## Not in scope

- Migrating the 66 `@react-aria/*` / `@react-stately/*` subpackage imports to `react-aria` / `react-aria/private/*` (separate ticket, ~2–3 days). Those subpackages are on a frozen line — Adobe did not republish them for 1.19.0 or 1.20.0 — so they will stop receiving fixes.
- Adopting new APIs: `MenuTrigger trigger="contextMenu"`, `keyboardNavigationBehavior="tab"` (may replace `TableEditableCell` internals), `PreviewTrigger`, `TokenField` (alpha, overlaps our `TagField`).
- An upstream issue for the commit-ordering assumption in `useComboBox` (it assumes the popover mounts in the same commit as `isOpen`). Worth filing so tray-style overlays are supported directly, but not needed for this PR.


sebald:
**VRT run:** [Chromatic build 170](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=170) — **73 visual changes** to review. ([workflow run](https://github.com/marigold-ui/marigold/actions/runs/30895862684))

The workflow shows as failed because the Chromatic action exits non-zero whenever changes are detected — it is not a crash. Slack was notified.

Treat the 73 with suspicion before reading it as 73 regressions. Two things inflate it:

1. This PR changes `package.json` and `pnpm-lock.yaml`, which are deliberately **not** in the workflow's `untraced` list so TurboSnap can detect runtime-dependency changes. A dependency bump therefore invalidates the trace and re-snapshots broadly rather than just the touched stories.
2. The baseline comes from `main`, which absorbed the v18 squash (#5304) minutes before this build, so some stories may be diffing against a baseline that itself just moved.

Still needs a human pass on Chromatic — a real rendering change from react-aria 1.20.0 would look identical to the above from here.

packages/components/src/utils/useComboBoxTrayRef.ts
```diff
@@ -0,0 +1,67 @@
...
+  // never measured against a stale element. It also has to run before
+  // `useComboBox`'s hide effect, which is a passive effect in an ancestor.
+  //
+  useLayoutEffect(() => {
+    if (popoverRef && !popoverRef.current) {
+      // `popoverRef` is owned by RAC, so `react-hooks/immutability` cannot tell
+      // that writing to it is the whole point here — this is the documented
+      // interop seam with RAC's ComboBox, not accidental shared mutation.
+      // eslint-disable-next-line react-hooks/immutability
+      popoverRef.current = portalContext?.getContainer?.() ?? document.body;
```
aromko, line 68: 💡 With the fallback in place `popoverRef.current` is never `null` again, which also feeds `useComboBox`'s blur path: `blurIntoPopover = nodeContains(popoverRef.current, e.relatedTarget)` (`useComboBox.mjs:206`). While the fallback holds `document.body`, that is true for *any* `relatedTarget` in the document, so `props.onBlur` and `setFocused(false)` get skipped.

Latent today, since the combobox `Input` only exists inside the dialog and therefore cannot blur while the fallback is armed. But the doc block above promises the opposite ("makes `useComboBox`'s blur handling work again"), so one sentence noting the fallback's over-broad `contains` would stop the next person from moving the input out of the tray subtree and losing blur silently.

Nit: line 55 is a stray empty `//`.

packages/components/src/Table/TableEditableCell.ssr.test.tsx
```diff
@@ -28,17 +28,26 @@ test('inline editing works after hydrating server-rendered markup', async () =>
   const container = document.createElement('div');
   container.innerHTML = renderToString(<EditableCell.Component />);
   document.body.appendChild(container);
-  const serverEditButton = within(container).getAllByLabelText('Edit')[0];
+  expect(within(container).getAllByLabelText('Edit').length).toBeGreaterThan(0);
```
aromko, line 34: 💡 This is a second, different assertion (the server markup emits the trigger) inside the arrange phase of a test named for post-hydration behaviour. Per the one-Arrange-Act-Assert rule, either drop it, the post-hydration query fails loudly anyway, or split the SSR-output check into its own `renderToString` test.

packages/components/src/ComboBox/MobileCombobox.tsx
```diff
@@ -69,13 +70,14 @@ const MobileComboBox = ({
   children,
 }: MobileComboBoxProps) => {
   const stringFormatter = useLocalizedStringFormatter(intlMessages);
+  const trayRef = useComboBoxTrayRef();
```
aromko, line 73: 💡 The regression is covered only implicitly: the mobile tests pass again because `findByRole` skips `aria-hidden` subtrees. That works, but a reintroduction shows up as "unable to find role combobox", which does not point at the accessibility tree.

One explicit assertion in the `mobile view` blocks of `ComboBox.test.tsx` / `Autocomplete.test.tsx` would name the invariant:

```ts
expect(screen.getByRole('dialog')).not.toHaveAttribute('aria-hidden');
```

vitest.config.shared.ts
```diff
@@ -67,6 +67,11 @@ const exclude = [
   '**/dist/**',
   '**/cypress/**',
   '**/.{idea,git,cache,output,temp}/**',
+  // Agent worktrees under .claude/worktrees/ are full checkouts of other
+  // branches, on their own (often stale) dependency state. Collecting them
+  // makes local runs report failures that have nothing to do with the working
+  // tree. Not an issue in CI (fresh checkout), but it breaks local verification.
+  '**/.claude/**',
```
aromko: 💡 The comment describes `.claude/worktrees/`, the glob excludes all of `.claude`. `**/.claude/worktrees/**` matches the stated cause and keeps any test that later lives under `.claude` collectable.

packages/components/src/Tray/Tray.tsx
```diff
@@ -99,6 +100,11 @@ export interface TrayProps extends Omit<
    * Children of the tray.
    */
   children?: ReactNode;
+
+  /**
+   * Ref to the tray's dialog element.
+   */
+  ref?: Ref<HTMLElement>;
```
aromko, line 107: ⚠️ This is a new prop on a publicly exported component, but both changesets are `patch`. Repo precedent for API additions is `minor` (e.g. `.changeset/card-padding-api-align-with-panel.md`). Suggest bumping `.changeset/combobox-tray-aria-hidden.md` to `minor`.

aromko (commented):
## Automated Code Review

Verified the fix mechanism against the installed `react-aria` 3.51.0 / RAC 1.20.0 sources rather than the description alone. Everything load-bearing checks out:

- RAC's ComboBox provides `PopoverContext` carrying the same `popoverRef` it hands `useComboBox` (`dist/private/ComboBox.mjs:117,126,248`), so the interop seam is real.
- `ariaHideOutside` is called only by `useComboBox`, `usePopover`, `useModalOverlay` and `DragManager`, which confirms scoping the fix to ComboBox/Autocomplete and leaving the Select/Menu/DatePicker trays untouched.
- The desktop path cannot be affected: `ComboBox.tsx` / `Autocomplete.tsx` render `MobileComboBox` only under `isSmallScreen`.
- The ref-ordering argument holds. `Dialog` is a descendant of `MobileComboBox`, so its ref attaches child-first before the hook's layout effect, and the `!popoverRef.current` guard cannot clobber a live dialog.
- The `getContainer()` branch is reachable, not speculative: `Provider/OverlayContainerProvider.tsx` wraps `UNSAFE_PortalProvider`, and RAC's `Modal` portals into the same container, so fallback and tray agree. A `null` container is caught by `??`.
- The hook stays internal (`index.ts` only re-exports `./utils/form.utils`), and `@react-aria/overlays` is already a declared dependency.

Dropping the fallback would re-introduce the bug on the open commit, since the tray portals later, so the hook is about as small as the constraint allows.

### Two things worth answering before merge

**1. What were the 5 accepted Chromatic baselines?**
The checklist leaves "Visual regression tests updated" unticked with "diffs need review", but the up-to-date run (`5e3e5f7`) reports *5 changes accepted as baselines*, and `visual-regression-tests.yml` sets no `autoAcceptChanges`, so they were accepted deliberately. This is the one part of the upgrade no automated check covers, and a reviewer cannot verify accepted baselines from the PR alone. A line naming the 5 stories would close it.

**2. The Renovate automerge risk is overstated.**
`.github/workflows/react-aria-dedupe.yml` runs `check-react-aria-dedupe.mjs` on every `pull_request` to `main`, and Renovate waits for green checks before automerging, so a duplicated lockfile cannot silently land, the PR just goes red and stalls. The follow-up is better framed as "run `pnpm dedupe` so react-aria Renovate PRs do not need manual unblocking" than "it can land a duplicated lockfile".

Remaining notes are inline: one latent `blurIntoPopover` widening, the SSR test's arrange-phase assertion, an explicit `aria-hidden` assertion for the mobile tests, the `.claude` glob breadth, and `patch` vs `minor` for the new public `Tray` `ref` prop.

No blockers. CI is green across all 25 checks and visual regression is up to date with the current head.

---

packages/components/src/utils/useComboBoxTrayRef.ts
sebald: Correct on both counts — fixed in 834083d.

`blurIntoPopover` is `nodeContains(popoverRef.current, e.relatedTarget)` (`useComboBox.mjs:206`), so while the closed-state fallback holds the portal container / `document.body` it is true for any `relatedTarget` in the document and `props.onBlur` + `setFocused(false)` are skipped. Latent as you say, because the combobox `Input` only exists inside the tray. The doc block now says so explicitly instead of only claiming the blur path is repaired, and the stray `//` is gone.

vitest.config.shared.ts
sebald: Agreed — narrowed to `**/.claude/worktrees/**` in 834083d. Nothing else under `.claude` was ever the problem, and `browserScanEntries` (`packages/*/src/**`) only ever matches inside a worktree checkout anyway.

packages/components/src/Tray/Tray.tsx
sebald: Fair — `.changeset/combobox-tray-aria-hidden.md` is `minor` as of 834083d. `Tray` gaining a public `ref` is an API addition, and the precedent you cite matches (`select-render-value`, `dst-1509-taggroup-collapse-at` are both `minor`; the `patch` ones are type widenings).

packages/components/src/ComboBox/MobileCombobox.tsx
sebald: Added in 834083d, with two adjustments after checking what actually fails.

`expect(screen.getByRole("dialog")).not.toHaveAttribute("aria-hidden")` cannot fail on the invariant: the default role query already skips `aria-hidden` subtrees, so a reintroduction throws "unable to find role dialog" — the same unhelpful message we have now. And the attribute does not land on the dialog; `ariaHideOutside` puts it on the tray's portal root, which the dialog renders inside.

So the assertion is:

```ts
const dialog = await screen.findByRole("dialog", { hidden: true });
expect(dialog.closest("[aria-hidden=\"true\"]")).toBeNull();
```

`ComboBox.test.tsx` had no `mobile view` block (only `Autocomplete.test.tsx` did), so it got one.

Verified it actually guards the regression. Note that reverting `ref={trayRef}` alone does **not** reproduce it — `useComboBoxTrayRef` writes the fallback to `popoverRef.current` from inside the hook, so the load-bearing half survives. With the hook stubbed to return `undefined`, the tray's dialog gains an `aria-hidden="true"` ancestor and both new tests fail (alongside the original `opens tray and shows empty state`); with the fix in place all 40 tests in the three touched files pass. That also nails down which half of the fix is doing the work.

packages/components/src/Table/TableEditableCell.ssr.test.tsx
sebald: Kept the assertion, with a comment naming why it is a precondition rather than a second act/assert pair — pushing back on this one.

`hydrateRoot` on markup that never contained the trigger client-renders one anyway, so the post-hydration query would still pass and the test would silently stop covering hydration at all. The line is what guarantees the server markup had the trigger, i.e. that hydration is really hydration. Splitting it into its own `renderToString` test would duplicate the render for the same guarantee.

Happy to split it if you would still rather have the strict one-AAA shape.

aromko (approved):

PR #5704, opened 2026-08-05
fix(DST-1686): make the control track fill ground-adaptive and enlarge the SegmentedControl indicator

sebald:
# Description

Three related changes to the "unselected control track" — the `Switch` groove, the `SegmentedControl` track and the `Slider` rail.

**1. The `Slider` rail was the wrong token, painted twice.** It used `bg-border`, the token for *structural lines* (dividers, grid lines, table rules), where the other two tracks use `bg-control`. On top of that, `Slider` applied its `track` style to two exactly-overlapping elements — the `SliderTrack` and an inner rail `div`, both `h-2` at the same position — so the translucent fill composited with itself and rendered at ~0.26 effective alpha instead of 0.14: `#c0bfbe` on white where the token specifies `#dddddc`. The two defects pushed in opposite directions, which is why the rail looked *darker* than `bg-control` while naming a token that would have made it *lighter*. The redundant element is gone; the `SliderTrack` is the rail. Geometry is unchanged, and `touch-none`, `select-none` and the disabled cursor stay on the interactive track element.

**2. `--color-control` is now translucent** — `charcoal-950 / 16%`, was the opaque `charcoal-300`. A track is not painted on one known background: it appears on a white Card, the gray page ground, a `muted` fill, and inside a hovered `Table` or `ListBox` row. A fixed palette step drifts across those — charcoal-300 measured 1.53:1 on white but only **1.21:1 inside a hovered row**, where the groove half-dissolved into the row it sits in. At 16% the four grounds land within 0.02 of each other (1.41–1.43:1), so a track weighs the same wherever it goes. Same rationale as `border` in DST-1672.

Before/after on a white surface: all three tracks go from `#d4d0ce` (and `#c0bfbe` for the Slider rail) to `#d8d8d7`.

**3. The `SegmentedControl` indicator is larger and its focus ring matches the rest of the system.** The frame of track around the selected thumb went 4px → 3px, so the thumb is 30px tall instead of 28px; the thumb's own 1px rim takes the innermost pixel, leaving 2px of bare track visible. Separately, the thumb's focus ring was a hand-rolled outline that reproduced only *half* of `ui-state-focus` — it drew the 3px `outline-ring/50` but never firmed `--ui-border-color` to the opaque ring colour, so a focused thumb read noticeably lighter than a focused `Input`. It now uses `ui-state-focus` itself, and resolves identically to an `Input`: 3px `outline-ring/50` at `outline-offset-0` plus a 1px opaque `oklch(0.52 0.008 54)` rim.

3px is the *minimum* frame that clears the ring: the list is a scroll container and clips at its padding box, so that padding is the only room an outset ring can grow into. Below 3px the first and last thumbs get a visibly shaved ring.

Closes DST-1686

## Screenshots / Preview

image

## Test Instructions

1. Run `pnpm sb` and open `Components/Slider` → **Basic**. The unfilled rail should be the same grey as the `Switch` track in `Components/Switch` → **Basic** and the `SegmentedControl` track in `Components/SegmentedControl` → **Basic**. All three should sample `#d8d8d7` on the white panel.
2. In `Components/SegmentedControl` → **Basic**, confirm the selected thumb is larger than before, with a thin frame of track around it.
3. Tab into the SegmentedControl. The focus ring should sit **outside** the thumb and look the same weight as the ring on a focused TextField (`Components/TextField` → **Basic**) — a 3px translucent ring plus a crisp dark 1px edge.
4. Arrow-key to the **first** and **last** segment and confirm the ring is not clipped at the track's edges.
5. Check a track on a non-white ground — e.g. a `Switch` inside a hovered `Table` row — and confirm the groove stays visible against the hover fill.
6. `Components/SegmentedControl` → **Overflow** / **OverflowSmallScreen**: confirm horizontal scrolling and the edge fade still behave, and that thumbs aren't clipped.

## Breaking Changes

No — no token was renamed or removed and no public API changed.

Worth flagging for consumers, though: `bg-control` is now **translucent**, so two stacked elements that both carry it composite into a darker track than the token specifies (that is exactly the `Slider` bug fixed here). The token docs gained a "Never paint it twice" callout. The changeset is currently a `patch`; a reviewer may reasonably argue for `minor`, since the rendered value of a public token changes.

## Checklist

### DIFF
--- a/.changeset/slider-track-control-surface.md
+++ b/.changeset/slider-track-control-surface.md
@@ -1,0 +1,18 @@
+---
+'@marigold/components': patch
+'@marigold/theme-rui': patch
+---
+
+fix: make `control` a ground-adaptive track fill, and give the `Slider` rail the same token as `Switch` and `SegmentedControl`
+
+**The Slider rail was the wrong token, painted twice.** It used `bg-border` — the token for structural lines (dividers, grid lines, table rules) — where the `Switch` groove and the `SegmentedControl` track both use `bg-control`. On top of that, `Slider` applied its `track` style to two exactly-overlapping elements (the `SliderTrack` and an inner rail `div`), so the translucent `bg-border` composited with itself and the rail rendered at ~0.26 effective alpha instead of 0.14 — measuring `#c0bfbe` on white where the token specifies `#dddddc`. The redundant inner element is gone; the `SliderTrack` itself is the rail. Geometry is unchanged (both were `h-2` at the same position) and `touch-none`, `select-none` and the disabled cursor stay on the interactive track element.
+
+**`control` is now translucent** (`charcoal-950 / 16%`, was the opaque `charcoal-300`). A track is not painted on one known background — it appears on a white Card, the gray page ground, a `muted` fill, and inside a hovered Table or ListBox row — and a fixed palette step drifts across those. charcoal-300 measured 1.53:1 on white but only 1.21:1 inside a hovered row, where the groove half-dissolved into the row it sits in. At 16% the four grounds land within 0.02 of each other (1.41–1.43:1), so a track weighs the same wherever it goes. Same rationale as `border` in DST-1672.
+
+The three tracks (`Switch`, `SegmentedControl`, `Slider`) read a touch lighter on white surfaces as a result: `#d8d8d7` rather than `#d4d0ce`. The trade-off is that the white Switch thumb and SegmentedControl indicator now vary against the track by ground (1.43:1 on white, 1.79:1 on a hovered row) where the opaque step held a flat 1.53:1 — track-vs-ground and thumb-vs-track cannot both be constant while the thumb is opaque, and ground legibility wins because it decides whether the control reads as a control at all.
+
+**The `SegmentedControl` indicator is larger.** The frame around the selected thumb went from 4px to 3px (`inset-y-[3px]` on the indicator, `p-[3px]` on the list), so the thumb is 30px tall instead of 28px. Since the thumb's own 1px rim occupies the innermost pixel of that frame, 2px of bare track is what you see.
+
+**The `SegmentedControl` thumb's focus ring is now the shared `ui-state-focus`.** It previously hand-rolled only the outline, so it missed the other half of that utility — firming `--ui-border-color` to the opaque ring colour — and read noticeably lighter than every other focused control. A focused thumb and a focused `Input` now resolve identically: 3px `outline-ring/50` at `outline-offset-0`, plus a 1px opaque `oklch(0.52 0.008 54)` rim. The ring stays outside the thumb, and 3px is the minimum frame that clears it: the list is a scroll container clipping at its padding box, so that padding is the only room an outset ring can grow into. Below 3px the first and last thumbs get a visibly shaved ring.
+
+If you use `bg-control` in your own code, note that it is now translucent: two stacked elements that both carry it composite into a darker track than the token specifies.
--- a/.changeset/slider-track-control-surface.md
+++ b/.changeset/slider-track-control-surface.md
@@ -3,7 +3,7 @@
 '@marigold/theme-rui': patch
 ---
 
-fix: make `control` a ground-adaptive track fill, and give the `Slider` rail the same token as `Switch` and `SegmentedControl`
+fix(DST-1686): make `control` a ground-adaptive track fill, and give the `Slider` rail the same token as `Switch` and `SegmentedControl`
 
 **The Slider rail was the wrong token, painted twice.** It used `bg-border` — the token for structural lines (dividers, grid lines, table rules) — where the `Switch` groove and the `SegmentedControl` track both use `bg-control`. On top of that, `Slider` applied its `track` style to two exactly-overlapping elements (the `SliderTrack` and an inner rail `div`), so the translucent `bg-border` composited with itself and the rail rendered at ~0.26 effective alpha instead of 0.14 — measuring `#c0bfbe` on white where the token specifies `#dddddc`. The redundant inner element is gone; the `SliderTrack` itself is the rail. Geometry is unchanged (both were `h-2` at the same position) and `touch-none`, `select-none` and the disabled cursor stay on the interactive track element.
 
@@ -15,4 +15,8 @@ The three tracks (`Switch`, `SegmentedControl`, `Slider`) read a touch lighter o
 
 **The `SegmentedControl` thumb's focus ring is now the shared `ui-state-focus`.** It previously hand-rolled only the outline, so it missed the other half of that utility — firming `--ui-border-color` to the opaque ring colour — and read noticeably lighter than every other focused control. A focused thumb and a focused `Input` now resolve identically: 3px `outline-ring/50` at `outline-offset-0`, plus a 1px opaque `oklch(0.52 0.008 54)` rim. The ring stays outside the thumb, and 3px is the minimum frame that clears it: the list is a scroll container clipping at its padding box, so that padding is the only room an outset ring can grow into. Below 3px the first and last thumbs get a visibly shaved ring.
 
-If you use `bg-control` in your own code, note that it is now translucent: two stacked elements that both carry it composite into a darker track than the token specifies.
+**The `SegmentedControl` track's corners are now concentric with the indicator's.** Both used the shared `rounded-surface`, but two rounded rectangles nested with a gap only look like parallel arcs when the outer radius is the inner radius plus that gap. With the thumb at 8px and 3px of frame around it, the track needs 11px; at 8px its corner read visibly tighter than the arc it frames. The track is now `calc(var(--radius-surface) + 3px)` — derived from the token, so it stays concentric if the radius is retuned. This is a deliberate exception to using `rounded-surface` everywhere.
+
+**The `SegmentedControl` indicator's resting rim is re-derived from the track.** `ui-frame` draws its rim as an *outset* ring, so the indicator's rim lands on the track and composites over `bg-control` — it reads denser than the same `control-border` token does on a field. The old compensation subtracted a flat 0.08, hand-tuned against the previous opaque `charcoal-300` track, which left the rim at an effective 0.31 against a field's 0.26. It is now derived from the track's own alpha (`--control-alpha`, newly exposed on the token), so a resting indicator rim and a resting `Input` edge render the same on every ground, and retuning the track cannot silently detune the rim.
+
+If you use `bg-control` in your own code, note that it is now translucent: two stacked elements that both carry it composite into a darker track than the token specifies, and anything that paints a translucent edge *over* a track has to account for the track underneath it.
--- a/packages/components/src/Slider/Slider.tsx
+++ b/packages/components/src/Slider/Slider.tsx
@@ -111,11 +111,7 @@ const _Slider = <T extends number | number[]>({
       >
         {({ state }) => (
           <>
-            {/* The `SliderTrack` above *is* the rail: it already carries
-                `classNames.track` at the same `h-2` geometry. Don't add a
-                second rail element here — it paints the track style twice, and
-                any translucency in the token then composites with itself into a
-                darker rail than the token specifies. */}
+            {/* No rail element here — the `SliderTrack` above is the rail. */}
             {/* fill */}
             <div
               className={cn(
--- a/packages/components/src/Slider/Slider.tsx
+++ b/packages/components/src/Slider/Slider.tsx
@@ -111,13 +111,11 @@ const _Slider = <T extends number | number[]>({
       >
         {({ state }) => (
           <>
-            {/* track */}
-            <div
-              className={cn(
-                'absolute top-[50%] h-2 w-full translate-y-[-50%]',
-                classNames.track
-              )}
-            />
+            {/* The `SliderTrack` above *is* the rail: it already carries
+                `classNames.track` at the same `h-2` geometry. Don't add a
+                second rail element here — it paints the track style twice, and
+                any translucency in the token then composites with itself into a
+                darker rail than the token specifies. */}
             {/* fill */}
             <div
               className={cn(
--- a/themes/theme-rui/src/components/SegmentedControl.styles.ts
+++ b/themes/theme-rui/src/components/SegmentedControl.styles.ts
@@ -3,16 +3,29 @@ import { ThemeComponent, cva } from '@marigold/system';
 export const SegmentedControl: ThemeComponent<'SegmentedControl'> = {
   // Outer track. Provides the surface (default variant). Doesn't clip: the inner
   // list is the scroll container, so option focus rings aren't cut at this edge.
+  // The radius is set per variant rather than once on `base`: twMerge doesn't
+  // recognise `rounded-surface` as a radius utility, so it won't drop it in favour
+  // of a variant's arbitrary `rounded-[…]` and emit order would decide the winner.
   group: cva({
-    base: 'group/segmented relative items-center rounded-surface',
+    base: 'group/segmented relative items-center',
     variants: {
       variant: {
         // Track matches the Switch groove and the Slider rail (`bg-control`, a
-        // translucent fill — see the token docs; never paint it twice). The 2px
-        // outer margin lives on the list (p-0.5), not here, so the list stays the
+        // translucent fill — see the token docs; never paint it twice). The 3px
+        // outer margin lives on the list (p-[3px]), not here, so the list stays the
         // one element that defines the frame around the thumbs.
-        default: 'bg-control',
-        ghost: '',
+        //   The radius is the one place this component departs from the shared
+        // `rounded-surface`, and it has to: two rounded rects nested with a gap are
+        // only concentric when the outer radius is the inner radius plus that gap
+        // (`R_outer = R_inner + d`). The thumb keeps `rounded-surface` (8px) and
+        // sits 3px in, so the track needs 11px. Sharing one radius makes the outer
+        // corner visibly tighter than the arc it frames. (The thumb's outset 1px
+        // rim doesn't change the sum: it widens the inner arc to 9px but closes the
+        // gap to 2px.) Derived from the token, so a retuned `--radius-surface`
+        // stays concentric.
+        default: 'bg-control rounded-[calc(var(--radius-surface)+3px)]',
+        // No track to frame, so no gap to compensate for.
+        ghost: 'rounded-surface',
       },
       size: {
         default: 'text-sm',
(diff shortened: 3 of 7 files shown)


themes/theme-rui/src/components/SegmentedControl.styles.ts
```diff
@@ -100,17 +108,24 @@ export const SegmentedControl: ThemeComponent<'SegmentedControl'> = {
...
+        // hand-rolled outline, so a focused thumb matches a focused Input exactly:
+        // that utility firms `--ui-border-color` to the opaque ring colour as well
+        // as drawing the 3px outline, and the outline alone (what this used to do)
+        // reads noticeably lighter than every other control without it.
+        //   At rest, control-border's ground-adaptive firming over-darkens the edge
+        // against the track, so the alpha is stepped down 0.08 (token-derived);
+        // ui-state-focus overrides that on focus.
         default:
-          'inset-y-[4px] left-0 w-full ui-control [--ui-border-color:oklch(from_var(--color-control-border)_l_c_h_/_calc(alpha_-_0.08))] group-has-[[data-focus-visible]]/segmented:outline-3 group-has-[[data-focus-visible]]/segmented:outline-solid group-has-[[data-focus-visible]]/segmented:outline-ring/50 group-has-[[data-focus-visible]]/segmented:outline-offset-1',
+          'inset-y-[3px] left-0 w-full ui-control [--ui-border-color:oklch(from_var(--color-control-border)_l_c_h_/_calc(alpha_-_0.08))] group-has-[[data-focus-visible]]/segmented:ui-state-focus',
```
sarahgm: 🔴 **Visual Regression Tests ran, but gave this component zero coverage**

Build 185 ran on the current head `57dfb48`, so it is not stale. The problem is what it snapshotted. TurboSnap traced the 4 traceable changed files to these 9 story files:

`DatePicker, DateRangePicker, Form, patterns/complex-form, Slider, Drawer, Card, ErrorState, Link`

`SegmentedControl.stories.tsx` and `Switch.stories.tsx` are not in that list, so both were among the 350 skipped tests and their snapshots were copied from baseline. The 4 changes in the build were Slider Basic, Slider Value Formatting, Slider Multiple Thumbs, and Complex Form: Employee Onboarding, all already accepted as baselines.

So none of this was ever snapshotted:

- the thumb going 28px → 30px (this line)
- the switch to `ui-state-focus` at `outline-offset-0`
- the first/last thumb ring clipping at the 3px scrollport padding, the tightest claim in the PR
- the Switch groove recolouring from `#d4d0ce` to `#d8d8d7`

Test instructions 2 through 6 all describe exactly this uncovered surface, and the green "UI Tests" check reads as coverage that does not exist.

Root cause: Storybook runs off `packages/components/src` and `packages/system/src` and consumes the theme as a built package, so `themes/theme-rui/src/**` is not in the Storybook module graph. TurboSnap traces it to nothing instead of falling back to a full run. Chromatic's `externals` input is the knob for this:

```yaml
externals: |
  themes/**/*.css
  themes/**/*.ts
```

For this PR: please get a full run before merge. A focused-thumb story would also help, since the ring is the thing that changed most and nothing pins it. The workflow fix is probably its own ticket, as this gap has been swallowing theme-only PRs for a while, DST-1672's `--color-border` change included.

themes/theme-rui/src/components/SegmentedControl.styles.ts
```diff
@@ -7,9 +7,10 @@ export const SegmentedControl: ThemeComponent<'SegmentedControl'> = {
     base: 'group/segmented relative items-center rounded-surface',
     variants: {
       variant: {
-        // Track matches the Switch's unselected track. The 4px outer margin lives
-        // on the list (p-1), not here, so the edge thumbs' focus ring stays inside
-        // the scroll container's clip.
+        // Track matches the Switch groove and the Slider rail (`bg-control`, a
+        // translucent fill — see the token docs; never paint it twice). The 2px
+        // outer margin lives on the list (p-0.5), not here, so the list stays the
```
sarahgm: ⚠️ This states the wrong value and the wrong class. The list is `p-[3px]`, so the margin is 3px, not 2px, and there is no `p-0.5` anywhere in the file. It also contradicts the `list` comment eight lines below, which correctly says 3px.

Looks like residue from the earlier 2px approach (the one DST-1686 still documents). In a file where the comments carry the whole design rationale, a wrong one costs more than none.

```suggestion
        // translucent fill — see the token docs; never paint it twice). The 3px
        // outer margin lives on the list (p-[3px]), not here, so the list stays the
```

themes/theme-rui/src/components/SegmentedControl.styles.ts
```diff
@@ -100,17 +108,24 @@ export const SegmentedControl: ThemeComponent<'SegmentedControl'> = {
...
+        // clearing the focus ring — see the note on `list` above.
+        //   The keyboard focus ring is drawn here, not on the cell, so it marks the
+        // thumb exactly. Selection follows focus, so the focused option is always
+        // the one under this thumb. It is the shared `ui-state-focus`, not a
+        // hand-rolled outline, so a focused thumb matches a focused Input exactly:
+        // that utility firms `--ui-border-color` to the opaque ring colour as well
+        // as drawing the 3px outline, and the outline alone (what this used to do)
+        // reads noticeably lighter than every other control without it.
+        //   At rest, control-border's ground-adaptive firming over-darkens the edge
+        // against the track, so the alpha is stepped down 0.08 (token-derived);
```
sarahgm: ⚠️ The step-down lost the reason it existed. The old comment justified `calc(alpha - 0.08)` with "On the dark charcoal-300 track, control-border's ground-adaptive firming over-darkens the edge". The track is no longer charcoal-300, it is `charcoal-950 / 16%` and composites to near-white on a Card. This rewrite keeps the value and restates the premise generically, but that premise was measured against the old darker ground.

The consequence works against the PR's own goal. On focus the thumb resolves identically to an Input, as claimed. At rest it does not: the thumb rim sits at alpha 0.18 while the Input's `control-border` is 0.26, so the resting thumb reads lighter than the field it is meant to mirror, on a ground that is now nearly white.

DST-1686's "Open" section flags this same item. Either re-derive the step, drop it and let `ui-control` stand, or say in the comment which ground the 0.08 was re-measured against.

.changeset/slider-track-control-surface.md
```diff
@@ -0,0 +1,18 @@
+---
+'@marigold/components': patch
+'@marigold/theme-rui': patch
+---
+
+fix: make `control` a ground-adaptive track fill, and give the `Slider` rail the same token as `Switch` and `SegmentedControl`
```
sarahgm: ⚠️ Missing the ticket scope. Every recent changeset carries it: `fix(DSTSUP-269): …`, `feat(DST-1381): …`, `style(DST-1586): …`.

```suggestion
fix(DST-1686): make `control` a ground-adaptive track fill, and give the `Slider` rail the same token as `Switch` and `SegmentedControl`
```

themes/theme-rui/src/components/SegmentedControl.styles.ts
```diff
@@ -23,18 +24,25 @@ export const SegmentedControl: ThemeComponent<'SegmentedControl'> = {
...
+  //   Don't drop below 3px without moving the ring inside the thumb: at 2px the
+  //   first and last thumbs get a visibly shaved ring (verified).
   //   - `motion-safe:scroll-smooth` makes the selection-reveal scroll animate for
   //     users who allow motion and jump instantly for those who don't; the
   //     component's `scrollTo` defers to it via `behavior: 'auto'` (matches Tabs).
   //   - `overscroll-x-contain` keeps horizontal overscroll from triggering the
   //     browser back/forward gesture at the track ends (matches Tabs).
   list: cva({
-    base: 'flex w-full items-center ui-scroll-mask-x p-1 -my-1 overscroll-x-contain motion-safe:scroll-smooth',
+    base: 'flex w-full items-center ui-scroll-mask-x p-[3px] -my-[3px] overscroll-x-contain motion-safe:scroll-smooth',
```
sarahgm, line 53: ⚠️ DST-1686 still documents the superseded implementation. Its Changes table says `inset-y-[4px]` → `inset-y-0.5`, `p-1 -my-1` → `p-0.5 -my-0.5`, `outline-offset-1` → `-outline-offset-3`, with the summary "the frame of track around it halved from 4px to 2px" and a trade-off headed "The `SegmentedControl` focus ring moved **inside** the thumb".

What shipped here is 3px plus the shared **outset** `ui-state-focus`, justified the opposite way (3px is the minimum that clears an outset ring). The code is the better call, but the ticket is now the wrong record of it, and it is where the stale `p-0.5` above came from. Worth updating the ticket.

packages/components/src/Slider/Slider.tsx
```diff
@@ -111,13 +111,11 @@ const _Slider = <T extends number | number[]>({
...
         {({ state }) => (
           <>
-            {/* track */}
-            <div
-              className={cn(
-                'absolute top-[50%] h-2 w-full translate-y-[-50%]',
-                classNames.track
-              )}
-            />
+            {/* The `SliderTrack` above *is* the rail: it already carries
```
sarahgm: 💡 Five lines describing markup that is not here. The same rule is now stated in `Slider.styles.ts` above the `track` cva and again in the token docs' "Never paint it twice" callout, so this is the third copy.

The styles file is the durable home, since that is where someone would reach to re-add a rail. One line here, or none.

sarahgm (changes requested):

themes/theme-rui/src/components/SegmentedControl.styles.ts
```diff
@@ -23,18 +24,25 @@ export const SegmentedControl: ThemeComponent<'SegmentedControl'> = {
     },
   }),
   // Inner scroll container: rows the options and scrolls them horizontally on
```
sarahgm, line 39: thats a very long comment - can you shorten the comment, so its more readable

themes/theme-rui/src/tokens.css
```diff
@@ -76,9 +76,37 @@
 
   /* --- Neutral Backgrounds & States --- */
   --color-muted: var(--color-charcoal-50);
-  --color-control: var(--color-charcoal-300);
   --color-hover: var(--color-charcoal-200);
 
+  /* Resting fill of an unselected control track, and only that: the Switch
```
sarahgm, line 81: this is also very long, can you shorten it?

themes/theme-rui/src/components/SegmentedControl.styles.ts
sebald: Fixed in 2e51d7ac1 — now reads "The 3px outer margin lives on the list (p-[3px])". You were right on both counts; it was residue from the 2px approach, and `p-0.5` appeared nowhere else in the file.

.changeset/slider-track-control-surface.md
sebald: Fixed in 2e51d7ac1 — `fix(DST-1686): …`.

packages/components/src/Slider/Slider.tsx
sebald: Cut to one line in 2e51d7ac1: `{/* No rail element here — the `SliderTrack` above is the rail. */}`. Agreed the styles file is the durable home — the full rule stays there and in the token docs.

themes/theme-rui/src/components/SegmentedControl.styles.ts
sebald: Trimmed in 2e51d7ac1, 18 lines → 13.

themes/theme-rui/src/tokens.css
sebald: Trimmed in 2e51d7ac1, 27 lines → 19.

themes/theme-rui/src/components/SegmentedControl.styles.ts
sebald: Re-derived in 2e51d7ac1. You were right that the step needed re-deriving, but the direction turned out to be the opposite of what we both read off the declared alphas — worth writing down because it changes the fix.

`ui-frame` draws its rim as an **outset** `ring`, so the rim never touches the thumb's own white fill: it lands on the track and composites over `bg-control`. Comparing `0.18` against a field's `0.26` misses the extra 16% of charcoal underneath. Sampled on rendered pixels in Storybook:

| | rendered rim |
| --- | --- |
| thumb, shipped `-0.08` | `#b4b3b2` (effective alpha **0.31**) |
| TextField, plain `ui-control` | `#c0bfbf` (0.26) |

So the resting thumb read *heavier* than the field it mirrors, not lighter.

Two layers of one colour stack to `control-alpha + (1 - control-alpha) * a`, so solving for `control-border`'s alpha:

```
a = (0.26 - 0.16) / (1 - 0.16) = 0.119
```

The ground cancels, so one value holds on `surface`, `background`, `muted` and a hovered row. Verified in an isolated harness across all four, then in the real story: the thumb rim is now `#c1c0bf` against a TextField's `#c0bfbf` — equal to within a rounding unit (the double composite costs 1/255).

Rather than hard-code `0.119` I exposed `--control-alpha` on the token and wrote the rim as a `calc()` over it, so retuning the track can't silently detune the rim again — which is exactly how the `-0.08` drifted. The comment now states the derivation instead of restating the premise.

Also, while in here: the track and the thumb both used `rounded-surface`, which can't be concentric across a gap — the outer radius has to be the inner plus the gap. Thumb 8px + 3px frame → 11px on the track. Set per variant, because twMerge doesn't treat `rounded-surface` as a radius utility and so won't drop it for a variant's arbitrary `rounded-[…]` (verified: it rendered 8px until I moved it).

themes/theme-rui/src/components/SegmentedControl.styles.ts
sebald: Ticket updated — DST-1686 now records what shipped rather than the superseded approach:

- Summary pt. 3: "halved from 4px to 2px" → "went from 4px to 3px, and its rim compensation was re-derived"
- Changes table: `inset-y-0.5` → `inset-y-[3px]`, `p-0.5 -my-0.5` → `p-[3px] -my-[3px]`, `-outline-offset-3` → the shared `ui-state-focus` (still **outset**), plus rows for `--control-alpha` and the rim
- Trade-offs: "The `SegmentedControl` focus ring moved **inside** the thumb" replaced with the outset / 3px-floor rationale — the opposite call, as you spotted; "the 2px band is 1px of visible ground" → "the 3px band is 2px of visible track"
- New section deriving the rim, with the measurements
- Open: the rim item is resolved and gone; replaced with the VRT-coverage gap from your other comment

themes/theme-rui/src/components/SegmentedControl.styles.ts
sebald: Confirmed from the build's own diagnostics artifact, and leaving this thread open since it isn't fixed yet.

Pulled `chromatic-diagnostics.json` from the run. Every claim holds: `git.commit` is `57dfb48` on this branch (not stale), `onlyStoryFiles` is exactly your nine files in that order, `actualCaptureCount: 48` / `inheritedCaptureCount: 350` / `changeCount: 4`, and neither `SegmentedControl.stories.tsx` nor `Switch.stories.tsx` is in the list. So the uncovered surface is exactly as you describe.

One correction to the root cause, because it changes which knob to reach for. Storybook does **not** consume the theme as a built package — `.storybook/decorators.tsx:8` imports `../themes/theme-rui/src/index.js`, and the trace log shows theme source is very much in the module graph:

```
— themes/theme-rui/src/components/Slider.styles.ts [changed]
  ∟ components/index.ts ∟ theme.ts ∟ index.ts
    ∟ packages/components/src/DatePicker/DatePicker.stories.tsx
```

It reached DatePicker/DateRangePicker only because those two stories import the theme source *directly* (`DatePicker.stories.tsx:12`). Every other story gets the theme through `decorators.tsx → preview.tsx`, and `decorators.tsx` is in our workflow's `untraced` list (`visual-regression-tests.yml:130`). The log spells it out:

```
We detected some untraced files, this may affect your traced changes as
    the untraced flag instructs TurboSnap to not trace dependencies for the files:
Files matching ./.storybook/decorators.tsx:
```

That's why `SegmentedControl.styles.ts` and `tokens.css` traced to nothing. So the lever is that `untraced` entry, not `externals` — `externals` would work, but by forcing full runs rather than by fixing the trace.

Not resolving this: the PR still needs a full run before merge, and that gap deserves its own ticket rather than being decided here.

sebald: Full run done — [build 189](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=189), green, **6 visual changes to review**.

Coverage this time, from the build's own diagnostics:

| | build 185 | build 189 |
| --- | --- | --- |
| `actualCaptureCount` | 48 | **398** |
| `inheritedCaptureCount` | 350 | **0** |
| `turboSnap` | traced to 9 story files | disabled |

So nothing inherited a baseline — every one of the 739 specs across 97 components was snapshotted for real, and the SegmentedControl/Switch surface your comment listed is finally covered.

Worth noting the blast radius came out small: 6 changed snapshots, consistent with `--color-control` having exactly three consumers (Switch groove, SegmentedControl track, Slider rail). The changed shots need your eyes in Chromatic — I can't read the spec list from CI, only the count.

Two notes on how the run happened:

- It went through `workflow_dispatch`, not `/run-chromatic`. For `issue_comment` events GitHub loads the workflow definition from the **default branch**, so a comment-triggered run would have used main's `onlyChanged: true` and reproduced this exact gap. `workflow_dispatch` resolves the workflow file from the dispatched ref.
- Disabling TurboSnap means also removing `untraced` and `traceChanged` — the CLI hard-errors on all three when `onlyChanged` is false (it gates exactly `--untraced`, `--externals` and `--trace-changed`). Cost me two failed runs before I read the CLI source.

Commits `605de4b43`, `ccbd3ea09` and `87d98f9be` are the temporary workflow change and **must be reverted before merge** — I'll do that once you've reviewed the build. Still leaving this thread open until then.

sarahgm: Good catch on `untraced`. That's the real cause, and a better fix than the `externals` I reached for, which would have hidden the broken trace behind forced full runs instead of repairing it.

Build 189 settles the coverage question for me: 398 actual captures, 0 inherited, so the SegmentedControl and Switch surface is genuinely snapshotted now.

Two things left before this can merge:

1. The three CI commits (`605de4b43`, `ccbd3ea09`, `87d98f9be`) are still the branch head, so `onlyChanged: false` with `untraced` and `traceChanged` commented out would land on main. Revert them and I'll approve.
2. The trace gap needs its own ticket. I searched DST and nothing covers it, closest is DST-1139 which is the original TurboSnap rollout. I'll write one up and link it here so it doesn't get lost with this PR.

On the changed shots: you accepted all 6 before I got to them. Fine by me for the 5 SegmentedControl ones, and the 6th is the ComboBox label-wrapping story, which is DSTSUP-269 drift from main rather than anything this PR did.

sarahgm (approved):

themes/theme-rui/src/components/SegmentedControl.styles.ts
sebald: Reverted in `627dbebc1` — one commit undoing all three (`605de4b43`, `ccbd3ea09`, `87d98f9be`), now the branch head.

The workflow file is byte-identical to its state at `2e51d7ac1`, i.e. before the temporary CI changes: `git diff 2e51d7ac1` on the tree is empty. So `onlyChanged: true`, the full `untraced` list and `traceChanged: expanded` are all back, and nothing but the theme/component changes lands on main.

Worth noting a push doesn't re-fire Chromatic — the workflow only triggers on merged PRs, `workflow_dispatch` and `/run-chromatic` — so build 189 stays the run of record for this branch.

Agreed on the ticket being yours to write up; link it here and I'll close this thread once it exists.

PR #5705, opened 2026-08-05
fix(DST-1647): honor RouterProvider useHref in sidebar links

sebald:
# Description

`SidebarLink` (single-column items) and `SidebarRail`'s `RailItemLink` rendered the raw `href` prop directly onto their anchor, which shadowed the value produced by `RouterProvider`'s optional `useHref`. Applications served from a prefix, such as a Next.js `basePath`, ended up with sidebar markup pointing at an unprefixed URL, so middle click and "copy link address" resolved to the wrong page.

`navigate` was never affected and always worked, since `handleLinkClick` reads the raw prop rather than anything `useLinkProps` produces. Both components now let `useLinkProps` supply the rendered href while still handing the unprefixed path to `navigate`, matching how react-aria's own `useLink` is built. Consumers that do not pass `useHref` see no change, because the default leaves the href untouched.

One wrinkle worth a reviewer's eye: `useLinkProps` applies `router.useHref(props?.href ?? '')` unconditionally, so an undefined href comes back as `href=""` rather than being omitted. The second commit therefore only handed the hook an href when one actually exists, which keeps no-href branch triggers and `onPress`-only rail tiles free of the attribute. CI caught that regression on the first commit. The third commit folds that guard into the hook call itself (`useLinkProps(href ? { href } : undefined)`) so the "no href, no attribute" contract lives at one site instead of being enforced by a spread guard in the JSX — `useLinkProps` returns `{}` for an undefined argument, and `router.useHref('')` is called either way, so the hook count and behaviour are unchanged.

These two are the only raw `<a>` elements in `packages/components/src`. Everything else renders through RAC `Link`, which already handled `useHref` correctly.

**This supersedes `507a47b8a`** (the DST-1609 review cleanup), which removed this same pattern from `RailItemLink` on the grounds that it was dead code. That premise was wrong: the spread was never dead, it was shadowed by the raw `href` rendered after it, so "immediately overridden by the raw href" described the bug rather than a reason to drop the spread. DST-1647 was originally filed on that same incorrect premise and has been rewritten to match the real conclusion. Nothing to correct in the changelog — `.changeset/dst-1609-two-level-sidebar.md` is still unreleased and never mentioned the cleanup, so the wrong claim only ever lived in a commit message.

**No visual changes.** Without a `useHref` provider the rendered markup is byte-identical, and only an `href` attribute value can differ. Chromatic ran green on this branch with no visual changes detected, so the Screenshots section is omitted deliberately.

Closes DST-1647

## Test Instructions

1. `pnpm test:unit --project=unit-tests packages/components/src/Sidebar/` — two new tests (one per component) assert both halves of the contract: the anchor renders `/base/...` while `navigate` receives the unprefixed path. Asserting only the DOM attribute would miss a regression that hands the transformed href to `navigate()`.
2. Wrap a sidebar in `<RouterProvider navigate={navigate} useHref={href => '/base' + href}>` and confirm rail tiles, panel items, and the mobile drawer rows all render prefixed hrefs while `navigate` still gets the app-level path.
3. Confirm the no-href cases stay non-links: a branch whose leaves have no `href`, and an `onPress`-only rail tile, should render `role="button"` and no `href` attribute.
4. Storybook → `Components/RouterProvider/WithBasePath` shows the prefixed href next to the path passed to `navigate`.

## Breaking Changes

No

## Checklist

### DIFF
--- a/packages/components/src/RouterProvider/RouterProvider.stories.tsx
+++ b/packages/components/src/RouterProvider/RouterProvider.stories.tsx
@@ -83,7 +83,12 @@ export const Basic = meta.story({
  * `useHref` prefixes the rendered href, while `navigate` still receives the
  * unprefixed path.
  */
+// No snapshot: the whole point of this story is an `href` attribute value, and
+// `/base/start` renders the same pixels as `/start`. The rendering itself is
+// two default `Link`s, already captured by `Link.Basic`. The behaviour is
+// covered by the unit tests in `Sidebar.test.tsx` and `SidebarRail.test.tsx`.
 export const WithBasePath = meta.story({
+  parameters: { chromatic: { disableSnapshot: true } },
   render: () => {
     const [navigated, setNavigated] = useState<string>('/start');
     return (
--- a/packages/components/src/RouterProvider/RouterProvider.stories.tsx
+++ b/packages/components/src/RouterProvider/RouterProvider.stories.tsx
@@ -1,6 +1,7 @@
 import { useState } from 'react';
 import { Tab, TabList, TabPanel, Tabs } from 'react-aria-components/Tabs';
 import preview from '.storybook/preview';
+import { Link } from '../Link/Link';
 import { RouterProvider } from './RouterProvider';
 
 const meta = preview.meta({
@@ -15,6 +16,15 @@ const meta = preview.meta({
       table: {
         type: { summary: 'string' },
         defaultValue: { summary: './start' },
+      },
+    },
+    useHref: {
+      control: false,
+      description:
+        'Rewrites the href that links render, for example to prepend a router base path. The unprefixed path is still what `navigate` receives.',
+      table: {
+        type: { summary: '(href: string) => string' },
+        defaultValue: { summary: 'href => href' },
       },
     },
   },
@@ -63,6 +73,32 @@ export const Basic = meta.story({
         <pre>
           <strong>URL:</strong>
           {url}
+        </pre>
+      </>
+    );
+  },
+});
+
+/**
+ * `useHref` prefixes the rendered href, while `navigate` still receives the
+ * unprefixed path.
+ */
+export const WithBasePath = meta.story({
+  render: () => {
+    const [navigated, setNavigated] = useState<string>('/start');
+    return (
+      <>
+        <RouterProvider
+          navigate={setNavigated}
+          useHref={href => `/base${href}`}
+        >
+          <div style={{ display: 'flex', gap: 8 }}>
+            <Link href="/start">Start</Link>
+            <Link href="/reports">Reports</Link>
+          </div>
+        </RouterProvider>
+        <pre>
+          <strong>Path passed to navigate</strong> {navigated}
         </pre>
       </>
     );
--- a/packages/components/src/Sidebar/Sidebar.test.tsx
+++ b/packages/components/src/Sidebar/Sidebar.test.tsx
@@ -136,6 +136,7 @@ test('branch items render as links with auto-derived href', () => {
 });
 
 test('router useHref rewrites the rendered href while navigate gets the raw path', async () => {
+  // Arrange
   const navigate = vi.fn();
   render(
     <RouterProvider navigate={navigate} useHref={href => `/base${href}`}>
@@ -152,10 +153,13 @@ test('router useHref rewrites the rendered href while navigate gets the raw path
   );
   const link = screen.getByRole('link', { name: 'Overview' });
 
+  // Assert (rendered href is prefixed)
   expect(link).toHaveAttribute('href', '/base/overview');
 
+  // Act
   await user.click(link);
 
+  // Assert (navigate still receives the unprefixed path)
   expect(navigate).toHaveBeenCalledWith('/overview', undefined);
 });
 
--- a/packages/components/src/Sidebar/Sidebar.test.tsx
+++ b/packages/components/src/Sidebar/Sidebar.test.tsx
@@ -135,6 +135,30 @@ test('branch items render as links with auto-derived href', () => {
   expect(managementLink).toHaveAttribute('href', '/users');
 });
 
+test('router useHref rewrites the rendered href while navigate gets the raw path', async () => {
+  const navigate = vi.fn();
+  render(
+    <RouterProvider navigate={navigate} useHref={href => `/base${href}`}>
+      <MarigoldProvider theme={theme}>
+        <Sidebar.Provider>
+          <Sidebar>
+            <Sidebar.Nav>
+              <Sidebar.Item href="/overview">Overview</Sidebar.Item>
+            </Sidebar.Nav>
+          </Sidebar>
+        </Sidebar.Provider>
+      </MarigoldProvider>
+    </RouterProvider>
+  );
+  const link = screen.getByRole('link', { name: 'Overview' });
+
+  expect(link).toHaveAttribute('href', '/base/overview');
+
+  await user.click(link);
+
+  expect(navigate).toHaveBeenCalledWith('/overview', undefined);
+});
+
 test('sub-panel opens when child is active', () => {
   render(<WithActiveBranch.Component />);
 
--- a/packages/components/src/Sidebar/SidebarLink.tsx
+++ b/packages/components/src/Sidebar/SidebarLink.tsx
@@ -32,14 +32,14 @@ export const SidebarLink = ({
   const ref = useObjectRef(forwardedRef);
   const router = useRouter();
   // Sole source of the rendered href, since a raw `href` would shadow the
-  // router's `useHref`. Only spread when there is one: without an href
-  // `useLinkProps` yields href="", which would make a branch trigger a link.
-  const routerLinkProps = useLinkProps({ href });
+  // router's `useHref`. Pass undefined when there is none, or `useLinkProps`
+  // yields href="", which would make a branch trigger a link.
+  const routerLinkProps = useLinkProps(href ? { href } : undefined);
   const { tabIndex, onFocus } = useRovingItem(dataKey!);
 
   return (
     <a
-      {...(href ? routerLinkProps : {})}
+      {...routerLinkProps}
       data-active={dataActive}
       data-key={dataKey}
       ref={ref}
--- a/packages/components/src/Sidebar/SidebarLink.tsx
+++ b/packages/components/src/Sidebar/SidebarLink.tsx
@@ -31,14 +31,15 @@ export const SidebarLink = ({
 }: SidebarLinkProps) => {
   const ref = useObjectRef(forwardedRef);
   const router = useRouter();
-  // Sole source of the rendered href. A raw `href` here would shadow the
-  // router's `useHref`, while `navigate` still wants the untransformed one.
+  // Sole source of the rendered href, since a raw `href` would shadow the
+  // router's `useHref`. Only spread when there is one: without an href
+  // `useLinkProps` yields href="", which would make a branch trigger a link.
   const routerLinkProps = useLinkProps({ href });
   const { tabIndex, onFocus } = useRovingItem(dataKey!);
 
   return (
     <a
-      {...routerLinkProps}
+      {...(href ? routerLinkProps : {})}
       data-active={dataActive}
       data-key={dataKey}
       ref={ref}
--- a/packages/components/src/Sidebar/SidebarLink.tsx
+++ b/packages/components/src/Sidebar/SidebarLink.tsx
@@ -31,6 +31,8 @@ export const SidebarLink = ({
 }: SidebarLinkProps) => {
   const ref = useObjectRef(forwardedRef);
   const router = useRouter();
+  // Sole source of the rendered href. A raw `href` here would shadow the
+  // router's `useHref`, while `navigate` still wants the untransformed one.
   const routerLinkProps = useLinkProps({ href });
   const { tabIndex, onFocus } = useRovingItem(dataKey!);
 
@@ -40,7 +42,6 @@ export const SidebarLink = ({
       data-active={dataActive}
       data-key={dataKey}
       ref={ref}
-      href={href}
       role={href ? undefined : 'button'}
       className={className}
       tabIndex={tabIndex}
--- a/packages/components/src/Sidebar/SidebarRail.tsx
+++ b/packages/components/src/Sidebar/SidebarRail.tsx
@@ -3,7 +3,7 @@ import type { ReactNode, Ref } from 'react';
 import { Focusable } from 'react-aria-components/Focusable';
 import { useLocalizedStringFormatter } from '@react-aria/i18n';
 import { isFocusVisible } from '@react-aria/interactions';
-import { handleLinkClick, useRouter } from '@react-aria/utils';
+import { handleLinkClick, useLinkProps, useRouter } from '@react-aria/utils';
 import { cn } from '@marigold/system';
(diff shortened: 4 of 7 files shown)


packages/components/src/Sidebar/Sidebar.test.tsx
```diff
@@ -135,6 +135,30 @@ test('branch items render as links with auto-derived href', () => {
   expect(managementLink).toHaveAttribute('href', '/users');
 });
 
+test('router useHref rewrites the rendered href while navigate gets the raw path', async () => {
```
sarahgm, line 138: can you write this test as aaa test? Or is this even a component test because of the interaction!

packages/components/src/Sidebar/SidebarRail.tsx
```diff
@@ -3,7 +3,7 @@ import type { ReactNode, Ref } from 'react';
 import { Focusable } from 'react-aria-components/Focusable';
 import { useLocalizedStringFormatter } from '@react-aria/i18n';
 import { isFocusVisible } from '@react-aria/interactions';
-import { handleLinkClick, useRouter } from '@react-aria/utils';
+import { handleLinkClick, useLinkProps, useRouter } from '@react-aria/utils';
```
sarahgm, line 6: 🔴 **DST-1647 now describes the opposite of what this PR does.**

The ticket asks to delete the `useLinkProps` spread from `SidebarLink` as dead code, on the grounds that DST-1609's review cleanup (`507a47b8a`) already removed the same pattern here. That premise was wrong. The spread was never dead, it was shadowed by the raw `href` rendered after it, so "immediately overridden by the raw href" described the bug rather than a reason to drop the spread.

The direction in this PR is right, and I verified it against react-aria 3.50.0: `useLink` renders `useLinkProps(props)` for the DOM href and passes the raw `props.href` to `handleLinkClick`, which is exactly what both components now do.

What is missing is the paper trail. Could you rewrite DST-1647's title, Problem, Expected Outcome and Scope to match the real conclusion, and add a line to the PR body noting that this supersedes `507a47b8a`? Otherwise anyone reading the ticket after merge concludes the PR did the reverse of what it says. Nothing to fix in the changelog: `.changeset/dst-1609-two-level-sidebar.md` is still unreleased and never mentioned the cleanup, so the wrong claim only lives in a commit message.

sarahgm (commented):
Reviewed the code against react-aria 3.50.0 and the surrounding Sidebar history. The fix itself is correct and well tested, both blockers below are process rather than code.

packages/components/src/RouterProvider/RouterProvider.stories.tsx
```diff
@@ -68,3 +78,29 @@ export const Basic = meta.story({
     );
   },
 });
+
+/**
+ * `useHref` prefixes the rendered href, while `navigate` still receives the
+ * unprefixed path.
+ */
+export const WithBasePath = meta.story({
```
sarahgm, line 90: 🔴 **Visual Regression Tests: stale**

The last successful `Visual-Regression-Tests` run was on commit `c778f97c`, but the branch has since moved to `0f994924`. Please re-run by commenting `/run-chromatic` on this PR.

Flagging it because it is the checklist rule, though the real risk here is nil. Commit 2 only removes an `href=""` attribute from elements that already carry `role="button"`, and no theme rule selects on `href` (`grep -rn href themes/ --include="*.ts"` is empty).

packages/components/src/Sidebar/SidebarLink.tsx
```diff
@@ -31,16 +31,18 @@ export const SidebarLink = ({
 }: SidebarLinkProps) => {
   const ref = useObjectRef(forwardedRef);
   const router = useRouter();
+  // Sole source of the rendered href, since a raw `href` would shadow the
+  // router's `useHref`. Only spread when there is one: without an href
+  // `useLinkProps` yields href="", which would make a branch trigger a link.
   const routerLinkProps = useLinkProps({ href });
```
sarahgm: 💡 Consider moving the guard into the hook call, so the "no href, no attribute" contract lives at one site instead of being enforced by a spread guard a few lines away in JSX:

```suggestion
  const routerLinkProps = useLinkProps(href ? { href } : undefined);
```

Line 42 then becomes a plain `{...routerLinkProps}` and the three-line comment can collapse to one. `useLinkProps` returns `{}` when `props` is undefined (it guards with `if (props)` before copying keys), so behavior is identical and `router.useHref('')` is called either way.

As written the invariant is easy to regress. Someone tidying the JSX sees `{...(href ? routerLinkProps : {})}` right next to `role={href ? undefined : 'button'}` and reads the ternary as redundant.

packages/components/src/Sidebar/SidebarRail.tsx
```diff
@@ -70,6 +70,10 @@ const RailItemLink = ({
   className,
 }: RailItemLinkProps) => {
   const router = useRouter();
+  // Sole source of the rendered href, since a raw `href` would shadow the
+  // router's `useHref`. Only spread when there is one: without an href
+  // `useLinkProps` yields href="", which would make a section tile a link.
+  const routerLinkProps = useLinkProps({ href: node.href });
```
sarahgm: 💡 Same suggestion as in `SidebarLink.tsx`:

```suggestion
  const routerLinkProps = useLinkProps(
    node.href ? { href: node.href } : undefined
  );
```

Line 88 then becomes a plain `{...routerLinkProps}`.

packages/components/src/RouterProvider/RouterProvider.stories.tsx
sarahgm, line 90: 💡 **Worth opting this story out of Chromatic.**

The "Disabling Chromatic snapshots" guidance in `docs/content/components/__internal__/component-guidelines.mdx` says to keep a snapshot only when the story is the only one capturing a particular visual. This one captures nothing new:

- Its entire observable effect is an `href` attribute value, and `href="/base/start"` renders the same pixels as `href="/start"`, so a visual diff can never see the thing the story exists to demonstrate.
- The rendering is two default `<Link>`s in a flex row. `Link.Basic` already snapshots the default and small Link, and `Link.Inline` covers the secondary variant. A `RouterProvider` ancestor changes no pixels.
- The behavior is covered by the two new unit tests in `Sidebar.test.tsx` and `SidebarRail.test.tsx`.

That matches the first "disable" bullet, a visual duplicate of a story that is still snapshotted. It is the same shape as `Sidebar.Controlled` (`Sidebar.stories.tsx:167`) and `SidebarRail.RailControlled` (`SidebarRail.stories.tsx:466`), which both opt out for exactly this reason: "exercised by the unit tests, so no snapshot or component test needed".

```suggestion
export const WithBasePath = meta.story({
  parameters: { chromatic: { disableSnapshot: true } },
```

A short comment above the export explaining why would match the precedent in the Sidebar stories.

sarahgm (commented):
One follow-up on the Chromatic coverage of the new story.

packages/components/src/Sidebar/SidebarLink.tsx
sebald: Good call, applied in e3ff41793.

I checked the `useLinkProps` body rather than taking the signature's word for it, and it behaves exactly as you describe:

```js
function useLinkProps(props) {
    let router = useRouter();
    const href = router.useHref(props?.href ?? '');   // unconditional
    let linkProps = {};
    if (props) {                                       // the guard
        for (let key of ['href', 'target', 'rel', 'download', 'ping', 'referrerPolicy'])
            if (key in props) linkProps[key] = key === 'href' ? href : props[key];
    }
    return linkProps;
}
```

Two things worth recording, since they are what make the swap safe:

- `router.useHref('')` runs on both paths, so the hook count is unchanged — no conditional-hook hazard.
- The old `{ href: undefined }` form failed *because* of `key in props`: the key is present with an undefined value, so it took the `href` branch and wrote back `router.useHref('')`, i.e. `href=""`. Passing `undefined` skips the loop entirely. That is the whole bug the second commit was working around, and it now cannot recur from the JSX side.

Your point about regression risk is the real win here. `{...(href ? routerLinkProps : {})}` sitting next to `role={href ? undefined : 'button'}` genuinely does read as a redundant ternary, and the failure mode is silent — a stray `href=""` turns a branch trigger into a link with no test-visible type error. One site is better.

Verified the invariant still holds: `an onPress-only rail item activates with Enter and Space` (which asserts `not.toHaveAttribute('href')`) passes, as do both `useHref` tests. Typecheck clean — which also confirms the optional `props?: LinkDOMProps` signature accepts this.

packages/components/src/Sidebar/SidebarRail.tsx
sebald: Applied in e3ff41793, same as the `SidebarLink` thread — `useLinkProps(node.href ? { href: node.href } : undefined)`, and line 88 is now a plain `{...routerLinkProps}`.

Worth noting this is the site where the invariant mattered most: a section tile (a `RailItem` wrapping a `Sidebar.Nav`) and an `onPress`-only tile both legitimately have no `href`, so there are two distinct no-href shapes flowing through here versus one in `SidebarLink`. Confirmed both still render `role="button"` with no `href` attribute.

packages/components/src/RouterProvider/RouterProvider.stories.tsx
sebald: Agreed, opted out in e3ff41793. Your reasoning holds on all three counts, and I verified the third: `grep -rn href themes/ --include="*.ts"` is empty, so nothing in the theme layer can turn an href value into pixels even indirectly.

Added the comment above the export as you suggested, matching the `Sidebar.Controlled` precedent:

```tsx
// No snapshot: the whole point of this story is an `href` attribute value, and
// `/base/start` renders the same pixels as `/start`. The rendering itself is
// two default `Link`s, already captured by `Link.Basic`. The behaviour is
// covered by the unit tests in `Sidebar.test.tsx` and `SidebarRail.test.tsx`.
```

The story stays in Storybook as the human-readable demo of the `useHref`/`navigate` split, which is what it is actually good for — it just no longer claims visual coverage it cannot provide.

packages/components/src/Sidebar/Sidebar.test.tsx
sebald: Both fair — done in e3ff41793, and on the second question I'd argue it should stay a unit test.

**AAA:** marked the phases in both new tests. The shape is assert-then-act-then-assert rather than a single Act block, because the test deliberately checks two halves of one contract:

```tsx
test('router useHref rewrites the rendered href while navigate gets the raw path', async () => {
  // Arrange
  const navigate = vi.fn();
  render(/* ...RouterProvider with useHref... */);
  const link = screen.getByRole('link', { name: 'Overview' });

  // Assert (rendered href is prefixed)
  expect(link).toHaveAttribute('href', '/base/overview');

  // Act
  await user.click(link);

  // Assert (navigate still receives the unprefixed path)
  expect(navigate).toHaveBeenCalledWith('/overview', undefined);
});
```

That matches the existing precedent in `Tabs.test.tsx:77` (`// Arrange` / `// Assert (initial state)` / `// Act` / `// Assert`), so I followed that labelling rather than inventing a variant. I kept it as one test on purpose: splitting it would let a regression that hands the *transformed* href to `navigate()` pass, since each half would still be green in isolation. The pairing is the assertion.

**Component test?** I don't think so, for two reasons:

1. This file is already full of interaction tests — `back button returns to root panel`, the Cmd+B shortcut tests, `an onPress-only rail item activates with Enter and Space`. And these unit tests already run in a real browser (Firefox via Playwright), which is why the file carries the `inert` panels remove elements from the a11y tree comment. So `await user.click` here isn't the thing that would make it a component test.
2. A story-based version would need a new Sidebar story wrapped in `RouterProvider` with a `useHref` — which would add a Chromatic snapshot whose entire observable effect is an `href` attribute value. That's exactly what your other comment argues against for `WithBasePath`, so it'd be pulling in opposite directions.

The assertion target is a DOM attribute plus a mock call, with no visual dimension at all, so the unit test is the cheaper and more honest home for it. Happy to move it if you still prefer a story.

packages/components/src/Sidebar/SidebarRail.tsx
sebald: You're right, and thanks for chasing the paper trail rather than just the diff — the ticket did describe the exact inverse. Both fixed.

**DST-1647 rewritten.** Title is now `🐛 Sidebar links ignore RouterProvider useHref` (was `🧹 Remove dead useLinkProps spread in SidebarLink`). Problem, Expected Outcome and Scope all rewritten to the real conclusion, including your framing explicitly: the spread was never dead, it was shadowed by the raw `href` rendered after it, so "immediately overridden by the raw href" described the bug rather than a reason to drop it. The Problem section now opens by saying the original premise was wrong, so the history is legible instead of quietly overwritten. Also recorded the no-href wrinkle in Scope, since that is the non-obvious part for anyone touching this later.

**PR body** now carries a `This supersedes 507a47b8a` paragraph with the same explanation.

I confirmed the two supporting facts before editing:

- `507a47b8a`'s message reads "Drop dead `useLinkProps` in `RailItemLink` — its spread was immediately overridden by the raw `href`", so that commit removed the only correct half of the behaviour on the dead-code premise. Your read is exact.
- `.changeset/dst-1609-two-level-sidebar.md` is unreleased and never mentioned the cleanup, so nothing to correct there and the wrong claim only ever lived in a commit message. Confirmed.

One thing worth naming, since it explains how this got past two reviews: `navigate` was never broken, only the rendered attribute was. Click-through testing passes completely while middle-click and "copy link address" are wrong — which is why "the spread contributes nothing" looked true from the outside. That's now in the ticket as the reason it stayed invisible.

packages/components/src/RouterProvider/RouterProvider.stories.tsx
sebald: Correct, and re-triggered: https://github.com/marigold-ui/marigold/actions/runs/31015730463 (`workflow_dispatch` on `fix/dst-1647-sidebar-router-usehref`).

Note the branch has moved again since your comment — it's now at e3ff41793, which carries the review follow-ups from your other three threads. So the run covers the current tip rather than 0f994924.

I agree the residual risk is nil, and independently confirmed your `grep`: `grep -rn href themes/ --include="*.ts"` returns nothing, so no theme rule can select on `href`. Commit 3 narrows the surface further — it only folds the no-href guard into the `useLinkProps` call (same rendered output, verified by the `onPress`-only rail-tile test still asserting `not.toHaveAttribute('href')`), adds comments, and sets `disableSnapshot` on `WithBasePath`.

One consequence worth flagging so the diff isn't a surprise: because I took your `disableSnapshot` suggestion on `WithBasePath`, this run will report that story as removed from the snapshot set rather than as a new baseline. That's intended, not a regression.

Happy to keep the checklist rule enforced as-written — cheap to re-run, and "the risk is nil" is exactly the kind of judgement that's better verified than asserted.

sebald: ✅ Green. Chromatic **build 193 passed** on `e3ff41793` (the current tip) — 740 stories across 97 components, 37 snapshots captured, no visual changes detected.

- Run: https://github.com/marigold-ui/marigold/actions/runs/31015730463
- Build: https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=193

So VRT is no longer stale, and the "No visual changes" claim in the PR body is now backed by a run against the tip rather than an ancestor. Resolving.

sarahgm (approved):

PR #5638, opened 2026-07-13
feat(DST-1489): add ListView component

OsamaAbdellateef:
# Description

Adds `<ListView>` and `<ListView.Item>`, a non-form interactive list built on React Aria's `GridList`/`GridListItem` for rows the user operates in place — toggling a `<Switch>`, dismissing with an `<IconButton>`, opening a per-row `<ActionMenu>`, or following a link — without leaving the page. Unlike `<SelectList>`/`<Select>`/`<ListBox>`, `<ListView>` has no selection, no hidden input, and never becomes a submitted form value.

This closes the gap between `<List>` (presentational), `<ListBox>` (selection only, forbids focusable controls inside a row), `<SelectList>`/`<Select>` (form fields), `<Table>` (tabular), and `<Menu>` (commands that close on activation). The motivating consumer is the Popover notifications panel; docs cover two more must-support scenarios: a resource list with a per-row action menu, and a settings list of toggles.

v1 is intentionally minimal: no selection, no bulk-action bar, no async loading, no drag-and-drop — tracked as follow-ups.

Closes DST-1489

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm sb` and open `Components/ListView` in Storybook — exercise each story.
2. Run `pnpm start` and open `/components/collection/listview` in the docs app — check the three demos (resource list with action menu, settings toggles, notifications panel).
3. Run `pnpm test:unit` to confirm `ListView.test.tsx` passes.

## Breaking Changes

No

## Checklist

- [ ] Storybook preview and Marigold docs preview are available
- [x] Stories added/updated (with `component-test` tag where applicable)
- [x] Unit tests added/updated
- [x] Component documentation added/updated (if it exists)
- [ ] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) (for new/changed interactive components)
- [ ] Visual regression tests updated (for UI changes)
- [x] Changeset added (`pnpm changeset`)

PR #5706, opened 2026-08-05
fix(DST-1679): stop SelectList advertising ignored props and double-warning on textValue

sebald:
# Description

`SelectList` publicly typed two props it discarded at runtime, and `SelectList.Option` warned about a missing `textValue` that RAC already warns about.

- **`layout` and `keyboardNavigationBehavior` left the public type.** `layout` was published in the docs props table as `"stack" | "grid"` defaulting to `'stack'`, while the component always rendered a grid — passing it type-checked and did nothing, because `layout="grid"` is applied *after* the consumer prop spread. `keyboardNavigationBehavior` had the same defect one step removed: RAC derives it as `'tab'` from the grid layout, so a consumer value never survived. Both now sit in `RemoveProps` alongside `selectionMode`, which is the correct pairing for a prop the component owns.

- **`layout="grid"` stays hardcoded.** The ticket suggested it was now redundant and could be dropped, mirroring `ListView`. It isn't — it's what gives the list arrow-key navigation on both axes, so a horizontal list moves with Left/Right and still moves with Up/Down after the container-query flip to a vertical stack. Under RAC's `"stack"` default a row instead captures Left/Right to walk its own focusable children (`useGridListItem` attaches `onKeyDownCapture` with `preventDefault`/`stopPropagation` when `keyboardNavigationBehavior === 'arrow'`), and horizontal navigation stops working. `ListView` could drop it because it omits `orientation` too and is vertical-only; `SelectList` exposes `orientation="horizontal"`, and the docs promise "left/right when horizontal". A new test guards this.

- **The local `textValue` warning is gone.** RAC's `GridList` already warns on the same condition and its warning can't be suppressed, so one authoring mistake produced two console lines. It was also a false positive in one case: RAC's own fallback chain accepts an `aria-label` in place of a `textValue`, so such a row is named correctly — but the local warning fired on it anyway while RAC stayed silent. The string-children fallback is unchanged.

No runtime behavior change, and no rendered output changes — nothing here needs visual regression coverage.

Closes DST-1679

## Test Instructions

1. `pnpm typecheck:only` — passes clean.
2. `pnpm vitest run --config vite.config.ts --project=unit-tests packages/components/src/SelectList/SelectList.test.tsx` — all 47 pass, including three new tests: horizontal ArrowRight/ArrowLeft row navigation, warn-exactly-once, and aria-label-does-not-warn.
   - Heads up: if you have a git worktree under `.claude/worktrees/`, vitest's project glob sweeps it in and you'll see unrelated failures from it. Those are pre-existing in that worktree, not from this branch.
3. To confirm `layout="grid"` is load-bearing: change it to `"stack"` in `SelectList.tsx` and re-run — the new horizontal navigation test fails (ArrowRight stops moving between rows).
4. To confirm the type narrowing bites: add `<SelectList layout="grid" />` anywhere and typecheck — `error TS2353: 'layout' does not exist in type 'SelectListProps'`.

## Breaking Changes

No runtime break — both removed props were already discarded at runtime, so no consumer's behavior changes.

Type-level note for reviewers: code currently passing `layout` or `keyboardNavigationBehavior` to `SelectList` will now fail to typecheck. Shipping as `patch` since nothing that worked stops working, and the ticket frames removing them now as the cheap moment — before someone passes them in good faith and debugs a silent no-op. Happy to bump if you'd rather signal it louder.

## Checklist

- [ ] Stories added/updated (with `component-test` tag where applicable) — N/A, no stories changed
- [ ] Component documentation added/updated (if it exists) — N/A, no documented prop changed
- [x] Visual regression tests updated (for UI changes) — N/A, no visual change


packages/components/src/SelectList/SelectList.tsx
```diff
@@ -37,6 +44,8 @@ type RemoveProps =
   | 'renderEmptyState'
   | 'selectionMode'
   | 'selectionBehavior'
+  | 'layout'
```
sarahgm, line 40: 🔴 **Visual Regression Tests: not run**

This PR changes files under `packages/components/src/**` and the `Visual-Regression-Tests` workflow has no runs for this branch.

To be fair to the change: I traced the full diff and there is no render-path change. The only runtime edit is deleting a dev-only `console.warn` in `SelectListOption.tsx`. Everything here is a type-union entry plus comments, so a Chromatic run should be a clean no-op.

Either trigger it as cheap insurance by commenting `/run-chromatic`, or waive it explicitly with a note on the PR. Worth closing out rather than leaving open.

.changeset/selectlist-layout-and-textvalue-parity.md
```diff
@@ -0,0 +1,20 @@
+---
+'@marigold/components': patch
+---
+
+fix(DST-1679): stop `SelectList` from advertising two props it silently ignores, and stop `SelectList.Option` from double-warning about a missing `textValue`.
+
+**What changed:**
+
+- `layout` and `keyboardNavigationBehavior` are no longer part of `SelectListProps`. Both were publicly typed but discarded at runtime: the list hardcodes `layout="grid"` after spreading consumer props, and RAC derives `keyboardNavigationBehavior: 'tab'` from that grid layout. They now sit in `RemoveProps` alongside `selectionMode`, which is the correct pairing for a prop the component owns.
```
sarahgm, line 9: ⚠️ **This removal belongs in `MIGRATION-v18.md` §A4.6.**

That section (`MIGRATION-v18.md:585`) is titled "Dead and mis-named props removed" and opens with "A cleanup of props that silently did nothing." That is precisely this change, and the section already tracks pure type-level removals in the same shape:

- **`TextField` `min` / `max`** removed (never forwarded to the input)
- **`FileTrigger` / `FileField` `acceptedFileType`** removed (filtering silently never applied)
- **`Label`** no longer exposes `style`

Since the props table in `docs/public/components/form/selectlist.md` is gitignored and regenerated from the type, it self-corrects and no manual docs edit is needed there. But that also makes the migration guide the only place a consumer will ever read that these two props are gone. Suggested bullet:

```markdown
- **`SelectList` `layout` / `keyboardNavigationBehavior`** removed. The list is
  always a `layout="grid"` GridList and RAC derives `keyboardNavigationBehavior:
  'tab'` from that, so neither value ever reached the DOM. Drop them; behavior is
  unchanged.
```

This is the one item I would hold the PR for. The checklist marks documentation N/A, which is right for the props table but misses this file.

packages/components/src/SelectList/SelectList.test.tsx
```diff
@@ -329,6 +355,53 @@ describe('SelectList', () => {
...
+
+      // RAC's GridList already warns about a missing `textValue`, so
+      // SelectList.Option deliberately adds none of its own — one authoring
+      // mistake must not produce two console lines.
+      expect(warnSpy).toHaveBeenCalledTimes(1);
+      expect(warnSpy).toHaveBeenCalledWith(
+        expect.stringContaining('<GridListItem>')
+      );
+
+      warnSpy.mockRestore();
```
sarahgm, line 381: 💡 `mockRestore()` here is not failure-safe. There is no `restoreMocks` or `clearMocks` in `vite.config.ts` or `vitest.config.shared.js`, so if either assertion above throws, `console.warn` stays stubbed for every later test in this file. That can mask a genuine warning and turn one failure into a confusing cascade.

Hoisting the restore into the enclosing `describe` covers both new tests and lets you drop the two manual calls:

```ts
afterEach(() => {
  vi.restoreAllMocks();
});
```

Flagging as a suggestion rather than a fix request, since this matches the pattern in the sibling `ListView` tests in #5638. If it changes, it is probably worth changing in both.

packages/components/src/SelectList/SelectList.test.tsx
```diff
@@ -278,6 +283,27 @@ describe('SelectList', () => {
...
+    // Right only move between rows because of it — with RAC's `"stack"` default
+    // the row captures those keys to walk its own focusable children instead,
+    // and a horizontal list stops being navigable.
+    test('moves focus between rows on ArrowRight/ArrowLeft when horizontal', async () => {
+      render(<Horizontal.Component />);
+      const standardRow = screen.getByRole('row', { name: /standard/i });
+      const expressRow = screen.getByRole('row', { name: /express/i });
+
+      await user.tab();
+      await user.keyboard('{ArrowRight}');
```
sarahgm, line 300: 💡 Nice test, and I confirmed it genuinely guards the hardcoded `layout="grid"`: `useGridListItem.mjs:152-170` shows that under `keyboardNavigationBehavior: 'arrow'` (what `layout="stack"` yields) ArrowRight *always* calls `preventDefault()` and `stopPropagation()`, refocusing the row itself when the row has no focusable child. The `Horizontal` rows hold only `TextValue` and `Description`, so focus would stay on `standardRow` and this test fails. The guard bites.

One robustness nit: the assertion chain assumes `tab()` lands on `standardRow`, which holds only because the story sets `defaultSelectedKeys={['standard']}`. Pinning that makes the precondition explicit, so a story-level change fails loudly instead of confusingly.

```suggestion
      await user.tab();
      expect(document.activeElement).toBe(standardRow);

      await user.keyboard('{ArrowRight}');
```

packages/components/src/SelectList/SelectList.tsx
```diff
@@ -29,6 +29,13 @@ import { SelectListOption } from './SelectListOption';
 
 export type SelectionMode = 'single' | 'multiple';
 
+// Props the component owns. Every entry here is either replaced by a Marigold
```
sarahgm: 💡 The "why grid is load-bearing" rationale now lives in four places: this block, the comment at the `layout="grid"` render site (`:359-365`), the new test's comment, and the changeset. Given the project's preference for terse one-line comments, this one could defer to the render site rather than restating the derivation:

```ts
// Props the component owns: each is either replaced by a Marigold prop or
// hardcoded on the `GridList` below. See the `layout="grid"` render site for
// why `layout` and `keyboardNavigationBehavior` cannot be public.
```

The comment at the render site is the one that earns its length, so I would keep that one as-is.

sarahgm (commented):
Automated review via Claude Code. Five inline comments below: one critical (visual regression), one warning (migration guide), three suggestions.

Two notes that are not findings:

- **On the `patch` vs `minor` question in the description:** `patch` is correct. `main` is in changesets pre-mode (`tag: rc`, currently `18.0.0-beta.4`), so this lands inside the unreleased v18 major. No bump needed.
- I checked all four react-aria claims in the description against the installed source rather than taking them on trust, and they all hold: `GridList.mjs:124` (`layout === 'grid' ? 'tab' : keyboardNavigationBehavior`), `GridList.mjs:325` (RAC's own `textValue` warning), `Document.mjs:218` (the `textValue || string children || aria-label` chain, which confirms the `aria-label` false positive), and `useGridListItem.mjs:250` plus `152-170` (why `layout="grid"` is load-bearing). The divergence from `ListView` is also justified: #5638 puts both `orientation` and `layout` in its `RemovedProps`, making it vertical-only, while `SelectList` keeps `orientation` public.

.changeset/selectlist-layout-and-textvalue-parity.md
sebald: Agreed, and it's a stronger case than the section shape alone. In v17 `SelectList`'s `RemoveProps` held only `style | className | onSelectionChange | dragAndDropHooks`, so both props were genuinely public API in the last stable release — a v17 consumer passing either one *will* fail to typecheck on v18. Added your bullet verbatim, with `RAC` spelled out as `react-aria` to match the surrounding prose (2b0c259).

packages/components/src/SelectList/SelectList.test.tsx
sebald: Good catch on the implicit precondition — pinned it. One correction to the snippet though: applied verbatim it puts `document.activeElement` *above* the `/* eslint-disable testing-library/no-node-access */` line, and that rule is active for `**/*.test.tsx` via `flat/dom` (`config/eslint/index.js:37-44`). I checked — it's exactly one error:

```
296:23  error  Avoid direct Node access. Prefer using the methods from Testing Library  testing-library/no-node-access
```

So the disable moved up instead:

```ts
      /* eslint-disable testing-library/no-node-access */
      await user.tab();
      // Tab lands here only because the story sets `defaultSelectedKeys`.
      expect(document.activeElement).toBe(standardRow);

      await user.keyboard('{ArrowRight}');
```

Still 47/47 green, which also confirms the precondition holds today (2b0c259).

packages/components/src/SelectList/SelectList.tsx
sebald: Fair — took your version. The render-site comment stays as-is and is now the single place the derivation lives; this block just points at it. I did a pass over the rest of the comments this PR adds on the same rule (keep only what the code doesn't already say), which also dropped the comment in the warns-exactly-once test — it restated its own assertions — and tightened the `SelectListOption.tsx` one (2b0c259).

packages/components/src/SelectList/SelectList.test.tsx
sebald: The premise checks out — there's no `restoreMocks`/`clearMocks` anywhere (the file is `vitest.config.shared.ts`, not `.js`, but same conclusion). Not taking the fix as scoped, for two reasons:

- Trailing `mockRestore()` is the house convention: 18 files under `packages/components/src` use it, and `vi.restoreAllMocks()` appears in **zero** test files. An `afterEach` here makes this file the outlier without helping the other 18.
- The cited precedent doesn't exist yet — #5638 is still open, so there are no `ListView` tests on `main` to stay consistent with.

The real fix is `restoreMocks: true` in `vitest.config.shared.ts`, which covers every file at once and lets all those manual calls go. Happy to open that as a separate chore if you think it's worth it — it shouldn't ride along in a `SelectList` patch.

packages/components/src/SelectList/SelectList.tsx
sebald: Waiving explicitly, per your own trace: the only runtime edit in this PR is deleting a dev-only `console.warn`. Everything else is a type-union entry, a comment or a test, so there is no render path for Chromatic to diff — a run would confirm a no-op at the cost of a run.

Treating this as closed rather than open: waived, not skipped. I'll tick the checklist box to match.

packages/components/src/SelectList/SelectList.tsx
sebald: Correction to my reply above: the render-site comment is gone too. e171df4 removes all three — this block, the `layout="grid"` comment I said would stay, and the `SelectListOption.tsx` one — so those two files now carry no added prose at all, just the `RemoveProps` entries and the deleted `console.warn`.

The rationale still has three homes: the changeset, the new `MIGRATION-v18.md` §A4.6 bullet, and the comment on `moves focus between rows on ArrowRight/ArrowLeft when horizontal`. That last one is the load-bearing guard anyway — it fails if `layout="grid"` is ever changed to `"stack"`, which a comment could never do.

sarahgm (approved):

PR #5707, opened 2026-08-05
refactor(DST-1627): extract shared CalendarBody for Calendar/RangeCalendar layout

sarahgm:
# Description

The multi-month and single-month layout markup was duplicated verbatim in `Calendar` and `RangeCalendar`, so every layout tweak had to be made twice and the two copies could silently drift. Both components now render that markup through one internal `<CalendarBody>`, which reads the theme classes, visible months, min/max value, disabled state and a required `isRange` flag off the existing `CalendarContext`.

The extraction also centralizes the dropdown view type. `main` declared `type ViewMapKeys = 'month' | 'year'` separately in both files and needed a `selectedDropdown as ViewMapKeys` cast at each use site. That becomes one `CalendarDropdownView` in `Context.tsx`, adopted by the three dropdown components in place of their looser `string` setter types, and the cast is gone.

Pure refactor: no visual, DOM or behavior change. The duplication predates #5631 and was filed as a follow-up rather than expanding that PR's scope.

Two decisions worth a reviewer's attention:

- **`isRange` is passed explicitly** rather than derived from `RangeCalendarStateContext`, because subscribing to RAC's state context would defeat the element-identity render bailout. It exists to keep the touch pointerup guard (DSTSUP-257) range-only, and it is a required field so a future provider cannot omit it and silently drop that guard.

  The flag's two directions are not symmetric, which `73ebe4e9` records in a comment at the `isRange: false` site. Omitting the guard for `RangeCalendar` reintroduces DSTSUP-257, and `RangeCalendar`'s "does not commit an in-progress range when tapping a month" story test covers that: flipping `isRange` to `false` kills that test and no other. A wrong `true` on the non-range `Calendar` has no observable effect, because the guard only stops pointerup from reaching `window` and plain `useCalendar` has no window range-commit to suppress. Flipping it to `true` leaves all 69 story tests and 97 unit tests passing, so there is no black-box assertion to add for that direction.
- **`selectedDropdown` stays owned by both parents** instead of moving into `<CalendarBody>`, because a picker's preset view switch unmounts the body and an open dropdown has to survive that. Both `DatePicker` and `DateRangePicker` now carry a `PresetsMobile` story test for it.

Closes DST-1627

## Screenshots / Preview

No visual change is expected anywhere, so a Before/After table would be misleading. The proof for this PR is a **zero-diff Chromatic run** rather than screenshots.

[Chromatic build 190](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=190) on `80973b9b6`, the current head: **"No changes were found in this build."** 742 stories across 97 components, 64 snapshots captured via TurboSnap. ([Build 188](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=188) was the same result on the earlier `73ebe4e9`.)

## Test Instructions

1. `pnpm sb`, then open `Components/Calendar` and `Components/RangeCalendar`. Confirm the single-month and multi-month (`visibleDuration`) layouts look unchanged, and that the month and year dropdowns still open, select and close.
2. In `RangeCalendar`, using touch or emulated touch, tap a start date to begin a range, then open the month dropdown and tap a month. The in-progress range must not commit (DSTSUP-257). The `Basic` story test "does not commit an in-progress range when tapping a month" covers this. Anchoring the range first is what makes the guard observable, so the earlier "Commits a month selection via touch" test does not cover it.
3. At phone width in `DatePicker` and `DateRangePicker` with presets, open the month dropdown, tap "Quick selection", then tap "Back". The dropdown must still be open.
4. `pnpm test:sb` and `pnpm test:unit`.

Verified locally: `typecheck:only` clean, `eslint` 0 errors, 69 story tests and 97 unit tests passing across Calendar, RangeCalendar, DatePicker and DateRangePicker. The new test was mutation-checked both ways, failing with the guard removed and passing with it in place.

## Breaking Changes

No.

`CalendarContextValue.isRange` changed from optional to required, but that interface is internal and is not exported from `packages/components/src/index.ts`, so no consumer can observe it.

## Checklist

- [x] Visual regression tests updated (for UI changes), see [build 190](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=190), zero diff

The three blank items, left blank rather than ticked because none of them happened:

- **Unit tests**: no new unit test. The extraction is covered by the existing 97 unit tests plus three new story tests, the two `PresetsMobile` dropdown-survival ones and the `RangeCalendar` range-commit regression, and the Coverage Report check passes.
- **Component documentation**: `<CalendarBody>` is internal and is not exported from `packages/components/src/index.ts`. No public API changed, so the `Calendar` and `RangeCalendar` docs are untouched.
- **Accessibility**: no new or changed interactive component. Every role, accessible name and focus behavior comes from subcomponents this PR does not touch (`CalendarGrid`, `CalendarHeader`, `CalendarListBox`, `MonthControls`), and the only addition to the rendered tree is a `ref` on the existing overlay `div`. So there was nothing new to audit against the APG.

### DIFF
--- a/packages/components/src/Calendar/Calendar.tsx
+++ b/packages/components/src/Calendar/Calendar.tsx
@@ -129,6 +129,7 @@ const _Calendar = ({
         minValue,
         maxValue,
         disabled,
+        // No window pointerup commit to guard here, unlike RangeCalendar (DSTSUP-257).
         isRange: false,
       }}
     >
--- a/packages/components/src/Calendar/Calendar.tsx
+++ b/packages/components/src/Calendar/Calendar.tsx
@@ -3,18 +3,9 @@ import type RAC from 'react-aria-components';
 import { Calendar, DateValue } from 'react-aria-components/Calendar';
 import { DatePickerStateContext } from 'react-aria-components/DatePicker';
 import { WidthProp, cn, createWidthVar, useClassNames } from '@marigold/system';
-import { CalendarGrid } from './CalendarGrid';
-import { CalendarHeader } from './CalendarHeader';
-import { CalendarListBox } from './CalendarListBox';
+import { CalendarBody } from './CalendarBody';
 import { CalendarPresets, CalendarPresetsShell } from './CalendarPresets';
-import { CalendarContext } from './Context';
-import MonthControls from './MonthControls';
-import MonthListBox from './MonthListBox';
-import YearListBox from './YearListBox';
-import {
-  hasOnlyOneSelectableMonth,
-  hasOnlyOneSelectableYear,
-} from './calendarListBoxSelectableCheck';
+import { CalendarContext, type CalendarDropdownView } from './Context';
 import type { DatePreset } from './presets';
 
 // Props
@@ -79,8 +70,6 @@ export interface CalendarProps extends Omit<
   presets?: readonly DatePreset[];
 }
 
-type ViewMapKeys = 'month' | 'year';
-
 // Component
 // ---------------
 const _Calendar = ({
@@ -98,7 +87,6 @@ const _Calendar = ({
   ...rest
 }: CalendarProps) => {
   const visibleMonths = visibleDuration?.months ?? 1;
-  const isMultiMonth = visibleMonths > 1;
   const hasPresets = !!presets?.length;
 
   const props: RAC.CalendarProps<DateValue> = {
@@ -114,8 +102,10 @@ const _Calendar = ({
 
   const classNames = useClassNames({ component: 'Calendar', size, variant });
 
+  // Lives here, not in <CalendarBody>: a picker's preset view switch unmounts
+  // the body, and an open dropdown has to survive that.
   const [selectedDropdown, setSelectedDropdown] = useState<
-    ViewMapKeys | undefined
+    CalendarDropdownView | undefined
   >();
 
   const pickerState = use(DatePickerStateContext);
@@ -124,66 +114,11 @@ const _Calendar = ({
     'calendar'
   );
 
-  const ViewMap = {
-    month: <MonthListBox setSelectedDropdown={setSelectedDropdown} />,
-    year: <YearListBox setSelectedDropdown={setSelectedDropdown} />,
-  } satisfies { [key in ViewMapKeys]: React.JSX.Element };
-
-  const content = isMultiMonth ? (
-    <div className={classNames.calendarContainer}>
-      {[...Array(visibleMonths).keys()].map(i => (
-        <div key={i} className={classNames.calendarMonth}>
-          <CalendarHeader
-            monthOffset={i}
-            showPrevious={i === 0}
-            showNext={i === visibleMonths - 1}
-          />
-          <CalendarGrid offset={{ months: i }} />
-        </div>
-      ))}
-    </div>
-  ) : (
-    <>
-      <div
-        className={cn(
-          'pointer-events-none absolute top-1/2 left-0 w-full -translate-y-1/2 opacity-0',
-          selectedDropdown && 'pointer-events-auto opacity-100'
-        )}
-      >
-        {ViewMap[selectedDropdown as ViewMapKeys]}
-      </div>
-
-      <div
-        className={cn(
-          'flex flex-col',
-          selectedDropdown && 'pointer-events-none opacity-0'
-        )}
-      >
-        <div className="mb-6 flex items-center justify-between gap-4 max-sm:gap-2">
-          <div className="flex w-fit gap-4 max-sm:gap-3">
-            <CalendarListBox
-              key="month"
-              type="month"
-              isDisabled={
-                hasOnlyOneSelectableMonth(minValue, maxValue) ||
-                props.isDisabled
-              }
-              setSelectedDropdown={setSelectedDropdown}
-            />
-            <CalendarListBox
-              key="year"
-              type="year"
-              isDisabled={
-                hasOnlyOneSelectableYear(minValue, maxValue) || props.isDisabled
-              }
-              setSelectedDropdown={setSelectedDropdown}
-            />
-          </div>
-          <MonthControls />
-        </div>
-        <CalendarGrid />
-      </div>
-    </>
+  const body = (
+    <CalendarBody
+      selectedDropdown={selectedDropdown}
+      setSelectedDropdown={setSelectedDropdown}
+    />
   );
 
   return (
@@ -194,6 +129,7 @@ const _Calendar = ({
         minValue,
         maxValue,
         disabled,
+        isRange: false,
       }}
     >
       <Calendar
@@ -217,10 +153,10 @@ const _Calendar = ({
               <CalendarPresets presets={presets} {...presetProps} />
             )}
           >
-            {content}
+            {body}
           </CalendarPresetsShell>
         ) : (
-          content
+          body
         )}
       </Calendar>
     </CalendarContext>
--- a/packages/components/src/Calendar/CalendarBody.tsx
+++ b/packages/components/src/Calendar/CalendarBody.tsx
@@ -15,10 +15,7 @@ import {
 // Props
 // ---------------
 interface CalendarBodyProps {
-  /**
-   * Which month/year dropdown is open, if any. Owned by the calendar so it
-   * outlives a preset view switch, which unmounts this component.
-   */
+  /** Which month/year dropdown is open, if any. Owned by the parent, see there for why. */
   selectedDropdown?: CalendarDropdownView;
   setSelectedDropdown: Dispatch<
     SetStateAction<CalendarDropdownView | undefined>
--- a/packages/components/src/Calendar/CalendarBody.tsx
+++ b/packages/components/src/Calendar/CalendarBody.tsx
@@ -1,0 +1,124 @@
+import { type Dispatch, type SetStateAction, useCallback } from 'react';
+import { cn } from '@marigold/system';
+import { CalendarGrid } from './CalendarGrid';
+import { CalendarHeader } from './CalendarHeader';
+import { CalendarListBox } from './CalendarListBox';
+import { type CalendarDropdownView, useCalendarContext } from './Context';
+import MonthControls from './MonthControls';
+import MonthListBox from './MonthListBox';
+import YearListBox from './YearListBox';
+import {
+  hasOnlyOneSelectableMonth,
+  hasOnlyOneSelectableYear,
+} from './calendarListBoxSelectableCheck';
+
+// Props
+// ---------------
+interface CalendarBodyProps {
+  /**
+   * Which month/year dropdown is open, if any. Owned by the calendar so it
+   * outlives a preset view switch, which unmounts this component.
+   */
+  selectedDropdown?: CalendarDropdownView;
+  setSelectedDropdown: Dispatch<
+    SetStateAction<CalendarDropdownView | undefined>
+  >;
+}
+
+// Component
+// ---------------
+/**
+ * Grid layout shared by `Calendar` and `RangeCalendar`. Renders unwrapped, so
+ * the dropdown overlay keeps positioning against the calendar root.
+ */
+export const CalendarBody = ({
+  selectedDropdown,
+  setSelectedDropdown,
+}: CalendarBodyProps) => {
+  const { classNames, visibleMonths, minValue, maxValue, disabled, isRange } =
+    useCalendarContext();
+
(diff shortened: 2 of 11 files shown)


packages/components/src/Calendar/CalendarBody.tsx
```diff
@@ -0,0 +1,124 @@
...
+    useCalendarContext();
+
+  // Range only: `useRangeCalendar` commits an in-progress range on any window
+  // `pointerup` outside a button. `usePress` listens on `document`, that commit on
+  // `window`, so stopping overlay pointerups at `document` (not the node, not
+  // `window`) keeps press handling alive and never reaches the commit (DSTSUP-257).
+  // Native listener because react-aria's is native too.
+  const dropdownOverlayRef = useCallback(
+    (node: HTMLDivElement | null) => {
+      if (!node || !isRange) return;
```
sebald, line 45: ⚠️ **This branch has no test coverage, contrary to the description.**

I flipped `isRange` to `false` on `RangeCalendar` — disabling this guard entirely — and all 24 RangeCalendar story tests stayed green. "Commits a month selection via touch" taps the month dropdown without ever starting a range, so there is no in-progress range to commit and the guard is invisible to it.

The guard itself is load-bearing (I verified it fails a test that *does* anchor a range first), and it's preserved correctly here. The issue is that `isRange` was made a required field specifically so this can't be silently dropped — and nothing would catch it if it were.

See the review comment for a ready-to-paste test that passes with the guard on and fails with it off.

packages/components/src/Calendar/Calendar.tsx
```diff
@@ -114,8 +102,10 @@ const _Calendar = ({
 
   const classNames = useClassNames({ component: 'Calendar', size, variant });
 
+  // Lives here, not in <CalendarBody>: a picker's preset view switch unmounts
```
sebald, line 105: 💡 This rationale appears three times: here, verbatim again at `RangeCalendar.tsx:150`, and restated in `CalendarBody`'s `selectedDropdown` JSDoc.

Two copies at most. The parents are where someone would actually delete the state, so keeping it here and having the prop doc point back (rather than restate) would carry the same protection for a third of the words.

packages/components/src/Calendar/Context.tsx
```diff
@@ -5,15 +5,19 @@ import { RangeCalendarStateContext } from 'react-aria-components/RangeCalendar';
...
+export type CalendarDropdownView = 'month' | 'year';
 
 export interface CalendarContextValue {
   classNames: CalendarSharedClassNames;
   visibleMonths: number;
   minValue?: DateValue | null;
   maxValue?: DateValue | null;
   disabled?: boolean;
+  /** Whether this is a `<RangeCalendar>`. Statically known by both parents. */
+  isRange: boolean;
```
sebald, line 19: 💡 Optional, take it or leave it — this is a range-only discriminator on shared infrastructure.

Leaving the guard in `RangeCalendar` and passing the ref down as an optional `overlayRef` prop would delete this field, its JSDoc, `isRange: false` and its comment in `Calendar.tsx`, and the `[isRange]` dep — and would remove the "a future provider omits it" worry structurally rather than by type.

It preserves the render bailout you were protecting: a parent `useCallback` with `[]` deps never subscribes to state context, so `CalendarBody` still renders 0 times during a range interaction (I measured it).

This cuts against your stated preference for context over prop-plumbing, so genuinely your call — not worth churning a green PR unless you like the shape.

sebald (commented):
## Automated Code Review

Reviewed at `73ebe4e9` in a clean worktree. I mutation-tested the three design decisions rather than reading them, so the notes below are measured.

### What checks out

- `typecheck:only` clean; `eslint` 0 errors, and `CalendarBody.tsx` emits nothing at all.
- Exactly 68 story tests and 97 unit tests pass — matching the description.
- VRT is green **on the current head** (build 188, 741 unchanged), so the zero-diff claim holds.
- **The "don't subscribe to RAC's state context" rationale is correct, and I measured it.** Same interaction (6 hovers → click → 6 hovers): this design gives **0** `CalendarBody` renders; deriving `isRange` via `use(RangeCalendarStateContext)` gives **7**. The control (`CalendarListBox`, a genuine state subscriber) is 14 either way, so the probe was live. Good call — that isn't obvious from reading.
- **`selectedDropdown` ownership is properly guarded.** Moving the state into `CalendarBody` fails exactly the two new `PresetsMobile` tests and nothing else.
- Pure-refactor claim holds: `props.isDisabled` is provably identical to the context's `disabled` in both parents (`isDisabled` is in `RemovedProps`, so `...rest` can't reintroduce it), and there are only two providers, so required `isRange` is compile-time enforced.

The unticked checklist boxes each carrying a written justification, and the Breaking-Changes note on the optional→required change, are both appreciated.

### Should fix: the DSTSUP-257 coverage claim isn't true

Test Instructions step 2 and commit `73ebe4e9`'s message both credit `Basic` → "Commits a month selection via touch" with covering DSTSUP-257. It doesn't. I set `isRange: false` on `RangeCalendar` — deleting the guard outright — and **all 24 RangeCalendar story tests stayed green**. That test taps the month dropdown without ever starting a range, and DSTSUP-257 is about an *in-progress* range being committed; with nothing to commit, the guard is invisible to it.

No defect in the shipped code — the guard is preserved exactly as before. But `isRange` was made *required* precisely so the guard can't be silently dropped, and the braces for that belt don't exist. The gap predates this PR; this PR makes it load-bearing and then reports it as covered.

It's closable. This passes with the guard on and fails with it off (I ran both):

```tsx
Basic.test(
  'does not commit an in-progress range when tapping a month',
  {
    args: { defaultFocusedValue: new CalendarDate(2026, 8, 1), onChange: fn() },
  },
  async ({ args, canvas, userEvent }) => {
    const tap = (target: Element) =>
      userEvent.pointer([{ keys: '[TouchA>]', target }, { keys: '[/TouchA]' }]);

    // Anchor an in-progress range: start picked, end not yet.
    await tap(canvas.getByLabelText(/Wednesday, August 12, 2026/i));
    await expect(args.onChange).not.toHaveBeenCalled();

    await tap(canvas.getByRole('button', { name: 'Aug' }));
    const monthList = canvas.getByRole('listbox', { name: 'month' });
    await tap(within(monthList).getByRole('option', { name: /Mar/i }));

    // The in-progress range must survive the trip (DSTSUP-257).
    await expect(args.onChange).not.toHaveBeenCalled();
  }
);
```

Minimum: correct the two statements. Better: add the test.

### Smaller notes

1. **Rationale stated three times** — `Calendar.tsx:105`, verbatim again at `RangeCalendar.tsx:150`, and restated in `CalendarBody`'s `selectedDropdown` JSDoc. Two is plenty; the parents are where someone would delete the state, so keep it there and let the prop doc point rather than repeat.
2. **`CalendarDropdownView` isn't mentioned in the description** — centralizing it and replacing the three `string` setter types is a real improvement (it kills the `as ViewMapKeys` cast) and a natural consequence of the extraction, but a reviewer reading only the body meets it cold. One line would cover it.
3. **Optional, take it or leave it:** `isRange` is a range-only discriminator threaded through shared infrastructure. Leaving the guard in `RangeCalendar` and passing the ref down as an optional `overlayRef` prop would delete the required context field, its JSDoc, `isRange: false` and its comment, and the `[isRange]` dep — and would remove the "a future provider omits it" worry structurally rather than by type. It keeps the 0-render bailout (a parent `useCallback` with `[]` deps never subscribes to state context). It runs against your stated preference for context over prop-plumbing, so genuinely your call — not worth churning a green PR unless you like it.
4. **Nit:** the JSDoc on `CalendarDropdownView` (`Context.tsx:10`) — a `'month' | 'year'` union under that name already says it. The rest of the comments earn their place; the 5-line pointerup one especially, and it's shorter than the 8 lines it replaced.

*Generated with Claude Code `review-pr` skill*

packages/components/src/Calendar/Context.tsx
```diff
@@ -5,15 +5,19 @@ import { RangeCalendarStateContext } from 'react-aria-components/RangeCalendar';
 import type { ComponentClassNames } from '@marigold/system';
 
 export type CalendarSharedClassNames =
-  | ComponentClassNames<'Calendar'>
-  | ComponentClassNames<'RangeCalendar'>;
+  ComponentClassNames<'Calendar'> | ComponentClassNames<'RangeCalendar'>;
+
+/** Which of the calendar header's dropdowns is open, if any. */
```
sebald: 💡 Nit: a `'month' | 'year'` union under the name `CalendarDropdownView` already says this.

Droppable. Noting it only because you asked for a comment-necessity pass — the rest of the comments in this PR earn their place, the 5-line pointerup one especially, and it's shorter than the 8 lines it replaced.

packages/components/src/Calendar/CalendarBody.tsx
sarahgm: You're right, and thanks for measuring it rather than reading it. Test added in `80973b9b6`, and I mutation-checked it both ways before trusting it.

I went with a simpler variant than your version: the meta args already put the view on August 2025, so anchoring on Aug 12 2025 needs no `defaultFocusedValue`, just `onChange: fn()`.

With the guard in place: 25 passed. With `isRange` flipped to `false` on `RangeCalendar`: 1 failed, 24 passed, and the failure is `onChange` receiving a range whose start is Aug 12 2025, which is the DSTSUP-257 signature exactly. So it is a live probe and it is specific to this guard.

Both statements are corrected too. Test Instructions step 2 now names the new test and says explicitly that anchoring the range first is what makes the guard observable. The commit body records that `73ebe4e9`'s message credited the wrong test and why that test cannot see the guard, so the reasoning does not have to be re-derived from scratch a second time.

packages/components/src/Calendar/Calendar.tsx
sarahgm: Fixed in `80973b9b6`. Kept both parent copies, since as you say those are the two sites where someone would actually delete the state, and cut the `CalendarBody` prop doc down to a pointer:

```ts
/** Which month/year dropdown is open, if any. Owned by the parent, see there for why. */
```

packages/components/src/Calendar/Context.tsx
sarahgm: Taking the "leave it" branch, but the reasoning is worth stating rather than just declining.

The required field's stated job was "a future provider cannot omit it and silently drop the guard". Your other finding is that this protection did not actually exist: nothing failed when the guard went away. Now something does, and it fails on the mutation and nothing else does. So the braces the required field was standing in for are real now, and they hold whether the flag arrives via context or via a prop.

That leaves the restructure as a shape preference rather than a protection gain, which is not enough to churn a green PR over. Noted for if this context grows a second range-only field, because at that point the discriminator-on-shared-infrastructure smell you are pointing at stops being theoretical.

Thanks for measuring that the render bailout survives it. That was the part I would have had to re-derive.

packages/components/src/Calendar/Context.tsx
sarahgm: Dropped in `80973b9b6`. Agreed, the name carries it.

sebald (approved):
## Approving — re-reviewed at `80973b9b6`

All four notes addressed. I re-ran the mutation check rather than taking the reply on trust:

- **The DSTSUP-257 gap is genuinely closed.** 25 RangeCalendar story tests pass at this head. Flipping `isRange` back to `false` — deleting the guard — fails **exactly one** test, the new one, and `onChange` receives a range starting `2025-08-12`, the DSTSUP-257 signature. Live probe, specific to this guard.
- **The simplification is sound, not lucky.** Dropping `defaultFocusedValue` works because `meta.args` pins `defaultValue` to Aug 7–14 **2025**, so the August view and the Aug 12 cell don't depend on today's date. I checked that before agreeing with it.
- Counts in the description match exactly: 69 story tests, 97 unit tests across the four components.
- VRT is green on this head — **Chromatic build 190, "No changes were found"**, 742 stories across 97 components. That supersedes the build-188 link in the checklist; the +1 story is the new `disableSnapshot` test, so the snapshot set is unchanged as described.

On the `overlayRef` note: taking the "leave it" branch was the right call, and for a better reason than I gave. You spotted that the required field's protection didn't actually exist when I wrote that note — nothing failed when the guard went away. Now that it does, the protection holds whether the flag arrives via context or via a prop, which demotes the restructure to a shape preference. Agreed on the trigger for revisiting it: a second range-only field in that context.

Nothing blocking left.

PR #5708, opened 2026-08-05
docs(DST-1687): document the data-entry grid pattern

sebald:
# Description

Extends the `table-record-management` decision framework from three options to four by documenting the **data-entry grid**, a `<Table>` of always-on form controls with `keyboardNavigationBehavior="tab"`.

Documentation, demos, and story tests only. **No component source change.** The prop already flows through our `{...props}` spread into RAC's `useTableState` and is not in `Table`'s `RemovedProps`, so nothing needed plumbing.

Closes DST-1687

### Why the prop is required, not a refinement

Without it the grid keeps the arrow keys, so pressing one inside a cell field throws focus onto the `<tr>` instead of moving the caret. A form control in a plain `Table.Cell` is keyboard-broken today.

### What the prop actually does

It moves Tab **in and out of a cell**, not from one cell to the next. Verified by hand in both Chromium and Firefox:

- Reaching a neighbouring field stays a grid movement and costs three keys (`Shift+Tab`, arrow, `Tab`)
- `Tab` pressed inside a field leaves the table altogether

That cost is what sets the documented limits of five editable columns and roughly twenty five rows, and it is why no table can offer the field to field tabbing a form does. [APG models a grid as a single tab stop](https://www.w3.org/WAI/ARIA/apg/patterns/grid/).

### Contents

- **New demo** `table-record-entry-grid` — a venues grid with a `Table.Footer` total and one batch Save
- **New demo** `table-record-entry-grid-compare` — one dataset as inline edit, grid, and drawer, placed in *Match the approach to the task* where the choice is actually made
- **Pattern page** — new *Filling in many rows at once* section, guardrail bullets, a `<Do>`/`<Dont>` pair, and a non-TOC *Notes on accessibility* subsection carrying the key model
- **Table component docs** — a two-line pointer from *Inline editing* to the pattern
- **Two story tests** — caret movement inside cell fields, the Tab round trip, and row checkboxes staying reachable under `selectionMode="multiple"`

## Test Instructions

1. `pnpm build && pnpm start`, open `/patterns/user-input/table-record-management`
2. In the *Seat allocation per venue* grid, click a Capacity field and press <kbd>←</kbd> or <kbd>Home</kbd>. The caret moves through the value instead of focus jumping to the row.
3. Press <kbd>Shift</kbd>+<kbd>Tab</kbd> to step out onto the cell, an arrow key to move cells, then <kbd>Tab</kbd> to step into that field.
4. Temporarily change the prop to `keyboardNavigationBehavior="arrow"` and repeat step 2 to see the broken behaviour the callout describes.
5. `pnpm test:sb --run packages/components/src/Table/Table.stories.tsx` — 38 pass.

### Known limitation, deliberately out of scope

`NumberField` + <kbd>Home</kbd> still reaches the grid in `tab` mode, likely upstream in `useSpinButton`. Recorded as a comment on DST-1687 rather than worked around here.

## Breaking Changes

No

## Checklist

- [x] Unit tests added/updated — story tests, per the repo's browser-test convention
- [x] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) — the single-tab-stop trade-off is documented rather than worked around
- [x] Visual regression tests updated — n/a, both stories set `chromatic: { disableSnapshot: true }`, matching the 10 existing Table story tests. The docs demos carry appearance coverage.
- [x] Changeset added — n/a, `@marigold/docs` is `private: true` and stories are not emitted to `dist`, so nothing consumer-visible changes.


packages/components/src/Table/Table.stories.tsx
```diff
@@ -1952,3 +1953,136 @@ StickyFooter.test(
...
+      async () => {
+        first.setSelectionRange(first.value.length, first.value.length);
+        const before = first.selectionStart as number;
+        await userEvent.keyboard('{ArrowLeft}');
+        expect(document.activeElement).toBe(first);
+        expect(first.selectionStart).toBe(before - 1);
+      }
+    );
+
+    // Home/End are deliberately not asserted here. What those keys do inside an
```
sarahgm: comments can be shorten for the stories

packages/components/src/Table/Table.stories.tsx
```diff
@@ -1952,3 +1953,136 @@ StickyFooter.test(
...
+    // input is platform dependent (on macOS they do not move the caret at all),
+    // and inside a NumberField they still reach the grid, so they are not part
+    // of what this prop reliably fixes.
+
+    // The prop moves Tab in and out of a cell, not from one cell to the next,
+    // so reaching the neighbouring field is a three key round trip.
+    await step('Shift+Tab steps back out onto the cell', async () => {
+      await userEvent.keyboard('{Shift>}{Tab}{/Shift}');
```
sarahgm, line 2050: missing blank lines in some steps like here

sarahgm: also the following ones

packages/components/src/Table/Table.stories.tsx
sebald: Done in f7972d4bd. Cut the four-line Home/End note and the five-line Tab note down to two lines each (keeping only why the assertion is absent), dropped the "prop moves Tab in and out of a cell" comment since the two step names below it already say that, and shortened the JSDoc. Also moved the `DataEntryGrid` description onto the exported story — it was sitting on the render helper, where Storybook never picks it up.

Went through the demo comments too while I was at it: the three `// Reach for this when …` comments in the compare demo duplicated the `<Text>` copy the tabs already render, so those are gone.

packages/components/src/Table/Table.stories.tsx
sebald: Done in f7972d4bd — added blank lines separating arrange/act/assert in all five new steps, not just this one, to match the rest of the file.

sebald:
@sarahgm both of your points are addressed in f7972d4bd, threads resolved.

- **Shorter comments** — trimmed the long notes in the test down to the reason an assertion is absent, and dropped the one the step names below it already stated. Went through the two demos as well: the three `// Reach for this when …` comments in the compare demo were duplicating the copy the tab panels already render, so those are gone. Also moved the `DataEntryGrid` description onto the exported story — it was on the render helper, where Storybook never picked it up.
- **Blank lines** — arrange/act/assert separated in all five new steps, not just the one you flagged.

Also removed the stale stacked-PR note from the description, since #5699 merged and this is on `main` now.

All 24 checks green. Ready for another look when you have a moment.

docs/content/patterns/user-input/table-record-management/index.mdx
```diff
@@ -97,6 +105,66 @@ The loop only feels good when the details are handled with care:
...
+    <DontDescription>
+      Don't turn a wide table into a grid of inputs. Eight editable columns
+      across fifteen rows is a hundred and twenty fields to step through three
+      keys at a time, and heavy controls like date pickers in every row crowd
+      the row past reading. Use a drawer.
+    </DontDescription>
+  </Dont>
+</GuidelineTiles>
+
+#### Notes on accessibility [!toc] [#accessibility]
```
sarahgm: 🔴 **`[!toc]` renders as literal text, and this heading skips a level**

fumadocs-core 16.9.3's `remarkHeading` only supports the trailing `[#slug]` form — there is no `[!toc]` handling. Running the actual plugin over this line:

```
OUTPUT MD: #### Notes on accessibility \[!toc]
TOC: [{"title":"Notes on accessibility [!toc]", "url":"#accessibility", "depth":4}]
```

So the published page shows **"Notes on accessibility [!toc]"**. It is also unnecessary: `docs/app/(docs)/[...slug]/page.tsx:53` already filters the TOC to `depth <= 3`, so an `h4` never reaches it.

Separately, `## Filling in many rows at once` (h2) → `#### Notes on accessibility` (h4) skips h3, which breaks the heading outline for anyone navigating by level — in the section about accessibility. The sibling Bulk Actions page uses a plain `## Accessibility`.

```suggestion
### Notes on accessibility [#accessibility]
```

docs/content/patterns/user-input/table-record-management/table-record-entry-grid.demo.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const changed = rows.filter((row, i) => {
+    const before = saved[i];
+    return (
+      row.capacity !== before.capacity ||
+      row.price !== before.price ||
+      row.note !== before.note
+    );
+  }).length;
+
+  const totalCapacity = rows.reduce((sum, row) => sum + row.capacity, 0);
```
sarahgm: 🔴 **Clearing a Capacity field makes the footer render `Total: NaN`**

RAC's NumberField commits `NaN` when the input is emptied (`useNumberFieldState` → `commit()` → `if (!newInputValue.length) setNumberValue(NaN)` → `onChange(NaN)`). Verified against the real component: the last `onChange` arg is `NaN`, and `(500 + NaN).toLocaleString('en-US')` is the string `"NaN"`, which is what line 140 prints.

Clearing a field is the most natural thing to do in a data-entry grid, and this is the demo the page points at right after the guardrail *"Validate in the field, and summarise on submit."*

```suggestion
  const totalCapacity = rows.reduce(
    (sum, row) => sum + (Number.isNaN(row.capacity) ? 0 : row.capacity),
    0
  );
```

The compare demo's `GridApproach` has the same `NaN`-into-state path (no total there, so it only skews the changed count). Given the guardrail this section preaches, putting an `error` on the empty field would make the demo teach its own rule.

docs/content/patterns/user-input/table-record-management/index.mdx
```diff
@@ -97,6 +105,66 @@ The loop only feels good when the details are handled with care:
   because a blank field is safer than a wrong default.
 </Callout>
 
+## Filling in many rows at once
```
sarahgm, line 108: 🔴 **Missing `@marigold/docs: patch` changeset**

The checklist reasons from `private: true`, but that only stops npm publishing. `@marigold/docs` is versioned (`18.0.0-rc.5`), has its own `CHANGELOG.md`, and sits in the `fixed` group with `@marigold/system` and `@marigold/components` in `.changeset/config.json`. Without a changeset, a whole new pattern section lands with no changelog or release-notes entry.

Direct precedent is the sibling pattern page: `.changeset/dst-1383-bulk-actions-pattern.md` is `@marigold/docs: patch` for exactly this kind of work, as are `docs-punctuation-scrub` and `adopt-sidebar-current-prop`.

packages/components/src/Table/Table.stories.tsx
```diff
@@ -1952,3 +1953,136 @@ StickyFooter.test(
...
+  parameters: { chromatic: { disableSnapshot: true } },
+  args: {
+    keyboardNavigationBehavior: 'tab',
+    selectionMode: 'multiple',
+  },
+  render: args => <DataEntryGridTable {...args} />,
+});
+
+DataEntryGridWithSelection.test(
+  'Row checkboxes stay reachable and selectable',
```
sarahgm: 💡 **This test doesn't assert what its name says**

*"stay reachable"* is only checked as presence plus a mouse click. I re-ran it with `keyboardNavigationBehavior: 'arrow'` and it still passes, so it isn't guarding the per-row tab stop the docs describe.

Either tab to a checkbox, or rename to *"Row checkboxes render and toggle"*. Worth noting the `DataEntryGrid` test above **is** load-bearing — it fails at the ArrowLeft caret assertion (line 2028) in arrow mode.

docs/content/patterns/user-input/table-record-management/table-record-entry-grid.demo.tsx
```diff
@@ -0,0 +1,176 @@
...
+          <Button
+            variant="secondary"
+            size="small"
+            disabled={changed === 0}
+            onPress={() => setRows(saved)}
+          >
+            Reset
+          </Button>
+          {/* One explicit Save for the whole grid, not per cell. */}
+          {/* [!code highlight:7] */}
```
sarahgm: 💡 The `<Button>` block spans 8 lines (164–171), so `</Button>` is left unhighlighted.

```suggestion
          {/* [!code highlight:8] */}
```

docs/content/patterns/user-input/table-record-management/table-record-entry-grid-compare.demo.tsx
```diff
@@ -0,0 +1,310 @@
...
+            </Table.Row>
+          ))}
+        </Table.Body>
+      </Table>
+    </Stack>
+  );
+};
+
+export default () => (
+  <Panel aria-label="Three ways to change the same table">
```
sarahgm: 💡 **Two different names for the same section**

`aria-label` wins over the visible `<Title>The same table, three ways</Title>` — `Panel.tsx` suppresses `aria-labelledby` whenever `aria-label` is set. Dropping it lets the Title supply the accessible name.

```suggestion
  <Panel>
```

docs/content/patterns/user-input/table-record-management/table-record-entry-grid-compare.demo.tsx
```diff
@@ -0,0 +1,310 @@
...
+} from '@marigold/components';
+
+interface Venue {
+  id: string;
+  name: string;
+  capacity: number;
+  price: number;
+}
+
+// One dataset, three interactions. Sharing it is what makes this a comparison
```
sarahgm: 💡 Only the *initial* array is shared. Each approach holds its own `useState(initialVenues)`, and RAC unmounts inactive `Tabs.Panel`s, so edits are discarded on tab switch. The mdx lead-in *"Try each below on one dataset"* sets the same expectation. Either lift the state to the parent or soften both.

sarahgm (changes requested):
## Automated Code Review

The pattern content is genuinely strong — the create-records vs fill-in-a-table distinction, the tab-stop arithmetic behind the ceilings, and the EditableCell-is-a-popover explanation all land, and the ticket's open questions are properly closed out. Three things to fix before merge, two of which ship visibly broken.

**Verified beyond reading:**

- The `DataEntryGrid` story test is load-bearing — flipping `keyboardNavigationBehavior` to `'arrow'` fails it at `Table.stories.tsx:2028`.
- `[!toc]` is not a directive fumadocs-core 16.9.3 supports — ran the actual `remarkHeading` plugin over the heading.
- NumberField emits `NaN` on clear — confirmed with a throwaway test against the real component.
- All props used in the demos exist, and both demos register in the demo registry.

**Visual regression: not needed.** Both new stories set `chromatic.disableSnapshot`, no existing story or shared code in `Table.stories.tsx` was touched, and everything else is `docs/**` (excluded by the workflow's `paths-ignore`).

---

docs/content/patterns/user-input/table-record-management/index.mdx
sebald: Fixed in `74bf68b4c`, but one correction: `[!toc]` does work here.

It is handled by `rehypeToc`, not `remarkHeading` — `fumadocs-core/dist/mdx-plugins/rehype-toc.js:6` defines `NoTocTag = "[!toc]"`, strips it from the heading text (`last.value = noToc`) and returns `"skip"` before `items.push`, so the heading is both cleaned and excluded. And `fumadocs-mdx` appends it to the default rehype plugins unconditionally (`dist/preset-NZprPlqc.js:48`), which our empty `mdxOptions` in `source.config.ts` inherits. Your snippet is remark-stage output — the tag is still there because rehype has not run yet.

Your other two points stand and are the reason I took the suggestion:

- **Unnecessary** — `page.tsx:53` filters to `depth <= 3`, so an h4 never reaches the TOC regardless.
- **The level skip is real.** h2 → h4 in the section about accessibility, and the sibling Bulk Actions page uses a plain `## Accessibility`.

Now `### Notes on accessibility [#accessibility]`, so it sits at the right level and appears in the TOC like its sibling.

docs/content/patterns/user-input/table-record-management/table-record-entry-grid.demo.tsx
sebald: Confirmed and fixed in `74bf68b4c`. `useNumberFieldState.mjs:87` does `setNumberValue(NaN)` on the empty commit, and line 36 shows empty ⇄ NaN is the representation, so the total was printing the string `NaN`.

Took your second suggestion too, since it is the better fix: the capacity field now sets `error={Number.isNaN(row.capacity)}` and Save is disabled while any row is invalid, so the demo follows the *"Validate in the field, and summarise on submit"* guardrail it sits under instead of contradicting it. The reduce is guarded as well, so the footer stays correct while a field is mid-edit.

Left the compare demo unguarded on purpose — it has no total, and `NaN !== saved` counting as "changed" is arguably right for an invalid row.

docs/content/patterns/user-input/table-record-management/index.mdx
sebald: You are right, fixed in `74bf68b4c` with a `@marigold/docs: patch` changeset.

My checklist reasoning was wrong: I argued from `private: true`, which only stops npm publishing. `.changeset/config.json` puts `@marigold/docs` in the `fixed` group with `@marigold/system` and `@marigold/components`, it is versioned `18.0.0-rc.5`, and it has its own `CHANGELOG.md` — so the section would have landed with no changelog entry.

Same mistake is in #5721 (the orphaned-demo cleanup), which I opened with the same n/a reasoning. Adding a changeset there too.

packages/components/src/Table/Table.stories.tsx
sebald: Good catch — renamed to *"Row checkboxes render and toggle"* in `74bf68b4c`, with a comment saying the per-row tab stop is not covered.

Went with the rename rather than tabbing to a checkbox because a Tab-based assertion would be unreliable for the same reason already noted in the `DataEntryGrid` test: the runner moves focus in DOM order instead of letting the grid handle the key, so it would pass or fail for the wrong reason. Better an honest name than a test that looks like it guards the tab stop and does not.

And agreed on `DataEntryGrid` being load-bearing — the ArrowLeft caret assertion is exactly the one that fails in arrow mode.

docs/content/patterns/user-input/table-record-management/table-record-entry-grid.demo.tsx
sebald: Fixed in `74bf68b4c` — `[!code highlight:8]`. The comment is on line 160 and the block runs 161–168, so `:7` stopped at 167 and left `</Button>` out.

docs/content/patterns/user-input/table-record-management/table-record-entry-grid-compare.demo.tsx
sebald: Confirmed against `Panel.tsx:142` — `aria-labelledby={!ariaLabel && hasTitle ? titleId : props[...]}`, so setting `aria-label` drops the Title as the accessible name. Removed in `74bf68b4c`, so *"The same table, three ways"* names the region.

docs/content/patterns/user-input/table-record-management/table-record-entry-grid-compare.demo.tsx
sebald: Right — softened both in `74bf68b4c` rather than lifting the state.

The comment now says each approach keeps its own copy and switching tabs discards edits, and the mdx lead-in says *"the same starting data … Each tab keeps its own copy, so switching tabs starts over."*

Chose that over lifting state because the three approaches deliberately model different commit semantics: inline edit commits per cell, the grid commits a batch on Save, the drawer commits per record. Sharing one state would have to pick one, which would blur the distinction the demo exists to draw.

aromko (commented):
## Automated Code Review

Reviewed against DST-1687's stated scope. Verified rather than assumed: I applied
`Table.stories.tsx` locally, flipped the arg to `keyboardNavigationBehavior: 'arrow'`,
and ran `pnpm test:sb` — the test fails at the `{ArrowLeft}` assertion, so the prop
genuinely works and the test is a real regression guard. I also regenerated
`docs/.registry/props.json` to check the props table, and confirmed `[!toc]`,
all new anchors, the orphan status of the deleted `table-input.demo.tsx`, and that
stories never reach `dist` (so no changeset is correct).

**Visual regression: not needed.** The `Table.stories.tsx` diff is purely additive,
both new stories set `chromatic: { disableSnapshot: true }`, and no component source
or theme file changed — Chromatic would capture zero new or changed snapshots.

### Critical

- **The required prop is missing from the Props table.** Two callouts say
  `keyboardNavigationBehavior="tab"` is mandatory, but `<AutoTypeTable path="Table"
  name="TableProps" />` on the same page does not list it. Regenerating
  `docs/.registry/props.json` on this branch confirms it is absent (100 entries,
  including `escapeKeyBehavior` and `treeColumn`, but not this one). This is
  DST-1687 open question 2, and scope item 2's second bullet.

### Warnings

- Heading level skipped: `##` → `####` at *Notes on accessibility*.
- The two new `####` subsections on the Table page never reach the TOC
  (`page.tsx:53` filters `depth <= 3`), so the page loses its *Inline editing* entry
  and the grid section never gets one.
- "Both work, so nothing breaks" states as fact what DST-1687 open question 1
  recorded as an inconclusive probe.

### Suggestions

- No test pins the `arrow` default, so the claim both callouts rest on can go stale
  silently.
- Home/End is uncovered; the stated blocker is `NumberField`-specific.
- `[!code highlight:7]` is one line short of the `<Button>` element.
- The compare demo's `aria-label` differs from its visible `<Title>`.

Excellent work otherwise — the prose is precise about the cost rather than selling
the pattern, and the guardrails all trace back to the ticket.

---

docs/content/components/collection/table/index.mdx
```diff
@@ -220,34 +224,36 @@ Editable cells always truncate their content, regardless of the table-level `ove
...
+
+When the task does fit, don't give a whole column editable cells. Put the form controls directly in the cells and let the table be the form. Every row stays visible for comparison while the user types, and nothing has to be opened first. See [Filling in many rows at once](/patterns/user-input/table-record-management#filling-in-many-rows-at-once) for the limits on columns and rows, which cells to keep read-only, and how the grid is saved.
+
+<ComponentDemo name="table-entry-grid" />
+
+<Callout type="warning" title="Set keyboardNavigationBehavior to tab">
```
aromko, line 249: 🔴 This callout makes the prop mandatory, but `<AutoTypeTable path="Table" name="TableProps" />` further down the same page does not list it.

I regenerated `docs/.registry/props.json` on this branch (`pnpm --filter @marigold/docs build:types`) to rule out a stale artifact — `TableProps` has 100 entries including `escapeKeyBehavior` and `treeColumn`, and no `keyboardNavigationBehavior`. It resolves for `tsc` but the extractor does not surface it. This is exactly what DST-1687 open question 2 anticipated, and scope item 2's second bullet ("Document `keyboardNavigationBehavior` under Props").

Cheapest fix, which also settles open question 3 by making the passthrough deliberate rather than an omission — in `packages/components/src/Table/Table.tsx`:

```ts
/**
 * Whether keyboard navigation between focusable elements inside cells uses the
 * arrow keys or the Tab key. Set to `tab` when cells contain always-on form controls.
 * @default 'arrow'
 */
keyboardNavigationBehavior?: 'arrow' | 'tab';
```

docs/content/patterns/user-input/table-record-management/index.mdx
aromko: ⚠️ Heading level skipped: `## Filling in many rows at once` (line 108) jumps straight to `####` with no `###` in between.

`[!toc]` keeps it out of the TOC either way, so `###` costs nothing:

```suggestion
### Notes on accessibility [!toc] [#accessibility]
```

docs/content/components/collection/table/index.mdx
```diff
@@ -209,7 +209,11 @@ Enable sorting to let users click column headers and dynamically reorder rows by
   </Dont>
 </GuidelineTiles>
 
-### Inline editing
+### Editing
+
+A table can be edited in place in two ways, and the one that fits follows from the shape of the task rather than the amount of data. Reach for **editable cells** when the table is mostly read-only and the user corrects one value at a time. Reach for a **data-entry grid** when every row needs the same fields filled in. If you are still weighing those against a drawer or a separate page, the [Table record management pattern](/patterns/user-input/table-record-management) sets all four side by side.
+
+#### Editable cells [#inline-editing]
```
aromko, line 216: ⚠️ These two `####` subsections never reach the on-this-page TOC — `docs/app/(docs)/[...slug]/page.tsx:53` filters `page.data.toc` to `depth <= 3`.

Net effect of the restructure: the page loses its existing *Inline editing* entry and the new *Data-entry grid* section never gets one, for a section this PR is specifically trying to make discoverable. It also folds into the parent chunk in `docs/lib/markdown/etl/chunker.ts` (splits on H2/H3), so docs search and the MCP see one large "Editing" chunk instead of two.

Promoting both to `###` fixes all of it — the `[#inline-editing]` anchor still preserves the old link target.

docs/content/patterns/user-input/table-record-management/index.mdx
```diff
@@ -97,6 +105,66 @@ The loop only feels good when the details are handled with care:
...
+
+Moving between two fields is a grid movement rather than a single <kbd>Tab</kbd> press, so every field the user visits costs a few keystrokes and the work grows faster than the table looks. That cost is what sets the limits.
+
+- **Five editable columns at most,** and only simple controls such as text, number, select, and switch. A date range, a lookup that opens its own popover, or a field whose validity depends on another field belongs in a drawer.
+- **A bounded set of rows,** roughly twenty five in view. Never put a data-entry grid behind infinite scroll or over an unbounded dataset.
+- **Keep everything the user does not edit read-only.** Identifiers, computed values, and totals stay plain cells, which also keeps them out of the tab order.
+- **Commit the whole grid with one Save,** and show how many rows are pending. Do not save per cell, because that is inline editing's model, and running both at once teaches two contradictory rules about when a change becomes real.
+- **Do not mix always-on inputs and `<Table.EditableCell>` in the same table.** Both work, so nothing breaks, but the user now has two different ways to edit depending on which column they are in.
```
aromko: ⚠️ "Both work, so nothing breaks" states as fact what DST-1687 open question 1 recorded as inconclusive:

> My probe was **inconclusive** — it failed identically with the prop on and off (the edit trigger sits in a `w-0 overflow-hidden` wrapper until row hover, so my synthetic click never landed). It isolates nothing and should not be read as evidence of a conflict. Needs a proper test during implementation.

Either add a story test covering `Table.EditableCell` alongside always-on inputs under `keyboardNavigationBehavior="tab"`, or soften the wording so it does not claim verification. The guidance itself (don't mix them) stands either way — it just doesn't need the "both work" premise.

packages/components/src/Table/Table.stories.tsx
```diff
@@ -1952,3 +1953,134 @@ StickyFooter.test(
...
+      ))}
+    </Table.Body>
+  </Table>
+);
+
+/**
+ * Form controls in cells need `keyboardNavigationBehavior="tab"`, otherwise the
+ * grid keeps the arrow keys and focus is pushed onto the row.
+ */
+export const DataEntryGrid = meta.story({
```
aromko, line 1991: 💡 Nothing in the suite pins the *default*, which is what the whole pattern rests on.

I checked this holds today: flipping this arg to `'arrow'` and running `pnpm test:sb` fails at line 2029 (`expect(document.activeElement).toBe(first)` after `{ArrowLeft}`), so the behaviour is real and this test is a genuine guard. But if RAC ever changes the default, both callouts go stale silently and CI stays green.

A sibling story without the arg, asserting focus leaves the field on `{ArrowLeft}`, would guard the claim the docs make.

packages/components/src/Table/Table.stories.tsx
```diff
@@ -1952,3 +1953,134 @@ StickyFooter.test(
...
+        const before = first.selectionStart as number;
+
+        await userEvent.keyboard('{ArrowLeft}');
+
+        expect(document.activeElement).toBe(first);
+        expect(first.selectionStart).toBe(before - 1);
+      }
+    );
+
+    // Home/End are not asserted: platform dependent inside an input, and still
```
aromko: 💡 Home/End is the strongest row in DST-1687's verification table (`arrow`: "focus lands on the `<tr>`" / `tab`: "stay inside the field") and it's the one behaviour left uncovered.

The stated blocker is `NumberField`-specific — the `TextField` in this same story looks assertable. Worth a step if it's stable across the runner's browsers.

docs/content/patterns/user-input/table-record-management/table-record-entry-grid.demo.tsx
```diff
@@ -0,0 +1,173 @@
...
```
aromko: 💡 Off by one — the `<Button>` element below spans 8 lines, so `</Button>` falls outside the highlight.

```suggestion
          {/* [!code highlight:8] */}
```

docs/content/patterns/user-input/table-record-management/table-record-entry-grid-compare.demo.tsx
```diff
@@ -0,0 +1,303 @@
...
+            </Table.Row>
+          ))}
+        </Table.Body>
+      </Table>
+    </Stack>
+  );
+};
+
+export default () => (
+  <Panel aria-label="Three ways to change the same table">
```
aromko: 💡 The accessible name ("Three ways to change the same table") differs from the visible title on line 261 ("The same table, three ways"), so screen reader users hear a different heading than everyone else reads.

Match the two, or drop the `aria-label` if `Panel` already derives its name from `<Title>`.

packages/components/src/Table/Table.stories.tsx
```diff
@@ -1952,3 +1953,134 @@ StickyFooter.test(
...
+  tags: ['component-test'],
+  parameters: { chromatic: { disableSnapshot: true } },
+  args: {
+    keyboardNavigationBehavior: 'tab',
+    selectionMode: 'none',
+  },
+  render: args => <DataEntryGridTable {...args} />,
+});
+
+DataEntryGrid.test(
```
aromko, line 2001: 💡 Four act-assert cycles in one test, where the house style is one Arrange–Act–Assert per test.

Flagging for your call only — the rest of this file already runs journeys of this shape (`Sorting.test` has 8 steps, `DynamicColumnsAndRows.test` has 12), so this matches the file rather than breaking from it. Splitting just these two would make them the odd ones out.

sebald:
@sarahgm all of your findings are integrated — thanks, the two 🔴 ones were real bugs that would have shipped. Everything is in `74bf68b4c`, plus `bcbcf5d78` for a TOC follow-up. All threads resolved.

### Your review

| Finding | What changed |
| --- | --- |
| 🔴 `Total: NaN` on a cleared Capacity field | Took your second suggestion, since it fixes the cause rather than the symptom: the field now sets `error={Number.isNaN(row.capacity)}` and Save is disabled while any row is invalid, so the demo models the validation it should have had. |
| 🔴 Missing `@marigold/docs: patch` changeset | Added. My checklist reasoning was wrong — I argued from `private: true`, which only stops npm publishing. |
| 🔴 Heading level skip | Fixed (see below). |
| 💡 Test name didn't match its assertion | Renamed to *"Row checkboxes render and toggle"*, with a comment noting the per-row tab stop is uncovered. Tabbing to a checkbox would pass under `arrow` too, for the reason already noted in the `DataEntryGrid` test. |
| 💡 `[!code highlight:7]` left `</Button>` out | `:8`. |
| 💡 `aria-label` overriding the visible `Title` | Removed, confirmed against `Panel.tsx:142`. |
| 💡 "one dataset" overstated the shared state | Softened both the comment and the mdx lead-in rather than lifting state — the three approaches deliberately model different commit semantics. |

### One correction

On the heading thread: you were right that it skipped a level, and that is fixed. But `[!toc]` was not the problem — it does work here.

It is handled by `rehypeToc`, not `remarkHeading`. `fumadocs-mdx`'s preset registers **both** (`preset-NZprPlqc.js:17` and `:48`), and it is `rehypeToc` that exports the `toc` binding the page reads. `rehype-toc.js:6` defines `NoTocTag = "[!toc]"`, strips it from the heading text (`last.value = noToc`) and `return "skip"`s before `items.push` — so the heading renders clean and stays out of the TOC. Your repro read `remarkHeading`'s `vfile.data.toc`, which is a different object, only used by `remark-include` for section extraction.

Worth knowing because `[!toc]` is the right tool if we ever want a real heading that stays out of the TOC.

### TOC follow-up (`bcbcf5d78`)

Sebastian noticed the aside was missing from the TOC on the preview. It turned out to be display lag — the preview predated the `####`→`###` fix, and depth 3 does render (`toc/default.js:91` only indents it). Either way it should not be listed next to the nine top-level sections, so it is bold text now rather than a deeper heading: `####` there would sit under an `##` and skip a level again, which is your original finding. Bold text has no outline entry to skip. That drops the `#accessibility` anchor — checked first, nothing links to it.

### Still open, deliberately

`NumberField` + <kbd>Home</kbd> still reaches the grid in `tab` mode, likely upstream in `useSpinButton`. Out of scope, recorded on DST-1687.

sarahgm (approved):

docs/content/components/collection/table/index.mdx
sebald: Fixed in `a91e6782a` — but the cause is not the extractor, which surfaces the prop fine. The stale artifact the page renders from is what was wrong.

`docs/scripts/build-types.ts` caches each extraction under `.next/cache/`, which Vercel preserves across deploys, and fumadocs-typescript keys those entries on `path:name:content:version` (`fumadocs-typescript/dist/index.js:68`) — the component source file only, never the `.d.ts` files the props are inherited from. `Table.tsx` last changed 2026-08-04; the react-aria bump landed 2026-08-05 in #5699. So Table's entry predates the prop, and nothing would ever invalidate it.

The cache directory is now salted with a hash of `pnpm-lock.yaml`, so any dependency move retires the generation. Superseded generations are pruned, along with the loose entry files the unsalted cache wrote straight into the root. Warm rebuild still 0.6s, cold 37s.

Verified on the deployment rather than locally, because a local regenerate is exactly what misled me first: the pre-fix preview contains 0 occurrences of the upstream description *"Whether keyboard navigation to focusable elements within the cells is…"*, the post-fix one contains 1, with `"arrow" | "tab" | undefined` and `@default 'arrow'`.

I also checked the blast radius, since a pipeline bug like this is rarely limited to one prop. Diffing react-stately 3.48→3.49 and react-aria 3.50→3.51, `keyboardNavigationBehavior` is the only added prop that reaches a Marigold-documented interface, and a full cold regenerate turns up exactly one affected table. So Table was the only casualty of this bump — the trap itself was general.

docs/content/patterns/user-input/table-record-management/index.mdx
sebald: You are right, and it was my wording. Fixed in `0301e196e` — by checking the claim rather than softening it, since that closes DST-1687 open question 1 instead of routing around it.

`DataEntryGridWithEditableCell` puts an always-on `TextField` and a `<Table.EditableCell>` in one table under `keyboardNavigationBehavior="tab"`, and asserts the always-on field keeps typing and caret movement, the editable cell still opens its editor focused, saving commits, and the always-on field is untouched by the edit. Flipping the arg to `arrow` fails at the first assertion — `expected <td …><div …></div></td> to be <input …>` — so it isolates the prop rather than passing incidentally.

Worth recording why the original probe was inconclusive: clicking the edit trigger works fine inside the story runner. The `w-0 overflow-hidden` wrapper only defeated a synthetic click from outside it.

Line 130 now says what the test shows — neither one breaks the other, so this is a consistency rule rather than a technical limit.

packages/components/src/Table/Table.stories.tsx
sebald: Added in `549321de3`. `DataEntryGridDefaultNavigation` omits the arg and asserts an arrow key is taken by the grid. Setting `keyboardNavigationBehavior: 'tab'` back on that story makes it fail, so it pins the upstream default rather than the prop.

One correction to my own first attempt: I asserted `role="gridcell"` and it failed with `role="rowheader"`. Focus does not merely fall out of the field — it moves one cell to the left, onto the `rowHeader` column. The real behaviour was a stronger demonstration of the trap than I had guessed, so the assertion names it.

packages/components/src/Table/Table.stories.tsx
sebald: Fair hit — the blocker did not justify skipping the whole row. Covered from both sides in `549321de3`:

- under `tab`, Home/End keep focus in the `TextField` and move the caret to either end
- on the default, Home leaves the field

Same key, same field, opposite outcomes, differing only by the prop, so the contrast is itself the proof of prop-dependence.

Still `TextField`-only, though. Home inside a `NumberField` continues to reach the grid, which looks like an upstream `useSpinButton` issue rather than something this table controls, so the comment now says that instead of implying the whole row is clean.

docs/content/patterns/user-input/table-record-management/index.mdx
sebald: Landed exactly as you suggested, in `0301e196e`: `### Notes on accessibility [!toc] [#accessibility]`.

Some history, since this heading has moved twice: `74bf68b4c` had already made it `###`, then `bcbcf5d78` went to bold text, because the toc is crowded and I wanted the aside out of it. Your form is the better answer — it sits directly under an `##` so nothing skips, `[!toc]` keeps it out anyway, and the `#accessibility` anchor comes back. Bold text had none of the heading semantics.

And `[!toc]` does work, which was disputed earlier in this PR. Now settled empirically on the rebuilt preview at depth 3, where the `depth <= 3` filter cannot be what excludes it: the heading renders as `as: "h3", id: "accessibility"`, `"!toc"` appears 0 times anywhere in the page, and there is no toc entry. The occurrence counts make that unambiguous rather than inferred — every section that *is* in the toc appears 4+ times, because the toc link adds its own pair, while "Notes on accessibility" appears exactly 2: the heading and its RSC payload.

docs/content/patterns/user-input/table-record-management/table-record-entry-grid.demo.tsx
sebald: Already fixed in `74bf68b4c`, before this review landed — same finding as sarahgm's r3733927023. The comment sits on line 160 and the block runs 161–168, so `:7` stopped at 167 and left `</Button>` out.

Heads up that this review was submitted against `392af7e91`, two commits behind the head at the time, so three of its findings were already addressed. Nothing lost — the other six were all new and all actionable.

docs/content/patterns/user-input/table-record-management/table-record-entry-grid-compare.demo.tsx
sebald: Already fixed in `74bf68b4c` — same finding as sarahgm's r3733927029. Confirmed against `Panel.tsx:142`, `aria-labelledby={!ariaLabel && hasTitle ? titleId : …}`, so setting `aria-label` suppresses the Title as the accessible name. Removing it lets the visible "The same table, three ways" name the region, which also settles the mismatch you spotted in the direction you preferred.

packages/components/src/Table/Table.stories.tsx
sebald: Agreed, leaving these as they are — and thanks for checking the file's convention before flagging it. Splitting just these two would make them the odd ones out next to `Sorting.test` (8 steps) and `DynamicColumnsAndRows.test` (12).

docs/content/components/collection/table/index.mdx
sebald: Right on the mechanism, but we are keeping the `####` deliberately: that toc is already long, and the call is not to add two more entries to it. Flagging rather than silently ignoring, so say if you disagree.

Two things about the framing that changed the decision:

Promoting both to `###` would flatten them into siblings of their own parent, `### Editing` on line 212 — which is exactly the nesting the restructure exists to express. The suggestion does not account for that parent, so it is not a free move.

And the page does not lose a toc entry: `### Inline editing` became `### Editing`, so the count is unchanged at one. The real cost is narrower than stated — *Data-entry grid* has no entry of its own — and that is what we are accepting.

The chunker point is also weaker than it reads. `chunk()` calls `splitAt(markdown, 2)` and only recurses deeper when a chunk exceeds `MAX_CHARS`, so `## Usage` splits at H3 and `### Editing` becomes one chunk containing both approaches either way. Promoting the two headings to `###` would not reliably split them, because the split at level 4 only happens on overflow.

If discoverability of the grid section turns out to matter more than toc length, the honest fix is to drop the `### Editing` umbrella and promote both children — two entries, no parent — rather than four levels of heading. Happy to do that instead if you would rather have it.

sebald:
## aromko's review — all nine findings addressed

Note for the record: this review was submitted against `392af7e91`, two commits behind the head at the time, so three findings had already been fixed in `74bf68b4c`. The other six were new.

| Finding | Outcome |
| --- | --- |
| 🔴 `keyboardNavigationBehavior` missing from the props table | Fixed, `a91e6782a` — but in the docs pipeline, not the page |
| ⚠️ Heading level skipped | Fixed, `0301e196e` — landed as your exact suggestion |
| ⚠️ Two `####` never reach the toc | **Declined**, deliberately — the only thread left open |
| ⚠️ Line 130 overclaims | Fixed, `0301e196e` — verified rather than softened |
| 💡 Nothing pins the default | Added, `549321de3` |
| 💡 Home/End uncovered | Added, `549321de3` — both sides |
| 💡 `[!code highlight:8]` | Already fixed, `74bf68b4c` |
| 💡 `aria-label` mismatch | Already fixed, `74bf68b4c` |
| 💡 Four act-assert cycles | No change, by your own reasoning |

### The props table was a pipeline bug, not a missing prop

Worth pulling out of the thread, because it was invisible and would have recurred. `keyboardNavigationBehavior` is real and typed — RAC 1.20 inherits it from `react-stately` 3.49.0. But fumadocs-typescript keys its cache on `path:name:content:version`: the component source file only, never the `.d.ts` files props are inherited from. `Table.tsx` last changed the day *before* the react-aria bump in #5699, so its cached entry predated the prop and nothing would ever have invalidated it. The cache lives in `.next/cache/`, which Vercel preserves across deploys, so the wrong table was being served indefinitely.

The cache directory is now salted with a hash of `pnpm-lock.yaml`. Verified against the deployments rather than a local rebuild — pre-fix 0 occurrences of the upstream description, post-fix 1.

Diffing react-stately 3.48→3.49 and react-aria 3.50→3.51, this was the only added prop reaching a Marigold-documented interface, and a cold regenerate of all 141 tables turns up exactly one affected. So Table was the only casualty this time.

### `[!toc]` works — settled empirically

This was disputed earlier in the PR and I want it on the record, since I was the one asserting it. On the rebuilt preview at depth 3 — where the `depth <= 3` filter cannot be the cause — the heading renders `as: "h3", id: "accessibility"`, `"!toc"` appears 0 times in the page, and there is no toc entry. Every section that *is* in the toc appears 4+ times in the HTML because the toc link adds its own pair; this one appears exactly 2.

### DST-1687 open question 1 is closed

The mixed-editing claim on line 130 is now backed by `DataEntryGridWithEditableCell`, which fails under `arrow`. The original probe was inconclusive only because a synthetic click could not reach the trigger inside its `w-0 overflow-hidden` wrapper; inside the story runner it works.

### Still open, deliberately

Home inside a `NumberField` reaches the grid in `tab` mode — an upstream `useSpinButton` issue rather than anything this table controls. Home/End is asserted for the `TextField` only, and the comment says so. Recorded on DST-1687.

Table story suite 42 passed, `typecheck:only` and prettier clean, `@marigold/docs: patch` changeset added for the cache fix.

sebald:
@aromko VRT is clean — [Chromatic build 202](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=202):

```
✔ Build 202 passed!
  No changes were found in this build.
  Tested 751 stories across 97 components; captured 391 snapshots in 1m 8s
```

Ran on `549321d` (the branch head), TurboSnap enabled against 2 parent builds and 92 changed files.

No diffs is the expected result rather than a lucky one: the three stories added in this PR all set `chromatic: { disableSnapshot: true }`, and nothing else here touches rendering — the Table changes are stories and tests, the rest is docs content plus the `build-types.ts` cache fix.

One unrelated thing the run surfaced, flagging it rather than acting on it: TurboSnap logged a config warning — *"You can either set the flags `storybook-base-dir` or `storybook-config-dir` to help TurboSnap find the correct storybook config file"* — next to a note about `untraced`. It did not degrade this build, but in a monorepo an unresolved config dir is the sort of thing that later widens or narrows the traced set without anyone noticing. Out of scope here; worth a look on its own.

docs/content/components/collection/table/index.mdx
sebald: @aromko — following up, since this is the last thing blocking the PR (approved by @sarahgm, 24/24 checks green).

Two shapes on the table. I'm fine with either, I just need a pick:

1. **Keep `####` as it stands** — one toc entry (*Editing*), *Data-entry grid* gets none. That's the cost we'd be accepting, in exchange for not growing an already-long toc.
2. **Drop the `### Editing` umbrella** and promote both children to `###` — two toc entries, both discoverable, no false sibling, `[#inline-editing]` anchor preserved, and the chunker splits them cleanly.

The one thing I'd still push back on is the suggestion exactly as written: promoting both to `###` while keeping `### Editing` above them makes them siblings of their own parent, which is the nesting the restructure exists to express.

Leaving this open rather than resolving it myself — it's your call to make.

aromko (approved):

PR #5711, opened 2026-08-06
fix(DST-1676): derive the contrast matrix from the theme palette

sarahgm:
# Description

`contrast-matrix.demo.tsx` hand-copied all 11 charcoal rungs and the palette hue into the demo, then recomputed the **published** WCAG and APCA numbers from that literal. Nothing linked it to `tokens.css`, so a rebrand would have left the documented contrast numbers reporting the old palette.

The demo now measures the shipped palette. Hidden probes painted with `var(--color-charcoal-*)` are read back through a 1×1 canvas, which is what converts the authored wide-gamut color down to the sRGB the contrast math needs. The swatch grid paints from the same custom properties, so it is correct on the first frame without JavaScript, and only the score badges wait for the measurement. A token that is missing, or that the canvas cannot parse, leaves the probe pixel clear, so the demo blanks the badges and says the palette is unreadable instead of publishing `1:1` for every pairing. The measurement retries once after a paint before giving up, and the swatch grid survives either way because it needs no measurement.

It also fixes the matrix's pinned first column, a defect that predates this change. No header cell painted a background, so the swatches scrolled visibly behind the row labels and, on a dark page background, the column labels sat at 3.4:1 on the page that documents contrast. The corner cell was not pinned sideways with the column it heads, so it scrolled out of the port and clipped mid-word. The header row also carried a `top` offset that pinned nothing, because the sticky containing block is the `overflow-x-auto` wrapper, which has no height to scroll vertically. That dead offset is gone rather than left to read as a working feature. While the header cells were being rewritten they also gained `scope`, and the table an accessible name, so a screen reader announces both rungs before a cell's numbers.

One inherent cost of measuring: the contrast numbers are no longer in the server-rendered HTML. The markdown export is unaffected, since file-based `<ComponentDemo>` nodes were never inlined into it.

Three notes for review, all on points the ticket raised:

- The ticket recommended reading the computed style of a `bg-charcoal-*` **utility** rather than the custom properties, because `styles.css` imports `tokens.css` with `theme(inline)`. The custom-property route is safe here: `tokens.css` opens `@theme static`, which emits every variable regardless of the `inline` import, and this matches how the neighbouring `color-palette.demo.tsx` already paints its swatches. Confirmed resolving in the browser.
- `oklchToRgb` is deleted even though the ticket lists the helpers under "Not included". It only ever served the literal palette and is dead code without it. The WCAG and APCA helpers themselves are untouched, and the APCA figures remain the simplified approximation the demo has always used.
- The ticket asks for the same treatment on `color-palette.demo.tsx` "if it carries its own copy". It does not. That demo already paints from `var(--color-charcoal-${step})` and holds no palette values, so it needed no change. The rung *names* are still authored in both files, so a rung added to the theme is omitted rather than reported, which the demo now notes in a comment.

Visual regression is N/A here: the diff has no `packages/**`, `themes/**` or `*.stories.tsx`, and the workflow's own `paths-ignore` excludes `docs/**`, `**/*.mdx` and `.changeset/**`.

Closes DST-1676

## Test Instructions

1. `pnpm start`, open `/foundations/token-overview` and scroll to **Contrast matrix**. Every cell shows an `Lc <n>` and an `<n>.<n>:1` badge, no blanks.
2. Narrow the window until the matrix scrolls horizontally, then scroll it right. The rung column stays pinned on an opaque background with a visible right edge, `Bg \ Text` stays fully readable, and no cells show through behind the labels.
3. The ticket's regression check. Change one rung in `themes/theme-rui/src/tokens.css` (e.g. `--color-charcoal-500` to `oklch(0.45 0.18 264)`), run `pnpm build:themes`, reload. Row and column 500 must move — 500-on-50 goes `3.5:1` → `7.5:1` — while untouched pairings such as 900-on-50 stay at `16.6:1`. Revert and rebuild to restore. On `main` those numbers do not move at all.

   Judge that by the WCAG ratio, which is exact across engines. The APCA figure moves the same way (`Lc 63` → `Lc 85` in Firefox) but can sit a point off in another engine, because the canvas readback rounds the palette to 8 bits and engines differ on the last unit. Chromium reads that pairing as `Lc 64`. That is inherent to measuring and predates this round of review feedback.
4. Failure path, optional. Rename `--color-charcoal-500` in the theme and rebuild. The matrix still renders, with its swatches intact and every score badge blank, under the note "The charcoal palette could not be read from the theme". No pairing publishes a zero or a `1:1`.

## Breaking Changes

No

## Checklist


docs/content/foundations/token-overview/contrast-matrix.demo.tsx
```diff
@@ -128,76 +103,217 @@ function wcagBadgeColor(score: number) {
...
                 <th
+                  key={s}
+                  scope="col"
                   className="text-secondary text-[11px] font-bold"
                   style={{
                     padding: '6px 4px',
                     minWidth: 65,
                     textAlign: 'center',
-                    borderRight: '1px solid var(--color-charcoal-200)',
                     borderBottom: '1px solid var(--color-charcoal-200)',
```
sebald, line 266: The column headers didn't get the fix the row headers got. `pinnedRail` went on the row header column only, so these still paint no background.

In fumadocs dark mode that leaves `text-secondary` (charcoal-600) on `#121212` — **3.4:1**, under the 4.5:1 AA floor, on the page that documents contrast. Rendering the page with `.dark` on `<html>` shows the white corner cell sitting right next to unstyled dark ones, so it reads as half-finished as well as failing.

Same defect class, same diff, one line.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
```diff
@@ -128,76 +103,217 @@ function wcagBadgeColor(score: number) {
...
                 className="text-secondary text-[11px] font-bold"
                 style={{
                   padding: '6px 4px',
                   minWidth: 65,
                   textAlign: 'center',
-                  borderBottom: '1px solid var(--color-charcoal-200)',
                   position: 'sticky',
                   top: 0,
+                  left: 0,
+                  zIndex: 2,
```
sebald: The vertical pin never engaged, so this `zIndex: 2` is inert.

The sticky containing block for the `<thead>` cells is the `.overflow-x-auto` wrapper. Its `overflow-y` computes to `auto`, but it has no height constraint, so it never scrolls vertically and `top: 0` has nothing to stick to. Measured on the running docs app after scrolling the page past the table:

```
colHeaderTop === cornerTop === tableTop === -200px
```

The header row scrolls away with everything else — on `main` and on this branch.

That matters mostly for the write-ups: the PR description and the changeset both say "the corner cell was pinned vertically but not horizontally, so it scrolled out of the port and clipped mid-word," which describes behaviour that doesn't exist. The horizontal pin and the opaque fill are real fixes and worth keeping.

Either give the wrapper a `max-height` so the header actually pins, or drop the dead `top: 0` / `zIndex` and adjust both write-ups.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
```diff
@@ -128,76 +103,217 @@ function wcagBadgeColor(score: number) {
...
+      borderRadius: 3,
+      lineHeight: 1.3,
+      backgroundColor: color ?? 'transparent',
+    }}
+  >
+    {children ?? '\u00A0'}
+  </span>
+);
+
+const Legend = () => (
```
sebald, line 159: Extracting this into `Legend` helps, but the thresholds and colours are still triplicated: the four tier colours live in `apcaBadgeColor`, in `wcagBadgeColor`, and again here at a different alpha across eight near-identical spans.

One `{ threshold, color, label }` table per scale could drive the badge functions *and* render this legend — roughly 60 lines lighter, and it removes the way a threshold change currently has to be made in two places to stay consistent.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
```diff
@@ -128,76 +103,217 @@ function wcagBadgeColor(score: number) {
...
-              Bg \ Text
-            </th>
-            {steps.map(s => (
+/**
+ * The row-header column is pinned while the matrix scrolls, so it needs an
+ * opaque fill or the cells travel visibly behind the labels. Collapsed borders
+ * belong to the table and stay behind at the unscrolled position, so the pinned
+ * edges are painted as inset shadows instead.
+ */
+const pinnedRail = {
```
sebald: `pinnedRail` is a good move. `padding: '6px 4px', minWidth: 65, textAlign: 'center'` is still repeated verbatim across all three `<th>` variants though — a `headerCell` const right here would finish the pattern.

sebald (commented):
Verified the core of this independently and it holds up.

Measuring against the built `theme-rui/dist/styles.css`, the measured palette is bit-identical to the deleted literal path on 10 of 11 rungs (one channel off by 1 on step 100). WCAG 500-on-50 = 3.50 and 900-on-50 = 16.57 by both routes, identical in Firefox. Both failure paths land on alpha 0 as intended: a missing token computes to `rgba(0, 0, 0, 0)`, and an unparseable value leaves `fillStyle` parked on `transparent`, so the `clearRect` + `transparent` pair is genuinely load-bearing rather than belt-and-braces. And `--color-charcoal-*` *is* emitted on `:root, :host` despite the `theme(inline)` import, so the rebuttal to the ticket's recommendation is correct — the custom-property route is safe here.

Against the ticket: the literal and the hardcoded `54` are gone, `color-palette.demo.tsx` was correctly checked and left alone (it already read the custom properties), the helpers are untouched, and no `54` survives anywhere in `index.mdx`. Deleting `oklchToRgb` despite the ticket's "Not included" is the right call and is flagged in the description.

One note on scope, not a defect: the contrast numbers are no longer in the server-rendered HTML. The generated `.md` is unaffected — I checked, and file-based `<ComponentDemo>` nodes have never survived `remarkResolveComponentDemo` (it only inlines demos with a `name` attribute), so the matrix was never in the markdown export on `main` either. It's an inherent cost of the ticket's preferred option 1, so nothing to fix, but it might be worth a line in the changeset.

Two things I'd change before merge, plus two bits of cleanup — inline.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
sarahgm: Fixed. Every header cell now shares a `headerCell` base that carries the opaque fill, so the column headers stop sitting on the page background. `pinnedRail` is down to just the sticky bits, which the column headers do not need.

Not quite the one line though. Their `borderBottom` stays a real border rather than folding in the rail's inset shadows, because those also paint a right edge, which would have put a vertical divider between every column header that the design does not have.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
sarahgm: Took the second option: `top: 0` and the `zIndex: 2` that served it are gone. A `max-height` would pin the header for real, but it is a layout change the ticket did not ask for, and eleven rows do not need one.

The `zIndex: 1` left on `pinnedRail` is not inert. It keeps the pinned column above the cells that scroll under it. Those cells are unpositioned so paint order covers it anyway, but the pin would break quietly if that ever stopped being true, hence the comment on the line.

Both write-ups corrected. I kept "scrolled out of the port", because that part did happen: the corner cell had no `left` at all on `main`, so it scrolled away sideways with the table. What is gone is the claim that it was pinned vertically.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
sarahgm: Done. One `{ min, rgb, label }` table per scale now drives the badge fills and the legend both, so a threshold moves in one place. The file is 34 lines lighter.

The last tier is `min: -Infinity` rather than a separate fallback, so the lookup is total and needs no non-null assertion.

Alpha is the one thing still split between the two: badges fill at 0.25 because they sit under text, legend swatches at 0.5.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
sarahgm: Added. `headerCell` holds the padding, width and centring, plus the opaque fill from your column-header comment, so the two concerns land in the same const. All three `<th>` variants spread it, and two of them are down to a single line.

sarahgm:
Thanks, that verification is more than I could have asked for.

All four inline points are addressed in 217fae8e4, each answered on its own thread. The scope note from your review body landed too: the changeset now carries a line saying the numbers leave the server-rendered HTML and that the markdown export is unaffected, and the PR description says the same.

sarahgm:
Verified 217fae8e4 on the fresh preview, with `.dark` on `<html>` so the page background is `rgb(18, 18, 18)`:

- All three header kinds compute `rgb(255, 255, 255)`, so the column labels are off the dark background and match the corner cell.
- `top` computes to `auto` on the header row, `left: 0px` / `zIndex: 1` on the two pinned cells. The dead offset is gone and the real pin is intact.
- 121 cells, no blank badges. Legend renders 8 tiers at the same colours and labels as before.

One thing the check turned up, unrelated to your comments. Test instruction 3 cited `Lc 63` for 500-on-50, which is the Firefox figure. Chromium reads `Lc 64`, because the canvas readback rounds the palette to 8 bits and engines differ on the last unit. The WCAG ratios are exact in both. I have reworded the instruction to hang the check on the ratio and to name the engine spread, rather than leave you chasing a one-point gap. The measurement and the APCA helpers are untouched by this commit, so that spread is inherent to the approach and not new.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
```diff
@@ -104,100 +82,198 @@ function getLuminance(r: number, g: number, b: number) {
...
+        frame = requestAnimationFrame(() => attempt(false));
+        return;
+      }
+      setPalette(result ?? 'unreadable');
+    };
+
+    // Deferred to satisfy `react-hooks/set-state-in-effect`, as Token.tsx does.
+    queueMicrotask(() => attempt(true));
+
+    return () => cancelAnimationFrame(frame);
```
sebald: 💡 Minor, and I checked before raising it: this cleanup cannot cancel the microtask, so there is one path where a `requestAnimationFrame` is scheduled *after* teardown and never cancelled.

If the component unmounts in the same task as it mounts, cleanup runs while `frame` is still `0`. The microtask then fires anyway, `probesRef.current` is `null` because React has detached the ref, so `attempt(true)` schedules a real rAF into `frame` — after the cleanup that was supposed to cancel it — and `attempt(false)` calls `setPalette` on a dead component.

I tested the paths that actually matter and **both are fine**, so please do not read this as a bug report:

| Scenario | Result |
|---|---|
| Route change (unmount in a later task) | rAF scheduled, cleanup cancels it. No leak |
| StrictMode dev double-invoke | No rAF at all, one redundant `setPalette`. No leak |
| Unmount in the same task as mount | rAF leaks, `setPalette` on unmounted component |

So the reachable trigger is essentially synchronous unmount in test code, and React 19 no longer warns about the stray `setPalette`. Nothing user-visible. The only reason I mention it is that the cleanup *reads* as a full teardown and is not quite one — a `let cancelled = false` checked at the top of `attempt` would make it match its appearance:

```suggestion
    return () => {
      cancelled = true;
      cancelAnimationFrame(frame);
    };
```

(with `let cancelled = false` beside `let frame = 0`, and an early `if (cancelled) return;` in `attempt`.) Entirely your call whether that is worth the two lines.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
```diff
@@ -1,5 +1,8 @@
 'use client';
 
+import { Fragment, useEffect, useRef, useState } from 'react';
+
+// Authored here, not measured, so a rung added to the theme is silently omitted.
```
sebald, line 5: 💡 Thank you for stating this in a comment rather than leaving it implicit — that is the honest way to ship a partial fix.

Flagging it only so the scope decision is recorded, since it is mine to make as ticket author: the rung *list* is still a hand-authored copy, so the drift class DST-1676 targets survives in narrower form. Add `--color-charcoal-1000` to the theme and the docs silently omit it rather than misreport it.

Omitting is strictly better than the old behaviour of publishing stale numbers, and I am not asking you to fix it here — a wrong number is the dangerous failure, a missing rung is visible to anyone comparing the page to the theme. I will decide whether that warrants a follow-up ticket. Nothing needed from you on this comment.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
```diff
@@ -104,100 +82,198 @@ function getLuminance(r: number, g: number, b: number) {
...
-              Bg \ Text
-            </th>
-            {steps.map(s => (
+  // A failed read only blanks the badges. The swatch grid needs no measurement.
+  const measured = palette === 'unreadable' ? null : palette;
+
+  return (
+    <>
+      {palette === 'unreadable' && (
+        <p className="text-secondary mb-2 text-sm">
```
sebald: 💡 Low priority, rare path: this message appears only after hydration, so a screen reader user who has already read past this point gets no notification that the numbers are missing — and the blank badges render as `&nbsp;`, which announces as nothing rather than as "unavailable".

`role="status"` would announce it politely when it appears:

```suggestion
        <p role="status" className="text-secondary mb-2 text-sm">
```

sebald (approved):
## Review

Approving. I verified the risky parts empirically rather than reading them, and they hold up.

### The `theme(inline)` warning in the ticket does not bite

DST-1676 carried a ⚠️ that reading custom properties might not work because `tokens.css` is imported with `theme(inline)`. Your defense is correct, and I confirmed it in the bundle the docs actually consume: `themes/theme-rui/dist/styles.css` emits `--color-charcoal-500: oklch(0.62 0.007 54)` under the selector `:root, :host`. `@theme static` wins over the `inline` import for variable emission, and because they land on `:root` they resolve regardless of the `[data-theme="rui"]` scoping that `postcss-prefix-selector` applies to utilities. The custom-property route was the right call.

### The measurement mechanism, checked in Firefox

I reproduced the readback in a scratch browser test:

- `getComputedStyle` on a child of a `display: none` container returns `"oklch(0.62 0.007 54)"` — the authored space, unconverted. Canvas readback gives `[138, 133, 130, 255]`. So the canvas genuinely is what converts to sRGB, exactly as your comment says, and `display: none` costs nothing because `background-color`'s resolved value needs no layout.
- The fail-closed path is doing real work. An undefined custom property computes to `"rgba(0, 0, 0, 0)"`, so alpha is 0. Without the `a === 0` guard that would paint pure black and publish `1:1` for all 121 pairings. Parking `fillStyle` on `transparent` before assigning is the detail that makes it safe, and it is the best part of this PR — on a page documenting contrast, blanking is the only acceptable failure.

### Other claims I spot-checked

- `--color-surface` has no appearance override (`var(--color-white)` unconditionally), so the pinned rail is opaque on any page background. The dark-mode 3.4:1 defect is genuinely fixed, not just papered over.
- The `flatten` cross-reference is real: `packages/components/src/contrast.utils.ts:36`.
- The `queueMicrotask` precedent is real: `docs/ui/Token.tsx:18`.
- The markdown-export claim is right. `remarkResolveComponentDemo` early-returns when there is no `name` attribute, so `file=`-based demos were never inlined. Losing the numbers from SSR HTML costs the export nothing.
- Visual regression is correctly N/A: nothing matches `packages/**`, `themes/**`, `.storybook/**`, `*.stories.tsx` or `*.styles.ts`, and the workflow's own `paths-ignore` covers `docs/**`, `**/*.mdx` and `.changeset/**`.

Deleting `oklchToRgb` despite the ticket listing it under "Not included" is the right call — it only ever served the literal and is dead code without it. The WCAG and APCA helpers are untouched, as the ticket required. Folding the four badge-colour functions into the `scales` tier tables that also drive the legend is a genuine improvement over what the ticket asked for; the duplicated thresholds were a second drift surface.

Three non-blocking notes inline. None of them need to hold up the merge; the `steps` one is a scope judgement that is mine to make as ticket author, not yours to fix here.

---
*Reviewed with Claude Code*

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
sarahgm: Taken. `cancelled` is set in the cleanup and checked at the top of `attempt`, so the same-task unmount path now stops before it can schedule anything.

I kept `cancelAnimationFrame(frame)` beside it. Strictly redundant once the flag is there, but cancelling scheduled work still beats letting a frame callback run to a no-op.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
sarahgm: Understood, nothing changed here. The comment on `steps` stays as the record that the rung list is still hand-authored, and the follow-up call is yours.

docs/content/foundations/token-overview/contrast-matrix.demo.tsx
sarahgm: Taken, with one change to the shape.

Your suggestion puts `role="status"` on the paragraph, but that paragraph only renders in the unreadable state, so the region and its text would mount together. `VenuesTable` in the filter example documents that exact failure at `docs/app/(examples)/examples/filter/VenuesTable.tsx:198`, where NVDA and JAWS on Chrome can drop the first announcement. So the region is now a wrapper that mounts empty and takes only the text. The `mb-2` stayed on the inner `<p>`, so the empty wrapper adds no spacing.

I also picked up the second half of your note, the one about the blank badges announcing as nothing. The table sets `aria-describedby` to the message when the read fails, so a reader arriving after the announcement gets the reason together with the table name, without 242 `unavailable` announcements in the cells. That part is beyond what you raised, so say the word and I will drop it.

sebald (approved):
Re-approving after the second round. I checked every reply against the branch head rather than taking them on trust, and all seven land:

- `headerCell` carries the surface fill and all three `<th>` variants spread it, so the column headers are opaque now. Keeping their `borderBottom` a real border instead of folding it into the rail's shadows is the right call — the inset shadow paints a right edge too, and that would have drawn a divider between every column header.
- `pinnedRail` is down to `position`/`left`/`zIndex: 1`; the dead `top: 0` and `zIndex: 2` are gone, and both write-ups match the code. Agreed on keeping "scrolled out of the port" — that part did happen.
- One `scales` table drives the badge fills and the legend, so a threshold moves in one place. `min: -Infinity` as the last tier is a nice touch; it also catches a `NaN` score, which a plain fallback would have handed a colour anyway.
- The `cancelled` flag makes the cleanup match how it reads. Keeping `cancelAnimationFrame` beside it is fine by me.
- The live region mounting empty is a better shape than what I suggested — you're right that my version mounts the region and its text together, which is the `VenuesTable` failure. The `aria-describedby` addition goes past what I raised; keep it, it's the part that helps a reader arriving after the announcement.

Nothing outstanding. The hand-authored rung list stays my call, and I'll decide separately whether it earns a follow-up.

PR #5712, opened 2026-08-06
feat(DST-1692): add consent banner, Impressum and privacy notice to docs

sebald:
# Description

The public docs site had no Impressum (§ 5 DDG) and no privacy notice (Art. 13 GDPR) — both mandatory for a business site — and ran Vercel Web Analytics with no disclosure. This adds all three, plus a consent gate.

Analytics now loads only after explicit consent, and consent can be withdrawn from the footer on any page. The legal basis in the notice moves from Art. 6(1)(f) legitimate interest to Art. 6(1)(a) consent accordingly.

The banner is built from fumadocs primitives (`buttonVariants`, `fd-*` tokens) so it reads as site chrome rather than an embedded demo.

Scope is `docs/` only — no component, system, or theme code is touched, so no changeset is required and there is no component-library UI change.

Closes DST-1692

> [!IMPORTANT]
> **The banner does not appear on this PR's Vercel preview** — that is intended. It runs on the production deployment only. To see it, run locally with `NEXT_PUBLIC_CONSENT_BANNER=1` in `docs/.env.local`.

### Implementation notes

Four constraints shaped this:

1. **Static rendering is preserved.** Consent lives in `localStorage`, not a cookie — reading a cookie in the root layout would opt all 173 prerendered routes into dynamic rendering.
2. **No flash, no hydration mismatch.** A before-paint inline script stamps `data-consent` on the root element and CSS hides the banner for visitors who already answered. This is the same technique fumadocs' own `Banner` uses internally.
3. **Analytics is genuinely gated, not hidden.** A `<script>` executes regardless of CSS, so it can't be rendered-and-hidden. `useSyncExternalStore` with an `unset` server snapshot keeps SSR and the first client render identical; the script mounts only post-hydration when consent is `granted`, with `beforeSend` as a second layer.
4. **Production only.** `CONSENT_BANNER_ENABLED` in `docs/lib/consent.ts` keys off `NEXT_PUBLIC_VERCEL_ENV`. Previews are internal, so they never show the banner and never load analytics — no consent prompts on every preview link, and no preview traffic in the analytics numbers. Locally it is off by default and opted into with `NEXT_PUBLIC_CONSENT_BANNER=1`. When the banner is off, the footer's **Cookie-Einstellungen** link is off too, so there is no dead control.

### Worth knowing for review

The live analytics script was inspected: it reads **no** `localStorage`, cookies, `sessionStorage` or `indexedDB` — only `navigator.userAgent` and `navigator.webdriver`, with the visitor hash derived server-side and dropped after 24h. So § 25 TDDDG is arguably not triggered and a banner is likely **not** legally required. It's added as a deliberate product decision. Expect analytics volume to drop substantially.

Accept and Decline are the same size and position, differing only in colour.

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

Add `NEXT_PUBLIC_CONSENT_BANNER=1` to `docs/.env.local` first — the banner is off locally without it. Then `pnpm start` and:

1. Open any docs page.
2. Banner appears bottom-left. Confirm in DevTools that **no** request to `va.vercel-scripts.com` has been made and `window.va` is undefined.
3. Click **Accept** → banner disappears, `mg-consent=granted` in localStorage, analytics script loads.
4. Reload → no banner, no flash of it during load, analytics loads.
5. Click **Cookie-Einstellungen** in the footer → banner reopens.
6. Click **Decline** → `mg-consent=denied`; reload and confirm analytics does not load.
7. Visit `/impressum` and `/datenschutz`; confirm footer links work from docs pages, the home page and the legal pages themselves.
8. Remove the env var and reload → no banner, no analytics, no **Cookie-Einstellungen** link, but Impressum and Datenschutz links remain.

## Breaking Changes

No

## Checklist

- [ ] Changeset added (`pnpm changeset`) — N/A, `docs/` only

## Before merging

The legal pages carry a `DRAFT` header and need Legal/DPO sign-off. Outstanding:

- Company data re-verified against the current HRB entry
- Dr. Hann Wagner still correct as the § 18(2) MStV contact
- Vercel DPA confirmed (the notice asserts one exists)
- SCC vs. EU-US DPF wording in § 8
- A concrete log retention period, if known
- The Art. 6(1)(a) switch is new text Legal hasn't seen


docs/app/_components/SiteFooter.tsx
```diff
@@ -0,0 +1,34 @@
+import { CONSENT_BANNER_ENABLED } from '@/lib/consent';
+import Link from 'fumadocs-core/link';
+import { ConsentSettingsButton } from './ConsentSettingsButton';
+
+const YEAR = new Date().getFullYear();
+
+export const SiteFooter = () => (
```
aromko, line 8: 🔴 **The footer never reaches the 16 public `/examples/*` routes.**

`SiteFooter` is wired into `(docs)/[...slug]`, `app/page.tsx` and `(legal)/layout.tsx`. `app/(examples)/examples/*` renders `ShellLayout` instead and gets no footer — those pages are publicly reachable, linked from pattern docs and indexed in ⌘K search.

Three consequences:

- § 5 DDG requires the Impressum to be *ständig verfügbar*; it isn't on those routes.
- The notice claims the withdrawal link is *"am Ende jeder Seite"* (`datenschutz/page.tsx:143`) / *"at the bottom of every page"* (`:398`). As written that's false.
- The banner **does** render there (root layout), so a visitor can be asked for consent on a page with no way to withdraw it — Art. 7(3) GDPR.

DST-1692's expected outcome says "reachable from every page". `app/(examples)/_shared/shell-layout.tsx:198` already has a `Sidebar.Footer` to hang them off.

docs/app/(legal)/datenschutz/page.tsx
```diff
@@ -0,0 +1,533 @@
...
+
+export const metadata: Metadata = {
+  title: 'Datenschutzerklärung | Marigold Design System',
+  description:
+    'Informationen zur Verarbeitung personenbezogener Daten auf marigold-ui.io gemäß Art. 13 DSGVO.',
+};
+
+const Page = () => (
+  <>
+    <h1>Datenschutzerklärung</h1>
```
aromko: 🔴 **German passages need `lang="de"` — WCAG 2.1 AA 3.1.2 (Language of Parts).**

The document is `<html lang="en">`, so screen readers will read the authoritative German version with an English voice. Wrapping each version also fixes the two competing `<h1>`s:

```tsx
const Page = () => (
  <>
    <section lang="de">
      <h1>Datenschutzerklärung</h1>
      …
    </section>
    <hr />
    <section lang="en">
      <h1>Privacy Notice</h1>
      …
    </section>
  </>
);
```

(No `suggestion` block on purpose — applying half of it would leave the file unbalanced.)

docs/app/(legal)/impressum/page.tsx
```diff
@@ -0,0 +1,202 @@
...
+ */
+
+export const metadata: Metadata = {
+  title: 'Impressum | Marigold Design System',
+  description: 'Impressum und Anbieterkennzeichnung gemäß § 5 DDG.',
+};
+
+const Page = () => (
+  <>
+    <h1>Impressum</h1>
```
aromko: 🔴 **Same as in `datenschutz/page.tsx` — the German half needs `lang="de"`, the English half `lang="en"`.** WCAG 2.1 AA 3.1.2.

docs/app/(legal)/layout.tsx
```diff
@@ -0,0 +1,17 @@
+import { SiteFooter } from '@/app/_components/SiteFooter';
+import { baseOptions } from '@/lib/layout.shared';
+import { HomeLayout } from 'fumadocs-ui/layouts/home';
+import type { PropsWithChildren } from 'react';
+
+const Layout = ({ children }: PropsWithChildren) => (
+  <HomeLayout {...baseOptions()}>
+    <div className="mx-auto w-full max-w-3xl px-(--page-padding-md) pt-24 pb-0 xl:px-(--page-padding-xl)">
+      <div className="prose prose-sm dark:prose-invert max-w-none">
```
aromko: ⚠️ **`prose-sm dark:prose-invert` are no-op classes.**

`@tailwindcss/typography` isn't a docs dependency and there's no `@plugin` line in `app/global.css`. The `.prose` that applies is fumadocs' own (`fumadocs-ui/dist/style.css:491`), which ships neither variant. So these pages render at default prose size, and dark mode works only because fumadocs' prose is already token-based.

```suggestion
      <div className="prose max-w-none">
```

docs/app/(legal)/impressum/page.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+    <p className="lead">Angaben gemäß § 5 Digitale-Dienste-Gesetz (DDG)</p>
```
aromko: ⚠️ **`className="lead"` is undefined.** No `.lead` rule in `app/global.css`, `@marigold/theme-rui` or fumadocs' stylesheet. Four occurrences across the two legal pages, all inert — these paragraphs get plain body styling.

docs/lib/consent.ts
```diff
@@ -0,0 +1,87 @@
...
+  } catch {
+    // Choice can't be persisted, but it still applies for this page view.
+  }
+
+  reopened = false;
+  document.documentElement.dataset.consent = choice;
+  notify();
+};
+
+export const openConsentSettings = () => {
```
aromko, line 65: ⚠️ **A reopened banner can't be dismissed, and `reopened` is sticky for the whole session.**

There's no close button and no Escape handler, so someone who clicks **Cookie-Einstellungen** out of curiosity is forced to re-answer. And since `reopened` is module state that only `writeConsent` clears, the banner follows them across every client-side navigation until they do. A close button that resets `reopened` covers both.

aromko (commented):
## Automated Code Review

### Overview

| Field | Value |
|-------|-------|
| Files Changed | 14 |
| Additions | +1062 |
| Deletions | -2 |
| CI | All 22 checks pass |
| Visual Regression | Not needed — no `packages/**`, `themes/**`, `.storybook/**`, `*.stories.tsx` or `*.styles.ts` changes |

### Linked ticket

**DST-1692** — ✨ Add GDPR consent banner, Impressum and privacy notice to the docs site (Unplanned, *In Review*)

Expected outcome, verbatim: *"Impressum and Datenschutzerklärung reachable from **every page**."* — see Critical #1.

### Checklist

Docs-only PR: no components, no RAC wrapping, no `useClassNames`, no stories. Those sections are N/A. `any`-free, `import type` used, functional components, hooks unconditional, no `fireEvent`/Jest, changeset correctly omitted. What's left is below.

---

### Issues found

#### Critical (must fix)

**1. `docs/app/_components/SiteFooter.tsx:7` — the footer misses the 16 public `/examples/*` routes**

`SiteFooter` is wired into three places: `(docs)/[...slug]` (all MDX content), `app/page.tsx`, and `(legal)/layout.tsx`. `app/(examples)/examples/*` — 16 publicly reachable pages, linked from pattern docs and indexed in ⌘K search — renders `ShellLayout` and gets no footer.

Three consequences:

- § 5 DDG requires the Impressum to be *ständig verfügbar*. It isn't on those routes.
- The privacy notice asserts the withdrawal link is *"am Ende jeder Seite"* (`datenschutz/page.tsx:143`) / *"at the bottom of every page"* (`:398`). That's factually false as written.
- The banner **does** render there (it's in the root layout), so a visitor can be asked for consent on a page that offers no way to withdraw it. Art. 7(3) GDPR requires withdrawal to be as easy as giving it.

`app/(examples)/_shared/shell-layout.tsx:198` already has a `Sidebar.Footer` — the links can go there without disturbing the product-shell illusion.

**2. `datenschutz/page.tsx:34`, `impressum/page.tsx:23` — German content inside `<html lang="en">` (WCAG 2.1 AA 3.1.2)**

Both pages put several hundred lines of German prose in a document declared `lang="en"`, with no `lang` on the German passages. Screen readers will read the authoritative German version with an English voice. That is a plain Level AA failure on the docs site of an accessibility-first design system. The `Cookie-Einstellungen` footer label has the same problem on every page.

One change fixes it and the double-`<h1>` structure at the same time:

```tsx
const Page = () => (
  <>
    <section lang="de">
      <h1>Datenschutzerklärung</h1>
      …
    </section>
    <hr />
    <section lang="en">
      <h1>Privacy Notice</h1>
      …
    </section>
  </>
);
```

#### Warnings (should fix)

**3. `docs/app/(legal)/layout.tsx:9` — `prose-sm dark:prose-invert` are no-op classes**

`@tailwindcss/typography` is not a docs dependency and there's no `@plugin` line in `app/global.css`. The `.prose` that applies comes from fumadocs (`fumadocs-ui/dist/style.css:491`), which ships neither a `-sm` nor an `-invert` variant. So the legal pages render at default prose size, and dark mode works only incidentally because fumadocs' prose is already token-based. Drop both, or add the plugin if the smaller scale was the intent.

**4. `impressum/page.tsx:25`, `datenschutz/page.tsx:36` — `className="lead"` is undefined**

No `.lead` rule exists in `app/global.css`, `@marigold/theme-rui`, or fumadocs' stylesheet. Four occurrences, all inert — the intro paragraphs get plain body styling.

**5. `docs/lib/consent.ts:66` — a reopened banner can't be dismissed, and `reopened` is sticky for the session**

There's no close button and no Escape handler, so a visitor who clicks **Cookie-Einstellungen** out of curiosity is forced to re-answer. Worse, `reopened` is module-scoped and only `writeConsent` clears it — until they choose, the banner follows them across every client-side navigation for the rest of the session. A close button that resets `reopened` covers both.

**6. `ConsentBanner.tsx:36` — `role="dialog"` with no focus management and no announcement**

Reopening re-renders the banner at the end of `<body>` with focus still on the footer button. Nothing moves, nothing is announced. APG requires focus to move into a dialog when it opens. Either move it (`tabIndex={-1}` + `ref.focus()` when `reopened`) or drop `role="dialog"` for `role="region"` with an `aria-live` region. Related: `aria-labelledby` points at a `<p>`, not a heading.

**7. `SiteFooter.tsx:5` — `new Date().getFullYear()` at module scope in a dual server/client module**

`DocsPageFooter.tsx` is `'use client'` and imports `SiteFooter`, so on every docs page this module is in the client graph. `YEAR` is evaluated at prerender on the server and again at hydration on the client — from 1 January until the next build, that's a text hydration mismatch on the whole docs site. `app/page.tsx` and `(legal)/layout.tsx` render it server-only, so the same file behaves differently per route. Laziest fix: drop the year.

**8. `docs/lib/consent.ts:38` — no tests for the consent state machine**

`docs/vitest.config.ts` already includes `lib/**/*.test.{ts,tsx}` and the node env is enough with a stubbed `localStorage`/`document`. Worth covering: an invalid stored value falls back to `unset`, a throwing `localStorage` doesn't break the page, `writeConsent` clears `reopened`, `subscribeConsent` unregisters both listeners. This is the one file in the PR where a bug means unlawful processing.

**9. `docs/lib/consent.ts:23` — `VERCEL_ENV !== 'preview'` makes the banner impossible to demo on a preview**

Legal/DPO sign-off is on your own pre-merge list, and they can't see the pages as shipped without cloning the repo — setting `NEXT_PUBLIC_CONSENT_BANNER=1` on the preview env still won't enable it. If preview traffic polluting the analytics numbers is the real concern, gate *analytics* on production and let the explicit opt-in enable the *banner* anywhere. That's also one condition less:

```ts
export const CONSENT_BANNER_ENABLED =
  VERCEL_ENV === 'production' || process.env.NEXT_PUBLIC_CONSENT_BANNER === '1';
```

#### Suggestions (nice to have)

**10. `ConsentInit.tsx:16` — the injected `<style>` has no runtime input**

`HIDE_WHEN_ANSWERED` is built from a constant, so it can live in `app/global.css` as-is. That removes a `<style>` element from `<body>` (non-conforming per the HTML spec, harmless in practice) and shrinks the component to just the script. The selector matches nothing when the banner is off.

**11. `datenschutz/page.tsx:32` — 735 lines of legal prose as hand-escaped JSX**

Every `{' '}`, `„…"` and `<br />` in here is friction for the people who will actually be editing this text (Legal/DPO). The repo already has an MDX pipeline under `content/` that would remove all of it and hand back prose styling plus a TOC. Was TSX chosen to keep these pages out of the page tree and search? If so, worth a comment saying that, because the next person will ask.

**12. `datenschutz/page.tsx:57` — plain `<a>` for internal links**

`<a href="/impressum">` triggers a full document reload between the two legal pages; the footer uses fumadocs `Link` for the same targets. Minor: `rel="noopener noreferrer"` on the GitHub anchors has no `target="_blank"`, so `noopener` is inert and `noreferrer` just drops the referrer to your own repo.

**13. `ConsentedAnalytics.tsx:22` — withdrawal after a grant leaves the loaded script in place**

`@vercel/analytics` appends the script and registers `beforeSend` in `useEffect` with no cleanup (`dist/react/index.mjs`), so unmounting removes nothing. `beforeSend` does keep blocking events live, so this is defensible — but *"das Analyse-Skript nicht geladen"* only holds before the first grant. A `location.reload()` when switching granted → denied would make the code match the notice.

**14. `SiteFooter.tsx:24` — "Cookie-Einstellungen" is the only German string in the UI**

Both page comments state the site is English-only. "Cookie settings" would fit the chrome; both notice versions name the link, so they'd need the same edit.

### Recommendation

The engineering is careful and the four constraints in the description are all genuinely met — static rendering intact, no flash, analytics truly gated, `useSyncExternalStore` used correctly with a matching server snapshot.

Two things stand out as blocking in substance: the footer doesn't reach `/examples/*`, which misses the ticket's own acceptance criterion and makes two sentences in the privacy notice untrue; and the German passages need `lang="de"`. #3 and #4 are dead classes that mean the legal pages don't look the way the code says they do.

---

docs/app/_components/ConsentBanner.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const choose = (choice: ConsentChoice) => writeConsent(choice);
+
+  if (state !== 'unset' && !reopened) return null;
+
+  return (
+    <div
+      id={CONSENT_BANNER_ID}
+      data-reopened={reopened || undefined}
+      role="dialog"
```
aromko, line 62: ⚠️ **`role="dialog"` with no focus management and nothing announced.**

Reopening re-renders the banner at the end of `<body>` with focus still on the footer button. APG requires focus to move into a dialog on open. Either move it (`tabIndex={-1}` + `ref.focus()` when `reopened`) or drop `role="dialog"` for `role="region"` plus an `aria-live` region.

Related: `aria-labelledby` points at a `<p>`, not a heading.

docs/app/_components/SiteFooter.tsx
```diff
@@ -0,0 +1,34 @@
+import { CONSENT_BANNER_ENABLED } from '@/lib/consent';
+import Link from 'fumadocs-core/link';
+import { ConsentSettingsButton } from './ConsentSettingsButton';
+
+const YEAR = new Date().getFullYear();
```
aromko: ⚠️ **Hydration mismatch across a year boundary.**

`DocsPageFooter.tsx` is `'use client'` and imports this module, so on every docs page it's in the client graph: `YEAR` is evaluated at prerender on the server and again at hydration on the client. From 1 January until the next build that's a text mismatch site-wide. `app/page.tsx` and `(legal)/layout.tsx` render it server-only, so the same file behaves differently per route.

Laziest fix: drop the year from the copyright line entirely.

docs/lib/consent.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+export const CONSENT_INIT_SCRIPT = `try{document.documentElement.dataset.consent=localStorage.getItem('${CONSENT_KEY}')||'unset'}catch(e){document.documentElement.dataset.consent='unset'}`;
+
+/** Set while the visitor is revisiting their choice via the footer link. */
+let reopened = false;
+
+const notify = () =>
+  window.dispatchEvent(new CustomEvent(CONSENT_CHANGE_EVENT));
+
+export const readConsent = (): ConsentState => {
```
aromko, line 37: ⚠️ **No tests for the consent state machine.**

`docs/vitest.config.ts` already picks up `lib/**/*.test.{ts,tsx}`, and the node env suffices with a stubbed `localStorage`/`document`. Worth covering: an invalid stored value falls back to `unset`; a throwing `localStorage` doesn't break the page; `writeConsent` clears `reopened`; `subscribeConsent` unregisters both listeners.

This is the one file here where a bug means unlawful processing.

docs/lib/consent.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Consent is only asked for on the live site. Preview deployments are internal
+ * and never show the banner; locally it is off unless you are working on it,
+ * which you opt into with `NEXT_PUBLIC_CONSENT_BANNER=1` in `docs/.env.local`.
+ *
+ * Where the banner is off, analytics stays off too — nobody is around to
+ * consent to it.
+ */
+export const CONSENT_BANNER_ENABLED =
+  VERCEL_ENV === 'production' ||
+  (VERCEL_ENV !== 'preview' && process.env.NEXT_PUBLIC_CONSENT_BANNER === '1');
```
aromko: ⚠️ **`VERCEL_ENV !== 'preview'` makes the banner impossible to demo on a preview.**

Legal/DPO sign-off is on your pre-merge list and they can't see the pages as shipped without cloning the repo — setting `NEXT_PUBLIC_CONSENT_BANNER=1` on the preview env won't help either. If preview traffic polluting analytics is the real concern, gate *analytics* on production and let the explicit opt-in enable the *banner* anywhere:

```suggestion
export const CONSENT_BANNER_ENABLED =
  VERCEL_ENV === 'production' || process.env.NEXT_PUBLIC_CONSENT_BANNER === '1';
```

docs/app/_components/ConsentInit.tsx
```diff
@@ -0,0 +1,19 @@
...
+ * logs an error.
+ *
+ * `:not([data-reopened])` limits the rule to the pre-hydration paint, so the
+ * footer link can still reopen the banner while a choice is stored.
+ */
+const HIDE_WHEN_ANSWERED = `html[data-consent="granted"] #${CONSENT_BANNER_ID}:not([data-reopened]),html[data-consent="denied"] #${CONSENT_BANNER_ID}:not([data-reopened]){display:none}`;
+
+export const ConsentInit = () => (
+  <>
+    <style>{HIDE_WHEN_ANSWERED}</style>
```
aromko: 💡 **This `<style>` has no runtime input.** `HIDE_WHEN_ANSWERED` is built from a constant, so it can move to `app/global.css` verbatim — one less `<style>` in `<body>` (non-conforming per spec, harmless in practice) and the component shrinks to just the script. The selector matches nothing when the banner is off.

docs/app/(legal)/datenschutz/page.tsx
```diff
@@ -0,0 +1,533 @@
...
+const LAST_UPDATED_DE = '6. August 2026';
+const LAST_UPDATED_EN = '6 August 2026';
+
+export const metadata: Metadata = {
+  title: 'Datenschutzerklärung | Marigold Design System',
+  description:
+    'Informationen zur Verarbeitung personenbezogener Daten auf marigold-ui.io gemäß Art. 13 DSGVO.',
+};
+
+const Page = () => (
```
aromko: 💡 **735 lines of legal prose as hand-escaped JSX.** Every `{' '}`, `„…"` and `<br />` is friction for the people who'll actually edit this (Legal/DPO), and the repo already has an MDX pipeline under `content/` that removes all of it and hands back prose styling plus a TOC.

Was TSX chosen to keep these out of the page tree and search? If so, worth saying so in the file comment — the next person will ask.

docs/app/(legal)/datenschutz/page.tsx
```diff
@@ -0,0 +1,533 @@
...
+      79098 Freiburg im Breisgau
+      <br />
+      Deutschland
+      <br />
+      Telefon: <a href="tel:+497618878800">+49 761 88788 0</a>
+      <br />
+      E-Mail: <a href="mailto:info@reservix.de">info@reservix.de</a>
+    </p>
+    <p>
+      Weitere Angaben finden Sie in unserem <a href="/impressum">Impressum</a>.
```
aromko: 💡 **Plain `<a>` for an internal link** triggers a full document reload between the two legal pages; the footer uses fumadocs `Link` for the same targets.

Minor: `rel="noopener noreferrer"` on the GitHub anchors has no `target="_blank"`, so `noopener` is inert and `noreferrer` just drops the referrer to your own repo.

docs/app/_components/ConsentedAnalytics.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * nothing; the script is mounted only after hydration, and only if accepted.
+ */
+export const ConsentedAnalytics = () => {
+  const consent = useSyncExternalStore(
+    subscribeConsent,
+    readConsent,
+    getServerConsent
+  );
+
+  if (consent !== 'granted') return null;
```
aromko, line 22: 💡 **Withdrawal after a grant leaves the loaded script in place.** `@vercel/analytics` appends the script and registers `beforeSend` in `useEffect` with no cleanup (`dist/react/index.mjs`), so unmounting removes nothing. `beforeSend` keeps blocking events live, so this is defensible — but *"das Analyse-Skript nicht geladen"* only holds before the first grant. A `location.reload()` on granted → denied would make the code match the notice.

docs/app/_components/SiteFooter.tsx
```diff
@@ -0,0 +1,34 @@
...
+        >
+          Impressum
+        </Link>
+        <Link
+          className="hover:text-fd-foreground transition-colors"
+          href="/datenschutz"
+        >
+          Datenschutz
+        </Link>
+        {CONSENT_BANNER_ENABLED && <ConsentSettingsButton />}
```
aromko, line 25: 💡 **"Cookie-Einstellungen" is the only German string in the site UI**, and both page comments state the site is English-only. "Cookie settings" would fit the chrome — both notice versions name the link, so they'd need the same edit.

docs/app/_components/SiteFooter.tsx
sebald: Fixed in d61ff0b6.

`SiteFooter` now renders from `app/(examples)/layout.tsx` — the group layout rather than `examples/layout.tsx`, so it covers every route in the group including `/examples` itself. It sits outside `Wrapper`, so it keeps the fumadocs chrome instead of inheriting the `rui` theme the demos run under.

Checked against the build output rather than by eye: `examples/teams.html`, `examples/users.html` and `examples/billing.html` each contain one `/impressum` and one `/datenschutz` link.

docs/app/(legal)/datenschutz/page.tsx
sebald: Fixed in d61ff0b6, in the shape you suggested — each version is wrapped in a `lang`-scoped `<section>`.

Built `/datenschutz`: `lang="de"` ×1 (the German section; the only other `lang` on the page is `<html lang="en">`), `<h1>` ×1.

On the competing `<h1>`s: wrapping alone would still leave two peers, so the English half nests one level below the German (`##` → `###`). That also matches what the page says in prose — the German version is the authoritative one, and the translation exists because the site is English-only.

docs/app/(legal)/impressum/page.tsx
sebald: Same fix in d61ff0b6. Built `/impressum`: `lang="de"` ×1, `<h1>` ×1, English half nested a level down.

docs/app/(legal)/layout.tsx
sebald: Confirmed and dropped in d61ff0b6.

No `.prose-sm` and no `.prose-invert` anywhere in `fumadocs-ui/dist/style.css`, and with no typography plugin there is nothing to generate them.

Also dropped `max-w-none` while there: fumadocs' `.prose` already sets `max-width: none` (`style.css:493`). The class list is just `prose` now.

docs/app/(legal)/impressum/page.tsx
sebald: This one doesn't hold — `lead` **is** styled, so I kept it. Flagging rather than resolving, in case you read it differently.

The rule is an attribute selector nested inside `.prose`, which is why grepping for `.lead` misses it:

```css
/* fumadocs-ui/dist/style.css:491 */
.prose {
  :where([class~="lead"]):not(:where([class~="not-prose"],[class~="not-prose"] *)) {
    font-size: 1.25em;
    line-height: 1.6;
    margin-top: 1.2em;
    margin-bottom: 1.2em;
    color: var(--tw-prose-lead);
  }
}
```

`--tw-prose-lead` is defined at line 789 as `var(--color-fd-foreground)`, so it is theme-aware as well — it is not relying on the plugin variants that this PR's other thread correctly identified as no-ops.

Built `/impressum` contains `class="lead"` ×2, rendering at 1.25em.

docs/lib/consent.ts
sebald: Fixed in d61ff0b6.

Added `closeConsentSettings()`, which clears the module flag without touching the stored choice, wired to both a close button and Escape.

The close button only appears while reopened. On a first visit there is no stored choice to fall back to, so a dismiss would have nothing to mean — the banner stays until it is answered.

Two of the new tests pin this: closing a reopened banner falls back to the stored choice, and answering again clears the reopened state.

docs/app/_components/ConsentBanner.tsx
sebald: Fixed in d61ff0b6.

On reopen the banner captures `document.activeElement`, focuses itself, and restores focus to the trigger when it closes. Escape closes it.

`tabIndex={-1}` is applied only while it acts as a dialog, so it never becomes a stray tab stop on a first visit. With `aria-labelledby` already in place, moving focus is what gets it announced.

docs/app/_components/SiteFooter.tsx
sebald: Fixed in d61ff0b6 by dropping the year: `© Reservix GmbH — Marigold Design System`.

Threading a build-time constant through the client boundary was the alternative, but the year carries no legal weight here — copyright arises automatically under Berne, the notice is not a condition of protection, and § 5 DDG does not ask for it either. Removing the input beats keeping two evaluations of it in sync.

docs/lib/consent.ts
sebald: Added in d61ff0b6: `docs/lib/consent.test.ts`, 13 tests in the node env with stubbed `localStorage` / `document` / `window`, as you described.

Covers the three you named — an invalid stored value falls back to `unset`, a throwing `localStorage` does not break the page (thrown on property *access*, not just on call), and `writeConsent` persists, stamps the dataset and notifies — plus the reopen/close transitions, subscribe/unsubscribe, that a failed write still applies the choice for the page view, and that the server snapshot stays `unset`.

`pnpm --filter @marigold/docs test`: 11 files, 100 tests passing.

docs/lib/consent.ts
sebald: Fixed in d61ff0b6.

The condition is now `production || NEXT_PUBLIC_CONSENT_BANNER === '1'` — the environment check no longer excludes preview, so setting the variable on a preview environment shows the banner exactly as it ships, which is what Legal and the DPO need to sign it off.

A preview that does not set it still shows nothing, so ordinary preview traffic is unaffected.

docs/app/_components/ConsentInit.tsx
sebald: Moved in d61ff0b6 — the rule lives in `app/global.css` and `ConsentInit` is just the script now.

One wrinkle worth recording: I kept it **unlayered** rather than putting it in `@layer base`. The inline `<style>` it replaces was unlayered too, and layered rules rank below utilities — inside `@layer base` any `display` utility on the banner would beat `display: none`. Nothing sets one today, but the precedence should not depend on that staying true.

docs/app/(legal)/datenschutz/page.tsx
sebald: Done in d61ff0b6. To answer the question directly: yes, keeping it out of `content/` was the reason. The docs collection globs all of `content`, so these pages would have joined the page tree, the sidebar, docs search and the MCP index.

So they went to `docs/legal/` with their own `legal` collection in `source.config.ts`, rendered by the existing `(legal)` routes. The prose is plain markdown now — `##`, `**…**`, `[…](…)`, `-` lists — with the `{' '}` and the hand-escaping gone.

Because this is text Legal and the DPO have to sign off on, I did not retype it. It was converted by script and diffed token-by-token against the TSX (visible text plus link targets, so a dropped `href` would show up). `impressum.mdx` is byte-identical to a pristine re-run of that conversion; `datenschutz.mdx` differs only by the two sentence edits from the other threads on this PR.

That process caught something a manual pass would probably have shipped. JSX drops a whitespace run that touches a tag boundary and contains a newline, so:

```jsx
eine zufällig erzeugte, nicht personenbezogene Kennung (
<code>anonymousId</code>), die lokal …
```

renders as `Kennung (anonymousId)`. Collapsing that run to a space instead — the obvious thing to do — silently gives `Kennung ( anonymousId)`. Built output confirms `(<code>anonymousId</code>)`.

docs/app/(legal)/datenschutz/page.tsx
sebald: Both handled in d61ff0b6, as part of the MDX conversion.

Internal links are markdown links now, so they resolve through the MDX pipeline's `Link` component and no longer force a full document reload between the two legal pages.

The `rel="noopener noreferrer"` is gone. It is actually how it got looked at: the converter refuses to drop attributes silently and errored on exactly those two anchors. Your reading was right — with no `target="_blank"` the `noopener` was inert, and `noreferrer` only suppressed the referrer on a same-tab navigation.

docs/app/_components/ConsentedAnalytics.tsx
sebald: Agreed on both counts — the behaviour is defensible, the wording was not. Fixed the wording in d61ff0b6 rather than forcing a reload on withdrawal.

Both versions of § 4 now say it plainly:

> Nach einem Widerruf werden keine weiteren Daten übertragen; das bereits geladene Analyse-Skript verbleibt bis zum nächsten Seitenaufruf im Speicher Ihres Browsers, sendet jedoch keine Ereignisse mehr.

> After a withdrawal no further data is transmitted; the analytics script that has already been loaded stays in your browser's memory until the next page load, but no longer sends any events.

The comment in `ConsentedAnalytics` now says the same thing from the code side: `beforeSend` reads the live choice on every event, so it is what actually enforces a withdrawal — the unmount does not.

docs/app/_components/SiteFooter.tsx
sebald: Renamed in d61ff0b6 to "Cookie settings", in `ConsentSettingsButton` and in both notice versions that name the link (§ 4, Widerruf / Withdrawal).

The German sentence around it stays German but quotes the English label, since it is naming a control the reader actually sees on the page.

sebald:
Pushed `d61ff0b6` — all 15 review threads addressed. 14 resolved; one left open where I think the finding doesn't hold (see the `className="lead"` thread — fumadocs styles it via an attribute selector, so a `.lead` grep misses it).

One thing not tied to any thread: **added the missing changeset** (`@marigold/docs: patch`). Same gap you flagged on #5708 — `private: true` only stops npm publishing, and `@marigold/docs` is in the `fixed` group with `@marigold/system` and `@marigold/components`.

Two judgment calls worth a second opinion:

- **Dropped the copyright year entirely** instead of threading a build-time constant through the client boundary. Copyright is automatic under Berne and § 5 DDG doesn't ask for a year, so removing the input beat keeping two evaluations of it in sync.
- **The English translation nests one heading level below the German** (`##` → `###`). Wrapping each version in a `lang`-scoped `<section>` fixes the screen-reader voice, but on its own it would still leave two peer `<h1>`s; nesting is what gets each page down to one, and it matches the pages' own statement that the German version prevails.

Verification: docs suite 11 files / 100 tests passing, `types:check` clean, `eslint` clean on changed files, and `next build` output checked for `lang="de"` ×1 and `<h1>` ×1 on both legal pages plus the footer links on the `/examples/*` routes.

Both legal pages remain **drafts** pending Legal / DPO review and release.

sebald:
Ready for another look, @aromko — all 15 findings from your review are addressed and pushed as `d61ff0b6`.

**14 threads answered and resolved.** The 3 🔴 blockers, the 7 ⚠️ should-fixes and the 5 💡 suggestions all landed, each with the verification in the thread rather than just a "done".

**1 thread left open on purpose** — `className="lead"`. It isn't inert, so I kept it. fumadocs styles it through an attribute selector nested inside `.prose`, which is why a `.lead` grep misses it:

```css
/* fumadocs-ui/dist/style.css:491 */
.prose {
  :where([class~="lead"]):not(:where([class~="not-prose"],[class~="not-prose"] *)) {
    font-size: 1.25em;
    color: var(--tw-prose-lead);
  }
}
```

`--tw-prose-lead` resolves to `--color-fd-foreground` (line 789), so it is theme-aware as well. Built `/impressum` contains `class="lead"` ×2. Resolve it if you agree; push back if you read it differently.

**One thing not on your list:** I added the missing changeset (`@marigold/docs: patch`) — same gap you flagged on #5708, and `@marigold/docs` is in the `fixed` group with `@marigold/system` and `@marigold/components`.

**Two judgment calls worth a second opinion**, both already noted above:

- The copyright year is gone rather than threaded through the client boundary. Copyright arises automatically under Berne, the notice isn't a condition of protection, and § 5 DDG doesn't ask for a year — so removing the input beat keeping two evaluations of it in sync across a January prerender/hydration boundary.
- The English translation nests one heading level below the German, so each legal page keeps a single `<h1>`. That also matches what the pages say in prose: the German version is the authoritative one.

22/22 checks green and mergeable. The only thing outstanding is an approval pass.

aromko (approved):

PR #5714, opened 2026-08-06
feat(DST-944): add destructive actions pattern with undo support

aromko:
# Description

Adds a `/patterns/feedback/destructive-actions` page that decides when a destructive action needs a confirmation dialog and when it can offer an undo toast, with a worked example of each.

The decision rule is **perceptibility and deferability, not reversibility**. Undo is built by deferring the request rather than reversing it, which makes almost anything reversible for a few seconds, so reversibility cannot choose between the two mechanisms. File deletion lands on a confirmation because near-identical filenames (per DSTSUP-259) make it impossible to tell which item vanished inside the undo window.

Supporting library changes:

- **`useToast` gains `addUndoToast({ title, onUndo, onCommit })`.** It owns the timing the pattern would otherwise leave to each caller: the commit rides the toast's own `onClose`, an undo cancels it, the toast carries no close button, and `timeout: 0` falls back to the default window instead of meaning "never commit". The undo button is named after the title, so stacked undo toasts stop announcing a bare "Undo".
- **`<FileField>` accepts `onBeforeRemove`.** Return `false`, or a promise resolving to `false`, to keep the file, which is what puts a confirmation in front of the built-in remove buttons. Each remove button is also named after its own file now instead of all announcing "Remove file", with `removeLabel` on `FileField.Item` for hand-composed lists.
- **`useToast` accepts `onClose`**, fired when the toast closes for any reason: its timeout ran out, it was dismissed, it was closed through `removeToast`, or the queue was cleared. This is what `addUndoToast` commits from. React Aria pauses the toast timer while the region is hovered or focused (`useToastRegion.mjs:49`), so a separate `setTimeout` commits the deletion while the Undo button is still on screen.
- **`ToastOptions.closeButton`**, ignored by a toast that never auto-dismisses (`warning`, `error`, `timeout: 0`), since dropping the button there would leave no way out of it. The rule lives in `Toast.tsx`, the one place every toast renders, so every path into the queue inherits it.
- **`clearToasts()` runs the close handlers of the toasts it removes.** React Stately's `clear()` empties the queue without closing its toasts, which would have left one close path silently skipping `onClose`. Invisible to existing consumers, since `onClose` is new here and no released version can have a handler to drop.
- **`Dialog.Trigger` no longer blocks <kbd>Escape</kbd> by default.** An unset `keyboardDismissable` was inverted into "keyboard dismiss disabled", so every dialog opened from a trigger trapped the keyboard unless the prop was passed explicitly. `Tray`, `TrayModal` and `useNonModal` all already defaulted to dismissable, and the Dialog page already documented Escape as a way out, so this was the only overlay contradicting both. Pass `keyboardDismissable={false}` to opt out.
- **Closing a confirmation without pressing a button now resolves `useConfirmation` as `cancelled`.** It previously never settled, so `await confirm(...)` stalled silently and its continuation never ran. Fails safe today, but the new page tells teams to `await` this hook.
- **`ConfirmationDialog` focuses cancel by default for the `destructive` variant**, so a reflexive <kbd>Enter</kbd> takes the safe path. That variant is now part of the prop's type instead of a bare `string`.
- **`ConfirmationProvider` forwards `autoFocusButton`**, which it accepted but dropped, and its confirm button falls back to a localized `confirm` message instead of a hardcoded "Confirm" sitting next to a translated "Abbrechen".

The [Dialog](https://www.marigold-ui.io/components/overlay/dialog) page picks up the parts of this that belong to the component rather than the pattern: what `useConfirmation` resolves to on each way out, that `destructive` now moves focus as well as styling, and how to opt out of keyboard dismissal. The [Toast](https://www.marigold-ui.io/components/overlay/toast) page documents `onClose` and `addUndoToast`, with generated props tables for both option types, and the [FileField](https://www.marigold-ui.io/components/form/file-field) page documents `onBeforeRemove` against the same demo the pattern page uses.

Both new APIs were the follow-up tickets this branch originally wrote for itself. Building them here is a deliberate reversal: the pattern page now documents an API rather than a recipe with three ways to lose data quietly, and the `addUndoToast` shape in particular is worth a look on its own, since it is a published export designed against one demo.

Closes DST-944
Closes DST-1515

### Behaviour changes worth a reviewer's eye

Not breaking (no removed exports, no renamed props, `minor` changeset), but visible:

1. `ConfirmationDialog`'s action buttons now call their handler **before** `close()`. This is what lets the provider settle `confirmed` before the close settles `cancelled`. Anyone relying on the previous ordering sees a change.
2. Destructive confirmations focus Cancel on open where previously nothing was focused by default.
3. <kbd>Escape</kbd> now closes a dialog opened from `<Dialog.Trigger>` (including `<ConfirmationDialog.Trigger>`) where it previously did nothing. Anyone who relied on the old default has to pass `keyboardDismissable={false}`.

### Out of scope

`keyboardDismissable` is mislabeled twice on its way down. `DialogTrigger` negates it into the context, `Dialog` passes that already-negated flag into a prop *named* `keyboardDismissable`, and `Underlay` assigns it straight to `isKeyboardDismissDisabled`. Net behaviour is correct, but two hops carry a boolean whose name means the opposite of its value, and `Loader.tsx:42` is the evidence it already misleads: it passes `keyboardDismissable` to `Underlay` and thereby disables keyboard dismiss. Pre-existing, and this PR is what makes the default load-bearing, so it belongs on the same follow-up ticket.

### Drive-by

`browserScanEntries` no longer crawls `packages/cli`, matching the `exclude` beside it. Main's CLI work added Node tests there, and pulling them into the browser dep pass reaches the optional `fsevents` on macOS, whose `.node` binary rolldown cannot load, so every local vitest run died with `UNLOADABLE_DEPENDENCY` before collecting a test. Linux CI never saw it, because fsevents does not install there.

`docs/scripts/build-types.ts` resolves a referenced type to `.ts` when there is no `.tsx`, which is what lets the Toast page generate a props table for `ToastOptions`.

## Screenshots / Preview

- Docs: [`/patterns/feedback/destructive-actions`](https://marigold-docs-git-feat-dst-944-destructive-actions-marigold.vercel.app/patterns/feedback/destructive-actions) and [`/components/overlay/toast`](https://marigold-docs-git-feat-dst-944-destructive-actions-marigold.vercel.app/components/overlay/toast)
- Storybook: [ConfirmationDialog](https://marigold-storybook-qmjt2keyi-marigold.vercel.app/?path=/docs/components-confirmationdialog--docs)

## Test Instructions

1. `pnpm build && pnpm start`, then open [`/patterns/feedback/destructive-actions`](http://localhost:3000/patterns/feedback/destructive-actions).
2. **Confirming demo** — upload two files, then press one row's remove button. The dialog names the file and Cancel has focus. Press <kbd>Escape</kbd>, then try again and press Cancel. Neither removes anything. Note each remove button is named after its own file.
3. **Offering undo demo** — delete a row. It disappears at once and a toast offers Undo, with no close button. Press Undo: the row returns *in its original position*. Delete another and let the toast expire: that row is deleted.
4. Delete two rows in quick succession so both toasts stack, and confirm each Undo button exposes a distinct accessible name (`Undo: “<list>” deleted`), and that undoing one does not affect the other.
5. Open [`/components/overlay/dialog`](http://localhost:3000/components/overlay/dialog) and [`/components/overlay/toast`](http://localhost:3000/components/overlay/toast) and check the Confirmation and `onClose` sections against the behaviour you just exercised.
6. `pnpm sb`, open any `Dialog` story with a trigger and press <kbd>Escape</kbd>: it closes without the story passing `keyboardDismissable`.
7. `pnpm test:unit packages/components/src/Toast packages/components/src/Dialog` and `pnpm test:sb packages/components/src/Dialog`.

## Breaking Changes

No — see "Behaviour changes worth a reviewer's eye" above for the three visible changes.

## Checklist


packages/components/src/Dialog/ConfirmationDialog.stories.tsx
```diff
@@ -124,9 +126,53 @@ Basic.test(
   }
 );
 
-// useConfirmation — the imperative hook opens a dialog built from its config and
-// resolves when the user confirms. A local `ConfirmationProvider` keeps the demo
-// self-contained instead of relying on the global one from the decorator.
+Basic.test(
+  'Focuses the cancel button on open when the variant is destructive',
```
sebald, line 145: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files (`packages/components/src/**`, `*.stories.tsx`) but the `Visual-Regression-Tests` workflow has never run for this branch — `gh run list --workflow=visual-regression-tests.yml --branch=feat/DST-944-destructive-actions` returns `[]`. The checklist item is unticked.

One nuance worth knowing before you decide how much this matters: `Basic` sets `parameters: { chromatic: { disableSnapshot: true } }`, and this new test does **not** re-enable it. So the one genuinely visual change in the diff — Cancel now carrying the focus ring on open for `destructive` — would not be snapshotted even with VRT green. If you want that covered, this test needs `parameters: { chromatic: { disableSnapshot: false } }` the way the first test in the file does.

docs/content/patterns/feedback/destructive-actions/index.mdx
```diff
@@ -0,0 +1,147 @@
...
+    <DontDescription>
+      Don't ask "Are you sure?" with OK and Cancel. It carries no information
+      about what is about to happen.
+    </DontDescription>
+  </Dont>
+</GuidelineTiles>
+
+## Offering undo
+
+An undo toast reports the action as done, hides its effect immediately, and sends the real request only when the toast closes. Until then the work is pending and a single press takes it back.
```
sebald, line 90: 🔴 **An undo toast built with `variant="warning"` or `"error"` never commits — and neither does `timeout: 0`**

This is the one finding on the page that can cost real data if someone follows it literally.

`resolveTimeout` in `ToastQueue.ts` returns `undefined` for `warning` and `error` (and for `timeout <= 0`), which means no `Timer` is created and the toast persists until manually dismissed. Combined with commit-on-`onClose`, the deletion then hangs indefinitely: the interface has already reported the row gone, and the request never goes out unless the user happens to dismiss the toast.

The trap is easy to walk into from two directions:

- reaching for `variant="warning"` because a pending deletion feels like it warrants weight, and
- reaching for `timeout: 0` thinking it gives the user *unlimited* time to undo, when it actually means "never commit".

The page never mentions `timeout` or variant choice for undo toasts at all. Worth an explicit **Don't**, since the mechanism this page introduces is what creates the hazard.

While you're here: the effective window is `MINIMUM_TIMEOUT_MS = 5000` with `Math.max`, so 5s is both the default *and* the floor — you cannot make an undo window shorter, and there is no ceiling. Showing `timeout` in the example would let teams size the window deliberately instead of inheriting it.

docs/content/patterns/feedback/destructive-actions/index.mdx
```diff
@@ -0,0 +1,147 @@
...
+</Callout>
+
+What the dialog owes the user:
+
+- **Name the object.** "Delete file?" with the filename in the body, not "Are you sure?".
+- **State the consequence plainly.** Say that it cannot be undone, and say what else goes with it.
+- **Repeat the verb in the confirm button.** "Delete file" rather than "OK", so the button reads as the decision it is.
+- **Leave the safe path as the default.** The `destructive` variant focuses Cancel on open, so a reflexive <kbd>Enter</kbd> cancels. Pass `autoFocusButton` to override.
+
+Dismissing the dialog with <kbd>Escape</kbd> resolves as `cancelled`. A press outside does not dismiss a confirmation at all, which is deliberate: a stray click next to the dialog should not decide anything. Never treat a dismissal as consent, and do not block <kbd>Escape</kbd> to prevent mis-clicks: an `alertdialog` the keyboard cannot escape is an accessibility defect, not a safety feature.
```
sebald, line 71: ⚠️ **`<ConfirmationDialog.Trigger>` blocks <kbd>Escape</kbd> by default — the exact thing this paragraph tells teams not to do**

The rule here is right: *"an `alertdialog` the keyboard cannot escape is an accessibility defect, not a safety feature."* But Marigold's own declarative API for confirmations violates it.

`DialogTrigger.tsx:35` does `isKeyboardDismissDisabled: !keyboardDismissable`, so an unset `keyboardDismissable` becomes `true`. That flows through `Dialog.tsx:155` → `Underlay.tsx:36` → RAC. Net effect:

- via `useConfirmation` (controlled, `DialogContext` defaults to `{}`) → <kbd>Escape</kbd> **works**
- via `<ConfirmationDialog.Trigger>` → <kbd>Escape</kbd> is **blocked**

Two independent confirmations that this isn't a misreading: `ConfirmationDialog.stories.tsx`'s `Basic` sets no `keyboardDismissable` and has no Escape test, while `Dialog.stories.tsx` has to set `keyboardDismissable: true` explicitly in `args` for `Dialog.test.tsx:39` ("supports closing the dialog with escape key") to pass.

Pre-existing and outside this diff — but this page is what turns it into a documented contradiction. Since DST-1515 already spawned one follow-up ticket for the `FileField` escape hatch, this belongs on it. In the meantime consider scoping the sentence to the hook, so the page isn't asserting something the trigger-based path doesn't honour.

docs/content/patterns/feedback/destructive-actions/index.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+// Do: one clock, owned by the toast.
+addToast({ title: 'Deleted', action: <UndoButton />, onClose: () => commit() });
+```
+
+### What undo cannot cover
+
+Deferring the request buys the undo window, and it costs these:
+
+- **Leaving the page abandons the commit.** Close the tab during the window and the request never goes out, while the interface already reported the deletion. Keep windows short, and prefer a confirmation wherever the user is likely to navigate away, such as inside a form.
```
sebald: ⚠️ **No undo-window duration, and nothing on WCAG 2.2.1**

"Keep windows short" is the only guidance the page gives on duration, and it never lands a number. In practice the window *is* 5s: `MINIMUM_TIMEOUT_MS = 5000` is both the default and a `Math.max` floor. Common guidance (Material, NN/g) is 5–10s, so saying so would cost one clause.

The larger gap: an auto-committing timer is a time limit on a user action, which puts this pattern squarely inside **WCAG 2.2.1 (Timing Adjustable)**. Five seconds is not a usable window for a screen-reader or switch-access user — they will not hear the toast, locate the Undo button, and press it in time. That has a real consequence for the decision table above: the accessibility answer is that confirmation is correct in *more* cases than four-times-yes implies, and undo is a mechanism with an access floor rather than a neutral alternative.

Right now the page's a11y content is the `FileField` naming limitation and "don't block Escape". Both good, but neither touches the timing question, which is the one this pattern actually creates.

docs/content/patterns/feedback/destructive-actions/index.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+<ComponentDemo name="undo-delete" />
+
+Three pieces make this work:
+
+1. **Hide the effect straight away** so the interface responds at once. Keep the committed data separate from what is hidden, so restoring puts the row back where it was.
+2. **Commit from the toast's `onClose`.** This is the part that is easy to get wrong, see below.
+3. **Guard the commit.** Pressing Undo has to both restore the item and prevent the commit, because closing the toast runs `onClose` either way.
+
+<Callout title="Two buttons, opposite outcomes">
```
sebald: ⚠️ **This callout names a design smell and then shrugs — it's the weakest part of the page**

The observation is sharp and correct: Close reads as "get this out of my way" while it actually means "go through with the deletion". Then the advice is *"say what happened in the title"*, which doesn't resolve the mismatch — it just hopes the user read the title before reaching for a button that means the opposite of what it looks like.

Two fixes are available and both are real designs:

1. **Hide the close button on undo toasts.** The toast already self-dismisses, and the Undo button is the way out. Close earns nothing here except the ambiguity.
2. **Commit only on timeout, so Close cancels the commit.** Then the button means what it looks like. This costs the `onClose` symmetry the section above argues for, so it's the bigger change — but it's the one that makes the affordance honest.

As written the page documents a footgun it is in a position to design away. Given that the same mechanism also produces the `clearToasts` behaviour below, I'd rather see this decided than described.

packages/components/src/Dialog/useConfirmation.tsx
```diff
@@ -52,18 +52,21 @@ export const ConfirmationProvider = ({ children }: PropsWithChildren) => {
...
+        onOpenChange={isOpen => {
+          setOpen(isOpen);
+          if (!isOpen) {
+            confirmation?.resolve('cancelled');
+          }
+        }}
         variant={confirmation?.variant}
         size="xsmall"
         title={confirmation?.title || ''}
         confirmationLabel={confirmation?.confirmationLabel || 'Confirm'}
```
sebald: ⚠️ **`confirmationLabel` falls back to hardcoded English while `cancelLabel` is localized**

```tsx
confirmationLabel={confirmation?.confirmationLabel || 'Confirm'}
cancelLabel={confirmation?.cancelLabel}   // → stringFormatter.format('cancel') → "Abbrechen"
```

There is no `confirm` key in `packages/components/src/intl/messages.ts` — only `cancel` (`'Abbrechen'` / `'Cancel'`) and `removeFile`. So a German UI calling `confirm({ title })` without an explicit `confirmationLabel` renders **"Confirm"** sitting next to **"Abbrechen"**.

The new pattern page's "repeat the verb in the confirm button" rule means well-behaved callers dodge this, but the default shouldn't be the broken one — and this PR is what puts the hook front and centre. Adding a `confirm` key and formatting it the way `cancelLabel` already does would close it.

packages/components/src/Dialog/ConfirmationDialog.tsx
```diff
@@ -32,6 +37,9 @@ export interface ConfirmationDialogProps extends Pick<
   onCancel?: () => void;
   /**
    * Button to focus by default when the dialog opens.
+   *
+   * Defaults to `cancel` for the `destructive` variant so that pressing
+   * <kbd>Enter</kbd> out of habit takes the safe path.
    */
   autoFocusButton?: 'cancel' | 'action';
```
sebald, line 44: ⚠️ **The `destructive` autofocus default can't be overridden to *no* autofocus**

`autoFocusButton = variant === 'destructive' ? 'cancel' : undefined` is a default parameter, so it only fires when the value is `undefined` — which is also the only value that means "leave focus alone". A consumer can override to `'action'`, but there is no way back to no autofocus at all.

This doc comment and the pattern page both say "Pass `autoFocusButton` to override", which overstates what's reachable. Either widen the type to `'cancel' | 'action' | 'none'` and handle `'none'`, or reword to something like:

```suggestion
   * Button to focus by default when the dialog opens.
   *
   * Defaults to `cancel` for the `destructive` variant so that pressing
   * <kbd>Enter</kbd> out of habit takes the safe path. Pass `'action'` to
   * focus the confirm button instead.
   */
  autoFocusButton?: 'cancel' | 'action';
```

packages/components/src/Dialog/ConfirmationDialog.tsx
```diff
@@ -60,16 +68,18 @@ export const ConfirmationDialog = ({
         <>
           <Dialog.Title>{title}</Dialog.Title>
           <Dialog.Content>{children}</Dialog.Content>
+          {/* Handlers run before `close()`, so an owner watching `onOpenChange`
+              sees the decision before it sees the close. */}
           <Dialog.Actions>
             <Button
-              onPress={() => chain(close(), onCancel?.())}
+              onPress={chain(onCancel, close)}
```
sebald: 💡 `onCancel` and `onConfirm` now receive a `PressEvent`

`chain` forwards its arguments to every callback, so both handlers are now called with the press event. The previous `() => chain(close(), onCancel?.())` called them bare.

The declared types (`() => void`) mean TypeScript won't flag it, but a consumer whose handler happens to accept an optional first parameter — `(opts?: SomeConfig) => void`, a curried helper, anything with a default argument — now silently gets a `PressEvent` in that slot. `close` picks it up too, though RAC ignores it.

Harmless in the common case, but it belongs in the changeset's "behaviour changes worth a reviewer's eye" list, which currently mentions only the ordering change.

packages/components/src/Dialog/useConfirmation.tsx
```diff
@@ -52,18 +52,21 @@ export const ConfirmationProvider = ({ children }: PropsWithChildren) => {
       {children}
       <ConfirmationDialog
         open={open}
-        onOpenChange={setOpen}
+        onOpenChange={isOpen => {
```
sebald, line 58: 💡 `confirm` isn't memoized, so every consumer re-renders whenever a confirmation opens or closes

`confirm` is recreated on each provider render and passed straight in as the context value, so every component calling `useConfirmation()` re-renders on each `setOpen` / `setConfirmation`. Pre-existing, but this hunk is already in the file.

`useCallback` with `[open]` (it closes over `open` for the already-open guard) would fix it. `useMemo` on the context value if you'd rather be explicit about what's being stabilized.

packages/components/src/Toast/ToastQueue.tsx
```diff
@@ -81,20 +90,22 @@ const resolveTimeout = (
...
+      }),
+    []
+  );
 
-  const clearToasts = useCallback(() => getToastQueue().clear(), []);
+  const clearToasts = useCallback(() => {
+    const queue = getToastQueue();
+    // `clear()` skips close handlers, so run them first. `visibleToasts` is the
+    // whole queue only while `maxVisibleToasts` stays at its `Infinity` default.
+    queue.visibleToasts.forEach(toast => toast.onClose?.());
```
sebald, line 203: 💡 Don't depend on a default for the invariant this line rests on

The comment is honest about the coupling, which is good. The cheap improvement is to stop *depending* on the default: pass `maxVisibleToasts: Infinity` explicitly in `getToastQueue()`.

Worth being precise about why this is more than pedantry — react-stately has two different defaults here:

- `new ToastQueue(options)` → `options?.maxVisibleToasts ?? Infinity`
- the `useToastState` hook → `maxVisibleToasts = 1`

Marigold happens to take the first path, so `visibleToasts === queue` holds. But it holds by accident, and the accident is one line away in a file nobody reviewing a future `maxVisibleToasts` change would think to read. Setting it explicitly turns the assumption into something a future editor has to walk past to break — and it costs nothing.

(Confirmed the rest of this is sound: `clear()` really does skip `onClose`, and the timers it orphans fire `close(toastKey)` against an empty queue → index `-1` → no double invocation.)

docs/content/patterns/feedback/destructive-actions/index.mdx
```diff
@@ -0,0 +1,147 @@
...
+  Reservix products. The undo half is new and has no product implementation yet.
+  If your team is about to build one, [get in
+  touch](/getting-started/get-in-touch) so we can observe it.
+</Callout>
+
+## Key principles
+
+- **Recoverable.** Every destructive action offers either a way to stop or a way to take it back.
+- **Specific.** The recovery path names the thing being destroyed, never "this item".
+- **Honest.** Never promise a recovery the system cannot deliver.
```
sebald, line 22: 💡 This principle and "reports the action as done" are in tension, and the page reconciles neither

*"Honest. Never promise a recovery the system cannot deliver."* Then line 87: an undo toast *"reports the action as done"* while the work is still pending.

That's the mirror image of the dishonesty this principle warns about — claiming a destruction that hasn't happened yet. It's defensible: the deferral is invisible, and the reported outcome is the one that arrives. But the page asserts both and leaves the reader to notice.

One sentence at line 87 would do it — something to the effect that reporting the deletion up front is honest *because* the commit is guaranteed by the toast's own lifecycle, which is exactly why the `setTimeout` variant below is not just fragile but dishonest.

docs/content/patterns/feedback/destructive-actions/index.mdx
```diff
@@ -0,0 +1,147 @@
...
+addToast({ title: 'Deleted', action: <UndoButton />, onClose: () => commit() });
+```
+
+### What undo cannot cover
+
+Deferring the request buys the undo window, and it costs these:
+
+- **Leaving the page abandons the commit.** Close the tab during the window and the request never goes out, while the interface already reported the deletion. Keep windows short, and prefer a confirmation wherever the user is likely to navigate away, such as inside a form.
+- **Dismissing the toast commits early.** A user who clears the toast to get it out of the way ends their own window.
+- **Clearing the queue commits everything at once.** `clearToasts` runs every close handler, so a global clear puts through every deferred deletion rather than dropping it. That is the safe direction, but it is still a decision the user did not make one item at a time.
```
sebald: 💡 "That is the safe direction" is asserted, and I think it's the wrong way round

Putting through a batch of deletions the user never individually confirmed looks like the *un*safe direction to me: dropping deferred deletions is recoverable (the data survives, the user retries), committing them is not.

If the reasoning is "the interface already said 'deleted', so dropping them would make the UI a liar" — that's a good argument, and it's the one that makes this consistent with the honesty principle. But it's doing real work here and it isn't on the page. Worth spelling out, because a reader who reaches the opposite conclusion has no way to tell whether it was considered.

docs/content/patterns/feedback/destructive-actions/index.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+First check the action is destructive at all. Some actions look it and are not, and the clearest example is replacing an uploaded file: the user picks the replacement in the operating system's own file dialog, which they can cancel before anything changes. The confirmation already exists there, and it is the platform's. Putting another one in front of it asks the same question twice, which is the [never use both](#never-use-both) problem wearing a different hat.
+
+For everything that is destructive, work through these four questions in order. The first "no" decides to use a confirmation.
+
+| Question                                                 | Why it matters                                                                                                            |
+| -------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------- |
+| **Is it a single item, rather than a bulk action?**      | Bulk multiplies the cost of a mistake by the size of the selection. Bulk always confirms.                                 |
```
sebald, line 39: 💡 Only two tiers — consider naming the typed-confirmation escalation

This row sends *all* bulk actions to a plain confirmation. That under-serves the high end: "delete 4,000 recipients" and "remove one row" get the same dialog, differing only in copy.

GitHub, Stripe and AWS all add a third tier for high-consequence irreversible deletes — type the resource name to enable the confirm button. It's deliberately high-friction, and that's the point: it defeats the click-through habit the intro correctly worries about.

Marigold has no component for this today, so I'd not expect a demo. But a sentence acknowledging the tier, and where the boundary sits, would keep teams from concluding the page has no answer for the case.

docs/content/patterns/feedback/destructive-actions/index.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+Toasts follow the [feedback message](/patterns/feedback/feedback-messages) guidelines. Two rules matter here in particular.
+
+Name what happened to what. "Newsletter August deleted" tells a user who mis-clicked exactly what they lost, and "Item deleted" does not. For bulk actions, carry the count instead, as described in [Bulk Actions](/patterns/user-input/bulk-actions).
+
+Label the recovery with the verb, not the outcome. "Undo" is understood. "Restore", "Revert" and "Cancel" all invite a second reading, and "Cancel" is actively ambiguous next to a completed action.
+
+## Related
+
+- [Dialog](/components/overlay/dialog) for the confirmation component and the `useConfirmation` hook.
```
sebald, line 157: 💡 Two existing pattern pages cover this ground and aren't linked from either side

`patterns/data/fetching-and-mutations/index.mdx:202` already has a "confirming destructive actions" section built on `useConfirmation`, and its `optimistic-delete.demo.tsx` is a *working implementation* of what the "When your api can restore" section above recommends in prose (confirm → optimistic removal → toast naming the row → rollback on error → invalidate on settle).

This list has five entries and misses the closest neighbour. Worth linking in both directions — and the "When your api can restore" section in particular should point at that demo rather than describing it, since the demo is more convincing than the paragraph.

docs/content/patterns/feedback/destructive-actions/index.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+### What undo cannot cover
+
+Deferring the request buys the undo window, and it costs these:
+
+- **Leaving the page abandons the commit.** Close the tab during the window and the request never goes out, while the interface already reported the deletion. Keep windows short, and prefer a confirmation wherever the user is likely to navigate away, such as inside a form.
+- **Dismissing the toast commits early.** A user who clears the toast to get it out of the way ends their own window.
+- **Clearing the queue commits everything at once.** `clearToasts` runs every close handler, so a global clear puts through every deferred deletion rather than dropping it. That is the safe direction, but it is still a decision the user did not make one item at a time.
+
+### When your api can restore
```
sebald: 💡 `api` → `API` in the heading (and the anchor it generates)

```suggestion
### When your API can restore
```

Note this also changes the slug to `#when-your-api-can-restore`, which is what line 39 already links to — so that in-page link keeps working either way, but check it after the edit.

docs/content/patterns/feedback/destructive-actions/index.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+| Question                                                 | Why it matters                                                                                                            |
+| -------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------- |
+| **Is it a single item, rather than a bulk action?**      | Bulk multiplies the cost of a mistake by the size of the selection. Bulk always confirms.                                 |
+| **Is the effect visible where the user is looking?**     | Undo depends on the user noticing inside a few seconds. An item they already scrolled past is not noticed at all.         |
+| **Is it unmistakable which item was affected?**          | If the list reads as near-identical rows, the user cannot verify that the right one disappeared before the window closes. |
+| **Is a short delay safe, with no external side effect?** | A notification, a payment or a webhook cannot be deferred ambiguously, and cannot be recalled once it is out.             |
+
+Four times yes means the action can offer undo.
```
sebald: 💡 The decision rule is easy to skim past, and this sentence reads awkwardly

Two small things in one place:

1. "Four times yes means the action can offer undo" parses as a puzzle for a beat.
2. More importantly, the rule that actually drives the table — *the first "no" decides to use a confirmation* — sits in prose at line 28, **above** the table, where a reader who jumps straight to the questions will miss it entirely.

```suggestion
If all four answers are yes, the action can offer undo. The first "no" decides a confirmation.
```

That restates the rule *below* the table as well as above it, so it's unmissable from either reading direction.

docs/content/patterns/feedback/destructive-actions/index.mdx
```diff
@@ -0,0 +1,147 @@
+---
+title: Destructive Actions
```
sebald, line 2: 💡 Two structural notes, neither blocking

**Placement.** This sits under `patterns/feedback`, but the page is mostly about *gating an action* rather than reporting one — the confirmation half especially. Its closest sibling, `bulk-actions`, lives under `patterns/user-input`. Debatable rather than wrong, and moving it later costs a redirect, so worth deciding now rather than after the links are out in the world.

**Missing section.** Other Marigold pattern pages pair "Key principles" with "When to use" / "When not to use". This page has the first and not the second. Given the intro's strong claim that a confirmation on a harmless action weakens every dialog in the product, a "When not to use" section has obvious material to work with — the replace-an-uploaded-file example at line 26 is already exactly that, currently buried mid-paragraph.

docs/content/patterns/feedback/destructive-actions/undo-delete.demo.tsx
```diff
@@ -0,0 +1,141 @@
...
+  { id: '2', name: 'Jazz Night Reminder', recipients: 312 },
+  { id: '3', name: 'Season Preview', recipients: 878 },
+];
+
+const Lists = () => {
+  // `lists` is committed data, `pending` holds rows that are hidden but not
+  // deleted yet. They share one state value because `commit` has to read the
+  // current `pending` to decide whether to delete, and an updater cannot call
+  // another setter.
+  const [state, setState] = useState({
```
sebald: 💡 This is a `useReducer` wearing a `useState` costume

The fused `{ lists, pending }` state plus a four-line comment apologising for it ("an updater cannot call another setter") is the shape `useReducer` exists for. With `dispatch({ type: 'delete' | 'undo' | 'commit' | 'reset' })` the comment disappears, `commit`'s three-way guard becomes one ordinary case in a switch, and "why is this one object" stops being a question the reader has to resolve before reading on.

This matters more than a normal style call because the file is *documentation*. Whatever shape ships here is the shape teams copy, and right now the shape they'd copy comes with a caveat attached.

(To be clear the current code is *correct* — I traced it. `undo()` queues its updater before `removeToast` triggers `commit`'s, React applies updaters in order, so `commit` sees `pending` already drained and no-ops. The concern is legibility, not behaviour.)

docs/content/patterns/feedback/destructive-actions/file-delete-confirm.demo.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    if (result === 'confirmed') {
+      setFiles(current => current.filter(item => item.id !== file.id));
+    }
+  };
+
+  return (
+    <Stack space={2}>
+      {files.length === 0 ? (
+        <Text>All files deleted.</Text>
```
sebald: 💡 Two demos on one page, two idioms for the same empty state

This uses `<Text>All files deleted.</Text>`; `undo-delete.demo.tsx:57` uses `<EmptyState title description />` for the same job. A reader comparing the two demos — which is exactly what this page invites, since they're the two halves of one decision — sees a difference that carries no meaning.

`EmptyState` in both is the obvious call.

packages/components/src/Dialog/ConfirmationDialog.stories.tsx
```diff
@@ -124,9 +126,53 @@ Basic.test(
...
+        }
+      >
+        Delete file
+      </Button>
+      {result ? <div>{`resolved: ${result}`}</div> : null}
+    </>
+  );
+};
+
+const ConfirmationResultDemo = (props: ConfirmationResultProps) => (
```
sebald, line 191: 💡 `ConfirmationResultDemo` and the inline `Demo` below it are near-duplicates

The pre-existing test at line 178 still builds its own `ConfirmationProvider` + button-that-calls-`confirm` wrapper, which is now the same component as this one modulo labels and the absence of `variant: 'destructive'`. Giving `ConfirmationResult` a prop for the config (or just accepting the destructive labels in that test) collapses roughly 30 lines.

Separately: this hunk deleted the comment that used to explain why that block exists —

> `useConfirmation` — the imperative hook opens a dialog built from its config and resolves when the user confirms. A local `ConfirmationProvider` keeps the demo self-contained instead of relying on the global one from the decorator.

The "local provider instead of the decorator's global one" part is the non-obvious bit and it's now unexplained. Worth restoring above whichever component survives the merge.

sebald (changes requested):
## Automated Code Review

Reviewed the full diff, and verified the PR's claims about React Aria internals against `react-stately@3.48.0`'s `useToastState.mjs` rather than taking them on trust. **They all hold:**

- `ToastQueue.close(key)` does call `onClose?.()` synchronously before splicing, so `removeToast` and the toast's own close button both fire it.
- `ToastQueue.clear()` sets `queue = []` and never touches `onClose` — the hole this PR patches is real.
- `maxVisibleToasts` defaults to `Infinity` only when the queue is constructed directly (the `useToastState` hook's default is `1`). `getToastQueue()` uses `new ToastQueue({wrapUpdate})`, so `visibleToasts === queue`. The comment in `clearToasts` is accurate.
- Timers left dangling after `clear()` fire `close(toastKey)` against an empty queue → index `-1` → no double `onClose`.
- The undo demo's ordering is correct: `undo()` queues its updater before `removeToast` triggers `commit`'s, and React applies updaters in order, so `commit` sees `pending` already drained and no-ops.
- Outside-press genuinely does not dismiss a `useConfirmation` dialog (controlled path → `DialogContext` default `{}` → `isDismissable: undefined`).

The confirm/close race is handled correctly — the promise settles once, `onConfirm` runs before `close`, so `'confirmed'` wins over the `'cancelled'` that `onOpenChange` then fires — and it's covered by an explicit test. Good.

The new pattern page is stronger than most design-system writing on this topic. Framing the choice as **perceptibility and deferability, not reversibility** is correct and unusual; since undo here works by deferring the request, almost anything is "reversible" for five seconds, so reversibility can't be the test. Stating that and then striking it off the checklist is a real contribution. "Never use both", the commit-from-`onClose` trap, and the honest "what undo cannot cover" section are all well judged.

### Blocking

1. **Visual regression tests have never run on this branch.** Worth noting the gap is narrower than the checklist implies: `Basic` sets `chromatic: { disableSnapshot: true }` and the new destructive-focus test doesn't re-enable it, so the one genuinely visual change (Cancel carrying the focus ring on open) wouldn't be snapshotted even if VRT ran.
2. **An undo toast built with `variant="warning"` / `"error"` — or `timeout: 0` — never commits.** The only finding here that could cost real data if someone follows the page literally.

### Worth your attention

- `<ConfirmationDialog.Trigger>` blocks <kbd>Escape</kbd> by default, which is the exact thing the new page tells teams not to do. Pre-existing, but this PR is what turns it into a documented contradiction.
- `confirmationLabel` falls back to hardcoded English while `cancelLabel` is localized — there is no `confirm` key in `intl/messages.ts`.
- The `destructive` autofocus default can't be overridden to *no* autofocus, only to the other button.

### The one structural reservation

Not a defect in the diff. The page says of commit-on-`onClose`: *"This is the part that is easy to get wrong."* Three pieces have to be assembled correctly — hide, commit-on-close, guard-the-commit — and getting any of them wrong silently either loses data or loses the deletion. When a documented pattern is that sharp, it's usually the moment to build the abstraction rather than the prose: an `addUndoToast({ title, onCommit, onUndo })` would make the trap unreachable. Worth a follow-up ticket before this page becomes the contract.

Full detail in the inline comments.

---

packages/components/src/Dialog/ConfirmationDialog.stories.tsx
aromko: Snapshot gap closed: the destructive-focus test now sets `chromatic: { disableSnapshot: false }`, since it is the only case in the file that renders the destructive variant at all. VRT itself is still mine to trigger and I will do that before merge.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: Covered in three places. The page now has an explicit **Don't** for `variant="warning"`, `variant="error"` and `timeout: 0` on an undo toast, the Do example passes `timeout` so the window is a decision rather than an inheritance, and a sentence states that 5000ms is both the default and a `Math.max` floor with no ceiling.

The same constraint went into `ToastOptions.onClose`'s JSDoc, which is where someone reaching for a warning variant will actually be reading.

No dev-time warning for the combination: `onClose` on a persistent toast is legitimate whenever the handler is cleanup rather than a deferred commit, so it would fire on correct code.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: Fixed rather than scoped. `DialogTrigger` now defaults `keyboardDismissable = true`, so an unset prop no longer inverts into "keyboard dismiss disabled". That matches `Tray`, `TrayModal` and `useNonModal`, which all already defaulted to `true` — `Dialog` was the outlier — and it makes the Dialog page's existing claim that "pressing the Esc key" closes a dialog true for the trigger path.

New story test on `ConfirmationDialog`, whose `Basic` sets no `keyboardDismissable`, so it exercises the default. I stashed the one-line fix and re-ran it to confirm it fails without it. The changeset lists it as a visible behaviour change, and the Dialog page documents the `keyboardDismissable={false}` opt-out with the reason not to reach for it.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: Added as a fourth bullet under "What undo cannot cover": the window is a time limit, so WCAG 2.2.1 applies; the extension mechanism does exist, since `useToastRegion` calls `pauseAll()` while the region is hovered or focused and the region is a `useLandmark({role: 'region'})` reachable with <kbd>F6</kbd>; and it is not discoverable, so undo carries an accessibility floor a confirmation does not. The bullet ends where you did: let the four questions fall towards confirmation wherever the user may not be watching.

The 5000ms default-and-floor is stated with the `timeout` guidance further up. I also dropped "keep windows short" from the navigation bullet, since it was pulling against this.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: Decided, though not one of the two. The callout now names the constraint instead of shrugging: `Toast` renders its close button unconditionally, so an undo toast cannot drop it today, which is why the title has to carry the meaning.

Option 1 needs a `closeButton` option on `ToastOptions` threaded to `Toast.tsx`. Small, but I would rather it land with `addUndoToast`, which can pass it itself and make the ambiguity unreachable rather than merely undocumented. Option 2 costs the `onClose` symmetry the rest of the page rests on, including the `clearToasts` behaviour, so it is the one change I do not want to make in the same PR that argues for it.

Both are on the follow-up ticket with `addUndoToast`.

packages/components/src/Dialog/useConfirmation.tsx
aromko: Added a `confirm` key to both locales and formatted it in the provider the way `cancelLabel` already is, so the hardcoded English is gone.

One correction to the repro: `confirm({ title })` without a `confirmationLabel` does not typecheck. `ConfirmationConfig` picks `confirmationLabel` up from `ConfirmationDialogProps`, where it is required, so the only way to reach the fallback is untyped JS or the closed dialog rendered before the first `confirm()`. It was a latent default rather than a live German-UI bug. Fixed regardless, since the default should not be the broken one.

packages/components/src/Dialog/ConfirmationDialog.tsx
aromko: Took the reword, on the doc comment and on the pattern page's bullet, which now says `autoFocusButton: 'action'` instead of "override".

Not widening to `'none'`: nobody has asked for a destructive confirmation with nothing focused, and it is the less safe of the two reachable states, so the type stays at what it can actually deliver.

packages/components/src/Dialog/ConfirmationDialog.tsx
aromko: Removed the leak rather than documenting it: both buttons now call the handler bare and `close()` after it, and the `chain` import is gone. Shorter than the changeset paragraph would have been.

Worth noting for the changeset question: the released code called them bare too, so nothing consumer-visible changes here. The `chain` form only ever existed in this PR.

packages/components/src/Dialog/useConfirmation.tsx
aromko: Skipping this one.

`useCallback` with `[open]` does not fix the symptom it names: `confirm` closes over `open` for the already-open guard, so its identity still changes on every open and close, which is exactly when the re-render was reported. Making it stable means moving that guard to a ref and depending on nothing, and at that point it is a real change to a pre-existing hook for a cost nothing has measured. Consumers of `useConfirmation()` are a handful of buttons per screen.

Happy to do the ref version if you think it earns its place, but the `[open]` version would be motion without effect.

packages/components/src/Toast/ToastQueue.tsx
aromko: `getToastQueue()` now passes `maxVisibleToasts: Infinity` explicitly, and the `clearToasts` comment states the invariant as a fact about our queue instead of pointing at a react-stately default. Agreed on the reasoning: the two different defaults are exactly the kind of thing a future editor would not think to check.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: Added at the top of "Offering undo", right after the sentence that introduces the deferral: reporting the deletion up front is honest because the toast's own lifecycle guarantees the commit, since every way a toast can close ends with the request going out. That is also what makes the `setTimeout` version dishonest rather than merely fragile, which sets up the section below it.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: The reasoning is on the page now, and it is the one you guessed: it is the safe direction because the interface has already reported each of those deletions, so dropping them would leave the user looking at a list that quietly grew its rows back. The sentence keeps the caveat that it is still a decision the user did not make one item at a time.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: Named it below the table: the boundary (a high-consequence irreversible delete, a whole account or thousands of recipients), what the escalation is (type the name of the thing before the confirm button enables), why it is deliberately high-friction, and that Marigold has no component for it, so it gets built in place for now with a pointer to get in touch.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: Linked in both directions. "When your API can restore" now points at `fetching-and-mutations#destructive-actions` and says what the `optimistic-delete` demo does, instead of describing the shape in prose; the Related list carries the same link; and that page's own destructive-actions section now points back here for the choice between confirming and offering undo.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: Applied. The slug is unchanged, since it lowercases either way, so the line-39 link still resolves.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: Took the sentence, and removed the duplicate clause from the paragraph above the table so the rule is not stated twice verbatim one line apart. It now reads once, immediately under the questions, where a reader who skipped straight to the table lands.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: **Missing section** is in: "When not to use", built from the replace-an-uploaded-file case, which is now a bullet of its own rather than buried in the paragraph, plus two more (nothing is lost, redoing it costs a click). It sits between "Key principles" and the decision table.

**Placement** stays under `patterns/feedback` for now. Both halves are about the recovery path the user sees, and the undo half is literally a toast. It is a genuine coin flip and nothing links to the page yet, so if you would rather have it next to `bulk-actions` under `user-input` say so and I will move it before merge, while a redirect still costs nothing.

docs/content/patterns/feedback/destructive-actions/undo-delete.demo.tsx
aromko: Leaving the demo as it is, and the diagnosis is fair: the fused state plus the comment defending it is a reducer in disguise, and the docs argument is the strongest part of the case.

What holds me back is the trade. A reducer here means an action union, a `dispatch` call site per action and a switch, in a file whose job is to be skimmed next to the prose that explains it. That is more to read than the four lines it deletes, and the `commit` guard does not get simpler, it just moves into a `case`. I would rather keep the comment and spend the reader's attention on the `onClose` wiring, which is the part they need to copy correctly.

Not a hill worth holding if you disagree, the rewrite is contained.

docs/content/patterns/feedback/destructive-actions/file-delete-confirm.demo.tsx
aromko: Both demos use `EmptyState` now, so the two halves of the decision no longer differ in a way that carries no meaning.

packages/components/src/Dialog/ConfirmationDialog.stories.tsx
aromko: Collapsed. The inline `Demo` is gone and that test renders `ConfirmationResultDemo`, narrowed to the one thing it uniquely covers: the dialog is built from the config passed to `confirm()`. Confirm-and-close is already asserted twice above it, so re-asserting it there was buying nothing and split the test across two acts.

Restored the local-provider comment above the surviving component, since that is the non-obvious part.

aromko:
Worked through all 21 findings. Two commits, `c05fe1636` (components) and `50d5b8ace` (docs), with per-thread replies above. Summary of where each blocking item landed, plus the one thing I did not build.

### Blocking

1. **VRT** — the destructive-focus test now re-enables its Chromatic snapshot, since it was the only case in the file rendering that variant. Triggering the run is mine to do before merge.
2. **An undo toast that never auto-dismisses** — explicit **Don't** for `variant="warning"`, `variant="error"` and `timeout: 0`, `timeout` shown in the Do example, the 5000ms default-and-floor stated, and the same constraint carried into `ToastOptions.onClose`'s JSDoc.

### Worth your attention

- **Escape** — fixed rather than scoped. `DialogTrigger` defaults `keyboardDismissable = true`, matching `Tray`, `TrayModal` and `useNonModal`; `Dialog` was the only overlay inverting an unset prop, and the Dialog page already claimed Escape worked. Covered by a story test that I confirmed fails without the fix, listed in the changeset as a visible behaviour change, and the opt-out is documented.
- **`confirmationLabel`** — `confirm` key added and formatted like `cancelLabel`. One correction: `confirm({ title })` without a label does not typecheck, since `ConfirmationConfig` inherits it as required, so this was a latent default rather than a live German-UI bug.
- **`autoFocusButton`** — took the reword. Not widening to `'none'`, which nothing asks for and which is the less safe state.

### The structural reservation

Agreed, and not in this PR. `addUndoToast({ title, onCommit, onUndo })` designed against a single demo would bake in whatever that demo happens to do, and the page says out loud that the undo half has no product implementation yet. Filing it as a follow-up, together with the `closeButton` option it needs, so the close-button ambiguity becomes unreachable inside the helper rather than documented on the page. Same for the `FileField` escape hatch and its fixed accessible name.

### Two open questions

- **Page placement.** Kept under `patterns/feedback`. Genuine coin flip, nothing links to it yet, so it is free to move to `user-input` next to `bulk-actions` if you would rather.
- **The `useReducer` rewrite.** Declined for now with reasoning in the thread. Contained change if you want it.

### Declined

`useCallback` on `confirm` with `[open]` — it does not fix the re-render it names, since the identity still changes on every open and close. A stable one needs the guard in a ref, for a cost nothing has measured.

packages/components/src/Dialog/ConfirmationDialog.stories.tsx
sebald: The snapshot fix is the right shape: that test is the only case in the file rendering the destructive variant, so re-enabling there covers the focus ring without re-snapshotting everything else.

Leaving this one open rather than resolving it, since the finding was that VRT has never run on this branch and it still hasn't. Yours to trigger — ping me once it's green and I'll close it out.

docs/content/patterns/feedback/destructive-actions/index.mdx
sebald: Covered, and `ToastOptions.onClose`'s JSDoc is the right second home for it: that is where someone reaching for a warning variant is actually reading.

One residual. `undo-delete.demo.tsx` still doesn't pass `timeout`. The page now says to set it deliberately rather than inherit it, and the Do snippet shows `timeout: 8000`, but the demo — the thing teams copy — inherits the default. That's your own argument against the reducer rewrite pointed the other way: whatever ships in the demo is the shape that gets copied, so the page and its reference implementation should agree.

Agreed on no dev-time warning. `onClose` as cleanup on a persistent toast is legitimate, so it would fire on correct code.

docs/content/patterns/feedback/destructive-actions/index.mdx
sebald: Fixing it beats scoping the sentence, and stashing the one-line fix to watch the test fail is the right way to prove the story covers the default rather than the arg. The finding itself is closed.

Two things next to it, neither blocking:

1. `Dialog.stories.tsx:47` still sets `keyboardDismissable: true` in `args`. That's now redundant, and it means `Dialog.test.tsx`'s Escape test would keep passing if the default ever regressed — the only coverage of the new default is the `ConfirmationDialog` story. Dropping the arg moves the coverage to where the behaviour now lives, for one line.

2. The inversion the default rests on is mislabeled twice on the way down: `DialogTrigger.tsx:35` puts `!keyboardDismissable` into the context, `Dialog.tsx:155` passes that already-negated flag into a prop *named* `keyboardDismissable`, and `Underlay.tsx:36` assigns it straight to `isKeyboardDismissDisabled`. Net behaviour is correct — I traced it before filing the original finding — but two hops carry a boolean whose name means the opposite of its value. `Loader.tsx:42` is the evidence that it already misleads: it passes `keyboardDismissable` to `Underlay` and thereby *disables* keyboard dismiss. Pre-existing and out of scope, but this PR is what makes the default load-bearing, so it belongs on the same follow-up ticket.

docs/content/patterns/feedback/destructive-actions/index.mdx
sebald: Better than either option I offered. Naming the constraint — `Toast` renders its close button unconditionally, so the title has to carry the meaning — is honest in a way "say what happened in the title" wasn't, because it says *why* the title is doing that work.

Agreed on the sequencing. Option 1 wants a `closeButton` option that `addUndoToast` can set itself, and option 2 would cost the `onClose` symmetry the rest of the page argues from; making that trade in the same PR that introduces the argument is the wrong order. Resolving, it's on the follow-up.

packages/components/src/Dialog/useConfirmation.tsx
sebald: You're right and I was wrong. `confirm` closes over `open`, so `[open]` changes identity on exactly the renders I claimed it would fix — the suggestion was motion without effect. The ref version is a real change to a pre-existing hook for a cost nothing has measured, and a handful of buttons per screen doesn't earn it. Dropping this; no need for the ref version.

docs/content/patterns/feedback/destructive-actions/index.mdx
sebald: "When not to use" reads well, and pulling the replace-an-uploaded-file case out of the paragraph into a bullet of its own is what makes it usable.

**Placement: keep it under `patterns/feedback`.** Your argument wins on the folder's coherence — `feedback-messages` and this page are both about what the system tells the user and how they get back out of it, whereas `user-input` is about how users express intent. A page whose second half is toast lifecycle sits badly there. My "it's really about gating" point holds for the confirmation half only, and that half is one dialog while the undo half is the mechanism nothing else documents.

The adjacency to `bulk-actions` is real either way, so do it with links rather than folders: this page already lists Bulk Actions under Related, and `bulk-actions` has a `Related` TeaserList with no link back. Adding the reverse entry puts them one hop apart from either direction, and nothing ever needs a redirect.

docs/content/patterns/feedback/destructive-actions/undo-delete.demo.tsx
sebald: Taking your call. The trade is real: an action union plus a dispatch site per action plus a switch is more to read than the four lines it deletes, in a file whose job is to be skimmed next to the prose that explains it. And you're right that the `commit` guard doesn't get simpler, it just moves into a `case`. Spending the reader's attention on the `onClose` wiring is the correct priority for a demo. Resolving.

sebald:
Went through all 21 findings against the code at `c24157b0` rather than the replies. Everything claimed is actually there, including the parts that went further than asked — removing `chain` outright instead of documenting the press-event leak, and fixing the `Escape` default at `DialogTrigger` instead of scoping the sentence to the hook.

**17 threads resolved.** Both open questions answered in their threads: the page **stays under `patterns/feedback`** (your reasoning wins on folder coherence; add the reverse link from `bulk-actions`'s Related list so the adjacency is carried by links rather than folders), and the `useReducer` rewrite is **dropped** — your trade is the right one for a file that gets skimmed next to its prose. The `useCallback` suggestion was wrong and I've withdrawn it.

**3 still open**, none of them a rewrite:

1. **VRT** — the snapshot gap is closed, the run itself hasn't happened. Yours to trigger.
2. **`undo-delete.demo.tsx` doesn't pass `timeout`** — the page now says to set it deliberately and the Do snippet does, but the demo teams copy from inherits the default.
3. **Two small things around the `Escape` fix** — the now-redundant `keyboardDismissable: true` in `Dialog.stories.tsx:47` (which hides a regression in `Dialog`'s own Escape test), and the double-mislabeled negation it flows through. Detail in the thread; the second is follow-up material.

One more, too small for its own thread: `docs/scripts/build-types.ts` now falls back to `.ts` without checking it exists, so a genuinely missing file produces a "no such file" error naming a `.ts` path even where the author meant `.tsx`. Only affects the diagnostic, not the resolution.

Happy to approve once VRT is green — the other two are your call whether to fold in or ticket.

packages/components/src/Dialog/ConfirmationDialog.stories.tsx
aromko: Still yours to trigger, and still not green — leaving it open is right.

Worth knowing before you look: the branch has moved a fair way since that finding. `9e32416a1` builds both follow-up tickets, so the diff now includes an undo toast with no close button (a new visual state), a real `<FileField>` in the confirm demo instead of composed items, and one more Chromatic-enabled story case. I'll ping here once the run is green.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: This one is aimed at wording that has since gone. The `/simplify` pass cut the "set `timeout` deliberately rather than inheriting it" sentence and the `timeout: 8000` snippet along with it — the snippet was a hand-copied duplicate of the demo seven lines above, and the demo already highlights those exact lines. What the page says now is that the window is 5000ms unless you pass `timeout`, that 5000 is also the floor, and that five to ten seconds is the common range, so pass a value when the decision needs longer. A three-row list at the default is consistent with that, so the disagreement you found is gone rather than fixed.

Happy to add an explicit `timeout` to the demo anyway if you'd rather it model the option existing — one line, and I take the point that the demo is the shape teams copy. Your call.

The residual worth more than that one: `addUndoToast` now owns the timeout, and it deliberately diverges from `addToast` on `0`. `addToast({ timeout: 0 })` persists; `addUndoToast({ timeout: 0 })` falls back to the default window, because an undo toast that never closes never commits. That divergence is in `UndoToastOptions.timeout`'s JSDoc and pinned by a test in the timeout-resolution block.

docs/content/patterns/feedback/destructive-actions/index.mdx
aromko: Both taken, in `960b37ed5`.

1. `keyboardDismissable: true` is out of `Dialog.stories.tsx`'s args. You're right about what it was hiding: `Dialog.test.tsx`'s Escape test was passing on the arg, not on the default, so the only coverage of the new behaviour was the `ConfirmationDialog` story. It now exercises the default, and the eight Dialog/Drawer/Tray/Overlay unit files plus the Dialog story tests are green without it.

2. The double-mislabeled negation is on the follow-up ticket, recorded in the PR description so it doesn't evaporate. `Loader.tsx:42` is the good catch there — it reads as "keyboard dismissable" and does the opposite, which is exactly what a name that means its own inverse buys you.

aromko:
Second round addressed, plus a reversal you should look at rather than skim.

### The reversal

Both follow-up tickets are built, in `9e32416a1`. I argued against `addUndoToast` in this PR and was overruled, so it is here with the reasoning intact rather than half-done: the page now documents an API instead of a recipe with three ways to lose data quietly.

- **`addUndoToast({ title, onUndo, onCommit })`** owns the timing. The commit rides the toast's own `onClose`, an undo cancels it, the toast has no close button, and `timeout: 0` falls back to the default window instead of meaning "never commit". The undo button is named after the title through an `undoNamed` message, so the stacked-toast ambiguity is handled by construction rather than by every caller writing an `aria-label`.
- **`<FileField onBeforeRemove>`** is the interception point DSTSUP-259 needed. Each remove button is also named after its own file now, so the "Remove file" limitation the page used to document as permanent is gone, and the confirm demo is a real `<FileField>` instead of composed items over a fake list.

Two consequences worth naming. `closeButton: false` on `ToastOptions` is new public surface, and the invariant that a toast which never auto-dismisses keeps its close button now lives in `Toast.tsx` rather than in `addToast` — the render layer is the only place every toast passes through, so `addUndoToast` inherits it by delegating instead of asserting it again. And the undo demo lost its toast-key ref, its commit guard and its fused state object, which retires the `useReducer` question from the other direction: the coupling that forced the fused state was the manual `onClose` wiring, not the choice of hook.

### Your three

1. **VRT** — still yours, still not run. The diff has grown a new visual state since the finding (an undo toast with no close button), so it is worth a look when it goes green.
2. **Demo `timeout`** — aimed at wording that has since gone; detail in the thread, along with one line you may still want.
3. **Both small things around `Escape`** — done in `960b37ed5`. The redundant arg is out of `Dialog.stories.tsx`, which moves the regression coverage onto the real default, and the double-mislabeled negation is recorded on the follow-up ticket with your `Loader.tsx:42` evidence.

Also fixed your no-thread item: `build-types.ts` falls back to `.ts` only when that file exists, so a missing file reports the `.tsx` path the author meant. And `bulk-actions` now links back here, so the adjacency is carried by links rather than folders, as agreed.

### Unprompted, since it affects you

`pnpm test` was dead on macOS on `main`. `browserScanEntries` still crawled `packages/cli` even though `exclude` skips it, and the CLI's Node tests drag the optional `fsevents` into the browser dep pass, whose `.node` binary rolldown cannot load — every run died with `UNLOADABLE_DEPENDENCY` before collecting a test. Linux CI never sees it because fsevents does not install there. Narrowed in the merge commit; worth cherry-picking if this branch is not the first thing to land.

### State

CI green on `c24157b0`; the two new commits are building. Full local run before pushing: 1539 unit tests, 41 story tests, root and docs typecheck, lint clean. A `/simplify` pass over the new API ran before the commit and moved the close-button invariant, dropped a `pendingRemovals` guard that could not fire (`ConfirmationProvider` is already single-flight and the underlay blocks the row), and replaced a hardcoded `"{label}: {title}"` with the message key — findings I would rather you saw in the diff than in this comment.

Still open on my side: nothing. Yours: VRT, and whether the demo takes an explicit `timeout`.

packages/components/src/Dialog/ConfirmationDialog.stories.tsx
aromko: VRT is green: [run 31387571326](https://github.com/marigold-ui/marigold/actions/runs/31387571326) on `960b37ed5`, which is current head, so it covers both new APIs rather than the state the finding was filed against.

Two things it now covers that it would not have then: the destructive `ConfirmationDialog` case re-enables its snapshot (`chromatic: { disableSnapshot: false }`), which was the gap in the original finding, and the diff has since grown an undo toast with no close button. Worth a look at that one in particular, since dropping the close button is the visual half of the argument the pattern page makes.

Yours to close out.

sebald: Green on `960b37ed5`, and confirmed it is current head, so the run covers both new APIs rather than the state this finding was filed against. The destructive `ConfirmationDialog` case re-enabling its snapshot was the actual gap here, and that is in.

Closing this out.

docs/content/patterns/feedback/destructive-actions/index.mdx
sebald: My call: **leave the demo at the default.**

My original argument was that the demo is the shape teams copy, and it held when the recipe was hand-assembled — a caller who never saw `timeout` had no idea the window was theirs to size. That is not the situation any more. `addUndoToast` owns the timeout, `0` falls back to the default window instead of meaning "never commit", and `timeout` is on the generated props table. The fallback is now a property of the API rather than something a caller can fumble, so a demo sitting at the default is the honest default rather than an omission.

The divergence from `addToast` on `0` is the right one, and putting it in `UndoToastOptions.timeout`'s JSDoc with a test in the timeout-resolution block is where it belongs — that is the one place someone reaching for "unlimited undo" will actually read.

Resolving.

docs/content/patterns/feedback/destructive-actions/index.mdx
sebald: Both confirmed on `960b37ed5`.

1. `keyboardDismissable: true` is out of `Dialog.stories.tsx`'s args, so the Escape test now passes on the default rather than on the arg, which is where the coverage needed to live.
2. The double-mislabeled negation is recorded in the PR description, so it survives the merge instead of evaporating with this thread.

Resolving.

sebald (approved):
## Approved

All twenty threads are closed. Both blockers are genuinely fixed rather than argued away: the undo toast can no longer be built into a state where it never commits, and VRT is green on current head with the destructive case re-enabling its snapshot — which was the real gap, since the run alone would not have caught the focus ring.

**This approval covers new API surface, not just the fixes.** `960b37ed5` built both follow-up tickets into the branch, so `addUndoToast`, `UndoToastOptions`, `ToastOptions.onClose`, `ToastOptions.closeButton`, `FileField.onBeforeRemove`, `FileField.Item.removeLabel` and four intl keys arrived after the review that requested changes. Approving them implicitly would have been the wrong way to ship a published export, so I read them, and verified the React Aria claims against `react-stately@3.49.0` in the tree rather than the 3.48 I checked the first time — the queue internals are identical between them.

What I checked and what holds:

- `add()` spreads `options` onto the queued toast, so `toast.timeout` and `toast.onClose` really do exist. Both new reads are sound.
- The close-button condition inverts correctly. Hidden only when `closeButton === false` **and** a timeout exists, so `warning`, `error` and `timeout: 0` keep their only way out. That matches what `ToastOptions.closeButton` documents.
- `undone` is set before `removeToast(key)`, so the undo press's own close does not commit.
- `clearToasts` runs the handlers before `clear()`, and the timers that survive fire `close()` against an empty queue → index `-1` → no double `onCommit`.
- `handleRemove` is the only removal path in `FileField`, so there is no second door around `onBeforeRemove`.
- `maxVisibleToasts: Infinity` is now explicit, which is what makes the `visibleToasts === queue` read in `clearToasts` a fact about our queue rather than an inherited default.

Building the two follow-ups here instead of deferring them was the right reversal. The page now documents an API rather than a three-part recipe where getting any part wrong silently loses either the data or the deletion, and `addUndoToast` makes the trap unreachable rather than merely documented. `timeout: 0` falling back to the default window instead of meaning "never commit" is the detail that earns it.

One nit, not blocking and fine to fix in passing or not at all: `onPress={() => void handleRemove(file)}` means a throwing `onBeforeRemove` becomes an unhandled rejection. The file is kept, which is the half that matters, but "the error propagates" in the JSDoc and changeset overstates where it goes — there is no caller to catch it.

Ship it.

PR #5717, opened 2026-08-06
docs(DST-1427): promote EmptyState from beta to stable

sarahgm:
# Description

`<EmptyState>` leaves beta. It shipped in January and has had no API churn since, and v18 is where its surface settled: the title became a semantic heading behind a configurable `headingLevel`, and the description moved onto the shared `<Description>` primitive. `title`, `description`, `action`, `headingLevel`, `variant` and `size` are covered by the usual semver guarantees from here on. No component code changes.

Beyond the ticket's stated scope (badge, changeset, release notes), this also brings the page up to the current docs standard, since promoting a page to stable is the moment to fix what it was missing:

- **Actions section** — the component had no guidance on its `action` prop. It maps the action's role to a Button variant and states the one-primary rule for the case where an empty state stands in for a page's content but the `Page.Header` above it already owns the page's primary action. Button's Surface actions section now names the components that reset the button cascade, since its "you rarely need to set a `variant` yourself" advice does not hold inside them.
- **Anatomy** — a themed inline `EmptyStateAnatomy` SVG replaces the static `emptystate-anatomy.jpg`, so the diagram follows light and dark mode instead of being a baked-in light-mode raster, and the section gains the labelled-parts list every other anatomy has.
- **Seven dead links** found while cross-referencing these pages, across the app frame pattern, the Button docs, the spacing page and the recommended libraries list. All of them 404 today.

N/A — no UI changes, so no visual regression coverage needed. All changes are under `docs/`.

Closes DST-1427

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm start` and open `/components/content/empty-state`. Confirm the `beta` badge is gone.
2. Check the Anatomy section: the diagram renders inline and is followed by the labelled-parts list.
3. Toggle dark mode. The diagram should follow the theme — the old `.jpg` did not.
4. Read the new Actions section and confirm its three Button links resolve (`#visual-hierarchy`, `#placement-and-order`, `#surface-actions`).
5. Spot-check the repointed links: `/patterns/layout/app-frame` (Empty state section), `/components/actions/button` (Button or link, Surface actions, the `useActionState` callout), `/foundations/spacing`, `/getting-started/recommended-libraries`.
6. Open `/releases/blog/release-2026-08-11` and confirm the new "`<EmptyState>` is stable" entry sits under Components, before `<ErrorState>`.

## Breaking Changes

No

## Checklist


docs/content/components/actions/button/index.mdx
```diff
@@ -306,7 +305,9 @@ Anchor each surface with at most one prominent `<Button>` for the real commitmen
...
+
+Some containers deliberately do the opposite and reset the cascade to a clean baseline, so their action never inherits the ghost or small treatment of a container further up. A `<Button>` in one of them keeps its own default `secondary`, and you pick the variant yourself. This covers the message components, [`<EmptyState>`](/components/content/empty-state), [`<ErrorState>`](/components/content/error-state), [`<SectionMessage>`](/components/content/section-message) and [`<ContextualHelp>`](/components/overlay/contextual-help), and it covers overlays like [`<Dialog>`](/components/overlay/dialog) and [`<Drawer>`](/components/overlay/drawer), which open a surface of their own that the cascade should not reach through.
```
sebald: Two things on this paragraph.

**`<Dialog>` does not reset the cascade itself.** `RESET_BUTTON_CONTEXT` / `<ResetButtonContext>` is used by `ContextualHelp`, `EmptyState`, `ErrorState`, `SectionMessage`, `Drawer`, `Popover`, `Modal` and `Tray` — `Dialog` is not among them. A `<Dialog>` gets the reset from the `Modal` or `Popover` it renders into, not from itself. The end result is the same for anyone using `Dialog.Trigger`, so the guidance is not misleading in practice, but it attributes the behaviour to the wrong component. Since `Popover`, `Modal` and `Tray` have no docs pages to link, something like "overlays like [`<Drawer>`](/components/overlay/drawer), and anything rendered inside a [`<Dialog>`](/components/overlay/dialog)" would stay accurate and still link only to pages that exist.

**This paragraph separates the demo from the paragraph it illustrates.** `<ComponentDemo name="button-in-context" />` two lines below demonstrates the *inheritance* case from the paragraph above this one, but now reads as an example of resetting. Moving this paragraph to after the demo keeps each claim next to its example.

sebald (commented):
Docs read well and I checked the link fixes against the checkout — all seven repointed routes and every anchor (`#async-forms-with-useactionstate`, `#validation`, `#visual-hierarchy`, `#placement-and-order`, `#surface-actions`) resolve, and the new Anatomy prose matches the component and theme (`max-w-sm`, no background, `aria-hidden` illustration, `headingLevel` default 3, `action` wrapper with no gap). Two notes on the Button page inline, plus one on the changeset:

**The `@marigold/components` bump is a no-op publish.** The changeset lists `@marigold/components` as `patch` while stating "No code changes". `.changeset/config.json` has `fixed: [["@marigold/system", "@marigold/components", "@marigold/docs"]]`, so the version moves regardless of which of the three is named — the actual effect of naming `components` is that this entry lands in the components CHANGELOG. That is arguably the right call, since leaving beta is consumer-facing information and the components changelog is where consumers look for it. Just flagging that it is a deliberate choice rather than a side effect, in case the intent was docs-only.

docs/content/components/actions/button/index.mdx
sarahgm: Both correct, fixed in 8fe9ad4fc.

**Attribution.** Confirmed against the source: `RESET_BUTTON_CONTEXT` / `<ResetButtonContext>` appears in `ContextualHelp`, `EmptyState`, `ErrorState`, `SectionMessage`, `Drawer`, `Tray`, `Overlay/Modal` and `Overlay/Popover`, and `Dialog` is indeed not among them. The paragraph now reads "overlays like [`<Drawer>`](/components/overlay/drawer) and anything rendered inside a [`<Dialog>`](/components/overlay/dialog), since an overlay opens a surface of its own that the cascade should not reach through." I reworded the trailing clause instead of taking your wording verbatim, because once the sentence ends on "anything rendered inside a `<Dialog>`", the old "which open a surface of their own" had no plural antecedent left to attach to.

**Placement.** Also right, and `button-in-context.demo.tsx` settles it: it renders a `<Button>` inside a `<SelectList.Option>`, which is the inheritance case from the paragraph above. The reset paragraph now sits below the demo, so each claim keeps its own example.

sarahgm:
On the changeset: deliberate, and I'd like to keep it.

You're right that the version moves either way because of the `fixed` group, so the only real effect of naming `@marigold/components` is which CHANGELOG the entry lands in. That is the reason it's named. A component leaving beta is consumer-facing information, and the components changelog is where a consumer goes looking for it. Routing this entry into `@marigold/docs` only would file "the `EmptyState` API is now covered by semver" under a package nobody installs.

So "No code changes" in the body is describing the diff, not arguing for a docs-only release. Happy to drop the line if you'd rather the components changelog stay strictly code-only, but my preference is to leave it.

The two Button page notes are addressed in 8fe9ad4fc, replies inline.

.changeset/dst-1427-empty-state-stable.md
```diff
@@ -0,0 +1,14 @@
+---
+'@marigold/components': patch
+'@marigold/docs': patch
+---
+
+docs(DST-1427): promote `EmptyState` from beta to stable
+
+`EmptyState` has been in beta since January. v18 is where its surface settled, with the title becoming a semantic heading behind a configurable `headingLevel` and the description moving onto the shared `Description` primitive, so the badge comes off now that the API being frozen is the one 18.0.0 ships. `title`, `description`, `action`, `headingLevel`, `variant` and `size` are covered by the usual semver guarantees from here on. No code changes.
```
aromko: ⚠️ **`variant` and `size` are the weakest part of this promotion**

`EmptyState.styles.ts` defines no `variants` at all, both props are typed bare `string` with no JSDoc, and the page's own Appearance section renders *"The selected theme does not has any options for `variant` and `size`"*. The props table shows them as `variant? string` / `size? string`, undocumented.

So the promotion promises semver stability on two props that accept anything and do nothing. Narrowing them to a union later, which is the natural move once EmptyState gets variants, would then be a breaking change: exactly the risk this promotion is meant to remove.

Either scope the guarantee to the four props that carry meaning, or say explicitly that `variant`/`size` are theme-driven and theme-rui defines no values today.

The same sentence is in `release-2026-08-11.mdx:717` and needs the same edit.

docs/content/releases/blog/release-2026-08-11.mdx
```diff
@@ -712,6 +712,12 @@ Built-in localized presets cover the common cases (`today`, `yesterday`, `tomorr
 
 `<Table>` gains a footer: a semantic `<tfoot>` rendered after `<Table.Body>` for summary rows like totals, counts, or averages, composed from `<Table.Row>` and `<Table.Cell>` just like the body. A `sticky` prop pins the footer to the bottom of the viewport while scrolling, mirroring sticky table headers.
 
+### `<EmptyState>` is stable
+
+`<EmptyState>` leaves beta. It shipped in January, and v18 is where its surface settled: the title became a semantic heading with a configurable `headingLevel`, and the description moved onto the shared `<Description>` primitive. That surface is now final, so `title`, `description`, `action`, `headingLevel`, `variant`, and `size` are covered by the usual semver guarantees.
```
aromko: ⚠️ Same `variant` / `size` point as in the changeset: theme-rui defines no variants or sizes for `EmptyState`, so freezing those two under semver over-promises. See the comment on `.changeset/dst-1427-empty-state-stable.md`.

.changeset/dst-1427-empty-state-stable.md
```diff
@@ -0,0 +1,14 @@
...
+
+Also fixes seven dead links found while cross-referencing these pages. The app frame pattern pointed `EmptyState` at `/components/feedback/empty-state`, and the Button docs named a `ButtonLink` component at `/components/actions/button-link`, linked `SelectList.Option` to `/components/form/select-list`, and sent the `useActionState` callout to `/patterns/user-input/form-implementation`. That last route was dead in two more places, both in the recommended libraries list. All of these routes 404. They are now `/components/content/empty-state`, `LinkButton` at `/components/actions/link-button`, `/components/form/selectlist`, `/patterns/user-input/forms#async-forms-with-useactionstate`, and `/patterns/user-input/forms` plus `/patterns/user-input/forms#validation` for the two library references. The seventh, the spacing page's link to `/components/overlay/tray`, has no page to point at because `Tray` is not documented, so it drops the link and keeps `Tray` as plain code.
```
aromko: 💡 **This paragraph lands in the `@marigold/components` changelog**

`@marigold/components` and `@marigold/docs` are in the same `fixed` group, so this text is published verbatim to the changelog npm consumers read, where "the app frame pattern pointed `EmptyState` at `/components/feedback/empty-state`" is noise.

Splitting into two changeset files, one listing `@marigold/components` with the promotion and one listing `@marigold/docs` with the housekeeping, keeps the consumer-facing entry about the API. The `fixed` group still bumps both together.

docs/content/components/content/empty-state/empty-state-anatomy.tsx
```diff
@@ -0,0 +1,173 @@
...
+    />
+
+    {/* Action */}
+    <rect
+      x="320"
+      y="240"
+      width="100"
+      height="30"
+      rx="6"
+      className="fill-fd-accent transition-colors duration-300"
```
aromko: 💡 **The Action rect nearly disappears in dark mode**

`fill-fd-accent` with no stroke sits very close to the page background in dark mode; on the preview the shape only reads because the connector dot lands on its edge. `section-message-anatomy.tsx` outlines its surfaces with `fill-fd-card stroke-fd-border` and `strokeWidth="2"`.

````suggestion
      className="fill-fd-accent stroke-fd-border transition-colors duration-300"
      strokeWidth="2"
````

File: .changeset/dst-1427-empty-state-stable.md (hunk 2 above)
Comment (sarahgm) [reply .changeset/dst-1427-empty-state-stable.md]: Half of this lands, and the half that lands is fixed in 9345cef2d.

**The prop typing is not the risk.** Bare `variant?: string` is the house style rather than an `EmptyState` slip: `Tabs`, `Sidebar`, `Drawer`, `Calendar`, `Radio`, `RadioGroup`, `TagGroup`, `Tag`, `HelpText`, `ButtonGroup`, `TimeField`, `Autocomplete`, `TagField`, `Panel` and `ContextualHelp` all type it the same way. And the pattern for adding values later is `'a' | 'b' | (string & {})`, which 35 files in `packages/components/src` already use. That widens the accepted set with autocomplete instead of narrowing it, so gaining variants is additive, not breaking. Nothing to guard against there.

**The wording was the problem.** You are right that `theme-rui`'s `EmptyState.styles.ts` has four base slots and no `variants` block, so listing `variant` and `size` alongside `title` and `description` as covered by semver read as though they did something. The guarantee is now scoped to `title`, `description`, `action` and `headingLevel`, followed by a sentence saying `variant` and `size` stay open ended as they do on every themed component and that `theme-rui` defines no values for either today. The props stay: they are the universal theme hook, and a no-op one is not a dead prop.

`release-2026-08-11.mdx:717` got the same edit.

File: docs/content/releases/blog/release-2026-08-11.mdx (hunk 3 above)
Comment (sarahgm) [reply docs/content/releases/blog/release-2026-08-11.mdx]: Fixed in 9345cef2d. The sentence now stops at `headingLevel` and adds: "`variant` and `size` stay open ended, as they do on every themed component, and `theme-rui` defines no values for either today." See the reply on the changeset for why the props themselves stay.

File: .changeset/dst-1427-empty-state-stable.md (hunk 4 above)
Comment (sarahgm) [reply .changeset/dst-1427-empty-state-stable.md]: Agreed, split in 9345cef2d.

`dst-1427-empty-state-stable.md` now lists only `@marigold/components` and carries the promotion. A new `dst-1427-empty-state-docs.md` lists only `@marigold/docs` and carries the Actions section, the Anatomy rework and the dead-link fixes. The `fixed` group still bumps all three together, so this only changes which changelog each paragraph lands in. Two changesets for one ticket has precedent (`dst-1480-panel-html-attributes.md` / `dst-1480-panel-aria-labelledby-guard.md`).

Worth noting this does not conflict with the point @sebald raised: naming `@marigold/components` was deliberate, and it stays named. The split just stops "the app frame pattern pointed `EmptyState` at `/components/feedback/empty-state`" from riding along into the changelog consumers read.

File: docs/content/components/content/empty-state/empty-state-anatomy.tsx (hunk 5 above)
Comment (sarahgm) [reply docs/content/components/content/empty-state/empty-state-anatomy.tsx]: The observation is right, the suggested fix is not. Addressed differently in 9345cef2d.

**The shape does read poorly, and in both modes.** Compositing the fumadocs tokens over the page background:

| | light (`#f5f5f5`) | dark (`#121212`) |
| --- | --- | --- |
| `fill-fd-accent` composited | `#e3e3e3`, 1.18:1 | `#2c2c2c`, 1.34:1 |

So it is not only dark mode. Light is marginally worse, it just reads as "flat page" rather than "missing shape".

**`stroke-fd-border` does not fix it.** `--color-fd-border` is `#cccccc80` in light and `#6663` in dark, both translucent. Composited over the background they land at `#e0e0e0` and `#232323`, which is **1.03:1** and **1.13:1** against the accent fill they are meant to outline. In dark mode the border is actually darker than the fill. The outline would be invisible either way.

`section-message-anatomy.tsx` gets away with `fill-fd-card stroke-fd-border` because that shape is a *surface*, where a hairline is the right register and the fill only needs to differ from the page a little. This one is a button on a container that deliberately paints no surface, so it has nothing to sit against.

**What I used instead:** `stroke-fd-muted-foreground` at `strokeWidth="2"`, which is opaque `#737373` in light (4.35:1) and `#b3b3b3cc` → `#939393` in dark (6.10:1). Clears 3:1 in both modes. Precedent is in the file you cited, `section-message-anatomy.tsx:63-65` draws its close button with exactly that.

```tsx
className="fill-fd-accent stroke-fd-muted-foreground transition-colors duration-300"
strokeWidth="2"
```

Out of scope but worth knowing: the same arithmetic puts the illustration placeholder (`fill-fd-muted-foreground/20`) at 1.34:1 and the description lines (`/40`) at 1.93:1 in dark mode, and those values are copied across every anatomy SVG in the docs, `section-message` included. That is a convention-wide call rather than something to change on one page in a promotion PR.

sarahgm:
Thanks @aromko, all four addressed in 9345cef2d (now pushed). Replies inline, summary here:

**Changeset split** and **the `variant`/`size` wording** both land. The promotion now lives in `.changeset/dst-1427-empty-state-stable.md` naming only `@marigold/components`, the docs rework and dead-link fixes moved to a new `.changeset/dst-1427-empty-state-docs.md` naming only `@marigold/docs`, and the semver sentence is scoped to `title`, `description`, `action` and `headingLevel`. Both the changeset and `release-2026-08-11.mdx` now say explicitly that `variant` and `size` stay open ended and that `theme-rui` defines no values for either today.

Two things I did not take as written.

**The props stay, and narrowing is not the risk.** Bare `variant?: string` is the house style, not an `EmptyState` slip: 15+ components type it identically. And the pattern for adding values later is `'a' | 'b' | (string & {})`, already in 35 files, which widens the accepted set instead of narrowing it. So gaining variants is additive, not breaking.

**The anatomy fix needed different colours than suggested.** Your read on the shape is right, and it turns out to be worse than "dark mode only" - `fill-fd-accent` composites to 1.18:1 in light and 1.34:1 in dark. But `--color-fd-border` is translucent in both themes, so `stroke-fd-border` composites to 1.03:1 (light) and 1.13:1 (dark) *against the fill it is meant to outline*, and in dark mode ends up darker than the fill. I used `stroke-fd-muted-foreground` at `strokeWidth="2"` instead, 4.35:1 and 6.10:1, which is what `section-message-anatomy.tsx:63-65` already uses for its close button.

One thing your comment surfaced that I left alone: the same arithmetic puts the illustration placeholder (`fill-fd-muted-foreground/20`) at 1.34:1 and the description lines (`/40`) at 1.93:1 in dark mode, and those values are copied across every anatomy SVG in the docs. Happy to file it, but it is a convention-wide call rather than something to change on one page in a promotion PR.

aromko (approved):

sebald (approved):

PR #5726, opened 2026-08-10
feat(DST-1700): make --bleed-px the documented container spacing contract

sebald:
# Description

Marigold had two spacing custom properties one hyphen apart that meant different things. `--panel-px` is set per instance on the `<Panel>` root from its padding props. `--ui-panel-px` was a static theme token for the shared anatomy of sectioned overlay surfaces (Dialog, Drawer, Tray, Sidebar) — never the Panel component, a distinction `ui.css` needed a comment to prevent people misreading. Both landed in the same v18 commit. This PR separates them, fixes a defect the overlap caused, and documents the result as public API.

**Renamed** the overlay anatomy out of Panel's namespace: `--ui-panel-px` → `--ui-surface-px`, and `ui-panel-{header,content,actions}` → `ui-surface-*`. They are now named after `ui-surface`, the role those containers already wear (`Dialog.container`, `Drawer.container` and `Tray.container` all apply it), so the utilities read as its parts rather than as Panel's. Dialog, Drawer, Tray and Sidebar are visually unchanged.

**Fixed** `Table`'s edge cell padding, which read `var(--panel-px, var(--bleed-px, var(--cell-x-padding)))`. `--panel-px` is set on the Panel *root*, so it inherits into non-bled content too: the chain always resolved at step one inside a Panel and the other two branches were unreachable. A table in a non-bled `Panel.Content` got 12px edge cells instead of the 10px cell default — offset twice from the Panel border, while the Panel title sat at 12px. It now reads `var(--bleed-px, var(--cell-x-padding))`. Every bleedable container already publishes `--bleed-px`, and because that is declared on the bled element rather than on a container root — and cleared with `initial` on every non-bled one — it is in effect only where edge alignment actually applies. A bled `Card.Content` and `Card.Footer` now publish it too, which they never did.

**Documented** `--panel-*`, `--card-*`, `--page-*` and `--bleed-px` as read-only public API in the spacing foundation, including the `gap-(--panel-gap)` recipe needed when a wrapper interrupts the Panel's flex gap.

Closes DST-1700

Consumer-side follow-up is tracked in DST-1701 (core repo, blocked on the release from this PR).

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

The one path with an intended visual change is `Table` / `Accordion` inside a bled `Card.Content`, which now aligns with the Card title instead of getting no edge alignment at all. In Storybook, `Components/Card/Table Edge Alignment` shows that case and `Components/Panel/Table Edge Alignment` shows the bled and non-bled Panel cases side by side.

## Test Instructions

1. `pnpm build && pnpm test:unit` — 1521 unit tests pass.
2. `pnpm test:sb` — 522 story tests pass, including three browser tests that measure resolved edge padding: `Components/Panel/Table Edge Alignment` (bled and non-bled `Panel.Content`), `Components/Card/Table Edge Alignment` (bled `Card.Content`, the new behaviour) and `Components/Drawer/Bleed Table` (bled `Drawer.Content`).
3. To confirm those tests are load-bearing: restore the old chain in `themes/theme-rui/src/components/Table.styles.ts` (`var(--panel-px, var(--bleed-px, var(--cell-x-padding)))`), run `pnpm build:themes --force`, then `pnpm test:sb --run Panel`. It fails with `expected 12 to be close to 10`. Dropping the `--bleed-px` publish from `DrawerContent.tsx` fails the Drawer test with `expected 10 to be close to 24`.
4. `pnpm sb` → open `Components/Drawer`, `Dialog`, `Tray`, `Sidebar` and confirm the renamed utilities changed nothing visually.
5. `pnpm start` → `/foundations/spacing#reading-container-spacing` for the new docs, linked from the Panel, Card and Page spacing sections.

## Breaking Changes

Yes — for CSS-level consumers on a prerelease only.

`--ui-panel-px` → `--ui-surface-px`, and the `ui-panel-header` / `ui-panel-content` / `ui-panel-actions` utilities → `ui-surface-*`. Rename any override or direct usage.

The token was undocumented and theme-internal (only theme-rui and `DrawerContent` read it), so no component prop or export changes and nothing in the documented API moves — hence the `minor` changeset. It also only ever existed in `18.0.0-rc.*`, never in a stable release, so only prerelease consumers can be affected.

Behavior change: a `Table` in a **non-bled** `Panel.Content` now uses the ordinary cell padding for its first and last cell rather than the Panel's horizontal padding. The docs steer tables to `bleed` and every demo does so, so most tables are unaffected.

## Checklist


packages/components/src/Card/CardContent.tsx
```diff
@@ -16,7 +16,14 @@ export interface CardContentProps {
 export const CardContent = ({ children, bleed }: CardContentProps) => {
   const { classNames } = useCardContext();
   return (
-    <div className={cn(!bleed && 'px-(--card-px)', classNames.content)}>
+    <div
+      className={cn(
+        // `--bleed-px` lets edge-aware children (Table, Accordion) realign with
+        // the Card title while their dividers reach the border. Like Panel.Content.
+        bleed ? '[--bleed-px:var(--card-px)]' : 'px-(--card-px)',
```
aromko: ⚠️ `--bleed-px` inherits into the whole subtree of the bled element, not just its direct children, and no non-bled branch resets it. So a `Table` in a **non-bled** `Card.Content` that itself sits inside a **bled** `Panel.Content` reads the Panel's `--bleed-px` and gets its edge cells offset twice. That is the same defect class this PR fixes for `--panel-px`, one nesting level in, and publishing from `Card.Content` widens the surface.

The changeset's "it is only set where edge alignment actually applies" is a bit stronger than what the CSS does. One token per non-bled branch fixes it (here, plus `PanelContent`, `PanelCollapsibleContent`, `DrawerContent`):

```suggestion
        bleed ? '[--bleed-px:var(--card-px)]' : 'px-(--card-px) [--bleed-px:initial]',
```

`--bleed-px: initial` yields the guaranteed-invalid value, so `var(--bleed-px, fallback)` falls through as intended. If you'd rather leave it, worth softening the claim in the changeset.

packages/components/src/Card/Card.test.tsx
```diff
@@ -168,6 +168,7 @@ describe('Card', () => {
 
       expect(content).toBeInTheDocument();
       expect(content!.className).not.toContain('px-(--card-px)');
+      expect(content!.className).toContain('[--bleed-px:var(--card-px)]');
```
aromko, line 171: ⚠️ DST-1700 Phase 2 asks for edge-padding tests in four positions: non-bled `Panel.Content`, bled `Panel.Content`, bled `Card.Content`, bled `Drawer.Content`. `TableEdgeAlignment.test` covers the first two. This assertion covers that the variable is *declared* on a bled `Card.Content`, but not that a `Table` actually aligns there, and there's no equivalent for bled `Drawer.Content`.

The Card case is the one that matters most: it's the only **new** behavior in this PR, and `Card.stories.tsx:216 WithBleedContent` renders a plain banner `<div>` rather than a Table or Accordion, so Chromatic won't cover it either.

.changeset/panel-spacing-contract.md
```diff
@@ -0,0 +1,24 @@
...
+
+**What changed:**
+
+- The `ui-panel-header` / `ui-panel-content` / `ui-panel-actions` utilities and their `--ui-panel-px` token are now `ui-surface-header` / `ui-surface-content` / `ui-surface-actions` and `--ui-surface-px`. They style the shared anatomy of sectioned overlay surfaces (`Dialog`, `Drawer`, `Tray`, `Sidebar`) — never the `Panel` component, which has its own per-instance `--panel-px`. They are now named after `ui-surface`, the role those containers already wear.
+- `Table` derives its edge cell padding from `--bleed-px` alone: `var(--bleed-px, var(--cell-x-padding))`. It previously read `var(--panel-px, var(--bleed-px, var(--cell-x-padding)))`.
+- A bled `Card.Content` now publishes `--bleed-px` (set to the Card's `--card-px`), which it never did. Non-bled content is unchanged and does not set it.
+- `--panel-px` / `--panel-py` / `--panel-gap`, their `--card-*` and `--page-*` counterparts, and `--bleed-px` are documented as read-only public API under [Reading a container's spacing](https://www.marigold-ui.io/foundations/spacing#reading-a-containers-spacing).
```
aromko: 🔴 Broken anchor. The heading added in `docs/content/foundations/spacing/index.mdx:168` is `### Reading container spacing`, which slugs to `#reading-container-spacing`. The three MDX cross-links in this PR use the correct one. This link ships into `CHANGELOG.md` and the public release notes page.

```suggestion
- `--panel-px` / `--panel-py` / `--panel-gap`, their `--card-*` and `--page-*` counterparts, and `--bleed-px` are documented as read-only public API under [Reading container spacing](https://www.marigold-ui.io/foundations/spacing#reading-container-spacing).
```

packages/components/src/Table/Table.test.tsx
```diff
@@ -17,6 +17,21 @@ const smallScreenQuery = `(width < ${theme.screens!.sm})`;
...
+describe('Edge cell padding', () => {
+  test('derives from --bleed-px only, never from --panel-px', () => {
+    render(<Basic.Component />);
+
+    const table = screen.getByRole('grid', { name: 'label' });
+
+    // `--panel-px` inherits into non-bled content, so reading it would offset
+    // those edge cells twice.
+    expect(table.className).toContain(
+      '[--cell-edge-padding:var(--bleed-px,var(--cell-x-padding))]'
```
aromko, line 29: 💡 This couples a `@marigold/components` test to theme-rui's exact `cva` output, so any whitespace or ordering change in the base array breaks it, and `TableEdgeAlignment.test` already proves the resolved geometry in a real browser. The `not.toContain('--panel-px')` line below is the part carrying the regression value.

docs/content/foundations/spacing/index.mdx
```diff
@@ -164,3 +164,41 @@ For example, consider the `related` token, which is typically used to separate d
...
+  <div className="flex flex-col gap-(--panel-gap)" aria-live="polite">
+    <Panel.Content bleed>…table…</Panel.Content>
+    <Panel.Content>…pagination…</Panel.Content>
+  </div>
+</Panel>
+```
+
+#### Aligning with a bled container
+
+`--bleed-px` is the one you want when content sits edge-to-edge and needs to line back up with the container's title. Any content area rendered with `bleed` publishes it — `Panel.Content`, `Panel.CollapsibleContent`, `Card.Content` and `Drawer.Content` — always holding that container's own horizontal padding.
```
aromko: 💡 `Card.Footer bleed` (`CardFooter.tsx:20`) drops its padding without publishing `--bleed-px`, so it's now the odd one out among the bleedable slots listed here. Either publish it there too for symmetry, or treat footers as deliberately out of scope for edge alignment and leave the list as is.

packages/components/src/Card/CardContent.tsx
sebald: Confirmed and fixed in e5480ec17 — thanks, this was a real leak.

I reproduced it in a browser before changing anything: a `Table` in a non-bled `Card.Content` inside a bled `Panel.Content` resolved to **12px** where the cell default is 10px. `--bleed-px: var(--panel-px)` substitutes at computed-value time on the bled element, so the *resolved* 12px is what inherits down, exactly as you described.

Applied your `[--bleed-px:initial]` suggestion, and extended it to every non-bled branch rather than just this one: `PanelContent.tsx`, `PanelCollapsibleContent.tsx`, `CardContent.tsx`, `CardFooter.tsx` and `DrawerContent.tsx`. Any single one left out would just move the leak.

Also corrected the claim in the changeset **and** the same sentence in `spacing/index.mdx:197`, which had the identical wording — with the reset in place both now describe what the CSS does.

The new `Card` story test covers this case and is load-bearing: reverting just the `CardContent` reset fails it with `expected 12 to be close to 10`.

.changeset/panel-spacing-contract.md
sebald: Fixed in e5480ec17 — applied your suggestion verbatim.

Verified the slug: the heading is `### Reading container spacing` (`spacing/index.mdx:168`), so `#reading-container-spacing`. The four MDX cross-links (card:192, page:102, panel:156, panel:264) were already correct; the changeset was the only one wrong, and as you note it is the one that ships into `CHANGELOG.md`.

packages/components/src/Card/Card.test.tsx
sebald: Fixed in e5480ec17 and 08c4af77b. Checked DST-1700 — Phase 2 does ask for all four positions, and you were right that Card was the gap that mattered.

- **Bled `Card.Content`**: new `Components/Card/TableEdgeAlignment` story tagged `component-test`, measuring resolved edge padding against the Card header. It also renders the nested case from your other comment, so it covers the new behavior and the leak in one story.
- **Bled `Drawer.Content`**: extended the existing `BleedTable` test to assert resolved geometry instead of just that the grid renders. Load-bearing — dropping the publish from `DrawerContent` fails it with `expected 10 to be close to 24`.
- Added two unit assertions in `Card.test.tsx` for the `--bleed-px` declaration on a bled `Card.Footer` and the reset on non-bled content/footer.

`WithBleedContent` still renders the banner `<div>` — I left it as the visual demo and put the Table coverage in the dedicated story.

VRT is still outstanding; not dispatched yet.

docs/content/foundations/spacing/index.mdx
sebald: Confirmed, and I went with publishing it — `Card.Footer bleed` now sets `--bleed-px` in e5480ec17, so all five bleedable slots behave the same.

`--bleed-px` is meant to be container-agnostic, so a footer holding a totals table or an `Accordion` should align like any other slot; treating footers as a special case would be the surprising half. One detail: the docs list you flagged says "content area" and omits `Card.Footer`, so it was not strictly wrong — the asymmetry was in the API. The list now includes the Footer.

(Line reference is `CardFooter.tsx:21` rather than `:20`, no consequence.)

packages/components/src/Table/Table.test.tsx
sebald: Fair point, and factually right: that assertion has to match `themes/theme-rui/src/components/Table.styles.ts:11` character for character, and it needs a theme rebuild to mean anything — a `@marigold/components` test gated on theme-rui's exact `cva` output.

Leaving it in for now rather than deciding it in passing. `TableEdgeAlignment` does prove the geometry in a real browser, so if you want the exact-match line gone and only `not.toContain('--panel-px')` kept, say so and I'll drop it — that's the one line carrying the regression value either way.

Leaving this thread open for that.

aromko (commented):
see vercel storybook

aromko (approved):

PR #5727, opened 2026-08-10
feat(DST-1702): add expandable rows to Table

sebald:
# Description

`<Table>` can now nest rows, so records that belong together — a settlement run and its individual clearings, an order and its line items — live in one table instead of being faked with a button and a menu.

Set `treeColumn` to the `id` of the column that carries the hierarchy and nest `<Table.Row>` inside `<Table.Row>`. Nested rows stay real table rows, which is the point: their values line up under the same headers, each can carry its own actions, and each can be selected in its own right.

The planned "jump to the row with ABR-Nr. X" search is **not** unblocked by this PR. Getting there needs one more piece, moving focus to a given row, which react-aria offers no public API for; that piece and the options for building it are written up in [DST-1713](https://reservix.atlassian.net/browse/DST-1713), together with a working-but-unshippable demo. The docs deliberately say nothing about it rather than half-document it.

Closes DST-1702 · driven by DSTSUP-268

### Three things worth knowing before reviewing

**Most of the state machinery already existed.** `treeColumn`, `expandedKeys`, `defaultExpandedKeys` and `onExpandedChange` are stable props on react-stately's `Expandable`, inherited into RAC's `TableProps` and therefore already reaching RAC through our `{...props}` spread. They worked before this PR — they were just invisible to the docs generator. So the diff is the control, the visuals, the theming, the docs and the tests, not state management.

**The indentation lives on a wrapper inside the cell, not on the cell's padding.** S2 puts it on `padding-inline-start`, but our `cell` style already owns that edge via `px-(--cell-x-padding)` / `first:ps-(--cell-edge-padding)`, which is derived from the `--bleed-px` contract that landed on this branch's base. Overriding it would start a specificity fight and risk regressing the edge-padding tests. The inner wrapper composes instead.

**`TableRow` writes two upstream prop names from one public prop.** react-aria's `useTableRow` reads `hasChildRows` for `aria-expanded`; RAC's `Row` reads `hasChildItems` for the visuals. Each alone gives you exactly one half — verified with a spike. `expandable` writes both, with a comment so it isn't "cleaned up" later, plus a test that fails if either half goes missing.

## Affordance: how a group row reads

The first implementation used indentation plus a vertical guide line. I explored six treatments in a throwaway story (committed at `7258421d1`, then removed) and measured them rather than eyeballing:

- **The guide line is gone.** It used the same `--color-border` as the horizontal row dividers it crossed — 1.37:1 — so it carried exactly the weight of the lines it needed to be distinguished from.
- **Group rows are now filled**, with `bg-foreground/5`. It has to hold one contrast on every ground a row can land on, and an alpha does that where a palette rung doesn't: 1.1039 / 1.1024 / 1.1033 / 1.1003 on `surface`, `background`, inside a `muted` table and under a hover wash — a spread of 0.0036. `--color-muted` was the obvious candidate and is unusable, at 1.04:1 invisible in practice. The fill marks the row *kind*, not the expansion state, so a collapsed group is filled too; that's what tells you there's something to open before you open it. **This started as a new `--color-band` token; see the simplification pass below for why it isn't one.**
- **The control is a 32px ghost icon button** (`ui-button-base` + `hover:ui-state-hover-ghost`), above the 24px WCAG 2.2 SC 2.5.8 floor rather than sitting on it, since it's pressed repeatedly. It is the tallest thing in a group row, so it sets that row's height at every size — see the row-height note below. Reuses `MorphCaret` from `<Accordion>`, so same caret, same 250ms morph, `prefers-reduced-motion` already honoured — no new animation code.
- **One x per level, and the header shares the root's.** The control sits in a leading grid track every row reserves, so a group row and a childless row at the same level align whether or not there's a control to draw. Levels below the root indent by `--tree-indent`. The tree column's *header* takes the same track, so the label sits above the values it heads.

This last point changed late, and the first version shipped it the other way — see below.

## Correction: every level used to share one x

The original design made all levels flush, on the argument that a staggered identifier column is harder to scan and finding an identifier is the point of the feature. Containment was left to the group row's fill.

Sebastian found the cost in review, on `Expandable Rows Dynamic`: `CLR-10240` is a **root-level** row that happens to follow an expanded group's three children. Flush, it is indistinguishable from a fourth child. I had documented that risk in this body as an acceptable edge case; it isn't — a row whose parentage you can't read is worse than a column you scan slightly more slowly.

So `--tree-indent-skip` went 2 → 1 and everything below the root now indents, which is also what S2 does (`(level - 1) * indent`). Separately, the tree column's header now takes the gutter, removing the 40px step between the label and its own values. That one is a deliberate **divergence** from S2, whose `columnStyles` keeps a flat `paddingX: 16` with no level term and therefore still has the step.

Annotated before/after with every coordinate: https://claude.ai/code/artifact/0f3a4a59-c4da-4819-b3bf-588b53a003d0

Measured x at default size — header label, group and root leaf 76; level 2 100; level 3 124. The old story test asserted the flush contract by name, so it is rewritten to assert the new one plus the regression above.

## Simplification pass

Ran after the feature was working, so some of the reasoning above describes a design the diff no longer contains. Three things changed:

**`--color-band` is gone; the fill is `bg-foreground/5`.** The token was justified on needing an alpha rather than a palette rung — which held up, but didn't require a *new* token. `bg-surface/90` two slots up in the same file was already the idiom, and the measurements land within 0.008 of the token (spread 0.0036 vs 0.0029). It also follows the theme's text colour instead of a fixed charcoal rung, so it adapts where the token wouldn't. One less global token.

Worth knowing if you re-check the numbers: my first measurement was wrong. `getComputedStyle` returns `oklch()` unchanged, so three of the four grounds were oklch components parsed as RGB. The figures quoted above are from a canvas re-run, which is why they differ from earlier commit messages.

**The invisible spacer is gone.** Reserving the control's width used to mean rendering an `aria-hidden` span on every row without a control, plus a `twMerge` per render for a string that never changed. `treeIndent` is now a two-column grid, so the fixed track holds the gutter open by itself and `TableCell` just renders `{hasChildItems && <TableExpandButton />}`. One less DOM node per leaf row.

**`--tree-indent-skip` replaces a bare `-2`** inside the indent calc, so "don't indent the first two levels" is a named property like every other number in that file instead of a comment explaining a magic number.

Comments came down by about half in the process. `Table.styles.ts` was 8 comment lines in 195 before this feature and my additions didn't match that; what's left is the part the code can't say — the WCAG threshold, that Tailwind's emit order rather than array order picks the winning background, and the react-aria/RAC prop split.

## Selection does not cascade — and that's deliberate

Selection works with grouped rows, but a group row doesn't select its children and shows no partial state. Verified against the installed RAC rather than assumed: ticking a group leaves children `aria-selected="false"` with the group checkbox `indeterminate: false`, and there's no `indeterminate` anywhere in RAC's Table source.

That matches the model. A nested row here is a real record with its own id and its own actions, so its checkbox already means "this record"; making it also mean "everything underneath" would overload it. AG Grid draws the same line — `groupSelectsChildren` explicitly does not work with their tree data, because there a group "is simply a normal row that has another row as a child". The grids that do cascade (MUI X `rowSelectionPropagation`) apply it to synthesised group headers, which have no other meaning to carry. react-spectrum, on these same primitives, ships neither.

The two features also don't separate: `aria-checked="mixed"` promises that pressing the checkbox resolves its children to all-or-none, so partial state without cascading misdescribes the control.

**Note for reviewers who read the first version of this PR body:** it claimed select-all "covers rendered rows only". That was wrong — inferred from collapsed rows being absent from the DOM. `selectAll()` sets the `'all'` sentinel and `isSelected` then falls through to `canSelectItem`, so select-all *does* cover collapsed nested rows. Fixed in `ac09c86d8`, and both claims are now pinned by tests.

## API naming

Renamed before shipping, while it's still free:

| Instead of | We ship | Why |
| --- | --- | --- |
| `Table.ChildRows` | `Table.ExpandableRows` | Named after the feature, so `expandable`, `Table.ExpandableRows`, the docs section and the stories all use one word. |
| `hasChildRows` | `expandable` | Lands the behaviour on the element that has it, and stays true before anything is loaded. `hasChildRows` asserts the row has children at exactly the moment it doesn't — its only purpose is rows whose children arrive later. |

## Screenshots / Preview

Storybook: `Components/Table` → **Expandable Rows**, **Expandable Rows Dynamic**, **Expandable Rows Lazy Children**. Docs: Table → *Grouped rows*.

| Before | After |
| ------ | ----- |
| Grouping faked with a button + menu; sub-rows unreachable | Nested real rows, filled group row, ghost caret in a shared track |

## Test Instructions

1. `pnpm sb` → `Components/Table` → **Expandable Rows**. Check that group, child and ungrouped rows all start their value at the same x, that both group rows are filled whether open or closed, and that the caret matches `<Accordion>`.
2. Switch `size` between `compact` / `default` / `spacious` — the 32px control should fit all three without clipping.
3. Keyboard: arrow to a group row, <kbd>→</kbd> expands, <kbd>←</kbd> collapses; <kbd>←</kbd> on an already-collapsed child jumps to its parent. <kbd>Tab</kbd> must skip the caret but still reach the per-row PDF buttons.
4. Screen reader: the row should announce expanded state and nesting level (the table is a `treegrid`).
5. Check RTL mirrors the arrow keys, and `prefers-reduced-motion` drops the morph.
6. `pnpm build && pnpm start` → Table docs: the *Grouped rows* demo renders, its group rows show a derived total, and the ungrouped clearing at the bottom reads as root-level rather than as a fourth child.
7. Automated: `pnpm typecheck:only`, `pnpm test:unit packages/components/src/Table` (56), `pnpm test:sb packages/components/src/Table` (62).

## Breaking Changes

**Yes, but only for tables that opt in.** Setting `treeColumn` changes the table's role from `grid` to `treegrid` — that's what makes screen readers announce expanded state and nesting level. Consumer tests asserting `getByRole('grid')` on such a table need `getByRole('treegrid')`. Tables that don't set `treeColumn` are untouched. Called out in the changeset and the docs.

Not breaking, but new public API: `Table.ExpandableRows`, `expandable` on `Table.Row`, the `treeIndent` / `expandButton` theme keys. No new token — the group-row fill is `bg-foreground/5`.

## Regression pass against the Table's existing features

Asked for explicitly in review, since a green suite didn't cover the *combinations*. All measured in Firefox.

| Feature × tree column | Result |
| --- | --- |
| Selection | Checkbox cell and tree column coexist; children get their own checkbox. The group-row fill correctly yields to hover (the `not-data-hovered` guard). |
| Drag & drop | Drag handle, checkbox and chevron all render in one group row; the chevron still works. |
| Inline editing in a nested row | Works. Story added. |
| `compact` / `default` / `spacious` | Control stays 32×32 and the indent stays 24px; only cell padding scales. A group row is therefore 12px taller than a leaf row that carries no control of its own — 33/45, 41/53, 53/65. [Corrected during review](https://github.com/marigold-ui/marigold/pull/5727#discussion_r3776037183); the earlier numbers here were wrong. |
| Panel bleed padding | `--bleed-px` → `--cell-edge-padding` intact at 12px in a bled `Panel.Content`; the control lands on the panel title's x. Measured before the `main` merge against the old `--panel-px` chain, re-verified after it — see below. |

**One real break, fixed.** `Table.EditableCell` implemented `Cell`'s render prop without the tree-column branch `TableCell` has, so putting it in the `treeColumn` dropped the gutter *and the chevron* while the row still reported `aria-expanded` — a row that announces itself as expandable with nothing to press. The cause was altitude: the tree column's anatomy lived inside one of the two components that can occupy it. Extracted to `TableTreeColumn` and rendered from both, which also makes inline editing in the tree column work (S2 does the same, spreading `treeColumnStyles` into `editableCell`).

Three new stories cover the combinations that had none: `ExpandableRowsWithEditableCell` (editable tree column *and* editable plain cell in one row, since the fix had to hold for both), `ExpandableRowsWithDragAndDrop` and `ExpandableRowsInPanel`.

## Drop indicators are level-aware

Ported from S2's `dnd-utils.tsx`: the insertion line starts at the x of the level the row would land at, so a reorder inside an expanded group reads as landing in that group; a root drop still spans the row.

Two things only measurement caught:

- **The collection's `level` is 0-based** (`g1=0, c1=1, gc1=2`) while `aria-level` and `--table-row-level` are 1-based. The first formula assumed 1-based and resolved a one-level-in drop to `0px` — identical to a root drop, i.e. the feature silently doing nothing.
- **RAC sets its own `--table-row-level` on the indicator row**, but at collection level + 2, so it doesn't line up with a row's own value and can't be reused as-is.

Verified per level: `0 / 76 / 100 / 124`, where 76 and 100 are exactly the level-1 and level-2 text positions.

## Out of scope (from the spec, deliberate)

No automatic cross-level sorting, no expand-all control, no elbow connector (needs collection introspection to find the last child), and no encouragement to nest deeper than two levels.

**Moving a row from one level to another is out**, decided deliberately after looking at what it would take. RAC 1.20 does ship a `TreeDropTargetDelegate` and wires it into `Table` (`Table.mjs:425`, `453-458`), so nested drop targets exist for pointer drags. But the **keyboard** path only offers root-level positions — I measured three indicators for a five-row tree, all at `aria-level=1`. Building reparenting on that would ship a feature mouse users can reach and keyboard users can't, which this design system doesn't get to do; closing that gap means a keyboard drop-target delegate or upstream support. Not worth bolting onto this PR.

What *is* here is the level-aware insertion line, which stands on its own: reordering within an expanded group is a nested drop, and the line now shows it landing in the group rather than at the root.

## Merged `main` — and what it changed about the edge padding

`5b5c7f376` merges `main` (v18 release, DST-1700, DST-1691). Two conflicts, both places where `main` and this branch fix the same thing, so neither was a real disagreement.

**`Table.styles.ts` — this branch now uses `main`'s `--cell-edge-padding`, which no longer reads `--panel-px`.** That is DST-1700's improvement and it matters here, because `treeIndent` and `dropIndicator` both resolve their x from `--cell-edge-padding`:

```diff
-'[--cell-edge-padding:var(--panel-px,var(--bleed-px,var(--cell-x-padding)))]',
+'[--cell-edge-padding:var(--bleed-px,var(--cell-x-padding))]',
```

`Panel` sets `--panel-px` on its root, where it inherits into non-bled content too and offsets those cells twice — so reading it was the bug DST-1700 removed. Only a bled container publishes `--bleed-px`. The tree column's alignment survives unchanged because `Panel.Content`, `PanelCollapsibleContent`, `CardContent`, `CardFooter` and `DrawerContent` all republish `--bleed-px` when bled, which is what `ExpandableRowsInPanel` uses.

**One behaviour does change, and it is `main`'s intent rather than a regression:** a tree table in a **non-bled** `Panel` now takes plain cell padding instead of aligning to the panel title. No story covered that case before, so nothing here broke — but if you were expecting the old chain, this is where it went.

Re-verified after the merge rather than assumed: typecheck, Table 56 unit + 60 story (including `The tree column starts at the panel edge padding`), and Panel/Card/Drawer/Accordion 105 unit — all green.

**`vitest.config.shared.ts` — `main`'s fix replaces this branch's.** `bd2b08649` had added a separate `'!packages/cli/**'` entry to `browserScanEntries` for the `chokidar -> fsevents` / `UNLOADABLE_DEPENDENCY` crash; `main` fixes the same thing with a negated glob (`packages/!(cli)/src/**`), which is one entry instead of two and matches the `exclude` list below it. Took `main`'s. Nothing of that commit's content remains, so there is no longer anything here to split into its own PR.

## The `CLI Tests` check, and why the manifest *is* touched

An earlier version of this body said `packages/cli/src/lib/codemod/manifests/v18.ts` was deliberately left alone and that `CLI Tests` was therefore red. `0d83bae` does the opposite — it adds `treeIndent` / `expandButton` to the v18 manifest, and the check is green. The diff is right and the old paragraph was stale; [caught in review](https://github.com/marigold-ui/marigold/pull/5727#discussion_r3776037216).

Adding them is the right call: the codemod's job is to get a consumer onto the newest version of the major, so listing the newest slot set matches the intent, and an extra stub is cheaper than a missing one.

What remains open is the *rule*, not this diff. `theme-drift.test.ts` states that it "pins the manifest of the CURRENT major … when v19 lands, point it at the v19 manifest" — which only covers a major boundary and says nothing about slots arriving in an 18.x minor, the case that actually recurs. **[DST-1718](https://reservix.atlassian.net/browse/DST-1718)** picks that up: decide whether the manifest tracks the latest minor or freezes at `x.0.0`, and make the test say so instead of forcing the next contributor to re-derive it from a red check. DST-1650 would dissolve the question by generating the manifests.

## Split out: the `dependencies` fix

This branch used to carry a second, unrelated changeset — `dependencies` on `Select`, `ComboBox`, `TagGroup`, `TagField` and `Autocomplete`. It now ships on its own in **#5740** (DST-1717), so the collection fix doesn't wait on a Chromatic run for the Table and both diffs stay reviewable. [Raised in review](https://github.com/marigold-ui/marigold/pull/5727#discussion_r3776037196); `6cdcfdc` removes it here, taking this PR from 29 files to 18.

Writing the missing tests for it turned up a second bug in that half: `Autocomplete` shipped a `dependencies` prop you could not use, because its `children` inherited React Aria's render-props type rather than an item render function. Fixed in the split PR.

## The children carry the group, not the parent

Two reviewers independently read the same gap on the preview: [aromko on the drag-and-drop story](https://vercel.com/marigold/marigold-storybook/c/InuxyrqJwqL9?s=15) — *"hard to see what belongs to the sub-rows… I even tried to drag a sub-row first"*. That is the affordance failing at the first thing it has to do.

The measurements agree. The group-row fill meant to carry containment was **1.10:1**, so in practice a 24px indent was the only cue.

**First attempt: a guide line** at the parent's caret. It measured well (1.83:1, clear of the 1.36:1 row divider that sank the *original* guide) and it tested green — but it reads as a stray tick rather than structure, and more importantly it can only mark the tree column. In the drag-and-drop story the handle and checkbox columns sit outside it, which is exactly where the confusion started.

**What shipped instead: the fill moved from the parent to the children.** Nested rows share a band across the full row width, so a group is one contiguous block and a root row that follows it is plainly outside. Because it spans the row, it covers the leading columns that don't indent — the case the guide could never reach.

`foreground/8`, measured at **1.17:1** on `surface`, `background` and inside a `muted` table (spread 0.006), yielding to the hover wash so hovering still reads over it. The group row keeps its caret and semibold label, which is what marks a collapsed group as openable — the fill was never doing that job alone.

Checked across `default`, `grid`, `muted`, `compact`, with selection checkboxes, and at depth 3, where levels 2 and 3 share one band and the indent carries depth.

**One defect the flip exposed, now fixed:** a full-row band outranked `admin` / `master` row colours, so an access-marked child lost its marking. Those backgrounds are now `!` for the same reason their borders already were — who may see a row outranks where it sits in the tree.

Story test `Nested rows are banded, rows at the root are not` pins all three states.

## Group rows are no longer taller than leaf rows

Found while re-measuring for the review. The 32px control was the tallest thing in a group row, so it set that row's height: **41px leaf vs 53px group** at `default`, and the same 12px gap at every size. Invisible in `ExpandableRows`, where every row carries a PDF button of its own; plain in `ExpandableRowsDynamic`, where they don't.

Row height belongs to the `size` variant, so the control gives it up. A negative block margin cancels its surplus over the line box — the same trick the inline direction already uses for `--tree-caret-inset`, which is why the caret can sit at the column's x without pushing it. The control keeps its 32×32 box and its full hit target; only its contribution to layout goes.

Measured after, on a tree whose group and leaf rows differ by nothing but the control:

| | compact | default | spacious |
| --- | --- | --- | --- |
| group row | **33px** | **41px** | **53px** |
| leaf row | **33px** | **41px** | **53px** |
| control | 32×32 | 32×32 | 32×32 |

Pinned by `The expand control does not make its row taller`, which asserts every row in `ExpandableRowsWithDragAndDrop` resolves to one height.

## Other review fixes

- The *Grouped rows* doc linked to `/components/disclosure/accordion`, which 404s — Accordion lives under `navigation`. Fixed, and switched to markdown link syntax like every other internal link.
- Three `querySelector` reaches in `Table.test.tsx` now go through roles (`within(row).getByRole('checkbox')`, and the chevron by its localized "Expand"/"Collapse" name), dropping three `eslint-disable`s. The old version clicked whichever `<button>` came first in the row, which this PR's own drag-and-drop story would have quietly retargeted.
- The `ExpandableRows` demo gave its Actions column `width={80}` while the group row's "Total PDF" button measures 79px. A fixed column doesn't clip, so the button rendered 9px past the table's right border — [spotted on the preview](https://vercel.com/marigold/marigold-storybook/c/_GLD-uOsd19i?s=15). Column widened to 120.

## What the docs actually say now

The *Grouped rows* section came down from about 1200 words to under 400 over review, on the rule that prose before a demo explains intent and UX while the API belongs after it. What's left: what grouped rows are for, that the hierarchy sits on the identifying column, and that a collapsed group row is the only thing most readers see so it needs summary values of its own. One callout survives, for `expandable`, because a row needing a control *before* its children exist is the one thing nobody guesses.

Three things were pulled out rather than trimmed. Selection, because it doesn't cascade and documenting the non-cascade invites the reader to expect it. Row search, because it needs focus movement we don't support (DST-1713). Cross-level moves, cross-level sorting and expand-all, because they read as a list of apologies in a component doc; they're in the changeset instead, where release notes are the right place for "not in this one".

## Checklist

- [x] Visual regression tests updated (for UI changes) — dispatched once the affordance pass and the row-height fix had landed, so the baseline covers the shipped visuals

[DST-1713]: https://reservix.atlassian.net/browse/DST-1713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DST-1718]: https://reservix.atlassian.net/browse/DST-1718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


themes/theme-rui/src/components/Table.styles.ts
```diff
@@ -47,6 +65,11 @@ export const Table: ThemeComponent<'Table'> = {
       'disabled:cursor-not-allowed',
       'data-hovered:cursor-pointer data-hovered:ui-state-hover',
       'dragging:opacity-50 dragging:transform-gpu',
+      // Fills a group row whether open or closed, so the row kind survives
+      // greyscale and print where font weight alone would not. `not-hovered` is
+      // explicit because Tailwind's emit order, not this array's, picks the
+      // winning background.
+      'data-has-child-items:not-data-hovered:bg-foreground/5',
```
aromko: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files but the `Visual-Regression-Tests` workflow has not run for this branch. `gh run list --workflow=visual-regression-tests.yml --branch=feat/DST-1702-table-expandable-rows` returns no runs at all.

This is the one check that would actually cover what this PR is about: the new group-row fill on this line, the indentation, the caret, and the four new snapshotting stories (`ExpandableRows`, `ExpandableRowsDynamic`, `ExpandableRowsInPanel`). The checklist in the PR body already flags it as outstanding.

docs/content/components/collection/table/index.mdx
```diff
@@ -181,6 +181,40 @@ If you cannot place all row actions at the end of the row, be mindful of clarity
...
+  <Do>
+    <DoDescription>
+      Keep the hierarchy shallow. One level of grouping is almost always enough
+      in a data table.
+    </DoDescription>
+  </Do>
+  <Dont>
+    <DontDescription>
+      Don't use grouped rows for content that isn't tabular. Reach for an{' '}
+      <a href="/components/disclosure/accordion">Accordion</a> when the expanded
```
aromko: ⚠️ **Broken link — `/components/disclosure/accordion` 404s.**

There is no `disclosure` category under `docs/content/components/`. The Accordion docs live at `docs/content/components/navigation/accordion/index.mdx`, i.e. `/components/navigation/accordion`.

While you're here: every other internal link in the docs uses markdown link syntax rather than a raw `<a>` plus the `{' '}` spacer (see `docs/content/components/form/radio/index.mdx:78`).

```suggestion
      Don't use grouped rows for content that isn't tabular. Reach for an
      [Accordion](/components/navigation/accordion) when the expanded
      content doesn't share the table's columns.
```

themes/theme-rui/src/components/Table.styles.ts
```diff
@@ -9,6 +9,24 @@ export const Table: ThemeComponent<'Table'> = {
...
+      // The track overlaps it instead of adding padding of its own, so the
+      // caret lines up with the column's values and still gets 32px to hit.
+      '[--tree-caret-inset:calc((var(--tree-chevron-size)-var(--spacing)*4)/2)]',
+      // Caret box plus the button's own trailing whitespace — the distance from
+      // where a row's content starts to where its value starts.
+      '[--tree-gutter:calc(var(--tree-chevron-size)-var(--tree-caret-inset))]',
+      // One level in is one gutter in, so a child's caret starts where its
+      // parent's value does. A smaller step would put it between the parent's
+      // caret and value, which reads as a misalignment rather than depth.
+      '[--tree-indent:var(--tree-gutter)]',
```
aromko, line 25: ⚠️ **`--tree-indent` doesn't scale with `size` — it's a constant 24px.**

All four tree tokens sit in the `table` cva's `base`, never in `variants.size`, and `var(--spacing)` is Tailwind's global 4px:

- `--tree-chevron-size` = `4px * 8` = **32px**
- `--tree-caret-inset` = `(32px - 4px*4) / 2` = **8px**
- `--tree-gutter` = `32px - 8px` = **24px**
- `--tree-indent` = `--tree-gutter` = **24px**, for `compact`, `default` and `spacious` alike.

Two things disagree with that:

1. DST-1702's *Included* scope lists "a `--tree-indent` step that scales with `size="compact" | "default" | "spacious"`".
2. The PR body's regression table claims "Control 32×32 against 45/53/65px rows; **indent 20/24/28px per level**" — that measurement can't come from this code.

The constant indent may well be the better call now that it's tied to the gutter (a size-varying indent would break the "a child's caret starts where its parent's value starts" invariant the comment above defends). But the PR body documents every other deviation from S2 and the spec explicitly, and this one isn't documented — it's asserted the other way. Either move the tokens into the `size` variants, or correct the body and note the deviation.

aromko (commented):
## PR Review: #5727 - feat(DST-1702): add expandable rows to Table

### Overview
| Field | Value |
|-------|-------|
| Author | @sebald |
| Branch | `feat/DST-1702-table-expandable-rows` → `main` |
| Files Changed | 29 |
| Additions | +1806 |
| Deletions | -76 |
| Head SHA | `0d83bae` |
| CI | All 25 checks green (incl. `CLI Tests`, `Typecheck`, `Lint`, `Storybook Tests`, `Unit Tests`) |
| Visual Regression | 🔴 **Not started** |

### Linked Ticket
**[DST-1702](https://reservix.atlassian.net/browse/DST-1702)**: ✨ Expandable rows for Table (tree grid)
- Status: In Review
- Assignee: Sebastian Sebald
- Driven by DSTSUP-268

**Scope coverage** — the ticket's *Included* list is met, with two documented renames (`Table.ChildRows` → `Table.ExpandableRows`, `hasChildRows` → `expandable`; both better, and both cheap to do now) and two intentional deviations that the PR body argues for:
- The vertical guide line was dropped (contrast measured at 1.37:1 against the row dividers it had to be distinguished from) in favour of a `bg-foreground/5` fill.
- Inline editing in the tree column was pulled *in* from the ticket's "Not included (v1)" list, because leaving it out shipped a real break: `Table.EditableCell` in the `treeColumn` dropped the chevron while the row still reported `aria-expanded`. Right call, and the extraction to `TableTreeColumn` is the right shape for it.

One *Included* item is not met — see Issue #3.

### Code Review Checklist

#### TypeScript Standards
- [x] No `any` types introduced (the pre-existing `children?: ReactNode | ((item: any) => ReactNode)` in `ComboBox.tsx` is untouched)
- [x] `import type` used for type-only imports (`CSSProperties`, `ReactNode`, `ReactElement`, `RAC`, `ListBoxProps`)
- [x] Proper interface definitions on every new component (`TableExpandableRowsProps`, `TableTreeColumnProps`, `TableExpandButtonProps`)
- [x] New public props declared explicitly on `TableProps` so they reach the docs generator, rather than relying on the `{...props}` spread

#### Component Patterns
- [x] `className` / `style` stay off every public props interface; `TableRowProps` additionally omits `hasChildItems` so `expandable` is the only way in
- [x] `useClassNames` drives all theming; the two new slots (`treeIndent`, `expandButton`) are registered in `packages/system/src/types/theme.ts` and the v18 codemod manifest
- [x] No z-index classes added to the theme file (`drop-target:before:z-10` stays in `TableDropIndicator.tsx`, per the CLAUDE.md rule)
- [x] Compound-component pattern followed for `Table.ExpandableRows`
- [n/a] `forwardRef` — none of the new components wrap a DOM element the consumer needs a handle on; `TableTreeColumn` and `TableExpandButton` are internal

#### React Best Practices
- [x] Hooks called unconditionally — `useDropLevel` calls `use(DragAndDropContext)` before its early return
- [x] Early returns used (`if (!items) return null`, `if (!isTreeColumn) return content`)
- [x] `treeColumn` correctly added to the `useMemo` dependency array for the table context

#### React Performance
- [x] The simplification pass removed a per-row `aria-hidden` spacer and a `twMerge` per render in favour of a two-column grid track — one less DOM node per leaf row
- [x] Context value memoised with a complete dependency list
- [x] No new dependencies, no bundle impact (`Size Limit` green); `MorphCaret` is reused from `Accordion`

#### Testing Standards
- [x] Vitest throughout, no Jest APIs
- [x] `userEvent` everywhere, no `fireEvent`
- [x] All new stories carry `tags: ['component-test']`
- [x] Unit tests import stories rather than building fixtures (`ExpandableRows.Component`, `ExpandableRowsDynamic.Component`)
- [x] `chromatic: { disableSnapshot: true }` on interaction tests; the four visual stories deliberately keep their snapshots
- [~] Accessible queries — mostly, but three spots reach for `querySelector` where a role query exists (Issue #6)
- [~] `step()` per act-assert cycle — followed everywhere except the new ComboBox test (Issue #7)
- [~] Two of the five components in the `dependencies` changeset ship with no test (Issue #4)

#### Accessibility
- [x] `treegrid` role, `aria-expanded`, `aria-level`, `aria-posinset`, `aria-setsize` all asserted in `Table.test.tsx`
- [x] Arrow-key expand/collapse and parent-jump covered by story tests
- [x] Chevron excluded from the tab order (`tabindex="-1"` asserted), so `Tab` still reaches per-row actions
- [x] 32px hit target, above the 24px WCAG 2.2 SC 2.5.8 floor, with the reasoning left in the theme file
- [x] Focus ring present — `ui-button-base` pulls in `ui-interactive` → `focus-visible:ui-state-focus`
- [x] The role change from `grid` to `treegrid` is called out in the changeset **and** the docs
- [x] Group-row fill yields to hover via the explicit `not-data-hovered` guard, so the hover affordance isn't swallowed
- [x] Reduced motion inherited from `MorphCaret`

#### Correctness spot-checks
Verified against the installed `react-aria-components@1.20.0` rather than taken on trust:
- `props.hasChildItems || state.collection.getItem(...)` (`private/Table.mjs:1039`, `:1203`) — writing `hasChildItems: undefined` for a non-`expandable` row does **not** clobber RAC's auto-detection.
- `DragAndDropContext` is a public export of `react-aria-components/useDragAndDrop`, not a reach into internals.
- The 0-based collection level vs. 1-based `--table-row-level` is handled correctly: `--drop-indent` resolves to `0` / `edge+gutter+indent` / `edge+gutter+2·indent` for collection levels 0/1/2, which matches the row text x at `--table-row-level` 1/2/3.
- React does not append `px` to custom properties, so `{'--drop-level': level}` emits a bare number as the `clamp()`/`max()` maths expects.
- `--table-row-level` is set per row by RAC and absent on the header, which is exactly what leaves the header label at the root's x.

### Issues Found

#### Critical (Must Fix)

**1. `themes/theme-rui/src/components/Table.styles.ts:72` — Visual Regression Tests have never run**

`gh run list --workflow=visual-regression-tests.yml --branch=feat/DST-1702-table-expandable-rows` returns no runs at all.

This is the one check that covers what this PR is actually about: the new group-row fill, the indentation, the caret, and the four new snapshotting stories (`ExpandableRows`, `ExpandableRowsDynamic`, `ExpandableRowsWithEditableCell` is opted out, `ExpandableRowsInPanel`). The PR checklist already flags it as outstanding.

#### Warnings (Should Fix)

**2. `docs/content/components/collection/table/index.mdx:212` — broken docs link**

`<a href="/components/disclosure/accordion">` 404s. There is no `disclosure` category under `docs/content/components/`; the Accordion docs live at `docs/content/components/navigation/accordion/index.mdx`, i.e. `/components/navigation/accordion`.

Every other internal link in the docs also uses markdown link syntax rather than a raw `<a>` plus a `{' '}` spacer (cf. `docs/content/components/form/radio/index.mdx:78`):

```mdx
      Don't use grouped rows for content that isn't tabular. Reach for an
      [Accordion](/components/navigation/accordion) when the expanded
      content doesn't share the table's columns.
```

**3. `themes/theme-rui/src/components/Table.styles.ts:25` — `--tree-indent` doesn't scale with `size`**

All four tree tokens sit in the `table` cva's `base`, never in `variants.size`, and `var(--spacing)` is Tailwind's global 4px:

- `--tree-chevron-size` = `4px * 8` = **32px**
- `--tree-caret-inset` = `(32px - 4px*4) / 2` = **8px**
- `--tree-gutter` = `32px - 8px` = **24px**
- `--tree-indent` = `--tree-gutter` = **24px**, for `compact`, `default` and `spacious` alike

Two things disagree with that:

1. DST-1702's *Included* scope lists "a `--tree-indent` step that scales with `size="compact" | "default" | "spacious"`".
2. The PR body's regression table claims "Control 32×32 against 45/53/65px rows; **indent 20/24/28px per level**" — that measurement cannot come from this code.

The constant indent may well be the better design now that it's tied to the gutter: a size-varying indent would break the "a child's caret starts where its parent's value starts" invariant the comment right above it defends. But this PR documents every other deviation from S2 and the spec explicitly, and this one isn't documented, it's asserted the other way round. Either move the tokens into the `size` variants, or correct the body and record the deviation.

**4. `packages/components/src/Autocomplete/Autocomplete.tsx:136` — `dependencies` ships untested on two of five components**

The `tidy-lists-refresh` changeset covers five components. Three got a `WithDependencies` story plus a `.test()` proving the stale render goes away (`ComboBox`, `Select`, `TagGroup`). `Autocomplete` and `TagField` (`TagField.tsx:78`, forwarded at `:177`) got the prop wired through with nothing exercising it.

These are the harder two, not the easier two: each forwards `dependencies` down *two* paths (popover `ListBox` and tray / `MobileAutocomplete` / `TagFieldDropdown`), which is exactly where a copy-paste pass drops one silently. A story mirroring the ComboBox one would catch it.

**5. `.changeset/tidy-lists-refresh.md:1` — this changeset is a second, unrelated PR**

`fix(Select, ComboBox, TagGroup, TagField, Autocomplete): expose dependencies` touches five components that have nothing to do with expandable rows: 7 source files and 3 story files, on its own changeset, inside a 29-file Table feature. `Table.ExpandableRows` and `Table.Body` get `dependencies` from RAC's collection props; none of them need the collection fix to land.

Splitting it would give both halves a reviewable diff and let the collection fix ship without waiting on a Chromatic run for the Table. Not blocking, but worth saying since the Table half alone is already a long review.

**6. `packages/components/src/Table/Table.test.tsx:534` — `querySelector('button')` where a role query exists**

`collapsed.querySelector('button')!` grabs whichever `<button>` comes first in the row. Today that's the chevron, but the moment this story gains a drag handle or a leading row action the click silently retargets and the test keeps passing while asserting nothing.

The story tests in this same PR already have the accessible version (`chevronOf` in `Table.stories.tsx`), and it drops the eslint-disable too:

```tsx
await userEvent.click(
  within(collapsed).getByRole('button', { name: /expand|collapse/i })
);
```

Same for `group.querySelector('input[type=checkbox]')` on line 500 and line 508 — `within(group).getByRole('checkbox')` reads better and matches the CLAUDE.md preference for accessible queries.

**7. `packages/components/src/ComboBox/ComboBox.stories.tsx:603` — two act-assert cycles, no `step()`**

The test asserts `early`, then switches shift, then asserts `late`, all in one flat body. CLAUDE.md: "Wrap each act-assert cycle in `step()` when a behaviour needs more than one" — which the `Select` version of this exact test and every new `Table` test in this PR do. Without the steps, a failure on the second assertion doesn't say which half broke.

**8. `packages/cli/src/lib/codemod/manifests/v18.ts:171` — the "Known red check" section of the PR body is stale**

The body says this manifest is "deliberately **not** touched: it is the shipped v18 manifest, and this feature lands in 18.1 at the earliest", and asks for a decision on the drift test rather than a fix "smuggled in here". Commit `0d83baec3` then does exactly that, and `CLI Tests` is green.

No objection to the change (a minor adding slots inside the same major makes the manifest listing them correct). But the body should say so, otherwise the next reader hunts for a red check that isn't there and files a follow-up ticket that may no longer be needed.

### Recommendation
**Request Changes**

The feature itself is in good shape and I'd approve it on the code alone. The extraction of `TableTreeColumn` after finding the `EditableCell` break is the right fix at the right altitude, the `hasChildItems` / `hasChildRows` split is handled and pinned by a test that fails if either half goes missing, the drop-indicator level maths is correct (I re-derived it), and the non-cascading selection decision is argued from the installed RAC rather than assumed. Test coverage on the Table side is genuinely good.

Blocking on one thing: **Visual Regression Tests have never run on this branch**, and this is a PR whose entire subject is visual — a new fill on group rows, a new indentation scheme, a new control, four new snapshotting stories. Per the review checklist that's a Critical.

Then the two quick fixes: the 404 docs link (#2), and reconciling `--tree-indent` with the ticket scope and the PR body (#3) — either by moving the tokens into the `size` variants or by correcting the claim.

---

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -127,6 +127,13 @@ export interface AutocompleteProps
    * The value of the input (controlled).
    */
   value?: RAC.ComboBoxProps<object>['inputValue'];
+  /**
+   * Values that invalidate the cached items, like a hook's dependency array.
+   * Only needed for `items` plus a render function: rendered items are cached
+   * per item object, so anything the function reads from outside the item —
+   * state, a lookup, a prop — has to be listed here or it renders stale.
+   */
+  dependencies?: ListBoxProps['dependencies'];
```
aromko: ⚠️ **`dependencies` lands on `Autocomplete` and `TagField` with no test at all.**

The `tidy-lists-refresh` changeset covers five components. Three got a `WithDependencies` story plus a `.test()` that proves the stale-render actually goes away (`ComboBox.stories.tsx`, `Select.stories.tsx`, `TagGroup.stories.tsx`). `Autocomplete` and `TagField` (`TagField.tsx:78`, forwarded at `:177`) got the prop wired through with nothing exercising it.

Both of these are the harder cases, not the easier ones — each forwards `dependencies` down *two* paths (popover `ListBox` and tray/`MobileAutocomplete` / `TagFieldDropdown`), so a missed forward in the mobile branch is exactly the kind of thing a copy-paste pass drops silently. A story mirroring the `ComboBox` one would catch it.

.changeset/tidy-lists-refresh.md
```diff
@@ -0,0 +1,22 @@
+---
```
aromko: 💡 **This changeset is a second, unrelated PR.**

`fix(Select, ComboBox, TagGroup, TagField, Autocomplete): expose dependencies` touches five components that have nothing to do with expandable rows — 7 source files and 3 story files, on its own changeset, inside a 29-file Table feature. `Table.ExpandableRows` and `Table.Body` get `dependencies` from RAC's collection props; none of them need the Select/ComboBox/TagGroup/TagField/Autocomplete work to land.

Splitting it out would give both halves a reviewable diff and let the collection fix ship without waiting on a Chromatic run for the Table. Not blocking if you'd rather keep them together, but worth saying since the Table half alone is already a long review.

packages/components/src/Table/Table.test.tsx
```diff
@@ -419,3 +423,127 @@ test('useTableContext throws outside Table', () => {
...
+    await userEvent.click(
+      screen.getByRole('checkbox', { name: /select all/i })
+    );
+
+    expect(onSelectionChange).toHaveBeenCalledWith('all');
+
+    // July was collapsed while select-all ran.
+    const collapsed = rowOf('Settlement run 1 Jul 2026');
+    // eslint-disable-next-line testing-library/no-node-access
+    await userEvent.click(collapsed.querySelector('button')!);
```
aromko: 💡 `collapsed.querySelector('button')!` grabs whichever `<button>` happens to come first in the row. Today that's the chevron, but the moment this story gains a drag handle or a leading row action the click silently retargets and the test passes while asserting nothing.

The story tests in this same PR already have the accessible version (`chevronOf` in `Table.stories.tsx`), and it drops the eslint-disable too:

```suggestion
    await userEvent.click(
      within(collapsed).getByRole('button', { name: /expand|collapse/i })
    );
```

Same applies to `group.querySelector('input[type=checkbox]')` on line 500 and line 508 — `within(group).getByRole('checkbox')` reads better and matches the CLAUDE.md preference for accessible queries.

packages/components/src/ComboBox/ComboBox.stories.tsx
```diff
@@ -567,3 +567,60 @@ Mobile.test(
...
+              {item.id} — {shift}
+            </ComboBox.Option>
+          )}
+        </ComboBox>
+      </Stack>
+    );
+  },
+});
+
+WithDependencies.test(
```
aromko: 💡 This test runs two act-assert cycles (assert `early`, then switch shift, then assert `late`) in one flat body. CLAUDE.md's story-test guidance: "Wrap each act-assert cycle in `step()` when a behaviour needs more than one" — which the `Select` version of this exact test and every new `Table` test in this PR do.

Without the steps, a failure on the second assertion doesn't say which half broke.

packages/cli/src/lib/codemod/manifests/v18.ts
```diff
@@ -168,6 +168,8 @@ export const v18: MigrationManifest = {
       'footer',
       'row',
       'cell',
+      'treeIndent',
```
aromko, line 171: 💡 **The PR body's "Known red check: `CLI Tests`" section is now stale and argues the opposite of the diff.**

It says `packages/cli/src/lib/codemod/manifests/v18.ts` is "deliberately **not** touched: it is the shipped v18 manifest, and this feature lands in 18.1 at the earliest", and asks for a decision on the drift test rather than a fix "smuggled in here". Commit `0d83baec3` then does exactly that — adds `treeIndent` / `expandButton` here — and `CLI Tests` is green.

No objection to the change itself (a minor adds slots inside the same major, so the manifest listing them looks right). But the body should say so, otherwise the next reader goes looking for a red check that isn't there and a follow-up ticket that may no longer be needed.

docs/content/components/collection/table/index.mdx
sebald: Fixed in `3c6f294`, suggestion taken as written — including the markdown-link form, so the raw `<a>` and the `{' '}` spacer are gone.

Worth noting the category was the whole bug: `disclosure` reads like it *should* be where Accordion lives, which is presumably how it got written. Nothing catches that today — `docs/scripts/validate-links.ts` is being built for exactly this, but it is not on this branch.

packages/components/src/Table/Table.test.tsx
sebald: Fixed in `3c6f294`. All three now go through roles:

- the two checkboxes via `within(group).getByRole('checkbox')`
- the chevron via `within(collapsed).getByRole('button', { name: EXPAND_CONTROL })`, with the same `/expand|collapse/i` constant the story tests use

Three `eslint-disable`s gone with them. The name is real rather than incidental: RAC's `slot="chevron"` supplies the localized "Expand"/"Collapse" label, so this keeps working when the row gains a drag handle — which is precisely the case you flagged, and one this PR already ships in `ExpandableRowsWithDragAndDrop`.

themes/theme-rui/src/components/Table.styles.ts
sebald: You are right on both counts, and the second one goes further than the comment does. Decision: **keep the constant indent, fix the body** — with the deviation written down rather than asserted away.

### Why constant is the right call here

Our indent *is* the gutter (`--tree-indent: var(--tree-gutter)`), and the gutter is the 32px control minus its own whitespace. The control is fixed across sizes on purpose — it is the WCAG 2.2 SC 2.5.8 hit target, and shrinking it at `compact` is exactly where you least want to. So a size-varying indent with a fixed control breaks the invariant the comment above `--tree-indent` defends, in both directions: 20px at `compact` puts a child's caret **left** of its parent's value, 28px at `spacious` puts it **right**. The alignment is the feature; the step size is not.

### Nobody else scales it either

I checked the three implementations worth checking rather than asserting it:

- **S2** — the ticket's own reference implementation — has `--indent: 16`, flat, in `TableView.tsx`. And it is not for lack of a density concept: the same file varies row heights per density (`ROW_HEIGHTS` = 32/40/48 at medium scale) *and* swaps the control size (`density === 'compact' → size = 'S'`). It scales the control and the row, and still leaves the indent term alone.
- **AG Grid** derives the leaf indent from the control, not the row: `.ag-row-group-leaf-indent { margin-left: $icon-size + $cell-horizontal-padding; }` — the same derivation as ours. Its density knobs (`rowHeight`, `headerHeight`) are a separate axis.
- **MUI X** hard-codes it from `rowNode.depth` in `GridTreeDataGroupingCell` with no density term at all; density scales row height by 0.7/1.0/1.3. Customising the indent is [an open feature request](https://github.com/mui/mui-x/issues/6178), not a setting.

So the ticket's "step that scales with `size`" is the outlier, and I would rather deviate from it deliberately than build a size-dependent misalignment to satisfy a bullet.

### The body was wrong twice, not once

Re-measuring to answer this turned up that the neighbouring sentence is off as well. "Control 32×32 against 45/53/65px rows … fits `compact` with 13px of slack" — those three numbers are the **group** row heights, and they are that tall *because* of the 32px control. It does not fit into them; it defines them.

Measured on the preview at `default` (`ExpandableRowsDynamic`, whose leaf rows carry no control of their own):

| | leaf row | group row |
| --- | --- | --- |
| default | **41px** | **53px** |

41 = 20px text + 2×10px padding + border; 53 = 32px control + the same 21px. Since only the padding changes per size (6/10/16), the 12px gap is constant: 33/45 at `compact`, 41/53 at `default`, 53/65 at `spacious`.

That is a real consequence worth its own decision — a group row is 12px taller than its children unless those children happen to carry a 32px control too (which is why `ExpandableRows`, with a PDF button per row, looks even and `ExpandableRowsDynamic` does not). Both the body and this thread now say so instead of implying the opposite. I am folding the question of whether to keep it that way into the affordance pass coming out of the preview comments, since it is the same block of styles.

Body corrected in the next push.

.changeset/tidy-lists-refresh.md
sebald: Agreed, and done: split out to **#5740** (DST-1717). This branch no longer carries any of it — `6cdcfdc` restores the five components and drops `tidy-lists-refresh.md`, so the diff here is 18 files of Table instead of 29.

You were right about the reason, too: the collection fix now ships without waiting on a Chromatic run for the Table, which matters more than usual given the affordance work still coming out of the preview comments.

packages/components/src/Autocomplete/Autocomplete.tsx
sebald: Both now have one, in **#5740** — a `WithDependencies` story plus a `.test()` each, mirroring the ComboBox pair.

Your "these are the harder cases" instinct paid off twice.

**They were verified to be load-bearing.** A test that passes whether or not the prop is wired is worse than no test, so I dropped `dependencies` from each story and confirmed both fail, then put it back and confirmed both pass:

```
× Autocomplete > Re-renders the options when a listed dependency changes
× TagField     > Re-renders the options when a listed dependency changes
```

**And Autocomplete shipped a prop you could not actually use.** Writing the story would not type-check: `AutocompleteProps` inherited `children` from RAC as `ChildrenOrFunction<ComboBoxRenderProps>` — render props, not an item function — so `items` plus a render function, the entire pattern `dependencies` exists for, only compiled through a cast. `ComboBox` hides this because it omits `children` and redeclares it as `(item: any) => ReactNode`.

Fixed in the split PR by giving `AutocompleteProps` an optional item type and declaring the render function on `children`, the shape `TagFieldProps` already uses — no `any`, type-level only, existing static-children uses untouched.

packages/components/src/ComboBox/ComboBox.stories.tsx
sebald: Fixed in **#5740** (this file moved there with the split). Both cycles are now wrapped:

- `The options render with the current outside state`
- `Changing that state re-renders them`

The two new Autocomplete and TagField tests are written the same way from the start, so all five read alike.

packages/cli/src/lib/codemod/manifests/v18.ts
sebald: Right, and thanks for catching it — the body argued against its own diff.

Two changes:

1. **The body is corrected in the next push.** The manifest edit stays; adding the slots is the right call for a codemod whose job is to get a consumer onto the newest version of the major, and `CLI Tests` is green because of it, not in spite of it.
2. **The follow-up is [DST-1718](https://reservix.atlassian.net/browse/DST-1718)**, and it is about the assumption rather than this diff. `theme-drift.test.ts` says it "pins the manifest of the CURRENT major … when v19 lands, point it at the v19 manifest" — a rule that only covers a major boundary and says nothing about slots arriving in an 18.x minor, which is the case that actually recurs. The ticket asks us to decide whether the manifest tracks the latest minor (edit in place, as here) or freezes at `x.0.0`, and to make the test express whichever answer instead of forcing the next contributor to re-derive it from a red check. DST-1650 would dissolve the question by generating the manifests; until then it needs writing down.

aromko (approved):

sebald:
@aromko — following up on your preview comment about the sub-rows: **we switched the colours around.**

Instead of shading the group row, the children now carry the fill. A nested block is one banded stretch across the full row width, and the group row goes back to plain — its caret and bold label are what mark it as openable.

Two reasons it went this way rather than the guide line I first replied with:

- The band covers the **drag handle and checkbox columns**, which don't indent. That is where you reached first, and a line in the tree column could never help there.
- It answers "what belongs to what" as a block rather than as a hairline, so a root row following an expanded group (`CLR-10240`) is plainly outside it.

`foreground/8`, ~1.17:1 on surface, background and muted, yielding to the hover wash. Checked on `default` / `grid` / `muted` / `compact`, with selection, and at depth 3.

It also turned up a bug worth knowing about: a full-row band outranked the `admin` / `master` row colours, so an access-marked child lost its marking. Fixed — those backgrounds are `!important` now, like their borders already were.

Preview is rebuilding on `9f91b29`; Chromatic [build 214](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=214) has the diffs. Happy to reopen the thread if it still does not read for you.

PR #5730, opened 2026-08-11
style(DST-1711): make the undo toast's action button visible without a hover

aromko:
# Description

`addUndoToast` rendered its Undo button as `variant="ghost"` — no fill and no border until it is hovered. Review feedback on the Destructive Actions pattern found it easy to overlook, and the placement makes that worse: the toast arrives far from the control that triggered it and leaves again after five seconds, which is the worst case for an affordance that only shows itself on hover.

It is now the default `secondary` button (`ui-soft` — background plus edge). Deliberately not `destructive`: undo is the safe action, and red beside a completed deletion reads as a second delete. The pattern page gains a paragraph recording the rule so it does not quietly get reverted.

Closes DST-1711

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm start` and open `/patterns/feedback/destructive-actions`.
2. In the "Offering undo" demo, delete a mailing list.
3. Confirm the Undo button in the toast has a visible fill and border before you hover it, and that pressing it restores the row.
4. Check the same toast in dark mode — `ui-soft` on a dark surface is the case worth a second look.

## Breaking Changes

No

## Checklist

### DIFF
--- a/packages/components/src/Toast/ToastQueue.tsx
+++ b/packages/components/src/Toast/ToastQueue.tsx
@@ -174,10 +174,6 @@ export function useToast() {
         action: (
           <Button
             size="small"
-            // Filled, not `ghost`: the toast sits far from the button that
-            // triggered it and leaves after five seconds, so the way back has
-            // to be visible without a hover. Not `destructive` either — undo is
-            // the safe action, and red would read as a second delete.
             variant="secondary"
             // Toasts stack, so a bare "Undo" is ambiguous to a screen reader.
             aria-label={stringFormatter.format('undoNamed', { title })}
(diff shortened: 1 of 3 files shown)


packages/components/src/Toast/ToastQueue.tsx
```diff
@@ -174,7 +174,11 @@ export function useToast() {
         action: (
           <Button
             size="small"
-            variant="ghost"
+            // Filled, not `ghost`: the toast sits far from the button that
```
sebald: do we need this long comment if it is already in the mdx? :D

aromko: Time to try a new skill :-D

aromko: First try it recommmends to keep.
Second try with the hint to the mdx -> remove

sebald (approved):

PR #5731, opened 2026-08-13
docs(DST-1703): migrate anatomy diagrams to theme-aware inline SVG

OsamaAbdellateef:
# Description

Replaces the 24 raster `.jpg` anatomy diagrams and 3 static `.svg` assets with hardcoded hex fills (`FileField`, `TagField`, `TopNavigation`) with hand-authored inline SVG components. Also removes 3 orphaned assets (`drawer`, `table`, `selectlist`) left over from earlier migrations. `docs/public/**/*-anatomy.*` no longer exists.

The old diagrams had one baked-in color scheme, so in dark mode they rendered as a bright block with low-contrast labels. The new inline SVGs consume Fumadocs theme tokens (`fill-fd-card`, `stroke-fd-border`, `fill-fd-foreground`, `fill-fd-primary`) with `transition-colors duration-300`, so they follow the theme toggle, stay sharp on HiDPI displays, and are reviewable as source rather than as a binary blob.

This is a format migration, not a redesign — diagrams depict the same parts under the same labels as before. `LinkButton` now imports `ButtonAnatomy` from the Button page instead of duplicating it.

Closes DST-1703

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm start` and open the docs site.
2. Visit each updated component page's Anatomy section (Badge, List, Tag, Autocomplete, Calendar, Checkbox, DateField, DatePicker, FileField, NumberField, Radio, SearchField, Slider, TagField, TextField, TimeField, Accordion, Breadcrumbs, Pagination, Tabs, TopNavigation, ContextualHelp, Dialog, Toast, Tooltip, Link, LinkButton) and confirm the SVG renders correctly.
3. Toggle dark/light mode and confirm the diagrams follow the theme instead of showing a fixed-color block.

## Breaking Changes

No

## Checklist

- [ ] Storybook preview and Marigold docs preview are available
- [ ] Stories added/updated (with `component-test` tag where applicable)
- [ ] Unit tests added/updated
- [x] Component documentation added/updated (if it exists)
- [ ] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) (for new/changed interactive components)
- [ ] Visual regression tests updated (for UI changes)
- [x] Changeset added (`pnpm changeset`)

PR #5733, opened 2026-08-13
docs(DST-1703): migrate anatomy diagrams for actions, content and collection

OsamaAbdellateef:
# Description

Part of the DST-1703 anatomy migration, split out of #5731 so each group can be
reviewed on its own. This one covers **Link, LinkButton, Tag, Badge and List**.

Each page's anatomy diagram moves from a raster image to a hand-authored inline
SVG that consumes Fumadocs theme tokens (`fill-fd-card`, `stroke-fd-border`,
`fill-fd-foreground`, `fill-fd-primary`). The old images carried one baked-in
color scheme, so in dark mode they rendered as a bright block in the middle of a
dark page and the labels lost contrast. The new ones follow the theme toggle,
stay sharp on HiDPI displays, and are reviewable as source rather than as a
binary blob.

The diagrams depict the same parts under the same labels as the images they
replace — this is a format migration, not a redesign.

Following [sebald's review of #5731](https://github.com/marigold-ui/marigold/pull/5731), every annotated part is now described in a
list below the illustration, matching the pattern already used on the Button
page.

Pages touched: Link, LinkButton, Tag, Badge, List. `LinkButton` imports
`ButtonAnatomy` from the Button page rather than duplicating it, which is what
the shared `button-anatomy.jpg` was already expressing. Also drops two orphaned
assets left over from earlier migrations (`button`, `table`).

Closes DST-1703

## Screenshots / Preview

Docs preview (deploy link appears below once the build finishes). Worth
toggling light/dark on each page — that is the point of the change.

## Test Instructions

1. Open the docs preview and visit each page listed above.
2. Check the `## Anatomy` section: the diagram renders, and every label in it
   has a matching bullet underneath.
3. Toggle the theme. The diagram's fills and strokes should follow, with no
   bright block left over in dark mode.

## Breaking Changes

No

## Checklist


docs/content/components/actions/link/link-anatomy.tsx
```diff
@@ -0,0 +1,55 @@
+export const LinkAnatomy = () => (
+  <svg
```
aromko, line 2: ⚠️ The `<svg>` has no accessible name. The `<ImageZoom>` this replaces carried `alt="Anatomy of a link"`; without `role="img"` + `aria-label`, an inline SVG is exposed as a graphics document and a screen reader reads its bare `<text>` children ("Label") with no context.

Six already-migrated anatomy components set this — see `docs/content/components/collection/table/table-anatomy.tsx:3-4`, plus `card-`, `empty-state-`, `keyboard-`, `menu-` and `segmentedcontrol-anatomy.tsx`.

```suggestion
  <svg
    role="img"
    aria-label="Anatomy of a Link: an underlined Label naming the destination"
```

Same issue in `tag-anatomy.tsx`, `badge-anatomy.tsx` and `list-anatomy.tsx`. `button-anatomy.tsx` (reused by the LinkButton page from this PR on) is missing it too, if you want to close the gap in one go.

docs/content/components/collection/tag/tag-anatomy.tsx
```diff
@@ -0,0 +1,102 @@
+export const TagAnatomy = () => (
+  <svg
```
aromko, line 2: ⚠️ No accessible name — same issue as `link-anatomy.tsx`. This one loses the most, since it has four `<text>` labels that a screen reader will read as loose words.

```suggestion
  <svg
    role="img"
    aria-label="Anatomy of a Tag group: a Group label above a Tag group of tags, each with a Tag label and a Remove button"
```

docs/content/components/content/badge/badge-anatomy.tsx
```diff
@@ -0,0 +1,84 @@
+export const BadgeAnatomy = () => (
+  <svg
```
aromko, line 2: ⚠️ No accessible name — same issue as `link-anatomy.tsx`.

```suggestion
  <svg
    role="img"
    aria-label="Anatomy of a Badge: a pill-shaped Container holding either a Text or an Icon"
```

docs/content/components/content/list/list-anatomy.tsx
```diff
@@ -0,0 +1,74 @@
+export const ListAnatomy = () => (
+  <svg
```
aromko, line 2: ⚠️ No accessible name — same issue as `link-anatomy.tsx`.

```suggestion
  <svg
    role="img"
    aria-label="Anatomy of a List: a List container holding several List items"
```

docs/content/components/collection/tag/index.mdx
```diff
@@ -3,19 +3,20 @@ title: Tag
 description: Used to manage related options.
 ---
 
+import { TagAnatomy } from './tag-anatomy';
+
 Tags help users explore, filter, or manage related options in a compact and accessible format. They are arranged in a container where they can be interacted with as a group. Tag groups are commonly used in search interfaces, content management systems, or anywhere users need to explore, refine, or manage categories and topics.
 
 ## Anatomy
 
 A tag group consists of a group label, a field label, individual tags with tag labels, and optional remove buttons for each tag.
```
aromko: 💡 This sentence names five parts ("a group label, **a field label**, individual tags with tag labels, and optional remove buttons"), but the new list below the diagram documents four — there is no "field label" bullet and no annotation for it.

`TagGroup` exposes a single `label` prop (`packages/components/src/TagGroup/TagGroup.tsx:45-47`, documented as "The label of the field"), so group label and field label are the same thing. Since the point of the bullet list is that every part gets described, dropping ", a field label" from the sentence resolves it.

```suggestion
A tag group consists of a group label, individual tags with tag labels, and optional remove buttons for each tag.
```

aromko (commented):
## PR Review: #5733 - docs(DST-1703): migrate anatomy diagrams for actions, content and collection

### Overview

| Field | Value |
|-------|-------|
| Author | @OsamaAbdellateef |
| Branch | `dst-1703_anatomy-actions-content` → `main` |
| Files Changed | 16 |
| Additions | +354 |
| Deletions | -35 |
| CI | All 24 checks passing |
| Visual Regression | Not needed (docs-only; no `packages/components/src`, `packages/system/src`, `themes/**`, `.storybook/**`, `*.stories.tsx` or `*.styles.ts` touched) |

### Linked Ticket

**[DST-1703](https://reservix.atlassian.net/browse/DST-1703)**: 📝 Migrate component anatomy diagrams from raster images to theme-aware inline SVG
- Status: In Review
- Scope: 24 raster `.jpg` + 3 hardcoded-hex `.svg` anatomy assets → inline `*-anatomy.tsx`, plus 3 orphaned assets to delete
- This PR covers 5 of them: Link, LinkButton, Tag, Badge, List

### What I verified

- Rendered all four new SVGs plus the reused `ButtonAnatomy` side by side in light and dark, and compared them against the JPGs pulled from the base commit. Same parts, same labels, same layout — the "format migration, not a redesign" claim holds for all five pages.
- Theme tokens only (`fill-fd-card`, `stroke-fd-border`, `fill-fd-foreground`, `fill-fd-primary`, `fill-fd-muted-foreground/25`), no hardcoded hex, `transition-colors duration-300` throughout, `viewBox` + no fixed `width`/`height` — matches the pattern the ticket prescribes.
- No dangling references to the six deleted assets. `docs/public/button/button-anatomy.jpg` was still in use by the LinkButton page until this PR removed it; `table-anatomy.jpg` was already orphaned by the earlier Table migration.
- `@marigold/docs` is private but versioned and routinely gets changesets (#5712, #5708), so the `patch` changeset is right.

### Code Review Checklist

#### TypeScript Standards
- [x] No `any` types
- [x] Proper interface definitions — n/a, all four components are zero-prop
- [x] `import type` — n/a, no type imports

#### Component Patterns
- [x] `forwardRef` — n/a, no DOM-wrapping component props
- [x] `isDisabled`/`isPending` renaming — n/a
- [x] Does not expose `className`/`style` props
- [x] `useClassNames` — n/a, these are docs-site SVGs styled with Fumadocs theme classes, which is the established pattern for `*-anatomy.tsx`

#### React Best Practices
- [x] Proper hook usage — n/a, no hooks
- [x] Named exports, PascalCase components in kebab-case docs files (matches sibling anatomy files)

#### React Performance
- [x] Static, prop-less, render-once SVGs — no memoization, data fetching or bundle concerns. The five deleted JPGs (~275 KB) are replaced by ~315 lines of inline markup, a net win for the docs bundle.

#### Testing Standards
- [x] n/a — no stories or tests in this PR, and none are expected for docs content

#### Accessibility
- [ ] The four new SVGs have no accessible name — see Warnings
- [x] Keyboard navigation — n/a, non-interactive illustrations
- [x] Every annotated part now has a matching text bullet, which is the improvement sebald asked for on #5731

### Issues Found

#### Critical (Must Fix)

None.

#### Warnings (Should Fix)

- `docs/content/components/actions/link/link-anatomy.tsx:2`, `collection/tag/tag-anatomy.tsx:2`, `content/badge/badge-anatomy.tsx:2`, `content/list/list-anatomy.tsx:2` — **the SVGs have no accessible name.** The `<ImageZoom>` blocks they replace each carried an `alt` ("Anatomy of a link", "Anatomy of a tag", …); an inline `<svg>` with no `role`/`aria-label` is exposed as a graphics document and a screen reader reads its bare `<text>` children instead. On the Tag page that is "Group label / Tag label / Remove button / Tag group" with no context. Six already-migrated components do it right — `collection/table/table-anatomy.tsx:3-4`, `card-`, `empty-state-`, `keyboard-`, `menu-`, `segmentedcontrol-anatomy.tsx` — so add `role="img"` + `aria-label` to match. Worth doing for `button-anatomy.tsx` in the same pass, since the LinkButton page now renders it too.

- **PR description — "Closes DST-1703" is premature.** The body itself says this is one group split out of #5731, and 23 anatomy assets are still in `docs/public/` on this branch, including two of the three orphans the ticket names for deletion (`drawer/drawer-anatomy.jpg`, `selectlist/selectlist-anatomy.jpg`). Change it to "Part of DST-1703" so the ticket does not read as done.

- `docs/content/components/collection/tag/index.mdx:12` — the intro sentence names five parts including "a field label", but the new bullet list documents four. `TagGroup` exposes a single `label` prop (`packages/components/src/TagGroup/TagGroup.tsx:45-47`, "The label of the field"), so group label and field label are the same thing. Dropping ", a field label" makes the prose and the list agree.

### Recommendation

**Comment**

Clean, well-scoped migration that does exactly what it says: I diffed the rendered output against the originals and nothing was redesigned, the tokens and structure follow the established pattern, and CI is green across the board. The accessible-name gap is the one thing I would fix before merge — it is a regression against the `alt` text these pages had, and the repo already contains the fix pattern. The rest are nits.

aromko (commented):
@OsamaAbdellateef maybe in general. Some picture lables looks bigger than others, e.g. LinkButton and ToggleButton. Can we achieve more consistency there? I would prefer the ToggleButton "size" as base. I checked a little but and it isn't the fontSize. Maybe it is possible to have the same viewbox everywhere?

OsamaAbdellateef:
Applied suggested solutions :heavy_check_mark:  @aromko

OsamaAbdellateef:
@aromko Good catch, and you were right that it isn't the `fontSize` — every diagram already uses `fontSize="14"`. The cause is the **viewBox width vs. the `max-w` fraction**: the SVGs scale to the docs column, so

```
label px = 14 × (column width × max-w fraction) ÷ viewBox width
```

Those combinations had drifted apart, so the same `14` meant something different on every page. Measured in the real 836px docs column:

| Diagram | viewBox / max-w | Label rendered |
| --- | --- | --- |
| Link | 260 @ 50% | 22.5px |
| Badge | 500 @ 90% | 21.1px |
| **LinkButton / Button** | 545 @ 90% | **19.3px** |
| **ToggleButton** | 660 @ 90% | **16.0px** |
| Panel / Drawer / Filter | 820 @ 90% | 12.9px |
| Select | 960 @ 100% | 12.2px |
| Table | 950 @ 90% | 11.1px |

A 2× spread — which is exactly the LinkButton vs. ToggleButton difference you spotted.

So yes to your suggestion: **all 32 anatomy diagrams now use ToggleButton's frame**, `viewBox` width `660` at `max-w-[90%]`. That pins the scale at 733.33 user units per content column, so `fontSize="14"` renders identically everywhere. Geometry was rescaled by the inverse factor, so **every drawing keeps the on-screen size it had before** — only the annotations change. `fontSize`, `strokeWidth="2"` and the `r="4"` connector dots are deliberately left unscaled, which is what makes them consistent across pages.

Verified at the real column width: **all 139 labels across the 32 diagrams now render at 15.96px** (was 11.1–22.5px), with no clipping and no label collisions.

Where the bigger labels no longer fit, I let the drawing give way rather than the frame:

- `select-anatomy` — drawn 14% smaller, plus elbows on the *Section header* / *Item description* connectors, and *Selected value* pulled left off the popover edge
- `list-anatomy` — drawn 6% smaller so *List item* clears the frame
- `table-anatomy` — viewBox extended upward (vertical padding doesn't affect the scale)
- `panel-anatomy` / `panel-collapsible-anatomy` — *Description* routed below *Title*

I also wrote the rule down in `.claude/skills/component-docs-writer/references/anatomy-diagrams.md` so this doesn't drift again: a diagram that needs more room gets **drawn smaller**, never a wider viewBox — a wider viewBox silently shrinks that page's labels relative to every other page.

Note this touches anatomy diagrams outside this PR's original scope (form, layout, overlay, navigation, patterns), since consistency only holds if all of them share the frame. Happy to split that into a follow-up PR if you'd rather keep this one narrow.

aromko (approved):

PR #5742, opened 2026-08-14
ci(DST-1721): validate docs links and fix the 49 broken ones

sebald:
# Description

Nothing checked that links in `docs/content` resolve. The first run of [`next-validate-link`](https://next-validate-link.vercel.app) found **52 errors across 23 files** — 49 of them real breakage that ships today. This PR adds the check and fixes every finding, so the new status check is green from day one rather than red on every docs PR.

**The class of bug nobody could have spotted by reading the MDX.** `createRelativeLink` → `source.resolveHref` resolves a `../`-style href by **file path**, calling `indexer.getPage('components/layout/panel')` — which misses, because the real key is `.../panel/index.mdx`. On a miss it returns the href **unchanged**, so the browser resolves it against the URL instead. `curl https://www.marigold-ui.io/components/collection/card` ships literal `href="../../layout/panel"` right now, and from `/components/collection/card` that lands on `/layout/panel`. 17 of these are on the Card page alone.

The rest are ordinary rot from page moves nobody chased down: Card `content/` → `collection/`, ActionBar `actions/` → `collection/`, ActionMenu split out of the Menu page, plus a `swich` typo and a `[…](<(/…)>)` double-paren.

| | count | fix |
| --- | --- | --- |
| Relative links Fumadocs leaves unresolved | 22 | rewritten as absolute, matching the convention the other ~600 links already use |
| Absolute links to pages that don't exist | 24 | repointed at the real page |
| Anchors to renamed headings | 3 | repointed (e.g. `menu#actionmenu` → `/components/overlay/action-menu`) |
| Trailing slash on a real page (works, via a 308) | 3 | slash dropped |

A second commit closes a blind spot the checker has: links written as `https://www.marigold-ui.io/...` are external as far as it's concerned, so they were never validated. **5 of the 15 in `content` were dead** — `/introduction/get-in-touch` (×2), a stale `?theme=b2b` query, plus two others. Rewriting them root-relative both fixes them and puts them under the checker for good, and it fixes a second bug: an absolute production URL in a preview deploy sends the reader to production instead of the preview. URLs that *are* the content stay absolute — the `/mcp` endpoint, the `manifest.json` and `.md` examples, the `MARIGOLD_DOCS_URL` default, and anything inside a code block.

Closes DST-1721

## Test Instructions

1. `pnpm install`
2. `pnpm --filter @marigold/docs lint:links` → `0 errored file, 0 errors`
3. Break something on purpose — e.g. change a link in `docs/content/components/collection/card/index.mdx` to `/components/layout/nope` — and re-run. It should report `not-found` with file, line and column, and exit 1.
4. Optional, on the docs preview: open **Card** and click through to Panel, Stack, Tiles, Inline, Table, SelectList and Title. On `main` those hit the not-found page.

## Breaking Changes

No — docs content and CI only, no published package touched.

## Notes for the reviewer

- **Why its own workflow and not a step in `build.yml`.** `build.yml` answers "do the packages build" and runs `pnpm build`; `lint.yml` and `typecheck.yml` also build packages first. This check needs none of that — it only reads `docs/content` and the App Router tree, so the whole job is install + **3.6s** (35s in CI). `paths`-filtered to `docs/content/**`, `docs/app/**`, the script, and the workflow file itself.
- **Why `on: push` and not `pull_request`** (third commit). It started as `pull_request`, like `react-aria-dedupe.yml`. Then the push of the second commit produced **no `pull_request` run at all** — not for this workflow, and not for Builds, CodeQL, Labeler or the dedupe guard either. Closing and reopening the PR to fire `reopened` produced nothing either, and GitHub's status API reported no incident. Two dropped activity types is enough to stop relying on the trigger: a check that silently doesn't run is worse than no check, because a stale green is indistinguishable from a fresh one. `Lint`, `Typecheck` and `Test` are all `on: ['push']` and fired for every commit here, so this now follows them — and it verified itself, since the paths filter covers the workflow file, so the commit that made the change exercised it. Same gap those three have: `push` doesn't fire for fork PRs, where Builds and CodeQL stay the only coverage.
- **Heads-up for the reviewer:** because of the above, the `Builds`, `CodeQL`, `Labeler` and `react-aria dedupe guard` checks shown on this PR are from the **first** commit only. They're `pull_request`-only, so nothing on this branch can refresh them. Treat them as unverified for the two later commits — though neither touches anything they cover (three MDX files and a workflow trigger).
- **Why a script and not a CLI step.** `next-validate-link@1.6.7` ships no `bin` — a script is the intended integration. `preset: 'next'` discovers the static routes; the only thing it can't infer is `(docs)/[...slug]`, populated from `content/**/*.mdx` via `getSlugs`, with per-page `hashes` from `getTableOfContents` so fragments are checked too.
- **Editorial calls worth a second opinion.** `/foundations/validation` (4 links) has no page any more → repointed at `/patterns/user-input/forms#validation`. `<MultiSelect>` has no page at all → the 2025-03-19 blog post now says it was superseded by [TagField](https://www.marigold-ui.io/components/form/tag-field) instead of linking into the void. The 2026-03-25 post's "App Shell pattern" is really `/patterns/layout/app-frame`, so the link text changed with it. Combobox said `useAsyncList`; the hook Marigold actually ships is `useAsyncListData`, so prose and link both changed.
- **Known gaps, deliberately out of scope.** Hrefs passed as props to `TeaserList`/`DemoLinks`/`GuidelineTiles` aren't scanned (`markdown.components` can opt them in later). The 12 generated, gitignored `content/releases/**/release.mdx` files are absent in clean CI, so their links go unchecked — covering them needs `pnpm build:changelog` and a `GH_TOKEN`. External links aren't checked, which keeps CI independent of third-party uptime.
- **Side finding: unknown docs URLs return HTTP 200, not 404.** Narrower than it first looks, and deliberately *not* fixed here. `notFound()` fires after the root layout's Suspense boundary has flushed the shell, and Next can only set 404 on a non-streamed response — documented behaviour, not a misconfiguration. It is **not** an SEO problem: Next injects `<meta name="robots" content="noindex">`, which is live on the response today. It is also **not** a blank page — the 404 UI client-renders on hydration (verified in a browser; `curl` only shows the SSR shell, which is what made it look blank). What remains is the status code itself, which matters for analytics, uptime checks and scripts. The documented fix is `export const dynamicParams = false` on the catch-all, so unknown slugs 404 at the routing layer before anything streams. Recorded on DST-1721 rather than folded in here, to keep a routing change out of a content PR.
- **Second side finding.** The `changed:` frontmatter on release blog posts isn't read by any component — dead data.
- **Verified:** `lint:links` clean, `prettier --check` clean on all 27 files, `eslint` clean, the script typechecks under the docs compiler options, and `CI=true pnpm install --frozen-lockfile` succeeds against the committed lockfile. I did **not** run the full `pnpm build` / `types:check` on this branch — packages aren't built here and nothing in this PR touches TypeScript that they'd affect.
- The lockfile diff is the new devDependency plus pnpm 11.8's own re-normalisation of some `eslint-import-resolver-typescript` peer keys. An unrelated `ws` bump that `pnpm install` drifted in was reverted.

## Checklist

- [ ] Storybook preview and Marigold docs preview are available — verify once CI builds them; the docs preview is where step 4 above is worth clicking
- [ ] Unit tests added/updated — N/A, the docs vitest project is deliberately node-only; the check *is* the test, and it runs in CI
- [x] Component documentation added/updated (if it exists) — 22 content files corrected


docs/content/components/form/combobox/index.mdx
```diff
@@ -46,7 +46,7 @@ This can be achieved by wrapping the options in the `<ComboBox.Section>` compone
 
 ### Working with asynchronous Data
 
-The ComboBox component supports working with asynchronous data. In the example below, the [`useAsyncList`](/hooks/useAsyncList) hook is used to handle the loading and filtering of data from the server.
+The ComboBox component supports working with asynchronous data. In the example below, the [`useAsyncListData`](/components/hooks-and-utils/useAsyncListData) hook is used to handle the loading and filtering of data from the server.
```
aromko: This renames the hook in prose from `useAsyncList` to `useAsyncListData`, but `useAsyncListData` is not exported by anything in this monorepo. `packages/components/src/hooks.ts:1` has `export { useAsyncList } from 'react-aria-components/useAsyncList'`, re-exported at `packages/components/src/index.ts:7`; grepping `packages/` and `themes/` for `useAsyncListData` returns nothing. The demo rendered immediately below this sentence, `docs/content/components/form/combobox/combobox-async.demo.tsx:1`, does `import { ComboBox, useAsyncList } from '@marigold/components'`. Every other page that links this URL keeps the correct name: `patterns/feedback/async-data-loading/index.mdx:28` reads "[useAsyncList documentation](/components/hooks-and-utils/useAsyncListData)" and `getting-started/recommended-libraries/index.mdx:43` reads "[`useAsyncList`](/components/hooks-and-utils/useAsyncListData)". Fix the href, keep the name. Separately worth a ticket: the target page itself is misnamed and `docs/content/components/hooks-and-utils/useAsyncListData/index.mdx:16` tells readers to `import { useAsyncListData } from '@marigold/components'`, which does not resolve.

docs/content/components/layout/panel/index.mdx
```diff
@@ -51,12 +51,12 @@ Every Panel separates from the page by fill, not by a shadow: a white surface ag
...
-| `default`     | Subtle surface tint.              | Most page sections: profile, notifications, billing.                               |
-| `master`      | Tint and accent for master users. | Sections only master users see. [Learn more.](../../../patterns/admin-master-mark) |
-| `admin`       | Tint and accent for admin users.  | Sections only admin users see. [Learn more.](../../../patterns/admin-master-mark)  |
-| `destructive` | Tint for danger zones.            | Irreversible actions, like deleting an account or revoking access.                 |
+| Variant       | Description                       | When to use                                                                       |
+| ------------- | --------------------------------- | --------------------------------------------------------------------------------- |
+| `default`     | Subtle surface tint.              | Most page sections: profile, notifications, billing.                              |
+| `master`      | Tint and accent for master users. | Sections only master users see. [Learn more.](/patterns/layout/admin-master-mark) |
```
aromko, line 57: This row is corrected to `/patterns/layout/admin-master-mark`, but line 359 of this same file still has `href: '/patterns/admin-master-mark'` inside `<TeaserList>`, and it 404s in production today (I confirmed with curl that the live Panel page ships both `href="/patterns/admin-master-mark"` and `href="../../content/card"`). Same file, lines 345 and 352: `../../content/card` resolves to `/content/card` and `../../navigation/accordion` resolves to `/navigation/accordion`, while line 84 of this PR fixes the identical Card target in the Callout. The other four in files this PR touched: `card/index.mdx:306` (`../../form/selectlist#options-as-cards`, fixed in prose at line 81), `card/index.mdx:328` (`../../layout/panel`, fixed in prose at line 52), `feedback-messages/index.mdx:244` (`../../foundations/validation`, the exact dead target this PR retires at line 109). Nine more of the same class sit elsewhere: `actionbar/index.mdx:166,189`, `timefield/index.mdx:71`, `formatters/numericformat/index.mdx:71`, `component-principles/index.mdx:264`, `fetching-and-mutations/index.mdx:241,247,252`, `loading-states/index.mdx:132`. The PR notes `markdown.components` can opt these in later, but it cannot: next-validate-link@1.6.7's default `onNode` skips any attribute where `typeof attr.value !== 'string'`, and `TeaserList` passes `items={[...]}` as an expression. Closing this needs a custom `markdown.onNode` walking `mdxJsxAttributeValueExpression`, or just fixing the 15 hrefs.

.github/workflows/validate-links.yml
```diff
@@ -0,0 +1,37 @@
+name: Docs links
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'docs/content/**'
+      - 'docs/app/**'
+      - 'docs/scripts/validate-links.ts'
+      - '.github/workflows/validate-links.yml'
```
aromko, line 15: The `paths` filter covers content, the App Router tree and the script, but not `docs/package.json`. A change to the `lint:links` entry point or a `next-validate-link` version bump will not re-run the check on the PR that makes it. Adding `- 'docs/package.json'` closes that.

aromko (commented):
## PR Review: #5742 - ci(DST-1721): validate docs links and fix the 49 broken ones

### Overview

| Field | Value |
|-------|-------|
| Author | @sebald |
| Branch | `chore/DST-1721-validate-links` → `main` |
| Files Changed | 27 |
| Additions | +169 |
| Deletions | -57 |
| CI | All green (26/26), including the new `Validate docs links` check |
| Visual Regression | Not needed (no `packages/**`, `themes/**`, `.storybook/**`, `*.stories.tsx` or `*.styles.ts` touched) |

### Linked Ticket

**[DST-1721](https://reservix.atlassian.net/browse/DST-1721)**: 🏗️ Guard docs links in CI and fix the 49 broken ones it found
- Status: In Review
- Scope match: yes. The workflow, the script, the entry point and all 52 findings are here, and the four documented exclusions (Fumadocs `resolveHref`, component-prop hrefs, generated release files, external links) are honoured.

### What I verified independently

- **The diagnosis is real.** `curl https://www.marigold-ui.io/components/collection/card` ships literal `href="../../layout/panel"`, `href="../../content/title"`, `href="../../form/selectlist#options-as-cards"`. From `/components/collection/card` those land on `/layout/panel`, `/content/title`, `/form/selectlist`. All 404.
- **The check actually fails.** Pulled `next-validate-link@1.6.7` from the registry: `printErrors(results, throwError)` does `process.exit(1)` when `totalErrors > 0`. The `true` is doing its job, this is not a silent no-op.
- **The validator matches browser semantics, not Fumadocs'.** Its resolver classifies a `../`-prefixed non-`.mdx` href as `relative-url` and resolves it against the page URL treating the base as a file. That is why `card`'s `../../layout/panel` is caught while `stack`'s `../layout/inline` is not, and it means the 40 relative links left in `docs/content` genuinely resolve today.
- **Every rewritten target exists.** Checked all of them against the page tree, including the anchors: `panel#panel-vs-card` (line 76), `selectlist#options-as-cards` (line 46), `forms#validation` (line 226), and `installation#set-up-a-marigold-project-with-tailwind-css-recommended` (line 44).
- **The editorial calls hold up.** `/patterns/user-input/forms#validation` covers everything the retired `/foundations/validation` page did (built-in, custom messages, custom validation, real-time, server errors, Zod). `tag-field` and `patterns/layout/app-frame` both exist. Three of the four are good; the fourth is Issue #1 below.

### Code Review Checklist

Docs content, one CI workflow and one Node script. Most component and TypeScript items are not applicable.

#### TypeScript Standards
- [x] No `any` types
- [x] Proper typing (`validate-links.ts` needs no annotations, the library types carry it)
- [x] N/A: `import type`, interface definitions

#### Component Patterns
- [x] N/A across the board, no component source touched

#### React Best Practices
- [x] N/A

#### React Performance
- [x] N/A

#### Testing Standards
- [x] N/A. The check is the test and it runs in CI. Reasonable call for a link validator.

#### Accessibility
- [x] N/A for the diff itself. Worth noting the fixes are an accessibility win in their own right: 15 dead links removed from screen-reader-reachable prose.

#### CI / Script
- [x] Workflow mirrors `lint.yml` conventions exactly (`checkout@v6`, `pnpm/action-setup@v6.0.9`, `.node-version`, `cache: 'pnpm'`, `CI: true`), minus the build step it does not need
- [x] `concurrency` group with `cancel-in-progress`
- [x] Error path handled: `validateLinks().catch(...)` plus `printErrors(..., true)`
- [ ] `paths` filter omits `docs/package.json` (Issue #3)

### Issues Found

#### Critical (Must Fix)

None.

#### Warnings (Should Fix)

**1. `docs/content/components/form/combobox/index.mdx:49` — the prose now names a hook the package does not export**

The line renames the hook from `useAsyncList` to `useAsyncListData`. It is the other way round: `useAsyncList` is what Marigold ships.

- `packages/components/src/hooks.ts:1` → `export { useAsyncList } from 'react-aria-components/useAsyncList';`, re-exported at `packages/components/src/index.ts:7`
- `useAsyncListData` appears nowhere in `packages/` or `themes/`
- The demo rendered directly under this sentence, `docs/content/components/form/combobox/combobox-async.demo.tsx:1`, does `import { ComboBox, useAsyncList } from '@marigold/components';`

The rest of the docs already handle this correctly by keeping the name and linking the stale slug: `patterns/feedback/async-data-loading/index.mdx:28` says "[useAsyncList documentation](/components/hooks-and-utils/useAsyncListData)" and `getting-started/recommended-libraries/index.mdx:43` says "[`useAsyncList`](/components/hooks-and-utils/useAsyncListData)". Follow that:

```suggestion
The ComboBox component supports working with asynchronous data. In the example below, the [`useAsyncList`](/components/hooks-and-utils/useAsyncListData) hook is used to handle the loading and filtering of data from the server.
```

Worth its own ticket: the target page is misnamed at the source, and `docs/content/components/hooks-and-utils/useAsyncListData/index.mdx:16` tells readers to `import { useAsyncListData } from '@marigold/components'`, which does not resolve.

**2. 15 `<TeaserList>` "Related" links still 404, six of them in files this PR fixed the same target in**

I resolved every `href:` prop in `docs/content` against the branch's page tree using the same URL semantics the validator uses. Fifteen are dead. Confirmed live: the Panel page ships `href="/patterns/admin-master-mark"`, `href="../../content/card"` and `href="../../navigation/accordion"` today.

In files this PR edits, where the identical target is corrected a few screens above:

| File | Line | Broken href | Lands on | Fixed in prose at |
|------|------|-------------|----------|-------------------|
| `components/layout/panel/index.mdx` | 345 | `../../content/card` | `/content/card` | line 84 |
| `components/layout/panel/index.mdx` | 352 | `../../navigation/accordion` | `/navigation/accordion` | — |
| `components/layout/panel/index.mdx` | 359 | `/patterns/admin-master-mark` | `/patterns/admin-master-mark` | lines 57, 58 |
| `components/collection/card/index.mdx` | 306 | `../../form/selectlist#options-as-cards` | `/form/selectlist` | line 81 |
| `components/collection/card/index.mdx` | 328 | `../../layout/panel` | `/layout/panel` | line 52 |
| `patterns/feedback/feedback-messages/index.mdx` | 244 | `../../foundations/validation` | `/foundations/validation` | line 109 |

Nine more of the same class elsewhere, listed for the follow-up ticket: `components/collection/actionbar/index.mdx:166,189`, `components/form/timefield/index.mdx:71`, `components/formatters/numericformat/index.mdx:71`, `foundations/component-principles/index.mdx:264`, `patterns/data/fetching-and-mutations/index.mdx:241,247,252`, `patterns/feedback/loading-states/index.mdx:132`.

One correction to the "known gaps" note: `markdown.components` **cannot** opt these in later. next-validate-link@1.6.7's default `onNode` skips any attribute where `typeof attr.value !== 'string'`, and `TeaserList` receives `items={[{ href: '...' }]}` as an expression, not a string literal. Closing the gap needs a custom `markdown.onNode` that walks `mdxJsxAttributeValueExpression`, or simply fixing the hrefs. At minimum the six in touched files should go in here, since they are the most prominent links on those pages and this PR already corrected their twins.

**3. `.github/workflows/validate-links.yml:10` — add `docs/package.json` to `paths`**

A change to the `lint:links` entry point or a `next-validate-link` bump will not re-run the check on the PR that makes it.

```suggestion
      - '.github/workflows/validate-links.yml'
      - 'docs/package.json'
```

### Recommendation

**Comment**

The infrastructure here is the good kind of lazy: 38 lines of script, a 3.6s job, no build step, modelled on the guard workflow that already exists. The diagnosis is genuinely sharp and I reproduced it. Nothing regresses and nothing about the approach needs rethinking.

Two things are worth fixing before merge, both small:

- Issue #1 is a one-line regression. `main` currently has the right hook name with a dead link; this changes it to the wrong hook name with a live link. Net worse for anyone copying from that page.
- Issue #2 is the headline claim landing at about 75 percent. Fixing `/patterns/admin-master-mark` in the Panel variants table while the Related card 250 lines down keeps the same dead path is the exact failure mode this PR exists to prevent, and it stays invisible for the same reason. The six in touched files are a ten-minute fix; the other nine and the `onNode` work can be a follow-up ticket.

Once those land this is a clear approve. Posting as a comment rather than a formal block — nothing here is a merge risk, but #1 is a regression against `main`.

docs/content/components/form/combobox/index.mdx
sebald: Right on both counts — fixed in 9444bc4.

The href was the correct part; the name was not. Restored to `useAsyncList` while keeping the target:

```
[`useAsyncList`](/components/hooks-and-utils/useAsyncListData)
```

I also went after the rest of the same defect rather than just this line, since the page itself was the worst offender — its Import block told readers to `import { useAsyncListData } from '@marigold/components'`, which does not compile, on a page whose own demo (`async.demo.tsx:1`) imports `useAsyncList`. Renamed in prose only, hrefs untouched:

- `hooks-and-utils/useAsyncListData/index.mdx` — frontmatter `title`, the opening sentence, and the Import block
- `form/autocomplete/index.mdx:151, 223, 225`
- `collection/table/index.mdx:543`

One correction: "every other page keeps the correct name" isn't quite right — `autocomplete/index.mdx:151` already read "The `useAsyncListData` hook provides…" before this PR. Pre-existing, and now fixed above.

On renaming the route itself: agreed it needs a ticket, and there's a concrete blocker worth recording. `docs/next.config.mjs` has a `rewrites()` block but no `redirects()`, and there is no `vercel.json` — so there is no redirect mechanism in the repo at all. Renaming the folder would break the public URL with nothing to catch it. The open question for that ticket is whether the docs site gets a redirects table and where it lives, which is too big a decision to make as a footnote here.

.github/workflows/validate-links.yml
sebald: Agreed — added in 9899c53:

```yaml
      - 'docs/scripts/validate-links.ts'
      # The `lint:links` entry point and the `next-validate-link` version both
      # live here, so a change to either has to re-run the check.
      - 'docs/package.json'
```

docs/content/components/layout/panel/index.mdx
sebald: Confirmed the mechanism against the installed `next-validate-link@1.6.7` — `dist/index.mjs:464`:

```js
if (attr.type !== "mdxJsxAttribute" || typeof attr.value !== "string") continue;
```

So `items={[…]}` is dropped as an `mdxJsxAttributeValueExpression`, exactly as you said. Worth adding: `validate-links.ts` passed no `markdown.components` either, so `components = {}` and the JSX branch of the default `onNode` was dead code — *no* JSX attribute was scanned, not even a plain string `href`.

Rather than choose between "custom `onNode`" and "fix the 15 hrefs", 9899c53 does both, since fixing them without widening the scan just lets the class rot again. The new `onNode` scans markdown links, `href` string attributes on any JSX element, and `href` keys inside expression attributes. I match the expression textually instead of walking the estree — the attribute value is arbitrary JS, and a regex over it cannot throw.

I audited all of `docs/content` first and got **exactly 15**, matching your list item for item, including the `/foundations/validation` that survives at line 244 of the same file this PR fixes at line 109. All 15 rewritten root-relative, per this branch's existing convention:

| file:line | was | now |
|---|---|---|
| `actionbar:166` | `../table` | `/components/collection/table` |
| `actionbar:189` | `../../overlay/drawer` | `/components/overlay/drawer` |
| `card:306` | `../../form/selectlist#options-as-cards` | `/components/form/selectlist#options-as-cards` |
| `card:328` | `../../layout/panel` | `/components/layout/panel` |
| `timefield:71` | `../../introduction/form-fields` | `/foundations/form-fields` |
| `numericformat:71` | `/components/form/numberfield` | `/components/form/number-field` |
| `panel:345` | `../../content/card` | `/components/collection/card` |
| `panel:352` | `../../navigation/accordion` | `/components/navigation/accordion` |
| `panel:359` | `/patterns/admin-master-mark` | `/patterns/layout/admin-master-mark` |
| `component-principles:264` | `../../foundations/design-tokens` | `/foundations/token-overview` |
| `fetching-and-mutations:241` | `../../feedback/loading-states` | `/patterns/feedback/loading-states` |
| `fetching-and-mutations:247` | `../../feedback/async-data-loading` | `/patterns/feedback/async-data-loading` |
| `fetching-and-mutations:252` | `../../user-input/forms` | `/patterns/user-input/forms` |
| `feedback-messages:244` | `../../foundations/validation` | `/patterns/user-input/forms#validation` |
| `loading-states:132` | `../../user-input/forms` | `/patterns/user-input/forms` |

Two were judgment calls, flagging them for you: `design-tokens` has no page, so it goes to `/foundations/token-overview`; `validation` goes to `/patterns/user-input/forms#validation`, matching what line 109 already does. Both anchors verified to exist.

**One correction.** These don't 404 in production — the site answers **200 with the generic not-found shell**, the same response a bogus URL gets:

| URL | status | `<title>` |
|---|---|---|
| `/patterns/admin-master-mark` | 200 | `Marigold Design System` |
| `/content/card` | 200 | `Marigold Design System` |
| `/definitely-not-a-real-page-xyz` | 200 | `Marigold Design System` |
| `/patterns/layout/admin-master-mark` | 200 | `Admin- and Mastermark` |

The links are dead as you described, but a status-code-based crawler would pass them, which makes catching them in source the only option rather than a convenience.

Verified the scan is genuinely live, not just green: `pnpm lint:links` reports 0 errors, and temporarily breaking one expression href produces `1 errored file, 1 errors` with a non-zero exit. Known wart — errors report the element's start line, not the href's, since the position belongs to the node.

docs/content/components/form/combobox/index.mdx
sebald: Following up on the route rename I said should be deferred — I've folded it in after all, in 61598aa. My reason for deferring it did not survive checking.

`/components/hooks-and-utils/useAsyncListData` → `/components/hooks-and-utils/useAsyncList`, plus `meta.json` and the five inbound hrefs. URL, page title and import statement now all agree.

**No redirect, and none was needed.** I had framed the missing redirect mechanism as a blocker. It is real — `next.config.mjs` has only `rewrites()`, there is no `vercel.json`, and nothing outside build artifacts mentions redirects — but the precedent runs the other way. Earlier restructures moved every component URL and left the old ones dead:

| URL | title today |
|---|---|
| `/components/button` | *(not-found shell)* |
| `/components/actions/button` | Button |
| `/components/table` | *(not-found shell)* |
| `/components/collection/table` | Table |

So renaming docs URLs without redirects is the established practice here, not a departure from it. I had set a bar higher than the project's own. Whether the docs site should get a redirects table is still a fair question, but it is about #5109 and #5210's fallout as much as this page, and it should not gate a two-line rename.

The new URL is enforced rather than assumed: pointing any link back at the old slug now fails the check with a non-zero exit.

aromko (approved):

Issue #418, opened 2020-08-11
Component - Message


ti10le:
Short component description (special logic used, use cases etc.)
- [x] create component
- [x] add CSS normalization
- [x] add theme styles
- [x] create story file
- [x] create test file
- [x] create index file

PR #5743, opened 2026-08-14
feat(DST-1723): add an RSS feed for the release blog

sebald:
# Description

The docs publish a release blog — 22 posts, dated, sorted newest first — with no way to subscribe to it. Every conventional feed path was a soft 404: `/rss.xml`, `/feed.xml` and `/releases/blog/rss.xml` all returned `200 text/html`, which is the catch-all, not a file.

`/rss.xml` is now a static route handler (`export const dynamic = 'force-static'`, so it's emitted at build like the rest of the site) over the existing `blogPosts` collection.

Three ways to reach it, because they cover different readers: `metadata.alternates.types` puts an autodiscovery `<link>` in every page's `<head>` for feed readers and crawlers; the **RSS** link in `SiteFooter` makes it findable from anywhere, since browsers stopped surfacing autodiscovery years ago; and the Release Notes page links it in context. At 375px the footer's four links still sit on one row with no horizontal overflow.

Two decisions worth knowing about, both about not changing bytes between deploys:

- **Item links and GUIDs use a new `canonicalUrl`, not `lib/config.ts`'s `baseUrl`.** `baseUrl` falls back to the per-deployment Vercel URL, which is fine for fetching fonts but not for an identifier a third party stores — a GUID that changes between deploys makes every reader re-notify every item.
- **`lastBuildDate` comes from the newest post, not from build time.** Rebuilding without new content produces a byte-identical feed.

XML is hand-rolled rather than adding `feed` + `xml-js` for the sake of five escaped characters. Descriptions reuse the `introduction` that `getAllBlogPosts()` already extracts for the Release Notes page, reduced to plain text — feed readers render a description as text, so `_don't_` and `` `@marigold/cli` `` would otherwise show their Markdown syntax.

Closes DST-1723

## Test Instructions

1. `pnpm install`
2. `pnpm --filter @marigold/docs test` → 115 passing, 9 of them new in `lib/rss.test.ts`
3. `pnpm --filter @marigold/docs dev`, then `curl -i http://localhost:3000/rss.xml` → `200`, `content-type: application/rss+xml; charset=utf-8`, 22 `<item>` elements newest-first
4. `curl -s http://localhost:3000/rss.xml | xmllint --noout -` → well-formed
5. Drop the feed URL into any reader (NetNewsWire, Feedly, Thunderbird) — or click **RSS** in the footer of any page
6. `curl -s http://localhost:3000/ | grep 'application/rss+xml'` → the autodiscovery `<link rel="alternate">` is in `<head>` on every page

## Breaking Changes

No — additive, and no published package is touched.

## Notes for the reviewer

- **Why a route handler and not a static file in `public/`.** The `.md` files under `public/` are generated by `build:md` because that pipeline needs to parse MDX outside the React tree. A feed needs none of that: the collection is already loaded, so a `force-static` handler is both less machinery and automatically correct when a post is added.
- **Only a summary feed.** `description` plus a link, no `content:encoded`. Rendering a post's MDX to standalone HTML is a genuinely separate problem (components, code highlighting, relative links) and not what a release feed needs.
- **`canonicalUrl` is hard-coded to production**, so the feed on a preview deploy points at production URLs. That's deliberate — a preview's feed items shouldn't get stored under a URL that stops existing. Two build scripts (`build-manifest.mjs`, `build-examples.mjs`) already hard-code the same origin; this is the first one in `lib/`, so if we ever want a single source of truth, those should move to it rather than the other way round.
- **Interaction with #5742 (DST-1721, the docs link checker).** `next-validate-link`'s `next` preset only globs `app/**/page.{js,jsx,tsx,md,mdx}` — route handlers are invisible to it — so the new `[RSS](/rss.xml)` link in `content/releases/release-notes.mdx` would be reported as `not-found`. I've added `/rss.xml` to the checker's whitelist on that branch. **Either merge order is safe:** if this PR lands first, main has the link but no checker yet; if #5742 lands first, the whitelist is already there.
- **The `xmllint` step above is the only structural validation.** The W3C Feed Validator needs a public URL, so it's worth a click on the preview deploy before merging — the feed satisfies RSS 2.0's required elements (`channel` title/link/description, and title-or-description per item) plus `atom:link rel="self"`, but that's my reading of the spec, not a validator's.

## Checklist

- [ ] Storybook preview and Marigold docs preview are available — verify once CI builds them; `/rss.xml` on the preview is where step 5 is worth a click
- [x] Unit tests added/updated — `docs/lib/rss.test.ts`, 9 cases (ordering, escaping, canonical GUIDs, deterministic `lastBuildDate`, Markdown reduction, empty feed)
- [x] Component documentation added/updated (if it exists) — the Release Notes page now links the feed

### DIFF
--- a/docs/app/_components/SiteFooter.tsx
+++ b/docs/app/_components/SiteFooter.tsx
@@ -29,12 +29,15 @@ export const SiteFooter = () => (
         >
           GitHub
         </Link>
-        <Link
+        {/* Not `Link`: it treats `/rss.xml` as internal, so the router
+            prefetches every in-viewport footer, pulling the whole feed on a
+            page that has nothing to do with it. */}
+        <a
           className="hover:text-fd-foreground transition-colors"
           href="/rss.xml"
         >
           RSS
-        </Link>
+        </a>
       </nav>
     </div>
   </footer>
--- a/docs/app/_components/SiteFooter.tsx
+++ b/docs/app/_components/SiteFooter.tsx
@@ -29,6 +29,12 @@ export const SiteFooter = () => (
         >
           GitHub
         </Link>
+        <Link
+          className="hover:text-fd-foreground transition-colors"
+          href="/rss.xml"
+        >
+          RSS
+        </Link>
       </nav>
     </div>
   </footer>
--- a/docs/app/rss.xml/route.ts
+++ b/docs/app/rss.xml/route.ts
@@ -1,0 +1,13 @@
+import { getAllBlogPosts } from '@/lib/blog';
+import { buildRssFeed } from '@/lib/rss';
+
+// Emitted at build time like the rest of the site; `GET` handlers are dynamic
+// by default.
+export const dynamic = 'force-static';
+
+export const GET = () =>
+  new Response(buildRssFeed(getAllBlogPosts()), {
+    headers: {
+      'Content-Type': 'application/rss+xml; charset=utf-8',
+    },
+  });
--- a/docs/lib/rss.ts
+++ b/docs/lib/rss.ts
@@ -1,0 +1,73 @@
+import type { BlogPost } from '@/lib/blog';
+import { canonicalUrl } from '@/lib/config';
+
+const FEED_PATH = '/rss.xml';
+const FEED_TITLE = 'Marigold Design System';
+const FEED_DESCRIPTION = "Release notes for Marigold, Reservix' Design System.";
+const BLOG_PATH = '/releases/blog';
+
+const XML_ENTITIES: Record<string, string> = {
+  '&': '&amp;',
+  '<': '&lt;',
+  '>': '&gt;',
+  '"': '&quot;',
+  "'": '&apos;',
+};
+
+const escapeXml = (value: string) =>
+  value.replace(/[&<>"']/g, char => XML_ENTITIES[char]);
+
+// Feed readers render a description as plain text, so inline Markdown would
+// show up as literal syntax.
+const toPlainText = (markdown: string) =>
+  markdown
+    .replace(/\[([^\]]+)\]\([^)]*\)/g, '$1')
+    .replace(/`([^`]+)`/g, '$1')
+    .replace(/\*\*([^*]+)\*\*/g, '$1')
+    // Only emphasis markers, not the underscores in an identifier.
+    .replace(/(?<![\w`])[*_]([^*_\n]+)[*_](?![\w`])/g, '$1')
+    .replace(/\s+/g, ' ')
+    .trim();
+
+const buildItem = (post: BlogPost) => {
+  const url = `${canonicalUrl}${post.slug}`;
+  const description = toPlainText(post.introduction);
+
+  return [
+    '    <item>',
+    `      <title>${escapeXml(post.title)}</title>`,
+    `      <link>${escapeXml(url)}</link>`,
+    `      <guid isPermaLink="true">${escapeXml(url)}</guid>`,
+    `      <pubDate>${post.date.toUTCString()}</pubDate>`,
+    ...(description
+      ? [`      <description>${escapeXml(description)}</description>`]
+      : []),
+    '    </item>',
+  ].join('\n');
+};
+
+export const buildRssFeed = (posts: BlogPost[]) => {
+  const sorted = [...posts].sort((a, b) => b.date.getTime() - a.date.getTime());
+
+  // Derived from the newest post rather than the build time, so rebuilding
+  // without new content produces a byte-identical feed.
+  const lastBuildDate = sorted[0]?.date.toUTCString();
+
+  return [
+    '<?xml version="1.0" encoding="UTF-8"?>',
+    '<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">',
+    '  <channel>',
+    `    <title>${escapeXml(FEED_TITLE)}</title>`,
+    `    <link>${canonicalUrl}${BLOG_PATH}</link>`,
+    `    <description>${escapeXml(FEED_DESCRIPTION)}</description>`,
+    '    <language>en</language>',
+    `    <atom:link href="${canonicalUrl}${FEED_PATH}" rel="self" type="application/rss+xml"/>`,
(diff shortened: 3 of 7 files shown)


docs/lib/rss.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const escapeXml = (value: string) =>
+  value.replace(/[&<>"']/g, char => XML_ENTITIES[char]);
+
+// Feed readers render a description as plain text, so inline Markdown would
+// show up as literal syntax.
+const toPlainText = (markdown: string) =>
+  markdown
+    .replace(/\[([^\]]+)\]\([^)]*\)/g, '$1')
+    .replace(/`([^`]+)`/g, '$1')
```
aromko, line 25: ⚠️ `toPlainText` unwraps inline code, so `` `<RangeCalendar>` `` becomes the literal text `<RangeCalendar>` and ships as `&lt;RangeCalendar&gt;`. After XML parsing a reader holds the string `<RangeCalendar>`, and readers render `<description>` as HTML rather than text (react.dev and nextjs.org both wrap descriptions in CDATA for exactly this reason), so an unknown tag is dropped: *"Back in v17.5.0 we shipped ."*

On the preview feed this hits 7 of 22 items, 12 occurrences (`<RangeCalendar>` x2, `<Dialog>`, `<Select>`, `<Switch>`, `<Sidebar>`, `<TopNavigation>`, `<TagField>`, `<Divider>`, `<Inset>`, `<Card>`, `<DateRangePicker>`). Reader-dependent, but the failure mode is losing the one word the sentence is about. The brackets only exist because the source wrote inline code, so dropping them is the smallest fix:

```suggestion
    .replace(/`([^`]+)`/g, '$1')
    // Component names lose their brackets too. A reader renders a description
    // as HTML, so an escaped `<Dialog>` comes back out of the XML parser as a
    // tag and gets dropped, taking the word the sentence is about with it.
    .replace(/<(\/?[A-Za-z][\w.-]*)>/g, '$1')
```

Verified over all 22 real intros: `"we shipped RangeCalendar"`, `"a wider Dialog"`, `"started with Sidebar and TopNavigation"`. Pushed as a commit on this branch, along with a test case.

docs/app/_components/SiteFooter.tsx
```diff
@@ -29,6 +29,12 @@ export const SiteFooter = () => (
         >
           GitHub
         </Link>
+        <Link
+          className="hover:text-fd-foreground transition-colors"
+          href="/rss.xml"
+        >
+          RSS
+        </Link>
```
aromko: ⚠️ `fumadocs-core/link` only renders a plain `<a>` for external hrefs, so `/rss.xml` goes through `next/link` and the App Router prefetches it once the footer scrolls into view. Measured on the preview while sitting on `/foundations/accessibility`: `GET /rss.xml?_rsc=...` -> `200`, the full 14 kB of XML (`x-matched-path: /rss.xml.rsc`), for a router cache entry that isn't flight data. The click still works via the MPA fallback, so this is waste rather than breakage, but it's 14 kB per navigation and it grows with every release post.

```suggestion
        {/* Not `Link`: it treats `/rss.xml` as internal, so the router
            prefetches every in-viewport footer, pulling the whole feed on a
            page that has nothing to do with it. */}
        <a
          className="hover:text-fd-foreground transition-colors"
          href="/rss.xml"
        >
          RSS
        </a>
```

`prefetch={false}` works too. I left the `[RSS](/rss.xml)` link in `release-notes.mdx` alone: `defaultMdxComponents` maps `a` onto the same component, and prefetching the feed on the page that advertises it is the one place it's plausibly wanted. Pushed as a commit on this branch.

docs/app/rss.xml/route.ts
```diff
@@ -0,0 +1,13 @@
+import { getAllBlogPosts } from '@/lib/blog';
+import { buildRssFeed } from '@/lib/rss';
+
+// Emitted at build time like the rest of the site; `GET` handlers are dynamic
+// by default.
+export const dynamic = 'force-static';
+
+export const GET = () =>
```
aromko, line 8: 💡 No test here, while `app/api/*/route.test.ts` is the established pattern. The bit worth locking is the `Content-Type`: it's what makes a reader accept the body instead of a browser downloading it. Fair to skip if mocking `getAllBlogPosts` costs more than the header is worth.

aromko (commented):
## Automated Code Review

Reviewed against DST-1723 and verified end to end on the preview deploy: `GET /rss.xml` returns `200` with `application/rss+xml; charset=utf-8` and `x-vercel-cache: PRERENDER` (so `force-static` really does emit it at build), 22 items newest-first, `xmllint --noout` clean, and the autodiscovery `<link rel="alternate">` is in `<head>` on `/` and on a docs page. No page under `docs/app` sets its own `alternates`, so nothing overrides the layout's. I also ran `extractIntroduction` + `toPlainText` over all 22 real posts: no leaked frontmatter or JSX. Scope matches the ticket 1:1, and no VRT is needed (docs-only, nothing under `packages/**`, `themes/**`, `*.stories.tsx` or `*.styles.ts`).

Two warnings, both inline, and I have **pushed the fixes to this branch** as two small commits (plus one test case) rather than leaving you to apply suggestions:

1. Component names in descriptions can vanish in feed readers (7 of 22 items).
2. The footer link pulls the whole feed as an RSC prefetch on every page view.

Two things that need no action:

- **Feed title.** `<title>Marigold Design System</title>` is the site name. In a reader's sidebar next to other feeds, "Marigold Releases" says what actually arrives. Also `"Reservix' Design System"` reads like a plural possessive; `Reservix's`, or "the Reservix Design System".
- **`canonicalUrl`.** It is the third hard-coded copy of the origin (`build-manifest.mjs:45`, `build-examples.mjs:37`). Agreeing with your note: those should move to it, not the reverse. Nothing to do here.

Unrelated to this PR, for the record: the Release Notes page logs a React #418 hydration error on the preview. It is pre-existing on `main` (same error on #5750's preview, none on production), not caused by the feed.

Nice work on the deterministic `lastBuildDate` and canonical GUIDs, that is the part feeds usually get wrong.

aromko (approved):

PR #5744, opened 2026-08-14
feat(DST-1724): publish llms.txt

sebald:
# Description

`/llms.txt` is the path an agent probes without being told anything. We already publish `manifest.json` and a `.md` per page — but that only helps someone who read [Usage with AI](https://www.marigold-ui.io/getting-started/usage-with-ai), or who is using the CLI or the (OAuth-gated) MCP server. This is the same index in the conventional format, so an agent that arrives cold discovers the docs *and* learns the `.md` convention exists.

**It's an index over the existing Markdown build, not a second one.** `scripts/build-llms.ts` reads `public/manifest.json` and links each page at the `.md` URL that's already there. It deliberately does not re-parse MDX: the existing pipeline (`build-md` → `lib/markdown/parser.ts`) resolves `<ComponentDemo>` to real demo source and `<PropsTable>` to Markdown tables, which a generic MDX-to-Markdown plugin would not. It runs after `build:md` and asserts every `.md` it advertises exists, so the index can't ship links to 404s.

Section order and labels come from the `meta.json` separators (`---Hooks and Utils---`, `---User Input---`), so `llms.txt` reads in sidebar order rather than alphabetically:

```
## Components / Application
## Components / Layout
## Components / Actions
...
- [Button](https://www.marigold-ui.io/components/actions/button.md): Buttons allow users to trigger actions.
- [ButtonGroup](https://www.marigold-ui.io/components/actions/button-group.md): (beta) Toolbar wrapper around a related cluster of buttons.
```

119 pages, 17 KB.

**No `llms-full.txt`.** I built it, measured it at 2 MB, and cut it: that's past what most agents ingest in one shot, the manifest-then-fetch-a-page path is better for the ones that won't, MCP already does semantic retrieval, and it would be a third full copy of the corpus that can drift from the other two.

### Two fixes that stand on their own

- **`remarkSimplifyJsx` never handled `mdxjsEsm`.** It strips JSX elements and expressions, but `import`/`export` nodes passed straight through — so **691 `import` statements for demo components were shipping inside the generated `.md` files**, which is what the CLI serves and what the MCP embedding index is built from. Their targets are relative MDX-only paths that resolve to nothing for a reader of the Markdown. `card.md` opened with `import { CardAnatomy } from './card-anatomy';` in production. Imports *inside* fenced demo source are untouched — those are the demo.
- **The `/docs/:path*.mdx` → `/llms.mdx/docs/:path*` rewrite is dead.** This site's `baseUrl` is `''`, so there's no `/docs` prefix and no `llms.mdx` route; `/docs/components/collection/card.mdx` resolves to the catch-all today. Leftover from the Fumadocs template.

Closes DST-1724

## Test Instructions

1. `pnpm install`
2. `pnpm --filter @marigold/docs build:md && pnpm --filter @marigold/docs build:llms` → `Built llms.txt (119 pages, 17 KB)`
3. `pnpm --filter @marigold/docs test` → 120 passing: 10 in `scripts/build-llms.test.ts` and 4 in `lib/markdown/plugins/simplify-jsx.test.ts`, all new
4. `pnpm --filter @marigold/docs dev`, then `curl -i http://localhost:3000/llms.txt` → `200`, `text/plain`, sidebar-ordered sections
5. Confirm the import fix: `curl -s http://localhost:3000/components/collection/card.md | head -5` → starts at `# Card`, no `import` line. Compare with production, which still shows 11 of them.
6. Delete a generated `.md` (e.g. `docs/public/components/actions/button.md`) and re-run `build:llms` → it fails with `Missing generated markdown for 1 page(s)` instead of publishing a broken link.

## Breaking Changes

No — build output and docs only, no published package touched.

## Notes for the reviewer

- **`public/llms.txt` is gitignored**, like `manifest.json` and the generated `.md` files. It's build output.
- **Why a separate script rather than folding it into `build-manifest.mjs`.** The manifest is written before the `.md` files exist; the existence check needs them. Chaining after `build:md` is what makes that check meaningful.
- **`vitest.config.ts` include widened** from `scripts/**/*.test.mjs` to `scripts/**/*.test.{ts,mjs}`. The existing scripts tests are `.mjs`; this one is TypeScript because it type-checks the manifest shape it consumes. Confirmed the new file actually runs (test count went 106 → 115, file count 12 → 13) — that config has silently skipped a whole directory before, per the comment in it.
- **`readCategories` parses a Fumadocs convention.** A `meta.json` mixes plain slugs, separators (`---Label---`) and folder spreads (`...folder`); a separator labels the spread that follows it. If a spread has no separator above it, the label falls back to the humanised folder name. Both paths are tested.
- **No link to `/rss.xml`.** The feed route arrives with DST-1723 (#5743), so linking it here would have shipped a link the catch-all answers with the 404 page. Removed, and the existence check now covers the hand-written lines too — so it can be added back safely once #5743 lands.
- **Deliberately out of scope: `fumadocs-core/negotiation`**, which would serve Markdown from the canonical URL by `Accept` header instead of requiring the `.md` suffix. Worth doing, but it changes what every request to a docs page can return — separate change, separate risk.
- **`llms.txt` is a convention, not a standard.** Google has said it doesn't consume it and adoption is patchy. The honest value is the cold-start case: 17 KB of generated output, no maintenance, and it's the file agents look for.
- **Verified:** `build:llms` clean, 115 tests passing, `prettier --check` and `eslint` clean on all changed files, `tsc --noEmit` clean apart from two pre-existing `table-expandable-rows.demo.tsx` errors that are on `main` and come from packages not being built in this worktree. `.md` files still served with `text/markdown` after the rewrite removal.

## Checklist

- [ ] Storybook preview and Marigold docs preview are available — verify once CI builds them; `/llms.txt` on the docs preview is the thing to open
- [x] Unit tests added/updated — `docs/scripts/build-llms.test.ts`, 9 cases (label derivation, sidebar ordering, unknown category last, badge notes, missing description, `.md` linking)
- [x] Component documentation added/updated (if it exists) — a bullet under "Manifest and direct ingestion" in Usage with AI


docs/scripts/build-llms.ts
```diff
@@ -0,0 +1,209 @@
...
+    '',
+    `> ${SUMMARY}`,
+    '',
+    'Every page is linked below as Markdown. Those `.md` files are generated from',
+    'the docs source with component demos inlined as source code and props tables',
+    'expanded, so they carry more than the rendered page shows. Append `.md` to any',
+    'docs URL to get the same for a page that is not listed here.',
+    '',
+    `- Page index as JSON: ${baseUrl}/manifest.json`,
+    `- Release notes: ${baseUrl}/releases/release-notes (feed: ${baseUrl}/rss.xml)`,
```
aromko: 🔴 **This links to a feed that doesn't exist.**

There's no RSS route in the docs app — no `rss` match anywhere under `docs/`, no `app/**/rss.xml/route.ts`, nothing in `public/`. Production *looks* fine (`/rss.xml` → `200`) but that's the catch-all:

```
content-type: text/html; charset=utf-8
x-matched-path: /[...slug]
```

…and the body is the 404 page. An agent following the link gets HTML with no failure signal.

This is also the one link `llms.txt` ships that the existence check below doesn't cover — `missing[]` only validates `.md` pages, so the "can't ship links to 404s" guarantee has a hole exactly where the hand-written lines are.

```suggestion
    `- Release notes: ${baseUrl}/releases/release-notes`,
```

docs/scripts/build-llms.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    expect(index).toContain(
+      '- [Button](https://www.marigold-ui.io/components/actions/button.md)\n'
+    );
+  });
+
+  it('points at the sibling machine-readable entry points', () => {
+    const index = buildIndex(manifest([page()]), CATEGORIES);
+
+    expect(index).toContain('https://www.marigold-ui.io/manifest.json');
+    expect(index).toContain('https://www.marigold-ui.io/rss.xml');
```
aromko: 🔴 Locks in the broken URL from `build-llms.ts:135`. Drop this line together with the `(feed: …)` parenthetical.

docs/lib/markdown/plugins/simplify-jsx.ts
```diff
@@ -413,5 +413,19 @@ export function remarkSimplifyJsx() {
           return cleanExpression(node, index, parent);
       }
     );
+
+    // The `import`s that pull demo components into a page. Their targets are
+    // relative MDX-only paths, so they resolve to nothing for a reader of the
+    // generated markdown.
+    visit(
+      tree,
+      'mdxjsEsm',
```
aromko, line 422: ⚠️ **The most consequential change in the PR, and the only one with no test.**

691 import lines were shipping into the `.md` files the CLI serves and the MCP embeddings are built from, and nothing caught it. I get that `lib/markdown/**` has no tests at all today, so this isn't a standard you're breaking — but `vitest.config.ts` already includes `lib/**/*.test.ts`, so a ~10-line `unified()` round-trip asserting (a) a top-level `import` is gone and (b) an `import` inside a fenced block survives would cost almost nothing and guard a path that has already regressed once, silently, in production.

docs/scripts/build-llms.ts
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env tsx
+// Generate public/llms.txt and public/llms-full.txt — the conventional
```
aromko: ⚠️ Stale on arrival — `llms-full.txt` isn't generated, so "Generate `public/llms.txt` and `public/llms-full.txt`" and "**Both** are indexes over…" don't match the file. Worth folding the *why not* from the PR description in here; it's the first place someone will look when they wonder where the full corpus went.

docs/scripts/build-llms.ts
```diff
@@ -0,0 +1,209 @@
...
+  areaMeta: Record<string, string[]>
+): Category[] => {
+  const categories: Category[] = [];
+
+  for (const area of topLevel) {
+    const areaLabel = humanizeCategory(area);
+    categories.push({ category: area, label: areaLabel });
+
+    let label: string | null = null;
+    for (const entry of areaMeta[area] ?? []) {
```
aromko, line 80: 💡 `areaMeta[area] ?? []` implies a missing `meta.json` is survivable, but `readMeta` throws on ENOENT before this can apply. Adding a top-level folder to `content/meta.json` without a `meta.json` in it dies with a raw ENOENT instead of something actionable. Either drop the `??` or let `readMeta` return `[]`.

docs/scripts/build-llms.ts
```diff
@@ -0,0 +1,209 @@
...
+        continue;
+      }
+      if (!entry.startsWith('...')) continue;
+
+      const folder = entry.slice(3);
+      categories.push({
+        category: `${area}/${folder}`,
+        label: `${areaLabel} / ${label ?? humanizeCategory(folder)}`,
+      });
+      label = null;
```
aromko, line 99: 💡 `label` only resets after a spread is consumed, so a separator followed by plain slugs (as in `content/releases/meta.json`) leaves a stale label for the next spread. Not reachable today, but resetting on any consumed entry stops it depending on that.

aromko (commented):
## Automated Code Review

### Overview

| Field | Value |
|-------|-------|
| Author | @sebald |
| Branch | `feat/DST-1724-llms-txt` → `main` |
| Files Changed | 8 (+355 / −10) |
| CI | 24/24 green |
| Visual Regression | Not needed (no `packages/`, `themes/`, `.storybook/`, `*.stories.tsx` or `*.styles.ts` touched) |

### Linked ticket

**[DST-1724](https://reservix.atlassian.net/browse/DST-1724)** — ✨ Publish llms.txt from the existing Markdown build · Unplanned · **In Review**

Scope deviation: the ticket's **Included** list still names `llms-full.txt`. This PR cut it (2 MB, third copy of the corpus, drift risk) and argues it well — but the ticket now records a deliverable that isn't shipping. Worth editing the ticket so it matches.

### What I verified independently

- **Dead rewrite removal is correct.** `lib/source.ts:9` has `baseUrl: ''`, there is no `llms.mdx` route under `app/`, and production confirms `/docs/…` resolves to the catch-all. Safe deletion.
- **Import stripping can't break demo resolution.** `parser.ts` runs `remarkResolveComponentDemo` *before* `remarkSimplifyJsx`, so demos are already inlined by the time the `mdxjsEsm` nodes are dropped. Imports inside fenced blocks are `code` nodes and survive — checked `patterns/data/fetching-and-mutations/index.mdx`, whose `export const useVenues` sits in a ```` ```tsx ```` fence and is untouched. The only top-level MDX export in `content/` is `Sw` in a release blog post, a render helper that *should* go.
- **`[SKIP, index]` after splice is the right pattern** and matches the three existing visitors in the file.
- **Sidebar ordering holds** against the real `meta.json` files, including `releases` (separators followed by plain slugs, no spreads) and `foundations`/`getting-started` (no spreads at all — they fall through to the top-level category).

### Issues found

#### 🔴 Critical

**`docs/scripts/build-llms.ts:135` — advertises an RSS feed that does not exist.**

There is no RSS route: no `rss` match anywhere under `docs/`, no `app/**/rss.xml/route.ts`, nothing in `public/`. Production returns `200` for `/rss.xml`, but:

```
content-type: text/html; charset=utf-8
x-matched-path: /[...slug]
```

…and the body contains "404" — it's the catch-all 404 page served with a 200 status. An agent following that link gets HTML with no failure signal at all.

This is also the one link in the file that the build's own guard doesn't cover — `missing[]` only validates `.md` pages, so the "can't ship links to 404s" guarantee has a hole exactly where the hand-written lines are. `docs/scripts/build-llms.test.ts:125` asserts the broken URL is present, so the test currently locks the bug in.

#### ⚠️ Warnings

**`docs/lib/markdown/plugins/simplify-jsx.ts:422` — the highest-impact change in the PR has no test.** 691 `import` lines were shipping into the `.md` files the CLI serves and the MCP embedding index is built from, and nothing caught it. `lib/markdown/**` has zero tests today, but `vitest.config.ts` already includes `lib/**/*.test.ts` — a ~10-line `unified()` round-trip would guard a path that has already regressed once, silently, in production.

**`docs/scripts/build-llms.ts:2` — header comment is stale on arrival.** Says "Generate public/llms.txt **and public/llms-full.txt**" and "**Both** are indexes over…". `llms-full.txt` was cut.

#### 💡 Suggestions

**`docs/scripts/build-llms.ts:74` — `areaMeta[area] ?? []` is unreachable.** It implies a missing `meta.json` is tolerated, but `readMeta` throws on ENOENT before the fallback can apply.

**`docs/scripts/build-llms.ts:87` — `label` only resets after a spread.** A separator followed by plain slugs (as in `content/releases/meta.json`) leaves a stale label for the next spread. Not reachable today.

### Checklist

- **TypeScript** ✅ no `any`, exported interfaces, `import type` where applicable
- **Testing** ✅ Vitest, pure-function tests, no fixtures — good coverage on the index builders; gap noted above on the plugin
- **Components / a11y / React performance** — N/A, no React code in this diff
- **Build wiring** ✅ `build:llms` chained after `build:md`; `public/llms.txt` gitignored alongside its siblings; the vitest include widening is confirmed by the passing Docs Script Tests job

### Recommendation

**Request changes** on one defect — and it's a two-line fix (drop the `(feed: …)` parenthetical at `build-llms.ts:135` and the assertion at `build-llms.test.ts:125`). Everything else is a comment fix or a suggestion. The core work is well-scoped and the two drive-by fixes genuinely stand on their own.

---

docs/scripts/build-llms.ts
sebald: Confirmed and fixed in fa990f2e6 — took your suggestion verbatim. There is no `rss` route anywhere on this branch; the only two matches under `docs/` were the link itself and the test pinning it. `app/rss.xml/route.ts` arrives with #5743, which is not merged.

The second half of your comment is the more useful one, so I fixed that too: `missing[]` only walked `manifest.pages`, and neither hand-written line is a manifest page — `/releases/release-notes` is not in the manifest at all (it has no `.md` twin; the manifest only covers `components`, `foundations`, `getting-started`, `patterns`). So the guarantee had a hole precisely where the hand-written lines live. Both are now checked against the source they point at *and* against the index still linking them, so the constant in `buildIndex` and the check cannot drift apart:

```
$ mv content/releases/release-notes.mdx /tmp/ && pnpm build:llms
Error: llms.txt links https://www.marigold-ui.io/releases/release-notes, but content/releases/release-notes.mdx does not exist
```

That makes re-adding the feed link after #5743 lands a self-checking change rather than a hopeful one.

One thing worth flagging that I found while in here: `scripts/validate-links.ts` on `main` (#5742) carries `whitelist: ["/rss.xml"]`, and nothing in `content/` links it. So `main` is already whitelisting a route that does not exist — same class of problem, different file. Out of scope here, but someone should look at it when #5743 merges.

docs/scripts/build-llms.test.ts
sebald: Dropped in fa990f2e6, along with the `(feed: …)` parenthetical. The test now pins `/releases/release-notes` instead, which is a page that exists.

docs/lib/markdown/plugins/simplify-jsx.ts
sebald: Fair, and you are right that this was the one change with real blast radius and no test. Added `lib/markdown/plugins/simplify-jsx.test.ts` in fa990f2e6 — four cases, `unified()` round-trip through the same plugin order as `parser.ts`:

- a top-level `import` is gone
- an `export` goes with it
- an `import` inside fenced demo source survives
- adjacent `import` blocks all go, not every other one

I checked each one guards something rather than just passing, by mutating the visitor:

| mutant | result |
|---|---|
| drop the `mdxjsEsm` visitor entirely | 3 of 4 fail (the fenced-source case correctly still passes) |
| `splice` without `return [SKIP, index]` | the adjacent-blocks case fails |

That last one is worth a note: my first draft wrote the three imports on consecutive lines, and it caught nothing — MDX parses consecutive `import` lines as a **single** `mdxjsEsm` node, so there is no index to get wrong. Blank lines between them are what makes three sibling nodes and exercises the splice. The test says so in a comment, since it is not obvious and the useless version looks identical at a glance.

docs/scripts/build-llms.ts
sebald: Fixed in fa990f2e6. The header no longer claims an `llms-full.txt`, and the *why not* is folded in where you suggested — ~2 MB, past what most agents ingest in one shot, and a third copy of the corpus free to drift from the manifest and the `.md` files. Also corrected "Both are indexes over…" to the singular.

docs/scripts/build-llms.ts
sebald: Right diagnosis: `readMeta` rejects inside the `Promise.all` before `readCategories` ever runs, so the `??` cannot apply at runtime and the failure is a raw ENOENT with no hint. Fixed in fa990f2e6 by naming the file:

```
Error: Missing content/foo/meta.json, which content/meta.json lists.
```

I kept the `?? []` rather than dropping it, which is where I would push back gently. `readCategories` is exported and unit-tested against hand-built objects, so being total for an area it was given no entry for is the better contract for it — dropping the `??` buys a `TypeError` on `undefined` instead. The misleading part was the raw ENOENT, and that is what changed.

docs/scripts/build-llms.ts
sebald: Confirmed and fixed in fa990f2e6 — `label` now resets on any consumed entry, not just a spread.

On "not reachable today", agreed, and here is why exactly: the leak needs separator → plain slug → spread with no separator between. `components/meta.json` and `patterns/meta.json` pair every separator with a spread immediately below it, and `releases/meta.json` has separators followed by plain slugs but no spreads at all. So no file today has the shape.

Added a test for it, and confirmed it fails against the old logic rather than just passing:

```
× does not leak a label past the plain slugs that follow it
  AssertionError: expected [ …(2) ] to deeply equal [ …(2) ]
```

aromko (approved):

PR #5745, opened 2026-08-14
feat(DST-1725): serve markdown from canonical docs URLs via content negotiation

sebald:
# Description

We generate an optimised Markdown twin of every docs page, but getting it requires knowing to append `.md`. That convention is documented on *Usage with AI*, listed in `manifest.json`, and (in #5744) indexed in `llms.txt` — all of which only help an agent that already found one of those three files. An agent handed a plain docs link from Slack or a PR review scrapes the HTML instead, which is the exact thing the generated Markdown exists to avoid.

This makes the canonical URL serve both:

```bash
curl https://www.marigold-ui.io/components/actions/button                          # → HTML
curl -H "Accept: text/markdown" https://www.marigold-ui.io/components/actions/button  # → Markdown
```

No new URLs, no new corpus. The rewrite target is the same `public/**/*.md` that `build:md` already writes, so this is a routing change on top of the existing pipeline. The `.md` URLs keep working exactly as before.

**`middleware.js` is deprecated in Next 16 and renamed to `proxy.js`**, so this is `docs/proxy.ts`. Proxy runs at step 3 of the routing chain, ahead of filesystem routes at step 5, which is what lets a rewrite land on a static file in `public/`.

## Why this is safe for browsers

The whole risk of `Accept` negotiation is serving Markdown to a human. Two things rule it out, both verified rather than assumed:

- **`isMarkdownPreferred` compares q-values**, it doesn't just look for a substring: `markdown > 0 && markdown >= html`. A browser sends `text/html,…,*/*;q=0.8`, which scores 0 for Markdown, so it gets HTML. Plain `curl` (`Accept: */*`) also scores 0 — **nothing that works today changes behaviour**. Only a client that explicitly names `text/markdown`, `text/x-markdown` or `text/plain` gets Markdown.
- **Vercel's CDN already includes `Accept` in its cache key by default** ([docs](https://vercel.com/docs/caching/cdn-cache): "Vercel's CDN already includes the `Accept` and `Accept-Encoding` headers as part of the cache key by default"). So a cached Markdown response cannot be served to a browser. `Vary: Accept` is set anyway, for any other cache in front of these URLs.

## Guards, and why each one exists

Both of these are real, not defensive padding — I hit them:

- **`__internal__`** is a live route (`/components/__internal__/component-guidelines` returns 200 in production) that `build-md` deliberately skips, so it has no `.md`. Without the guard, an agent asking for Markdown there gets a 404 instead of the page.
- **A path that already carries an extension** would rewrite `/…/button.md` → `/…/button.md.md`.
- **A trailing slash** would ask for `/…/button/.md`. Worth knowing: this needs `new URL(request.url)` rather than `request.nextUrl.clone()`, because `NextURL` re-appends the trailing slash it was constructed with when you assign `pathname` — a test caught that.

# Test Instructions

1. `pnpm install && pnpm build` (component packages must be current — `main` just added ListView)
2. `pnpm --filter @marigold/docs test` → 118 passing, 12 of them new in `proxy.test.ts`
3. `pnpm --filter @marigold/docs dev`, then:

```bash
curl -sI -H 'Accept: text/markdown' localhost:3000/components/actions/button   # 200, text/markdown, Vary: Accept
curl -sI localhost:3000/components/actions/button                              # 200, text/html
curl -s  -H 'Accept: text/markdown' localhost:3000/components/actions/button | head -1   # → # Button
```

4. Confirm the tricky paths: `/components/form` (see below) returns `# Form`; `/components/__internal__/component-guidelines` with `Accept: text/markdown` returns HTML, not a 404; `/manifest.json` is untouched.

# Breaking Changes

No. Every URL that works today returns exactly what it returns today, including for `curl` with no `Accept` header. The only new behaviour is for clients that explicitly ask for Markdown.

# Notes for the reviewer

- **The matcher is limited to the five content roots** (`/components`, `/foundations`, `/getting-started`, `/patterns`, `/releases`), which is exactly where generated Markdown exists. Verified: 155 generated `.md` = 120 manifest pages + 35 releases, and the manifest's own build asserts every page has one. `/`, `/examples/**`, `/mcp`, `/api/**` and the legal pages are deliberately out.
- **`/components/form` is the one page where the naive rule could have been wrong.** `slugFromRelPath` collapses a doubled last segment, so `content/components/form/form/index.mdx` is written to `public/components/form.md`. I checked the manifest before relying on it: fumadocs routes that page at `/components/form`, so `${pathname}.md` is right there too. No manifest slug contains a dot, either.
- **A path with no `.md` degrades to HTML rather than erroring.** Found this by accident probing `/foundations/colors`, which isn't a page at all: the rewrite misses in `public/`, falls through to the catch-all, and the client gets HTML. That's the behaviour I'd want anyway.
- **Next.js owns `Vary` on rendered pages.** A successful page render comes back with `Vary: rsc, next-router-state-tree, next-router-prefetch, next-router-segment-prefetch, Accept-Encoding` — my `Accept` is replaced. So `Vary: Accept` is only set on the Markdown response, where it survives, and the HTML side relies on Vercel's default `Accept` keying. I removed the `Vary` I had originally set on the pass-through branch rather than leave code that looks like it does something it doesn't.
- **The docs' `unstable_doesProxyMatch` doesn't exist in 16.3.0.** The bundled Next docs name it, but the shipped package only exports `unstable_doesMiddlewareMatch`, which is what the tests use. Flagging it because copying the doc example verbatim fails.
- **`docs/proxy.ts` is not covered by the root `pnpm typecheck`** — `tsconfig.check.json` excludes `docs/**/*`. I verified with `--listFiles` and ran `pnpm --filter @marigold/docs types:check` separately, which is the one that covers it.
- **Two small overlaps with #5744**, if that lands first: a bullet in the same list in `usage-with-ai/index.mdx`, and one line in `vitest.config.ts`'s `include` (that PR widens the `scripts/` pattern; this one adds `proxy.test.ts`, since `proxy.ts` has to sit next to `app/` so its test does too). Both are trivial to resolve.
- **Deliberately out of scope:** negotiating anything other than Markdown, and touching how the `.md` files are generated.

# Checklist

- [ ] Storybook preview and Marigold docs preview are available — verify once CI builds them; the thing to run against the docs preview is the `curl -H 'Accept: text/markdown'` above
- [x] Unit tests added/updated — `docs/proxy.test.ts`, 12 cases (matcher inclusion and exclusion, browser vs. Markdown vs. no-preference clients, q-value ranking, `Vary`, and the three paths a naive rewrite breaks)
- [x] Component documentation added/updated (if it exists) — a bullet under "Manifest and direct ingestion" in Usage with AI
- [ ] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) — N/A, no interactive components
- [ ] Visual regression tests updated — N/A, Chromatic snapshots Storybook, not the docs site

Closes DST-1725


sebald:
## Verified on the docs preview

All 24 checks pass. Probed `marigold-docs-git-feat-dst-1725-content-negotiation-marigold.vercel.app`, i.e. real Vercel edge rather than a local server:

| Request | Result |
| --- | --- |
| `Accept: text/markdown` | 200 `text/markdown`, `Vary: Accept`, body starts `# Button` |
| Browser `Accept` | 200 `text/html`, `<title>Button</title>` |
| `Accept: */*` (curl default) | 200 `text/html` — unchanged from today |
| `/components/form` | 200 `text/markdown` — the collapsed-path page |
| `/foundations/spacing`, `/releases/blog/release-2026-06-09` | 200 `text/markdown` |
| `/components/actions/button.md` | 200 `text/markdown`, unproxied |
| `/components/__internal__/component-guidelines` | 200 `text/html` — guard holds, no 404 |

### Cache isolation, measured

The one thing worth proving rather than asserting is that a cached Markdown response can't reach a browser. Four sequential requests to the same URL:

```
1. markdown (cold)   -> text/markdown  cache=MISS       # Forms
2. markdown (warm)   -> text/markdown  cache=HIT        # Forms
3. browser           -> text/html      cache=PRERENDER  <!DOCTYPE html>
4. markdown          -> text/markdown  cache=HIT        # Forms
```

Step 3 is the point: with a warm Markdown entry in the CDN, a browser hitting the identical URL still gets HTML. The two variants occupy separate cache entries, which confirms Vercel's documented default of including `Accept` in the cache key.

Also confirms the production build registers the proxy — `next build` lists `ƒ Proxy (Middleware)` in the route table.

docs/proxy.ts
```diff
@@ -0,0 +1,43 @@
...
+import { NextResponse } from 'next/server';
+import type { NextRequest } from 'next/server';
+
+// Every docs page has a generated Markdown twin in `public/` (see
+// scripts/build-md.ts). Hand it to clients that ask for Markdown so an agent
+// doesn't need to know about the `.md` suffix to get the good version. Proxy
+// runs ahead of filesystem routes, so the rewrite resolves to the static file.
+
+export const proxy = (request: NextRequest) => {
+  if (!isMarkdownPreferred(request)) return NextResponse.next();
```
aromko, line 11: 🔴 **The q-value comparison this relies on is not in the installed `fumadocs-core`.**

The PR body and DST-1725 both state that `isMarkdownPreferred` computes `markdown > 0 && markdown >= html`, and the ticket says "We are on `fumadocs-core` 16.14.3". `docs/package.json` declares `^16.6.5` and `pnpm-lock.yaml` resolves **16.9.3**, where there is no ranking at all:

```js
// fumadocs-core@16.9.3, dist/negotiation
const mediaTypes = getNegotiator(request).mediaTypes();
return markdownMediaTypes.some(type => mediaTypes.includes(type));
```

The `markdown >= html` implementation landed in **16.13.0** (checked 16.10.0 / 16.11.0 / 16.12.0 → membership, 16.13.0 / 16.14.x → ranking).

Running the copy this repo installs:

| `Accept` | installed 16.9.3 | intended |
|---|---|---|
| browser default | `false` | HTML ✅ |
| `*/*`, or no header | `false` | HTML ✅ |
| `text/x-component` (RSC) | `false` | HTML ✅ |
| `text/html,text/markdown;q=0.1` | **`true`** | HTML ❌ |
| `text/html,text/plain;q=0.1` | **`true`** | HTML ❌ |

So browsers are safe, but not for the stated reason: they are safe because they never name `text/plain` or `text/markdown` at all, not because Markdown loses on q. As merged, any client that mentions `text/plain` anywhere in `Accept`, at any q, gets Markdown from these five roots.

One loose end: the preview *does* return HTML for both failing rows. I could not reconcile that with 16.9.3 running there, because `x-vercel-cache: HIT` came back for `Accept` strings that had never been requested, so the CDN answered without invoking the proxy. Worth a look at the build log to see which version Vercel actually installed; either way the code should not depend on it.

Cheapest fix, same module, no dependency movement:

```ts
import { getNegotiator } from 'fumadocs-core/negotiation';

const MARKDOWN = ['text/markdown', 'text/x-markdown', 'text/plain'];
```

```ts
const best = getNegotiator(request).mediaTypes(['text/html', ...MARKDOWN])[0];
if (!best || !MARKDOWN.includes(best)) return NextResponse.next();
```

I ran this against all twelve headers: it matches your intended model everywhere, with one nuance worth deciding on. Equal q (`text/html,text/markdown`) goes to whichever type the client lists first, so HTML here, where 16.13.0's `>=` would give Markdown.

The alternative is bumping `fumadocs-core` to `>= 16.13.0`, but note `fumadocs-ui@16.9.1` declares `fumadocs-core: 16.9.1` as an exact peer, so that is a coordinated fumadocs bump rather than a one-line change.

aromko (commented):
## Automated Code Review

Nice piece of routing work, and the guards are the right three. Confirmed working end to end on the preview: plain `GET` gives HTML, `Accept: text/markdown` gives `# Button` with `vary: Accept`, `/components/form` resolves through the collapsed slug, `__internal__` stays HTML instead of 404ing, and a section index with no generated `.md` degrades to HTML as documented. `PROXY_FILENAME`/`mod.proxy` resolution and `unstable_doesMiddlewareMatch` both check out against the shipped 16.3.0, and `test.yml` runs `pnpm --filter @marigold/docs test`, so the `vitest.config.ts` include really does put the 12 new tests in CI. Visual regression is not applicable here: nothing under `packages/**`, `themes/**` or `.storybook/**` is touched.

One blocking thing. The `markdown > 0 && markdown >= html` behaviour that the description and DST-1725 lean on landed in `fumadocs-core` **16.13.0**; the lockfile resolves **16.9.3**, where `isMarkdownPreferred` is a membership check with no ranking. Browsers still get HTML, but because they never name `text/plain` or `text/markdown`, not because Markdown loses on q. Details, the measured table and a three-line fix are in the inline comment on `proxy.ts`, plus the two assertions that would have caught it.

Three smaller notes inline: the `BROWSER_ACCEPT` comment and one test name encode the ranking model too, trailing slashes are 308-redirected before the proxy ever sees them in production, and the matcher tests would report better as `test.each`.

---

docs/proxy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    expect(getRewrittenUrl(response)).toBe(
+      'https://www.marigold-ui.io/components/actions/button.md'
+    );
+  });
+
+  test('accepts text/plain as a markdown preference', () => {
+    expect(isRewritten('/components/actions/button', 'text/plain')).toBe(true);
+  });
+
+  test('prefers markdown when it outranks html', () => {
```
aromko, line 76: ⚠️ This suite only covers headers where "mentions markdown" and "ranks markdown highest" give the same answer, which is why the version gap above slipped through. Two assertions pin the difference, and both fail on the installed 16.9.3:

```ts
test('serves html when html outranks markdown', () => {
  expect(
    isRewritten('/components/actions/button', 'text/html,text/markdown;q=0.1')
  ).toBe(false);
});

test('serves html when text/plain is only a fallback', () => {
  expect(
    isRewritten('/components/actions/button', 'text/html,text/plain;q=0.1')
  ).toBe(false);
});
```

docs/proxy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+import {
+  getRewrittenUrl,
+  isRewrite,
+  unstable_doesMiddlewareMatch,
+} from 'next/experimental/testing/server';
+import { NextRequest } from 'next/server';
+import { config, proxy } from './proxy';
+
+// What a browser actually sends. The `*/*;q=0.8` at the end is why a naive
+// "does Accept mention markdown" check would serve Markdown to every visitor.
```
aromko: ⚠️ This attributes the browser's safety to the trailing `*/*;q=0.8`, but with the installed implementation the browser is safe because its `Accept` never names a markdown media type at all: a literal "does `Accept` mention markdown" check returns `false` for this header too. Same for the name of the test at line 71, "prefers markdown when it outranks html", which describes a property 16.9.3 does not have (it passes because markdown is *mentioned*, not because it outranks). Worth rewording both once the check above is settled, so the comments do not outlive the behaviour they describe.

docs/proxy.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const pathname = request.nextUrl.pathname.replace(/\/$/, '');
```
aromko, line 16: 💡 Verified on the preview: `/components/actions/button/` comes back as a 308 to the un-slashed path, so the trailing-slash strip never runs in production. It is still worth keeping for `next start` and self-hosting, but the `NextURL` note above reads as if it is load-bearing on Vercel. One clause saying "only reachable when the trailing-slash redirect does not run first" would place it.

docs/proxy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    headers: accept ? { accept } : {},
+  });
+
+const matches = (url: string) => unstable_doesMiddlewareMatch({ config, url });
+
+const isRewritten = (path: string, accept?: string) =>
+  isRewrite(proxy(request(path, accept)));
+
+describe('matcher', () => {
+  test('covers the content areas that have generated markdown', () => {
```
aromko: 💡 Five assertions in one test means a failure reports the test name and you have to read the line number to learn which path regressed. `test.each` gives each path its own name, and `docs/lib/consent.test.ts` already uses it:

```ts
test.each([
  '/components/actions/button',
  '/foundations/spacing',
  '/getting-started/installation',
  '/patterns/user-input/forms',
  '/releases/blog/release-2026-06-09',
])('covers %s', path => {
  expect(matches(path)).toBe(true);
});
```

docs/proxy.ts
sebald: Thanks for digging into the negotiation semantics — but I think this one rests on a stale checkout. On this branch `fumadocs-core` is pinned to **16.14.3**, not `^16.6.5`/16.9.3:

- `docs/package.json` → `"fumadocs-core": "16.14.3"` (exact, no range)
- `pnpm-lock.yaml` → only `fumadocs-core@16.14.3`; no `16.6.5`, `16.9.x`, or `16.1[0-2]` entry anywhere
- installed on disk → `docs/node_modules/fumadocs-core` reports `16.14.3`
- the installed `isMarkdownPreferred` contains verbatim `return markdown > 0 && markdown >= html;`

So the ranking the PR body and DST-1725 describe *is* the code that runs. I ran your two "both fail on 16.9.3" assertions against the installed module and both pass:

```
✓ html outranks markdown    text/html,text/markdown;q=0.1 → false
✓ text/plain only fallback  text/html,text/plain;q=0.1    → false
```

The timeline also rules out a later fix: the 16.14.3 upgrade (#5738, `88cba1593`) is an ancestor of this branch from **Aug 14 12:38**, my commit is **15:31**, your review is **Aug 17 07:37**, and the head had not moved since Aug 14 — so the state you reviewed is this state. `^16.6.5` matches an older `main`, which is my guess at what got inspected.

That also settles the `fumadocs-ui` peer concern: installed `fumadocs-ui@16.14.3` declares `fumadocs-core@16.14.3` as its exact peer, so the two are already aligned — no coordinated bump needed.

Your coverage point stands on its own merits though, so I took it: ec9c24451 adds both assertions. They are green today, and that is the value — they pin the `>=` comparison so a downgrade past 16.13.0 fails in CI instead of in production.

On the preview `x-vercel-cache: HIT` you could not reconcile: with 16.14.3 running there, HTML for those two rows is the *expected* answer, so the CDN hit was masking a correct result rather than a wrong one.

docs/proxy.test.ts
sebald: The premise does not hold — both of these pass on the pinned 16.14.3 (see the version evidence in the `proxy.ts:11` thread), so they do not demonstrate a gap in behaviour.

The coverage argument is fair on its own terms though: nothing in the suite failed if a mentions-only check crept back in. Added both in ec9c24451, with a comment saying what they pin and why they are green — a downgrade past fumadocs-core 16.13.0 now fails here rather than in production.

docs/proxy.test.ts
sebald: Half right, and worth fixing either way. Under a literal "does `Accept` name a markdown type" check the browser is safe regardless, as you say. But under negotiation with provided types, `*/*` genuinely does match `text/plain`, and there the `q=0.8` is exactly what saves it. So the comment was not wrong, just ambiguous about which naive check it meant.

Reworded in ec9c24451 to name both reasons. The `prefers markdown when it outranks html` test name is accurate on 16.14.3, so I left it.

docs/proxy.ts
sebald: Good catch on the placement. Added the clause in ec9c24451: the strip is only reachable when Next own trailing-slash redirect has not run first, so it is there for `next start` and self-hosting.

For the record I did not reproduce the 308 myself — it needs a deploy or `next start` — so that part of the comment rests on your preview check.

docs/proxy.test.ts
sebald: Agreed, and the precedent is right where you point (`docs/lib/consent.test.ts:58`). Converted both five-assertion matcher tests in ec9c24451, so a failure names the path. Suite goes 12 → 22 tests, all green.

aromko (approved):

PR #5749, opened 2026-08-17
fix(DSTSUP-272): prevent tray dismissal on gestures starting in content

OsamaAbdellateef:
# Description

Fixes drag-to-dismiss on `<Tray>` incorrectly triggering when a gesture starts inside the tray's content. Previously, any vertical pointer-down on the tray (including inside `<Tray.Content>`) armed Motion's drag handler, so scrolling a list or swiping over a calendar (e.g. `DatePicker`/`DateRangePicker` open on mobile) closed the tray instead of scrolling it.

The drag gesture is now manually controlled (`useDragControls` + `dragListener={false}`) and only starts when the pointer-down does **not** originate inside the tray's content — `startsInTrayContent` is a one-liner over `target.closest('[data-tray-content]')`, using the shared internal `TRAY_CONTENT_ATTR` constant.

Disarming Motion's own listener is the other half of the fix: while it is armed, Motion forces `touch-action: pan-x` and `user-select: none` on the whole tray, and `pan-x` blocks vertical panning for the entire subtree, so the content could never scroll on touch. Both are now scoped to the chrome that can actually start a drag (drag handle, title row, actions) via `touch-none`/`select-none`, which also means text inside the tray is selectable again.

**Not in this PR:** requirement 3 of the ticket, a ≥44px touch target on the drag handle. The handle is unchanged at 48×6px; tracked separately as DST-1738. A pre-existing height bug found during review (`TrayContent`'s `min-height` pinning pushing `Tray.Actions` off screen when the viewport shrinks) is tracked as DST-1739.

No opt-in required; no changes needed in consumer code.

Closes DSTSUP-272

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. `pnpm test:sb --project=storybook-tests packages/components/src/Tray` — `ScrollableContent` drives a real touch drag through Motion's `PanSession` in Firefox and asserts the tray's own `translateY`: a downward drag inside `Tray.Content` leaves it at 0, the same drag from the chrome moves it and dismisses. It also asserts the preconditions the fix rests on (`touch-action` on the modal is not `pan-x`, content is still selectable, the chrome is `touch-none`).
2. `pnpm test:unit packages/components/src/Tray` — wiring checks on `startDrag` against a mocked Motion.
3. Manually: run `pnpm sb`, open `Components/Tray → Scrollable Content` in a mobile viewport, open the tray and swipe vertically over the list inside `Tray.Content` — the list scrolls and the tray stays open. Drag down from the handle or title row — the tray dismisses.
4. Repro from the ticket: `Components/DatePicker → Basic` on a mobile viewport, open the calendar, swipe down inside a calendar cell — the tray no longer closes.

### Mutation results for the story test

| Mutation to `TrayModal.tsx` | Story tests |
| --- | --- |
| none | 10 passed |
| whole fix removed (`dragListener={false}` + `dragControls` + `onPointerDown`) | 1 failed — `expected 'pan-x' not to be 'pan-x'` |
| only the `startsInTrayContent` guard removed | 1 failed — `expected 1002.43 to be +0` |
| `touch-none` dropped from the drag handle | 1 failed — `expected 'auto' to be 'none'` |

## Breaking Changes

No

## Checklist


.changeset/tray-content-drag-dismiss.md
```diff
@@ -0,0 +1,15 @@
+---
+'@marigold/components': patch
+'@marigold/theme-rui': patch
```
sebald: **Blocker: this bump isn't supported by the diff.**

No file under `themes/` is touched on this branch — I checked with `git diff main...HEAD --stat`, which lists only the changeset, `Tray.stories.tsx`, `Tray.test.tsx`, `TrayContent.tsx` and `TrayModal.tsx`.

Publishing a `theme-rui` patch whose changelog entry describes a touch-target change that isn't in that package means a version bump consumers can't act on and a changelog line that doesn't match the code. Either land the touch target (see my note on lines 11-13) or drop this line.

.changeset/tray-content-drag-dismiss.md
```diff
@@ -0,0 +1,15 @@
...
+---
+
+fix(Tray): don't dismiss the tray on gestures that start in its content
+
+A vertical swipe inside `<Tray.Content>` was read as drag-to-dismiss, so
+scrolling a list or swiping over a calendar closed the tray instead of
+scrolling it. including the trays `DatePicker` and `DateRangePicker` open on
+mobile. The gesture now only starts on the tray's chrome (drag handle, title,
+actions), and the drag handle's row carries a 44px touch target instead of the
+6px pill. Motion no longer sets `touch-action: pan-x` on the tray either, so the
```
sebald: **Blocker: the 44px touch target isn't in this PR.**

I opened the tray in Chromium at 390x844 and measured the handle:

```js
document.querySelector('[class*="grid-area:drag"]').getBoundingClientRect()
// => { width: 48, height: 6 }
```

`Tray.tsx:189` still renders `<div className={cn('[grid-area:drag]', classNames.dragHandle)} />` unchanged, and `Tray.styles.ts:16-18` is still `dragHandle: cva({ base: 'bg-border mx-auto mt-2 h-1.5 w-12 rounded-full' })` — byte-for-byte identical to `main`. So it's still the 6px pill the sentence says it replaced.

This is also **requirement 3** of the ticket ("with a touch target of at least 44px rather than the handle pill alone"), and the ticket calls the small handle out directly as one of its two flagged pitfalls. Worth actually doing rather than dropping — `theme-rui` already has a `ui-touch-hitbox` utility built for this (`themes/theme-rui/src/ui.css`), currently at 24px with a comment noting 44px is what WCAG 2.2 wants.

*(nit, same block)* The `touch-action: pan-x` sentence on line 13 **is** accurate, but it's non-obvious enough to deserve a code comment rather than only a changelog line — see my note on `dragListener={false}`.

*(nit)* Line 10 reads "scrolling it. including the trays `DatePicker` and ..." — stray period, and "the trays ... open" is garbled.

packages/components/src/Tray/TrayModal.tsx
```diff
@@ -44,6 +46,30 @@ const staticTransition = {
...
+    if (
+      SCROLLABLE_OVERFLOW.has(overflowY) &&
+      node.scrollHeight > node.clientHeight
+    ) {
+      return true;
+    }
+  }
+
+  return false;
+};
```
sebald: **This is the main simplification: the scrollable-ancestor branch never fires in the cases we care about, and the whole function collapses to one line.**

I verified this in the browser rather than by reading, because it's the crux of the fix. Walking up from the actual event targets:

| Case | Nearest `overflow-y:auto` ancestor | `scrollHeight > clientHeight`? | Branch that fires |
| --- | --- | --- | --- |
| DatePicker calendar cell | `Tray.Content` | **No** — `382/382` | `data-tray-content` only |
| Select `role="option"` (short list) | RAC ListBox | **No** | `data-tray-content` only |
| Long list in `Tray.Content` | `Tray.Content` | Yes — `1304/662` | both, same result |

So the overflow branch either agrees with the marker or is the reason the bug would still reproduce. It contributes nothing on its own. That's structural, not accidental: `ui-surface-content` (`themes/theme-rui/src/ui.css:333`) puts `overflow-y-auto` on the *same element* that now carries `data-tray-content`, and nothing in the chrome (`drag`/`title`/`actions` areas) has any overflow at all.

The `boundary` parameter is also unnecessary — the tray is portalled through `ModalOverlay`, and separate `<Tray>` instances portal to sibling subtrees, so `closest()` can't reach another tray's content marker.

```suggestion
const startsInTrayContent = (target: Element) =>
  target.closest('[data-tray-content]') !== null;
```

...with `startDrag` becoming a 3-line function (see my note below). That drops `SCROLLABLE_OVERFLOW`, the manual walk, the `getComputedStyle` call and the `currentTarget` argument, with behaviour identical on every case I measured.

**If you'd rather keep a genuine scroll check** as a safety net for consumers who skip `<Tray.Content>`, please don't hand-roll it — `@react-aria/utils` is already a direct dependency of this package (`packages/components/package.json:65`) and exports the same thing, more thoroughly (it tests `overflow` + `overflowX` + `overflowY` and both axes, where this checks only `overflowY`):

```ts
import { getScrollParent } from '@react-aria/utils';

const scrollParent = getScrollParent(target, true);
if (boundary.contains(scrollParent) && scrollParent !== boundary) return;
```

A second, subtly narrower scroll-detection implementation in the codebase is the real cost here: any upstream fix react-aria makes for a browser quirk won't reach this copy.

*(nit)* `startsInScrollableBody` is misnamed either way — it returns `true` for content that isn't scrollable at all (both DatePicker and Select above), which is precisely the behaviour we want but not what the name promises.

*(nit)* A `new Set` for two string literals is heavier than `overflowY === 'auto' || overflowY === 'scroll'`, moot if the branch goes.

packages/components/src/Tray/TrayModal.tsx
```diff
@@ -44,6 +46,30 @@ const staticTransition = {
...
+  for (
+    let node: Element | null = target;
+    node && node !== boundary;
+    node = node.parentElement
+  ) {
+    if (node.hasAttribute('data-tray-content')) {
+      return true;
+    }
+
+    const { overflowY } = getComputedStyle(node);
```
sebald: **Performance: this runs on the first event of every gesture and forces both style recalc and layout.**

`getComputedStyle(node)` forces a style recalc, and reading `node.scrollHeight`/`node.clientHeight` forces a **layout flush** — on `pointerdown`, the single most latency-sensitive moment of a drag or scroll, where a stall costs the first frame.

Concrete worst case, traced through the rendered DOM for a DatePicker tray. Between `[data-tray-content]` and a calendar day cell there are 6 intermediate elements — RAC Calendar root -> CalendarBody wrapper -> `table` -> `tbody` -> `tr` -> `td` -> target `div` — so **every day-cell tap costs ~7 `getComputedStyle` calls and ~14 forced layout reads**, none of which can return `true` (none of those nodes is scrollable), to re-derive an answer the marker 7 levels up already had. Add a level for `CalendarPresetsShell` or `visibleDuration.months > 1`.

`Element.closest()` is a pure DOM-tree walk in native code with zero style or layout cost. Folding into the note above.

packages/components/src/Tray/TrayModal.tsx
```diff
@@ -56,6 +82,15 @@ export const TrayModal = ({
...
   const y = useMotionValue(h);
+  const dragControls = useDragControls();
+
+  const startDrag = (event: ReactPointerEvent<Element>) => {
+    if (startsInScrollableBody(event.target as Element, event.currentTarget)) {
+      return;
+    }
+
+    dragControls.start(event);
+  };
```
sebald, line 77: **Worth a design conversation: this is a blacklist where `useDragControls` is built for an allowlist, and it fails open.**

`useDragControls` + `dragListener={false}` exists so you can call `start()` from **a handle's own** `onPointerDown` — that's the documented use. Here it's reattached to the same broad root that `dragListener={true}` would have used, with a custom DOM-walking blacklist substituted for motion's built-in text-input blacklist. Same shape as before, different exclusion list.

The chrome is already individually addressable: `Tray.tsx:189` renders a dedicated `[grid-area:drag]` div, and `TrayTitle`/`TrayHeader`/`TrayActions` are separate components. Attaching `onPointerDown={e => dragControls.start(e)}` to that small fixed set instead would:

- need no cross-file string contract and no DOM walk at all (also kills the perf note above);
- **fail safe** — anything a consumer composes that isn't recognised chrome simply has no drag listener, so drag-to-dismiss defaults to *off*, which is the right default for an overlay;
- match the ticket's own primary suggestion, which is framed in terms of `[grid-area:drag]` plus `[grid-area:title]`.

The current blacklist fails the other way: drag is *on* everywhere except what the predicate recognises. Since it only recognises `Tray.Content` and native `overflow` scrolling, a consumer dropping a swipeable carousel, a slider, or any custom gesture widget directly under `<Tray>` re-opens exactly this bug. There's also no escape hatch in either direction — no way to opt a custom region out, or to opt content-area drag back in.

Not asking you to rewrite it in this PR if you'd rather ship the fix — the behaviour I measured is correct. But please pick deliberately, because the blacklist is the version that needs extending every time someone puts something new in a tray.

Also, for requirement 9: the decision here is made once at `pointerdown` and is permanent for the gesture, so the native scroll-to-top-then-keep-dragging handoff can't work — the user has to lift and start again on the chrome. That's a fine scope cut, but this file already has a precedent for writing such things down (the reduced-motion/FocusScope comment on lines 102-106 ends with "follow-up needed for a full fix"). A comment here saying the same would stop it reading as a closed question.

packages/components/src/Tray/TrayModal.tsx
```diff
@@ -101,8 +136,11 @@ export const TrayModal = ({
               animate={{ y: 0 }}
               exit={{ y: h }}
               transition={staticTransition}
-              style={{ y }}
+              style={{ y, userSelect: 'none', WebkitUserSelect: 'none' }}
```
sebald: **This is now mis-scoped — it kills text selection in content that can no longer be dragged.**

Motion only applies these itself when drag is armed on the element (`framer-motion/dist/es/render/html/use-props.mjs`):

```js
if (props.drag && props.dragListener !== false) {
    htmlProps.draggable = false
    style.userSelect = style.WebkitUserSelect = style.WebkitTouchCallout = "none"
    style.touchAction = props.drag === true ? "none" : `pan-${props.drag === "x" ? "y" : "x"}`
}
```

With `dragListener={false}` that guard is now false, which is what these two lines compensate for. But blanket `userSelect: 'none'` on the whole `MotionModal` was only *correct* while the whole tray was draggable. Now that drags can only start on the chrome, a user can no longer select or copy text inside `Tray.Content` — a date, an address, a reference number — for no remaining reason. I confirmed `user-select: none` is computed on the content element in the browser.

Two smaller things in the same change:

- It restores **2 of the 4** things the guard dropped. `WebkitTouchCallout: 'none'` and `draggable: false` are gone. The callout is arguably good to lose now, but `draggable` means images and links in a tray can start a native HTML5 drag with a ghost image.
- `'none'` string literals in `style` are the sort of thing CLAUDE.md's no-inline-styles rule targets. `style={{ y }}` is unavoidable (it's the motion value binding); these aren't.

Suggest dropping both from here and putting `select-none` on the three chrome pieces that can actually start a drag — `Tray.tsx:189` (`[grid-area:drag]`), `TrayHeader`/`TrayTitle` (`[grid-area:title]`), `TrayActions` (`[grid-area:actions]`) — in line with how z-index is applied in component implementations rather than theme files.

```suggestion
              style={{ y }}
```

Worth knowing: vaul goes further in the opposite direction — it *aborts* a drag when `window.getSelection()` is non-empty, rather than suppressing selection to protect the drag.

sebald (commented):
## Review: verified in a real browser, plus one bug this PR didn't cause

I ran the fix through Chromium at 390x844 with real touch events (CDP `Input.dispatchTouchEvent`), read DSTSUP-272 requirement by requirement, and checked whether this behaviour should be ours to own at all. Short version: **the fix works and the approach is right — but the changeset describes a change that isn't in the diff, nothing automated covers the actual fix, and the predicate is ~20 lines where 3 would do.**

### 1. It fixes the ticket — measured, not assumed

| Test | Result |
| --- | --- |
| Touch swipe **up** inside `Tray.Content` | `scrollTop 0 -> 326` — **scrolls** |
| Touch swipe **down** 250px inside content | stays open, `scrollTop 0` |
| Mouse drag down 300px inside content | stays open |
| **Ticket repro:** 240px down-swipe on a DatePicker calendar cell | **stays open** (was dismissing) |
| Select tray: down-swipe on a `role="option"` | stays open |
| Drag from the chrome (handle) | dismisses |
| `touch-action` on the modal | `auto` (was `pan-x`) — requirement 2 |

One finding worth carrying into the code: **`data-tray-content` is doing all the work; the scrollable-ancestor branch never fires in the real cases.** DatePicker's calendar content measures `382/382` (not scrollable) and Select's ListBox is `overflow-y: auto` but `scrollable: false` with a short list. In both, the overflow branch returns `false` and only the marker saves us. Details inline on `TrayModal.tsx`.

### 2. Requirements scorecard (DSTSUP-272)

- **Pass:** 1, 2, 4, 7, 8. 5 passes for mouse; reduced-motion still has no drag at all (unchanged — fine, but see below).
- **Fail: 3.** I measured the drag handle at **48x6px**. The changeset and PR description both claim a 44px touch target; it is not in the diff. `themes/theme-rui/src/components/Tray.styles.ts` is untouched.
- **Not done: 9** — explicitly optional in the ticket. Fine to defer, but please leave a marker in the code (inline suggestion below).

### 3. Should we own this at all? Yes.

I checked the alternatives properly, since it's a fair question:

- **React Aria has no helper.** `TrayModal` is a direct fork of RAC's official *Gesture Driven Sheet* example — identical `inertiaTransition` (300/40/300), `cubicBezier(0.32, 0.72, 0, 1)`, `dragConstraints={{ top: 0 }}`, `offset.y > innerHeight * 0.75 || velocity.y > 10`, even the `mx-auto w-12 mt-2 h-1.5` pill. **That example does not address scroll-vs-drag at all.** RAC 1.20 ships no Tray/Sheet primitive; `useMove` normalises deltas but knows nothing about scroll; `useDrag`/`useDrop` are data-transfer DnD.
- **vaul doesn't compose.** It's built on Radix Dialog. `Tray` depends on RAC's `Modal`/`ModalOverlay`/`FocusScope`, `useIsHidden` collection interop, `ResetButtonContext`, and the DSTSUP-261 hidden-pass guard. Adopting it means a second overlay/focus stack in the bundle, or reimplementing all of that.
- **One thing we should borrow:** `@react-aria/utils@^3.34.1` is already a direct dependency of `@marigold/components` and exports `isScrollable(node, checkForOverflow)` and `getScrollParent` — the exact parent-walk this PR hand-rolls, and more thorough (checks `overflow` + `overflowX` + `overflowY`, both axes).

So ~40 lines of our own is the right call. See the inline note for how to make it the right 40.

**For reference, how vaul solves the harder version** (its `shouldDrag`, evaluated on *pointermove*, not pointerdown): a `SELECT` tag bail; a `data-vaul-no-drag` public opt-out; a 500ms lockout during the open animation; a sticky "already dragging" check; a `window.getSelection()` check so it never fights text selection; a **100ms `scrollLockTimeout`** after any scroll; a direction check; then an upward walk testing `scrollHeight > clientHeight` **and `scrollTop !== 0`**, with `role="dialog"` as the sentinel that stops the walk. Plus an `isAllowedToDrag` latch and an iOS `touchend` listener (iOS doesn't fire `mouseup` after a scroll). The two bits hand-rolled versions always miss are the scroll-lock timeout and the once-true latch. Notably vaul also ships `handleOnly` — **exactly this PR's strategy** — as the simple tier next to the scroll-aware default. So this PR is a legitimate tier, just the coarser of vaul's two.

### 4. A separate bug I reproduced — not caused by this PR

`TrayContent`'s `min-height` pinning freezes the content height at mount. When the viewport then shrinks — rotation, keyboard opening in a ComboBox tray, mobile browser UI appearing — the tray box shrinks but the content row cannot:

| Viewport height | Tray box | Content `clientHeight` | `Tray.Actions` |
| --- | --- | --- | --- |
| 844 | 802 | 662 (pinned) | at viewport bottom |
| 500 | 475 | **662** | **327px below the fold** |
| 390 (landscape) | 371 | **662** | **432px below the fold** |

Confirm/Close leave the screen and you can't scroll to them — the tray container is `overflow: visible`, and the inner content scroll doesn't move the actions row. From a user's seat this reads as "the tray is broken when the content is long", so it may well be part of what's being reported alongside DSTSUP-272. **Suggest a sibling ticket rather than growing this PR.** Related: `Tray` uses `max-h-[95vh]` while `Drawer` already uses `h-(--visual-viewport-height)`; `vh` is the *large* viewport on mobile, so the same clipping is expected whenever browser UI is visible. I could not reproduce that variant headless (all of `vh`/`dvh`/`svh`/`lvh` resolved to 802), so treat that half as a hypothesis to confirm on a device — the resize half above is reproduced.

### 5. On the height model — current design is right, keep it

Worth stating since it came up: the tray **sizes to content and caps at the viewport, with the inner content area scrolling**. Verified — the DatePicker tray measured 522px tall, not the 802px cap. That matches iOS `UISheetPresentationController` detents and Material 3 modal bottom sheets, and it's the reason to *not* move to a "whole tray scrolls, extends above the viewport" model:

1. `Tray.Actions` must stay pinned — Select/DatePicker put Confirm/Cancel there, and a scrolling sheet scrolls the primary action away (which is exactly the failure in section 4).
2. The drag handle must stay put, or the dismiss affordance moves under the user's thumb.
3. Drag-to-dismiss and page-scroll would become the same gesture on the same element — this bug, generalised to the whole sheet.

The full-screen-takeover model already exists as `Drawer`. (vaul, for what it's worth, sets no `max-height` in JS at all — height is the consumer's CSS, detents come from snap points.)

### Blockers before merge

1. **The changeset claims a 44px touch target that doesn't exist and bumps `@marigold/theme-rui` for it.** Either implement it — `theme-rui` already has a `ui-touch-hitbox` utility for exactly this — or strike both. Requirement 3 asks for it explicitly.
2. **Nothing automated covers the fix.** The unit tests assert `dragControls.start` was called against a fully mocked motion; they'd pass if the tray never scrolled. The new `ScrollableContent` story is the natural home for a real gesture test.
3. **Document the interim decision** — the pointerdown-time choice can't support the continuous scroll-to-top-then-drag handoff of requirement 9.

Everything else is inline. Non-blocking notes are marked *(nit)*. Lint and Prettier are clean, and all 10 Tray unit tests pass on this branch.

packages/components/src/Tray/TrayModal.tsx
```diff
@@ -101,8 +136,11 @@ export const TrayModal = ({
...
               drag="y"
+              dragListener={false}
```
sebald, line 131: *(nit)* Please add a comment here. This one prop silently changes four things via motion's internals (`draggable`, `userSelect`, `WebkitTouchCallout`, `touchAction`) — and dropping `touch-action: pan-x` is *the* reason content can pan vertically on touch, i.e. half the actual fix. I confirmed `touch-action: auto` on the modal in the browser, and requirement 2 of the ticket depends on it.

Right now that only lives in the changeset. Anyone later wondering why there's a bare `dragControls` with no handle element, or tempted to "tidy up" by restoring `dragListener`, would silently reintroduce the scroll bug with no test to catch it (see my note on `Tray.test.tsx`).

Something like:

```ts
// `dragListener={false}` is what lets content scroll: motion only sets
// `touch-action: pan-x` (plus `user-select: none`, `draggable`) on the dragged
// element while its own listener is armed, and `pan-x` blocks vertical panning
// for the whole subtree — a descendant can't re-allow it. Drag is started
// manually from `startDrag` instead. See DSTSUP-272.
```

packages/components/src/Tray/TrayContent.tsx
```diff
@@ -28,6 +28,7 @@ export const TrayContent = ({ children, className }: TrayContentProps) => {
           node.style.minHeight = `${node.offsetHeight}px`;
```
sebald, line 28: **Separate bug, pre-existing (DST-900) — flagging it here because it lands in the same file and reads to users as "the tray is broken with long content".**

This pins `min-height` to the height measured at mount. If the viewport later shrinks, the tray box shrinks but this row can't, so the grid overflows a container that is `overflow: visible` — and `Tray.Actions` walks off the bottom of the screen. Reproduced by resizing with the tray open (`ScrollableContent`, 40 items):

| Viewport height | Tray box | Content `clientHeight` | `Tray.Actions` |
| --- | --- | --- | --- |
| 844 | 802 | 662 (pinned) | at viewport bottom |
| 500 | 475 | **662** | **327px below the fold** |
| 390 (landscape) | 371 | **662** | **432px below the fold** |

Confirm/Close become unreachable, and the inner content scroll doesn't help because it doesn't move the actions row. Real triggers: device rotation, the keyboard opening in a ComboBox/Autocomplete tray, and mobile browser UI appearing.

Related, same area: `Tray.styles.ts` caps at `max-h-[95vh]` while `Drawer` uses `h-(--visual-viewport-height)` (`DrawerModal.tsx:12`). `vh` is the *large* viewport on mobile, so the tray can be laid out taller than what's actually visible. I could not reproduce that variant headless — `vh`/`dvh`/`svh`/`lvh` all resolved to 802 in Chromium — so it needs a device check, but the resize case above is reproduced and the inconsistency with `Drawer` looks unintentional.

**Not asking you to fix it here** — it's out of DSTSUP-272's scope and the ticket explicitly scopes the height model out. A sibling ticket would be good. When it's picked up, the fix is likely a `min-h-` floor on the grid row (the pinning exists to stop the tray resizing as a filtered list changes length, which a floor also achieves) rather than freezing an absolute pixel value, plus the visual-viewport unit.

packages/components/src/Tray/TrayContent.tsx
```diff
@@ -28,6 +28,7 @@ export const TrayContent = ({ children, className }: TrayContentProps) => {
...
         }
       }}
+      data-tray-content
```
sebald: *(nit)* This creates an untyped contract between two files: the string is written here and read in `TrayModal.tsx` with `hasAttribute('data-tray-content')`, with no shared constant. A rename or typo in either place silently re-opens the bug with nothing failing at compile time — and per my other note, nothing failing at test time either.

Two reasonable directions:

1. **Keep the attribute but own it.** It's a defensible mechanism — vaul does the same with `data-vaul-no-drag` — but then hoist it to a shared `const TRAY_CONTENT_ATTR` and document it, since it's now real markup that consumer CSS and e2e selectors can key off.
2. **Use the context you already have.** `Tray/Context.tsx` is the house pattern for exactly this and already carries `titleSlotRef` via `useSlot()`. A `contentRef` alongside it, checked with `contentRef.current?.contains(event.target as Node)`, keeps it internal and type-safe.

I'd lean (1) if you want it to be a public opt-out marker (which would also give consumers the escape hatch the ticket asks for), (2) if it should stay private. Either beats a bare string in two files.

packages/components/src/Tray/Tray.test.tsx
```diff
@@ -155,6 +158,36 @@ test('small drag snaps tray back', async () => {
   expect(mockAnimate).toHaveBeenCalled();
 });
 
+const openTray = async () => {
+  renderWithOverlay(<Basic.Component />);
+  await user.click(screen.getByRole('button', { name: 'Open Tray' }));
+  await waitFor(() => {
+    expect(screen.getByRole('dialog')).toBeInTheDocument();
+  });
+};
```
sebald: *(nit)* Half-done extraction. This helper is byte-for-byte the first three lines of three tests directly above it — `'drag down far enough dismisses the tray'` (113-118), `'fast swipe dismisses the tray'` (130-135), `'small drag snaps tray back'` (146-151) — none of which were changed to use it.

So the file now has two representations of "open the Basic tray", and a future change to the open sequence has to be made twice with no signal which is canonical. Moving `openTray` above line 113 and collapsing those three would leave one.

packages/components/src/Tray/Tray.test.tsx
```diff
@@ -155,6 +158,36 @@ test('small drag snaps tray back', async () => {
...
+test('starts a drag from the tray chrome', async () => {
+  await openTray();
+
+  await user.pointer({
+    keys: '[MouseLeft>]',
+    target: screen.getByText('Tray Title'),
+  });
+
+  expect(mockDragControls.start).toHaveBeenCalled();
+});
```
sebald, line 177: **These pass without the bug being fixed — they assert the mock, not the behaviour.**

`motion/react` is mocked wholesale at the top of this file, so `useDragControls` returns `mockDragControls` and `dragControls.start` is a `vi.fn()` that does nothing. What's verified is "`startDrag` called `start()`" — a direct restatement of the four lines under test. Nothing here exercises what the ticket is actually about:

- that `Tray.Content` **scrolls** on a touch swipe (this depends on `touch-action` no longer being `pan-x`, which is a motion-applied inline style the mock never produces);
- that the tray survives a real down-swipe rather than `start()` merely going uncalled;
- the `getComputedStyle`/`scrollHeight` branch at all — jsdom reports `scrollHeight === clientHeight === 0`, so it can't ever fire here.

Concretely: revert `TrayModal.tsx` to `dragListener` armed and drop `startDrag`, and I'd expect `'starts a drag from the tray chrome'` to still pass, because motion's real listener isn't in play either way. That's the coverage gap behind blocker 2 — the regression guard for a gesture bug needs to be a gesture test.

These are still worth keeping as cheap wiring checks; they just shouldn't be the only thing standing between us and a repeat of DSTSUP-272. See my note on `ScrollableContent` for where the real test belongs.

packages/components/src/Tray/Tray.stories.tsx
```diff
@@ -274,6 +274,28 @@ SlotPrimitives.test(
   }
 );
 
+export const ScrollableContent = meta.story({
+  parameters: { chromatic: { disableSnapshot: true } },
```
sebald, line 281: **This story is the right idea but currently carries no automated weight — and it's the natural home for the missing regression test.**

As written it has `chromatic: { disableSnapshot: true }`, no `tags: ['component-test']`, and no attached `.test()`. So `pnpm test:sb` never runs it and Chromatic never snapshots it: it's a manual demo only. Meanwhile the unit tests that nominally cover this PR run against a fully mocked motion (see my note on `Tray.test.tsx`), so **right now nothing automated proves the fix**.

Story tests run in a real browser via Playwright, which is exactly what this needs — I verified the fix that way by hand and the assertions are straightforward:

```ts
export const ScrollableContent = meta.story({
  tags: ['component-test'],
  render: args => (
    // ...unchanged
  ),
});

ScrollableContent.test(
  'scrolls its content instead of dismissing',
  async ({ canvas, userEvent, step }) => {
    await userEvent.click(canvas.getByRole('button', { name: 'Open Tray' }));
    const dialog = await canvas.findByRole('dialog');
    const content = dialog.querySelector('[data-tray-content]')!;

    await step('a swipe inside the content scrolls it', async () => {
      await userEvent.pointer([
        { keys: '[TouchA>]', target: content, coords: { x: 150, y: 400 } },
        { pointerName: 'TouchA', coords: { x: 150, y: 200 } },
        { keys: '[/TouchA]' },
      ]);
      await expect(content.scrollTop).toBeGreaterThan(0);
      await expect(dialog).toBeInTheDocument();
    });
  }
);
```

For reference, the numbers I got driving this story manually at 390x844: content `clientHeight` 662 / `scrollHeight` 1304, and a 200px upward touch swipe moved `scrollTop` from 0 to 326 while the dialog stayed mounted. A down-swipe from `scrollTop: 0` left it at 0 and open.

Two more worth adding while you're here, both of which the current unit tests can't express: a **down**-swipe on the content asserting the dialog is still mounted (the actual reported symptom), and a drag from `[grid-area:drag]` asserting it *does* dismiss (so the fix can't regress into disabling drag altogether).

*(nit)* If the story stays snapshot-disabled and untested, it'd be more honest to drop it — `Basic` already demonstrates the component. But I'd much rather it grew the test.

.changeset/tray-content-drag-dismiss.md
OsamaAbdellateef: Dropped the `theme-rui` bump — nothing under `themes/` is touched now that we're not landing the 44px touch target in this PR (see reply below). Changeset is `@marigold/components` only now.

.changeset/tray-content-drag-dismiss.md
OsamaAbdellateef: Fixed the typo. On the 44px target: decided not to implement it in this PR — DSTSUP-272 is an external support ticket rather than a binding product requirement, so I'm not treating its "requirement 3" as a spec we have to satisfy here. Reworded the changeset to only describe what's actually in the diff (drag detection + `touch-action`), dropped the touch-target claim. Happy to pick up the bigger hit target as its own ticket — `ui-touch-hitbox` is a reasonable starting point, though I'd want to scope it to the Tray handle rather than bump the shared utility, since that's also used by Button/Tag/Collapsible/DatePicker/ContextualHelp.

packages/components/src/Tray/TrayModal.tsx
OsamaAbdellateef: Done — replaced `startsInScrollableBody` with `startsInTrayContent`, a one-liner using `target.closest('[data-tray-content]') !== null`. Dropped the `boundary` param and the `getComputedStyle`/`scrollHeight` branch entirely, per your table.

packages/components/src/Tray/TrayModal.tsx
OsamaAbdellateef: Resolved by the same simplification above — `closest()` is a pure DOM-tree walk, no style recalc or layout cost.

packages/components/src/Tray/TrayModal.tsx
OsamaAbdellateef: Kept the denylist for now rather than restructuring to an allowlist on the individual chrome pieces — didn't want to expand this PR's surface further. Added a comment above `startsInTrayContent` documenting the tradeoff explicitly (a custom widget outside `Tray.Content` re-opens the bug, no per-region opt-out in either direction), so it reads as a deliberate choice rather than an oversight. Open to revisiting as a follow-up if it turns out to matter in practice.

packages/components/src/Tray/TrayModal.tsx
OsamaAbdellateef: Done — dropped `userSelect`/`WebkitUserSelect` from the modal's `style` (back to `style={{ y }}`) and added `select-none` directly to the three chrome pieces instead: the drag handle in `Tray.tsx`, the title row in `TrayHeader.tsx`, and the actions row in `TrayActions.tsx`. Content is selectable again.

packages/components/src/Tray/TrayModal.tsx
OsamaAbdellateef: Added, close to what you suggested — explains why `dragListener={false}` is what lets content scroll and points back to DSTSUP-272.

packages/components/src/Tray/TrayContent.tsx
OsamaAbdellateef: Agreed this is out of scope here — not touching it in this PR. Thanks for writing up the repro, that'll be useful for whoever picks up a follow-up on the height model.

packages/components/src/Tray/TrayContent.tsx
OsamaAbdellateef: Went with option 1 — hoisted the string to `TRAY_CONTENT_ATTR` in `Context.tsx`, documented there. Both `TrayContent` (sets it) and `TrayModal` (reads it via the new `startsInTrayContent`) import it from that single source.

packages/components/src/Tray/Tray.test.tsx
OsamaAbdellateef: Done — moved `openTray` above the three drag tests and switched them to use it. One representation of "open the Basic tray" now instead of two.

packages/components/src/Tray/Tray.test.tsx
OsamaAbdellateef: Agreed, and addressed via the new story test rather than here (see reply on `ScrollableContent` below) — these unit tests stay as cheap wiring checks on top of that.

packages/components/src/Tray/Tray.stories.tsx
OsamaAbdellateef: Added `tags: ['component-test']` and a `.test()` that runs in a real browser (Playwright) and swipes inside `Tray.Content` in both directions, asserting the dialog survives — this is the test that would actually fail if the fix regressed. Two things I couldn't get working with `userEvent.pointer` specifically, both synthetic-event limitations rather than gaps in the fix: asserting `scrollTop` moves (synthetic PointerEvents don't drive real touch-scroll physics in this harness), and asserting a drag from the handle still dismisses (motion's `PanSession` didn't register the gesture from either a synthetic touch or mouse sequence here). Documented both limitations in a comment on the test; that half still needs manual/device verification, same as what you did by hand.

packages/components/src/Tray/Tray.stories.tsx
```diff
@@ -274,6 +274,60 @@ SlotPrimitives.test(
...
+        </Tray.Content>
+        <Tray.Actions>
+          <Button slot="close">Close</Button>
+        </Tray.Actions>
+      </Tray>
+    </Tray.Trigger>
+  ),
+});
+
+ScrollableContent.test(
```
sarahgm, line 345: 🔴 **This test cannot fail. Blocker 2 from the last round is still open.**

I mutation-tested it locally:

| Mutation to `TrayModal.tsx` | `Tray.stories.tsx` | `Tray.test.tsx` |
| --- | --- | --- |
| none (branch as-is) | 9 passed | 10 passed |
| whole fix removed (`dragListener={false}` + `dragControls` + `onPointerDown`) | **9 passed** | 1 failed |
| only the `startsInTrayContent` guard removed, so the bug is fully back | **9 passed** | 1 failed |
| guard removed **and** this test's downward-swipe assertion inverted to `expect(canvas.queryByRole('dialog')).not.toBeInTheDocument()` | **dialog still present** | n/a |

The last row is the diagnosis. With the bug fully restored, a 700px downward synthetic touch swipe inside the content does not dismiss the tray, so nothing was ever dragged. Both `expect(...).toBeInTheDocument()` calls here are tautologies, not assertions that merely fire too early.

That is the same limitation you described for the handle-dismiss assertion you dropped, and it applies just as much to the swipe you kept. So this does not replace the unit tests, it repeats them one layer up. Credit where it is due: the unit tests do catch both wiring mutations above, which is more than I gave them last round. What is still unguarded is the behaviour.

Two ways out:

- drive real touch, for example Playwright's `page.touchscreen` in a context with `hasTouch: true`, which does go through the browser's gesture arbitration;
- or assert the observable preconditions of the fix instead of the gesture: computed `touch-action` on the modal is not `pan-x`, and `content.scrollTop` moves after a scroll. Both break if the fix is reverted, and both are reachable with the current harness.

*(also here)* Your reply said "documented both limitations in a comment on the test". There is no comment on this test in the branch. See my note on `TrayModal.tsx:50`.

packages/components/src/Tray/Tray.stories.tsx
```diff
@@ -274,6 +274,60 @@ SlotPrimitives.test(
...
+  'a swipe that starts inside the content never dismisses the tray',
+  async ({ canvas, step }) => {
+    await userEvent.click(canvas.getByRole('button', { name: 'Open Tray' }));
+    const dialog = await waitFor(() => canvas.getByRole('dialog'));
+    const content = dialog.querySelector('[data-tray-content]') as HTMLElement;
```
sarahgm: ⚠️ The `'[data-tray-content]'` string is hardcoded here. `TRAY_CONTENT_ATTR` was introduced in this PR precisely to remove the untyped cross-file contract, and this re-creates it in a third file.

```suggestion
    const content = dialog.querySelector(
      `[${TRAY_CONTENT_ATTR}]`
    ) as HTMLElement;
```

with `import { TRAY_CONTENT_ATTR } from './Context';` at the top.

packages/components/src/Tray/Tray.stories.tsx
```diff
@@ -274,6 +274,60 @@ SlotPrimitives.test(
   }
 );
 
+export const ScrollableContent = meta.story({
+  tags: ['component-test'],
+  parameters: { chromatic: { disableSnapshot: true } },
```
sarahgm, line 281: 💡 With `disableSnapshot` on the story, a tray whose content actually overflows is never snapshotted, and this is the only story where that happens. That is a visual state worth guarding, especially since the fix touches how the content area behaves.

CLAUDE.md's pattern keeps per-test parameters in the `.test()` options so the story itself stays snapshottable:

```ts
export const ScrollableContent = meta.story({
  tags: ['component-test'],
  render: args => (
    // ...unchanged
  ),
});

ScrollableContent.test(
  'a swipe that starts inside the content never dismisses the tray',
  { parameters: { chromatic: { disableSnapshot: true } } },
  async ({ canvas, step }) => {
    // ...
  }
);
```

packages/components/src/Tray/TrayModal.tsx
```diff
@@ -44,6 +47,9 @@ const staticTransition = {
   ease: cubicBezier(0.32, 0.72, 0, 1),
 };
 
+const startsInTrayContent = (target: Element) =>
```
sarahgm, line 54: ⚠️ **The comment you described in your reply is not in the branch.**

Your reply on the denylist thread says a comment was added above `startsInTrayContent` documenting the tradeoff explicitly. `grep -n '^\s*//' packages/components/src/Tray/TrayModal.tsx` returns only the three comments that were already on `main` (lines 33, 82, 106), and nothing anywhere in `packages/components/src/Tray/` mentions DSTSUP-272, the denylist tradeoff, `touch-action`, or the story-test limitations.

The same holds for three sibling replies in this round:

- the `dragListener={false}` explanation, "added, close to what you suggested" (see `:121`);
- the story-test limitations, "documented both limitations in a comment on the test" (see `Tray.stories.tsx:300`);
- `TRAY_CONTENT_ATTR` being "documented there" (`Context.tsx:26` is a bare export).

Keeping the denylist is your call and I am not reopening it. The reason the comment matters is that the next person to read this thread will trust the reply rather than re-diff the file, and the tradeoff then reads as an oversight again. The requirement 9 marker was part of the same ask.

packages/components/src/Tray/TrayModal.tsx
```diff
@@ -103,6 +118,9 @@ export const TrayModal = ({
               transition={staticTransition}
               style={{ y }}
               drag="y"
+              dragListener={false}
```
sarahgm, line 131: ⚠️ **Nothing declares `touch-action` any more, and the comment for this prop is also missing.**

Dropping motion's own listener is what removes the inline `touch-action: pan-x`, which is the good half of this fix and what requirement 2 depends on. It also removes the only declaration telling the browser that a vertical gesture on the chrome is JS-owned. There is no `touch-action` anywhere in `Tray` now.

While the tray is open, RAC's `usePreventScroll` locks the page and the chrome has no scrollable ancestor, so gesture arbitration should leave the drag alone. But that is implicit rather than stated, and it is unverified on real hardware: the harness cannot drive a real gesture (see `Tray.stories.tsx:300`) and CDP-dispatched touch is not a finger. `touch-none` on the three chrome areas would state the intent, and would mirror exactly how you scoped `select-none` in this PR.

*(nit)* `draggable={false}`, one of the four things motion's guard used to set, is still not restored, so images and links in a tray can start a native HTML5 drag. Raised last round, not answered.

packages/components/src/Tray/Context.tsx
```diff
@@ -22,3 +22,5 @@ export const TrayContext = createContext<TrayContextProps>(
 );
 
 export const useTrayContext = () => use(TrayContext);
+
+export const TRAY_CONTENT_ATTR = 'data-tray-content';
```
sarahgm, line 31: ⚠️ Hoisting this was the right move and it does remove the duplicated string inside `src`. Two loose ends:

- Your reply says it is documented here. It is a bare export with no comment. Documenting it was the point of option 1, since this is now real markup that consumer CSS and e2e selectors can key off.
- It is not re-exported from the package (`src/index.ts` exports only `Tray` and `TrayProps`), so it cannot serve as the public opt-out marker option 1 described, and the ticket's request for an escape hatch is still unmet.

Pick one deliberately: document it as internal, or export it and mention it in the changeset. Also worth a word on why it lives in `Context.tsx`, since it is not context.

packages/components/src/Tray/Tray.tsx
```diff
@@ -186,7 +186,12 @@ export const Tray = ({
               classNames.container
             )}
           >
-            <div className={cn('[grid-area:drag]', classNames.dragHandle)} />
+            <div
+              className={cn(
+                'select-none [grid-area:drag]',
```
sarahgm: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files but the `Visual-Regression-Tests` workflow has not run for this branch. `gh run list --workflow=visual-regression-tests.yml --branch=dstsup-272_fix-tray-drag` returns an empty list, and the checklist item is unticked.

I expect no diffs, since the drag wiring is non-visual and the new story is snapshot-disabled, but this line is a style change on the drag handle and the run is what confirms that.

.changeset/tray-content-drag-dismiss.md
```diff
@@ -0,0 +1,13 @@
+---
+'@marigold/components': patch
+---
+
+fix(Tray): don't dismiss the tray on gestures that start in its content
```
sarahgm, line 5: ⚠️ The changeset is accurate now and the `theme-rui` bump is gone, thanks. I verified no file under `themes/` is touched, and the `theme-rui` and `icons` rows in the changeset-bot comment are dependent bumps rather than something this changeset asks for.

Two things did not follow it:

- **The PR description still claims the 44px hit target.** It says the drag handle's hit target "was also widened to 44px (up from a 6px pill)", and it still describes `startsInScrollableBody` bailing on "any scrollable ancestor", which is the code your second commit deleted. Since requirement 3 is deliberately out, the description is now the only place claiming it shipped, and the reporter is tracking requirements on DSTSUP-272.
- **Requirement 3 has no follow-up ticket.** Declining it here is a fair call and scoping `ui-touch-hitbox` to the handle rather than bumping the shared utility is the right instinct, but with nothing filed it just gets lost.

*(nit)* The title reads `fix([DST-272])`. The ticket is `DSTSUP-272`, and `DST-272` resolves to a different project. Convention here is `fix(DSTSUP-272): ...` with no brackets, and the title becomes the squash commit subject.

sarahgm (changes requested):

.changeset/tray-content-drag-dismiss.md
sarahgm: Verified and resolving. `gh pr diff` and `git diff origin/main -- themes/` both show no file under `themes/` on this branch, and the changeset is `@marigold/components` only. The `theme-rui` and `icons` rows in the changeset-bot comment are dependent bumps rather than something the changeset asks for.

.changeset/tray-content-drag-dismiss.md
sarahgm: Resolving the changeset half of this: the 44px claim is gone and the text now matches the diff. The declined requirement and the stale PR description are no longer about this line, so I picked them up in a fresh note on the changeset file instead.

packages/components/src/Tray/TrayModal.tsx
sarahgm: Verified and resolving. It is the one-liner, `boundary` is gone, and so are `SCROLLABLE_OVERFLOW`, the manual walk and the `getComputedStyle`/`scrollHeight` reads.

One thing I checked on top of your table, since the marker now carries the whole fix: every internal `Tray` consumer goes through `Tray.Content`. DatePicker, DateRangePicker, Select, Menu, ActionMenu, mobile ComboBox, mobile Autocomplete, TagField and CalendarPresets all wrap their body in it, so requirement 8 and the flagship repro are covered by the marker alone.

packages/components/src/Tray/TrayModal.tsx
sarahgm: Resolving, this went away with the ancestor walk. `closest()` is a native tree walk with no style or layout cost.

packages/components/src/Tray/TrayModal.tsx
sarahgm: Verified and resolving. `style={{ y }}` is back, and `select-none` now sits on the drag handle (`Tray.tsx`), the title row (`TrayHeader.tsx`) and the actions row (`TrayActions.tsx`), in the component implementations rather than the theme.

`draggable={false}` is still not restored, so images and links in a tray can start a native HTML5 drag. I re-raised that as a nit on the `dragListener={false}` line rather than holding this thread open for it.

packages/components/src/Tray/TrayContent.tsx
sarahgm: Resolving as agreed out of scope. Worth filing as a sibling ticket so the repro you wrote up does not get lost, since the failure mode reads to users as "the tray is broken with long content".

packages/components/src/Tray/Tray.test.tsx
sarahgm: Verified and resolving. `openTray` moved above the three drag tests and all three now use it, so there is one representation of opening the Basic tray.

packages/components/src/Tray/TrayModal.tsx
sarahgm: Leaving this open, but only for the documentation half.

The design decision is settled as far as I am concerned. You were explicit that you were not asking for a rewrite, keeping the denylist is a legitimate call, and the behaviour measured correct. What is not there is the comment: your reply says one was added above `startsInTrayContent` documenting the tradeoff explicitly, and `grep -n '^\s*//' packages/components/src/Tray/TrayModal.tsx` returns only the three comments that were already on `main` (lines 33, 82, 106). The requirement 9 marker you asked for in the same note is missing too.

Detail in my new comment on `TrayModal.tsx:50`.

packages/components/src/Tray/TrayModal.tsx
sarahgm: Still open. There is no comment on `dragListener={false}` in the branch, and nothing anywhere in `packages/components/src/Tray/` mentions DSTSUP-272 or `touch-action`.

I also added a related point on the same line: dropping motion's listener removes the inline `touch-action: pan-x`, which is the good half of the fix, and nothing declares `touch-action` in its place. `usePreventScroll` means the chrome has no scrollable ancestor while the tray is open, so arbitration should leave the drag alone, but that is implicit and unverified on a real device.

packages/components/src/Tray/TrayContent.tsx
sarahgm: Half done, so leaving it open.

The constant exists and the duplication inside `src` is gone, which removes the main risk you flagged. What is still outstanding: it has no doc comment (your reply says it is documented there), it is not re-exported from the package, so it cannot be the public opt-out marker option 1 was about, and the ticket's request for an escape hatch stays unmet. The new story test also hardcodes `'[data-tray-content]'` again, which puts the bare string back in a third file.

Detail on `Context.tsx:26` and `Tray.stories.tsx:305`.

packages/components/src/Tray/Tray.test.tsx
sarahgm: Partly resolved, with a correction in both directions.

In your favour: the story test that was meant to take over does not work, so these unit tests are still the only automated guard on this PR.

Against: they are better than "assert the mock" suggests. I mutation-tested them. Removing only the `startsInTrayContent` guard fails `'does not start a drag from inside the tray content'`, and removing the whole wiring fails `'starts a drag from the tray chrome'`. They also prove RAC's `Modal` really does forward `onPointerDown` to the DOM, since the mock passes it through to the real component. So they are a genuine wiring guard, just not a behaviour guard.

Leaving open because the gap you identified is still the gap. Detail on `Tray.stories.tsx:300`.

packages/components/src/Tray/Tray.stories.tsx
sarahgm: Still open, and this is the blocker.

The story has the tag and a `.test()` now, but the test cannot fail. Mutation results:

| Mutation to `TrayModal.tsx` | Story tests |
| --- | --- |
| none | 9 passed |
| whole fix removed | 9 passed |
| only the `startsInTrayContent` guard removed, bug fully back | 9 passed |
| guard removed, plus this test's downward-swipe assertion inverted to expect a dismissal | dialog still present |

The last row is the point. With the bug fully restored, a 700px downward synthetic touch swipe inside the content does not dismiss the tray, so no gesture ever happened and both `toBeInTheDocument()` calls are tautologies. That matches what the reply says about `PanSession` not registering synthetic gestures, but it applies to the swipe that was kept, not only to the handle assertion that was dropped.

Full detail and two workable alternatives in my new comment on `Tray.stories.tsx:300`.

packages/components/src/Tray/Tray.stories.tsx
OsamaAbdellateef: Rebuilt as a real gesture test, and mutation-tested it against your table before pushing.

The reason synthetic swipes did nothing: motion reads `pageX`/`pageY` (`extractEventInfo` in `framer-motion/dist/es/events/event-info.mjs`) and throttles `updatePoint` through `requestAnimationFrame`. `userEvent.pointer` with `coords: { x, y }` sets neither page coordinate, so `PanSession` computed a `NaN` offset, never passed its 3px threshold, and no gesture ever started. The new `dragVertically` helper passes explicit `pageX`/`pageY` and awaits a frame per move — motion's `PanSession` then runs for real, through the same code path a finger uses.

The assertions also stopped being tautologies. `toBeInTheDocument()` was never going to fail here even when the tray *was* being dismissed, because `AnimatePresence` keeps the dialog mounted through the 0.5s exit. So the test now reads the tray's own `translateY` off the modal, which moves the instant a drag starts:

| Mutation to `TrayModal.tsx` | Story tests |
| --- | --- |
| none | 10 passed |
| whole fix removed (`dragListener={false}` + `dragControls` + `onPointerDown`) | **1 failed** — `expected 'pan-x' not to be 'pan-x'` |
| only the `startsInTrayContent` guard removed, bug fully back | **1 failed** — `expected 1002.43 to be +0` |
| `touch-none` dropped from the drag handle | **1 failed** — `expected 'auto' to be 'none'` |

Both of your suggested routes are in there: the preconditions (`touch-action` on the modal is not `pan-x`, content is still selectable, chrome is `touch-none`) and the gesture itself (a downward drag in the content leaves `translateY` at 0 and the tray open; the same drag from the handle moves it and dismisses).

On the missing comment — you were right, and I should have re-diffed rather than reported from memory. The limitations note went into a working copy that never made it into the commit. Same root cause as the three sibling threads you flagged; all of those comments are in the branch now.

packages/components/src/Tray/Tray.stories.tsx
OsamaAbdellateef: Done — `Tray.stories.tsx:12` imports `TRAY_CONTENT_ATTR` from `./Context`, and the query is now `` dialog.querySelector(`[${TRAY_CONTENT_ATTR}]`) `` (`:377-379`). The bare string exists in one place again.

packages/components/src/Tray/Tray.stories.tsx
OsamaAbdellateef: Agreed on the goal, but took a different route to it — worth a look since it deviates from your snippet.

Enabling the snapshot on the story itself captures the story's *initial* state, which for `ScrollableContent` is the trigger button with the tray closed. That's a near-duplicate of `Basic`'s snapshot and never shows the overflowing content you want guarded. The state you're after only exists after the tray opens.

So I followed the precedent already in this file (`Basic` / `'Opens the tray'`, `:109-124`): the story keeps `disableSnapshot: true` with a comment saying why (`:280-282`), and a new snapshot-enabled test opens the tray and leaves it open (`:353-366`). The gesture test keeps `disableSnapshot: true` in its own `.test()` options, per the CLAUDE.md pattern you quoted.

Net effect is what you asked for: exactly one Chromatic snapshot for this story, and it's the tray open with content overflowing. Happy to flip to the story-level version if you'd rather have both.

packages/components/src/Tray/TrayModal.tsx
OsamaAbdellateef: You were right to check, and the correction is fair: the comment was in a working copy I never committed, so the reply described something that didn't exist in the branch. Fixed for all four you listed.

`TrayModal.tsx:50-64` now carries the denylist note above `startsInTrayContent`: that `useDragControls` is built for the opposite (fail-safe) shape, that this one fails open, and the concrete case that re-opens the bug — a carousel, slider or custom gesture widget dropped directly under `<Tray>`, outside `Tray.Content` — plus that there's no opt-out in either direction and what the allowlist rewrite would look like if it ever matters.

The requirement 9 marker is the second paragraph of the same block (`:60-64`): the decision is made once at `pointerdown` and holds for the whole gesture, so the scroll-to-top-then-keep-pulling handoff can't work and the user has to lift and restart on the chrome. Written in the same "follow-up needed" shape as the reduced-motion/FocusScope note further down.

The other three: `dragListener={false}` explained at `:136-151`, `TRAY_CONTENT_ATTR` documented at `Context.tsx:26-40`, and the story test's limitations are moot now — it drives a real gesture, detail on that thread.

packages/components/src/Tray/TrayModal.tsx
OsamaAbdellateef: Both points taken, one implemented and one declined with a reason.

**`touch-action`:** `touch-none` now sits on the three chrome areas next to the `select-none` I scoped last round — the drag handle (`Tray.tsx:198`), the title row (`TrayHeader.tsx:45`) and the actions row (`TrayActions.tsx:22`). Your read of why it was safe is right, but it was implicit; now the intent is declared where the gesture is owned, and the story test asserts `getComputedStyle(handle).touchAction === 'none'` so it can't quietly disappear (dropping it fails that test).

**`draggable={false}`, declined — I think the nit doesn't hold up.** `draggable` is a per-element content attribute and does not inherit, so setting it on the modal `<div>` has never had any effect on images or links *inside* the tray; each of those carries its own default `draggable="true"` and remains draggable regardless. And the element motion sets it on is a `<div>`, which isn't draggable to begin with — motion is being defensive about its own drag target, not about descendants. Restoring it would look like it addressed the concern while changing nothing.

What actually suppressed content drags before was the blanket `user-select: none`, and that's the thing you correctly asked me to stop applying to content last round. If native drags of tray content turn out to be a real problem, the honest fix is `-webkit-user-drag`/`draggable` on the offending elements, which is a consumer-side concern. Reasoning is recorded at `TrayModal.tsx:144-149` so the next person doesn't re-litigate it from the changelog.

The `dragListener={false}` comment you asked for is directly above, at `:136-151`.

packages/components/src/Tray/Context.tsx
OsamaAbdellateef: Picked "internal", and documented it at `Context.tsx:26-40`.

The doc block says what it marks and why (`TrayModal` telling a content gesture from a chrome one), states plainly that it is internal and not re-exported, and points at the `startsInTrayContent` note for the escape-hatch question — a public opt-out marker is a real API decision that belongs with the allowlist-vs-denylist call, not smuggled in as a side effect of hoisting a string. So the ticket's escape hatch stays unmet on purpose, and it's now written down as such rather than being an accident of where the constant lives.

It also answers your last question: it sits in `Context.tsx` because that module already holds the contracts shared between the tray parts, and putting it in `TrayContent.tsx` would make `TrayModal` import from the component it's checking against.

The story test no longer re-creates the bare string either — it imports the constant (`Tray.stories.tsx:12`).

packages/components/src/Tray/Tray.tsx
OsamaAbdellateef: Correct, it hasn't run — not triggering it in this round.

The handle's box is unchanged: this PR only adds `touch-none select-none` to the `[grid-area:drag]` row (`Tray.tsx:198`), both non-visual. The dimensions stay 48x6px, since the 44px target is deliberately out of scope and now tracked as DST-1738. The new `ScrollableContent` snapshot test does add one Chromatic snapshot of the opened tray with overflowing content (see the thread on `:279`).

Say the word and I'll comment `/run-chromatic` — I'd rather you make that call than spend the snapshot quota unasked.

.changeset/tray-content-drag-dismiss.md
OsamaAbdellateef: All three done.

**PR description** rewritten. The 44px claim and the `startsInScrollableBody` "any scrollable ancestor" description are gone; it now describes what's actually in the diff (`closest('[data-tray-content]')` via `TRAY_CONTENT_ATTR`, and `dragListener={false}` as the reason `touch-action: pan-x` disappears), says explicitly that requirement 3 is not in this PR with the handle still at 48x6px, and carries the mutation table for the story test.

**Follow-ups filed:**
- **DST-1738** — 💄 the 44px touch target. Scoped to the tray handle, with your `ui-touch-hitbox` pointer, and a note that the enlarged area has to stay inside the drag row or it will swallow content taps and re-open this bug from the other side.
- **DST-1739** — 🐛 the `TrayContent` `min-height` pinning, with your resize measurements and the `max-h-[95vh]` vs `h-(--visual-viewport-height)` inconsistency, flagged as needing a device check since it isn't reproducible headless.

**Title** fixed: `fix(DSTSUP-272): prevent tray dismissal on gestures starting in content` — no brackets, and it points at the project the ticket actually lives in.

packages/components/src/Tray/TrayModal.tsx
OsamaAbdellateef: Documentation half done — the comment is in the branch this time, at `TrayModal.tsx:50-64`, covering both the denylist tradeoff and the requirement 9 marker. Detail on your `:50` thread. The design decision stays as agreed: denylist, deliberately, now stated as such in the code rather than only in a review reply.

packages/components/src/Tray/TrayModal.tsx
OsamaAbdellateef: Both halves addressed: the `dragListener={false}` comment is at `TrayModal.tsx:136-151`, and `touch-none` now declares the intent on the three chrome areas that can start a drag. The `draggable={false}` nit I've declined, with the reasoning on your `:121` thread — short version, `draggable` doesn't inherit, so it never protected the tray's images or links in the first place.

packages/components/src/Tray/TrayContent.tsx
OsamaAbdellateef: Loose ends closed: `TRAY_CONTENT_ATTR` is documented at `Context.tsx:26-40` and deliberately kept internal (the public escape hatch belongs with the allowlist decision, not with hoisting a string), and the story test imports the constant instead of hardcoding the attribute a third time. Detail on your `Context.tsx:26` thread.

packages/components/src/Tray/Tray.test.tsx
OsamaAbdellateef: Thanks for the correction in both directions — and the mutation results are why the story test got rebuilt rather than patched. It now fails on the same two mutations these unit tests catch, plus on the behaviour they can't reach, so these stay as the cheap wiring layer underneath rather than being the only guard. Table on the `Tray.stories.tsx:300` thread.

packages/components/src/Tray/Tray.stories.tsx
OsamaAbdellateef: The test can fail now. Root cause of the tautology was two-fold: motion reads `pageX`/`pageY` and throttles through rAF, so the old synthetic swipe never started a gesture at all; and `toBeInTheDocument()` stays true through `AnimatePresence`'s 0.5s exit even while the tray is being dismissed. The rewrite drives a real `PanSession` and asserts the tray's `translateY`. Full mutation table on the `:300` thread — every row you listed now fails, plus `touch-none` being dropped from the handle.

packages/components/src/Tray/TrayModal.tsx
sarahgm: Resolving. The denylist stays, deliberately, and the tradeoff is now stated in the code rather than only in a review reply.

`TrayModal.tsx:50-64` says that `useDragControls` is built for the opposite, fail-safe shape, that this one fails open, and names the concrete case that re-opens the bug: a carousel, slider or custom gesture widget dropped directly under `<Tray>`, outside `Tray.Content`. It also records that there is no opt-out in either direction, and sketches the allowlist over `[grid-area:drag]`/`title`/`actions` as the revisit if that shows up in practice.

The requirement 9 handoff is the second paragraph (`:60-64`), in the same "follow-up needed" shape as the reduced-motion note further down.

packages/components/src/Tray/TrayModal.tsx
sarahgm: Resolving, the comment is at `TrayModal.tsx:136-151` and covers what you asked for: motion's four side effects, `pan-x` blocking vertical panning for the whole subtree with no way for a descendant to re-allow it, that drag is started manually from `startDrag` instead, and that restoring `dragListener` would silently bring the scroll bug back. It points at DSTSUP-272 and names where the pieces worth keeping were re-applied (`select-none`/`touch-none` on the chrome).

It also records why `draggable={false}` is not restored, which was the open half of your list. Detail on the `:152` thread below.

packages/components/src/Tray/TrayContent.tsx
sarahgm: Resolving. `TRAY_CONTENT_ATTR` is documented at `Context.tsx:26-40`, and the story test imports it (`Tray.stories.tsx:12`) instead of re-creating the bare string a third time, so the untyped cross-file contract you flagged is down to a single definition.

Osama took option 1's mechanism but deliberately kept it internal rather than making it the public opt-out marker. That call and the reasoning are on the `Context.tsx:26` thread.

packages/components/src/Tray/Tray.test.tsx
sarahgm: Resolving. These are no longer the only automated guard: the story test drives a real `PanSession` now and asserts the tray's own `translateY`, so it fails on both wiring mutations these catch plus the behaviour they cannot reach. Mutation table on the `Tray.stories.tsx:353` thread.

Keeping these as the cheap wiring layer underneath is the right outcome. Your read of what they verify was fair at the time, and they turned out to catch slightly more than "assert the mock" suggested, which I noted last round.

packages/components/src/Tray/Tray.stories.tsx
sarahgm: Resolving. The story carries automated weight now: `tags: ['component-test']` plus two attached tests. One opens the tray and is snapshot-enabled, the other drives a real touch gesture through motion's `PanSession` and asserts `translateY` on the modal, so a regression in the guard fails it.

Your instinct about where the test belonged was right. What it needed on top was explicit `pageX`/`pageY` and a frame per move, since motion reads page coordinates and throttles through `requestAnimationFrame`, which is why the first synthetic version never started a gesture at all. Full mutation table on the `:353` thread.

packages/components/src/Tray/Tray.stories.tsx
sarahgm: Blocker cleared, resolving.

The assertions can fail now. `translateY(modal)` reads the computed transform and `getComputedStyle(handle).touchAction` reads a real style, so neither one is a tautology the way `toBeInTheDocument()` was through `AnimatePresence`'s 0.5s exit. The `pageX`/`pageY` plus one-frame-per-move diagnosis matches what I measured last round, when a 700px synthetic swipe with the bug fully restored moved nothing at all: `PanSession` never passed its 3px threshold, so there was no gesture to assert on.

Storybook Tests are green across all four CI shards on `fceee670`, so the rewrite runs and passes in Firefox. I have not re-run your four mutations myself. I am taking that table on the strength of the assertions no longer being tautological plus the green run, rather than re-deriving it.

packages/components/src/Tray/Tray.stories.tsx
sarahgm: Your route is better than my snippet, resolving.

You are right that a story-level snapshot captures the initial state, which for `ScrollableContent` is the closed trigger button and a near-duplicate of `Basic`. The state I wanted guarded only exists after the tray opens.

What is in the branch does what I was after: the story keeps `disableSnapshot: true` with the reason at `:280-282`, the snapshot lives in `'Opens a tray whose content overflows'` (`:353-366`) which captures the opened tray with content overflowing, and the gesture test opts out in its own `.test()` options per the CLAUDE.md pattern. One snapshot for this story and it is the right one. No need to flip it.

packages/components/src/Tray/TrayModal.tsx
sarahgm: Verified and resolving. All four are in the branch this time.

- the denylist note above `startsInTrayContent`, `TrayModal.tsx:50-64`, with the requirement 9 handoff as its second paragraph at `:60-64`
- `dragListener={false}` explained at `:136-151`
- `TRAY_CONTENT_ATTR` documented at `Context.tsx:26-40`
- the story test's limitations note, moot now that the test drives a real gesture

Thanks for saying plainly what happened with the uncommitted working copy. That is the useful version of that answer, and re-diffing before reporting is all this needed.

packages/components/src/Tray/TrayModal.tsx
sarahgm: Verified and resolving, and the `draggable` decline is accepted.

**`touch-action`:** `touch-none` sits on the three chrome areas next to the `select-none` from last round, at `Tray.tsx:198`, `TrayHeader.tsx:45` and `TrayActions.tsx:22`, so the intent is declared where the gesture is owned instead of resting on RAC's scroll lock implicitly. The story test asserts `getComputedStyle(handle).touchAction === 'none'`, so it cannot quietly disappear.

**`draggable={false}`:** you are right and my nit does not hold up. It is a per-element content attribute that does not inherit, so on the modal `<div>` it never protected images or links inside the tray, and restoring it would have looked like a fix while changing nothing. Good catch that the blanket `user-select: none` was what actually suppressed those drags, which is exactly the thing I asked you to stop applying to content. Recording the reasoning in the code was the right call so nobody re-litigates it from the changelog.

packages/components/src/Tray/Context.tsx
sarahgm: Verified and resolving. Both loose ends are closed.

`Context.tsx:26-40` documents what the attribute marks and why `TrayModal` needs it, states plainly that it is internal and not re-exported, and answers my last question about placement: it sits here because this module already holds the contracts shared between the tray parts, and putting it in `TrayContent.tsx` would make `TrayModal` import from the component it checks against. `index.ts` still exports only `Tray` and `TrayProps`, which matches the "internal" choice.

Agreed that a public opt-out marker is a real API decision belonging with the allowlist-vs-denylist call rather than a side effect of hoisting a string. The ticket's escape hatch stays unmet, now on purpose and written down, which is what I was asking for. `Tray.stories.tsx:12` importing the constant closes the third copy too.

.changeset/tray-content-drag-dismiss.md
sarahgm: Verified and resolving, all three followed through.

- **Title** is `fix(DSTSUP-272): prevent tray dismissal on gestures starting in content`, no brackets and pointing at the project the ticket lives in.
- **PR description** no longer claims the 44px hit target or the "any scrollable ancestor" walk, and it now says explicitly that requirement 3 is out with the handle unchanged at 48x6px. The mutation table in there is a good addition for the reporter.
- **Follow-ups** both exist in Backlog: DST-1738 for the 44px touch target with the note that the enlarged area has to stay inside the drag row, and DST-1739 for the `TrayContent` `min-height` pinning with the resize measurements and the device-check flag. Thanks for carrying both repros over rather than letting them die in the thread.

packages/components/src/Tray/Tray.tsx
sarahgm: Saying the word: please comment `/run-chromatic` on the PR.

I agree the handle's box is unchanged and I expect no diffs on the existing stories, `touch-none select-none` are both non-visual. The reason to run it anyway is the new `'Opens a tray whose content overflows'` test, which adds a Chromatic snapshot of the opened tray with content overflowing. That is a new baseline rather than a no-op, and it is the one visual state this fix touches, so it should be reviewed and accepted deliberately.

Thanks for leaving the quota call to me. Keeping this thread open until the run is green and the checklist item is ticked.

One mechanical note: it has to be a top-level comment on the PR. A reply inside a review thread does not trigger the workflow, and an absent run shows up as no check at all rather than as a failure.

OsamaAbdellateef: Done :heavy_check_mark:  @sarahgm

packages/components/src/Tray/Context.tsx
```diff
@@ -22,3 +22,20 @@ export const TrayContext = createContext<TrayContextProps>(
 );
 
 export const useTrayContext = () => use(TrayContext);
+
+/**
+ * Marks the tray's scrollable body so `TrayModal` can tell a gesture that
```
sarahgm: so can you run the skill I wrote in the channel over it for the comments check?

OsamaAbdellateef: image
@sarahgm

sarahgm: but maybe shorten the comments to the main point

OsamaAbdellateef: now they should be shorter  @sarahgm

sarahgm (approved):

PR #5750, opened 2026-08-17
chore(DST-1528): retire the unused AI tools and normalize the rest

sebald:
# Description

Cuts the committed AI toolkit from 12 tools to 3 and puts the survivors on one standard. Covers two tickets that were always meant to run back to back: DST-1528 decides what earns its place, DST-1529 fixes the shape of what's left.

Nine tools were never reached for and are retired: the `a11y-audit` and `component-scaffold` agents, the `create-component` command, and the `component-docs-writer`, `testing-codequality-check`, `typescript-codequality-check`, `vercel-composition-patterns` and `vercel-react-best-practices` skills. Also gone is `.cursorrules`, whose only line included `@.ai/agents.md` — a path that has never existed in this repo.

`review-pr` called the deleted vercel pack in five places, so it now calls `react-best-practices` from the `vercel` plugin already enabled in `.claude/settings.json`: 66 maintained rules instead of the 64 stale ones the vendored copy had drifted to.

The survivors — `create-pr`, `review-pr`, `vrt` — all live at `.claude/skills/<name>/SKILL.md`. `vrt` moves out of `.claude/commands`, which is now empty and deleted. Because it spends Chromatic snapshot quota, its description rules out proactive use and confirmation is step 1 of its workflow rather than a footnote. Two workflow behaviours the old command never mentioned are now written down: a dispatch is silently skipped when the `VRT_ENABLED` repository variable is `false`, and commenting `/run-chromatic` on a PR triggers the same run.

`.claude/README.md` documents the conventions and deliberately lists no tools — Claude Code already injects every skill's name and description into each session, so an inventory here would be a hand-kept copy of that list, drifting a little further with every change. DST-1529 is the evidence: it was itself written as a verified-on-disk inventory and three of its five findings were wrong eight weeks later. `CLAUDE.md` points at the conventions instead of restating them.

Retiring `component-docs-writer` also removes 2,823 lines of documentation conventions that exist nowhere else in the repo. That is a deliberate trade, recorded on DST-1528 — the content stays recoverable from git history.

Closes DST-1528
Closes DST-1529

> [!NOTE]
> **Base of a three-PR stack — nothing blocks this one.** #5753 (DST-1521) and #5754 (DST-1530) are both stacked on it and cannot merge until this does, so it is the one to review first.

## Test Instructions

1. `git ls-files .claude` — expect only `README.md`, `settings.json`, and the three skills.
2. Open a fresh Claude Code session in this repo. `/create-pr`, `/review-pr` and `/vrt` should be available; `/create-component` and the retired skills should be gone.
3. Read `.claude/README.md` and check the conventions match what you'd want a new teammate to follow.
4. **Do not run `/vrt` to test it** — it dispatches a real Chromatic run. It was verified structurally and by confirming the session resolves it as a skill.

## Breaking Changes

No — nothing published changes.

Note for reviewers: this removes tools other people may have been using, and the team triage session DST-1528 called for never happened. This PR is the only gate — please object here if one of the nine retired tools is something you rely on.

## Checklist

- [ ] Unit tests added/updated — N/A, no source changes
- [ ] Component documentation added/updated (if it exists) — N/A, no `docs/` changes
- [ ] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) (for new/changed interactive components) — N/A


.claude/skills/vrt/SKILL.md
```diff
@@ -0,0 +1,68 @@
+---
+name: vrt
+description: Trigger the Visual-Regression-Tests (Chromatic) GitHub Actions workflow on the current or a given branch. Use ONLY when the user explicitly asks to run visual regression tests, run Chromatic, or types `/vrt`. Dispatching consumes Chromatic snapshot quota, so never invoke this proactively, never as a follow-up to unrelated work, and never to "verify" a change nobody asked to verify.
+allowed-tools: Bash(gh workflow run *), Bash(gh run list *), Bash(gh run view *), Bash(gh run watch *), Bash(git branch --show-current), Bash(git rev-parse *)
```
jim761: ⚠️ **Step 2 asks for a check this `allowed-tools` list forbids.**

Line 36 says: *"If the resolved branch has unpushed commits, say so before asking."* Detecting that needs `git rev-list @{u}..HEAD`, `git status -sb`, or `git log origin/x..x` — none of which are permitted here. `git rev-parse HEAD` vs `git rev-parse origin/<branch>` can detect *divergence* but not its direction, and with no `git fetch` allowed the remote ref may be stale anyway.

This matters because `.claude/README.md` in this same PR calls `allowed-tools` "the one guardrail in a skill that is structural rather than a matter of prose" — so the first skill written to that standard has a workflow step the guardrail blocks. Either widen the list:

```suggestion
allowed-tools: Bash(gh workflow run *), Bash(gh run list *), Bash(gh run view *), Bash(gh run watch *), Bash(git branch --show-current), Bash(git rev-parse *), Bash(git rev-list *)
```

…or restate the line-36 check in terms of what's already permitted.

.claude/skills/vrt/SKILL.md
```diff
@@ -0,0 +1,68 @@
...
+
+Dispatch the `Visual-Regression-Tests` workflow (`.github/workflows/visual-regression-tests.yml`) via `workflow_dispatch`, then report the run so the user can follow it.
+
+This skill spends money. The `allowed-tools` list above is deliberately narrow, and step 1 is a hard gate — treat both as part of the skill, not as boilerplate.
+
+## Usage
+
+```
+/vrt              # current branch
+/vrt <branch>     # named branch
```
jim761, line 17: ⚠️ **The branch argument has no defined entry point.**

Usage documents `/vrt <branch>`, but step 2 only says "If a branch name was given" — it never says where the name arrives. The sibling skill names it explicitly (`review-pr/SKILL.md:20` → *"Extract the PR number from `$ARGUMENTS`"*), and the `.claude/commands/vrt.md` this replaces parsed `$ARGUMENTS` too.

For a skill whose whole confirmation gate is about *which branch* burns snapshot quota, this is the wrong place to leave intake implicit.

.claude/skills/vrt/SKILL.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+## Workflow
+
+### 1. Ask before dispatching
```
jim761: 💡 **Step 1 forward-references step 2.**

"Resolve the branch first (step 2), then state it back" makes a numbered workflow read backwards. Ordering them *determine branch* → *confirm* reads straight and still satisfies the README rule, since nothing is dispatched until step 3.

.claude/skills/review-pr/SKILL.md
```diff
@@ -139,10 +139,10 @@ For each changed file:
 
 ### 8. Apply Vercel React Best Practices
 
-Invoke the `vercel-react-best-practices` skill to check for React/Next.js performance patterns:
+Invoke the `react-best-practices` skill from the `vercel` plugin (enabled in `.claude/settings.json`) to check for React/Next.js performance patterns:
 
 ```
-Skill(vercel-react-best-practices)
+Skill(vercel:react-best-practices)
```
jim761, line 145: ⚠️ **No fallback if the `vercel` plugin isn't installed.**

`Skill(vercel:react-best-practices)` is now a hard dependency in the middle of the workflow. Plugins install per-user — `.claude/README.md` in this PR says so explicitly — so a teammate without `claude-plugins-official` hits a failure at step 8 of 11. The vendored copy at least always resolved.

One line ("skip this step if the plugin isn't available") turns a break into a graceful degrade.

.claude/skills/review-pr/SKILL.md
```diff
@@ -349,7 +349,7 @@ The following tools are needed for this skill:
 
 - Review is displayed locally first, then user is asked if they want to post to GitHub
 - The skill focuses on Marigold-specific patterns documented in CLAUDE.md
-- Integrates `vercel-react-best-practices` skill for React performance checks
+- Integrates the `vercel` plugin's `react-best-practices` skill for React performance checks
 - Jira integration is optional and gracefully handles missing tickets
 - Automated checks (typecheck, lint) require user confirmation before running
 - GitHub posting is opt-in and supports full review comments, inline code comments, or both
```
jim761: 💡 **File has no trailing newline.**

Pre-existing, not introduced here — but `.claude` is listed in `.prettierignore`, so CI will never flag it. Free fix while the file is already open.

CLAUDE.md
```diff
@@ -145,6 +145,30 @@ _Dialog.Title = DialogTitle;
 export const Dialog = _Dialog as DialogComponent;
 ```
 
+### Composition
```
jim761, line 148: 💡 **Worth noting that this section is the *only* successor to the deleted composition pack.**

The `vercel` plugin ships `react-best-practices` but there is no `vercel:composition-patterns` — I checked the plugin's skill list. So unlike the react rules, which moved to a maintained upstream, these composition rules have no upstream at all: this section is it.

That's a fine outcome, but the PR description explains the react swap and is silent here. Saying so in the description or on DST-1528 prevents a future reader assuming a plugin still covers it.

(The `Sidebar/Context.tsx` pointer checks out, by the way — `createContext<SidebarContextValue | null>(null)` plus a `useSidebar()` that throws outside its provider.)

jim761 (commented):
## Automated Code Review

**Visual Regression:** not needed — no `packages/**`, `themes/**`, `.storybook/**`, `*.stories.tsx` or `*.styles.ts` files touched. CI is green across all 25 checks.

### What I verified (not just read)

- **Every factual claim in `vrt/SKILL.md` matches `.github/workflows/visual-regression-tests.yml`**: workflow name `Visual-Regression-Tests`, `workflow_dispatch` takes no inputs (so `--ref` is right), the `vars.VRT_ENABLED != 'false'` fail-open gate, `/run-chromatic` restricted to `OWNER|MEMBER|COLLABORATOR`, and the merge-into-`main` trigger with `docs/**`/`*.md`/`*.mdx`/`.changeset/**` ignored. The two "behaviours the old command never mentioned" are both real.
- **No dangling references** to any of the nine retired tools anywhere in the tree — one grep across the whole PR ref returns a single hit (`.prettierignore`, below).
- `git ls-tree .claude` matches the PR body's claim (README, settings.json, 3 skills + review-pr's `references/`).
- `packages/components/src/Sidebar/Context.tsx` really does model what the new CLAUDE.md Composition section points at — `createContext<SidebarContextValue | null>(null)` plus a `useSidebar()` that throws outside its provider.
- `vercel:react-best-practices` exists in the enabled `vercel@claude-plugins-official` plugin, so the five rewired `review-pr` call sites resolve.

### Issues Found

#### Critical (Must Fix)

None.

#### Warnings (Should Fix)

**1. `.claude/skills/vrt/SKILL.md:36` — step 2 asks for a check `allowed-tools` (line 4) forbids.**
"If the resolved branch has unpushed commits, say so before asking" needs `git rev-list @{u}..HEAD`, `git status -sb`, or `git log origin/x..x`. The frontmatter allows only `git branch --show-current` and `git rev-parse *`. `git rev-parse HEAD` vs `git rev-parse origin/<branch>` can detect *divergence* but not direction, and there's no `git fetch` allowed, so the remote ref may be stale anyway. This one matters because the README added in the same PR calls `allowed-tools` "the one guardrail in a skill that is structural rather than a matter of prose" — the first skill written to that standard has a workflow step the guardrail blocks. Either add `Bash(git rev-list *)` or restate the check in terms of what's permitted.

**2. `.claude/skills/vrt/SKILL.md:16-17` — the branch argument has no defined entry point.**
Usage documents `/vrt <branch>`, but step 2 only says "If a branch name was given". The sibling `review-pr/SKILL.md:20` names it explicitly (`Extract the PR number from $ARGUMENTS`), and the command file this replaced parsed `$ARGUMENTS` too. For a skill whose confirmation gate is explicitly about *which branch* burns quota, that's the wrong place to leave intake implicit.

**3. `.claude/skills/review-pr/SKILL.md:145` — no fallback if the vercel plugin is missing.**
`Skill(vercel:react-best-practices)` is now a hard dependency mid-workflow. Plugins install per-user (the new README table says so), so a teammate without `claude-plugins-official` gets a failure in step 8 of 11 — the vendored copy at least always resolved. One line ("skip this step if the plugin isn't available") turns a break into a degrade.

#### Suggestions

4. **`.prettierignore:16`** still lists `.cursorrules` after the file is deleted — stale entry, one-line removal. (No inline comment on this one — the file isn't part of the diff, so GitHub can't anchor there.)
5. **`vrt/SKILL.md:22`** — step 1 says "Resolve the branch first (step 2)". Ordering steps 1→2 as *determine branch* → *confirm* reads straight and still satisfies the README rule, since nothing is dispatched until step 3.
6. **`review-pr/SKILL.md`** has no trailing newline (pre-existing; `.claude` is prettier-ignored so CI won't ever flag it). Free fix while the file is open.
7. **`vercel-composition-patterns` has no plugin equivalent.** The vercel plugin ships `react-best-practices` but not composition-patterns, so the new CLAUDE.md §Composition is its *only* successor — a stronger and more fragile claim than the react one. The PR body explains the react swap but is silent here; saying so prevents a future "the plugin covers it" assumption.

### Against the tickets

Two deliberate deviations worth a team ack rather than a fix:

- **DST-1528 also scoped the plugins** ("`code-simplifier` and the other enabled plugins — confirm each still earns its slot"). `.claude/settings.json` is untouched and unmentioned; all six plugins keep their slot by default.
- **Both tickets ask for the verdicts/inventory to land in `CLAUDE.md`** (plus the Confluence §2 adoption table and a quarterly re-check). This PR argues the opposite — that a hand-kept inventory is precisely the drift trap DST-1529 demonstrated — and the argument is right and well evidenced. But it inverts a stated acceptance criterion, and the Confluence table + re-check are still open regardless.

The PR's own note stands too: nine tools disappear without the triage session the ticket called for, and this PR is the only gate.

### Recommendation

**Comment** — no blockers. Findings 1–3 are small and worth fixing before merge; the rest are nits. The deletions are clean, the survivors are consistent, and every claim I could check held up.

---

.claude/skills/vrt/SKILL.md
aromko: Fixed by narrowing rather than widening. `git rev-parse *` is called nowhere in the skill body, so it goes, and the pattern the check actually needs comes in:

```suggestion
allowed-tools: Bash(gh workflow run *), Bash(gh run list *), Bash(gh run view *), Bash(gh run watch *), Bash(git branch --show-current), Bash(git rev-list *)
```

The line-36 prose is now a real command. Worth reporting how that landed, because my first attempt was wrong: `git rev-list --count origin/<branch>..<branch>` needs a **local** branch of that name and dies with `unknown revision` otherwise, which `/vrt some-other-branch` always would. So the check is `git rev-list --count @{u}..HEAD`, scoped to the current branch, and its failure mode is the useful one: no upstream means never pushed, which is exactly what the user needs told.

A branch named in `$ARGUMENTS` gets no check, and the skill now says so: there is no local counterpart for it to be ahead of, and `gh workflow run --ref` fails with a clear error on an unknown ref.

No `git fetch` in the list, deliberately. A stale remote ref means occasionally failing to warn, and the failing dispatch is the backstop.

.claude/skills/vrt/SKILL.md
aromko: Fixed. Step 1 names the intake the way `review-pr/SKILL.md:20` does:

> If `$ARGUMENTS` names a branch, use it. Otherwise take the current branch and check that it is pushed:
>
> ```bash
> git branch --show-current
> git rev-list --count @{u}..HEAD
> ```

.claude/skills/vrt/SKILL.md
aromko: Fixed, by merging steps 1 and 2 rather than swapping them.

Swapping would have made `.claude/README.md:42` literally false for the one skill it was written for: *"Confirmation is step 1 of the workflow ... the first numbered step."* One step titled **"Resolve the branch, then ask"** reads straight, keeps confirmation in step 1, and takes the workflow from five steps to four.

.claude/skills/review-pr/SKILL.md
aromko: Fixed, one line after the call:

> If the `vercel` plugin is not installed, skip this step and note it in the report. The rest of the review does not depend on it.

Left `## Tools Required` as it was. It already labels the entry `(vercel plugin)`, and the workflow step is where the degrade has to be readable.

.claude/skills/review-pr/SKILL.md
aromko: Fixed, the file ends with `0a` now.

aromko:
Worked through @jim761's six actionable findings. All six are fixed and each thread has the details.

**Finding 4 (`.prettierignore:16`) has no inline anchor, so here:** `.cursorrules` is out of the ignore list. `grep -rn '\.cursorrules'` across the tree now returns nothing.

Finding 7 (`vercel-composition-patterns` has no plugin successor) is a description or ticket note rather than a code change, so I left that one to you, @sebald.

All six sit on one commit, `4a891a3bd`, pushed to this branch. *(Edited: this paragraph previously said the commit was local and unpushed.)*

<details>
<summary>Diff (3 files, +21 −16)</summary>

- `.claude/skills/vrt/SKILL.md`: steps 1 and 2 merged, `$ARGUMENTS` intake named, pushed-check is `git rev-list --count @{u}..HEAD`, `allowed-tools` swaps the never-called `git rev-parse *` for `git rev-list *`, steps renumbered 3–5 → 2–4.
- `.claude/skills/review-pr/SKILL.md`: plugin-missing degrade after the `Skill()` call, trailing newline.
- `.prettierignore`: drop `.cursorrules`.

</details>

---

One thing the review missed, and it bears directly on this PR's "one standard" claim rather than being a nit:

`.claude/README.md:35-44` defines the side-effect class as skills that *"spend money **or touch the outside world**"*, then imposes two obligations plus narrow `allowed-tools`. `create-pr` pushes and opens PRs. `review-pr` posts review comments. Both touch the outside world, both keep the prose `## Tools required` list that line 44 explicitly dismisses, and neither has `allowed-tools` or a description that rules out proactive use. So two of the three survivors do not meet the standard the same PR writes down.

Adding `allowed-tools` to them is the wrong fix: `review-pr`'s own list omits `Grep` and `Glob`, so making it structural would break the skill. The honest fix is one line, narrowing the class at README:37 to *"spend money or take an action that cannot be undone"*, which leaves `vrt` as the only member and makes the rule true as written. Both skills already gate their outward step behind an explicit confirmation, so nothing is lost. Not in my commit, since it is a call about your own convention, not a defect.

.claude/skills/vrt/SKILL.md
aromko: Pushed as `4a891a3bd` (rebased onto main). `allowed-tools` now ends `Bash(git branch --show-current), Bash(git rev-list *)`, and step 1 carries the concrete `git rev-list --count @{u}..HEAD`.

.claude/skills/vrt/SKILL.md
aromko: Pushed as `4a891a3bd` (rebased onto main). Step 1 opens with *"If `$ARGUMENTS` names a branch, use it."*

.claude/skills/vrt/SKILL.md
aromko: Pushed as `4a891a3bd` (rebased onto main). Steps 1 and 2 are one step now, `### 1. Resolve the branch, then ask`, and the workflow is four steps instead of five.

.claude/skills/review-pr/SKILL.md
aromko: Pushed as `4a891a3bd` (rebased onto main). The degrade line sits directly after the `Skill()` fence.

.claude/skills/review-pr/SKILL.md
aromko: Pushed as `4a891a3bd` (rebased onto main). File ends with a newline.

aromko:
Rebased onto `main` (was 15 commits behind and `CONFLICTING`; now `MERGEABLE`). Head is `ace240f4c`. **Force-pushed**, so #5753 and #5754 need rebasing onto the rewritten base.

One conflict, and it is worth a look rather than a rubber stamp, because it is evidence rather than mechanics.

`modify/delete` on `.claude/skills/component-docs-writer/references/anatomy-diagrams.md`: this branch deletes it, and #5733 (DST-1703, @OsamaAbdellateef, merged 2026-08-18, the day after this PR opened) added 19 lines to it. Not a drive-by either. It is a **canonical-frame contract**: viewBox width must be 660, wrapper must be `max-w-[90%]`, so `fontSize="14"` labels and `strokeWidth="2"` strokes come out the same size on every page.

I took the delete, since retiring the skill is the point of the PR. But I did not let the contract go with it:

- **59** `*-anatomy.tsx` files under `docs/content/**` are built to it.
- DST-1703 is still migrating more of them.
- No script enforces it (`scripts/` has `check-rac-first-imports`, `check-react-aria-dedupe`, `check-theme-zindex`, nothing for this).
- `git grep` finds the rule stated in that one file and nowhere else.

So dropping it silently leaves the next author to pick a viewBox width and quietly break label scale across every other page. "Recoverable from git history" does not help someone who never knew the rule existed. It is now verbatim in `docs/CLAUDE.md` (`ace240f4c`), which is hand-written, scoped to the only directory where the rule applies, and out of reach of the `next dev` generator that owns `docs/AGENTS.md`. Separate commit, so it is trivial to drop if you disagree.

Two things this raises that are yours and @OsamaAbdellateef's to call, not mine:

1. The PR retires `component-docs-writer` as never reached for. Someone extended one of its reference files six days ago. That is the objection the PR body invited, arriving as a merge conflict instead of as a comment.
2. The PR notes that retiring it drops 2,823 lines of documentation conventions that exist nowhere else. The anatomy contract turned out to be one concrete instance of exactly that cost, found only because the rebase forced it into view. Worth asking what else in those 2,823 lines is load-bearing before this merges.

CLAUDE.md
aromko: Checked this against the installed plugin rather than the skill list, and you are right: `vercel@claude-plugins-official` **0.45.1** ships 30+ skills including `react-best-practices`, and no composition equivalent. So §Composition is the only copy, not a pointer to an upstream.

Recorded on DST-1528 rather than in the PR description, since the description will not be read again after merge. It includes the coverage mapping of the 8 deleted rule files:

- **Carried in full (5):** `architecture-avoid-boolean-props` → "No mode booleans", `architecture-compound-components`, `patterns-children-over-render-props` → "Children over render props", `state-context-interface` → "Share state through context", `state-lift-state` → "Lift state into a provider".
- **Compressed (2):** `patterns-explicit-variants`, `state-decouple-implementation`.
- **Deliberately dropped (1):** `react19-no-forwardref`.

That last one is worth surfacing here, because it changes the framing of this thread. The rule told agents `forwardRef` is unnecessary in React 19 and showed it as the *incorrect* pattern. `CLAUDE.md` mandates the opposite in two places (*"Don't skip `forwardRef`"*), and the Compound Components example directly above this section is built on `forwardRef`.

So the vendored pack was instructing agents against a project rule for as long as it sat in `.claude/skills/`. `vercel-react-best-practices` was retired for being stale at 64 rules against upstream's 70; this one was *wrong for this repo*, which is worse and less visible. §Composition having no upstream is a real fragility, but it is 7-of-8 rules with the eighth removed on purpose, which reads better than the vendored pack did.

Nothing to change in the code, so resolving. The durable note is on the ticket.

aromko (approved):

PR #5751, opened 2026-08-17
style(DST-1386): add hover affordance to Checkbox, Radio and Switch

sebald:
# Description

`Checkbox`, `Radio` and the off-state `Switch` had no hover affordance — the only interactive controls in the theme without one. DST-1295 widened their hit area, so a larger, more forgiving target was moving silently under the cursor.

**Current vs. new behaviour**

| | Current | New |
| --- | --- | --- |
| Checkbox / Radio, unselected | border stays at the resting edge | border darkens (alpha 0.32 → 0.44, 2.14:1 → 3.06:1) |
| Switch, off | track stays flat `bg-control` | track darkens (alpha 0.16 → 0.30, 1.43:1 → 2.05:1), instantly |
| Selected / indeterminate / disabled / read-only | — | unchanged on hover |
| **Indeterminate Checkbox (bug)** | white box, resting border, dark minus | filled `selected-bold` box with a light minus |

All three land on the same perceived step — a ~1.4× contrast multiplier, matching `ToggleButton`'s existing hover (1.40×). Measured from composited pixels rather than derived from token lightness.

**No new token.** `Checkbox`/`Radio` step the alpha on `--color-control-border` (the `ToggleButton` idiom). The `Switch` steps `--color-control`, which has been alpha-based since DST-1672 — so the opaque `--color-control-hover` the ticket originally proposed isn't needed, and the naming dependency on DST-1674 Scope B doesn't apply.

**The Switch hover is instant, and the toggle still eases.** DST-1436 removed background from hover transitions system-wide but deliberately left the track's `transition-colors` for the *toggled* state. Since one property can't distinguish the two triggers, the hover tint is a `background-image` gradient instead — not a transitioned property, so it lands on the next frame in both directions while switching on/off keeps its 150ms ease. Rendered pixels are identical to a plain `background-color` at the same value. (React Aria offers no help here: `Switch` renders no transition-state attribute; `data-entering`/`data-exiting` exist only on overlay components.)

**Also fixed: the indeterminate `Checkbox` has never been styled.** Its three state declarations used `group-[indeterminate]`, whose arbitrary-variant syntax takes a *selector* — so the bare word compiled to `:is(indeterminate)`, a type selector for a nonexistent element, and the rules never matched. Now `group-indeterminate`, which reads the `data-indeterminate` React Aria actually renders. Folded in here because the hover rule has to exclude that state, and guarding against a dead variant guards nothing. It is the only bare-word arbitrary group variant in the codebase.

### Reviewer note: every `:not()` is load-bearing

Worth knowing before simplifying the class strings. A *bare* `group-hover:` rule already loses to `group-selected:`/`group-disabled:` — the selectors tie at (0,2,0) and `tailwindcss-react-aria-components` registers `hover` before them, so those compile later and win. But each `:not()` adds a class-level weight, so **the moment one exclusion is added the rule outranks those state rules**. Removing any clause makes hover repaint that state. This bit me mid-implementation; the story tests now cover each state, and deleting `not-group-selected` was confirmed to fail them.

Closes DST-1386

## Screenshots / Preview

Storybook preview from CI is the best place to see this — hover states don't screenshot from a static build. Chromatic has **not** been run yet (see checklist).

The indeterminate fix is the only change to a *resting* state, so it is the one to look at closely:

| Before | After |
| ------ | ----- |
| `border oklch(0.15 0.008 54 / 0.32)`, `background rgb(255,255,255)` — dark minus on white | `border` and `background` both `oklch(0.22 0.008 54)` — light minus on the bold fill |

## Test Instructions

1. `pnpm sb`, open **Components / Checkbox → Basic**. Hover the box: the border darkens. Toggle `defaultChecked`, `indeterminate`, `disabled`, `readOnly` in the controls — hover changes nothing in any of those states.
2. Set `indeterminate` on its own: the box should be a filled dark square with a light minus (on `main` it is white with a dark minus).
3. **Components / Radio → Basic.** Hover "Option 2" (unselected): border darkens. Hover "Option 1" (selected) and "Option 3" (disabled): unchanged.
4. **Components / Switch → Basic.** Hover the off-state track: it darkens *immediately*, with no fade. Then click to toggle: the colour change still eases over 150ms. That asymmetry is the point.
5. Compare the three hovers side by side — they should read as the same weight of change. Cross-component consistency is the main argument for the ticket.
6. Automated: `pnpm test:sb` (574 passing, 13 new) and `pnpm test:unit` (1569 passing; one `Checkbox` className snapshot updated).

## Breaking Changes

No. Visual-only, patch changeset. No API, token or class-name surface changes for consumers.

## Checklist

- [x] Accessibility reviewed against ARIA APG 


themes/theme-rui/src/components/Switch.styles.ts
```diff
@@ -24,6 +24,12 @@ export const Switch: ThemeComponent<'Switch'> = {
       'border-2 border-transparent',
       'group-disabled/switch:bg-disabled-surface group-disabled/switch:text-disabled group-selected/switch:group-disabled/switch:bg-disabled-surface group-selected/switch:group-disabled/switch:text-disabled',
       'group-selected/switch:bg-selected-bold bg-control',
+      // background-image, not background-color: the track keeps
+      // `transition-colors` for the toggle (DST-1436), and background-image is
+      // not transitioned, so hover lands instantly. It layers over the track
+      // rather than replacing it, hence 0.14/(1-alpha) to reach the same step as
+      // Checkbox/Radio. Exclusions are load-bearing, see Checkbox.styles.ts.
```
jim761: are the comments necessary here, 5 lines of comments for 1 line of code

jim761 (commented):
## Automated Code Review

### Overview
| Field | Value |
|-------|-------|
| Author | @sebald |
| Branch | `feat/DST-1386-control-hover-affordance` → `main` |
| Files Changed | 8 |
| Additions | +254 |
| Deletions | -2 |
| CI | All 25 checks passing |

### Linked Ticket
**DST-1386** — 💄 Add hover affordance to Checkbox, Radio (and Switch) controls · Status: *Wird überprüft* · Assignee: Sebastian Sebald

The implementation resolves all three open decisions the ticket left: **A** → `+0.18` (step, not value) as recommended; **B** → no new token, steps the now-alpha-based `--color-control`, so the DST-1674 naming gate genuinely doesn't apply; **C** → hover flips instantly via `background-image` while the toggle keeps its 150ms ease. Scope matches (`SegmentedControl` correctly untouched). The `Checkbox.test.tsx` snapshot the ticket predicted would break was updated as instructed.

### Verification behind this review

The three new utilities were compiled through Tailwind v4 with the repo's patched `tailwindcss-react-aria-components` to check the specificity claims rather than take them on trust, and `react-aria-components@1.19.0` was checked to confirm `data-readonly` / `data-indeterminate` are actually rendered on all three groups. **The PR's central claim holds**: `not-*` compiles to `:not(*:is(…))`, each clause adds a class-level weight, and the exclusions really are load-bearing. The `group-indeterminate` fix is also genuine — RAC renders `data-indeterminate`, and the one *positive* test that survived (`backgroundColor === borderColor`) proves the fix works, since a white box with a resting border would fail it.

### Checklist
| | |
|---|---|
| TypeScript — no `any`, `import type`, strict | ✅ |
| Component patterns — theme-only change, no API/`className` surface | ✅ |
| Styling — token-derived, no hardcoded colors, no theme z-index | ✅ |
| Testing — Vitest, `userEvent`, `getByRole`, `component-test` tag, `.test()` not `play` | ✅ |
| React performance | n/a (no component logic changed) |
| Accessibility | ⚠️ see Critical #1 |
| Changeset | ✅ patch, `@marigold/theme-rui` |

---

### Issues Found

#### 🔴 Critical

**1. `Checkbox.styles.ts:19` — the hover border outranks the focus-visible border, degrading the focus indicator below 3:1.**

This is your own documented trap, applied in the direction that didn't get checked. Compiled specificity:

```
group-focus-visible/checkbox:border-(--ui-border-color)   → (0,2,0)
group-hover/…:not-…:not-…:not-…:border-[oklch(…+0.18)]    → (0,5,0)   ← 3 × :not()
```

Both set `border-color`, so on an **unselected, enabled checkbox that is keyboard-focused *and* hovered** (Tab to it, then move the mouse over it) the hover edge wins and the border never flips to `--color-ring`. `ui.css:187-192` states the consequence in the theme's own words:

> both parts matter: the *border* flips to full-opacity `--color-ring` … the halo alone measures 2.08:1 — below the 3:1 a focus indicator needs

So the indicator falls back to the halo alone, at 2.08:1 — failing WCAG 1.4.11. The PR body ticks "Accessibility reviewed" and argues *against* `!important` on the selected border precisely to protect this rule, then outranks it anyway via the `:not()` chain. Fix is one more clause, consistent with the file's idiom:

```
'group-hover/checkbox:not-group-focus-visible/checkbox:not-group-read-only/checkbox:not-group-selected/checkbox:not-group-indeterminate/checkbox:border-[…]'
```

`Radio` and `Switch` are **not** affected — neither applies `border-(--ui-border-color)`, so there's no conflicting declaration (see suggestion #6).

#### ⚠️ Warnings

**2. Every new hover test is a negative assertion — the deleted positive ones were the only proof the rules match at all.** Commit `508e513` reasons that "hover darkens…" is implied by the "leaves … alone" cases. It isn't: a rule that never matches passes all eleven remaining tests. That is exactly the failure signature of the bug this PR fixes — `group-[indeterminate]` compiled to a dead selector and CI stayed green for its whole life. Two concrete ways the current suite would stay green on a broken feature:

- The `Switch` tint is an unusual single-stop `linear-gradient(color 0 0)`. If it ever failed to parse, `backgroundImage` is `none` and all three `expect(tintOf(track)).toBe('none')` assertions pass.
- Combined with #3, pointing `getTrack` at the wrong element also yields `'none'`.

Restoring one positive assertion per component (the deleted Switch one was the strongest — it asserted the tint appears **and** `backgroundColor` stays put, which is the entire architectural argument for using `background-image`) would cost three tests and close the gap. Their `disableSnapshot: false` also made them the only Chromatic coverage of a hovered state; that's gone too.

**3. `Switch.stories.tsx:153` — `querySelector('[class*="rounded-full"]')` also matches the thumb.** It works only because the track precedes the thumb in document order, and since every assertion on it is `toBe('none')`, a wrong-element match fails silently rather than erroring. `.closest('label')?.querySelector('div > div')`, or asserting the element also carries `w-7`, would fail loudly instead.

#### 💡 Suggestions

**4. `Checkbox.styles.ts:16-18` / `Radio.styles.ts:27` — say why `disabled` is absent from the exclusion chain.** It's covered by the `!` on `group-disabled/*:border-disabled-surface!`, but the comment directly above says "removing one lets hover repaint that state", so the omission reads as a bug — especially next to `Switch.styles.ts:32`, which *does* exclude `disabled` explicitly (correctly, since its disabled rule isn't `!important` and the gradient layers over `background-color` rather than replacing it).

**5. `getIcon` / `borderOf` are duplicated verbatim** in `Checkbox.stories.tsx:130-138` and `Radio.stories.tsx:93-101`. Worth a shared helper.

**6. Follow-up, pre-existing and out of scope:** `Radio.styles.ts:24` and `Switch.styles.ts:33` apply `ui-state-focus` on slots whose border never reads `--ui-border-color`, so those focus indicators are already the halo alone — the exact case `ui.css:191-192` warns about ("On a borderless row it silently degrades to the halo"). Not caused by this PR, but the darker hover edge sits right next to that halo now. Probably its own ticket.

### Summary

The engineering here is unusually careful: the specificity analysis is correct and verifiable, the three open ticket decisions are each resolved with a documented rationale, the `background-image` trick is a genuinely elegant answer to DST-1436's constraint, and the alpha math (`0.14/(1-alpha)` → 0.30) checks out. #1 is the one that should land before merge — it's the same insight as the reviewer note, just aimed at `focus-visible` instead of `selected`, and one added clause fixes it. #2 is the one judgement call worth pushing back on: that commit traded away the only tests that could catch a silently-dead rule, in a PR whose headline bug fix *was* a silently-dead rule.

---

themes/theme-rui/src/components/Checkbox.styles.ts
```diff
@@ -11,8 +11,12 @@ export const Checkbox: ThemeComponent<'Checkbox'> = {
...
       'group-disabled/checkbox:group-selected/checkbox:bg-disabled-surface group-disabled/checkbox:border-disabled-surface! group-disabled/checkbox:text-disabled group-disabled/checkbox:cursor-not-allowed',
       'group-selected/checkbox:border-selected-bold group-selected/checkbox:bg-selected-bold group-selected/checkbox:text-selected-bold-foreground',
-      'group-[indeterminate]/checkbox:border-selected-bold group-[indeterminate]/checkbox:bg-selected-bold group-[indeterminate]/checkbox:text-selected-bold-foreground',
+      'group-indeterminate/checkbox:border-selected-bold group-indeterminate/checkbox:bg-selected-bold group-indeterminate/checkbox:text-selected-bold-foreground',
       'group-hover/checkbox:group-disabled/checkbox:bg-disabled-surface',
+      // Matches ToggleButton's hover step, not its value — this slot idles at
+      // +0.06, so its +0.12 would be half the step. Each `:not()` outranks a
+      // state rule above; removing one lets hover repaint that state.
+      'group-hover/checkbox:not-group-read-only/checkbox:not-group-selected/checkbox:not-group-indeterminate/checkbox:border-[oklch(from_var(--color-control-border)_l_c_h_/_calc(alpha_+_0.18))]',
```
jim761: 🔴 **This rule outranks the focus-visible border, degrading the focus indicator below 3:1.**

Your own documented trap, aimed at the state you did not check. Compiled through Tailwind v4 with the patched `tailwindcss-react-aria-components`:

```
group-focus-visible/checkbox:border-(--ui-border-color)   → (0,2,0)
group-hover/…:not-…:not-…:not-…:border-[oklch(…+0.18)]    → (0,5,0)   ← 3 × :not()
```

Both set `border-color`, so on an **unselected, enabled checkbox that is keyboard-focused *and* hovered** (Tab to it, then move the mouse over it) this wins and the border never flips to `--color-ring`. Per `ui.css:187-192`:

> both parts matter: the *border* flips to full-opacity `--color-ring` … the halo alone measures 2.08:1 — below the 3:1 a focus indicator needs

So the indicator falls back to the halo alone at 2.08:1, failing WCAG 1.4.11. This is the same mechanism your reviewer note describes — the note applies it to `selected`/`disabled`, but `group-focus-visible:border-(--ui-border-color)` on line 11 sits at the same (0,2,0) and loses too. One more clause fixes it:

```suggestion
      'group-hover/checkbox:not-group-focus-visible/checkbox:not-group-read-only/checkbox:not-group-selected/checkbox:not-group-indeterminate/checkbox:border-[oklch(from_var(--color-control-border)_l_c_h_/_calc(alpha_+_0.18))]',
```

`Radio` and `Switch` are not affected — neither slot applies `border-(--ui-border-color)`, so there is no conflicting declaration.

themes/theme-rui/src/components/Radio.styles.ts
```diff
@@ -24,6 +24,8 @@ export const Radio: ThemeComponent<'Radio'> = {
       'group-focus-visible/radio:ui-state-focus outline-none',
       'group-disabled/radio:group-selected/radio:bg-disabled-surface group-disabled/radio:border-disabled-surface! group-disabled/radio:cursor-not-allowed',
       'group-selected/radio:border-selected-bold group-selected/radio:bg-selected-bold group-selected/radio:text-selected-bold-foreground',
+      // Same step and the same load-bearing exclusions as Checkbox.
+      'group-hover/radio:not-group-read-only/radio:not-group-selected/radio:border-[oklch(from_var(--color-control-border)_l_c_h_/_calc(alpha_+_0.18))]',
```
jim761, line 33: 💡 Worth a word on why `disabled` is *not* in the exclusion chain here or in `Checkbox.styles.ts:19`.

It is covered by the `!` on `group-disabled/radio:border-disabled-surface!` — but the comment directly above says "removing one lets hover repaint that state", so the omission reads as an oversight rather than a decision. It looks more inconsistent next to `Switch.styles.ts:32`, which *does* exclude `disabled` explicitly (correctly — its disabled rule is not `!important`, and the gradient layers over `background-color` instead of replacing it).

packages/components/src/Switch/Switch.stories.tsx
```diff
@@ -145,6 +145,83 @@ Basic.test(
   }
 );
 
+// There is no accessible way to reach the track. It is the first rounded
+// element in the label; the thumb is nested inside it.
+const getTrack = (switchEl: HTMLElement) => {
+  const track = switchEl
+    .closest('label')
+    ?.querySelector('[class*="rounded-full"]');
```
jim761: ⚠️ `[class*="rounded-full"]` also matches the thumb — `thumb` carries `rounded-full` too, so this works only because the track precedes it in document order.

That would be harmless if a wrong match failed loudly, but every assertion on the result is `expect(tintOf(track)).toBe('none')`, and the thumb also has no `background-image`. So pointing this at the wrong element passes all three tests silently.

`.closest('label')?.querySelector('div > div')`, or additionally asserting the element carries `w-7`, would fail instead of quietly passing.

packages/components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -126,6 +126,101 @@ Basic.test(
...
+  const icon = checkbox.closest('label')?.querySelector('[aria-hidden="true"]');
+
+  expect(icon).not.toBeNull();
+
+  return icon!;
+};
+
+const borderOf = (el: Element) => getComputedStyle(el).borderColor;
+
+// One test per state, because each needs its own `args`. Note `.test()` calls are
```
jim761, line 183: ⚠️ **Every hover test left here is a negative assertion, and the deleted positive ones were the only proof the rules match at all.**

`508e513` reasons that "hover darkens…" is implied by the "leaves … alone" cases. It is not: a rule that never matches passes all eleven remaining tests. That is precisely the failure signature of the bug this PR fixes — `group-[indeterminate]` compiled to `:is(indeterminate)` and CI stayed green for its entire life.

Two concrete ways the suite as it stands would stay green on a broken feature:

- The `Switch` tint is an unusual single-stop `linear-gradient(color 0 0)`. If that ever failed to parse, `backgroundImage` is `none` and all three `expect(tintOf(track)).toBe('none')` assertions pass.
- Combined with the `getTrack` selector, aiming at the wrong element also yields `'none'`.

Restoring one positive assertion per component would close the gap for three tests. The deleted `Switch` one was the strongest of them — it asserted the tint appears **and** that `backgroundColor` stays put, which is the whole architectural argument for using `background-image` over `background-color`. Their `disableSnapshot: false` also made them the only Chromatic coverage of a hovered state.

packages/components/src/Radio/Radio.stories.tsx
```diff
@@ -88,6 +89,60 @@ export const Basic = meta.story({
   ),
 });
 
+// There is no accessible way to reach the element that paints the radio.
+const getIcon = (radio: HTMLElement) => {
```
jim761: 💡 `getIcon` and `borderOf` are duplicated verbatim from `Checkbox.stories.tsx:130-138`. Worth lifting into a shared test helper.

themes/theme-rui/src/components/Checkbox.styles.ts
sebald: Confirmed, fixed in 66474bcb8 — took your suggestion. I verified the specificity by compiling rather than by reading, and your arithmetic is exact:

```
group-focus-visible/checkbox:border-(--ui-border-color)   (0,2,0)
group-hover/…:not-…:not-…:not-…                           (0,5,0)
```

Both `:is(...)` wrappers resolve to (0,1,0) via the `:where()` collapse, and each of the three `:not()`s adds another (0,1,0). So hover wins, as you said.

Then I rendered it and measured, and the outcome is **less severe than the comment says** — worth correcting the record since the number is quotable:

| state | border | contrast on white |
|---|---|---|
| idle | control-border +0.06 | 2.16:1 |
| hover | +0.18 | 3.06:1 |
| focus | `--color-ring` | **5.50:1** |
| focus + hover (the bug) | +0.18 | **3.06:1** |

The border is not lost and the indicator does not "fall back to the halo alone at 2.08:1" — the border is still painted at 3.06:1, which clears 3:1 on its own. The real defect is narrower and still worth fixing: **focus + hover renders pixel-identically to plain hover**, so the halo is the only thing distinguishing focused from hovered, and that is the 2.08:1 part. It is a focus-*distinguishability* failure rather than a non-text-contrast one.

Same fix either way, and after it `focus + hover` is back to `--color-ring` at 5.50:1.

I also checked the other exclusions while the harness was up, and they all hold: `selected` and `indeterminate` land on selected-bold (17.28:1), `read-only` stays at idle, and `disabled` correctly resolves to disabled-surface — the `!important` wins, as you note in the Radio thread.

Guarded with a story test using the existing `contrast.utils` helpers. It fails in Firefox against the unfixed rule:

```
× Hover does not weaken the focus indicator
```

**One finding beyond this PR.** `ui-state-focus` sets `--ui-border-color: var(--color-ring)`, but on Radio and Switch *nothing consumes it* — I compiled both and got zero `border-color: var(--ui-border-color)` declarations. So your "not affected" is right about the conflict, and the reason is that their focus indicator is **already** the halo alone at 2.08:1, in every state, hover or not. That is exactly the degradation `ui.css:187-192` warns about. It is on `main` today and predates this PR, so I have left it alone rather than widen the scope — but it wants a ticket, and it is a bigger problem than the one you found here.

packages/components/src/Checkbox/Checkbox.stories.tsx
sebald: You are right and I was wrong in 508e513 — reverted in 66474bcb8, one positive assertion restored per component.

The reasoning in that commit ("implied by the negative tests") does not hold, and I confirmed it empirically rather than by argument. Mutating the hover rule so it can never match — `not-group-nonsense/checkbox:`, the same shape as the `group-[indeterminate]` bug:

```
Tests  2 failed | 22 passed (24)
× Hover darkens the border
× Hover darkens the border of an unselected radio
```

22 tests green against a feature that does nothing. That is your point, exactly, and the two restored assertions are the only things that caught it.

I restored the Switch one in full, including the `fillOf` half — you are right that it is the one test pinning the architectural choice (tint arrives as `background-image`, transitioned `background-color` untouched), and it now sits next to the `transitionProperty` test that pins the other half. All three keep `disableSnapshot: false`, so the Chromatic coverage of a hovered state comes back too.

packages/components/src/Switch/Switch.stories.tsx
sebald: Fixed in 66474bcb8. Confirmed the thumb carries `rounded-full` too, so the selector was resolving on document order alone.

Now `.closest("label")?.querySelector("div > div")`, plus the width assertion you suggested — I used `28px` (the computed `w-7`) rather than the class, since asserting the class would pass on a wrong element that happened to carry it:

```ts
expect(getComputedStyle(track!).width).toBe(28px);
```

The thumb is `size-3`, so pointing this at it now fails in `getTrack` instead of silently reading `"none"` off it three times.

themes/theme-rui/src/components/Radio.styles.ts
sebald: Both documented in 66474bcb8. You are right that the omission read as an oversight — and it is load-bearing in the opposite direction from the rest of the chain, which makes it worth spelling out.

Radio now says why `disabled` is out (`border-disabled-surface!` carries `!important` and wins regardless), why Switch has to name it explicitly (its disabled rule does not), and why Radio omits `focus-visible` as well (this slot never flips its border on focus, so there is nothing for hover to outrank). Checkbox carries the matching note next to its new `focus-visible` clause.

Verified rather than assumed: `disabled + hover` resolves to disabled-surface at 1.11:1, so the `!` does win.

packages/components/src/Radio/Radio.stories.tsx
sebald: Done in 66474bcb8 — lifted into `packages/components/src/control.utils.ts` as `controlIcon` and `borderOf`, following the `contrast.utils.ts` pattern (browser-only, kept out of `index.ts`, documented as internal test tooling).

Left the Switch helpers where they are: `getTrack`/`tintOf`/`fillOf` are track-specific and have no second caller, so sharing them would only add indirection.

themes/theme-rui/src/components/Switch.styles.ts
sebald: Trimmed to four lines in 66474bcb8, keeping what a reader cannot recover from the code: why `background-image` and not `background-color` (the track keeps `transition-colors` from DST-1436, and `background-image` is not transitioned, so hover lands instantly), and why the alpha is `0.14/(1 - alpha)` rather than a flat step (it layers over the track instead of replacing it).

I would rather not cut further. That one line encodes two non-obvious decisions, and both have a test pinning them precisely because they are easy to "simplify" back into a bug.

themes/theme-rui/src/components/Checkbox.styles.ts
sebald: Filed the Radio/Switch finding as **DST-1733** so it does not ride along here: https://reservix.atlassian.net/browse/DST-1733

Scoping it turned up more than the two slots. `ui-state-focus` only sets `--ui-border-color`; something else has to consume it — one of `ui-frame`/`ui-soft`/`ui-contrast`/`ui-control`/`ui-input` (each applies `ring-1 ring-[var(--ui-border-color, …)]`), or an explicit `border-(--ui-border-color)` as Checkbox writes by hand. **23 slots across 15 components apply it with no consumer in scope**, so the halo is the only thing that changes on focus.

Nine are fine (`ActionBar.container`, `Checkbox.checkbox`, `DateField.field`, `Input.input`, `NumberField.group`, `SegmentedControl.indicator`, `Select.select`, `Tag.tag`, `TagField.trigger`), and Menu + FileField were already fixed in DST-1661 (#5683) — which is also the precedent for the fix shape, `ui-state-focus-item`.

Only Radio and Switch are confirmed by measurement; the other 21 are static-sweep candidates the ticket flags as needing individual confirmation. Two false-positive traps are written down there, because I hit both: the `ui-*` utilities consume the variable as a `ring` rather than a `border` (my first sweep said 34 slots and was wrong), and several of these files mention `ui-state-focus` in prose, so an uncommented grep over them over-reports again.

jim761 (approved):

OsamaAbdellateef (approved):

PR #5753, opened 2026-08-17
feat(DST-1521): commit the /grill scoping skill and the .memory store

sebald:
# Description

Moves the `grill-me` skill out of personal config (`~/.claude/skills/`) and into the repo, so the team gets it from a fresh clone and it is reviewed like code. Adds the `.memory/` store it maintains: a glossary at `CONTEXT.md` and ADRs under `adr/`, both committed and deliberately **not** gitignored.

Closes DST-1521

> [!IMPORTANT]
> **Draft — blocked by #5750 (DST-1528/1529).**
> This builds on the skill conventions committed there, so it cannot merge until that does. GitHub retargets this to `main` automatically when #5750 merges, at which point this comes out of draft.
>
> #5754 (DST-1530) is in turn stacked on this one.

## What's here

| Path | |
| --- | --- |
| `.claude/skills/grill/SKILL.md` | The skill. One interrogation workflow; the documentation layer is gated on step 3 and stays off for sessions that aren't about the code. |
| `.memory/README.md` | Scope, the write rule, the boundary with `CLAUDE.md`, merge-safety and ADR conventions. |
| `.memory/CONTEXT.md` | Glossary skeleton. **Intentionally empty** — the ticket scopes seeding to DST-1530. |
| `.memory/adr/TEMPLATE.md` | ADR template with a mandatory "alternatives rejected" section. |
| `CLAUDE.md` | A "Scoping work: `/grill`" entry under AI Toolkit. |

## Five additions beyond the ticket

The ticket's decisions all stand — location, commit policy, one skill not two. Research into how teams are actually running repo-committed agent memory surfaced five gaps worth closing before the pattern sets. Recorded in full [on the ticket](https://reservix.atlassian.net/browse/DST-1521).

1. **ADR frontmatter**, not just a filename — a stable citable `id`, a `status` with a supersede chain, and `applies_to` globs so an agent loads only the records governing the file it's editing.
2. **An explicit boundary with `CLAUDE.md`** — rules there, meanings in `CONTEXT.md`, history in `adr/`. No rule lives in two places. Duplicated rules disagree within a couple of months and the model then picks one arbitrarily; DST-1529 was this exact failure with an inventory instead of a rule.
3. **Merge safety** — one term per heading, one decision per file. Two branches appending contradictory notes to a shared blob merge cleanly in git and leave a file holding two things that can't both be true, with nothing failing to signal it.
4. **A write rule** — persist only what's durable, cross-session and settled.
5. **A stated ceiling** — these records are advisory. Nothing checks compliance. Recorded so nobody later mistakes an ADR for a guarantee.

`.memory/` being committed is also the security control: persistent memory poisoning is tracked as OWASP ASI06 in the 2026 Agentic AI Top 10, and Anthropic's agent-memory docs warn that a writable store plus untrusted input lets an injection write content later sessions read as trusted. `/grill` reads Jira tickets, PR comments and web pages. Review-on-merge is what stops a bad entry becoming something every future session believes.

## Potential follow-up: namespacing first-party skills

This PR takes the cheap version — every repo skill's description now opens with `Marigold repo — `, so ours group visibly in a `/` menu that also lists plugin and personal skills. Invocation is unchanged (`/grill`, not `/mg-grill`).

The heavier option is a real prefix. Deliberately **not** done here; noted so the trade-off is on record.

**Pros**
- Unambiguous ownership at the call site — `/mg-review-pr` can't be confused with a personal or plugin skill of the same name.
- Pre-empts collisions. `review-pr` and `create-pr` are generic enough that someone's personal config could plausibly claim them.
- Makes "is this reviewed team tooling?" answerable without knowing where skills live.

**Cons**
- **A colon impersonates a real mechanism.** The harness already namespaces plugin skills (`vercel:react-best-practices`) and directory-scoped skills (`apps/web:deploy`) with `:`. A hand-written `mg:grill` is one flat literal name that every reader parses as a plugin called `mg`.
- **The problem is largely already solved.** Plugin skills arrive pre-namespaced. What's left unprefixed is personal skills — few, per-developer, and currently zero collisions against our four.
- **Permanent per-invocation cost for a cosmetic gain.** Typing the prefix forever, on every use.
- **Nothing enforces it.** The first skill that lands without the prefix makes absence meaningless and the convention becomes misleading rather than merely useless — the class of drift DST-1529 concluded we should stop committing.
- **Location already answers it and can't drift.** `.claude/skills/` is ours; `.claude/README.md` says so.
- Per our own conventions the **description** does the triggering, not the name — so a prefix adds tokens without buying any.

If we ever want real namespacing, the honest route is packaging these as a local plugin and letting the harness produce `marigold:grill` for free. The cost is that plugins are installed per-user and versioned upstream, which gives up the "committed, ours to edit" property that made moving `/grill` in-repo worth doing in the first place.

## Test Instructions

1. `.memory/` is committed on purpose — confirm `git check-ignore .memory` reports nothing.
2. Open a fresh session and check `/grill` appears, with `Marigold repo — ` leading its description alongside `/create-pr`, `/review-pr`, `/vrt`.
3. Read `.memory/README.md` and push on the boundary rule — it's the one thing here that decides whether this store stays useful or rots.
4. A live `/grill` run on a real under-specified idea is the actual test; that's happening next and findings will be reported here.

## Breaking Changes

No. Nothing outside `.claude/`, `.memory/` and `CLAUDE.md`; no package or theme code touched, so no changeset.

## Checklist

### DIFF
--- a/.claude/skills/grill/SKILL.md
+++ b/.claude/skills/grill/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: grill
-description: Interrogate an under-specified idea one question at a time until every decision branch is resolved, recommending an answer for each. When the work concerns this codebase, also maintain the glossary in .memory/CONTEXT.md and record hard-to-reverse decisions as ADRs under .memory/adr/. Use before writing a ticket, before starting non-trivial work, when a plan needs stress-testing, or when the user types `/grill`.
+description: Marigold repo — Interrogate an under-specified idea one question at a time until every decision branch is resolved, recommending an answer for each. When the work concerns this codebase, also maintain the glossary in .memory/CONTEXT.md and record hard-to-reverse decisions as ADRs under .memory/adr/. Use before writing a ticket, before starting non-trivial work, when a plan needs stress-testing, or when the user types `/grill`.
 ---
 
 # Grill
--- a/.claude/skills/grill/SKILL.md
+++ b/.claude/skills/grill/SKILL.md
@@ -1,0 +1,82 @@
+---
+name: grill
+description: Interrogate an under-specified idea one question at a time until every decision branch is resolved, recommending an answer for each. When the work concerns this codebase, also maintain the glossary in .memory/CONTEXT.md and record hard-to-reverse decisions as ADRs under .memory/adr/. Use before writing a ticket, before starting non-trivial work, when a plan needs stress-testing, or when the user types `/grill`.
+---
+
+# Grill
+
+Turn an under-specified idea into a scoped one by interrogating it, then leave behind the part worth keeping.
+
+Two jobs, and the second only sometimes:
+
+1. **Always** — interrogate. One question at a time, each with your recommended answer, until every branch of the decision tree is resolved.
+2. **When the session concerns this codebase** — record. Vocabulary goes to `.memory/CONTEXT.md`; decisions that are hard to reverse go to `.memory/adr/`.
+
+Read [.memory/README.md](../../../.memory/README.md) before writing anything to `.memory/`. It defines what belongs there and — more importantly — what does not.
+
+## Usage
+
+```
+/grill                 # scope whatever is being discussed
+/grill <idea>          # scope a specific idea
+```
+
+## Workflow
+
+### 1. Establish what is being scoped
+
+State the idea back in one or two sentences before asking anything. If that restatement is already wrong, the interrogation is about to go the wrong way, and the user can correct it for free.
+
+### 2. Interrogate, one question at a time
+
+The rules that make this useful rather than annoying:
+
+- **One question per turn.** A list of six questions gets one answer and five shrugs.
+- **Always recommend.** Every question carries your recommended answer and the reason. The user is confirming or overruling a position, not filling in a blank form.
+- **Answer from the code, don't ask.** If a question is answerable by reading the repo, read the repo. Asking the user what the code already says wastes the turn and reduces trust in the questions that genuinely need them.
+- **Resolve dependencies in order.** When answer A changes which questions B and C even are, ask A first.
+- **Challenge vague terms against the code.** When someone says "the layout component" or "the wrapper", find out which one they mean. Half of scoping is discovering that two people have been using one word for two things — that discovery is what step 4 records.
+
+Stop when the remaining questions no longer change what gets built.
+
+### 3. Decide whether the documentation layer applies
+
+It applies when the session is about this codebase — a component, an API, an architectural call, a convention.
+
+It does not apply to one-off questions, throwaway exploration, or scoping that concerns process rather than code. In those sessions, finish at step 6 and write nothing. **A `/grill` session that records nothing is a normal outcome, not a failure.**
+
+### 4. Update the glossary
+
+If the session settled what a term means, add or correct its entry in `.memory/CONTEXT.md`.
+
+One term per `###` heading, kept alphabetical. Define the term and name what it is *not* when there is a neighbouring term it gets confused with — that confusion is usually why the entry is needed.
+
+Do not add rules here. A rule the agent must follow belongs in `CLAUDE.md`. The glossary says what words mean; it does not say what to do.
+
+### 5. Write an ADR when the bar is met
+
+Write one only when at least one of these holds:
+
+- **Hard to reverse.** Undoing it means a migration, a breaking change, or touching many files.
+- **Surprising without context.** Someone reading the code cold would "fix" it back, not understanding why it is the way it is.
+- **A live alternative was rejected.** Without the record, the same alternative gets proposed again next quarter.
+
+Copy `.memory/adr/TEMPLATE.md` to `.memory/adr/NNNN-<slug>.md`, taking the next free number. Fill in the frontmatter — `id`, `status`, `date`, and an `applies_to` glob narrow enough that the record only loads for work it governs.
+
+Set `status: proposed`. It becomes `accepted` when the PR carrying it merges.
+
+**Never edit an accepted ADR to change its decision.** Write a new one and set the old record's status to `superseded-by ADR-NNNN`. The history is the point; correcting the past in place destroys it.
+
+If none of the three criteria hold, write nothing. Records with the thinking removed are worse than absence — they look like documentation while doing nothing, and they cost context on every future session that loads them.
+
+### 6. Close with a scoped summary
+
+End with a short summary the user can act on: what is being built, the decisions that were settled, and anything explicitly ruled out of scope.
+
+Offer to hand it to `/create-ticket` once that skill exists. Do not file the ticket from here.
+
+If you wrote to `.memory/`, say exactly which files changed. These are committed, reviewed files — the user needs to know they are in the diff.
+
+## Limits worth knowing
+
+Records under `.memory/` are **advisory**. They enter the context window and influence what the agent does; nothing verifies compliance. An ADR that no check enforces is a well-argued comment, and a probabilistic reader cannot turn it into a guarantee. Write them because they compress genuine context, not because they constrain anything.
--- a/.claude/skills/grill/SKILL.md
+++ b/.claude/skills/grill/SKILL.md
@@ -47,11 +47,9 @@ It does not apply to one-off questions, throwaway exploration, or scoping that c
 
 ### 4. Update the glossary
 
-If the session settled what a term means, add or correct its entry in `.memory/CONTEXT.md`.
+If the session settled what a term means, add or correct its entry in `.memory/CONTEXT.md`, following the glossary conventions in [.memory/README.md](../../../.memory/README.md).
 
-One term per `###` heading, kept alphabetical. Define the term and name what it is *not* when there is a neighbouring term it gets confused with — that confusion is usually why the entry is needed.
-
-Do not add rules here. A rule the agent must follow belongs in `CLAUDE.md`. The glossary says what words mean; it does not say what to do.
+The one worth repeating because it is what gets this step wrong: do not add rules here. A rule the agent must follow belongs in `CLAUDE.md`. The glossary says what words mean; it does not say what to do.
 
 ### 5. Write an ADR when the bar is met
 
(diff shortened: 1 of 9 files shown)


.memory/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+One term per `###` heading, alphabetical. Define the term, and name what it is _not_ when a neighbouring term gets confused with it.
+
+Split the file once it passes roughly 200 lines — long files get read less faithfully, and a glossary nobody finishes reading is a glossary that quietly stops working.
+
+**One term per heading is a merge-safety rule, not just tidiness.** Two branches appending different definitions of the same term in the same place produce a git conflict, which someone resolves. Two branches appending to a shared blob merge cleanly and leave the file holding two definitions that cannot both be true — with nothing failing to tell you.
+
+## ADR conventions
+
+One decision per file, `adr/NNNN-<slug>.md`, numbered sequentially from `0001`. Start from `TEMPLATE.md`.
```
aromko: ⚠️ **ADR numbering has the merge hole this file warns about**

Numbers are hand-allocated — here and at `grill/SKILL.md:64` ("taking the next free number"). Two branches both taking `0002` get *different filenames*, so git merges both cleanly and you end up with two `ADR-0002`. Nothing fails.

That breaks the property sold three lines down at :52 — `id` "stable and citable" — and makes `superseded-by ADR-0002` ambiguous.

This is the same failure mode you call out for the glossary at :44. Merge-safety got applied to the headings but not to the numbering scheme.

Cheapest fix with no new tooling: take `NNNN` from the DST ticket rather than from a sequence. Tickets are already unique, already on the branch and the commits, and it links the record to its ticket for free.

.memory/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Frontmatter carries:
+
+- **`id`** — stable and citable (`ADR-0001`), so a review comment can name the record it means.
+- **`status`** — `proposed` → `accepted`, then `superseded-by ADR-NNNN`. Never anything else.
+- **`applies_to`** — globs for the paths the decision governs. Narrow ones let an agent load only the records relevant to the file it is editing rather than the whole set.
```
aromko: ⚠️ **`applies_to` is described as a loading mechanism that nothing implements**

"Narrow ones let an agent load only the records relevant to the file it is editing rather than the whole set."

Nothing reads `.memory/`. Only `CLAUDE.md`, this file and `SKILL.md` mention the path; no script or config consumes the glob, and nothing tells an agent to load `adr/` at all. As it stands the store is write-only.

The disclaimer at :62 covers *compliance* ("nothing verifies compliance"), which is a different claim from *retrieval*. Either name who does the loading, or soften this to what it actually is — a hint for whoever, human or agent, has been pointed at the directory.

.claude/skills/grill/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+
+It applies when the session is about this codebase — a component, an API, an architectural call, a convention.
+
+It does not apply to one-off questions, throwaway exploration, or scoping that concerns process rather than code. In those sessions, finish at step 6 and write nothing. **A `/grill` session that records nothing is a normal outcome, not a failure.**
+
+### 4. Update the glossary
+
+If the session settled what a term means, add or correct its entry in `.memory/CONTEXT.md`.
+
+One term per `###` heading, kept alphabetical. Define the term and name what it is *not* when there is a neighbouring term it gets confused with — that confusion is usually why the entry is needed.
```
aromko: ⚠️ **Four rules are stated twice, against this PR's own headline rule**

`.memory/README.md:18` — "No rule lives in two places. Copies do not stay in sync." But:

| Rule | `.memory/README.md` | this file |
| --- | --- | --- |
| Glossary conventions | :40 | :52 (here) |
| ADR file naming | :48 | :64 |
| Never edit an accepted ADR | :58 | :68 |
| Advisory ceiling | :62 | :82 |

Line 15 already sends the reader to the README before writing anything, so these can be pointers instead of copies. As written, the PR that establishes the no-duplication rule is the first thing to break it.

.claude/skills/grill/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+
+Write one only when at least one of these holds:
+
+- **Hard to reverse.** Undoing it means a migration, a breaking change, or touching many files.
+- **Surprising without context.** Someone reading the code cold would "fix" it back, not understanding why it is the way it is.
+- **A live alternative was rejected.** Without the record, the same alternative gets proposed again next quarter.
+
+Copy `.memory/adr/TEMPLATE.md` to `.memory/adr/NNNN-<slug>.md`, taking the next free number. Fill in the frontmatter — `id`, `status`, `date`, and an `applies_to` glob narrow enough that the record only loads for work it governs.
+
+Set `status: proposed`. It becomes `accepted` when the PR carrying it merges.
```
aromko: 💡 **`proposed` → `accepted` has no actor**

"It becomes `accepted` when the PR carrying it merges" is passive with no mechanism behind it — same at `.memory/README.md:53`. Nothing flips it, so records will sit at `proposed` indefinitely and the field stops carrying information.

Name the step. The author flipping it in the same PR once the review approves is the cheapest version, and it keeps the transition inside something a human already looks at.

.memory/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Anything task-local, fast-changing or uncertain stays in the session. The store is only useful while it is small enough to load and trustworthy enough to believe; every marginal entry costs context on every future session and dilutes both.
+
+Prune on sight. A wrong entry is worse than a missing one, because the agent has no way to doubt it.
+
+## Glossary conventions
+
+One term per `###` heading, alphabetical. Define the term, and name what it is _not_ when a neighbouring term gets confused with it.
+
+Split the file once it passes roughly 200 lines — long files get read less faithfully, and a glossary nobody finishes reading is a glossary that quietly stops working.
```
aromko: 💡 **"Split the file" doesn't say into what**

`CONTEXT.md` is the single path named in `CLAUDE.md:335`, `grill/SKILL.md:13` and the table at the top of this file. A split silently breaks all three pointers, and the rule gives no successor shape.

Either name it now (`CONTEXT/<area>.md` with `CONTEXT.md` kept as the index) or drop the rule — a threshold nobody can act on is one more line of context on every session.

.memory/adr/TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
+---
+id: ADR-NNNN
+status: proposed # proposed | accepted | superseded-by ADR-NNNN
+date: YYYY-MM-DD
+applies_to:
+  - 'packages/components/src/**'
```
aromko: 💡 **This placeholder is a plausible real value**

`'packages/components/src/**'` will get copied unedited and silently over-scope a record. Its siblings — `ADR-NNNN`, `YYYY-MM-DD` — are obviously placeholders; this one isn't.

```suggestion
  - 'CHANGEME/**'
```

.memory/README.md
```diff
@@ -0,0 +1,64 @@
...
+**This directory is committed and must not be added to `.gitignore`.** That is the whole point. Shared context has to travel with the repo and go through review like code — an agent-written note that nobody reviewed is not team knowledge, it is a rumour with a file path. It also happens to be the control that matters most for safety: `/grill` sessions read Jira tickets, PR comments and web pages, and content from those sources can end up proposed as a memory. Review is what stops a bad one from becoming something every future session treats as fact.
+
+`/grill` writes here. So can you, by hand.
+
+## What lives here
+
+| Path                 | Holds                          | Nature                              |
+| -------------------- | ------------------------------ | ----------------------------------- |
+| `CONTEXT.md`         | Glossary — what our terms mean | Descriptive, corrected in place     |
+| `adr/NNNN-<slug>.md` | Why a decision was made        | Historical, immutable once accepted |
```
aromko, line 14: 💡 **The table omits `.memory/tasks/`**

DST-1521 names it as part of this same store — "the per-ticket specs `/pick-up` writes to `.memory/tasks/`" — under the same commit policy.

A row marked reserved stops the next author inventing a different location for it.

aromko (commented):
## Automated Code Review

Reviewed against `9649a4f61`. Nine markdown files, no React or TypeScript, so no performance or a11y pass applies. Visual regression is **not needed** — nothing under `packages/`, `themes/`, `.storybook/`, `*.stories.tsx` or `*.styles.ts` is touched.

**Verified**

- `git check-ignore .memory` reports nothing, so the commit policy holds.
- All four relative links in the changed files resolve against the tree, including `../../../.memory/README.md` from `grill/SKILL.md`. Worth knowing: `validate-links.yml` only covers `docs/`, so nothing checks these in CI.
- All four first-party skills carry the `Marigold repo — ` prefix, and the `/` menu renders it, so the raw-description claim holds.
- `grill/SKILL.md` matches the shape `.claude/README.md:29` mandates.
- Scope matches DST-1521. All four ticket decisions are honoured, including the ones that were easy to skip, and seeding is correctly left to DST-1530.

**Recommendation: Comment.** Nothing here blocks the merge.

Seven findings inline: three worth fixing, four nits. Findings 1 and 3 are the ones worth resolving before DST-1530 seeds content — numbering gets expensive to change once records exist and are cited, and `applies_to` is currently a promise with no consumer. Finding 2 matters mostly because it is the one rule this PR argues hardest for.

---

.memory/README.md
aromko: Fixed in 107b28dcb — `NNNN` now comes from the DST ticket. `1521-memory-store-conventions.md` is `ADR-1521`.

Unique for free, already on the branch and the commits, and it links the record to the discussion at no cost. The reasoning is written into the README so the next person does not reintroduce a sequence. The one collision left is two ADRs from one ticket, which happens inside a single branch where you can see it — noted as `NNNN-b-<slug>`.

.memory/README.md
aromko: Fixed in 107b28dcb — the retrieval claim is gone.

The bullet now says what the field actually does (lets a reader pointed at the directory tell in one line whether a record concerns their file), states plainly that nothing loads records on its own, and keeps the narrow-glob guidance as preparation rather than as a description of something that exists.

.claude/skills/grill/SKILL.md
aromko: Fixed in 107b28dcb — all four are now stated once, in `.memory/README.md`, with this file pointing at it.

Kept one restatement on purpose: "do not add rules in the glossary". It is the thing that actually gets step 4 wrong, and it is worth the line at the point of the mistake rather than one link away.

.claude/skills/grill/SKILL.md
aromko: Fixed in 107b28dcb — the author flips it, in the same PR, once the review approves and before merge.

Named in both places: `.memory/README.md` on the `status` bullet, and here as something the skill tells the user when it writes a record, so the handoff does not depend on anyone rereading the README.

.memory/README.md
aromko: Fixed in 107b28dcb — splits into `CONTEXT/<area>.md`, with `CONTEXT.md` kept as the index that links them.

That keeps the single path `CLAUDE.md` and the skill both point at valid through a split, which was the actual risk.

.memory/adr/TEMPLATE.md
aromko: Fixed in 107b28dcb — now `CHANGEME/**`.

Also gave `id: ADR-NNNN` an inline comment naming `NNNN` as the DST ticket number, since finding #1 changed where that number comes from.

.memory/README.md
aromko: Fixed in 107b28dcb — added as a row marked reserved, "Not in use yet; `/pick-up` fills it".

aromko (approved):

PR #5754, opened 2026-08-17
docs(DST-1530): seed the .memory store with the specified ADRs and glossary

sebald:
# Description

Seeds the `.memory` store that #5753 established, following DST-1530's specified scope: the three ADRs the ticket names, the glossary terms it lists, and the template at `0000-template.md`.

Two further ADRs are included. They were written first, from a repo analysis rather than from the ticket, and have been **renumbered to 0004/0005** to free the slots DST-1530 allocates.

Was stacked on #5753 (DST-1521) and #5750 (DST-1528/1529); both have merged, so this targets `main` directly and stands on its own.

Closes DST-1530

> [!NOTE]
> **One blocker left: DST-1731.**
>
> The stack has merged — #5750 (DST-1528/1529) and #5753 (DST-1521) are both on `main`, and this now targets `main` directly.
>
> Of the three refactoring tickets, two are done and each shipped a guard, so ADR-0001 and ADR-0003 have been updated to say what is enforced and where each guard stops:
>
> | Ticket | Violation | State |
> | --- | --- | --- |
> | **DST-1729** | Four files imported `I18nProvider` from `@react-aria/i18n` — contradicted ADR-0001 | Fixed in #5756, which added `check-rac-first-imports.mjs` |
> | **DST-1730** | `z-index` back in `ToggleButton.styles.ts`, plus scale drift in `CLAUDE.md` — contradicted ADR-0003 | Fixed in #5757, which added `check-theme-zindex.mjs` |
> | **DST-1731** | `Input` clones children to inject positioning classes — contradicts ADR-0005 | **Open** — `Input.tsx:36,48` |
>
> Rather than soften a record to fit the code, the code gets fixed first; otherwise the store ships describing rules the repo already breaks, which is how records become decorative.

### ADR-0001 — Import from `react-aria-components` whenever it re-exports the API

The best-evidenced of the three. RAC pins its internals exactly; every `@react-aria/*` shell keeps a caret range **by design**, so a consumer's lockfile can legitimately resolve two `react-aria` copies — which means two module instances and two distinct contexts with the same name. A provider from one does not satisfy a `useContext` from the other, and it never reproduces here because our own dedupe hides it.

Sourced from DST-1512, and verified against `packages/components/package.json` (every shell on `^`, RAC on a single `^1.20.0`).

Keeps the exception that looks like a violation: `packages/system`'s formatter tests stay on `@react-aria/i18n`, because a test provider must write to the context its subject reads and `packages/system` has no RAC dependency at all.

### ADR-0002 — Name the container for its structure (`AppShell`), keep "chrome" for the role

The ticket cites a chat-log decision nobody has. `MIGRATION-v18.md` turned out to document it: A3.1 records `<AppLayout>` → `<AppShell>` with the pass-through parts removed, and section 5 states **"The frame: `<AppShell>`"** outright. Meanwhile "chrome" appears throughout the codebase as a *role* at every scale — dialog chrome, page chrome, header chrome — and never as a component name.

**Marked explicitly as reconstructed.** The outcome is well evidenced; the arguments under *Alternatives rejected* are inferred from what the change did, not transcribed. The record says so, and says that anyone who was in that discussion overrides it.

### ADR-0003 — Stacking order belongs to components, not themes

Records DST-1547's reasoning: stacking is a **correctness** property, not an aesthetic one. A toast must be above a modal; a tooltip must be below it. Putting `z-*` in a theme file makes that a per-theme opinion, so a second theme can break overlays without touching component code.

Deliberately does **not** restate the scale — that lives in `CLAUDE.md`, and the store's own rule is that no rule lives in two places. Notes `ui.css`'s `z-100` as the one real exception (the rule targets `*.styles.ts`, not everything under `themes/`).

### ADR-0004 — Slot-driven composition *(was 0001)*

When a container should publish a slot config: only when it injects **semantics or placement** (`level`, `id`/`ref`, `elementType`, `as`, or a `[grid-area:…]`), when the role recurs, and when the alternative is a per-container `<PanelTitle>`/`<DialogTitle>`/`<CardTitle>` differing only in level and styling.

Equally important is when not to — above all **when the region holds arbitrary consumer content.** RAC throws `"A slot prop is required when using slots"` for any unslotted `<Heading>`/`<Text>` nested inside a publishing container, which is why `noSlot` exists. Also captures slot config as a *layout* mechanism, which the public docs under-sell.

### ADR-0005 — Place parts with CSS, not by rearranging React children *(was 0002)*

Default to `grid-template-areas` and `:has()`; touch the React tree only for runtime measurement or collection semantics. `AppShell` is the clearest case — the whole application frame reflows on the presence of a child via `:has([data-rail])`, with no React code observing it.

## Glossary

Ten terms. Added the ticket's set — *app chrome*, *frame*, *RAC-first imports*, *theme component*, *trigger props* — alongside *grid area*, *role primitive*, *slot config* and *slot-driven composition*.

*Compound part* was renamed to **compound component** to match the ticket and `CLAUDE.md`'s own heading, with the part defined inside it.

## The structural guard

`scripts/check-memory-store.mjs` (`pnpm check:memory`, CI: `memory-store.yml`), added after reviewing the store against itself. It does **not** check whether decisions are followed — that stays unenforceable. It closes two holes in the store's own design:

**ADR numbering had no allocation mechanism.** Two branches each adding `0003-<their-own-slug>.md` have different filenames, so git merges both cleanly and `ADR-0003` names two decisions. `.memory/README.md` argued one-decision-per-file makes the store merge-safe — true for contradiction, not for ID collision.

**`applies_to` was decorative.** Nothing read it, yet it looked like it scoped which records load: a false assumption invited, maintenance still charged. It is now documented as an index for finding records by path, and a glob matching zero files fails — so it cannot rot silently past a rename.

Also covers `id`/filename agreement, the status vocabulary, supersede targets that resolve, and duplicate or misordered glossary terms. Follows `check-react-aria-dedupe.mjs`: plain node, no install, own light workflow. Deliberately unfiltered by path, because a glob rots when the code it points at is renamed — a PR that never touches `.memory/`.

It also made the renumbering in this PR verifiable rather than careful.

## What these records are not

**Advisory.** No deterministic gate reads an ADR and blocks a violating change — they influence an agent's context and give a review comment something concrete to cite. All five say so in their own text.

All are `status: accepted`. There is no `proposed` state: deciding happens in review, so a record only reaches `main` by being accepted, and the guard rejects any other value.

## Review focus

The *Alternatives rejected* sections are the ones worth arguing about — they're what stops a settled question being reopened next quarter. ADR-0002's is the weakest by construction, since it is reconstructed rather than transcribed.

Three things review already caught on this branch, worth knowing because they show the failure modes:

- **ADR-0005 contradicted its own Decision section.** The Context claimed none of the five children-inspecting components do visual placement. `Aside` does — `Children.toArray` into `[left, right]`, each in a positioning `<div>`, exactly the wrapper-per-region approach the record rejects. The rule survived (the Decision already allowed it as "an enumerated position is part of the API contract"); the Context was wrong, and now names `Aside` as the counter-example, justified by its tuple-typed `children`.
- **A count drifted within a day.** The `noSlot` figure was right in the ADR and wrong in this PR body, the Jira comment and the discussion — written while writing the rule against exactly that. Bare counts are gone from the records, replaced by the grep that produces them.
- **The delivered content didn't match the ticket.** Caught late, which is the process failure worth naming: the deviation should have been raised before the records were written, not after.

One thing found while writing ADR-0003, **not** fixed here because it is a code change in a docs ticket: `CLAUDE.md` documents `z-20` for "Dropdowns (Select, ComboBox)", but `z-20` appears nowhere in any source and neither component carries a `z-*` at all. `SegmentedControl` uses `z-0`, which isn't in the scale either. Worth a follow-up — either the scale or the components are wrong.

## Test Instructions

1. Run `pnpm check:memory` — it should pass (5 ADRs, 10 terms). To confirm it isn't vacuous, copy an ADR to a duplicate number or point an `applies_to` glob at a path that doesn't exist, and check it fails.
2. Read ADR-0001's per-package table against `packages/system` — the formatter-test exception is the part most likely to be "fixed" by a future PR.
3. Check ADR-0002's reconstructed reasoning against your memory of the naming discussion. If it's wrong, it should be superseded rather than edited.
4. Read ADR-0004's "do not publish a slot config when" list against a container you know well.
5. Check the glossary's "not to be confused with" pairs match how the team actually uses the words.

## Breaking Changes

No. Documentation plus one CI script — nothing under `packages/*/src` or `themes/` is touched, so no changeset.

## Checklist


scripts/check-memory-store.mjs
```diff
@@ -0,0 +1,266 @@
...
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Glossary
+// ---------------------------------------------------------------------------
+
+const glossarySource = readFileSync(resolve(memory, 'CONTEXT.md'), 'utf8');
+// Strip HTML comments first — CONTEXT.md documents its entry shape with a sample `### Term`.
+const glossaryBody = glossarySource.replace(/<!--[\s\S]*?-->/g, '');
```
aromko: 🔴 **CodeQL fails here — 1 new high-severity alert, and it blocks the merge**

Rule `js/incomplete-multi-character-sanitization`. A single pass can leave `<!--` behind on nested input: `<!--<!---->` → removing the inner comment leaves `<!--`.

Not a security problem *here* — local script, committed file, and the output only feeds a `###` regex, never an HTML sink. But it is a red check on a security rule, so the PR cannot merge past it.

Loop until stable, which is the fix CodeQL accepts:

```suggestion
let glossaryBody = glossarySource;
let previousBody;
do {
  previousBody = glossaryBody;
  glossaryBody = previousBody.replace(/<!--[\s\S]*?-->/g, '');
} while (glossaryBody !== previousBody);
```

.memory/adr/0001-rac-first-imports.md
```diff
@@ -0,0 +1,81 @@
...
+**Re-export everything through `@marigold/components` and forbid both doors internally.** Attractive, and already what `docs` does. Rejected as a blanket rule because `packages/components` is where the public surface is defined — it cannot import from itself — and `packages/system` deliberately has no RAC dependency to re-export from.
+
+## Consequences
+
+**What this buys.** One `react-aria` instance in a consumer's tree is the normal outcome rather than a lucky one. Context-splitting bugs — the ones that only appear in a consumer's app, never in ours — stop being reachable through our own imports.
+
+**What it costs.**
+
+- **The rule differs per package**, and the reason is not visible at the import site. Someone moving code from `packages/system` to `packages/components` has to change the import, and nothing says so.
+- **Nothing enforces it.** No lint rule checks the import source. `check-react-aria-dedupe.mjs` catches the duplicate _tree_, not a wrong import path.
```
aromko: ⚠️ **This is no longer true — the guard exists**

"**Nothing enforces it.** No lint rule checks the import source."

`scripts/check-rac-first-imports.mjs` does exactly that. It landed with DST-1729 (#5756) and is on `main`; its own header says *"`check-react-aria-dedupe.mjs` guards the duplicate *tree*; this guards the *import path* that turns such a duplicate into a bug"* and *"this guard exists because four of them were missed and nothing noticed."*

The record was written before that ticket landed — and this PR lists DST-1729 as one of its own blockers, so the fix that closed the blocker is what invalidated the sentence.

Worth keeping the honest half: the guard's scope is a hardcoded glob over `packages/components`, so `docs` and any future RAC-consuming package are still unchecked.

.memory/adr/0003-z-index-scale.md
```diff
@@ -0,0 +1,56 @@
...
+**No convention — set what each situation needs.** What produced the original mess. Rejected for the obvious reason: every value becomes a local negotiation with whatever it currently overlaps, and the result only holds until something new appears between them.
+
+## Consequences
+
+**What this buys.** Stacking behaves identically under every theme. The order is discoverable by reading the component, which is also where the behaviour that depends on it lives. A theme package cannot introduce an overlay bug.
+
+**What it costs.**
+
+- **A rule with a shape people do not expect.** "Styling goes in the theme" is otherwise near-absolute in this codebase; z-index is the exception, and it looks like an oversight until you know why. Expect `z-*` to be "tidied" back into `*.styles.ts`.
+- **Nothing enforces either half.** No check rejects `z-*` in a `*.styles.ts` file, and none verifies that a component picked the right rung. Both are review-time concerns.
```
aromko: ⚠️ **Same as `0001-rac-first-imports.md:79` — the first half is no longer true**

"**Nothing enforces either half.** No check rejects `z-*` in a `*.styles.ts` file"

`scripts/check-theme-zindex.mjs` does, added by DST-1730 (#5757) and on `main`. `CLAUDE.md` already says so: *"`pnpm check:theme-zindex` enforces this in CI — a theme must not be able to reorder the layers."*

The **second** half stands and is worth keeping — nothing verifies a component picked the right *rung*, which is the genuinely unenforceable part. Worth noting the guard's own stated limit too: it reads source text, so a runtime-assembled class (`'z-' + level`) is invisible to it, and it only scans `*.styles.{ts,tsx}`, not other theme `.css`.

scripts/check-memory-store.mjs
```diff
@@ -0,0 +1,266 @@
...
+      `ADR-${number} is claimed by two records: ${byNumber.get(number)} and ${name}. ` +
+        'Two branches most likely took the same next number; renumber the later one to the ' +
+        'next free slot and update its `id`, its `# NNNN.` heading, and any links to it.'
+    );
+  } else {
+    byNumber.set(number, name);
+  }
+
+  // 2. id must agree with the filename.
+  if (front.id !== `ADR-${number}`) {
```
aromko, line 131: ⚠️ **The guard misses the one failure its own error message names**

Line 113 tells the reader a renumber must update "its `id`, its `# NNNN.` heading, and any links to it." This checks the `id`; the citation pass checks the links. **Nothing checks the H1 heading.**

Verified by inducing it — rewriting `# 0003.` to `# 0009.` in `0003-z-index-scale.md` passes clean:

```
✅ .memory store check passed: 5 ADR(s), 10 glossary term(s).
```

This PR renumbers 0001→0004 and 0002→0005, which is precisely that operation. Both headings happen to be correct — the guard just didn't confirm it. One check alongside the `id` one closes it.

scripts/check-memory-store.mjs
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  if (!/^\d{4}-\d{2}-\d{2}$/.test(front.date ?? '')) {
+    errors.push(
+      `${where}: date is \`${front.date ?? '(missing)'}\`; expected YYYY-MM-DD.`
+    );
+  }
+
+  // 4. applies_to must be present and every glob must still match something.
+  const globs = Array.isArray(front.applies_to) ? front.applies_to : [];
```
aromko: 💡 **An inline `applies_to` reports the wrong problem**

`applies_to: 'packages/**'` on one line is valid YAML, but `Array.isArray` is false, so `globs` is `[]` and line 143 fires "applies_to is missing or empty" — for a field that is present and populated.

It fails loudly, so this is a message bug rather than a hole. Coercing a lone string to a one-item list is the smaller fix:

```suggestion
  const raw = front.applies_to;
  const globs = Array.isArray(raw) ? raw : raw ? [raw] : [];
```

scripts/check-memory-store.mjs
```diff
@@ -0,0 +1,266 @@
...
+          'If something was renumbered, its slug moved too.'
+      );
+    } else if (!target.startsWith(`${cited}-`)) {
+      errors.push(
+        `${where}: cites ADR-${cited} but links to \`${target}\`. The number and the file disagree.`
+      );
+    }
+  }
+
+  const bare = body.replace(/\[ADR-\d{4}\]\([^)]+\)/g, '');
```
aromko: 💡 **The bare-`ADR-NNNN` check scans fenced code blocks**

`body` has frontmatter stripped but not fences. No record trips it today, but the first one that illustrates the convention — showing `status: superseded-by ADR-0002` in an example block, say — errors with no way to opt out, and the fix would be to make the documentation worse.

Strip fences the same way frontmatter is stripped, before the bare-mention scan.

aromko (changes requested):
## Automated Code Review

Reviewed against `96b2e9935`. Disclosure: I rebased this branch, so a couple of findings touch my own conflict resolutions.

**Verified against the code** — every factual claim I could check holds:

| Claim | Where |
| --- | --- |
| `noSlot` is a `null as unknown as string \| undefined` cast | `utils/noSlot.ts:27` ✓ |
| `Aside` declares `children: [ReactElement, ReactElement]` | `Aside/Aside.tsx:11` ✓ |
| `AppShell` reflows on `:has([data-rail])` | `AppShell.tsx:67-68` ✓ |
| `Input` clones children to inject placement (DST-1731) | `Input.tsx:36,48` ✓ still true |
| shells on caret, RAC on a single range | `packages/components/package.json` ✓ |

`node scripts/check-memory-store.mjs` passes locally (5 ADRs, 10 terms). The workflow matches `theme-conventions.yml` and `react-aria-dedupe.yml` exactly — same trigger, same concurrency, no install. Glossary is alphabetical, one term per `###`, each with a "not to be confused with".

ADR-0004 and 0005 refusing to enumerate instances, and saying in the record *why* (an earlier draft was wrong about one component and had missed four), is the right call and the best thing in this PR. The guard script is also a better answer to the ADR-numbering hole than the prose fix that shipped in #5753, which it correctly replaces.

---

### One finding with no line to anchor to: the PR description is stale in two places

- **"All are `status: proposed`; they become `accepted` when this merges."** Commit `dd074f1f9` removed the `proposed` state entirely. Every record ships `status: accepted`, and the guard's `STATUS_PATTERN` now *rejects* `proposed`.
- **The blocker callout** still lists #5750 and #5753 (both merged) and DST-1729 / DST-1730 (both on `main`). Only DST-1731 is outstanding — `Input.tsx:36,48` confirms it. A reviewer told to "**Review those first**" follows dead links.

---

### Recommendation: Request Changes

On the CodeQL finding alone, since CI is red on a security rule and nothing merges past it.

The one that matters for the store's credibility is the pair of "nothing enforces this" claims: two of the five records ship saying the opposite of what `main` does, and in both cases the fix *and its guard* landed while the record sat in review. The remaining three are cheap.

---

scripts/check-memory-store.mjs
aromko: Fixed in 2780da229 — the strip now loops until the source stops changing, via a small `stripRepeatedly` helper.

Same helper does the fence stripping for the other finding on this file, so there is one place that gets this right rather than two.

.memory/adr/0001-rac-first-imports.md
aromko: Fixed in 2780da229 — the bullet now names `check-rac-first-imports.mjs` and DST-1729.

Kept the honest half you pointed at, which is the more useful sentence anyway: the guard's scope is a hardcoded glob over `packages/components`, so `docs` and any package that later takes on RAC are unchecked until someone adds them.

.memory/adr/0003-z-index-scale.md
aromko: Fixed in 2780da229 — split into what is enforced and what is not.

`check-theme-zindex.mjs` (DST-1730) covers the first half. The second half stayed, since which rung a component picked is the genuinely unenforceable part. Added the two limits from the guard's own header: it reads source text, so `'z-' + level` is invisible to it, and it scans `*.styles.{ts,tsx}` only — which is also the shape of the `ui.css` exception the record documents just above.

scripts/check-memory-store.mjs
aromko: Fixed in 2780da229 — the `# NNNN.` heading is now checked against the filename, right after the `id` check.

Verified by inducing it. Rewriting `# 0003.` to `# 0009.` used to pass; it now fails with:

```
• .memory/adr/0003-z-index-scale.md: heading says `# 0009.` but the filename says ADR-0003. A renumber has to move the heading too.
```

A missing heading is its own error, since a record with no `# NNNN. Title` has nothing to check.

scripts/check-memory-store.mjs
aromko: Fixed in 2780da229 — a lone string is now read as a one-item list.

Verified: `applies_to: 'packages/components/src/**/*.tsx'` inline now passes instead of reporting a populated field as missing. An absent or empty field still errors as before.

scripts/check-memory-store.mjs
aromko: Fixed in 2780da229 — fenced blocks are stripped before the bare-mention scan, using the same `stripRepeatedly` helper as the CodeQL fix.

Verified both directions: a fenced `status: superseded-by ADR-0002` now passes, and a bare `ADR-0002` in prose still errors.

aromko (approved):

Issue #11, opened 2020-01-15
Setup greenkeper or depandabot


sebald:
Or does Github do this now for us???

Also: https://renovatebot.com/

---> check which is the best tool

PR #5755, opened 2026-08-17
feat(DST-1391): add Stepper, a progress indicator for multi-step tasks

aromko:
# Description

Adds `<Stepper>`, a progress indicator for multi-step tasks: checkouts, onboarding flows, multi-page forms. Several product flows had each built their own "Step 1 of 4" widget, so this replaces them with one accessible component.

It renders a `<nav>` landmark around an ordered list. Each step announces its label, its position, and its state ("Choose plan, Step 2 of 4, current step"), and no state is carried by colour alone.

**State is entirely consumer-owned.** The component never infers that a step is finished, because only your code knows whether validation passed or a save succeeded.

- `completedKeys` is a set, not a high-water mark, so non-contiguous completion coming from a server is expressible.
- `selectableKeys` replaces the default "completed, errored, or current" rule when a backend decides what is reachable. `disabledKeys` always wins over both.
- Errored steps stay clickable on purpose: a user told a step failed needs a way back to it.
- Steps with an `href` render as real links and route through `RouterProvider`; steps without one render as buttons. Unreachable steps render as plain text rather than as disabled controls, since an unreachable step is not a disabled widget.
- `hideLabels` drops labels visually for flows too long to label, keeps them for screen readers, and adds a visible "Step 3 of 5" counter in exchange.

Per-step state is styled via `data-state` on each `<li>` plus Tailwind `in-data-[state=…]` selectors, following the pattern `Sidebar` already established. `useClassNames` only carries `variant`/`size` and the parent renders steps in a loop, so a `cva` variant cannot express it.

Scope was amended before implementation. The [decision log on DST-1391](https://reservix.atlassian.net/browse/DST-1391)
(2026-08-14, approved 2026-08-17) resolves 21 branches against the original spec: decision 6 drops
the Adobe `useStepList*` port, decision 10 cuts `small`/`large` sizes, the `optional` state and
per-item `description` as YAGNI, and decision 19 fixes the theme contract at six slots. Three
acceptance criteria are voided there. Grade this PR against that comment, not against the ticket's
original Scope section, which decision 21 deliberately left intact as history.

Closes DST-1391

## Screenshots / Preview

New component, so there is no "before". Worth looking at in the deploy preview:

| Where | What to look at |
| ----- | --------------- |
| Docs: `/components/navigation/stepper` | Full page: anatomy diagram, states, multi-step form, validation errors |
| Storybook: `Components/Stepper` | `Basic`, `States`, `ManySteps`, `MultiStepForm` (the four snapshotted stories) |

## Test Instructions

1. `pnpm sb` and open **Components → Stepper**.
2. In **MultiStepForm**, press Continue. The panel content should swap, the step just left should turn green, and you should be able to click back to it from the stepper itself.
3. In **ValidationErrors**, press Continue with an invalid email. The step marker should turn red and the field should show its own message. Editing the field should clear the red marker.
4. In **WithHrefs**, the *reachable* steps are real anchors: hover shows a URL, Cmd-click opens a new tab, plain click routes without a page reload. Steps still ahead render as plain text and have no URL, so click forward first to watch one become a link.
5. In **ManySteps**, the labels are hidden and a "Step 5 of 12" counter appears to the left.
6. Tab through any story. The focus ring should appear on keyboard focus only, never on mouse click, and never on steps that are still ahead (those are plain text, not controls).
7. Automated: `pnpm test:unit` and `pnpm test:sb`.

## Breaking Changes

No. New component, additive `Stepper` slot on the `Theme` type.

## Checklist

### Notes on the two unchecked boxes

**VRT not triggered.** Four base stories are snapshotted (`Basic`, `States`, `ManySteps`, `MultiStepForm`); the other four are deliberately opted out, each with a stated reason in the story file. Leaving the Chromatic run to you.

**Screen reader pass outstanding.** The a11y work here was done against WCAG and the APG, and verified with axe plus measured contrast ratios, but it has not been through NVDA/Firefox or VoiceOver/Safari by hand. Worth doing before this leaves BETA.

Two things a reviewer may want to weigh in on:

- `a11y.test` is `'todo'` repo-wide, so axe findings never fail CI for any component. Every axe result quoted during development came from temporarily flipping it to `'error'` locally. That is a theme-team decision, not something I changed here.
- Contrast: the disabled step uses `text-secondary` (4.99:1) rather than `text-disabled` (2.09:1), because an unreachable step renders as plain text and so does not get WCAG 1.4.3's exemption for inactive controls. The completed connector uses `success-accent` (2.90:1) rather than `success` (green-100, 1.01:1 as a hairline); it is `aria-hidden` and decorative, so it needs to be perceivable rather than to clear 1.4.11 on its own.


packages/components/src/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,585 @@
+import type { Key } from 'react';
```
jim761, line 1: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files (component source, theme, stories) but the `Visual-Regression-Tests` workflow has not run for this branch — `gh run list --workflow=visual-regression-tests.yml --branch=feat/DST-1391-stepper` returns `[]`.

packages/components/src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,267 @@
...
+  );
+
+  const completed = new Set(completedKeys);
+  const errors = new Set(errorKeys);
+  const disabled = new Set(disabledKeys);
+  const selectable = selectableKeys ? new Set(selectableKeys) : undefined;
+
+  const getState = (key: Key): StepState => {
+    if (disabled.has(key)) return 'disabled';
+    if (errors.has(key)) return 'error';
```
jim761, line 178: 🟠 A step that is **current *and* errored** loses every "you are here" signal.

`state` is a single value, and it drives both `data-state` (:222) and the hidden `stateText` (:219). This is the normal state in `ValidationErrors`: the marker drops `bg-primary`, the label loses `font-medium`, and the announcement reads "Step 1 of 3, error" — no mention of current. Only `aria-current="step"` survives, and for a non-selectable step that lands on a bare `<span>` (:112) where SR support is weakest. Same collapse for `disabled` + selected.

Suggest emitting currency as its own axis: `data-current={id === selectedKey || undefined}` on the `<li>`, an `in-data-current:` rule in the theme, and appending `stepCurrent` to `stateText` when the id is selected regardless of `state`.

jim761 (commented):
## Automated Code Review

Reviewed with three parallel agents (component/TypeScript/theming, accessibility/testing, docs/spec conformance). Findings below were verified against the source on `71de373`; all 39 unit + 16 story tests pass locally.

### Overview

| Field | Value |
|-------|-------|
| Branch | `feat/DST-1391-stepper` → `main` |
| Files Changed | 22 (all new) |
| Additions / Deletions | +1895 / −0 |
| CI | All 27 checks passing |

### Linked ticket — DST-1391

The ticket's original description is superseded by the decision-log comment (2026-08-14) that rewrote the design and voided three ACs, approved 2026-08-17. Graded against the amended spec, **every acceptance criterion is met** except the two this PR itself flags as open (axe-clean CI, manual NVDA/VoiceOver pass).

Two things shipped beyond spec, both defensible: a seventh theme slot `count` (decision 19 fixed the contract at six slots) and the `v18.ts` codemod manifest entry. The `count` slot is a theme-contract change the decision log does not record.

### Issues found

#### 🔴 Critical

**1. Visual regression tests never run.** `gh run list --workflow=visual-regression-tests.yml --branch=feat/DST-1391-stepper` returns `[]`. This PR adds a component, 585 lines of stories and a 112-line theme file. Trigger with `/run-chromatic`.

#### 🟠 Warnings

**2. `Stepper.tsx:174` — a step that is current *and* errored loses every "you are here" signal.** `state` is a single value driving both `data-state` and the hidden state text. Reachable from the shipped `ValidationErrors` story, not a corner case.

**3. `Stepper.tsx:53` — the `errorKeys` doc comment ("Always selectable") states a guarantee the code does not make** when `selectableKeys` is passed. Repeated in `.changeset/stepper.md:11` and `index.mdx:87,103`. Untested combination.

**4. `Stepper.styles.ts:15` — `list-none` strips list semantics in Safari/VoiceOver (WCAG 1.3.1).** The existing test cannot catch it: Testing Library ignores CSS and the browser project runs Firefox.

**5. `stepper-basic.demo.tsx:9` — the page's first interactive demo dead-ends after one click.** `stepper-links.demo.tsx:22` has the identical bug.

**6. `index.mdx:131` — "an `href` renders a real link" is false for unreachable steps,** and the demo directly below it disproves the sentence.

**7. `Stepper.tsx:155` — the child filter accepts any element, not just `Stepper.Item`.** Fragments and stray elements become steps with `id === undefined`.

**8. `Stepper.tsx:138` — no `ref` exposed.** **9. `Stepper.tsx:197` — the root is styled outside the theme.** **10. `Stepper.tsx:12` — type-only import as a value import.** **11. `index.mdx:4` — `badge: BETA` is the only uppercase badge in the docs** (leaks into `llms.txt`).

Each of the above is an inline comment on the relevant line.

#### 💡 Suggestions

- **`hideLabels` vs the "No mode booleans" rule** — judged *not* a genuine violation: the label stays in the a11y tree (appearance, not structure), and the counter is compensation the component owes for its own visual choice, so it cannot be a compound part without making a11y correctness the consumer's job. `variant="compact"` would be worse. If you want strict compliance, `labels?: 'visible' | 'hidden'`.
- **Test gaps**, all cheap to close: `defaultSelectedKey` is entirely unexercised; the uncontrolled write path is never asserted (every click test pins `selectedKey`); non-contiguous `completedKeys` — the headline promise of the prop — has no test; `errorKeys` + `selectableKeys` untested; the `stepDisabled` announcement is only checked as a raw string, never rendered.
- `Stepper.tsx:194` — an empty `<nav>` landmark ships when `children` is empty; `if (items.length === 0) return null`.
- `Stepper.tsx:161` — the uncontrolled default is captured on first render, so async-arriving steps never acquire a current step. Worth a line on the JSDoc.
- `Stepper.tsx:82` — accept `aria-labelledby` so a page with a visible "Checkout progress" heading need not translate the string twice.
- `Stepper.tsx:247` — the `", "` separator and the ordering are hardcoded where a translator cannot reach them. Also: accname joins text nodes with a space, so the real accessible name is `"Choose plan Step 2 of 4, current step"` — `index.mdx:151` advertises a comma.
- `intl/messages.test.ts` — a one-line locale key-parity assertion would stop the *next* component adding to one locale only.
- `.changeset/stepper.md` — `@marigold/cli` is missing despite `v18.ts` changing (the ListView PR omitted it too, so possibly deliberate).
- `packages/cli/src/lib/codemod/manifests/v18.ts:152` — `Stepper:` breaks the block's alphabetical order.
- `Stepper.tsx:76` — `variant` is the only prop without a doc comment, and there is no `size` though every sibling has one.
- Docs polish: `index.mdx:19` (the last step has no connector), `:114-127` (snippet lacks `onSelectionChange`), `:112` (does not say `selectableKeys` revokes the errored-step guarantee), `:9` (omits `disabled` from the state list), `:58-79` (a 3-item Do list against a one-sentence Don't renders unbalanced).

### Done well

Element choice is genuinely reasoned and the reasoning is *in the code*: real `<a>`, real `<button>`, plain text rather than a fake disabled control for unreachable steps. State is never colour-alone (check / triangle / dashed ring / fill plus `font-medium`), backed by hidden state text. The `hideLabels` counter is `aria-hidden` *because* every step already announces its own position — the detail most stepper implementations get wrong — and it is tested three ways. Label-first ordering satisfies WCAG 2.5.3 and is pinned with an anchored regex rather than a loose match. Focus rings are tested against real computed styles in a real browser. Z-index rules respected, RAC-first deep imports, theme addition additive and mirrored in `v18.ts`, `data-state` + `in-data-[state=…]` matching the `Sidebar` precedent. The anatomy diagram conforms exactly to the DST-1703 inline-diagram convention. `Stepper.styles.ts` records measurements and rejected alternatives, which is precisely what stops the next person "fixing" it back.

### Recommendation

Posted as a **non-blocking comment**. Two items are worth resolving before merge — the missing Chromatic run (#1) and the current+error state collapse (#2), which is reachable from a shipped story. #3–#6 are small, concrete fixes; everything else is optional.

The code quality here is well above average for the repo. The length of this list is density of review, not density of defect.

---

packages/components/src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,267 @@
...
+  onSelectionChange?: (key: Key) => void;
+
+  /**
+   * The steps that are done. A set, not a high-water mark, so non-contiguous
+   * completion is expressible. Never inferred by the component.
+   */
+  completedKeys?: Iterable<Key>;
+
+  /**
+   * The steps that failed. Always selectable, so the user can go back and fix
```
jim761: 🟠 This guarantee isn't one the code makes. `isSelectable` (:187) short-circuits — `if (selectable) return selectable.has(key);` — so an errored key absent from `selectableKeys` renders as inert text: the user is told a step is broken with no way back to it. `.changeset/stepper.md:11` and `index.mdx:87,103` repeat the absolute claim.

Either honour it (`return selectable.has(key) || state === 'error'`) or reword to the implemented precedence. Untested either way.

```suggestion
  /**
   * The steps that failed. Selectable by default, so the user can go back and
   * fix them — unless `selectableKeys` is given, which replaces that rule.
   */
```

themes/theme-rui/src/components/Stepper.styles.ts
```diff
@@ -0,0 +1,112 @@
...
+   * The `gap` matters: the connector is the last child of its own `<li>`, so
+   * without a gap between list items it butts straight into the next marker
+   * while still sitting 8px clear of its own step. The line has to breathe on
+   * both sides or it reads as an arm growing out of the next circle.
+   */
+  container: cva({
+    // `grow` because the list is itself a flex child of the `<nav>` (which may
+    // also hold the step counter): without it the list sizes to its content and
+    // the steps have no surplus left to share.
+    base: 'm-0 flex grow list-none items-start gap-2 p-0',
```
jim761, line 6: 🟠 `list-none` strips list semantics in Safari/VoiceOver (WCAG 1.3.1 Info and Relationships). WebKit drops the `list`/`listitem` roles when `list-style: none` is set without an explicit role. This is the only `list-none` in the theme package, and `index.mdx` promises the opposite ("renders a `<nav>` landmark around an ordered list"). The unit test at `Stepper.test.tsx:13` can't catch it — Testing Library ignores CSS and the browser project runs Firefox.

Best fixed in the component so a theme override can never remove it again — `Stepper.tsx:211`: `<ol role="list" className={classNames.container}>`.

packages/components/src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,267 @@
...
+  selectableKeys,
+  hideLabels,
+  variant,
+  'aria-label': ariaLabel,
+}: StepperProps) => {
+  const stringFormatter = useLocalizedStringFormatter(intlMessages);
+  const classNames = useClassNames({ component: 'Stepper', variant });
+
+  const items = Children.toArray(children).filter(
+    isValidElement<StepperItemProps>
```
jim761: 🟠 This filter accepts any element, not just `Stepper.Item`. `isValidElement<T>` only asserts the props type — it doesn't check element type — and `Children.toArray` doesn't flatten fragments. So `<Stepper><>{a}{b}</></Stepper>` becomes one step with `id === undefined`: React key warning, `upcoming` state, fragment children rendered as the label.

```suggestion
  const items = Children.toArray(children).filter(
    (child): child is ReactElement<StepperItemProps> =>
      isValidElement(child) && child.type === StepperItem
  );
```

packages/components/src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  return (
+    <Button className={className} aria-current={ariaCurrent} onPress={onSelect}>
+      {children}
+    </Button>
+  );
+};
+
+const _Stepper = ({
```
jim761, line 143: 🟠 No `ref` exposed, so consumers can't reach the `<nav>`. The repo review checklist grades a missing ref *Critical*; repo practice is React 19 ref-as-prop (`Breadcrumbs.tsx:45-53` does it, `Tabs`/`Pagination` don't). A stepper is a plausible scroll-into-view / measurement target.

packages/components/src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,267 @@
...
+    if (selectable) return selectable.has(key);
+    return state !== 'upcoming';
+  };
+
+  const currentIndex = items.findIndex(item => item.props.id === selectedKey);
+
+  return (
+    <nav
+      aria-label={ariaLabel ?? stringFormatter.format('progress')}
+      className="flex items-center gap-3"
```
jim761, line 198: 🟠 The root is styled outside the theme, so the counter/list gap and the root layout aren't themeable while everything else in the component is. `Pagination.tsx:171` threads its theme slot through the root — either add a `root` slot to `Theme['components'].Stepper` (+ `v18.ts`), or compose with `cn()`.

packages/components/src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,267 @@
...
+import { Button } from 'react-aria-components/Button';
+import { Link } from 'react-aria-components/Link';
+import { VisuallyHidden } from 'react-aria-components/VisuallyHidden';
+import { useLocalizedStringFormatter } from '@react-aria/i18n';
+import { useControlledState } from '@react-stately/utils';
+import { useClassNames } from '@marigold/system';
+import { Check } from '../icons/Check';
+import { TriangleAlert } from '../icons/TriangleAlert';
+import { intlMessages } from '../intl/messages';
+import { StepperItem, StepperItemProps } from './StepperItem';
```
jim761: 💡 Type-only import brought in as a value import.

```suggestion
import { StepperItem, type StepperItemProps } from './StepperItem';
```

docs/content/components/navigation/stepper/stepper-basic.demo.tsx
```diff
@@ -0,0 +1,24 @@
+import type { Key } from 'react';
+import { useState } from 'react';
+import { Stepper } from '@marigold/components';
+
+const STEPS = ['signin', 'plan', 'pay', 'done'];
+
+export default () => {
+  const [selectedKey, setSelectedKey] = useState<Key>('plan');
+  const completedKeys = STEPS.slice(0, STEPS.indexOf(selectedKey as string));
```
jim761: 🟠 The page's first interactive demo dead-ends after one click. `completedKeys` is derived from the current index, so clicking back to `signin` empties it — `plan`, `pay` and `done` all become `upcoming`, render as `<span>`s, and there's no other control to move forward. That contradicts the Do guideline at `index.mdx:69` ("Let users return to steps they have already finished").

Hold completion in its own state that only grows, as `stepper-multi-step-form.demo.tsx:50-58` already does correctly.

docs/content/components/navigation/stepper/stepper-links.demo.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const [path, setPath] = useState('/checkout/plan');
+  const index = ROUTES.findIndex(route => route.href === path);
+
+  return (
+    <RouterProvider navigate={setPath}>
+      <Stack space={4}>
+        <Stepper
+          aria-label="Checkout progress"
+          selectedKey={ROUTES[index].id}
+          completedKeys={ROUTES.slice(0, index).map(route => route.id)}
```
jim761: 🟠 Same dead end as `stepper-basic.demo.tsx:9` — clicking "Sign in" empties completion, so "Choose plan" and "Pay" stop being anchors and the routing demo can't be re-driven. Keep a `visited` set alongside `path` and pass that as `completedKeys`.

docs/content/components/navigation/stepper/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+    <Stepper.Item key={step.id} id={step.id}>
+      {step.label}
+    </Stepper.Item>
+  ))}
+</Stepper>
+```
+
+### Links or buttons
+
+When each step of your flow has its own URL, the steps should behave like the links they are, so middle-click, open in a new tab, and copy link all work. Give a `<Stepper.Item>` an `href` and it renders as a real link. Without an `href`, the step renders as a button and only calls `onSelectionChange`.
```
jim761: 🟠 Not true for unreachable steps, and the demo directly below disproves it: selectability is checked *before* `href` (`Stepper.tsx:110-116`), so "Pay" in `stepper-links.demo.tsx` carries an `href` and still renders as a `<span>`. PR test instruction 4 ("hover shows a URL, Cmd-click opens a new tab") doesn't hold for it. Same imprecision at `:25-26`.

Suggest: "Give a **reachable** `<Stepper.Item>` an `href` and it renders as a real link; a step that is still ahead stays inert text even with an `href`."

docs/content/components/navigation/stepper/index.mdx
```diff
@@ -0,0 +1,180 @@
+---
+title: Stepper
+description: A component for showing a user's position in a multi-step task.
+badge: BETA
```
jim761: 💡 Only uppercase badge in the docs — every sibling uses lowercase (`beta` on `sidebar`, `topnavigation`, `listview`, …). The sidebar chip is CSS-uppercased so it looks identical, but `docs/scripts/build-llms.ts:130` interpolates the raw value, so `llms.txt` will read `(BETA)`.

```suggestion
badge: beta
```

packages/components/src/Stepper/Stepper.stories.tsx
aromko: I have it in mind, after it gets near the merge.

packages/components/src/Stepper/Stepper.tsx
aromko: Half right, and the half that's wrong matters.

`isCurrent` is passed independently of `state` (`isCurrent={id === selectedKey}`), so `aria-current="step"` **is** on the rendered `<button>` in the `ValidationErrors` story. SR users do hear "current step" there. The bare-`<span>` case you describe only fires when `selectableKeys` excludes the errored step — at which point you have already opted out of the errored-step guarantee deliberately.

What genuinely collapses is the visual: no `bg-primary`, no `font-medium`.

Fixed by emitting currency as its own axis, as you suggested: `data-current` on the `<li>`, and the label's weight rule now keys off `in-data-current:` instead of `in-data-[state=current]:`. Same output for a plain current step, so no Chromatic diff; the weight now survives `error` and `disabled`.

Skipped the `stateText` half on purpose — appending `stepCurrent` would double-announce against `aria-current` in the common case.

What the *marker* should look like when current and error coincide is a design question, not a code one: the `marker` slot deliberately keeps error as a ring so red cannot out-shout the current step's fill. `data-current` is now there for whatever design picks.

packages/components/src/Stepper/Stepper.tsx
aromko: Confirmed — `isSelectable` short-circuits on `selectable.has(key)`, so the guarantee is not one the code makes.

Reworded rather than changed the behaviour: `selectableKeys` "replaces the default rule entirely" is the documented and coherent design, and making `errorKeys` punch through it would give `disabledKeys` a second thing to lose to.

Updated in all three places you named, plus the `selectableKeys` bullet at `index.mdx:112`, which said nothing about revoking the guarantee.

Now tested: `drops the errored-step guarantee when selectableKeys leaves it out`.

themes/theme-rui/src/components/Stepper.styles.ts
aromko: Fix is right, diagnosis isn't.

Tailwind v4's preflight already ships `ol, ul, menu { list-style: none }` and `themes/theme-rui/src/styles.css:18` imports it. So deleting `list-none` from `Stepper.styles.ts` would change nothing, and *every* list in the design system loses its role in Safari — Breadcrumbs included. "The only `list-none` in the theme package" points at the wrong cause.

Added `role="list"` on the `<ol>` in the component, per your reasoning about theme overrides. Kept `list-none` in the theme: explicit beats inheriting it from preflight, and `list-style: none` + `role="list"` is the correct pairing.

The repo-wide gap is worth its own ticket rather than a drive-by here.

packages/components/src/Stepper/Stepper.tsx
aromko: Correct — `isValidElement<T>` only asserts the props type, so the predicate was lying. Applied your suggestion.

Now tested: `ignores children that are not a Stepper.Item`.

packages/components/src/Stepper/Stepper.tsx
aromko: Skipping this one.

`Tabs` and `Pagination` don't expose a ref either, so there is no settled convention to violate — and "the review checklist grades it Critical" is the checklist grading itself.

"A plausible scroll-into-view target" is speculation, and adding a ref later is non-breaking, so there is no cost to waiting for a consumer who actually wants one.

packages/components/src/Stepper/Stepper.tsx
aromko: Not taking this.

`flex items-center gap-3` on the `<nav>` exists for exactly one job: placing the counter next to the list. Your own review flags that the `count` slot already put the component past the six-slot contract decision 19 fixed — adding an eighth slot to make a three-token flex row themeable makes that worse, not better.

If the root has to be themeable, that is a decision-log amendment with the `count` question settled alongside it, not a change smuggled in on this PR.

packages/components/src/Stepper/Stepper.tsx
aromko: Applied.

docs/content/components/navigation/stepper/stepper-basic.demo.tsx
aromko: Confirmed and fixed. Completion now lives in its own state that only grows, and `onSelectionChange` adds the step being left before it moves the selection — so walking back to "Sign in" keeps "Choose plan" and "Pay" clickable and the demo can be re-driven.

docs/content/components/navigation/stepper/stepper-links.demo.tsx
aromko: Same fix. A `visited` set now sits alongside `path` and feeds `completedKeys`, so the steps stay real anchors after walking back.

Worth noting the route alone could never carry this: it says where you are, not where you have been.

docs/content/components/navigation/stepper/index.mdx
aromko: Confirmed — `isSelectable` is checked before `href` (`Stepper.tsx:110-116`), so an unreachable step with an `href` renders as a `<span>`.

Took your wording. Also fixed test instruction 4 in the PR description, which made the same claim.

docs/content/components/navigation/stepper/index.mdx
aromko: Applied. The `llms.txt` interpolation is the part that made this worth changing rather than leaving as a cosmetic inconsistency.

aromko:
Worked through the review on `655670585`. Each inline thread has its own reply; this is the summary, including the three findings that did not survive checking.

### Applied

| Finding | What changed |
| --- | --- |
| #2 current + error | `data-current` on the `<li>`, independent of `data-state`. The label's weight rule keys off currency now, so it survives `error` and `disabled` |
| #3 `errorKeys` guarantee | Doc reworded to the implemented precedence in all four places, behaviour unchanged |
| #4 `list-none` | `role="list"` on the `<ol>` |
| #5 child filter | `child.type === StepperItem` |
| #6 `href` claim | Scoped to reachable steps, in the docs and in the PR description |
| #7 demo dead-ends | Both demos hold a set that only grows |
| #10 type-only import | Applied |
| #11 `badge: BETA` | Lowercase |
| changeset | `@marigold/cli: patch` added for the `v18.ts` manifest entry |
| test gaps | `errorKeys` + `selectableKeys`, non-contiguous `completedKeys`, `defaultSelectedKey`, the uncontrolled write path, plus the empty-stepper and child-filter guards. 21 -> 29 unit tests |

Two extras from the suggestion list: an empty `<Stepper>` no longer ships an empty landmark, and `index.mdx:151` now shows the space-joined accessible name, since accname joins text nodes with a space rather than the comma it advertised.

### Not applied

**#8 missing `ref`.** `Tabs` and `Pagination` don't expose one either, so there is no convention being violated, and "the repo review checklist grades it Critical" is the checklist grading itself. A ref is non-breaking to add later, so there is no cost to waiting for a consumer who wants one.

**#9 root theme slot.** The root's `flex items-center gap-3` exists only to place the counter. This review also flags that the `count` slot already pushed the component past the six-slot contract decision 19 fixed — adding an eighth slot for a three-token flex row makes that worse. If the root must be themeable, that is a decision-log amendment, not a drive-by.

**`v18.ts:152` alphabetical order — this one is wrong.** The block is not alphabetical: `Select` sits after `Slider`, and `NumberField` after `SelectList`. `Stepper` after `Slider` is in order.

### Corrections to the review's reasoning

Two findings were right about the fix and wrong about the cause, which matters for anyone reading the thread later.

**#4** is not a Stepper defect. Tailwind v4's preflight sets `list-style: none` on every list and `themes/theme-rui/src/styles.css:18` imports it, so removing `list-none` from `Stepper.styles.ts` would change nothing, and every list in the design system has this in Safari — Breadcrumbs included. Filing it separately.

**#2** overstates the a11y half. `isCurrent` is passed independently of `state`, so `aria-current="step"` is on the rendered `<button>` in the shipped `ValidationErrors` story; SR users do hear "current step". The bare-`<span>` path only fires when `selectableKeys` excludes the errored step, which is an explicit opt-out. What actually collapsed was the visual, which is what the fix addresses. I skipped the `stateText` half deliberately: appending `stepCurrent` would double-announce against `aria-current`.

### Still open

- **VRT.** Mine to trigger, near merge.
- **The current+error marker.** The label weight is restored; whether the marker changes shape or fill is a design call. The `marker` slot deliberately keeps error as a ring so red cannot out-shout the current step's fill, so I am not inventing a treatment here. `data-current` is in place for whatever design picks. cc @sarahgm
- **Manual screen reader pass**, as the description says.

Verified: `pnpm typecheck:only` clean, `pnpm lint` 0 errors, 29 unit + 16 story tests pass.

packages/components/src/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,610 @@
...
+  // No snapshot: this demonstrates behaviour, and its appearance is a 3-step
+  // subset of Basic. Same call the repo makes for every other `Controlled` story.
+  parameters: { chromatic: { disableSnapshot: true } },
+  args: {
+    'aria-label': 'Onboarding progress',
+  },
+  render: function Render(args) {
+    const steps = ['account', 'profile', 'invite'];
+    const [selectedKey, setSelectedKey] = useState<Key>('profile');
+    const completedKeys = steps.slice(0, steps.indexOf(selectedKey as string));
```
jim761: 🟠 `Controlled` walks into a dead end, and its own test performs the click that causes it.

`completedKeys` is derived from the current index, so clicking "Account" sets `selectedKey='account'` and `slice(0, 0)` empties completion — "Profile" and "Invite team" resolve to `upcoming`, render as `<span>`s, and the story has no other control to move forward. This is the finding that was fixed in `stepper-basic.demo.tsx` and `stepper-links.demo.tsx`; the stories teaching the same thing weren't touched.

Unlike a docs demo, this one can't be made static — the story is named for demonstrating flow-owned selection. Completion has to grow:

```tsx
const [completedKeys, setCompletedKeys] = useState<Key[]>(['account']);

const select = (key: Key) => {
  setCompletedKeys(keys =>
    keys.includes(selectedKey) ? keys : [...keys, selectedKey]
  );
  setSelectedKey(key);
};
```

Note it records the step being *left* (`selectedKey`), not the one being entered — recording the entered key is a no-op, since a step has to be selectable before it can be clicked.

packages/components/src/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,610 @@
...
+    const [path, setPath] = useState('/checkout/plan');
+    const index = CHECKOUT_ROUTES.findIndex(route => route.href === path);
+
+    return (
+      <RouterProvider navigate={setPath}>
+        <Stack space={4}>
+          <Stepper
+            {...args}
+            selectedKey={CHECKOUT_ROUTES[index].id}
+            completedKeys={CHECKOUT_ROUTES.slice(0, index).map(
```
jim761: 🟠 Same derive-from-index pattern as `Controlled` above (`:142`), and here it also falsifies test instruction 4 in the PR description.

`CHECKOUT_ROUTES.slice(0, index)` means that at the initial route `/checkout/plan`, "Pay" is `upcoming` — so despite carrying an `href` it renders as a `<span>`. "hover shows a URL, Cmd-click opens a new tab, plain click routes without a page reload" doesn't hold for it. The `index.mdx` sentence was rescoped to reachable steps last round; the PR body still reads unchanged.

Clicking "Sign in" also empties completion, leaving "Sign in" the only reachable step. `stepper-links.demo.tsx:17-23` already has the shape this needs — a `visited` set alongside the route:

```tsx
const [visited, setVisited] = useState(['/checkout/signin']);

const navigate = (href: string) => {
  setVisited(paths => (paths.includes(path) ? paths : [...paths, path]));
  setPath(href);
};
```

packages/components/src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,280 @@
...
+  selectableKeys?: Iterable<Key>;
+
+  /**
+   * Renders markers only, keeping the labels for screen readers. Use it when
+   * there are too many steps for labels to fit.
+   * @default false
+   */
+  hideLabels?: boolean;
+
+  variant?: 'default' | (string & {});
```
jim761, line 81: 🟠 Three items from DST-1391's *Scope → Included* shipped as nothing, with no note anywhere in the PR:

- **Sizes** `small` / `default` / `large`. There is no `size` prop and the theme defines no size variants. It defines no `variant` values either, so `<AppearanceTable component="Stepper" />` on the docs page renders `variant: -` and `size: -` while this prop advertises `'default'`.
- **`optional` state + `optionalKeys`**, including its place in the documented priority order (`disabled > error > current > completed > optional > upcoming`).
- **`Stepper.Item description`** and the theme's `description` slot.

The other deviations from the ticket are deliberate and argued in the changeset — `completedKeys` as a set over `lastCompletedKey`, real `<a>`/`<button>`/text over `role="link"` + `aria-disabled`, no ported Adobe hooks. These three are simply absent. Descope them on the ticket, or add a line to the PR body saying so; as it stands a reviewer reading the ticket can't tell which are decisions and which are gaps.

packages/components/src/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,610 @@
...
+  }
+);
+
+// An unreachable step renders as a <span>, so a pointer there would promise a
+// click that does nothing.
+Basic.test(
+  'shows no pointer cursor on a step that is still ahead',
+  { parameters: { chromatic: { disableSnapshot: true } } },
+  async ({ canvas }) => {
+    const upcoming = canvas.getByText('Pay').closest('span');
```
jim761: 💡 `closest('span')` resolves to the label span itself, not the step wrapper carrying the `link` slot class — so this asserts the cursor of the element *inside* the one under test. It passes only because `cursor` inherits.

```suggestion
    const upcoming = canvas.getByText('Pay').parentElement;
```

packages/components/src/Stepper/Stepper.test.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+// Tailwind's preflight sets `list-style: none` on every list, and WebKit reads
+// that as "not a list". The explicit role is what keeps the semantics.
+test('keeps the list role against the stylesheet', () => {
+  render(<Basic.Component />);
+
+  expect(screen.getByRole('list')).toHaveAttribute('role', 'list');
+});
+
+test('ignores children that are not a Stepper.Item', () => {
```
jim761, line 266: 💡 These two tests hand-build the tree with `MarigoldProvider`, against CLAUDE.md's "import stories instead of creating fixtures". Justified — no story expresses a non-`Stepper.Item` child or an empty stepper — but a `NotAStep` / `Empty` story with `chromatic: { disableSnapshot: true }` would keep the rule intact and put both behaviours in Storybook where they're discoverable.

jim761 (commented):
## Automated Code Review

Second pass on this branch: checking the fixes from the last round, then re-running the checklist against `fe76ae4b1`.

### Previous round

Ten of twelve comments verified closed in the code; two declined with reasons I'm not reopening (`ref`, root theme slot — both non-breaking to add later).

Spot-checks worth recording:

- `data-current` + `in-data-current:font-medium` is live CSS, not a dead variant — compiling it yields `:where([data-current]) .in-data-current\:font-medium`.
- Your correction on `list-none` is right: preflight ships `list-style: none`, so my stated cause was wrong. `role="list"` on the `<ol>` is the fix either way, and it's there with a test.
- `stepper-links.demo.tsx` behaves: after walking back to "Sign in", "Choose plan" is still a real anchor.

### This round

All 27 checks green. No `any`, `import type` throughout, no `className`/`style` on the props, `useClassNames` on every visible element, no effects at all, Vitest + `userEvent` + accessible queries, `component-test` on every tested story, no z-index in a `*.styles.ts`. Applying `vercel:react-best-practices` surfaced nothing worth raising — state is derived during render, `StepContent` is module-level, key lookups already use `Set`.

Three findings and two suggestions, inline. The first two are the same defect the docs demos were fixed for, in the two stories the test instructions point a reviewer at.

---

packages/components/src/Stepper/Stepper.stories.tsx
aromko: Applied in 0f8d928. Completion is state now and only grows, recording the step being left rather than the one being entered, as you noted: recording the entered key is a no-op since a step has to be selectable before it can be activated.

Added `keeps the step ahead reachable after walking back` — clicks "Account", asserts "Profile" is still a button. The old shape passed its own test precisely because the test never looked past the click.

packages/components/src/Stepper/Stepper.stories.tsx
aromko: Split verdict: the dead-end half is fixed in code, the PR-description half was a real miss and is fixed in the description.

`WithHrefs` now keeps a `visited` set alongside the route, in the shape you quoted, plus a regression test that routes to "Sign in" and asserts "Choose plan" is still a link. 0f8d928.

The second half stands and I had missed it. At the initial route `/checkout/plan` the visited set holds only `/checkout/signin`, so "Pay" resolves to `upcoming` and renders as a `<span>` — test instruction 4 promised a URL on hover for every step, which was never true for it. Reworded to scope the claim to reachable steps and to say the steps ahead are plain text, matching what `index.mdx` was rescoped to last round.

packages/components/src/Stepper/Stepper.tsx
aromko: These three are approved cuts, not gaps.

The decision-log comment on DST-1391 (2026-08-14), decision 10, is titled "YAGNI cuts" and reads: "`small`/`large` sizes, the `optional` state and `optionalKeys`, per-item `description`, and `useNumberFormatter` are all cut." Decision 19 fixes the theme contract at six slots with no `description`. The "Acceptance criteria superseded" section voids "All six states render correctly across the three sizes" to "five states, one size, per decision 10". Sebastian approved the log on 2026-08-17.

Your first-round review cited that same document in its opening paragraph and graded the PR against it — "the ticket's original description is superseded by the decision-log comment (2026-08-14) ... graded against the amended spec, every acceptance criterion is met". This round grades against the Scope section that comment supersedes.

On the `AppearanceTable` row: `variant: -` is the norm for navigation components rather than a Stepper defect. `Tabs`, `TopNavigation` and `Sidebar` each expose a `variant` prop against a theme that defines no variants block; `Pagination` defines none either.

The part I am taking: nothing in the PR itself pointed at the decision log, which is how this got raised twice. Added a line to the description pointing at it. No code change.

packages/components/src/Stepper/Stepper.stories.tsx
aromko: Applied in 0f8d928. `closest()` starts at the element itself, so it matched the label span and the assertion only passed through inheritance. Now `.parentElement`, with a comment recording why.

packages/components/src/Stepper/Stepper.test.tsx
aromko: Not taking this one.

The rule is "Don't create test fixtures/themes - Import stories instead of using `setup()` or creating inline themes". Both tests import the real `theme` from `@marigold/theme-rui`: no fixture, no inline theme, no `setup()`. Nine component test files in the repo do the same, `Drawer` / `ListView` / `Sidebar` / `SidebarRail` / `SelectList` / `Panel` among them.

The cost is two catalogue entries whose only job is hosting a guard test. A `NotAStep` story renders a stray `<span>` inside a `Stepper` to document an internal child filter, and Storybook is where people go to copy patterns from, so "discoverable" works against it here.

aromko:
> [!NOTE]
>  VRT still missing needs a second reviewer before.

jim761 (approved):

packages/components/src/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,660 @@
...
+  ),
+});
+
+/**
+ * The step is owned by the flow. Marigold never advances it, and never marks a
+ * step complete on its own.
+ */
+export const Controlled = meta.story({
+  tags: ['component-test'],
+  // No snapshot: this demonstrates behaviour, and its appearance is a 3-step
```
sarahgm: unneeded comment

packages/components/src/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,660 @@
...
+  { id: 'delivery', label: 'Delivery' },
+  { id: 'payment', label: 'Payment' },
+  { id: 'texts', label: 'Texts' },
+  { id: 'images', label: 'Images' },
+  { id: 'preview', label: 'Preview' },
+  { id: 'publish', label: 'Publish' },
+];
+
+/**
+ * The case `hideLabels` exists for: twelve steps, where no amount of horizontal
```
sarahgm, line 202: comment can also be shortened

themes/theme-rui/src/components/Stepper.styles.ts
```diff
@@ -0,0 +1,120 @@
+import type { ThemeComponent } from '@marigold/system';
+import { cva } from '@marigold/system';
+
+export const Stepper: ThemeComponent<'Stepper'> = {
+  /**
+   * The `gap` matters: the connector is the last child of its own `<li>`, so
```
sarahgm: The style file has a looooot of long comments, you can also shorten them or remove

packages/system/src/types/theme.ts
```diff
@@ -209,6 +209,21 @@ export type Theme = {
       'container' | 'track' | 'thumb' | 'output' | 'selectedTrack',
       ComponentStyleFunction<string, string>
     >;
+    // Per-step state (`completed` / `current` / `error` / `disabled` /
```
sarahgm: not neeeded comment

docs/content/components/navigation/stepper/stepper-basic.demo.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const [selectedKey, setSelectedKey] = useState<Key>('plan');
+  // Completion only ever grows. Deriving it from the current step would undo it
+  // the moment the user walks back, and every later step would stop being
+  // clickable, so there would be no way forward again.
+  const [completedKeys, setCompletedKeys] = useState<Key[]>(['signin']);
+
+  const select = (key: Key) => {
+    setCompletedKeys(keys => (keys.includes(key) ? keys : [...keys, key]));
+    setSelectedKey(key);
+  };
```
sarahgm, line 14: 🔴 **The primary docs demo dead-ends after one click.**

`select` marks the step being *entered* complete instead of the step being *left*. I rendered this file verbatim and clicked "Sign in" once:

```
before: ["Sign in … completed", "Choose plan … current step"]
after:  ["Sign in … current step"]
states: ['current', 'upcoming', 'upcoming', 'upcoming']
```

"Choose plan" loses its progress and reverts to inert text, so after a single click nothing else on the page is clickable.

This is the same failure 0f8d9280 fixed in the `Controlled` story, whose comment says recording the entered key is a no-op. The demo did not get the same fix.

```suggestion
  const select = (key: Key) => {
    setCompletedKeys(keys =>
      keys.includes(selectedKey) ? keys : [...keys, selectedKey]
    );
    setSelectedKey(key);
  };
```

packages/components/src/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,660 @@
+import type { Key } from 'react';
```
sarahgm, line 1: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files (component source, `Stepper.styles.ts`, and these stories) but the `Visual-Regression-Tests` workflow has never run for this branch:

```
$ gh run list --workflow=visual-regression-tests.yml --branch=feat/DST-1391-stepper
[]
```

The decision log's "Still open" section names the VRT run as verification owed, so this is worth doing before merge rather than after. Trigger it with a top-level `/run-chromatic` comment on the PR. A reply inside a review thread will not fire it.

Worth running after the connector fix below, since that changes rendered output.

packages/components/src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,280 @@
...
+
+  const completed = new Set(completedKeys);
+  const errors = new Set(errorKeys);
+  const disabled = new Set(disabledKeys);
+  const selectable = selectableKeys ? new Set(selectableKeys) : undefined;
+
+  const getState = (key: Key): StepState => {
+    if (disabled.has(key)) return 'disabled';
+    if (errors.has(key)) return 'error';
+    if (key === selectedKey) return 'current';
```
sarahgm, line 179: ⚠️ **The connector un-fills when you walk back to a completed step.**

`current` is tested before `completed`, so a step that is both reports `data-state="current"`. The theme fills the connector only on `in-data-[state=completed]` (`themes/theme-rui/src/components/Stepper.styles.ts:117`), so that step's outgoing line goes grey.

Verified with `completedKeys={['contact','delivery']}` and `selectedKey="contact"`:

```
states: ['current', 'completed', 'upcoming']
```

Segment 1 renders grey while segment 2 is green, so the trail reads as broken in exactly the "jump back to a finished step" flow the docs recommend. Reachable from both `MultiStepForm` and `Controlled`.

The priority order itself is right, the connector just needs a different input. Two options: key it off the step it leads *to*, or emit a separate `data-completed` on the `<li>` so "current" and "completed" can both be true at once. `data-current` already exists for a similar reason.

packages/components/src/Stepper/Stepper.tsx
sarahgm, line 81: ⚠️ **No `size` prop**, and 💡 **`variant` has no JSDoc.**

On the missing `size`: decision 10 cut small/large sizes and decision 19 says no size variants in the theme, but neither says the prop should not exist. `size` is the theme's second universal hook. `Breadcrumbs` and `Tabs` both expose it, and having it means a theme can add sizes later without a component change.

On the JSDoc: every other prop in `StepperProps` has one, and `AutoTypeTable` renders a description column, so `variant` will show up blank on the docs page.

```suggestion
  /**
   * Selects a themed appearance.
   * @default 'default'
   */
  variant?: 'default' | (string & {});

  /**
   * Selects a themed size.
   * @default 'default'
   */
  size?: 'default' | (string & {});
```

If you take the `size` half, it also needs destructuring in `_Stepper` and passing to `useClassNames({ component: 'Stepper', variant, size })`, plus the `Stepper` entry in `packages/system/src/types/theme.ts`.

packages/system/src/types/theme.ts
```diff
@@ -209,6 +209,21 @@ export type Theme = {
...
+    // `upcoming`) is not a slot variant: `useClassNames` only carries `variant`
+    // and `size`, and the parent renders every step in a loop, so it cannot call
+    // the hook per item. The component emits `data-state` on each `<li>` and the
+    // theme selects on it instead. See theme-rui's Stepper.styles.ts.
+    Stepper?: Record<
```
sarahgm, line 212: 💡 **Seven slots here, but the PR body and decision 19 both say six.**

`count` is the extra one, added for the `hideLabels` counter. It is disclosed in the PR description and it is clearly needed, so this is not a design objection. The decision log is the stated grading baseline for this PR though, so a one-line amendment there (or in the body) would keep the record and the code in agreement for whoever reads this next.

packages/components/src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,280 @@
...
+
+  const currentIndex = items.findIndex(item => item.props.id === selectedKey);
+
+  // A landmark with nothing in it is noise in the landmark list.
+  if (items.length === 0) return null;
+
+  return (
+    <nav
+      aria-label={ariaLabel ?? stringFormatter.format('progress')}
+      className="flex items-center gap-3"
```
sarahgm, line 198: 💡 **Decision 19 says the `<nav>` stays unstyled, but it carries three layout classes.**

`flex items-center gap-3` is hardcoded here, so a theme cannot change the gap between the counter and the list. There is plenty of precedent for hardcoded layout classes on structural wrappers (`Calendar`, `Sidebar`, `HelpText` all do it), so this is fine as-is. It is only worth a mention because decision 19 explicitly promised the opposite.

docs/content/components/navigation/stepper/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+<ComponentDemo name="stepper-hide-labels" />
+
+### Accessibility
+
+- The stepper renders a `<nav>` landmark around an ordered list, so screen reader users can jump to it or past it. Give it an `aria-label` when a page has more than one landmark, otherwise it falls back to a localized "Progress".
+- Avoid labelling both the stepper and a `<Panel>` wrapped directly around it. Two nested labelled regions add noise without adding information.
+- Each step announces its label, its position, and its state, for example "Choose plan Step 2 of 4, current step".
+- Activation follows the element: Enter on a link step, Enter or Space on a button step.
+- Colour transitions are removed when the user prefers reduced motion.
```
sarahgm, line 166: 💡 **Focus management after a step change is not covered.**

Activating a step swaps the panel but leaves focus on the stepper, with nothing announcing that the content changed. Neither this section nor `MultiStepForm` mentions it, and consumers will hit it immediately since this component navigates as well as reports.

One bullet would do it, something like: move focus to the new step's heading after a step change, so keyboard and screen reader users land in the content they just navigated to.

docs/content/components/navigation/stepper/stepper-links.demo.tsx
```diff
@@ -0,0 +1,45 @@
...
+    setVisited(paths => (paths.includes(path) ? paths : [...paths, path]));
+    setPath(href);
+  };
+
+  return (
+    <RouterProvider navigate={navigate}>
+      <Stack space={4}>
+        <Stepper
+          aria-label="Checkout progress"
+          selectedKey={ROUTES[index].id}
```
sarahgm: 💡 **`ROUTES[index].id` throws when `findIndex` returns `-1`.**

Safe today because every href handed to `navigate` comes from `ROUTES`. Worth guarding anyway, since docs demos get copy-pasted and adding a single route outside the list crashes the page rather than degrading. Same shape at `packages/components/src/Stepper/Stepper.stories.tsx:280`.

sarahgm (changes requested):
Automated review via the `review-pr` skill, graded against the [DST-1391 decision log](https://reservix.atlassian.net/browse/DST-1391) as the description asks. All 21 decisions hold, including the ones easiest to get wrong: the selectability rule (3 + 16), no Adobe port (6), unreachable steps as text (8), and the 44px target that decision 10 flagged as a watch item.

Two blockers and one visible styling bug below, plus four smaller notes. Everything else is in good shape.

packages/components/src/Stepper/Stepper.stories.tsx
aromko: Gone, along with every other implementation comment in the branch — 176 deletions in 9fd196d5. `Stepper.test.tsx`, `Stepper.styles.ts` and all eight docs demos are now at zero comment lines.

Three things that look like comments stayed, because they render as documentation rather than commentary: the `StepperProps` / `StepperItemProps` JSDoc (it fills the description column of `AutoTypeTable` on the docs page), the `/** */` block above each story (Storybook renders it as the story's description on the docs tab), and one `// eslint-disable-next-line`. Say the word if you want those gone too.

packages/components/src/Stepper/Stepper.stories.tsx
aromko: Shortened, then removed. The paragraph that went first was the "stacking the labels was measured at 492px → 12px" one: that block is a `/** */` above the story, so it renders on the Storybook docs page, which made it internal design rationale sitting in user-facing prose. It lives in the DST-1391 decision log instead. The `//` comments in the story went with the branch-wide sweep in 9fd196d5.

The story's opening paragraph stayed — it is the description a reader sees on the docs tab, not a note about the code.

themes/theme-rui/src/components/Stepper.styles.ts
aromko: Removed, all of them: the file is 120 → 54 lines with zero comment lines in 9fd196d5.

Worth naming the cost so it is a decision and not an accident. Four of those comments recorded measurements rather than restating the code:

- `grow` over `flex-1`, because `flex-1`'s zero basis produced connectors of 77/98/117/92px in one row
- `success-accent` over `success`, because green-100 is 1.01:1 and invisible as a hairline
- not `ui-control` on the marker, because it composes `ui-frame` and its 8px radius plus white fill overrode `rounded-full`, rendering the current step as a tiny empty TextField
- `text-secondary` over `text-disabled` on a disabled step, because it renders as a `<span>` and so owes the full 4.5:1

The code still does the right thing in each case; nothing now says why, so the next person to touch those lines rediscovers it. Happy to move those four findings to the decision log if you would rather they lived somewhere than nowhere.

packages/system/src/types/theme.ts
aromko: Removed in 9fd196d5. It had already been cut from five lines to two in the previous pass; now it is gone entirely and the `Stepper` entry sits flush with the other slot records.

docs/content/components/navigation/stepper/stepper-basic.demo.tsx
aromko: Confirmed and fixed in 9fd196d5. I rendered the demo in a browser test rather than trusting the read, and it reproduces exactly as you describe:

```
after one click on "Sign in":
  states:  ['current', 'upcoming', 'upcoming', 'upcoming']
  buttons: ['1Sign inStep 1 of 4, current step']   <- the only control left
```

Took your suggestion verbatim. It is the same fix 0f8d9280 made in the `Controlled` story, so the demo and the story now agree.

The part worth flagging: the demo's own comment already said "Completion only ever grows", which describes the *fixed* behaviour. The comment was right and the code was not, which is presumably how this got past both of us.

packages/components/src/Stepper/Stepper.stories.tsx
aromko: Agreed that it is owed, and it has not run. I am not dispatching it.

One ordering note: run it after 9fd196d5 rather than before. The `data-completed` fix below changes rendered output in `MultiStepForm` and `Controlled` (the connector behind a revisited step now fills), so a run against the previous commit would need doing twice.

packages/components/src/Stepper/Stepper.tsx
aromko: Confirmed and fixed in 9fd196d5. Reproduced in a browser test with your exact inputs:

```
completedKeys={['contact','delivery']} selectedKey="contact"
  states: ['current', 'completed', 'upcoming']
```

So connector 1 sits under `data-state="current"` and `in-data-[state=completed]` never matches it, while connector 2 is green. The trail reads as broken.

Took your second option: a separate `data-completed` on the `<li>`, with the connector keyed off `in-data-completed:`. That is the axis `data-current` already established for the identical `current` + `error` collapse last round, and it is two lines.

Rejected the first option — keying the connector off the step it leads *to* — because `completedKeys` is documented as a non-contiguous set. Completing step 3 but not step 2 would then fill the 2→3 line underneath a ringed marker.

Marker and label needed no change: current-wins is the intended design for the marker, and completed/current both resolve to `text-foreground` on the label. Regression test added at `Stepper.test.tsx` ("marks a completed step even when its state resolves to current"). The green itself still wants the Chromatic run, since the unit environment has no compiled Tailwind — `in-data-completed:` is the same variant form as the shipping `in-data-current:font-medium`.

packages/components/src/Stepper/Stepper.tsx
aromko: Took both halves in 9fd196d5, though the `size` half for a different reason than the one you gave.

"A theme could add sizes later without a component change" is the same argument I declined the `ref` request on earlier in this PR, so on its own it would not carry. What does carry is the release coupling: a `ref` addition is one package, but a theme adding sizes without the prop needs a `@marigold/components` release to land first. That asymmetry is worth three lines. And you are right that decision 10 cut small/large as designs, not the theme hook.

One correction to the suggestion: no `theme.ts` change is needed. `ComponentStyleFunction`'s `Variants` / `Sizes` / `Additional` type parameters are documented as non-constraining (`theme.ts:19-21`, "retained for backward source-compatibility ... they no longer constrain the parameter shape"), and both `variant` and `size` are typed `any` on the call signature. So `size` flows through `useClassNames` with the slot record untouched.

JSDoc on `variant` was unarguable — it was rendering as a blank description in `AutoTypeTable`.

packages/system/src/types/theme.ts
aromko: Agreed, and no code change. Handling it as a record fix on the ticket rather than here, since the decision log is the stated grading baseline: an amendment to decision 19 covering the seventh slot, the `<nav>`'s layout classes (see the thread below — same root cause, decision 19 predates the `hideLabels` counter existing as a rendered element), and the `size` prop from the thread above. Net contract: seven slots, one size, `<nav>` layout owned by the component.

packages/components/src/Stepper/Stepper.tsx
aromko: No action, and thanks for saying so explicitly — it saves re-litigating it.

For the record, this was raised in the first review round and declined on the grounds that an eighth theme slot for a three-token flex row makes the slot-contract problem worse rather than better, given `count` had already put the component past six. Your precedent list is the same one: `Calendar`, `Sidebar`, `HelpText`.

Where you are right is that decision 19 promised the opposite, so the mismatch is real. That goes into the decision-19 amendment alongside the `count` slot rather than into the code.

docs/content/components/navigation/stepper/index.mdx
aromko: Added in 9fd196d5, in the Accessibility section:

> Move focus into the new panel after a step change, to its heading for example. Activating a step swaps the content but leaves focus on the stepper, so without this nothing tells a keyboard or screen reader user that the page moved on.

Genuine gap — this component navigates as well as reports, and nothing else on the page said so.

docs/content/components/navigation/stepper/stepper-links.demo.tsx
aromko: Guarded in 9fd196d5, in both places, and smaller than the suggestion — the index goes away instead of gaining a check:

```diff
- const index = ROUTES.findIndex(route => route.href === path);
+ const current = ROUTES.find(route => route.href === path) ?? ROUTES[0];
…
- selectedKey={ROUTES[index].id}
+ selectedKey={current.id}
```

Net one line shorter and there is no `-1` to handle. Same in `Stepper.stories.tsx`.

Agreed on the reasoning: it cannot fire today, but a docs demo is copy-paste material, and degrading beats crashing when someone adds a route outside the list.

packages/components/src/Stepper/Stepper.tsx
sarahgm: ⚠️ **Follow-up on the fix: `data-completed` has no ladder, so `error` and `disabled` no longer suppress a green connector.**

Taking the `data-completed` option was the right call and this is its tail, not a reason to go back to the other one.

`getState` (`:176-181`) returns a single winner, but `data-completed` (`:224`) is a raw set test. The two disagree whenever a completed step is also errored or disabled:

| Step is | `data-state` | `data-completed` | Connector before | Connector after |
| completed + current | `current` | yes | grey (the bug above) | green, intended |
| completed + **error** | `error` | yes | grey | **green** |
| completed + **disabled** | `disabled` | yes | grey | **green** |

Reachable in the shipped `ValidationErrors` story, which keeps `completed` and `failed` as independent state with nothing holding them disjoint:

1. Valid email, press Continue. `completed = ['contact']`, lands on Payment.
2. Click "Contact" in the stepper. Selectable because it is completed.
3. Drop the `@`. `edit` clears `failed` for contact, `completed` keeps it.
4. Press Continue. Now `failed = ['contact']` **and** `completed = ['contact']`.

Contact renders a `TriangleAlert` in a destructive marker with a success-green line leaving it. The marker says "come back here" and the trail says the segment is cleared. The `disabled` row is the stranger of the two: a dashed inert marker with a green line out of it.

Traced from source rather than rendered, so worth a click before acting on it.

Nothing covers it in either direction. `States` (`Stepper.stories.tsx:101`) is the only story that snapshots the state matrix and its four keys are disjoint, and `ValidationErrors` sets `disableSnapshot: true`. So the Chromatic run will not catch this one. The new test at `Stepper.test.tsx:222` pins completed + current only.

If you want the ladder back everywhere except `current`, it is one guard on the emit:

```diff
-              data-completed={completed.has(id) || undefined}
+              data-completed={
+                (completed.has(id) && !disabled.has(id) && !errors.has(id)) ||
+                undefined
+              }
```

Happy to be told green is correct here, since the step genuinely was completed and the connector describes the segment behind it. It is worth deciding rather than inheriting, because `States`'s own description advertises `disabled > error > current > completed > upcoming` as the render priority order, and the connector now sits outside it.

aromko: Confirmed, and taken as you wrote it in `efcce8f31`.

Traced it the same way you did and it holds: `getState` returns a single winner, `data-completed` was a raw set test, and the connector is keyed off `in-data-completed:` (`Stepper.styles.ts:51`), so it sat outside the ladder entirely.

Reachability in `ValidationErrors` checks out too. `edit` clears `failed` for the step but never touches `completed` (`:504-508`), and the failure branch of `onContinue` adds to `failed` and returns before reaching `completed` (`:511-517`). Nothing holds them disjoint, so `completed=['contact']` + `failed=['contact']` is real state.

On your question of whether green is correct here: it is not, and the guard is the answer.

```diff
-              data-completed={completed.has(id) || undefined}
+              data-completed={
+                (completed.has(id) && !disabled.has(id) && !errors.has(id)) ||
+                undefined
+              }
```

`current` stays exempt because that is the case we fixed last round: walking back to a finished step, where the trail behind it genuinely is cleared. `error` is not the same case. An errored step is a "come back here" signal and it already outranks `current` in the ladder, so it should outrank the exemption too. `disabled` is the row you called the stranger of the two, and I agree: a green line out of a dashed inert marker says the segment is cleared while the marker says the step is not even available.

Worth saying where the fix went, since there was a tempting wrong place for it. `ValidationErrors` could have dropped `completed` on failure and the reachable path would have disappeared, but state is consumer-owned by design and `getState` already publishes a precedence rather than forbidding overlap. Fixing the story would have hidden one path and left every other consumer holding the same bug.

Coverage, since you were right that nothing covered it in either direction:

- two unit tests pinning `data-completed` absent for completed + error and completed + disabled. I checked they actually bite: reverting the guard fails exactly those two and nothing else. Attribute assertions, so the missing compiled Tailwind in the unit environment does not matter here
- `States` now carries the two overlapping rows so Chromatic covers the rendered green. They go in **before** `later`, not appended, for two reasons: the connector only renders while `index < items.length - 1`, so an overlap row placed last would render no connector and guard nothing, and indices 0-3 stay stable for the five existing tests that index into the story. Only the priority test needed updating, and asserting the overlap rows there suits a test called "derives data-state for each step in priority order"

Also documented the thing your last paragraph was really about, that the connector now sits outside the advertised order, in the States section of the docs page rather than only in the story docstring.

Full run on `efcce8f31`: 32 unit tests, 20 story tests, `typecheck:only` clean, `typecheck:docs` clean after a build, eslint 0 errors.

packages/components/src/Stepper/Stepper.stories.tsx
aromko: Still owed, still not dispatched by me.

Updating the ordering note: run it after `efcce8f31` rather than after `9fd196d5`. Your follow-up on the connector thread was right, and the guard for it changes rendered output again. `States` also gained two rows so that the completed + error and completed + disabled combinations are actually in a snapshotted story, which they were not before, and that is the whole reason the run is worth having on this commit rather than the previous one.

Nothing else has landed since, so this is the commit to run.

sarahgm (approved):

PR #5758, opened 2026-08-20
fix(DST-1675): treat `collapsed` as axis-less when resolving inset padding

OsamaAbdellateef:
# Description

`resolveInsetAxes` appended an axis suffix to **every** non-numeric inset token, including `collapsed` — the one token where an axis split is meaningless, since it means "no spacing" on both axes. So `<Panel p="collapsed">` resolved to `var(--spacing-collapsed-x)` rather than `var(--spacing-collapsed)`.

That variable went undeclared until DST-1672 (#5693) added it, so the declaration was invalid at computed-value time and — `padding` not being inherited — fell back to its initial value of `0`. The right answer for entirely the wrong reason. The cost was that every theme implementing this API had to ship three variables that all mean `0` to make one documented prop value work, with nothing to tell it so.

**Current → new behaviour:** `<Panel p="collapsed">` still computes `padding: 0px`, but now because `var(--spacing-collapsed)` *resolves*, not because the reference fails. A theme declares one token instead of three.

Three things worth a reviewer's attention:

- **`isAxislessToken` is a new export from `@marigold/system`**, sitting next to `isScale`. It's backed by a `readonly NoSpacingToken[]`, so adding a member to that type is a type error until it's listed here — the concept gets one home rather than a string check inlined at two call sites.
- **`SelectList` got the same guard applied in place, not consolidated.** `SelectList.tsx` hand-rolls the same axis logic and had the identical bug on its `p` prop. The existing comment documents three deliberate divergences from `resolveInsetAxes`, so merging the two functions would be wrong; I left that comment intact and added a line noting that the axis-suffix rule specifically *is* shared.
- **`isScale` is untouched, deliberately.** `collapsed` must stay non-numeric there because `createWidthVar`, `createHeightVar` and `createSpacingVar` all depend on that classification. A test now pins it.

Closes DST-1675

## Screenshots / Preview

This is a "nothing should look different" change — the point is that `p="collapsed"` already rendered at `0`, just for the wrong reason. Rather than a before/after image, here is the computed-style evidence from Storybook after rebuilding `@marigold/system` and `@marigold/theme-rui`, with `--spacing-collapsed-x` / `-y` genuinely absent from the document:

| Panel (`Components/Panel/CustomPadding`) | inline `--panel-px` | computed | section padding |
| ---------------------------------------- | ------------------- | -------- | --------------- |
| `p="square-loose"` | `var(--spacing-square-loose-x)` | `calc(0.25rem * 6)` | `24px` |
| `px="padding-relaxed" py="padding-snug"` | `var(--spacing-padding-relaxed)` | `calc(0.25rem * 4)` | `16px` |
| **`p="collapsed"`** | **`var(--spacing-collapsed)`** | **`0px`** | **`0px`** |

The first two rows are the unchanged control cases. The third is the fix.

## Test Instructions

1. `pnpm install && pnpm build` (the system + theme packages must be rebuilt — Storybook resolves `@marigold/system` from `dist/`).
2. `pnpm test:unit` — new `resolveInsetAxes` cases for `collapsed` via both `p` and `defaultInset`, a new `isAxislessToken` block, and a `SelectList` case for the inline vars.
3. `pnpm test:sb` — `SelectList/WithCustomPadding` gains a `.test()` that asserts computed `paddingLeft`/`paddingTop` are `0px` in a real browser. This is the assertion the unit suite structurally *cannot* make, and the reason the original defect was invisible: a CSS custom-property declaration can't be checked from jsdom.
4. `pnpm sb` → `Components/Panel/CustomPadding`. The third panel (`p="collapsed"`) should sit flush to its edges; the first two should be unchanged.
5. To confirm the fix is load-bearing: re-add `--spacing-collapsed-x`/`-y` to `themes/theme-rui/src/tokens.css` and note that nothing changes — they are genuinely dead now.

## Breaking Changes

No — but flagging one judgment call for the reviewer.

`--spacing-collapsed-x` and `--spacing-collapsed-y` are **removed** from `@marigold/theme-rui`. They are CSS custom properties that did ship in a released version (they came in with the v18 changes, #5304), so by a strict reading this is a public removal.

I've treated it as a patch because they existed *only* to satisfy a suffix the resolver no longer generates: nothing in the repo referenced them, they were never documented, and the documented token has always been `collapsed`. A consumer would have had to hand-write `var(--spacing-collapsed-x)` against an internal workaround to be affected. Migration in that unlikely case is a one-word edit to `var(--spacing-collapsed)`.

Happy to bump this to a minor and add the 🚨 label if you read it the other way — I didn't want a breaking-change label sitting on a patch changeset without agreement.

## Checklist

### DIFF
--- a/.changeset/dst-1675_axisless-collapsed-token.md
+++ b/.changeset/dst-1675_axisless-collapsed-token.md
@@ -14,4 +14,4 @@ That variable went undeclared until it was added to `@marigold/theme-rui`, so th
 
 `isScale` is unchanged: `collapsed` stays non-numeric there, because `createWidthVar`, `createHeightVar` and `createSpacingVar` depend on that classification.
 
-**Theme:** `--spacing-collapsed-x` and `--spacing-collapsed-y` are removed from `@marigold/theme-rui`. They existed only to satisfy the suffix that is no longer generated, and nothing else referenced them.
+**Theme:** `--spacing-collapsed-x` and `--spacing-collapsed-y` are removed from `@marigold/theme-rui`. They shipped in `theme-rui` 6.0.0 stable but existed only to satisfy the suffix that is no longer generated — nothing else referenced them, and the documented token has always been `collapsed`.
--- a/.changeset/dst-1675_axisless-collapsed-token.md
+++ b/.changeset/dst-1675_axisless-collapsed-token.md
@@ -1,0 +1,17 @@
+---
+'@marigold/system': patch
+'@marigold/components': patch
+'@marigold/theme-rui': patch
+---
+
+fix(DST-1675): treat `collapsed` as axis-less when resolving inset padding
+
+`resolveInsetAxes` appended an axis suffix to every non-numeric inset token, including `collapsed` — the one token where an axis split is meaningless, since it means "no spacing" on both axes. So `<Panel p="collapsed">` resolved to `var(--spacing-collapsed-x)` rather than `var(--spacing-collapsed)`.
+
+That variable went undeclared until it was added to `@marigold/theme-rui`, so the declaration was invalid at computed-value time and `padding` fell back to its initial value of `0` — the right answer for the wrong reason. The cost was that every theme implementing this API had to ship three variables that all mean `0` (`--spacing-collapsed`, `-x`, `-y`) to make one documented prop value work, with nothing to tell it so.
+
+`collapsed` is now recognised as axis-less and passed through unsuffixed, so a theme declares `--spacing-collapsed` only. The same guard is applied to the deliberately-divergent copy of the axis logic in `SelectList`, which had the identical bug on its `p` prop. A new `isAxislessToken` predicate is exported from `@marigold/system` so the concept has one home instead of being inlined at both call sites.
+
+`isScale` is unchanged: `collapsed` stays non-numeric there, because `createWidthVar`, `createHeightVar` and `createSpacingVar` depend on that classification.
+
+**Theme:** `--spacing-collapsed-x` and `--spacing-collapsed-y` are removed from `@marigold/theme-rui`. They existed only to satisfy the suffix that is no longer generated, and nothing else referenced them.
--- a/packages/components/src/Panel/Panel.stories.tsx
+++ b/packages/components/src/Panel/Panel.stories.tsx
@@ -903,5 +903,21 @@ export const CustomPadding = meta.story(() => (
         <Button>Action</Button>
       </Panel.Footer>
     </Panel>
+
+    <Panel p="collapsed">
+      <Panel.Header>
+        <Title>No padding</Title>
+        <Description>
+          Using <code>p="collapsed"</code> — the universal no-spacing token, for
+          wrappers that should render edge to edge.
+        </Description>
+      </Panel.Header>
+      <Panel.Content>
+        <Text>Every section sits flush against the panel edge.</Text>
+      </Panel.Content>
+      <Panel.Footer>
+        <Button>Action</Button>
+      </Panel.Footer>
+    </Panel>
   </Stack>
 ));
--- a/packages/components/src/SelectList/SelectList.stories.tsx
+++ b/packages/components/src/SelectList/SelectList.stories.tsx
@@ -755,10 +755,11 @@ WithCustomPadding.test(
       'var(--spacing-collapsed)'
     );
 
-    // And it resolves, rather than falling back to 0 because it is undefined.
-    const styles = getComputedStyle(standardRow);
-    expect(styles.paddingLeft).toBe('0px');
-    expect(styles.paddingTop).toBe('0px');
+    // Resolution, not fallback: a failed substitution would compute to the
+    // guaranteed-invalid value, which reads back as ''.
+    const listStyles = getComputedStyle(list);
+    expect(listStyles.getPropertyValue('--selectlist-item-px')).not.toBe('');
+    expect(listStyles.getPropertyValue('--selectlist-item-py')).not.toBe('');
   }
 );
 
--- a/packages/components/src/SelectList/SelectList.stories.tsx
+++ b/packages/components/src/SelectList/SelectList.stories.tsx
@@ -737,6 +737,31 @@ WithCustomPadding.test(
   }
 );
 
+WithCustomPadding.test(
+  'resolves the axis-less `collapsed` token to zero padding',
+  {
+    parameters: { chromatic: { disableSnapshot: true } },
+    args: { p: 'collapsed' },
+  },
+  async ({ canvas }) => {
+    const standardRow = await canvas.findByRole('row', { name: /standard/i });
+    const list = standardRow.closest('[role="grid"]') as HTMLElement;
+
+    // No `-x` / `-y` suffix: a theme only declares `--spacing-collapsed`.
+    expect(list.style.getPropertyValue('--selectlist-item-px')).toBe(
+      'var(--spacing-collapsed)'
+    );
+    expect(list.style.getPropertyValue('--selectlist-item-py')).toBe(
+      'var(--spacing-collapsed)'
+    );
+
+    // And it resolves, rather than falling back to 0 because it is undefined.
+    const styles = getComputedStyle(standardRow);
+    expect(styles.paddingLeft).toBe('0px');
+    expect(styles.paddingTop).toBe('0px');
+  }
+);
+
 export const EmptyState = meta.story({
   args: {
     items: [],
--- a/packages/components/src/SelectList/SelectList.stories.tsx
+++ b/packages/components/src/SelectList/SelectList.stories.tsx
@@ -17,6 +17,7 @@ import { SelectList } from './SelectList';
 // it back to `undefined` so the component falls through to the theme's
 // variant-based fallback.
 const insetTokens = [
+  'collapsed',
   'square-tight',
   'square-snug',
   'square-regular',
--- a/packages/components/src/SelectList/SelectList.test.tsx
+++ b/packages/components/src/SelectList/SelectList.test.tsx
@@ -458,6 +458,21 @@ describe('SelectList', () => {
       );
     });
 
+    test('`p="collapsed"` writes the unsuffixed token on both axes', () => {
+      render(<Basic.Component p="collapsed" />);
+
+      const grid = screen.getByRole('grid') as HTMLElement;
+
+      // `collapsed` means "no spacing" on both axes, so there is no
+      // `--spacing-collapsed-x` / `-y` for a theme to declare.
+      expect(grid.style.getPropertyValue('--selectlist-item-px')).toBe(
+        'var(--spacing-collapsed)'
+      );
+      expect(grid.style.getPropertyValue('--selectlist-item-py')).toBe(
+        'var(--spacing-collapsed)'
+      );
+    });
+
     test('axis-specific `px` / `py` write only the matching axis var inline', () => {
       render(<Basic.Component px="padding-relaxed" py="padding-tight" />);
 
--- a/packages/system/src/utils/css-variables.utils.test.ts
+++ b/packages/system/src/utils/css-variables.utils.test.ts
@@ -38,6 +38,11 @@ describe('isScale', () => {
   ])('should return %s for "%s"', (input, expected) => {
     expect(isScale(input)).toBe(expected);
   });
+
+  it('does not classify the no-spacing token as a scale value', () => {
+    // createWidthVar / createHeightVar / createSpacingVar depend on this.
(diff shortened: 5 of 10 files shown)


packages/system/src/utils/css-variables.utils.ts
```diff
@@ -1,4 +1,31 @@
...
+/**
+ * The one spacing token that has no per-axis form.
+ *
+ * Kept in sync with `NoSpacingToken` by the annotation: adding a member to that
+ * type is a type error here until it is listed.
+ */
+const AXISLESS_TOKENS: readonly NoSpacingToken[] = ['collapsed'];
+
+/**
+ * Checks whether a spacing token is axis-less, i.e. splitting it into `-x` /
```
sarahgm, line 11: can you shorten the comment to the main point?

packages/components/src/Panel/Panel.stories.tsx
```diff
@@ -903,5 +903,21 @@ export const CustomPadding = meta.story(() => (
         <Button>Action</Button>
       </Panel.Footer>
     </Panel>
+
+    <Panel p="collapsed">
```
sarahgm, line 907: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files but the `Visual-Regression-Tests` workflow has never run on this branch (`gh run list --workflow=visual-regression-tests.yml --branch=dst-1675_append-axis-to-collapsed` returns an empty list).

It matters concretely here: `CustomPadding` sets no `chromatic: { disableSnapshot: true }`, so it is snapshotted, and this adds a third Panel to it. That is a guaranteed snapshot delta to accept — which is what the PR description predicts too.

packages/system/src/utils/css-variables.utils.ts
```diff
@@ -1,4 +1,31 @@
...
+ *
+ * Note this is deliberately *not* folded into {@link isScale}: `collapsed` is
+ * not numeric, and `createWidthVar` / `createHeightVar` / `createSpacingVar`
+ * rely on it being classified that way.
+ *
+ * @param value - The token to test.
+ * @returns `true` if the token has no per-axis form, otherwise `false`.
+ */
+export const isAxislessToken = (value: string) =>
+  (AXISLESS_TOKENS as readonly string[]).includes(value);
```
sarahgm: ⚠️ The sync guarantee this comment promises isn't there.

`readonly NoSpacingToken[]` constrains what may go *into* the array, not that every union member is present. Checked against the repo's own tsc (6.0.3): widening `NoSpacingToken` to `'collapsed' | 'flush'` leaves `['collapsed']` compiling with zero errors.

Runtime is correct today. The cost lands on the next person to add a no-spacing token: they'll trust a guard that doesn't exist, and their token silently keeps getting the `-x`/`-y` suffix — the exact bug this PR fixes.

A `Record` annotation does deliver it (verified: `Property 'flush' is missing in type '{ collapsed: true; }'`) and it drops the cast on line 28. `Object.hasOwn` rather than `in` — `in` walks the prototype chain, so `isAxislessToken('toString')` would return `true`, and this test file already guards that class of input for `createWidthVar`.

```suggestion
/**
 * The one spacing token that has no per-axis form.
 *
 * Kept in sync with `NoSpacingToken` by the annotation: adding a member to that
 * type is a type error here until it is listed.
 */
const AXISLESS_TOKENS: Record<NoSpacingToken, true> = { collapsed: true };

/**
 * Checks whether a spacing token is axis-less, i.e. splitting it into `-x` /
 * `-y` variants would be meaningless.
 *
 * `collapsed` means "no spacing" on both axes, so there is nothing for an axis
 * suffix to distinguish. Treating it as axis-less keeps themes from having to
 * declare three variables (`--spacing-collapsed{,-x,-y}`) that all mean `0`.
 *
 * Note this is deliberately *not* folded into {@link isScale}: `collapsed` is
 * not numeric, and `createWidthVar` / `createHeightVar` / `createSpacingVar`
 * rely on it being classified that way.
 *
 * @param value - The token to test.
 * @returns `true` if the token has no per-axis form, otherwise `false`.
 */
export const isAxislessToken = (value: string) =>
  Object.hasOwn(AXISLESS_TOKENS, value);
```

packages/components/src/SelectList/SelectList.stories.tsx
```diff
@@ -737,6 +737,31 @@ WithCustomPadding.test(
...
+      'var(--spacing-collapsed)'
+    );
+    expect(list.style.getPropertyValue('--selectlist-item-py')).toBe(
+      'var(--spacing-collapsed)'
+    );
+
+    // And it resolves, rather than falling back to 0 because it is undefined.
+    const styles = getComputedStyle(standardRow);
+    expect(styles.paddingLeft).toBe('0px');
+    expect(styles.paddingTop).toBe('0px');
```
sarahgm: 💡 These two assertions pass with or without the fix, so they don't guard what the comment says.

Under the bug, `--selectlist-item-px: var(--spacing-collapsed-x)` is a failed substitution, so `padding-inline` is invalid at computed-value time and falls back to its initial `0`. That is precisely the accident the PR body describes, and it yields `0px` here too. The inline-var assertions above are what actually catch the regression, and they already exist verbatim in `SelectList.test.tsx:465-472`.

Reading the computed *custom property* does discriminate: a failed substitution computes to the guaranteed-invalid value, which `getPropertyValue` reports as `''`.

```suggestion
    // Resolution, not fallback: a failed substitution would compute to the
    // guaranteed-invalid value, which reads back as ''.
    const listStyles = getComputedStyle(list);
    expect(listStyles.getPropertyValue('--selectlist-item-px')).not.toBe('');
    expect(listStyles.getPropertyValue('--selectlist-item-py')).not.toBe('');
```

Unrelated to the code: the description says this assertion "can't be checked from jsdom" — the unit suite runs in real Firefox via `@vitest/browser-playwright`. The real distinction is that theme CSS isn't loaded there.

packages/components/src/SelectList/SelectList.stories.tsx
```diff
@@ -737,6 +737,31 @@ WithCustomPadding.test(
   }
 );
 
+WithCustomPadding.test(
+  'resolves the axis-less `collapsed` token to zero padding',
+  {
+    parameters: { chromatic: { disableSnapshot: true } },
+    args: { p: 'collapsed' },
```
sarahgm, line 745: 💡 Worth adding `collapsed` to the `p` control while you're here.

`insetTokens` (line 19) enumerates all 15 `square-*` / `squish-*` / `stretch-*` values but not `collapsed`, and `paddingTokens` omits it for `px` / `py`. `collapsed` is a member of `InsetSpacingTokens`, so the enumeration is incomplete — and the token this PR fixes can't be reached from the Storybook control, only through this test's `args`.

packages/components/src/SelectList/SelectList.test.tsx
```diff
@@ -458,6 +458,21 @@ describe('SelectList', () => {
       );
     });
 
+    test('`p="collapsed"` writes the unsuffixed token on both axes', () => {
```
sarahgm, line 461: 💡 Good test. The Panel path has no equivalent, though.

Panel is the running example throughout — the ticket body, the PR description and the changeset — but its coverage here is a story with no assertions. `Panel.test.tsx:126` already has "`p` recipe drives both --panel-px and --panel-py" inside `describe('Spacing props …')`; a sibling case asserting `var(--spacing-collapsed)` would pin the Panel path the way this pins SelectList. `Card.test.tsx:102` has the same hook if you want symmetry.

packages/system/src/utils/css-variables.utils.test.ts
```diff
@@ -205,6 +206,42 @@ describe('resolveInsetAxes', () => {
...
+    'reports "%s" as having per-axis forms',
+    token => {
+      expect(isAxislessToken(token)).toBe(false);
+    }
+  );
+
+  it('does not classify the no-spacing token as a scale value', () => {
+    // createWidthVar / createHeightVar / createSpacingVar depend on this.
+    expect(isScale('collapsed')).toBe(false);
+  });
```
sarahgm: 💡 This asserts an `isScale` guarantee, so `describe('isScale')` (line 15) is arguably its home. Worth pinning either way, given the ticket's explicit "no change to `isScale`". Minor.

.changeset/dst-1675_axisless-collapsed-token.md
```diff
@@ -0,0 +1,17 @@
...
+
+`collapsed` is now recognised as axis-less and passed through unsuffixed, so a theme declares `--spacing-collapsed` only. The same guard is applied to the deliberately-divergent copy of the axis logic in `SelectList`, which had the identical bug on its `p` prop. A new `isAxislessToken` predicate is exported from `@marigold/system` so the concept has one home instead of being inlined at both call sites.
+
+`isScale` is unchanged: `collapsed` stays non-numeric there, because `createWidthVar`, `createHeightVar` and `createSpacingVar` depend on that classification.
+
+**Theme:** `--spacing-collapsed-x` and `--spacing-collapsed-y` are removed from `@marigold/theme-rui`. They existed only to satisfy the suffix that is no longer generated, and nothing else referenced them.
```
sarahgm: 💬 On the breaking-change call you flagged in the description: patch is right, but one fact for the record.

These shipped in `@marigold/theme-rui` **6.0.0 stable**, not only in an rc, and they sit inside `@theme static`, so Tailwind also generated utilities from them (`p-collapsed-x`, `gap-collapsed-x`, …). The surface is a little wider than a hand-written `var(--spacing-collapsed-x)`.

Doesn't change the conclusion — undocumented, nothing referenced them, `p-collapsed` is the documented spelling. But adding "shipped in 6.0.0" to this paragraph makes the removal traceable from the changelog rather than only from this PR.

sarahgm (changes requested):

packages/system/src/utils/css-variables.utils.ts
OsamaAbdellateef: Shortened — dropped the prose, kept just the sync-guarantee statement. In 3c449b5f.

packages/components/src/Panel/Panel.stories.tsx
OsamaAbdellateef: Agreed — this can't be triggered locally, it needs a comment on the PR itself. Will comment `/run-chromatic` separately once we're ready for it (didn't want to fire it silently).

packages/system/src/utils/css-variables.utils.ts
OsamaAbdellateef: Applied your suggested diff as-is: `AXISLESS_TOKENS` is now `Record<NoSpacingToken, true>` and the check uses `Object.hasOwn`. Also added the `__proto__`/`constructor`/`toString` cases to `css-variables.utils.test.ts` mirroring the existing `createWidthVar` guard, so the prototype-chain gap is pinned. In 3c449b5f.

packages/components/src/SelectList/SelectList.stories.tsx
OsamaAbdellateef: Applied your suggested fix — replaced the two `paddingLeft`/`paddingTop` assertions with the `getPropertyValue('--selectlist-item-px'/'-py')).not.toBe('')` check, since that's the one that actually distinguishes resolution from fallback. In 3c449b5f.

packages/components/src/SelectList/SelectList.stories.tsx
OsamaAbdellateef: Added `collapsed` to `insetTokens` so it's reachable from the `p` control. In 3c449b5f.

packages/components/src/SelectList/SelectList.test.tsx
OsamaAbdellateef: Added a mirroring case in `Panel.test.tsx` (``p="collapsed"` writes the unsuffixed token on both axes``), asserting `--panel-px`/`--panel-py` resolve to `var(--spacing-collapsed)` via the existing `CustomPadding` story's 'No padding' region. In 3c449b5f.

packages/system/src/utils/css-variables.utils.test.ts
OsamaAbdellateef: Moved the `isScale('collapsed')` pin into `describe('isScale')`. In 3c449b5f.

.changeset/dst-1675_axisless-collapsed-token.md
OsamaAbdellateef: Added: the changeset now notes these shipped in `theme-rui` 6.0.0 stable, not only an rc. In 3c449b5f.

sarahgm (approved):

sarahgm (approved):
The Chromatic needs to be checked and approved

PR #5761, opened 2026-08-24
fix(DST-1607): align boolean-field controls to the first line of their label

OsamaAbdellateef:
# Description

`Checkbox` and `Radio` anchor their control to the **first line** of the label with `items-start` — correct, since it keeps the box on line one of a label that wraps. The bug: the label line is `text-sm leading-4` (16px), while a default `Badge` is 20px, so a badge placed inline in the label inflated the first line and the 16px control, pinned to its top, stopped reaching the line's optical centre. `Switch` had the mirror bug via `items-center`, plus a `leading-none` label line (14px) against its 16px track.

Flipping to `items-center` fixes the badge but breaks wrapping labels (control floats to the block's middle). So the fix keeps `items-start`/first-line anchoring and instead stops tall decorations from inflating the line:

- **`Badge` gains `size="inline"`** — 16px tall, sized to sit *inside* a line of text rather than beside one.
- **`Checkbox`, `Radio` and `Switch` gain a `labelAdornment` slot** that takes the label line's height (`1lh`) and centres the decoration in it, so a badge lands dead on the line and an oversized decoration overflows symmetrically instead of pushing the control off-centre.
- **`Switch` now renders its own label slot** (16px line height) instead of the shared `Label`, matching `Checkbox`/`Radio`.

`WithBadge` and `LongMultilineLabel` stories for all three components pin both cases under Chromatic, each with a test asserting the control is within 0.5px of the first line's centre.

**Theme authors:** `Switch` has a new `label` style slot.

Closes DST-1607

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Run `pnpm sb` and open `Components/Checkbox`, `Components/Radio`, `Components/Switch` — check the `WithBadge` and `LongMultilineLabel` stories: the control should sit on the first line's centre in both cases.
2. Open `Components/Badge` and check the new `size="inline"` variant, including inline in `docs/content/components/content/badge/badge-inline.demo.tsx`.
3. Run `pnpm test:sb` to confirm the pinned Chromatic interaction tests (control-to-line-centre assertions) pass.

## Breaking Changes

No — `labelAdornment` and `Badge`'s `size="inline"` are additive; `Switch`'s new `label` theme slot is additive for theme authors.

## Checklist


packages/components/src/Radio/Radio.stories.tsx
```diff
@@ -174,6 +180,89 @@ export const WithOwnWidth = meta.story({
...
+      <Radio value="standard">Standard admission</Radio>
+      <Radio
+        value="early-bird"
+        labelAdornment={
+          <Badge variant="master" size="inline">
+            Master
+          </Badge>
+        }
+      >
+        Lorem ipsum, dolor sit amet consectetur adipisicing elit. Qui tempore
```
sarahgm: 🔴 Story and test disagree, and CI is red on it. The `early-bird` radio renders lorem ipsum, but the test below queries `findByRole('radio', { name: /Early bird/ })` and `getByText('Early bird pricing')`. Storybook Tests (2/4) fails with `Unable to find role="radio" and name /Early bird/`.

There is a second problem behind it: this label wraps, so once the query is fixed `expect(isSingleLine(labelBlock)).toBe(true)` on line 218 fails too. The story needs a short label like the Checkbox and Switch versions.

```suggestion
        Early bird pricing
```

packages/system/src/types/theme.ts
```diff
@@ -137,7 +137,7 @@ export type Theme = {
       ComponentStyleFunction<string, string>
     >;
     Switch?: Record<
-      'container' | 'track' | 'thumb',
+      'container' | 'label' | 'track' | 'thumb',
```
sarahgm, line 140: 🔴 Adding the `label` slot here needs the same slot added to the codemod manifest. `packages/cli/src/lib/codemod/manifests/v18.ts:100` still reads `Switch: ['container', 'track', 'thumb']`, and `theme-drift.test.ts` compares the manifest against this type. CLI Tests currently fails with `Switch: missing [label], extra []`. That guard exists to catch exactly this, so it is working as intended.

themes/theme-rui/src/components/Radio.styles.ts
```diff
@@ -11,7 +11,11 @@ export const Radio: ThemeComponent<'Radio'> = {
   }),
   label: cva({
     base: [
-      'text-sm leading-4 font-normal cursor-pointer w-full',
+      // Same row as Checkbox's label, and `items-start` for the same reason: the
+      // dot is top-aligned to the label, so a `labelAdornment` has to stay on
+      // the first line rather than centre itself over a wrapped block.
+      'flex items-start gap-1',
```
sarahgm: 🔴 Turning the Radio label into a flex row collapses element children that relied on filling the label width. The label was a block, so a block child was full width. As a flex item it gets `flex-grow: 0` and shrinks to content width.

Measured in Firefox with these exact utilities: an `<Inline alignX="between">` inside the label goes from 400px wide to 236px, and the space between its two children goes from 172px to 8px, so `justify-between` stops spreading.

This breaks `docs/app/(examples)/examples/inventory/Inventory.tsx:88-108`, where three Radio labels use `<Inline alignX="between">` to push a price to the right edge. `Inline` is `flex gap flex-wrap` with no `w-full`, so nothing restores the stretch. `min-w-0` helps shrinking, not growing. Radio takes arbitrary `children`, so consumer code hits this the same way.

Worth considering the alternative: `LabelAdornment` is already `inline-flex`, so leaving this label a text block and giving the wrapper `align-middle` would avoid the flex change entirely and put the badge genuinely after the last word of line one. As a flex item it currently lands at the right edge of the label column, which is not what the prop docs and the MDX prose promise.

themes/theme-rui/src/components/Switch.styles.ts
```diff
@@ -3,7 +3,12 @@ import { type ThemeComponent, cva } from '@marigold/system';
 export const Switch: ThemeComponent<'Switch'> = {
   container: cva({
     base: [
-      'grid gap-x-2 items-center',
+      // `items-start`, matching Checkbox and Radio: the track belongs on the
+      // label's *first* line. Under `items-center` a wrapping `settings` label
+      // floated the track to the middle of the block instead (DST-1607).
+      // One-line labels are unaffected — the label slot below gives them a
+      // 16px line box, the same height as the track.
+      'grid gap-x-2 items-start',
```
sarahgm, line 6: 🔴 **Visual Regression Tests - not run**

It matters more than usual here, because the PR moves real geometry in three places: Checkbox and Radio label rows go from `items-center` to `items-start`, this container goes from `items-center` to `items-start`, and the Switch label moves off the shared `Label` onto its own slot. That last one takes the label line box from 14px (`text-sm` plus `leading-none`) to 16px (`leading-4`), so the label text re-centres about 1px lower on every labelled Switch. The container height is unchanged, the text position is not.

docs/app/(examples)/examples/event-form/registrationSettings.tsx
```diff
@@ -42,11 +42,11 @@ export const RegistrationSettings = () => (
             </Select>
           </Inline>
           <Checkbox
-            label={
-              <Inline space="related" alignY="center">
-                Enable early bird pricing
-                <Badge variant="master">Master</Badge>
-              </Inline>
+            label="Enable early bird pricing"
+            labelAdornment={
```
sarahgm: ⚠️ This call site is migrated, but the canonical access-mark references are not:

- `docs/content/patterns/layout/admin-master-mark/admin-master-mark-placement.demo.tsx:60-67` renders `<Checkbox label={<>Enable Diagnostics <Badge variant="admin">Admin</Badge></>} />`. That is the published "placement" demo consumers copy from.
- `packages/components/src/patterns/admin-master-mark.stories.tsx:76-80` has the same shape in the `Form` story.

Both put a default 20px badge inline in the label. With the new `items-start` label row the box no longer misaligns, but the badge now hangs below the text line instead of centring on it, so they look wrong in a new way rather than being fixed.

The prose at `docs/content/patterns/layout/admin-master-mark/index.mdx:37` ("place the mark next to the label in form fields") should also point at `labelAdornment` for boolean fields.

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -100,6 +101,13 @@ export interface CheckboxProps extends Omit<
    *
    */
   label?: ReactNode;
+  /**
+   * A decoration shown at the end of the label's first line, e.g. a
+   * `<Badge size="inline">`. Use this instead of putting the decoration in
+   * `label` yourself — the slot keeps it from making the line taller than the
+   * checkbox next to it.
+   */
+  labelAdornment?: ReactNode;
```
sarahgm: ⚠️ `labelAdornment` is hard to read as a name, and it has no precedent in Marigold. Nothing in `packages/`, `docs/` or `themes/` uses the word "adornment". It comes from MUI (`InputAdornment`, `startAdornment`). The Marigold convention is to name a `ReactNode` slot after what goes in it: `action`, `actions`, `icon`.

The tell is the doc comment above. It takes four lines to explain what the prop is for, and it still does not say whether a decoration, an action or a hint belongs there. Every example, demo and story in this PR passes a `<Badge>` and nothing else, and DST-1607 itself proposed "`labelAdornment` (or `badge`)", so the name was left open.

`labelBadge` would say exactly what it takes and need no doc comment. `labelEnd` works if the slot is meant to stay open to other content. This becomes public API on a `minor`, so it is cheapest to settle now.

packages/components/src/utils/LabelAdornment.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * line-height the theme gives the label) and centres the decoration inside it.
+ * A decoration that fits — `Badge size="inline"`, a 16px icon — lands dead on
+ * the line. A taller one overflows symmetrically instead of pushing the line
+ * apart, so the control stays put either way.
+ *
+ * The classes live here rather than in a theme file on purpose: this is the
+ * guardrail, and a theme must not be able to reopen the bug.
+ */
+export const LabelAdornment = ({ children }: { children?: ReactNode }) => (
+  <span className="inline-flex h-[1lh] shrink-0 items-center">{children}</span>
```
sarahgm: ⚠️ Consider having this slot supply `size="inline"` itself instead of asking the consumer to remember it.

DST-1607's scope B frames the slot as the component taking control of "how a decoration is sized/aligned". Right now it controls alignment (this `1lh` box keeps the control from moving) but not sizing, so passing `labelAdornment={<Badge variant="master">Master</Badge>}` still leaves the badge hanging about 4px below the line. The three MDX sections all have to end with the same reminder, "Use `<Badge size="inline">`", which is the sign the API should be doing it.

There is a direct template in the repo. `TagGroup.tsx:304` cascades a look onto a bare child `<Button>`:

```tsx
const removeAllButtonContext = { variant: 'link', size: 'small' } as const;
<Provider values={[[ButtonContext, removeAllButtonContext]]}>
```

`Button.tsx:54-60` reads it with `useSlottedContext` and an explicit prop still wins. `Drawer`, `ToggleButton`, `Checkbox` and legacy `TableRow` all do the same `size ?? ctx.size`.

Applied here: add `Badge/Context.ts` mirroring `Button/Context.ts`, resolve `size: size ?? ctx?.size` in `Badge`, and provide `{ size: 'inline' }` from this wrapper. Worth avoiding `cloneElement` with a `child.type === Badge` check, which breaks on a fragment, a `memo` or a consumer wrapper and would pull `Badge` into every boolean field's bundle.

Two things stay either way: this `1lh` box, because it protects the control from decorations context cannot reach, and a public `size="inline"`, because `badge-inline.demo.tsx` needs it inside a `<Text>` where there is no slot to infer from.

themes/theme-rui/src/components/Badge.styles.ts
```diff
@@ -2,11 +2,15 @@ import { type ThemeComponent, cva } from '@marigold/system';
...
 export const Badge: ThemeComponent<'Badge'> = cva({
   base: [
-    'inline-flex items-center justify-center rounded-full px-2 text-xs font-medium leading-normal',
+    'inline-flex items-center justify-center rounded-full text-xs font-medium',
     'focus-visible:ui-state-focus outline-none',
     'has-[svg]:gap-1',
   ],
   variants: {
+    size: {
+      default: 'px-2 leading-normal',
```
sarahgm: ⚠️ Moving `px-2 leading-normal` out of `base` into `size.default` silently strips them for any unrecognised size. `size` used to be inert (`size?: string` with no size variants), so `<Badge size="small">` still got padding and line-height from `base`.

`cva@1.0.0-beta.4` resolves `variantKey = falsyToString(variantProp) || falsyToString(defaultVariantProp)`, so a truthy unknown value wins over `defaultVariants` and looks up `variants.size['small']`, which is `undefined`. The result is a badge with no horizontal padding and no line-height. The new type `'default' | 'inline' | (string & {})` explicitly invites arbitrary strings, and a downstream theme adding its own sizes is normal.

Keeping the neutral bits in `base` and letting `inline` override them avoids the cliff.

.changeset/dst-1607_boolean-fields-label-adornment.md
```diff
@@ -0,0 +1,37 @@
...
+
+**`Switch` carried the mirror of the same bug** and is now consistent with the other two. It used `items-center`, so a wrapping `variant="settings"` label dropped the track to the middle of the block instead of the first line — measured 28.5px off. It also rendered its label through the shared `Label`, whose `leading-none` gives a 14px line against a 16px track; it now uses its own label slot with a 16px line box, matching `Checkbox` and `Radio`. The accessible name is unchanged — it comes from the wrapping `<label>`'s text either way — and single-line switches keep their exact height and position.
+
+Checkbox's and Radio's label rows moved from `items-center` to `items-start`, which is identical for a single-line label and correct for a wrapping one.
+
+`WithBadge` and `LongMultilineLabel` stories for all three components pin both cases under Chromatic, each with a test asserting the control is within 0.5px of the first line's centre.
+
+**Theme authors:** `Switch` has a new `label` style slot.
```
sarahgm: ⚠️ This is a type-level break for external themes, not purely additive. `Theme['components']['Switch']` is a `Record` with required keys, so any theme outside this repo that defines `Switch` without `label` now fails `tsc`. The PR body says the same thing ("additive for theme authors").

A `minor` may well be the house rule for a new required slot, but the changeset should say what breaks rather than say nothing does.

packages/components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -264,6 +270,77 @@ Basic.test(
...
+      expect(
+        firstLineOffset(canvas.getByText('Master'), labelBlock)
+      ).toBeLessThanOrEqual(0.5);
+    });
+  }
+);
+
+// The other half of DST-1607: `items-center` would fix the badge above and
+// break this one, floating the box to the middle of the block.
+export const LongMultilineLabel = meta.story({
```
sarahgm, line 306: 💡 The two conditions are never pinned together. `WithBadge` covers adornment plus a single line, this one covers wrapping with no adornment. The case the prop exists for, an adornment on a label that wraps, is untested, and it is where the flex-item placement noted on `Radio.styles.ts` shows up. Adding a `labelAdornment` to this story would cover it.

packages/components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -264,6 +270,77 @@ Basic.test(
   }
 );
 
+export const WithBadge = meta.story({
```
sarahgm, line 273: 💡 The guardrail's headline promise is never asserted. `LabelAdornment`'s doc comment says a too-tall decoration "overflows symmetrically instead of pushing the line apart, so the control stays put either way", but every story passes a fitting `size="inline"` badge. A story passing a plain `<Badge>` would pin the claim that the slot survives misuse, which is the main argument for the slot over just documenting `size="inline"`.

packages/components/src/control.utils.ts
```diff
@@ -22,3 +22,52 @@ export const controlIcon = (control: HTMLElement) => {
...
+
+/**
+ * Whether a label block is still exactly one line tall. A decoration that fits
+ * the line leaves this true; one that inflates it -- the DST-1607 bug -- does
+ * not, and that is what pushes the control off centre.
+ */
+export const isSingleLine = (labelBlock: Element) => {
+  const lineHeight = parseFloat(getComputedStyle(labelBlock).lineHeight);
+
+  return labelBlock.getBoundingClientRect().height === lineHeight;
```
sarahgm: 💡 `height === lineHeight` is exact float equality. Fine at 16px today, but a theme with a fractional line-height makes `isSingleLine` return `false` for a perfectly good single line, and the failure then reads as a layout bug rather than a measurement one. A sub-pixel epsilon would match the tolerance the offset assertions already use.

docs/content/components/form/checkbox/index.mdx
```diff
@@ -168,6 +168,12 @@ Adding a description to a checkbox is helpful when the label alone isn't enough
 
 <ComponentDemo name="checkbox-description" />
 
+### Label adornment
+
+To mark a checkbox with a `<Badge>` — an access level, a "New" flag — pass it to `labelAdornment` rather than building the decoration into `label` yourself. The slot holds it to the height of the label's line, so the box stays centred on the text instead of drifting above it, and a label that wraps still keeps the box on its first line. Use `<Badge size="inline">`, which is drawn to fit a line of text.
```
sarahgm: 💡 Worth a sentence that the adornment joins the accessible name. The badge sits inside the `<label>`, so the checkbox is announced as "Enable early bird pricing Master", and the Radio test relies on that. Probably right for an access level, but a purely decorative adornment needs `aria-hidden` and nothing here says so.

sarahgm (changes requested):

packages/components/src/Checkbox/Checkbox.tsx
sarahgm: Following up with concrete alternatives, grounded in what this repo already does.

### The naming rule already in the codebase

Every `ReactNode` prop across `packages/components/src`:

```
children  label  description  emptyState  action  actions
title     errorMessage  input  icon  header  content  actionBarOverlay
```

All 13 are **nouns naming the content that goes in the slot**. Not one is an abstract role or position word. `emptyState` and `errorMessage` are compounds, but the second word is still the content noun and the first only scopes it.

`labelAdornment` follows the shape but breaks the rule: "adornment" names a *role* the content plays rather than the content itself. That is why it needs a four-line doc comment while `icon` and `action` need none.

### The wrinkle that decides it

`Radio` has no `label` prop. Its label is `children`. So on `Radio` any `label*` prefix points at a prop that does not exist, which is why the doc comment there has to say "instead of putting the decoration in `children` yourself". The prefix carries a concept rather than a prop name, and only on two of the three components.

Applying the rule and dropping the dead prefix lands on `badge`, which is what DST-1607 proposed as the alternative:

```tsx
<Checkbox label="Enable early bird pricing" badge={<Badge variant="master">Master</Badge>} />
<Radio value="early-bird" badge={<Badge variant="master">Master</Badge>}>Early bird pricing</Radio>
<Switch label="Require approval" badge={<Badge variant="master">Master</Badge>} />
```

No component in the repo has a `badge` prop today, so there is no collision. It sits beside `icon` and `action`, reads the same on all three components, and pairs cleanly with the auto-sizing idea on `LabelAdornment.tsx`: a `BadgeContext` only reaches `Badge`, so a prop named `badge` and a slot that sizes badges describe the same thing. The 1:1 mapping to the `<Badge>` component means the name teaches itself.

### Ranked alternatives

| Name | Case for | Case against |
| --- | --- | --- |
| **`badge`** | Matches the bare-noun convention. Works on `Radio`, where there is no `label`. The ticket's own suggestion. Needs no doc comment. | Closes the slot to non-badge content, which is arguably the point, since that open-endedness is what caused the sizing bug. |
| `labelBadge` | Explicit that it attaches to the label rather than the control. | Reintroduces the phantom `label` on `Radio`, and is redundant on Checkbox and Switch where there is nowhere else it could go. |
| `labelMark` | "Mark" is established Marigold vocabulary for this decoration. The pattern is titled *Admin- and Mastermark* and its prose says "the mark next to the label". Stays content-agnostic. | The docs also promise the slot for a "New" flag, which is not a mark. "Mark" has no component behind it, so it is less discoverable than `badge`. |
| `labelEnd` / `labelAfter` | Positional and honest about intent, without MUI's vocabulary. | Currently inaccurate. As a flex item the badge lands at the label column's right edge, not after the last word of line one. |

### Worth dismissing explicitly

Dropping the prop and exporting the wrapper as something like `<Badge.Inline>` for use inside `label` is the status quo that scope B rejected, because it lets arbitrary tall content back into the line. Not worth revisiting.

Recommendation is `badge`. If the team wants the slot open to future non-badge content, `labelMark` at least borrows vocabulary the design system already publishes.

docs/content/components/content/badge/index.mdx
```diff
@@ -40,6 +40,25 @@ Badges use color to convey meaning at a glance. Choose the variant that matches
 
 A `<Badge>` is a component used to provide information on new updates and notifications. It is used with other components and cannot be used alone. The content of a badge can be arbitrary, but it is preferred to use a string and/or a icon.
 
+### Sizes
+
+A badge at its default size is taller than a line of text, which is right when it sits _next to_ text — in a table cell, a card header, a list row.
```
sarahgm: 💡 The new docs prose leans on em dashes, and we would rather not have them in the published pages.

It is also the minority style here already: only 13 of the 121 `docs/content/**/index.mdx` files contain an em dash, so the other 108 pages are the convention this is departing from. Nothing enforces it mechanically (there is no markdownlint config), which is exactly why it is worth catching in review.

In every case in this PR the dash is doing a job a colon, a comma, or a plain rewrite does better:

```suggestion
A badge at its default size is taller than a line of text, which is right when it sits _next to_ text: in a table cell, a card header, a list row.
```

Four more on the same pattern, all flagged individually below: line 47 of this file, `form/checkbox/index.mdx:173`, `form/radio/index.mdx:122`, and `form/switch/index.mdx:76`.

Note the three form pages also each end with the same "Use `<Badge size=\"inline\">`" sentence, which disappears entirely if the slot supplies the size itself (see the thread on `LabelAdornment.tsx`), and the prop name in all three changes if the rename lands. Worth doing all of it in one pass rather than three.

docs/content/components/content/badge/index.mdx
```diff
@@ -40,6 +40,25 @@ Badges use color to convey meaning at a glance. Choose the variant that matches
...
+
+Inside a line of text it is not. A default badge in the label of a `<Checkbox>`, `<Radio>` or `<Switch>` makes that line taller than the control beside it, and the control ends up sitting above the words it belongs to. Use `size="inline"` there: it is drawn to the height of a text line, so the line — and the control — stay put.
```
sarahgm: 💡 Em dash, same as line 45. The paired dashes around "and the control" read as an aside, but both items are the subject of "stay put", so a plain conjunction is clearer as well as shorter.

```suggestion
Inside a line of text it is not. A default badge in the label of a `<Checkbox>`, `<Radio>` or `<Switch>` makes that line taller than the control beside it, and the control ends up sitting above the words it belongs to. Use `size="inline"` there: it is drawn to the height of a text line, so both the line and the control stay put.
```

docs/content/components/form/checkbox/index.mdx
sarahgm: 💡 Em dashes, same as `content/badge/index.mdx:45`. The pair here wraps an appositive, so commas do the same work.

```suggestion
To mark a checkbox with a `<Badge>`, such as an access level or a "New" flag, pass it to `labelAdornment` rather than building the decoration into `label` yourself. The slot holds it to the height of the label's line, so the box stays centred on the text instead of drifting above it, and a label that wraps still keeps the box on its first line. Use `<Badge size="inline">`, which is drawn to fit a line of text.
```

docs/content/components/form/radio/index.mdx
```diff
@@ -117,6 +117,12 @@ When a radio group is long enough to create clutter, overload, or excessive scro
 
 Options should be arranged by relevance to help users find what they need quickly. Prioritizing by factors such as frequency of use, overall popularity, or recency of interaction ensures the most useful items appear first. Alphabetical order should be considered only when no meaningful ranking criteria are available, as it may not align with user priorities or usage patterns.
 
+### Label adornment
+
+To mark a single option with a `<Badge>` — an access level, a "New" flag — pass it to `labelAdornment` rather than putting it inside the radio's children. The slot holds it to the height of the label's line, so the dot stays centred on the text instead of drifting above it, and a label that wraps still keeps the dot on its first line. Use `<Badge size="inline">`, which is drawn to fit a line of text.
```
sarahgm: 💡 Em dashes, same as `content/badge/index.mdx:45`.

```suggestion
To mark a single option with a `<Badge>`, such as an access level or a "New" flag, pass it to `labelAdornment` rather than putting it inside the radio's children. The slot holds it to the height of the label's line, so the dot stays centred on the text instead of drifting above it, and a label that wraps still keeps the dot on its first line. Use `<Badge size="inline">`, which is drawn to fit a line of text.
```

docs/content/components/form/switch/index.mdx
```diff
@@ -71,6 +71,12 @@ Add a description when the label alone doesn't explain what toggling actually do
 
 <ComponentDemo name="switch-description" />
 
+### Label adornment
+
+To mark a switch with a `<Badge>` — an access level, a "New" flag — pass it to `labelAdornment` rather than building the decoration into `label` yourself. The slot holds it to the height of the label's line, so the track stays centred on the text instead of drifting above it, and a label that wraps still keeps the track on its first line. Use `<Badge size="inline">`, which is drawn to fit a line of text.
```
sarahgm: 💡 Em dashes, same as `content/badge/index.mdx:45`.

```suggestion
To mark a switch with a `<Badge>`, such as an access level or a "New" flag, pass it to `labelAdornment` rather than building the decoration into `label` yourself. The slot holds it to the height of the label's line, so the track stays centred on the text instead of drifting above it, and a label that wraps still keeps the track on its first line. Use `<Badge size="inline">`, which is drawn to fit a line of text.
```

themes/theme-rui/src/components/Badge.styles.ts
sarahgm: Worth adding: this is not just a footgun, it inverts a documented pattern. `docs/content/components/__internal__/component-guidelines.mdx:299` says "To prevent overwriting base styles through variants and sizes we should add `default` values for the styles... In the end only necessary and common classes should be in the base styles", and the `Select` example right below it keeps the shared classes in `base` and uses empty-string defaults in the variant map:

```tsx
icon: cva({
  base: 'text-secondary-400',
  variants: { size: { default: '', small: 'size-3' } },
  defaultVariants: { size: 'default' },
})
```

`px-2 leading-normal` are common classes by that definition, since they applied to every badge before this change. Putting them back in `base` and letting `inline` override them satisfies the guideline and closes the unknown-size hole in one move, so there are two independent reasons for it.

packages/components/src/Radio/Radio.stories.tsx
OsamaAbdellateef: Fixed. The story now carries the short `Early bird pricing` label, matching the Checkbox and Switch versions, so `findByRole('radio', { name: /Early bird/ })` and `getByText('Early bird pricing')` both resolve and `isSingleLine` holds. Storybook Tests are green.

packages/system/src/types/theme.ts
OsamaAbdellateef: Fixed. `v18.ts:100` now reads `Switch: ['container', 'label', 'track', 'thumb']`, so `theme-drift.test.ts` and CLI Tests pass.

Also verified the codemod end-to-end rather than just the unit test: running `marigold migrate v18 <theme> --only stub-missing-slots` against a v17 theme with `container`/`track`/`thumb` correctly stubs in `label: cva({})`. That is the migration path external theme authors need for the breaking type change, so it is worth having exercised.

Two migrate tests had to move with it: they asserted `stub-missing-slots` does *not* fire because "the fixture Switch already has every v18 slot", which stopped being true. I updated the expectations rather than adding `label` to those fixtures, since they model a genuine pre-migration v17 theme and faking completeness there would have hidden the new behaviour instead of pinning it.

themes/theme-rui/src/components/Radio.styles.ts
OsamaAbdellateef: Fixed, and thanks for the measurement, it was the right call.

The Radio label is a plain text block again (`flex items-start gap-1` removed), so an `<Inline alignX="between">` in `children` still fills the label the way it fills any other block container. The badge lands on the first line through normal inline flow via `LabelAdornment` instead of through the label being a flex row.

One thing worth flagging from implementing your suggestion: `align-middle` did not hold. `vertical-align: middle` aligns to half the font's x-height above the baseline, not the line box centre, which inflated the label line by ~1.17px and failed `isSingleLine`. `align-top` is exact, since the wrapper is already `1lh` tall and matches the line's own strut, so aligning its top to the line box top adds zero growth regardless of font metrics. Measured 0.00px offset afterwards.

docs/app/(examples)/examples/event-form/registrationSettings.tsx
OsamaAbdellateef: Fixed. Both canonical references now use the slot instead of building the badge into the label:

- `docs/content/patterns/layout/admin-master-mark/admin-master-mark-placement.demo.tsx` (the published placement demo)
- `packages/components/src/patterns/admin-master-mark.stories.tsx` (the `Form` story)

`admin-master-mark/index.mdx` now points boolean fields at the `badge` slot alongside the existing placement prose.

packages/components/src/Checkbox/Checkbox.tsx
OsamaAbdellateef: Renamed to `badge`, following your recommendation and the bare-noun convention. Applied across `Checkbox`, `Radio` and `Switch`, plus every story, demo, MDX section and the changeset.

The four-line doc comment is gone, as you predicted it would be: the name carries it, and the remaining sentence only explains why the slot exists rather than what it takes.

packages/components/src/utils/LabelAdornment.tsx
OsamaAbdellateef: Done, using the `TagGroup`/`Button` template you pointed at.

- Added `Badge/Context.ts` mirroring `Button/Context.ts`.
- `Badge` resolves `size ?? context.size` via `use()`, so an explicit `size` on the badge still wins.
- `LabelAdornment` provides `{ size: 'inline' }`.

No `cloneElement`, so fragments, `memo` and consumer wrappers all still work and `Badge` is not pulled into every boolean field's bundle.

Both things you said should stay, stayed: the `1lh` box is still there for decorations context cannot reach, and `size="inline"` is still public for `badge-inline.demo.tsx`. Every story and demo now passes a bare `<Badge variant="master">Master</Badge>`, and the "Use `<Badge size="inline">`" reminder is gone from all three MDX sections.

themes/theme-rui/src/components/Badge.styles.ts
OsamaAbdellateef: Fixed. `px-2 leading-normal` moved back into `base` and `size.default` is now `''`, which matches the `Select` icon example in `component-guidelines.mdx:299` exactly. An unrecognised size keeps its padding and line-height instead of falling off the cliff.

Added a story test on the new `Inline` story that asserts the inline badge renders shorter than the default one, so a future refactor that drops `inline`'s overrides fails loudly rather than silently collapsing the two sizes.

.changeset/dst-1607_boolean-fields-label-adornment.md
OsamaAbdellateef: Fixed. The changeset now says it outright instead of claiming the change is purely additive:

> **Breaking for external themes:** `Theme['components']['Switch']` is a `Record` with required keys, so a theme outside this repo that defines `Switch` without the new `label` slot now fails `tsc`.

The `badge` prop and `Badge`'s `size="inline"` are still called out separately as additive.

packages/components/src/Checkbox/Checkbox.stories.tsx
OsamaAbdellateef: Added `WrappingLabelWithBadge`, which pins exactly that combination: a badge on a label that also wraps. It asserts the label still wraps, and that both the box and the badge stay on the first line.

This is also the case that caught the `align-middle` problem noted on `Radio.styles.ts`, so it earned its place immediately.

packages/components/src/Checkbox/Checkbox.stories.tsx
OsamaAbdellateef: Added `OversizedBadge`. It passes an explicit `size="default"` badge into the slot, deliberately overriding the context auto-sizing, and asserts the line stays single and that both the box and the badge remain centred on it.

That pins the overflow guarantee itself, which as you say is the main argument for the slot over just documenting the size.

packages/components/src/control.utils.ts
OsamaAbdellateef: Fixed. `isSingleLine` now compares within a named `SUBPIXEL_TOLERANCE` of 0.5px, matching the tolerance the offset assertions already use, instead of an exact `===`.

docs/content/components/form/checkbox/index.mdx
OsamaAbdellateef: Added. The Checkbox docs now note that the badge joins the accessible name and is announced after the label text, that this is right for an access-level mark, and that a purely decorative badge should carry `aria-hidden` instead.

docs/content/components/content/badge/index.mdx
OsamaAbdellateef: Applied, and done in one pass across all five as you suggested. `grep -c '—'` is now 0 in all four changed `index.mdx` files.

The trailing "Use `<Badge size="inline">`" sentence is gone from the three form pages too, since the slot supplies the size itself now, and the prop name changed to `badge` throughout.

docs/content/components/content/badge/index.mdx
OsamaAbdellateef: Applied. The paired dashes are gone and the sentence now reads as one clause, with the wording adjusted for the `badge` slot supplying the size.

docs/content/components/form/checkbox/index.mdx
OsamaAbdellateef: Applied, with the wording adjusted for the rename to `badge` and the slot now supplying the size. The section is retitled "Badge" and gained the accessible-name note from the other thread.

docs/content/components/form/radio/index.mdx
OsamaAbdellateef: Applied, with the wording adjusted for the rename to `badge` and the slot now supplying the size.

docs/content/components/form/switch/index.mdx
OsamaAbdellateef: Applied, with the wording adjusted for the rename to `badge` and the slot now supplying the size.

packages/system/src/types/theme.ts
sarahgm: Verified. `v18.ts:100` reads `Switch: ['container', 'label', 'track', 'thumb']` and CLI Tests are green.

Moving the two migrate expectations was the better call over padding the fixtures. `codemod.test.ts` now spells out a complete v18 Switch inline for the "leaves complete slot objects unchanged" case and leaves `PORTAL_SWITCH` as the v17 theme it actually is, so both the fires and the does-not-fire path stay honest. Resolving.

themes/theme-rui/src/components/Radio.styles.ts
sarahgm: Verified, and thanks for chasing the `align-middle` half of it down. You are right that `vertical-align: middle` resolves against half the font's x-height rather than the line box, so `align-top` is the correct choice and the reason is now in the doc comment where the next person will find it.

The label is a plain block again. With `RadioButton` as `flex items-start` and the label carrying `w-full min-w-0`, a block child like `<Inline alignX="between">` gets the full label width back, and long unbreakable content can still shrink. Resolving.

docs/app/(examples)/examples/event-form/registrationSettings.tsx
sarahgm: Verified. Both canonical references go through the slot now, and `admin-master-mark/index.mdx` covers all three cases explicitly: `size="inline"` for a badge you put in a field label yourself, the `badge` slot for boolean fields, and the default size for tabs and section headings.

That last sentence also answers the tabs question from the preview comment, and adding `Switch` to the `Form` story alongside the `TextField` answers the other one. Resolving.

packages/components/src/Checkbox/Checkbox.tsx
sarahgm: Verified. `badge` across `Checkbox`, `Radio` and `Switch`, plus every story, demo, MDX section and the changeset. The doc comment is down to the one sentence saying why the slot exists rather than four explaining what belongs in it. Resolving.

packages/components/src/utils/LabelAdornment.tsx
sarahgm: Verified. `Badge/Context.ts`, `size ?? context.size` read through `use()` in `Badge`, and `LabelAdornment` providing `{ size: 'inline' }`. No `cloneElement`, so fragments, `memo` and consumer wrappers all still work.

`WithBadge` passing a bare `<Badge variant="master">` with a comment saying why there is no `size` on it is exactly the outcome I was after, and the "Use `<Badge size=\"inline\">`" reminder is gone from all three MDX sections. The `1lh` box and the public `size="inline"` both survived. Resolving.

themes/theme-rui/src/components/Badge.styles.ts
sarahgm: Verified. `px-2 leading-normal` is back in `base` and `size.default` is `''`, which is the `component-guidelines.mdx` shape exactly. An unrecognised size keeps its padding and line-height.

One note for the record: the new `Inline` story test guards a slightly different thing than this comment did. It pins inline being shorter than default, not the unknown-size fallback. That is fine, the style change is what closes the hole, and a `size="small"` case would only be belt and braces. Resolving.

.changeset/dst-1607_boolean-fields-label-adornment.md
sarahgm: Verified. The changeset states the `Switch` `label` slot break outright and keeps the `badge` prop and `size="inline"` listed separately as additive. Resolving.

packages/components/src/Checkbox/Checkbox.stories.tsx
sarahgm: Verified. `WrappingLabelWithBadge` pins the combination, and it asserts the label still wraps before checking the box and the badge, so it cannot quietly pass for free once the label stops wrapping. Resolving.

packages/components/src/Checkbox/Checkbox.stories.tsx
sarahgm: Verified. `OversizedBadge` pushes an explicit `size="default"` through the slot and asserts the line stays single while both the box and the badge stay centred on it. That is the overflow guarantee from the doc comment, pinned. Resolving.

packages/components/src/control.utils.ts
sarahgm: Verified. `SUBPIXEL_TOLERANCE` of 0.5px, matching the offset assertions, and the `NaN` guard for a `normal` line-height is still in place above it. Resolving.

docs/content/components/form/checkbox/index.mdx
sarahgm: Verified. The Checkbox section now says the badge joins the accessible name, that it is announced after the label text, that this is right for an access-level mark, and that a decorative badge should carry `aria-hidden` instead. Resolving.

themes/theme-rui/src/components/Switch.styles.ts
sarahgm: Still outstanding, and it is now the only thing left on this review.

There is no `Visual-Regression-Tests` run on this branch at all. `gh run list --workflow=visual-regression-tests.yml --branch dst-1607_boolean-fields` comes back empty.

One practical note, since my original comment was imprecise about it: the workflow listens on `issue_comment`, which only fires for a comment on the PR's conversation tab. A reply inside a review thread like this one will not trigger it. So it has to be `/run-chromatic` as a top-level PR comment, or a manual `workflow_dispatch` on the branch.

The geometry it needs to cover has grown since I wrote this: Checkbox and Radio label rows moving to `items-start`, the Switch container doing the same, the Switch label moving off the shared `Label` onto its own 16px line box, and the Switch label now being a flex row where it used to be a `Label` span. Every other thread is verified and resolved, so once the snapshots look right this is good to go.

docs/content/components/form/checkbox/index.mdx
```diff
@@ -168,6 +168,14 @@ Adding a description to a checkbox is helpful when the label alone isn't enough
 
 <ComponentDemo name="checkbox-description" />
 
+### Badge
+
+To mark a checkbox with a `<Badge>`, such as an access level or a "New" flag, pass it to `badge` rather than building the decoration into `label` yourself. The slot holds it to the height of the label's line and sizes it to fit, so the box stays centred on the text instead of drifting above it, and a label that wraps still keeps the box on its first line.
+
+The badge joins the checkbox's accessible name, announced right after the label text. That's right for an access-level mark, but a purely decorative badge should get `aria-hidden` instead.
+
+<ComponentDemo name="checkbox-label-adornment" />
```
sarahgm: 💡 Leftover from the rename, spotted on the re-review. The three demo files are still `checkbox-label-adornment.demo.tsx`, `radio-label-adornment.demo.tsx` and `switch-label-adornment.demo.tsx`, so a `<ComponentDemo>` named `checkbox-label-adornment` sits under a heading called "Badge" and points at a word the API no longer uses. Same for the changeset filename, though that one gets consumed at release and does not matter.

Renaming them to `*-badge.demo.tsx` keeps the docs greppable by the prop name. It needs a `build:registry` run to pick the new names up, since the registry is generated.

Non-blocking, and the only thing actually gating the review is the Chromatic run.

docs/app/(examples)/examples/event-form/registrationSettings.tsx
```diff
@@ -42,12 +42,8 @@ export const RegistrationSettings = () => (
...
           </Inline>
           <Checkbox
-            label={
-              <Inline space="related" alignY="center">
-                Enable early bird pricing
-                <Badge variant="master">Master</Badge>
-              </Inline>
-            }
+            label="Enable early bird pricing"
+            badge={<Badge variant="master">Master</Badge>}
```
sarahgm, line 46: ⚠️ The sibling file in this same directory was missed. `organizerInfo.tsx:103` still puts a default-size badge in a `<Select>` label:

```tsx
<Select
  label={
    <Inline space="related" alignY="center">
      Preferred contact method
      <Badge variant="admin">Admin</Badge>
    </Inline>
  }
```

Measured on the branch preview, label heights in this example app:

| Case | Label height |
| --- | --- |
| Bare field label (`Organizer Name`, `Organizer Type`) | 14px |
| `size="inline"` badge (`Associated Team` in the placement demo) | 16px |
| This one, default badge | 20px |

So it runs 4px taller than the labels this PR did migrate, and it pushes the Select below it down. That is the case the new prose on the pattern page describes, and it is the same shape as the placement demo's Select label, which got `size="inline"`. With this call site migrated and its neighbour not, the event-form example is now inconsistent with itself.

Smallest fix is adding the size:

```tsx
<Badge variant="admin" size="inline">Admin</Badge>
```

Optionally match the shape the placement demo settled on and drop the wrapper, since a fragment with `{' '}` does the same job without an `Inline`.

Worth saying what I checked, so this does not turn into a wider sweep than it needs to be. I went through every `<Badge>` in `docs/` and `packages/`, and this is the only miss:

- No boolean field anywhere still builds a badge into its label. I scanned every `<Checkbox>`, `<Switch>` and `<Radio>` for a JSX `label=` expression containing a `Badge` and got zero hits, so the `badge` slot migration is complete.
- The other three field labels carrying a badge all have `size="inline"` already: the placement demo's `Select`, and the `TextField` and `Radio.Group` in `admin-master-mark.stories.tsx`.
- Tabs and Sidebar nav items are correct on the default size, so please leave them alone. I was suspicious of the "a tab already reserves the badge's height" line since the `tab` slot has no fixed height, so I measured it: the badged tab is 32px and its unbadged siblings are also 32px, because the tab's line box is already 20px. `Sidebar`'s `navLink` is a fixed `h-7.5` flex row, so a 20px badge is absorbed there too.
- Everything else, roughly a hundred usages, sits beside text rather than inside it: table cells, card and panel headers, accordion headers, list rows, Select options, button and filter counts. Default is right for all of them.

docs/content/components/form/checkbox/index.mdx
OsamaAbdellateef: Done in `5ef112a6a`. The three demo files are now `checkbox-badge.demo.tsx`, `radio-badge.demo.tsx` and `switch-badge.demo.tsx`, renamed with `git mv` so the history follows, and the three `<ComponentDemo name=...>` references point at the new names.

Registry regenerated (`pnpm build:registry`) — it resolves `checkbox-badge`, `radio-badge` and `switch-badge`, and no `label-adornment` reference survives in `.registry/demos.json`. Also re-ran `build:md`, since `docs/public/components/**` is generated and still carried the old names; it's gitignored, so nothing to review there, but a local checkout would have been stale otherwise.

I renamed the changeset file to `dst-1607_boolean-fields-badge-slot.md` too. You're right that it gets consumed at release, but it costs nothing and the word is gone from the branch now.

Loaded all three pages against a local dev server to confirm the demos resolve rather than falling back to a not-found state.

I left the internal `LabelAdornment` util under its own name — it is the generic wrapper the three `badge` slots share, not the prop, and its doc comment already says so. Happy to rename it if you'd rather have one word for the whole thing.

docs/app/(examples)/examples/event-form/registrationSettings.tsx
OsamaAbdellateef: Good catch, fixed in `5ef112a6a`. `organizerInfo.tsx` now reads:

```tsx
<Select
  label={
    <>
      Preferred contact method{' '}
      <Badge variant="admin" size="inline">
        Admin
      </Badge>
    </>
  }
```

I took the optional half too — the `<Inline space="related" alignY="center">` wrapper is gone, so it is the same shape as the placement demo's Select label rather than a second way of writing it.

Re-measured on a local dev server after the change, same three cases as your table:

| Case | Before | After |
| --- | --- | --- |
| Bare field label (`Alternative Contact`) | 14px | 14px |
| `size="inline"` badge (placement demo `Associated Team`) | 16px | 16px |
| `Preferred contact method` | 20px | **16px** |

So it lands exactly on the migrated labels and stops pushing the Select down, and the event-form example agrees with itself again.

Thanks for scoping the sweep in the comment — I took the four negative results at your word rather than re-running them, and touched only this one call site. Tabs and Sidebar untouched.

themes/theme-rui/src/components/Switch.styles.ts
OsamaAbdellateef: Dispatched. `Visual-Regression-Tests` is running on this branch now:

https://github.com/marigold-ui/marigold/actions/runs/32980330738 — `workflow_dispatch`, head `5ef112a6`.

Thanks for the correction about `issue_comment`; I went the `workflow_dispatch` route rather than a top-level `/run-chromatic` comment, which sidesteps the question entirely.

It covers the geometry you listed, since `5ef112a6` is the branch tip: Checkbox and Radio label rows on `items-start`, the Switch container likewise, the Switch label on its own 16px line box instead of the shared `Label`'s 14px, and that label now being a flex row. The 1px re-centring on every labelled Switch is the diff to look hardest at — the container height should be unchanged, only the text position.

Two docs-only commits also landed since your last pass, so a few docs snapshots will move for reasons unrelated to the boolean-field geometry: the Badge sizes demo was rebuilt around a table cell and a real Checkbox (`4a39638b`, per your Vercel thread), and the event-form example's `Preferred contact method` label dropped from 20px to 16px (`5ef112a6`, your `organizerInfo.tsx` comment).

packages/components/src/Badge/Badge.stories.tsx
```diff
@@ -72,6 +80,37 @@ export const Basic = meta.story({
...
+      <Text>
+        Inline sits inside one: <Badge {...args} size="inline" />
+      </Text>
+    </div>
+  ),
+});
+
+Inline.test(
+  'The inline size is drawn shorter than the default',
+  { parameters: { chromatic: { disableSnapshot: false } } },
```
sarahgm: 💡 **These nine tests each add a second Chromatic snapshot of a state that is already snapshotted.**

None of the nine new stories carries a story-level `chromatic` parameter, and `.storybook/preview.tsx` sets only `pauseAnimationAtEnd` and `prefersReducedMotion` globally. So every one of them is snapshotted by default, and the attached test is then snapshotted on top of it. That is 18 new snapshots for 9 distinct visual states.

The two are identical because all nine tests are pure measurement: `findByRole`, then `getBoundingClientRect`/`getComputedStyle` through `firstLineOffset` and `isSingleLine`. None of them clicks, hovers, types or changes state, so the test's end state is the story's initial state, pixel for pixel.

That is the case CLAUDE.md's rule excludes: default a `.test()` to `disableSnapshot: true`, and use `false` only when *the interaction's* end state is a visual state worth guarding. The legitimate uses are in these same files, for contrast. `Switch`'s `Basic.test('Hover darkens the track when off, without touching the eased fill')` keeps `false` because hover is a state the story never shows.

```suggestion
  { parameters: { chromatic: { disableSnapshot: true } } },
```

Same change in the other eight:

- `Checkbox.stories.tsx:286`, `:328`, `:363`, `:402`
- `Radio.stories.tsx:204`, `:245`
- `Switch.stories.tsx:324`, `:367`

To be clear about what I am **not** suggesting: keep the story-level snapshots. This PR moves real geometry, so the initial rendered layout is exactly what wants a baseline, and the nine states are genuinely distinct (badge on one line, wrapping label, both at once, oversized badge, across three components). Two of them exist because I asked for them earlier in this review. The duplication is the only thing worth removing.

Worth deciding before the run you just dispatched gets reviewed. If 18 baselines are accepted and 9 of them then disappear, that is avoidable churn in the Chromatic history, whereas changing it now costs one more run.

packages/components/src/control.utils.ts
```diff
@@ -22,3 +22,60 @@ export const controlIcon = (control: HTMLElement) => {
...
+
+/**
+ * Vertical centre of the *first* line box of a label block.
+ *
+ * All three controls anchor to line one, so this is the number they have to
+ * meet whether the label is one line or five. Derived from the block's own
+ * line-height rather than from a `Range`: range rects are the glyph box
+ * (ascent + descent) and sit ~0.5px off the line box, which is the same order
+ * as the misalignment being guarded against. The block carries no padding or
+ * border, so its top *is* the top of line one.
```
sarahgm: can you also check to remove comments with /remove-dumb-comments?

sarahgm:
Nice work on this. One thing I noticed while reading through: should the `badge` slot extend to `TextField` and friends as a follow-up?

The placement demo shows the asymmetry pretty clearly:

```tsx
// boolean field: slot handles the sizing
<Checkbox label="Enable Diagnostics" badge={<Badge variant="admin">Admin</Badge>} />

// everything else: hand-built fragment, and you have to remember size="inline"
<Select label={<>Associated Team <Badge variant="master" size="inline">Master</Badge></>} />
```

And the pattern page now has a line telling people to remember `size="inline"` in field labels, which is the kind of instruction a slot usually exists to remove.

I don't think it belongs in this PR though. It's a different problem: here the control sits *beside* the label, so an inflated first line pushes the box off centre. In `FieldBase` the label sits above the input, so nothing misaligns, the label row just grows from 14px to 20px and shifts the field down. Cosmetic plus API consistency rather than a broken control. And the surface is much wider: `FieldBase` feeds around twenty public components, most of which `Pick<FieldBaseProps, 'label' | 'description' | 'errorMessage'>`, so each one needs the prop, a prop table and a docs section. This PR is already a minor with a breaking theme change for Switch.

The good news is the mechanism you built transfers as is. `LabelAdornment` clamping to `1lh` plus `BadgeContext` means the follow-up is basically one line in `FieldBase`:

```tsx
<Label variant={variant} size={size}>
  {label}
  {badge && <LabelAdornment>{badge}</LabelAdornment>}
</Label>
```

Open questions I'd want settled before we write the ticket:

- Every `FieldBase` consumer, or a named subset? A badge on `Slider` or `ListView` seems odd.
- Do `Checkbox.Group` and `RadioGroup` get one? Their group label goes through `FieldBase`, so after this PR the items can take a badge but the group label still can't, which is the most visible leftover.
- The badge lands inside `<label>`, so it joins the accessible name ("Associated Team Master"). Probably what we want for an access mark, but worth deciding rather than inheriting.
- Does `utils/LabelAdornment` stay internal, or become the shared label-decoration primitive?

Happy to write it up as a follow-up ticket if you agree.

OsamaAbdellateef:
> Nice work on this. One thing I noticed while reading through: should the `badge` slot extend to `TextField` and friends as a follow-up?
> 
> The placement demo shows the asymmetry pretty clearly:
> 
> ```tsx
> // boolean field: slot handles the sizing
> <Checkbox label="Enable Diagnostics" badge={<Badge variant="admin">Admin</Badge>} />
> 
> // everything else: hand-built fragment, and you have to remember size="inline"
> <Select label={<>Associated Team <Badge variant="master" size="inline">Master</Badge></>} />
> ```
> 
> And the pattern page now has a line telling people to remember `size="inline"` in field labels, which is the kind of instruction a slot usually exists to remove.
> 
> I don't think it belongs in this PR though. It's a different problem: here the control sits _beside_ the label, so an inflated first line pushes the box off centre. In `FieldBase` the label sits above the input, so nothing misaligns, the label row just grows from 14px to 20px and shifts the field down. Cosmetic plus API consistency rather than a broken control. And the surface is much wider: `FieldBase` feeds around twenty public components, most of which `Pick<FieldBaseProps, 'label' | 'description' | 'errorMessage'>`, so each one needs the prop, a prop table and a docs section. This PR is already a minor with a breaking theme change for Switch.
> 
> The good news is the mechanism you built transfers as is. `LabelAdornment` clamping to `1lh` plus `BadgeContext` means the follow-up is basically one line in `FieldBase`:
> 
> ```tsx
> <Label variant={variant} size={size}>
>   {label}
>   {badge && <LabelAdornment>{badge}</LabelAdornment>}
> </Label>
> ```
> 
> Open questions I'd want settled before we write the ticket:
> 
> * Every `FieldBase` consumer, or a named subset? A badge on `Slider` or `ListView` seems odd.
> * Do `Checkbox.Group` and `RadioGroup` get one? Their group label goes through `FieldBase`, so after this PR the items can take a badge but the group label still can't, which is the most visible leftover.
> * The badge lands inside `<label>`, so it joins the accessible name ("Associated Team Master"). Probably what we want for an access mark, but worth deciding rather than inheriting.
> * Does `utils/LabelAdornment` stay internal, or become the shared label-decoration primitive?
> 
> Happy to write it up as a follow-up ticket if you agree.

Yes sure : ) 
@sarahgm

packages/components/src/control.utils.ts
OsamaAbdellateef: removed :heavy_check_mark:  @sarahgm

sarahgm: in general i mean not only this file

OsamaAbdellateef: Including  the comments that aren't included in this PR changes ? @sarahgm

sarahgm: No but for example the stories.tsx files you touched, and the styles.tsx files

OsamaAbdellateef: removed comments :heavy_check_mark:  @sarahgm

themes/theme-rui/src/components/Switch.styles.ts
sarahgm: The Chromatic run from 26 Aug no longer covers this branch. It was dispatched on `5ef112a6` and the tip is `aeed507`, four commits later.

The gap is not just distance. The merge in `859a16fbc` resolved `Switch.tsx` in main's favour and dropped the `badge` prop, `labelSlot` and the `LabelAdornment` import. `b40f72fb5` put them back, but now on top of DST-1740 (#5765), which moved `relative` onto `SwitchButton` and removed the wrapper `<div>` around the track. So the track is a direct grid item rather than sitting in its own block, next to a label slot that did not exist when DST-1740 landed. Neither build has seen that combination: the 26 Aug one snapshotted the pre-DST-1740 Switch with the wrapper, main's snapshotted DST-1740 without the label slot.

Every labelled Switch is in scope, `settings` most of all, since its label sits in the wide first column and is the variant that actually wraps.

Please re-dispatch on the tip, either a top-level `/run-chromatic` comment or the same `workflow_dispatch` you used last time.

packages/components/src/Switch/Switch.stories.tsx
```diff
@@ -145,12 +147,10 @@ Basic.test(
...
 
-// There is no accessible way to reach the track. It is the direct child div of
-// the label; the thumb is nested inside it. Matching on `rounded-full` would
-// also match the thumb, and every assertion here reads `'none'` off a wrong
-// element just as happily as off the right one — so pin the shape instead.
 const getTrack = (switchEl: HTMLElement) => {
-  const track = switchEl.closest('label')?.querySelector(':scope > div');
+  const track = switchEl
+    .closest('label')
+    ?.querySelector('[aria-hidden="true"]');
```
sarahgm: `b40f72fb5` swapped `:scope > div` for `[aria-hidden="true"]` here, and added `aria-hidden` to the track in `Switch.tsx` to make it match. That is a component attribute added to serve a test selector, and it widens the search from a direct child to any descendant at any depth, so an aria-hidden node RAC might add earlier in the tree would win instead. The `28px` guard below still catches that, which is why this is not blocking, but `:scope > div` was already exact and asked nothing of the component.

If the `aria-hidden` is wanted on its own merits, and it does match what Checkbox and Radio do with their control graphic, keep it and add a line to the description, since it is an a11y change the PR body does not mention. Otherwise I would put the selector back.

packages/components/src/Badge/Badge.stories.tsx
OsamaAbdellateef: Fixed in `9a78e3071`. The thread got marked resolved earlier, but only `Badge.stories.tsx` had actually picked up the change — the other 8 pure-measurement tests you listed (`WithBadge`, `LongMultilineLabel`, `WrappingLabelWithBadge`, `OversizedBadge` across Checkbox, Radio and Switch stories) were still on `disableSnapshot: false`. All 8 are now `true`; the story-level baselines stay as the source of truth for these visual states, and the hover/indeterminate tests that genuinely need their own snapshot are untouched.

packages/components/src/Switch/Switch.stories.tsx
OsamaAbdellateef: Reverted in `9a78e3071` — removed `aria-hidden="true"` from the track div in `Switch.tsx` and moved the test off the attribute selector, as you suggested.

A literal `:scope > div` on its own doesn't hold up though: the label is now its own `div` (not the old `Label` `span`), and the `settings` variant renders that label div *before* the track, so `:scope > div` would grab the label instead. Went with `:scope > div:has(> div)` — the track is the only one of the two with a `div` as a direct child (the thumb); `LabelAdornment` wraps a badge in a `span`, so the label slot never has a direct `div` child in any of the current stories. Ran the full `Switch.stories.tsx` suite to confirm: 21/21 pass, including `LongMultilineLabel` (`settings` variant) and `WithBadge` (default variant), both of which go through `getTrack`.

themes/theme-rui/src/components/Switch.styles.ts
OsamaAbdellateef: Re-dispatched on the tip. `Visual-Regression-Tests` is running now:

https://github.com/marigold-ui/marigold/actions/runs/33865607934 — `workflow_dispatch`, head `9a78e3071`.

That commit is also where the two threads from this pass landed: the `aria-hidden` revert on the track (back to a structural selector, `:scope > div:has(> div)`, since a literal `:scope > div` breaks on the `settings` variant now that the label is its own div) and the `disableSnapshot: true` fix on the 8 measurement tests that were still snapshotting twice. So this run covers the full geometry list from your comment plus those two.

sarahgm (approved):

PR #5763, opened 2026-08-25
fix(DST-1673): keep example tables readable at narrow widths

sarahgm:
# Description

`/examples/users` kept all five columns below roughly 800px of viewport width instead of scrolling, squeezing them under their content until the name cell's text lay on top of the Title column.

Two gaps cause it, and they only bite together. A `Table.Column` without a `width` falls back to react-aria's default `minWidth` of 75px, and the examples shell still gives the sidebar 352px at a 768px viewport, leaving 401px for content. Five columns at 75px do not fit, so every column pins at that floor. The Name cell then holds an `Inline noWrap` (avatar, name, email) that cannot shrink, and because cells are `overflow: visible` it paints 146px past its own edge. Measured identically at 768px and 420px, and on production as well as the beta deployment named in the ticket.

Columns now carry widths (a fraction plus a floor for the flexible ones, a fixed width for the predictable ones) and the tables sit in `Scrollable`. The users table becomes 790px inside a 369px slot and scrolls, with every cell's content back inside its own box.

Auditing the rest turned up no example table that gets this right. Nine had no widths at all, and `bulk-actions/EventsTable` had widths but no scroll container, so 329px of it was clipped by `body { overflow-x: clip }` with no scrollbar at all. This PR covers the two pages named in the ticket plus the three worst of the remainder.

**One thing a reviewer should know.**

`filter/VenuesTable` gets fraction-only widths and no floors, because it is the one table here that gets no `Scrollable`. Its `actionBar` overlay renders *inside* `Table` as `sticky bottom-…`, so wrapping the component from outside would re-anchor that sticky to the wrapper. That matters for the widths, because a static `width` also becomes the column's `minWidth` (`TableColumn.tsx`, `ensureWidth`). Floors on eleven columns put the table's minimum at 1538px inside a 1289px container with nothing to scroll, which is 249px unreachable at 1440px where `main` fits all twelve. Fractions leave the floor at react-aria's 75px per column, exactly where `main` has it, so the table fills its container at 1440px with the name cells no longer overlapping, and narrow viewports clip no more than they already do.

`bulk-actions/EventsTable` has the same shape. Both wait on the same follow-up, whether `Table` should own horizontal scroll, with one caveat for whoever takes it. Putting `overflow-x-auto` on `ResizableTableContainer` makes `overflow-y` compute to `auto` as well, which would break `Table.Header sticky` inside a consumer's `Scrollable`.

**The second half of the ticket does not reproduce.** The dashboard at 1440x900 renders the two panels side by side at exactly 2:1 (662px and 331px, 48px gap), with nothing crossing the viewport edge, on the same Chromium and the same deployment the ticket names. Production renders it the same way, without this PR.

Two commits landed after the 2026-07-31 report and are plausible causes: `bb2318685` `feat(DST-1700): make --bleed-px the documented container spacing contract` (#5726, 2026-08-11), whose own message says it rewrote how `Table` derives edge padding, and `b7122c0b7` `chore: Marigold v18 changes` (#5304, 2026-08-04). Worth checking with the reporter against those two before treating that half as done. The dashboard's new `Scrollable` guards the symptom either way, because its own min-content is zero, so the activity table can no longer inflate its panel and wrap the pair before `collapseAt` asks for it.

Closes DST-1673

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

Measured on `/examples/users` at a 768px viewport, which leaves a 369px content area:

|                     | Before                     | After                        |
| ------------------- | -------------------------- | ---------------------------- |
| Column widths       | 75 / 75 / 75 / 75 / 75     | 280 / 170 / 110 / 110 / 120  |
| Table width         | 375px                      | 790px                        |
| Scrolls?            | no (`overflow-x: visible`) | yes                          |
| Worst cell overhang | **+146px**                 | none, inside its own cell    |

And on `/examples/filter` at 1440x900, which gives the table a 1289px container:

|                     | `main`  | This PR |
| ------------------- | ------- | ------- |
| Table minimum width | 861px   | 861px   |
| Table width         | 1289px  | 1289px  |
| Clipped             | 0px     | 0px     |

## Test Instructions

1. `pnpm build && pnpm start`, then open http://localhost:3000/examples/users
2. Set the viewport to 768px wide, then to 420px. The table scrolls horizontally inside its panel and no cell's text overlaps its neighbour.
3. Open `/examples` at 1440x900. "Recent activity" and "Your team" stay side by side at 2:1 with nothing running past the right edge.
4. Open `/examples/filter` at 1440x900. All twelve columns are reachable and the table fills its container, with the Name column wider than it is on `main`. At 768px it clips as much as `main` does, which is what the `Table` follow-up will fix.
5. Open `/examples/pick` at 768px and check the same as step 2 for those tables.
6. Open http://localhost:3000/components/collection/table and read the new "Column widths" section. Its demo scrolls horizontally when the container is narrower than the columns' floors.

## Breaking Changes

No

## Checklist

On the unticked items: this PR adds no stories and no unit tests. A narrow-container `Table` story with an overhang assertion was written and then dropped on request, so the new "Column widths" section on the Table docs page is what stands in its place: it records that a static `width` is also the column's floor, that the floors added together are the table's minimum, and that a table whose minimum can outgrow its container needs a `Scrollable`. Visual regression does not apply here. `visual-regression-tests.yml` carries `paths-ignore: ['docs/**', '.changeset/**']`, which is this entire diff.

### DIFF
--- a/.changeset/example-table-narrow-width.md
+++ b/.changeset/example-table-narrow-width.md
@@ -11,13 +11,17 @@ the sidebar 352px, leaving 401px for content, which pinned all five columns of
 146px over the Title column.
 
 Columns now carry widths (a fraction plus a floor for the flexible ones, a
-fixed width for the predictable ones), and the users, dashboard, report and
-picker tables sit in `Scrollable`, so they scroll instead of overlapping. On
-the dashboard that also stops the activity table inflating its panel, which is
-what could wrap the 2:1 pair before `collapseAt` asked for it.
+fixed width for the predictable ones), and the users, dashboard and report
+tables sit in `Scrollable`, so they scroll instead of overlapping. The picker
+dialog's table already scrolled and only gains the widths. On the dashboard
+that scroll container also stops the activity table inflating its panel, which
+is what could wrap the 2:1 pair before `collapseAt` asked for it.
 
-The filter table keeps the widths but no `Scrollable`: its `actionBar` overlay
-renders inside `Table` as `sticky bottom-…`, so wrapping the component from
-outside would re-anchor it. Its cells no longer overlap, but the columns past
-the container edge stay unreachable until `Table` can own that scroll
-container itself.
+The filter table gets fraction-only widths and no floors. It cannot have a
+`Scrollable` of its own yet, because its `actionBar` overlay renders inside
+`Table` as `sticky bottom-…` and an outer wrapper would re-anchor it. Since a
+static width also becomes that column's floor, floors on eleven columns would
+push the table past its container at 1440px with nothing to scroll. Fractions
+leave the floor where it already was, so the table keeps filling its container
+at desktop width, where the name cell no longer overlaps, and narrow viewports
+stay as they are until `Table` can own that scroll container itself.
--- a/.changeset/example-table-narrow-width.md
+++ b/.changeset/example-table-narrow-width.md
@@ -1,0 +1,23 @@
+---
+'@marigold/docs': patch
+---
+
+fix(DST-1673): keep example tables readable when they outgrow their container
+
+Example tables declared no column widths, so every column fell back to
+react-aria's 75px default. In the examples shell a 768px viewport still gives
+the sidebar 352px, leaving 401px for content, which pinned all five columns of
+`/examples/users` at that floor and let the nowrap name-and-email cell paint
+146px over the Title column.
+
+Columns now carry widths (a fraction plus a floor for the flexible ones, a
+fixed width for the predictable ones), and the users, dashboard, report and
+picker tables sit in `Scrollable`, so they scroll instead of overlapping. On
+the dashboard that also stops the activity table inflating its panel, which is
+what could wrap the 2:1 pair before `collapseAt` asked for it.
+
+The filter table keeps the widths but no `Scrollable`: its `actionBar` overlay
+renders inside `Table` as `sticky bottom-…`, so wrapping the component from
+outside would re-anchor it. Its cells no longer overlap, but the columns past
+the container edge stay unreachable until `Table` can own that scroll
+container itself.
--- a/docs/app/(examples)/examples/filter/VenuesTable.tsx
+++ b/docs/app/(examples)/examples/filter/VenuesTable.tsx
@@ -242,25 +242,48 @@ export const VenuesTable = () => {
           )}
         >
           <Table.Header>
-            <Table.Column id="name" rowHeader allowsSorting>
+            <Table.Column
+              id="name"
+              rowHeader
+              allowsSorting
+              width="2fr"
+              minWidth={200}
+            >
               Name
             </Table.Column>
-            <Table.Column id="type">Type</Table.Column>
-            <Table.Column id="address">Address</Table.Column>
-            <Table.Column id="capacity" alignX="right" allowsSorting>
+            <Table.Column id="type" width={130}>
+              Type
+            </Table.Column>
+            <Table.Column id="address" width="1fr" minWidth={190}>
+              Address
+            </Table.Column>
+            <Table.Column
+              id="capacity"
+              alignX="right"
+              allowsSorting
+              width={110}
+            >
               Capacity
             </Table.Column>
-            <Table.Column id="price" alignX="right" allowsSorting>
+            <Table.Column id="price" alignX="right" allowsSorting width={120}>
               Max. Price
             </Table.Column>
-            <Table.Column id="traits">Traits</Table.Column>
-            <Table.Column id="amenities">Amenities</Table.Column>
-            <Table.Column id="parking">Parking</Table.Column>
-            <Table.Column id="rating" alignX="right">
+            <Table.Column id="traits" width="1fr" minWidth={160}>
+              Traits
+            </Table.Column>
+            <Table.Column id="amenities" width="1fr" minWidth={160}>
+              Amenities
+            </Table.Column>
+            <Table.Column id="parking" width={140}>
+              Parking
+            </Table.Column>
+            <Table.Column id="rating" alignX="right" width={100}>
               Rating
             </Table.Column>
-            <Table.Column id="available">Available</Table.Column>
-            <Table.Column id="actions">
+            <Table.Column id="available" width={120}>
+              Available
+            </Table.Column>
+            <Table.Column id="actions" width={72}>
               <VisuallyHidden>Actions</VisuallyHidden>
             </Table.Column>
           </Table.Header>
--- a/docs/app/(examples)/examples/filter/VenuesTable.tsx
+++ b/docs/app/(examples)/examples/filter/VenuesTable.tsx
@@ -241,9 +241,6 @@ export const VenuesTable = () => {
             </ActionBar>
           )}
         >
-          {/* Fraction-only widths. A static width becomes the column's floor
-              too, and eleven floors do not fit this container, which has no
-              scroll container of its own to fall back on. */}
           <Table.Header>
             <Table.Column id="name" rowHeader allowsSorting width="2fr">
               Name
--- a/docs/app/(examples)/examples/filter/VenuesTable.tsx
+++ b/docs/app/(examples)/examples/filter/VenuesTable.tsx
@@ -241,49 +241,46 @@ export const VenuesTable = () => {
             </ActionBar>
           )}
         >
+          {/* Fraction-only widths. A static width becomes the column's floor
+              too, and eleven floors do not fit this container, which has no
+              scroll container of its own to fall back on. */}
           <Table.Header>
-            <Table.Column
-              id="name"
-              rowHeader
-              allowsSorting
-              width="2fr"
-              minWidth={200}
-            >
+            <Table.Column id="name" rowHeader allowsSorting width="2fr">
               Name
             </Table.Column>
-            <Table.Column id="type" width={130}>
+            <Table.Column id="type" width="1fr">
               Type
             </Table.Column>
-            <Table.Column id="address" width="1fr" minWidth={190}>
+            <Table.Column id="address" width="2fr">
               Address
             </Table.Column>
             <Table.Column
               id="capacity"
               alignX="right"
               allowsSorting
-              width={110}
+              width="1fr"
             >
               Capacity
             </Table.Column>
-            <Table.Column id="price" alignX="right" allowsSorting width={120}>
+            <Table.Column id="price" alignX="right" allowsSorting width="1fr">
               Max. Price
             </Table.Column>
-            <Table.Column id="traits" width="1fr" minWidth={160}>
+            <Table.Column id="traits" width="1.5fr">
               Traits
             </Table.Column>
-            <Table.Column id="amenities" width="1fr" minWidth={160}>
+            <Table.Column id="amenities" width="1.5fr">
               Amenities
             </Table.Column>
-            <Table.Column id="parking" width={140}>
+            <Table.Column id="parking" width="1fr">
               Parking
             </Table.Column>
-            <Table.Column id="rating" alignX="right" width={100}>
+            <Table.Column id="rating" alignX="right" width="1fr">
               Rating
             </Table.Column>
-            <Table.Column id="available" width={120}>
+            <Table.Column id="available" width="1fr">
               Available
             </Table.Column>
-            <Table.Column id="actions" width={72}>
+            <Table.Column id="actions" width="1fr">
               <VisuallyHidden>Actions</VisuallyHidden>
             </Table.Column>
           </Table.Header>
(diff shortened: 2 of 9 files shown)


docs/app/(examples)/examples/filter/VenuesTable.tsx
```diff
@@ -242,25 +242,48 @@ export const VenuesTable = () => {
           )}
         >
           <Table.Header>
-            <Table.Column id="name" rowHeader allowsSorting>
+            <Table.Column
+              id="name"
+              rowHeader
+              allowsSorting
+              width="2fr"
```
aromko: 🔴 **These widths make the table unreachable at 1440px, where `main` currently fits.**

Measured at 1440×900 on this branch's preview vs. production:

| | main | this PR |
|---|---|---|
| table width | 1289px | **1538px** |
| container | 1289px | 1289px |
| clipped | 0px | **249px** |

The declared widths sum to 1538px once RAC's `selectionMode="multiple"` checkbox column is counted. `body` computes to `overflow-x: clip`, so there is no scrollbar at any level: `Rating` is cut mid-column and `Available` plus the `Actions` row menu cannot be reached at all. On production every one of the 12 columns is reachable at 114px each.

I follow the reasoning for not wrapping this one in `Scrollable` (the `actionBar` overlay is `sticky bottom-…` inside `Table`, so an outer wrapper re-anchors it), but the consequence is that this page goes from correct to broken at the exact viewport DST-1673 uses as its reference, on a page the ticket calls out as article-linked.

Three ways out, in order of laziness:
1. Leave this file's widths off until `Table` can own horizontal scroll, and land it with the follow-up ticket. Narrow widths stay as broken as they are on `main` today (already 244px clipped at 768px), but nothing regresses.
2. Trim the widths so the sum lands at or under ~1289px, which keeps every column reachable at 1440px and still stops the cell overlap.
3. Fix the scroll container in `Table` as part of this PR.

docs/app/(examples)/examples/users/page.tsx
```diff
@@ -36,49 +37,62 @@ const UsersPage = () => (
...
-                      alt=""
-                      className="block size-8 shrink-0 rounded-full"
+        {/* Columns carry widths so they never fall back to the 75px default,
+            which squeezes them under their content and lets the name cell
+            paint over its neighbour. `Scrollable` then gives the table
+            somewhere to go when the sum no longer fits. */}
+        <Scrollable>
+          <Table aria-label="Members">
+            <Table.Header>
+              <Table.Column rowHeader width="2fr" minWidth={240}>
```
aromko: ⚠️ **`minWidth={240}` sits exactly on the content minimum, with no headroom.**

Measured at 768px, where this column resolves to exactly 240px: the longest email (`diana.martinez@example.com`, 26 chars at `fontSize="xs"`) ends **7px** short of the cell's padding edge. Roughly one character of slack.

Since the cell content is an `Inline noWrap`, it cannot shrink, and `overflow="wrap"` only applies `wrap-break-word`, which a nowrap flex row ignores. So a 28-character address restores exactly the overlap this PR fixes, and with no test on this the regression would be silent. Either give the floor some room, or let this one cell truncate:

```suggestion
              <Table.Column rowHeader width="2fr" minWidth={280}>
```

aromko (commented):
## Automated Code Review

**Recommendation: Request Changes.** The `/examples/users` half is correct and the measurements in the description hold up. Verified independently at 768px on this branch's preview: table 750px inside a 369px scroller, scrolls, worst cell overhang -10px (inside its own box). The blocker is `/examples/filter` regressing at 1440px, detailed inline.

### What I measured

Chromium, this branch's preview vs. production (`main`), 1440×900 and 768px.

| | main | this PR |
|---|---|---|
| `/examples/users` @768 | columns pinned at 75px, cells overlap | 750px table in a 369px scroller, scrolls, no overhang ✅ |
| `/examples` @1440 | 662 / 331, ratio 2.00, nothing past 1409 | identical, ratio 2.00 |
| `/examples/filter` @1440 | 1289px table in a 1289px container, all 12 columns reachable | 1538px table, **249px clipped, no scrollbar** ❌ |

### On the second half of the ticket

Your call that it does not reproduce is right: production at 1440×900 already renders the two panels side by side at exactly 2:1 (662 / 331), inside the content edge, without this PR.

But the stated reason is not: "No layout or token commit landed between the report and that branch's HEAD." Two did, both after the 2026-07-31 report:

- `bb2318685` `feat(DST-1700): make --bleed-px the documented container spacing contract` (#5726, 2026-08-11), whose own message says it rewrote how `Table` derives edge padding and which containers publish `--bleed-px`
- `b7122c0b7` `chore: Marigold v18 changes` (#5304, 2026-08-04)

Either is a plausible cause, which turns "I cannot point at a fix" into a specific candidate the reporter can confirm against. Worth updating the description, since "Closes DST-1673" currently closes a ticket on a half that reads as unexplained.

### Follow-ups worth filing before merge

DST-1673 has no issue links, so the two deferrals in the description exist only in this PR's prose:

1. **`Table` owning horizontal scroll**, with your caveat recorded: `overflow-x-auto` on `ResizableTableContainer` makes `overflow-y` compute to `auto` too, which breaks `Table.Header sticky` inside a consumer's `Scrollable`. This unblocks both `filter/VenuesTable` and `bulk-actions/EventsTable`.
2. **The five example tables still with no widths at all**: `analytics/page.tsx` (×2), `billing/page.tsx`, `inventory/UserTable.tsx`, `security/page.tsx`. All of them carry the exact defect the ticket reports, on the `/examples` pages the ticket describes as the area articles link to.

### One structural thought

The root cause lives in the component, not the call sites: RAC's 75px default column minimum, plus cells that never clip. `themes/theme-rui/src/components/Table.styles.ts` sets no `overflow` on `cell`, and `overflow="wrap"` only adds `wrap-break-word`, which a `noWrap` `Inline` ignores. Any consumer who omits widths reproduces this, and `docs/content/components/collection/table/index.mdx` currently says nothing about widths being needed or about horizontal scrolling.

Given the description notes that the narrow-container story with the overhang assertion was dropped, a short "Column widths and horizontal scrolling" section in that docs page is the only regression guard this change would have. It would also make the "Component documentation added/updated" checkbox accurate, since no file under `docs/content/` is touched.

### Verified fine

No `any`, the `as const` width map is correctly typed and catches key typos, Typecheck and all 27 checks pass. VRT genuinely does not apply: `visual-regression-tests.yml` carries `paths-ignore: ['docs/**', '.changeset/**']` and this diff is entirely those two. Accessibility holds: `Scrollable` needs no `tabindex` here because the RAC grid carries `tabindex="0"`, so arrow-key navigation scrolls the container and WCAG 2.1.1 is satisfied.

---

.changeset/example-table-narrow-width.md
```diff
@@ -0,0 +1,23 @@
...
+fix(DST-1673): keep example tables readable when they outgrow their container
+
+Example tables declared no column widths, so every column fell back to
+react-aria's 75px default. In the examples shell a 768px viewport still gives
+the sidebar 352px, leaving 401px for content, which pinned all five columns of
+`/examples/users` at that floor and let the nowrap name-and-email cell paint
+146px over the Title column.
+
+Columns now carry widths (a fraction plus a floor for the flexible ones, a
+fixed width for the predictable ones), and the users, dashboard, report and
```
aromko: 💡 Two accuracy notes on the wording.

"the users, dashboard, report and picker tables sit in `Scrollable`" reads as though this PR wrapped four tables. The picker dialog table already sat in a `Scrollable height="calc(100dvh - 22rem)"` before this change; only the widths are new there.

And the filter paragraph below describes the caveat as columns past the container edge being unreachable, which sounds like a narrow-viewport tradeoff. Measured at 1440×900 it is 249px of clipped table with no scrollbar, including the whole Actions column, where `main` fits all 12 columns exactly. Consumers reading this changeset should see that it applies at desktop width.

docs/app/(examples)/examples/filter/VenuesTable.tsx
sarahgm: Fixed in c575ed5, taking a stricter version of your option 2.

Trimming the widths to land under 1289px would have fixed 1440px but made narrow viewports worse than `main`, because the floors would still sit far above `main`'s 861px. So the widths are now fraction-only with no `minWidth` at all, which leaves the floor at react-aria's 75px per column, exactly where `main` has it.

Name `2fr`, Address `2fr`, Traits and Amenities `1.5fr`, `1fr` for the other seven. That is 14fr over the 1253px content pool, so 89.5px per fr with no column falling below its 75px floor and nothing to clamp:

| | `main` | this PR |
|---|---|---|
| floor (sum of minimums) | 861px | 861px |
| @1440, 1289px container | fills, 114px each | fills, Name 179px, Address 179px |
| @768 | 244px clipped | 244px clipped |

So it is better than `main` at desktop width and identical to it below, rather than trading one for the other. Worth recording why static widths were the problem: `TableColumn`'s `ensureWidth` promotes any non-fraction `width` to `minWidth`, so each of those seven fixed widths was also a floor. That is what put the minimum at 1538px, and it is now written up in the docs section below.

The remaining narrow-width clipping is DST-1742, with your `overflow-y` caveat recorded in it.

docs/app/(examples)/examples/users/page.tsx
sarahgm: Taken, `minWidth={280}` in c575ed5.

Confirmed your reading of the data: the two longest addresses in the fixture are both 26 characters (`diana.martinez@example.com`, `chip.patterson@example.com`), so 240px really was sitting on the content minimum. Truncation is not available for this cell, since the `Inline noWrap` ignores `wrap-break-word`, so the floor is the only lever. The table is in a `Scrollable`, so the higher floor only means it starts scrolling a little sooner.

.changeset/example-table-narrow-width.md
sarahgm: Both corrected in c575ed5.

The picker table is out of the `Scrollable` list, with a sentence saying it already scrolled and only gains the widths. And the filter paragraph now leads with the desktop case rather than describing it as a narrow-viewport tradeoff: it says a static width becomes that column's floor, that floors on eleven columns would push the table past its container at 1440px with nothing to scroll, and that fractions are what keep the floor where it already was.

sarahgm:
Thanks, all four points are addressed in c575ed5. The blocker and the two inline notes are answered in their threads. The rest:

**The two commits.** You are right and I was wrong to say nothing landed. Both are ancestors of this branch, and `bb2318685`'s own message says it rewrote how `Table` derives edge padding. The description now names both as candidates for the reporter to check against, instead of claiming there is nothing to point at.

I have left `Closes DST-1673` in place for now. If the reporter cannot recognise either commit for the dashboard half, that should become a plain reference and they can close it themselves.

**Follow-ups filed**, both linked to DST-1673:

- **DST-1742** `✨ Let Table own its horizontal scroll container`, carrying your `overflow-x-auto` / `overflow-y` caveat and covering `filter/VenuesTable` and `bulk-actions/EventsTable`
- **DST-1743** `🐛 Example tables without column widths squeeze their columns onto the 75px floor`, for the five you listed: `analytics/page.tsx` (×2), `billing/page.tsx`, `inventory/UserTable.tsx`, `security/page.tsx`

**The structural point, and the missing guard.** Agreed that the cause lives in the component rather than the call sites, and that with the narrow-container story dropped there was nothing to catch a regression. There is now a `### Column widths` section on the Table docs page plus a `table-column-widths` demo, recording the three things that were nowhere in the docs:

- a static `width` is also the column's floor, via `TableColumn`'s `ensureWidth`
- the floors added together are the table's minimum width
- a table whose minimum can outgrow its container needs a `Scrollable`, and one that cannot have a `Scrollable` needs to keep its floors low

That also makes the ticked "Component documentation added/updated" box honest, since the diff now touches `docs/content/`.

Not re-measured in a browser since the change: the reasoning above is arithmetic from `ensureWidth` and `TableHeader`'s 36px selection column, so worth a look at `/examples/filter` at 1440px and 768px on the new preview before you re-review.

docs/app/(examples)/examples/pick/ReportVenues.tsx
```diff
@@ -43,42 +44,64 @@ export const ReportVenues = () => {
...
-                    <Badge variant={statusVariant[venue.status]}>
-                      {venue.status}
-                    </Badge>
-                  </Table.Cell>
-                  <Table.Cell alignX="right">{venue.rate}</Table.Cell>
-                </Table.Row>
-              )}
-            </Table.Body>
-          </Table>
+          {/* Eleven columns never fit a narrow panel, so the table scrolls
```
aromko: also run /remove-dumb-comments

aromko (approved):

docs/app/(examples)/examples/pick/ReportVenues.tsx
sarahgm: shorten and removed comments

PR #5764, opened 2026-08-26
fix([DSTSUP-275]): format file sizes to reflect magnitude and locale

OsamaAbdellateef:
# Description

`<FileField>` rendered every selected file's size with a fixed megabyte divisor and two decimals (`(file.size / 1024 / 1024).toFixed(2)` + `" MB"`), so anything under ~5 kB read `0.00 MB`. For consumers importing CSVs — routinely 1–50 kB — the item description carried no information at all, and there was no way to override it from the outside since `<FileField>` owns the file list in internal state and renders the items itself.

Sizes now step through `B`, `kB`, `MB`, `GB`, `TB` and pick the unit that fits the file's magnitude, formatted through `Intl.NumberFormat` for the active locale. The step stays at 1024 (unchanged), so a given file keeps the number it had before — only its unit and trailing zeros change.

Closes DSTSUP-275

## Screenshots / Preview

No visual redesign — only the item description text changes.

| File | Before | After |
| ---- | ------ | ----- |
| 2,400-byte CSV | `0.00 MB` | `2.34 kB` |
| 0.5 MiB image | `0.50 MB` | `512 kB` |
| 2 MiB PDF | `2.00 MB` | `2 MB` |
| Same CSV, `de-DE` locale | `0.00 MB` | `2,34 kB` |

## Test Instructions

1. Open the `FileField` story in Storybook and upload a small file (a few KB, e.g. a CSV) alongside a multi-MB file.
2. Confirm the small file shows a `kB` size instead of `0.00 MB`, and the large file shows a clean `MB`/`GB` value without a fixed two-decimal MB suffix.
3. Wrap the story in an `I18nProvider` with `locale="de-DE"` and re-upload; confirm the number uses a comma decimal separator (e.g. `2,34 kB`) while the unit stays `kB`.
4. `pnpm test:unit -- FileField` and `pnpm test:sb -- FileField` to run the added/updated tests.

## Breaking Changes

No

## Checklist

### DIFF
--- a/.changeset/dstsup-275_file-size-units.md
+++ b/.changeset/dstsup-275_file-size-units.md
@@ -6,8 +6,8 @@ fix(DSTSUP-275): pick the file size unit from the file's magnitude
 
 `<FileField>` rendered every selected file's size with a fixed megabyte divisor and two decimals — `(file.size / 1024 / 1024).toFixed(2)` — so anything under ~5 kB read `0.00 MB`. For consumers importing CSVs, where files are routinely 1–50 kB, the item description carried no information at all, and there was no way to override it from the outside: `<FileField>` owns the file list in internal state and renders the items itself.
 
-Sizes now step through `B`, `kB`, `MB`, `GB` and `TB`, picking the unit that fits: a 2,400-byte CSV reads `2.34 kB`, a 2 MiB PDF reads `2 MB`, and a `0`-byte file reads `0 B`. Sizes past the top unit stay in `TB` rather than running off the end of the scale.
+Sizes now step through `B`, `kB`, `MB`, `GB` and `TB`, picking the unit that fits: a 2,400-byte CSV reads `2.4 kB`, a 2,000,000-byte PDF reads `2 MB`, and a `0`-byte file reads `0 B`. Sizes past the top unit stay in `TB` rather than running off the end of the scale.
 
-The step stays at 1024, which is what the field has always divided by, so a given file keeps the number it had before — only its unit and trailing zeros change (`0.50 MB` is now `512 kB`, `2.00 MB` is now `2 MB`).
+The step is 1000, not the 1024 the field used to divide by, because `kB`/`MB`/`GB`/`TB` are SI symbols and 1000 is their SI value. That matches Finder, GNOME Files and the browser download UIs a user has open next to the field. Numbers therefore shift slightly against the old output beyond the unit change (`0.50 MB` is now `524.29 kB` for the same file, `2.00 MB` is now `2.1 MB`).
 
-The number is run through `Intl.NumberFormat` for the active locale, so a German consumer gets `2,34 kB` next to the field's already-localized labels.
+The number is run through `Intl.NumberFormat` for the active locale, so a German consumer gets `2,4 kB` next to the field's already-localized labels.
--- a/.changeset/dstsup-275_file-size-units.md
+++ b/.changeset/dstsup-275_file-size-units.md
@@ -1,0 +1,13 @@
+---
+'@marigold/components': patch
+---
+
+fix(DSTSUP-275): pick the file size unit from the file's magnitude
+
+`<FileField>` rendered every selected file's size with a fixed megabyte divisor and two decimals — `(file.size / 1024 / 1024).toFixed(2)` — so anything under ~5 kB read `0.00 MB`. For consumers importing CSVs, where files are routinely 1–50 kB, the item description carried no information at all, and there was no way to override it from the outside: `<FileField>` owns the file list in internal state and renders the items itself.
+
+Sizes now step through `B`, `kB`, `MB`, `GB` and `TB`, picking the unit that fits: a 2,400-byte CSV reads `2.34 kB`, a 2 MiB PDF reads `2 MB`, and a `0`-byte file reads `0 B`. Sizes past the top unit stay in `TB` rather than running off the end of the scale.
+
+The step stays at 1024, which is what the field has always divided by, so a given file keeps the number it had before — only its unit and trailing zeros change (`0.50 MB` is now `512 kB`, `2.00 MB` is now `2 MB`).
+
+The number is run through `Intl.NumberFormat` for the active locale, so a German consumer gets `2,34 kB` next to the field's already-localized labels. The unit symbol is not localized: `style: 'unit'` spells bytes out in its short form (`340 byte`), which reads inconsistently next to the abbreviated `kB` a row above it in the same list, and `B`/`kB`/`MB` are the same in every locale Marigold ships messages for.
--- a/.changeset/dstsup-275_file-size-units.md
+++ b/.changeset/dstsup-275_file-size-units.md
@@ -10,4 +10,4 @@ Sizes now step through `B`, `kB`, `MB`, `GB` and `TB`, picking the unit that fit
 
 The step stays at 1024, which is what the field has always divided by, so a given file keeps the number it had before — only its unit and trailing zeros change (`0.50 MB` is now `512 kB`, `2.00 MB` is now `2 MB`).
 
-The number is run through `Intl.NumberFormat` for the active locale, so a German consumer gets `2,34 kB` next to the field's already-localized labels. The unit symbol is not localized: `style: 'unit'` spells bytes out in its short form (`340 byte`), which reads inconsistently next to the abbreviated `kB` a row above it in the same list, and `B`/`kB`/`MB` are the same in every locale Marigold ships messages for.
+The number is run through `Intl.NumberFormat` for the active locale, so a German consumer gets `2,34 kB` next to the field's already-localized labels.
--- a/.changeset/dstsup-275_file-size-units.md
+++ b/.changeset/dstsup-275_file-size-units.md
@@ -4,7 +4,7 @@
 
 fix(DSTSUP-275): pick the file size unit from the file's magnitude
 
-`<FileField>` rendered every selected file's size with a fixed megabyte divisor and two decimals — `(file.size / 1024 / 1024).toFixed(2)` — so anything under ~5 kB read `0.00 MB`. For consumers importing CSVs, where files are routinely 1–50 kB, the item description carried no information at all, and there was no way to override it from the outside: `<FileField>` owns the file list in internal state and renders the items itself.
+`<FileField>` rendered every selected file's size with a fixed megabyte divisor and two decimals, `(file.size / 1024 / 1024).toFixed(2)`, so anything under ~5 kB read `0.00 MB`. For consumers importing CSVs, where files are routinely 1–50 kB, the item description carried no information at all, and there was no way to override it from the outside: `<FileField>` owns the file list in internal state and renders the items itself.
 
 Sizes now step through `B`, `kB`, `MB`, `GB` and `TB`, picking the unit that fits: a 2,400-byte CSV reads `2.4 kB`, a 2,000,000-byte PDF reads `2 MB`, and a `0`-byte file reads `0 B`. Sizes past the top unit stay in `TB` rather than running off the end of the scale.
 
--- a/packages/components/src/FileField/FileField.stories.tsx
+++ b/packages/components/src/FileField/FileField.stories.tsx
@@ -189,9 +189,9 @@ UploadFile.test(
     const input = document.querySelector(
       'input[type="file"]'
     ) as HTMLInputElement;
-    const fileA = makeFile('abc.pdf', 'application/pdf', 2 * 1024 * 1024);
-    const fileB = makeFile('test.txt', 'text/plain', 5 * 1024 * 1024);
-    const fileC = makeFile('pic1.jpg', 'image/*', 0.5 * 1024 * 1024);
+    const fileA = makeFile('abc.pdf', 'application/pdf', 2_000_000);
+    const fileB = makeFile('test.txt', 'text/plain', 5_000_000);
+    const fileC = makeFile('pic1.jpg', 'image/*', 512_000);
     // Small enough that a fixed MB divisor rendered it as "0.00 MB" (DSTSUP-275).
     const fileD = makeFile('import.csv', 'text/csv', 2400);
 
@@ -206,7 +206,7 @@ UploadFile.test(
     await expect(canvas.getByText('2 MB')).toBeInTheDocument();
     await expect(canvas.getByText('5 MB')).toBeInTheDocument();
     await expect(canvas.getByText('512 kB')).toBeInTheDocument();
-    await expect(canvas.getByText('2.34 kB')).toBeInTheDocument();
+    await expect(canvas.getByText('2.4 kB')).toBeInTheDocument();
   }
 );
 
--- a/packages/components/src/FileField/FileField.stories.tsx
+++ b/packages/components/src/FileField/FileField.stories.tsx
@@ -185,17 +185,21 @@ UploadFile.test(
     const fileA = makeFile('abc.pdf', 'application/pdf', 2 * 1024 * 1024);
     const fileB = makeFile('test.txt', 'text/plain', 5 * 1024 * 1024);
     const fileC = makeFile('pic1.jpg', 'image/*', 0.5 * 1024 * 1024);
+    // Small enough that a fixed MB divisor rendered it as "0.00 MB" (DSTSUP-275).
+    const fileD = makeFile('import.csv', 'text/csv', 2400);
 
     // Act
-    await userEvent.upload(input, [fileA, fileB, fileC]);
+    await userEvent.upload(input, [fileA, fileB, fileC, fileD]);
 
     // Assert
     await expect(canvas.getByText('abc.pdf')).toBeInTheDocument();
     await expect(canvas.getByText('test.txt')).toBeInTheDocument();
     await expect(canvas.getByText('pic1.jpg')).toBeInTheDocument();
-    await expect(canvas.getByText('2.00 MB')).toBeInTheDocument();
-    await expect(canvas.getByText('5.00 MB')).toBeInTheDocument();
-    await expect(canvas.getByText('0.50 MB')).toBeInTheDocument();
+    await expect(canvas.getByText('import.csv')).toBeInTheDocument();
+    await expect(canvas.getByText('2 MB')).toBeInTheDocument();
+    await expect(canvas.getByText('5 MB')).toBeInTheDocument();
+    await expect(canvas.getByText('512 kB')).toBeInTheDocument();
+    await expect(canvas.getByText('2.34 kB')).toBeInTheDocument();
   }
 );
 
--- a/packages/components/src/FileField/FileField.stories.tsx
+++ b/packages/components/src/FileField/FileField.stories.tsx
@@ -176,6 +176,13 @@ UploadFile.test(
       label: 'Multifile Upload',
       multiple: true,
     },
+    decorators: [
+      Story => (
+        <I18nProvider locale="en-US">
+          <Story />
+        </I18nProvider>
+      ),
+    ],
   },
   async ({ canvas, userEvent }) => {
     // Arrange
--- a/packages/components/src/FileField/FileField.test.tsx
+++ b/packages/components/src/FileField/FileField.test.tsx
@@ -205,10 +205,10 @@ test('shows a small file with a unit that fits it, not "0.00 MB"', async () => {
 
   await user.upload(input, [
     makeFile('import.csv', 'text/csv', 2400),
-    makeFile('report.pdf', 'application/pdf', 2 * 1024 * 1024),
+    makeFile('report.pdf', 'application/pdf', 2_000_000),
   ]);
 
-  expect(screen.getByText('2.34 kB')).toBeInTheDocument();
+  expect(screen.getByText('2.4 kB')).toBeInTheDocument();
   expect(screen.getByText('2 MB')).toBeInTheDocument();
   expect(screen.queryByText('0.00 MB')).not.toBeInTheDocument();
 });
@@ -226,7 +226,7 @@ test('formats the file size for the active locale', async () => {
 
   await user.upload(input, [makeFile('import.csv', 'text/csv', 2400)]);
 
-  expect(screen.getByText('2,34 kB')).toBeInTheDocument();
+  expect(screen.getByText('2,4 kB')).toBeInTheDocument();
 });
 
 test('onBeforeRemove receives the file it is about to remove', async () => {
--- a/packages/components/src/FileField/FileField.test.tsx
+++ b/packages/components/src/FileField/FileField.test.tsx
@@ -194,7 +194,11 @@ test('names the remove button in German when the locale is de-DE', async () => {
 
 test('shows a small file with a unit that fits it, not "0.00 MB"', async () => {
   const user = userEvent.setup();
-  render(<UploadFile.Component multiple />);
+  render(
+    <I18nProvider locale="en-US">
+      <UploadFile.Component multiple />
+    </I18nProvider>
+  );
   const input = document.querySelector(
     'input[type="file"]'
   ) as HTMLInputElement;
--- a/packages/components/src/FileField/FileField.test.tsx
+++ b/packages/components/src/FileField/FileField.test.tsx
@@ -192,6 +192,39 @@ test('names the remove button in German when the locale is de-DE', async () => {
   ).toBeInTheDocument();
 });
 
+test('shows a small file with a unit that fits it, not "0.00 MB"', async () => {
+  const user = userEvent.setup();
+  render(<UploadFile.Component multiple />);
+  const input = document.querySelector(
+    'input[type="file"]'
+  ) as HTMLInputElement;
+
+  await user.upload(input, [
+    makeFile('import.csv', 'text/csv', 2400),
+    makeFile('report.pdf', 'application/pdf', 2 * 1024 * 1024),
+  ]);
+
+  expect(screen.getByText('2.34 kB')).toBeInTheDocument();
+  expect(screen.getByText('2 MB')).toBeInTheDocument();
+  expect(screen.queryByText('0.00 MB')).not.toBeInTheDocument();
+});
+
+test('formats the file size for the active locale', async () => {
+  const user = userEvent.setup();
+  render(
+    <I18nProvider locale="de-DE">
+      <UploadFile.Component multiple />
+    </I18nProvider>
+  );
+  const input = document.querySelector(
+    'input[type="file"]'
+  ) as HTMLInputElement;
+
+  await user.upload(input, [makeFile('import.csv', 'text/csv', 2400)]);
+
+  expect(screen.getByText('2,34 kB')).toBeInTheDocument();
+});
+
 test('onBeforeRemove receives the file it is about to remove', async () => {
   const user = userEvent.setup();
   const onBeforeRemove = vi.fn(() => true);
--- a/packages/components/src/FileField/FileField.tsx
+++ b/packages/components/src/FileField/FileField.tsx
@@ -1,7 +1,8 @@
 import { useEffect, useRef, useState } from 'react';
 import type RAC from 'react-aria-components';
 import { DropZone } from 'react-aria-components/DropZone';
-import { useLocale, useLocalizedStringFormatter } from '@react-aria/i18n';
+import { useLocale } from 'react-aria-components/I18nProvider';
+import { useLocalizedStringFormatter } from '@react-aria/i18n';
 import { WidthProp, cn, useClassNames } from '@marigold/system';
 import { FieldBase, type FieldBaseProps } from '../FieldBase/FieldBase';
 import { intlMessages } from '../intl/messages';
--- a/packages/components/src/FileField/FileField.tsx
+++ b/packages/components/src/FileField/FileField.tsx
@@ -1,14 +1,17 @@
 import { useEffect, useRef, useState } from 'react';
 import type RAC from 'react-aria-components';
 import { DropZone } from 'react-aria-components/DropZone';
-import { useLocale } from 'react-aria-components/I18nProvider';
-import { useLocalizedStringFormatter } from '@react-aria/i18n';
+import {
+  useLocalizedStringFormatter,
+  useNumberFormatter,
+} from '@react-aria/i18n';
 import { WidthProp, cn, useClassNames } from '@marigold/system';
 import { FieldBase, type FieldBaseProps } from '../FieldBase/FieldBase';
 import { intlMessages } from '../intl/messages';
 import { FileFieldItem } from './FileFieldItem';
 import { FileTrigger } from './FileTrigger';
 import {
+  FILE_SIZE_FORMAT_OPTIONS,
   fileKey,
   formatFileSize,
   isFileDropItem,
@@ -81,7 +84,7 @@ export const FileField = ({
   const [files, setFiles] = useState<File[] | null>(null);
   const hiddenInputRef = useRef<HTMLInputElement>(null);
   const stringFormatter = useLocalizedStringFormatter(intlMessages);
-  const { locale } = useLocale();
+  const sizeFormatter = useNumberFormatter(FILE_SIZE_FORMAT_OPTIONS);
   const dropZoneLabel = stringFormatter.format('dropZoneLabel');
   const buttonLabel = stringFormatter.format('uploadLabel');
 
@@ -207,7 +210,7 @@ export const FileField = ({
                 classNames.itemDescription
               )}
             >
-              {formatFileSize(file.size, locale)}
+              {formatFileSize(file.size, sizeFormatter)}
             </div>
           </FileField.Item>
         ))}
--- a/packages/components/src/FileField/FileField.tsx
+++ b/packages/components/src/FileField/FileField.tsx
@@ -1,13 +1,18 @@
 import { useEffect, useRef, useState } from 'react';
 import type RAC from 'react-aria-components';
 import { DropZone } from 'react-aria-components/DropZone';
-import { useLocalizedStringFormatter } from '@react-aria/i18n';
+import { useLocale, useLocalizedStringFormatter } from '@react-aria/i18n';
 import { WidthProp, cn, useClassNames } from '@marigold/system';
 import { FieldBase, type FieldBaseProps } from '../FieldBase/FieldBase';
 import { intlMessages } from '../intl/messages';
 import { FileFieldItem } from './FileFieldItem';
 import { FileTrigger } from './FileTrigger';
-import { fileKey, isFileDropItem, normalizeAndLimitFiles } from './fileUtils';
+import {
+  fileKey,
+  formatFileSize,
+  isFileDropItem,
+  normalizeAndLimitFiles,
+} from './fileUtils';
 
 type RemovedProps =
   'className' | 'style' | 'children' | 'isDisabled' | 'isRequired';
@@ -75,6 +80,7 @@ export const FileField = ({
   const [files, setFiles] = useState<File[] | null>(null);
   const hiddenInputRef = useRef<HTMLInputElement>(null);
   const stringFormatter = useLocalizedStringFormatter(intlMessages);
+  const { locale } = useLocale();
   const dropZoneLabel = stringFormatter.format('dropZoneLabel');
   const buttonLabel = stringFormatter.format('uploadLabel');
 
@@ -200,7 +206,7 @@ export const FileField = ({
                 classNames.itemDescription
               )}
             >
-              {(file.size / 1024 / 1024).toFixed(2)} MB
+              {formatFileSize(file.size, locale)}
             </div>
           </FileField.Item>
         ))}
--- a/packages/components/src/FileField/fileUtils.ts
+++ b/packages/components/src/FileField/fileUtils.ts
@@ -56,6 +56,8 @@ const matchesAcceptedToken = (file: File, token: string): boolean => {
   return fileType === t;
 };
 
+// Identity of a file for de-duplication and removal: two files with the same
+// name, size, and last-modified time are treated as the same file.
 export const fileKey = (file: File): string =>
   `${file.name}:${file.size}:${file.lastModified}`;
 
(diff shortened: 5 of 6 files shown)


packages/components/src/FileField/FileField.tsx
```diff
@@ -200,7 +207,7 @@ export const FileField = ({
                 classNames.itemDescription
               )}
             >
-              {(file.size / 1024 / 1024).toFixed(2)} MB
+              {formatFileSize(file.size, locale)}
```
jim761: 🔴 **Visual Regression Tests — not run**

No `Visual-Regression-Tests` run exists for this branch, and this line changes the rendered text for *every* file (`0.50 MB` → `512 kB`, `2.00 MB` → `2 MB`). Chromatic will show diffs that someone should accept deliberately.

Please trigger it by commenting `/run-chromatic` on this PR. The "Visual regression tests updated" checklist item is also still unticked.

packages/components/src/FileField/fileUtils.ts
```diff
@@ -56,6 +56,37 @@ const matchesAcceptedToken = (file: File, token: string): boolean => {
...
 };
 
+// Sizes step by 1024, which is what the field has always divided by. The
+// symbols stay the same in every locale Marigold ships messages for, so only
+// the number goes through `Intl.NumberFormat` - that keeps `2,34 kB` in de-DE
+// next to `2.34 kB` in en-US while a list of files stays on one set of units.
+// `style: 'unit'` was the alternative, but its short form spells bytes out
+// ("340 byte"), which reads inconsistently next to the abbreviated `kB` above
+// it in the same list.
+const FILE_SIZE_UNITS = ['B', 'kB', 'MB', 'GB', 'TB'] as const;
```
jim761, line 72: ⚠️ **SI symbols on a binary ladder**

`kB`/`MB`/`GB`/`TB` are 1000-based by SI/IEC, but `FILE_SIZE_STEP` is 1024 — so a 524,288-byte file reads `512 kB` when it is really 524 kB, and the error compounds ~2.4% per step (`1.5 GB` is actually 1.61 GB). Every OS file manager the user compares against reports the 1000-based number.

The trade-off *is* documented in the changeset and the ticket's own snippet uses 1024, so this is a decision rather than an oversight — but the two coherent options are `KiB/MiB/GiB/TiB` with the 1024 step, or SI symbols with a 1000 step. The current pairing is the one combination that's wrong in both readings.

My suggestion is `KiB/MiB/…`, since it preserves the "a given file keeps the number it had before" property the changeset is protecting:

```suggestion
const FILE_SIZE_UNITS = ['B', 'KiB', 'MiB', 'GiB', 'TiB'] as const;
```

Changing user-facing unit symbols twice is worse than getting them right once.

packages/components/src/FileField/fileUtils.ts
```diff
@@ -56,6 +56,37 @@ const matchesAcceptedToken = (file: File, token: string): boolean => {
...
+ * Formats a file size with the unit that fits its magnitude, so a 2,400-byte
+ * CSV reads as `2.34 kB` instead of rounding away to `0.00 MB`.
+ */
+export const formatFileSize = (size: number, locale: string): string => {
+  const bytes = Number.isFinite(size) && size > 0 ? size : 0;
+  const exponent =
+    bytes === 0
+      ? 0
+      : Math.min(
+          Math.floor(Math.log(bytes) / Math.log(FILE_SIZE_STEP)),
```
jim761: 💡 **The exponent is chosen before rounding, so `1,024 kB` is reachable**

`maximumFractionDigits: 2` can round the value up to `1024` after the unit has already been fixed. Measured windows:

| Range (bytes) | Renders as | Window |
| --- | --- | --- |
| `1048571..1048575` | `1,024 kB` | 5 bytes |
| `1073736582..1073741823` | `1,024 MB` | ~5.2 kB |
| just below `1024 ** 4` | `1,024 GB` | ~5.4 MB |

Very unlikely to be hit, but a ladder shouldn't be able to print its own step. Bump the exponent when the rounded value reaches `FILE_SIZE_STEP` and a larger unit exists, and add `[1024 ** 2 - 1, '1 MB']` to the `it.each` table in `fileUtils.test.ts` to pin it.

packages/components/src/FileField/fileUtils.ts
```diff
@@ -56,6 +56,37 @@ const matchesAcceptedToken = (file: File, token: string): boolean => {
...
+          FILE_SIZE_UNITS.length - 1
+        );
+
+  const value = new Intl.NumberFormat(locale, {
```
jim761: 💡 **A new `Intl.NumberFormat` per file per render**

`formatFileSize` is called inside the `files.map()` in `FileField.tsx:210`, so a 20-file selection constructs 20 formatters on every render. A module-level `Map<string, Intl.NumberFormat>` keyed on locale fixes it with no signature change and no new import:

```ts
const formatters = new Map<string, Intl.NumberFormat>();
const formatterFor = (locale: string) => {
  let formatter = formatters.get(locale);
  if (!formatter) {
    formatter = new Intl.NumberFormat(locale, { maximumFractionDigits: 2 });
    formatters.set(locale, formatter);
  }
  return formatter;
};
```

Small absolute cost — Intl constructors are just the expensive kind of cheap.

jim761 (commented):
## Automated Code Review

Spec match against **DSTSUP-275** is good: unit-aware sizes via the ticket's own `Math.log(size)/Math.log(1024)` approach, plus a top-unit clamp and non-finite/negative guards the ticket didn't ask for. The one deviation — literal unit symbols instead of `Intl.NumberFormat` `style: 'unit'` — is documented in the changeset and I agree with it (`340 byte` next to `kB` in the same list reads badly).

Also correct and worth noting: `useLocale` comes from `react-aria-components/I18nProvider`, not the `@react-aria/i18n` shell — exactly what `check-rac-first-imports.mjs` guards — and pure-function cases live in `fileUtils.test.ts` rather than the component test.

All 28 CI checks pass. Four inline comments: one blocker (visual regression), one decision about the unit symbols, and two minor ones.

---

packages/components/src/FileField/fileUtils.ts
OsamaAbdellateef: Fixed — the exponent now bumps to the next unit when rounding to 2 decimals would print the current unit's own step (e.g. `1,024 kB`). Added the suggested `[1024 ** 2 - 1, '1 MB']` case to `fileUtils.test.ts`. See `faf191f81`.

OsamaAbdellateef: Fixed — the exponent now bumps to the next unit when rounding to 2 decimals would print the current unit's own step (e.g. `1,024 kB`). Added the suggested `[1024 ** 2 - 1, '1 MB']` case to `fileUtils.test.ts`. See `faf191f81`.

packages/components/src/FileField/fileUtils.ts
OsamaAbdellateef: Fixed — `Intl.NumberFormat` instances are now memoized in a module-level `Map` keyed by locale, so a re-render with the same locale reuses the formatter instead of constructing a new one per file. Same commit as above (`faf191f81`).

packages/components/src/FileField/fileUtils.ts
OsamaAbdellateef: Good catch on the ambiguity — `kB` is 1000-based by SI/IEC while the step here is 1024, so strictly speaking `KiB/MiB/…` is the more precise pairing. I'm keeping `kB/MB/GB/TB` for this PR: it's the label every OS file manager and most consumer software already shows, so it matches what users compare against day to day, even though it's the less exact reading. Noted as a deliberate trade-off, not an oversight — open to revisiting if the team prefers strict precision here.

jim761 (commented):
## Automated Code Review

Reviewed against `CLAUDE.md`, DSTSUP-275, and the repo's own CI guards. Ran `vitest run --project=unit-tests FileField fileUtils` (**54 passed**) and probed `formatFileSize` directly across 19 boundary values in `en-US` / `de-DE` / `ar-EG`.

**Verdict: non-blocking.** Careful work — the fix matches the ticket, the rounding carry at unit boundaries is handled, and coverage is layered sensibly across the pure util, the component, and the browser. Seven notes, all nits, left inline on the relevant lines.

### Spec match

DSTSUP-275 asked for "a unit-aware size, e.g. `2.35 kB` / `1.2 MB` / `340 B`, picking the unit from the magnitude", localized via `Intl.NumberFormat`. Delivered exactly that. The ticket's `style: 'unit'` suggestion was deliberately declined and the changeset says why. The ticket's "Related, possibly a separate issue" (no `onSelect` on `FileField`) is correctly left out of scope.

### What's good

- **The i18n import path dodges a real bug class.** `useNumberFormatter` from `@react-aria/i18n` would have been shorter, but it reads the *shell's* `I18nContext` — the DSTSUP-261 / DST-1505 duplicate-context bug that `scripts/check-rac-first-imports.mjs` exists to prevent. Taking `useLocale` from `react-aria-components/I18nProvider` and building `Intl.NumberFormat` by hand is the right call; worth keeping if this is ever refactored.
- **The carry at unit boundaries is handled.** `1048575` renders `1 MB`, not `1024 kB` — the `Math.round(scaled * 100) / 100 >= FILE_SIZE_STEP` check, with `[1024 ** 2 - 1, '1 MB']` pinning it.
- Module-level `Map<string, Intl.NumberFormat>` is the correct caching shape for a formatter.
- No `any`, no `@ts-ignore`, `as const` on the unit tuple, explicit return types. Hook called unconditionally; no new state or effects; nothing for `memo`/`useMemo` to earn here.

### One note that has no line to sit on

**PR title uses `fix([DSTSUP-275]):` instead of `fix(DSTSUP-275):`.** The repo's Conventional-Commits scope carries the bare ticket key — `fix(DST-1675):`, `docs(DST-1703):` in recent history. The brackets land in `main`'s history verbatim on squash.

### Visual regression

Chromatic build [#244](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=244) ran on head `faf191f8` and passed, reporting **3 changes accepted as baselines**. Since the item description text is the only thing that changed, three diffs is the expected count — but the "Visual regression tests updated" checkbox is unticked, so worth confirming those three were the FileField stories and that accepting them was intentional.

---

packages/components/src/FileField/fileUtils.ts
```diff
@@ -56,8 +56,47 @@ const matchesAcceptedToken = (file: File, token: string): boolean => {
...
+    exponent += 1;
+    scaled = bytes / FILE_SIZE_STEP ** exponent;
+  }
+
+  const value = fileSizeFormatterFor(locale).format(scaled);
+
+  return `${value} ${FILE_SIZE_UNITS[exponent]}`;
+};
+
 export const fileKey = (file: File): string =>
```
jim761: ⚠️ **The doc comment on `fileKey` was dropped.**

The new size block was inserted where `fileKey`'s explanation used to live, and took it with it:

```diff
-// Identity of a file for de-duplication and removal: two files with the same
-// name, size, and last-modified time are treated as the same file.
```

`fileKey` is now the only undocumented export in this file, and it's the one whose contract (name + size + lastModified = same file) is least guessable from the body. Worth restoring those two lines.

Putting the size helpers *after* `fileKey` / `dedupeFiles` rather than between the accept-token helpers and `fileKey` would also keep the accept-matching code contiguous.

packages/components/src/FileField/fileUtils.ts
```diff
@@ -56,8 +56,47 @@ const matchesAcceptedToken = (file: File, token: string): boolean => {
...
+      ? 0
+      : Math.min(
+          Math.floor(Math.log(bytes) / Math.log(FILE_SIZE_STEP)),
+          FILE_SIZE_UNITS.length - 1
+        );
+
+  let scaled = bytes / FILE_SIZE_STEP ** exponent;
+  if (
+    exponent < FILE_SIZE_UNITS.length - 1 &&
+    Math.round(scaled * 100) / 100 >= FILE_SIZE_STEP
```
jim761: 💡 **The magic `100` here silently mirrors `maximumFractionDigits: 2` on line 66.**

The two have to stay in lockstep, 30 lines apart, with nothing connecting them — change the formatter to 1 or 3 digits and this boundary carry drifts without a single test noticing (the `[1024 ** 2 - 1, '1 MB']` case is exactly what would break).

```ts
const FILE_SIZE_FRACTION_DIGITS = 2;
const FILE_SIZE_ROUNDING = 10 ** FILE_SIZE_FRACTION_DIGITS;
```

packages/components/src/FileField/fileUtils.ts
```diff
@@ -56,8 +56,47 @@ const matchesAcceptedToken = (file: File, token: string): boolean => {
...
+ * Formats a file size with the unit that fits its magnitude, so a 2,400-byte
+ * CSV reads as `2.34 kB` instead of rounding away to `0.00 MB`.
+ */
+export const formatFileSize = (size: number, locale: string): string => {
+  const bytes = Number.isFinite(size) && size > 0 ? size : 0;
+  let exponent =
+    bytes === 0
+      ? 0
+      : Math.min(
+          Math.floor(Math.log(bytes) / Math.log(FILE_SIZE_STEP)),
```
jim761: 💡 **Optional: make the input guard total.**

Line 77 already absorbs `-1`, `NaN` and `Infinity`, and `fileUtils.test.ts:172` tests all three. A size in `(0, 1)` slips past it — `Math.floor(Math.log(0.4) / Math.log(1024))` is `-1`, `Math.min` keeps it, and `FILE_SIZE_UNITS[-1]` renders as the literal string `undefined`:

```
formatFileSize(0.4, 'en-US')  // → "409.6 undefined"
```

**Unreachable today**, to be clear: `File.size` is spec'd as an integer, and `formatFileSize` isn't re-exported from the package index — one production caller, `FileField.tsx:210`, passing `file.size`. So this is only worth doing if you'd rather the guard cover the whole domain than depend on that caller staying the only one.

```suggestion
          Math.max(Math.floor(Math.log(bytes) / Math.log(FILE_SIZE_STEP)), 0),
```

packages/components/src/FileField/fileUtils.ts
```diff
@@ -56,8 +56,47 @@ const matchesAcceptedToken = (file: File, token: string): boolean => {
   return fileType === t;
 };
 
-// Identity of a file for de-duplication and removal: two files with the same
-// name, size, and last-modified time are treated as the same file.
+const FILE_SIZE_UNITS = ['B', 'kB', 'MB', 'GB', 'TB'] as const;
```
jim761, line 72: 💡 **SI prefixes with a binary step.**

`2400` bytes renders `2.34 kB`, but 2.34 is the *kibibyte* count — 2400 B is `2.4 kB` in SI, or `2.34 KiB` in binary.

The changeset is upfront that 1024 was kept so numbers don't shift for existing users, and the ticket's own suggested implementation did the same, so this is a knowingly-taken trade rather than a slip. Flagging it only so it stays a team decision rather than quietly becoming precedent for the next component that formats bytes: either label `KiB`/`MiB`, or move the step to 1000 and accept the number change. Fine as-is.

packages/components/src/FileField/FileField.test.tsx
```diff
@@ -192,6 +192,39 @@ test('names the remove button in German when the locale is de-DE', async () => {
...
+  const input = document.querySelector(
+    'input[type="file"]'
+  ) as HTMLInputElement;
+
+  await user.upload(input, [
+    makeFile('import.csv', 'text/csv', 2400),
+    makeFile('report.pdf', 'application/pdf', 2 * 1024 * 1024),
+  ]);
+
+  expect(screen.getByText('2.34 kB')).toBeInTheDocument();
```
jim761: 💡 **This assertion depends on the machine's locale.**

There's no `I18nProvider` here, so `2.34 kB` is asserted against whatever `useLocale` resolves from the environment: green in CI (en-US), red for a developer on a German-locale machine, who gets `2,34 kB`. The de-DE test right below is correctly wrapped.

The file already carries this assumption in its older label tests, so it's pre-existing rather than introduced by this PR — but wrapping the two new assertions in `<I18nProvider locale="en-US">` would make them say what they mean, and would state the contrast with the de-DE test explicitly.

packages/components/src/FileField/FileField.stories.tsx
```diff
@@ -185,17 +185,21 @@ UploadFile.test(
...
     await expect(canvas.getByText('test.txt')).toBeInTheDocument();
     await expect(canvas.getByText('pic1.jpg')).toBeInTheDocument();
-    await expect(canvas.getByText('2.00 MB')).toBeInTheDocument();
-    await expect(canvas.getByText('5.00 MB')).toBeInTheDocument();
-    await expect(canvas.getByText('0.50 MB')).toBeInTheDocument();
+    await expect(canvas.getByText('import.csv')).toBeInTheDocument();
+    await expect(canvas.getByText('2 MB')).toBeInTheDocument();
+    await expect(canvas.getByText('5 MB')).toBeInTheDocument();
+    await expect(canvas.getByText('512 kB')).toBeInTheDocument();
+    await expect(canvas.getByText('2.34 kB')).toBeInTheDocument();
```
jim761: 💡 Same ambient-locale coupling as `FileField.test.tsx:207` — `2.34 kB`, `512 kB` and the two `MB` assertions all pin en-US number formatting without an `I18nProvider` to guarantee it.

Slightly more consequential here than in the unit test, since `pnpm test:sb` runs in a real browser and picks up the developer's actual system locale.

.changeset/dstsup-275_file-size-units.md
```diff
@@ -0,0 +1,13 @@
...
+
+Sizes now step through `B`, `kB`, `MB`, `GB` and `TB`, picking the unit that fits: a 2,400-byte CSV reads `2.34 kB`, a 2 MiB PDF reads `2 MB`, and a `0`-byte file reads `0 B`. Sizes past the top unit stay in `TB` rather than running off the end of the scale.
+
+The step stays at 1024, which is what the field has always divided by, so a given file keeps the number it had before — only its unit and trailing zeros change (`0.50 MB` is now `512 kB`, `2.00 MB` is now `2 MB`).
+
+The number is run through `Intl.NumberFormat` for the active locale, so a German consumer gets `2,34 kB` next to the field's already-localized labels. The unit symbol is not localized: `style: 'unit'` spells bytes out in its short form (`340 byte`), which reads inconsistently next to the abbreviated `kB` a row above it in the same list, and `B`/`kB`/`MB` are the same in every locale Marigold ships messages for.
```
jim761: 💡 **This paragraph is implementation rationale, not a changelog entry.**

Why `style: 'unit'` was rejected is genuinely useful and shouldn't be lost — but consumers reading release notes aren't its audience. It belongs in a comment next to `fileSizeFormatterFor`, where the next person to reach for `style: 'unit'` will actually find it.

Paragraphs 1–2 plus the "same number, different unit" note are the changelog; four paragraphs is a lot for a patch.

packages/components/src/FileField/fileUtils.ts
OsamaAbdellateef: Fixed — restored `fileKey`'s identity comment and moved the size helpers after `fileKey`/`dedupeFiles` so the accept-matching code stays contiguous. The reorder landed in `3639c794b`; the comment itself got dropped from that push by mistake and is now back in `6db08bfef`.

packages/components/src/FileField/fileUtils.ts
OsamaAbdellateef: Fixed — extracted `FILE_SIZE_FRACTION_DIGITS` and `FILE_SIZE_ROUNDING` so the formatter's `maximumFractionDigits` and the boundary-rounding check read from the same constant instead of drifting independently. See `3639c794b`.

packages/components/src/FileField/fileUtils.ts
OsamaAbdellateef: Fixed — added the `Math.max(..., 0)` guard so the exponent can't go negative for a size in `(0, 1)`, making the domain guard total rather than relying on `File.size` staying the only caller. See `3639c794b`.

packages/components/src/FileField/fileUtils.ts
OsamaAbdellateef: Agreed, no change made — same trade-off as the `kB`/binary-step decision above, so leaving it as-is per your note.

packages/components/src/FileField/FileField.test.tsx
OsamaAbdellateef: Fixed — wrapped that render in `<I18nProvider locale="en-US">` so the assertion states its own locale instead of depending on the machine's. See `3639c794b`.

packages/components/src/FileField/FileField.stories.tsx
OsamaAbdellateef: Fixed — rather than hardcoding a locale into the shared `UploadFile` story's `render` (which would have shadowed the `de-DE` wrapper other tests apply around `UploadFile.Component`), scoped an `en-US` decorator to just this `.test()` call. See `3639c794b`.

.changeset/dstsup-275_file-size-units.md
OsamaAbdellateef: Fixed — trimmed the changeset to the changelog-relevant paragraphs and moved the `style: 'unit'` rationale into a comment next to `fileSizeFormatterFor` in `fileUtils.ts`. Changeset trim landed in `3639c794b`; the comment is now in `6db08bfef`.

packages/components/src/FileField/fileUtils.ts
```diff
@@ -71,6 +71,56 @@ const dedupeFiles = (files: File[]): File[] => {
   });
 };
 
+const FILE_SIZE_UNITS = ['B', 'kB', 'MB', 'GB', 'TB'] as const;
+const FILE_SIZE_STEP = 1024;
```
aromko: ⚠️ Picking this up because you left it to the team ("open to revisiting if the team prefers strict precision here"), so here is a decision rather than another flag.

I do not think the rationale holds. The argument for `kB` + 1024 was that it is "the label every OS file manager and most consumer software already shows". That describes Windows Explorer, which is binary — and which writes `KB`, not `kB`. The other places a user would compare against are decimal:

- macOS Finder has been base-1000 since Snow Leopard, and writes `kB`
- GNOME Files is base-1000 by default, and writes `kB`
- Chrome and Firefox download UIs are base-1000

So the current pairing does not match either camp. It borrows the SI symbol from the decimal group and the divisor from the binary one, and the result is that a macOS user looking at the same 2,400-byte CSV sees `2.4 kB` in Finder and `2.34 kB` in this field. That is a worse outcome than either convention on its own, and it is the specific failure mode the "matches what users compare against" argument was trying to avoid.

My preference, in order:

1. **Move the step to 1000, keep `kB`/`MB`/`GB`/`TB`.** Correct by SI, matches Finder, GNOME and both browsers, and needs one constant changed.
2. `KiB`/`MiB`/`GiB`/`TiB` with the step at 1024. Also correct, less familiar.

The one argument for the status quo is the changeset's "a given file keeps the number it had before" — but that is already not true, and the changeset says so itself two lines later (`0.50 MB` is now `512 kB`). The numbers are changing in this PR regardless; this is the cheapest moment to land on a defensible pair.

Worth settling properly because `grep -rn "1024"` across `packages/` and `docs/` finds no other byte formatting anywhere in the repo. Whatever ships here is the precedent, exactly as `jim761` predicted.

packages/components/src/FileField/fileUtils.ts
```diff
@@ -71,6 +71,56 @@ const dedupeFiles = (files: File[]): File[] => {
...
+const FILE_SIZE_FRACTION_DIGITS = 2;
+const FILE_SIZE_ROUNDING = 10 ** FILE_SIZE_FRACTION_DIGITS;
+
+const fileSizeFormatters = new Map<string, Intl.NumberFormat>();
```
aromko: 💡 This came out of `jim761`'s "a new `Intl.NumberFormat` per file per render", which was a fair catch, but there is a smaller answer than a hand-rolled cache.

`@react-aria/i18n` exports `useNumberFormatter`, which memoizes formatters internally and reads the locale straight from context. This repo already uses it, in `packages/system/src/components/Formatters/NumericFormat.tsx:39`, so it is the established idiom rather than a new dependency.

Passing the formatter in keeps `formatFileSize` a pure, directly testable function while deleting `fileSizeFormatters`, `fileSizeFormatterFor` and the `useLocale()` import:

```ts
export const formatFileSize = (size: number, formatter: Intl.NumberFormat): string => {
```

and in the component:

```ts
const sizeFormatter = useNumberFormatter({ maximumFractionDigits: 2 });
```

The `fileUtils` tests then build a formatter per case instead of passing a locale string, which is barely more code and drops an unbounded module-level `Map` from the bundle. Entirely optional, the current version is correct.

packages/components/src/FileField/fileUtils.ts
OsamaAbdellateef: Taken option 1 — `FILE_SIZE_STEP` is now 1000 with `kB`/`MB`/`GB`/`TB` kept. You're right that the "matches what users compare against" argument pointed at the decimal camp all along, and that the "keeps the number it had before" property was already spent by this PR, so there was nothing left holding the pairing together.

What moved with it:

- `formatFileSize` docblock and the constants comment now say why the step is 1000 (SI symbols, Finder/GNOME/browser download UIs)
- `fileUtils.test.ts` cases re-derived on the decimal ladder: `2400 → 2.4 kB`, `999 B`/`1 kB` at the first boundary, and the rounding-boundary case is now `[1000 ** 2 - 1, '1 MB']`
- `FileField.test.tsx` and the `UploadFile` story test use decimal fixture sizes (`2_000_000`, `512_000`) so the expected strings stay clean; de-DE now asserts `2,4 kB`
- the changeset's "step stays at 1024" paragraph is replaced by the SI rationale, and states plainly that numbers shift against the old output (`0.50 MB` → `524.29 kB`)

See `1655a44ab`.

packages/components/src/FileField/fileUtils.ts
OsamaAbdellateef: Done — `useNumberFormatter` it is. `fileSizeFormatters`/`fileSizeFormatterFor` and the `useLocale` import are gone; the component does

```ts
const sizeFormatter = useNumberFormatter(FILE_SIZE_FORMAT_OPTIONS);
```

and `formatFileSize(size, formatter)` is now pure. Following the `NumericFormat.tsx` precedent rather than hand-rolling a cache was the right call.

One wrinkle worth naming: the boundary-rounding check inside `formatFileSize` derives `FILE_SIZE_ROUNDING` from `maximumFractionDigits`, so moving the formatter to the call site would have re-opened exactly the "magic `100` mirrors line 66" coupling from the earlier review, now across two files. So the options object is exported as `FILE_SIZE_FORMAT_OPTIONS` and the component passes that to `useNumberFormatter` — one source of truth, and a module-level constant keeps the identity stable for the hook's memo. The tests build their formatter from the same constant.

See `1655a44ab`.

packages/components/src/FileField/FileField.tsx
OsamaAbdellateef: Still open, and now more so. I commented `/run-chromatic` on 27 Aug, but `gh run list --branch dstsup-275_fix-filefield` shows no `Visual-Regression-Tests` run on any commit of this branch, so the dispatch never took.

The rendered text has also changed again since: `1655a44ab` moves the step to 1000 per `aromko`'s decision, so the same file now reads `524.29 kB` where this PR previously showed `512 kB`. Any run from before that commit would have been stale anyway.

Leaving the checklist item unticked until a run exists on `1655a44ab` and someone accepts the diffs deliberately — the snapshots should be taken against the final unit scale, not an intermediate one.

packages/components/src/FileField/fileUtils.ts
```diff
@@ -71,6 +71,62 @@ const dedupeFiles = (files: File[]): File[] => {
   });
 };
 
+// `kB`/`MB`/`GB`/`TB` are SI symbols, so the step is their SI value of 1000 -
```
aromko: Please remove the wall of texts.

packages/components/src/FileField/FileField.stories.tsx
```diff
@@ -176,26 +176,37 @@ UploadFile.test(
...
     await expect(canvas.getByText('test.txt')).toBeInTheDocument();
     await expect(canvas.getByText('pic1.jpg')).toBeInTheDocument();
-    await expect(canvas.getByText('2.00 MB')).toBeInTheDocument();
-    await expect(canvas.getByText('5.00 MB')).toBeInTheDocument();
-    await expect(canvas.getByText('0.50 MB')).toBeInTheDocument();
+    await expect(canvas.getByText('import.csv')).toBeInTheDocument();
+    await expect(canvas.getByText('2 MB')).toBeInTheDocument();
+    await expect(canvas.getByText('5 MB')).toBeInTheDocument();
+    await expect(canvas.getByText('512 kB')).toBeInTheDocument();
+    await expect(canvas.getByText('2.4 kB')).toBeInTheDocument();
```
aromko, line 209: 🔴 **Visual Regression Tests: stale**

The last successful `Visual-Regression-Tests` run is the one your `/run-chromatic` comment triggered on 27 Aug ([run 33079717179](https://github.com/marigold-ui/marigold/actions/runs/33079717179)). It ran against `faf191f8`. The branch has since moved to `1655a44a`, and this test is the one FileField snapshot Chromatic keeps (`disableSnapshot: false`), so its end state changed with the 1000 step and the fourth file.

Please comment `/run-chromatic` again so the baselines are taken from the final unit scale.

On "the dispatch never took": comment-triggered runs execute on the default branch and check the PR branch out inside the job, so `gh run list --branch dstsup-275_fix-filefield` never lists them. Filter by workflow and look for the PR title in the run name instead.

.changeset/dstsup-275_file-size-units.md
```diff
@@ -0,0 +1,13 @@
+---
+'@marigold/components': patch
+---
+
+fix(DSTSUP-275): pick the file size unit from the file's magnitude
+
+`<FileField>` rendered every selected file's size with a fixed megabyte divisor and two decimals — `(file.size / 1024 / 1024).toFixed(2)` — so anything under ~5 kB read `0.00 MB`. For consumers importing CSVs, where files are routinely 1–50 kB, the item description carried no information at all, and there was no way to override it from the outside: `<FileField>` owns the file list in internal state and renders the items itself.
```
aromko: ⚠️ **Two em dashes in a changeset**

#5779 (DST-1526) puts `.changeset/*.md` under Vale with the em dash rule at `error`. Whichever of the two PRs merges second turns the Prose check red, so it is cheaper to drop them here now. Commas carry the aside just as well:

```suggestion
`<FileField>` rendered every selected file's size with a fixed megabyte divisor and two decimals, `(file.size / 1024 / 1024).toFixed(2)`, so anything under ~5 kB read `0.00 MB`. For consumers importing CSVs, where files are routinely 1–50 kB, the item description carried no information at all, and there was no way to override it from the outside: `<FileField>` owns the file list in internal state and renders the items itself.
```

The `1–50 kB` range is fine, the en dash rule only catches spaced asides.

aromko (commented):
## Automated Code Review

Reviewed `1655a44a` against `CLAUDE.md`, DSTSUP-275 and the earlier review threads. All CI checks pass on the head. Spec match is complete: unit-aware sizes with the unit picked from the magnitude, localized through `Intl.NumberFormat`, and both decisions from the last round (1000 step with SI symbols, `useNumberFormatter` following the `NumericFormat` precedent) are in.

I re-ran the boundary maths on the head's `formatFileSize`: no byte count near any unit boundary prints `1,000 <unit>`, `999,995 B` carries to `1 MB`, and `Math.round` and `Intl` agree at every half-cent probe. The `[1000 ** 2 - 1, '1 MB']` case pins it.

Three inline comments: one blocker, one warning, one nit. Two things with no line to sit on:

- **The PR description still describes the 1024 build.** "The step stays at 1024 (unchanged)", the Before/After table (`2.34 kB`, "0.5 MiB image", "2 MiB PDF") and test step 3 (`2,34 kB`) all predate `1655a44a`. With the 1000 step the same fixtures read `2.4 kB`, `512 kB` for a 512,000-byte image, `2 MB` for a 2,000,000-byte PDF, and `2,4 kB` in `de-DE`. Worth fixing before merge, since the description is what the ticket links to.
- **PR title:** `fix([DSTSUP-275]):` should be `fix(DSTSUP-275):`. The repo squash-merges with the PR title as the commit title, so the brackets land in `main`'s history. Recent commits use the bare key: `fix(DST-1607):`, `fix(DSTSUP-272):`.

Still open from the previous round: the request to trim the comment blocks in `fileUtils.ts` (lines 74-77, 83-90, 95-101).

---

packages/components/src/FileField/fileUtils.ts
```diff
@@ -71,6 +71,62 @@ const dedupeFiles = (files: File[]): File[] => {
...
+    Math.round(scaled * FILE_SIZE_ROUNDING) / FILE_SIZE_ROUNDING >=
+      FILE_SIZE_STEP
+  ) {
+    exponent += 1;
+    scaled = bytes / FILE_SIZE_STEP ** exponent;
+  }
+
+  const value = formatter.format(scaled);
+
+  return `${value} ${FILE_SIZE_UNITS[exponent]}`;
```
aromko: 💡 **Both `let`s can go**

CLAUDE.md asks for `const` over `let`, and the carry reads more plainly as a second derived value than as a mutation. Same output for every input, checked against the current version on 200,000 random sizes plus the boundary cases in `fileUtils.test.ts`, and already prettier-clean:

```suggestion
  const top = FILE_SIZE_UNITS.length - 1;
  const magnitude =
    bytes === 0
      ? 0
      : Math.min(
          Math.max(Math.floor(Math.log(bytes) / Math.log(FILE_SIZE_STEP)), 0),
          top
        );
  const rounded =
    Math.round((bytes / FILE_SIZE_STEP ** magnitude) * FILE_SIZE_ROUNDING) /
    FILE_SIZE_ROUNDING;
  const exponent =
    magnitude < top && rounded >= FILE_SIZE_STEP ? magnitude + 1 : magnitude;
  const value = formatter.format(bytes / FILE_SIZE_STEP ** exponent);

  return `${value} ${FILE_SIZE_UNITS[exponent]}`;
```

packages/components/src/FileField/fileUtils.ts
OsamaAbdellateef: Trimmed — the three comment blocks around the size constants are down from 26 lines to 10, keeping only the parts that are not guessable from the code: why the step is 1000, why not `style: 'unit', and why the options object is exported. See `3e93e1e73`.

packages/components/src/FileField/fileUtils.ts
OsamaAbdellateef: Applied as suggested — both `let`s are gone, with the carry derived as `rounded`/`exponent` instead of a mutation. Checked the two versions agree on the `fileUtils.test.ts` boundary cases plus 300,000 random sizes, and `pnpm test:unit`/`test:sb` for FileField are green. See `3e93e1e73`.

.changeset/dstsup-275_file-size-units.md
OsamaAbdellateef: Applied — your suggestion landed as `21e6db288`, so the changeset has no em dashes left. Good call getting ahead of #5779 rather than racing it.

packages/components/src/FileField/FileField.stories.tsx
OsamaAbdellateef: Thanks for the correction on how comment-triggered runs are listed — that explains the empty `gh run list --branch`, and my earlier "the dispatch never took" was wrong.

Not triggered yet. The branch is now at `3e93e1e73`, which is comment trimming and a `let` → `const` refactor only, so the rendered text is unchanged since `1655a44a`. A `/run-chromatic` on this commit would take the baselines from the final unit scale. Leaving the checklist item unticked until that run exists and the diffs are accepted deliberately.

packages/components/src/FileField/fileUtils.ts
aromko: Confirmed on `3e93e1e73`. `FILE_SIZE_STEP` is `1000` with `kB`/`MB`/`GB`/`TB` kept, and the two-line comment above the constants carries the reason, so the next person formatting bytes in this repo finds the precedent and the rationale in the same place.

Spot-checked the follow-through rather than just the constant: `fileUtils.test.ts` now pins `999 B`/`1 kB` at the first boundary, `2400 → 2.4 kB`, and the rounding carry as `[1000 ** 2 - 1, '1 MB']`, and the changeset states the shift against the old output (`0.50 MB` → `524.29 kB`) instead of the withdrawn "keeps the number it had before" claim.

Resolving.

packages/components/src/FileField/fileUtils.ts
aromko: Done, and the wrinkle you named is handled better than I would have asked for. `fileSizeFormatters`/`fileSizeFormatterFor` and the `useLocale` import are gone, `formatFileSize(size, formatter)` is pure, and `FileField.tsx:87` does `useNumberFormatter(FILE_SIZE_FORMAT_OPTIONS)`.

Exporting the options object rather than inlining `{ maximumFractionDigits: 2 }` at the call site is the right call: `FILE_SIZE_ROUNDING` stays derived from the same constant the formatter uses, so the "magic `100` mirrors line 66" coupling cannot reopen across two files, and `fileUtils.test.ts:149` builds its formatter from the same export. Stable module-level identity for the hook's memo is a bonus.

Resolving.

packages/components/src/FileField/fileUtils.ts
aromko: Much better. Ten lines across the three blocks, and each says something the code does not: why the step is 1000, why not `style: 'unit'`, why the options object is exported. Nothing left restating the mechanics.

Resolving.

packages/components/src/FileField/FileField.stories.tsx
aromko: Settled. Your `/run-chromatic` took this time: `UI Tests: marigold_2026` is green on `3e93e1e7`, the current head ([build 270](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=270)), so the baselines come from the final unit scale rather than an intermediate one.

That clears the "Visual regression tests updated" checklist item, feel free to tick it.

Resolving.

packages/components/src/FileField/fileUtils.ts
aromko: Applied verbatim. `magnitude`/`rounded`/`exponent` read as three derived values now, and the boundary carry is visible as a condition instead of a reassignment.

Resolving.

packages/components/src/FileField/fileUtils.ts
```diff
@@ -71,6 +71,48 @@ const dedupeFiles = (files: File[]): File[] => {
   });
 };
 
+// `kB`/`MB`/`GB`/`TB` are SI symbols, so the step is 1000, matching what
```
aromko: Maybe remove all these comments pls. So much noise

aromko (approved):

aromko (approved):

PR #5765, opened 2026-08-26
fix(DST-1740): give the Checkbox and Switch labels a containing block

sarahgm:
# Description

React Aria renders the real `<input>` inside the control's `<label>` wrapped in `VisuallyHidden`, which is `position: absolute` with no offsets. Both the `Checkbox` and `Switch` labels were `position: static`, so the input anchored to whatever positioned ancestor the consumer happened to have. In a list inside a dialog that is the dialog itself, a dozen static elements up.

An absolutely positioned box laid out against an ancestor outside a scroll container is not part of that container's scrolled content, so the input stopped travelling with the row it belongs to. Clicking a row far down the list focused an input the browser believed was over a thousand pixels away, and it scrolled the dialog to reveal it. In the reported case the dialog also set `overflow: hidden`, so there was no scrollbar to get back and it ended up blank.

Both labels are now `relative`. React Aria's [own docs](https://react-aria.adobe.com/VisuallyHidden) name this as a precondition of `VisuallyHidden` ("it must have a `position: relative` or `position: absolute` ancestor"), citing stray scrollbars. The focus jump is the sharper consequence, which is probably why it went unnoticed.

**On the scope.** The ticket describes this as a `Checkbox`-only oversight, on the grounds that `Radio` and `Switch` already position their labels. `Radio` does. `Switch` only looked like it did: its `relative` sat on a wrapper `<div>` around the track, a sibling of the hidden input rather than an ancestor, so it never reached it. Reproduced identically for both controls, so both are fixed here. That wrapper had no other job (the thumb translates rather than absolutely positioning, and `ui-state-focus` is an outline), so it is gone and the track is now a direct child of the label.

`Checkbox.Group` renders `Checkbox` children and needs no change of its own.

Consumers who worked around this by setting `position: relative` on their own scroll containers can drop it, though leaving it in place is harmless.

Closes DST-1740

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Open Storybook → `Components/Checkbox` → the `Hidden input travels with the control inside a scroll container` test, and the same one under `Components/Switch`. Each renders a fixed-height list with rows below the fold and every ancestor `static`, and asserts the hidden input travels exactly as far as its row. Both fail on `main` (the row moves 82px, the input moves 0).
2. `pnpm test:sb` and `pnpm test:unit` — 581 story tests and 1583 unit tests pass.
3. Visual check on `Components/Switch`: the track wrapper `<div>` was removed, so confirm the track and thumb are unchanged at both `default` and `settings` variants. Chromatic is the real gate here.

## Breaking Changes

No

## Checklist

### DIFF
--- a/.changeset/dst-1740_boolean-control-containing-block.md
+++ b/.changeset/dst-1740_boolean-control-containing-block.md
@@ -1,0 +1,15 @@
+---
+'@marigold/components': patch
+---
+
+fix(DST-1740): give `Checkbox` and `Switch` labels a containing block for their hidden input
+
+React Aria renders the real `<input>` inside the control's `<label>` wrapped in `VisuallyHidden`, which is `position: absolute` with no offsets. The label was `position: static`, so the input anchored to whatever positioned ancestor the consumer happened to have. In a list inside a dialog that is the dialog itself, a dozen static elements up.
+
+An absolutely positioned box laid out against an ancestor outside a scroll container is not part of that container's scrolled content, so the input stopped travelling with the row it belongs to. Clicking a row far down the list focused an input the browser believed was over a thousand pixels away, and it scrolled the dialog to reveal it. In the reported case the dialog also set `overflow: hidden`, so there was no scrollbar to get back and it ended up blank.
+
+Both labels are now `relative`. React Aria's own docs name this as a precondition of `VisuallyHidden` ("it must have a `position: relative` or `position: absolute` ancestor"), citing stray scrollbars — the focus jump is the sharper consequence, which is probably why it went unnoticed.
+
+`Radio` already positioned its label and is unchanged. `Switch` looked like it did too, but its `relative` sat on the wrapper around the track, a sibling of the input rather than an ancestor, so it never reached it. That class moves up to the label, where it was meant to be. `Checkbox.Group` renders `Checkbox` children and needs no change of its own.
+
+Consumers who worked around this by setting `position: relative` on their own scroll containers can drop it, though leaving it in place is harmless.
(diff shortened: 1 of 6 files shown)


OsamaAbdellateef (commented):
## Automated Code Review

### Overview
| Field | Value |
|-------|-------|
| Files Changed | 6 |
| Additions | +124 |
| Deletions | -8 |
| Visual Regression | Passed (up to date — last successful run matches current head `3d3338a`) |
| CI | All 27 checks passing |

### Linked Ticket
**DST-1740**: *Checkbox: label is missing `position: relative`, hidden input escapes scroll containers*

The ticket asks only for a `Checkbox` fix. I verified this PR's fix matches exactly what the ticket prescribes, and independently verified the PR's expanded claim: `Radio` already has `relative flex w-full items-start` on its `RadioButton`, while `Switch`'s `relative` really was misplaced on a sibling wrapper `<div>` rather than an ancestor of the hidden input — so extending the fix to `Switch` is justified, not scope creep. `CheckboxGroup.tsx` only composes `Checkbox` children and has no wrapper of its own, so the PR's claim that it needs no change also checks out.

### Checklist
- **TypeScript / Component Patterns** — clean. No `any`, no new props, no `className`/`style` exposure, `useClassNames` used correctly.
- **React Performance** — not applicable; pure CSS positioning change (add `relative`, remove one wrapper `div`) plus two new interaction tests. No re-render/data-fetching/bundle concerns.
- **Testing Standards** — both new tests use `Basic.test(...)`, and `Basic` is already tagged `component-test` in both story files, so the tag is correctly inherited. Accessible queries used throughout, no `fireEvent`/`getByTestId`. The `getTrack` selector update (`div > div` → `:scope > div`) correctly reflects the new DOM shape after the wrapper `div` was removed.
- **Accessibility** — verified the theme style files (`Checkbox.styles.ts`, `Switch.styles.ts`) don't define any conflicting `static`/position utility on `container` that could clobber the new `relative` class via `cn()`'s last-wins merge.

### Issues Found
No critical or warning-level issues.

**Suggestion (nice to have):** the PR checklist item "Visual regression tests updated" is left unchecked, but VRT actually ran and passed against the current head commit — worth ticking for accuracy, though not blocking.

### Recommendation
**Approve** — root-cause analysis is precise and independently verified, the fix is minimal and correctly scoped, new regression tests target the actual failure mode, and CI/Chromatic are fully green.

---

OsamaAbdellateef (approved):

PR #5766, opened 2026-08-27
fix(DSTSUP-276): anchor the Calendar year list at year 1

aromko:
# Description

Opening the year dropdown on a year below 21 produced a list that counted down and then back up — focused on AD 5 it read `16 15 14 … 2 1 1 2 3 … 25`, with two options sharing the accessible name `1`.

`CalendarYearPicker` centres a fixed `visibleYears` window on the focused year, so a 41-year window reaches 20 years back. There is no year 0 in the Gregorian calendar, so from AD 5 that lands in 16 BC, and react-aria renders an era marker only when the *focused* date is BC, never the individual entries. The first 16 options were BC years wearing bare AD numbers.

The window is now anchored at year 1 rather than centred whenever the focused year sits too close to the boundary to centre it. The list stays 41 entries long: AD 5 offers `1 … 41`, as does AD 1. Since react-aria only shifts its window for `minValue`, which this component does not own, the window is instead sized to guarantee 41 in-era years and the entries that rolled into the previous era are dropped.

`minValue` calendars are untouched: react-aria re-anchors its own window there, and a BC `minValue` is an explicit opt-in, so neither the anchor nor the era filter applies. A `maxValue`-only calendar focused below AD 21 does change — the clamp drags the window back across the era boundary and those entries are dropped the same way, so `maxValue` AD 10 focused on AD 5 lists `1 … 10` instead of `31 BC … AD 10`. Every other bounded calendar renders exactly as before: the window still reaches whichever bound is farther and renders every in-range year. Verified against every bound combination (`min` only, `max` only, both, wide range, single-year range, BC bounds) before and after the change.

Closes DSTSUP-276

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Open Storybook → `Components/Calendar/Basic` and set `defaultValue` to a year below 21, e.g. `new CalendarDate(5, 6, 15)`.
2. Open the year dropdown. It should list `1` through `41` ascending — no duplicate `1`, no descending run at the top.
3. Repeat with `minValue` and/or `maxValue` set and confirm those lists are unchanged.
4. `pnpm test:sb` — the new `Anchors the year list at year 1 instead of rolling into the previous era` story test covers step 2.

## Breaking Changes

No

## Checklist


packages/components/src/Calendar/YearListBox.tsx
```diff
@@ -13,21 +13,17 @@ const YearListBox = ({ setSelectedDropdown }: YearDropdownProps) => {
...
+  const max = state.maxValue ? state.maxValue.year : start + 40;
   const visibleYears = 2 * Math.max(focusedYear - min, max - focusedYear) + 1;
 
   return (
     <CalendarYearPicker visibleYears={visibleYears}>
       {({ items, value, onChange, 'aria-label': ariaLabel }) => (
         <ListBox
           ariaLabel={ariaLabel}
-          items={items}
+          items={items.filter(item => item.date.era === state.focusedDate.era)}
```
jim761: ⚠️ **The era filter is unconditional — this truncates the list in non-Gregorian calendars.**

react-aria derives the calendar from the locale, so `<I18nProvider locale="ja-JP-u-ca-japanese">` yields the Japanese calendar, whose eras change every few decades. `useCalendarYearPicker` warns about exactly this where it builds the items ("in some calendars, such as the Japanese, the era may also change"), and it gates its own era formatting to `calendar.identifier === 'gregory' && era === 'BC'`.

Under Reiwa this filter cuts the 41-year list down to ~8 entries and makes pre-2019 years unreachable from the dropdown — a navigation regression rather than a cosmetic one. Since the bug being fixed only exists for Gregorian BC, the filter can be gated the same way (typechecks and passes Prettier as written):

```suggestion
          items={
            state.focusedDate.calendar.identifier === 'gregory'
              ? items.filter(item => item.date.era === state.focusedDate.era)
              : items
          }
```

jim761 (commented):
## Automated Code Review

Four notes on the year-window change. The core logic checks out — I traced `useCalendarYearPicker` in `react-aria@3.51.0` and confirmed the window math (focused AD 5 → `visibleYears: 73` → `1 … 41` after the filter), and that filtering `items` is safe for selection because RAC's `id` is an index into the unfiltered array that `onChange` resolves by lookup. One warning and three nits below.

---

packages/components/src/Calendar/YearListBox.tsx
```diff
@@ -13,21 +13,17 @@ const YearListBox = ({ setSelectedDropdown }: YearDropdownProps) => {
...
   const focusedYear = state.focusedDate.year;
 
-  // `CalendarYearPicker` centers a fixed `visibleYears` window on the focused
-  // year and clamps it to the bounds, so size the window to reach whichever
-  // bound is farther. Unbounded sides keep the focused-year ±20 window. A wide
-  // both-bounds range still renders every in-range year, so virtualizing the
-  // listbox is the path forward if the DOM size ever becomes a concern.
-  const min = state.minValue ? state.minValue.year : focusedYear - 20;
-  const max = state.maxValue ? state.maxValue.year : focusedYear + 20;
+  const start = Math.max(focusedYear - 20, 1);
```
jim761: 💡 `20` here and `40` on line 18 have to stay in a 1:2 relationship or the 41-year invariant silently breaks. A module-scope `const YEARS_BEFORE = 20;` used as `focusedYear - YEARS_BEFORE` and `start + YEARS_BEFORE * 2` makes the coupling explicit.

.changeset/dstsup-276_year-list-era-boundary.md
```diff
@@ -0,0 +1,13 @@
...
+
+Opening the year dropdown on a year below 21 produced a list that counted down and then back up — focused on AD 5 it read `16 15 14 … 2 1 1 2 3 … 25`, with two options sharing the accessible name `1`.
+
+`CalendarYearPicker` centres a fixed `visibleYears` window on the focused year, so a 41-year window reaches 20 years back. There is no year 0 in the Gregorian calendar, so from AD 5 that lands in 16 BC, and react-aria renders an era marker only when the *focused* date is BC — never the individual entries. The first 16 options were BC years wearing bare AD numbers.
+
+The window is now anchored at year 1 rather than centred whenever the focused year sits too close to the boundary to centre it. `Calendar` and `RangeCalendar` both keep a 41-year list: AD 5 offers `1 … 41`, as does AD 1. Since react-aria only shifts its window for `minValue`, which this component does not own, the window is instead sized to guarantee 41 in-era years and the entries that rolled into the previous era are dropped.
+
+Calendars with a `minValue` or `maxValue` are untouched — the bounded window still reaches whichever bound is farther and renders every in-range year.
```
jim761: 💡 Not quite true when `minValue` is set **and** the focused year is ≤ 21. `useCalendarYearPicker` clamps `minDate` to `minValue` and then recomputes `maxDate = minDate.add({ years: visibleYears - 1 })`, so `minValue: new CalendarDate(1, 1, 1)` with a focused AD 5 now lists `1 … 73` where it previously listed `1 … 41`. Every in-range year still renders, so the stated invariant holds — it's the "untouched" wording that overclaims. Realistic bounds are genuinely unaffected: for any focused year > 21, `start + 40 === focusedYear + 20`, identical to before.

.changeset/dstsup-276_year-list-era-boundary.md
```diff
@@ -0,0 +1,13 @@
...
+'@marigold/components': patch
+---
+
+fix(DSTSUP-276): anchor the `Calendar` year list at year 1
+
+Opening the year dropdown on a year below 21 produced a list that counted down and then back up — focused on AD 5 it read `16 15 14 … 2 1 1 2 3 … 25`, with two options sharing the accessible name `1`.
+
+`CalendarYearPicker` centres a fixed `visibleYears` window on the focused year, so a 41-year window reaches 20 years back. There is no year 0 in the Gregorian calendar, so from AD 5 that lands in 16 BC, and react-aria renders an era marker only when the *focused* date is BC — never the individual entries. The first 16 options were BC years wearing bare AD numbers.
+
+The window is now anchored at year 1 rather than centred whenever the focused year sits too close to the boundary to centre it. `Calendar` and `RangeCalendar` both keep a 41-year list: AD 5 offers `1 … 41`, as does AD 1. Since react-aria only shifts its window for `minValue`, which this component does not own, the window is instead sized to guarantee 41 in-era years and the entries that rolled into the previous era are dropped.
```
jim761: 💡 `RangeCalendar` is asserted here, but only `Calendar` has a story test. `YearListBox` is shared through `useCalendarOrRangeState`, so the risk is low — the claim just rests on inference rather than coverage. A mirrored test in the RangeCalendar stories would pin it.

packages/components/src/Calendar/YearListBox.tsx
aromko: Fixed in 623ba28 — but gating only the filter would have made this worse, so the anchor now gates on `gregory` too.

`focusedDate.year` is era-relative outside Gregorian, so Reiwa 8 hits `Math.max(8 - 20, 1)`, anchors at Reiwa 1 and inflates `visibleYears` from 41 to 67. With the suggested filter gate alone you'd ship a 67-entry listbox — a bigger DOM than before the PR, for no benefit.

Measured, focused = Reiwa 8 (2026):

| | main | PR as reviewed | PR + filter gate only | now |
| entries | 41 | 41 | **67** | 41 |
| range | Heisei 18 … Reiwa 28 | Reiwa 1 … Reiwa 41 | Heisei 5 … Reiwa 41 | Heisei 18 … Reiwa 28 |

Two corrections to the report: the list was not cut to ~8 entries. Reiwa gives 41 (Reiwa 1 … 41), Heisei 8 gives 30. The unreachable-pre-2019 conclusion held, the figure did not — worth not quoting onward.

The concern itself was right to raise: non-Gregorian is documented support (`docs/content/components/form/calendar/index.mdx:146`), not speculative.

Caveat on the fix: branch coverage for `YearListBox` is now 90%, and what is uncovered is exactly the non-Gregorian side of the new ternaries. No story pins it yet.

packages/components/src/Calendar/YearListBox.tsx
aromko: Done in 623ba28 — `YEARS_AROUND_FOCUS = 20` at module scope, used as `focusedYear - YEARS_AROUND_FOCUS` and `start + YEARS_AROUND_FOCUS * 2`.

.changeset/dstsup-276_year-list-era-boundary.md
aromko: Confirmed — 41 → 73 entries with `minValue: new CalendarDate(1, 1, 1)` and a focused AD 5.

Fixed in the code rather than the wording: react-aria already re-anchors the window at `minValue`, so the year-1 anchor is redundant there and only inflates `visibleYears`. It is now skipped whenever `minValue` is set.

`min AD 1` + focused AD 5 is back to `1 … 41`, and `min 1990` + focused 2026 is 73 both before and after — the pre-existing overshoot, unchanged. "Untouched" is now literally true, so the sentence stands with no caveat.

.changeset/dstsup-276_year-list-era-boundary.md
aromko: Added in 623ba28. Stronger case than you put it: `RangeCalendar.stories.tsx` already mirrors Calendar's `year picker is clamped to minValue/maxValue`, so the convention existed and the new test was the one that skipped it. The mirror now sits right next to it.

packages/components/src/Calendar/YearListBox.tsx
```diff
@@ -12,22 +14,28 @@ interface YearDropdownProps {
...
+  const gregorian = state.focusedDate.calendar.identifier === 'gregory';
 
-  // `CalendarYearPicker` centers a fixed `visibleYears` window on the focused
-  // year and clamps it to the bounds, so size the window to reach whichever
-  // bound is farther. Unbounded sides keep the focused-year ±20 window. A wide
-  // both-bounds range still renders every in-range year, so virtualizing the
-  // listbox is the path forward if the DOM size ever becomes a concern.
-  const min = state.minValue ? state.minValue.year : focusedYear - 20;
-  const max = state.maxValue ? state.maxValue.year : focusedYear + 20;
+  const start =
```
jim761: ⚠️ **The anchor and the filter are gated differently — intentional?**

The anchor is gated on `gregorian && !state.minValue`, but the era filter on line 36 is gated on `gregorian` alone. The last commit is titled *"limit the year-list anchor to unbounded Gregorian calendars"* — the anchor got limited, the filter didn't.

Consequence for bounded Gregorian calendars (replayed against `useCalendarYearPicker`'s arithmetic):

| Case | `main` | this PR |
| --- | --- | --- |
| `maxValue={AD 10}`, focus AD 5 | 41 options, `BC31…AD10` | **10** options, `AD1…AD10` |
| `minValue={30 BC}`, focus AD 5 | 41 options, `BC16…AD25` | **25** options, `AD1…AD25` |

The first is an improvement — main's 31 BC entries were exactly the mislabelled ones this PR is fixing. The second drops 16 years the consumer *explicitly* opted into via `minValue`. Vanishingly rare, and those entries were mislabelled on `main` anyway, so arguably no worse — but worth a deliberate decision either way: gate the filter the same as the anchor, or leave it and adjust the changeset (see the comment there).

---

⚠️ **The explanatory comment is gone and wasn't replaced.**

`main` carried five lines explaining that `CalendarYearPicker` centres a fixed window and clamps it to the bounds, hence sizing the window to reach the farther bound — plus *"a wide both-bounds range still renders every in-range year, so virtualizing the listbox is the path forward if the DOM size ever becomes a concern"*, which is now the only record of a known scaling limit.

The new code is strictly harder to read than the code that comment was explaining. Three things aren't self-evident: why `Math.max(…, 1)`; why `max` derives from `start` rather than `focusedYear` (that coupling is what keeps the list at 41 entries once the low side is clamped); and why the filter is needed on top of the anchor at all.

Minor, related: `YEARS_AROUND_FOCUS` reads as a radius, but line 26 uses `YEARS_AROUND_FOCUS * 2` as a span. A second named constant for the 41 both tests assert would make the intent land.

.changeset/dstsup-276_year-list-era-boundary.md
```diff
@@ -0,0 +1,13 @@
...
+
+`CalendarYearPicker` centres a fixed `visibleYears` window on the focused year, so a 41-year window reaches 20 years back. There is no year 0 in the Gregorian calendar, so from AD 5 that lands in 16 BC, and react-aria renders an era marker only when the *focused* date is BC — never the individual entries. The first 16 options were BC years wearing bare AD numbers.
+
+The window is now anchored at year 1 rather than centred whenever the focused year sits too close to the boundary to centre it. This applies to the Gregorian calendar only, whose era floor is the cause; other calendar systems change era every few decades, so their windows stay centred and keep crossing eras as before. `Calendar` and `RangeCalendar` both keep a 41-year list: AD 5 offers `1 … 41`, as does AD 1. Since react-aria only shifts its window for `minValue`, which this component does not own, the window is instead sized to guarantee 41 in-era years and the entries that rolled into the previous era are dropped.
+
+Calendars with a `minValue` or `maxValue` are untouched — the bounded window still reaches whichever bound is farther and renders every in-range year.
```
jim761: ⚠️ This paragraph isn't accurate, for both bound types.

> Calendars with a `minValue` or `maxValue` are untouched — the bounded window still reaches whichever bound is farther and renders every in-range year.

- **`maxValue`-only is not untouched.** The `start` clamp is gated on `!state.minValue`, not on "no bounds", so with `maxValue={AD 10}` and focus at AD 5, `min` moves from `-15` to `1` and the list goes from 41 options to 10.
- **`minValue` calendars don't render every in-range year.** The era filter runs regardless of bounds. `minValue={30 BC}`, focus AD 5, no `maxValue`: same window as `main` (`BC16…AD25`, 41 raw items), but 25 rendered — the 16 BC years are in range and get dropped.

See the comment on `YearListBox.tsx` for whether the code or the sentence should move. The same wording is in the PR description.

packages/components/src/Calendar/Calendar.stories.tsx
```diff
@@ -227,6 +227,27 @@ Basic.test(
...
+    args: {
+      defaultValue: new CalendarDate(5, 6, 15),
+    },
+  },
+  async ({ canvas, userEvent }) => {
+    await userEvent.click(canvas.getByRole('button', { name: '5' }));
+
+    const years = within(canvas.getByRole('listbox', { name: 'year' }))
+      .getAllByRole('option')
+      .map(option => option.textContent);
```
jim761, line 243: 💡 The filter's riskiest invariant isn't pinned by these tests.

The filter decouples two index spaces: `item.id` indexes RAC's **unfiltered** `years` array (RAC's `onChange` does `years[key].date`), while `ListBox` renders `id={String(index)}` off the **filtered** array. I traced it and it's correct as written — but neither new test ever selects a year, only reads `textContent`.

So if someone later "simplifies" `onSelect` to use the rendered index, focusing AD 5 and picking `1` would jump to 32 BC, and both new tests would still pass. Every other year-dropdown test in this file and in `RangeCalendar.stories.tsx` asserts the selection round-trip; these two are the exception.

A click plus a trigger assertion closes it — appended after the existing `expect`:

```ts
await userEvent.click(canvas.getByRole('option', { name: '1' }));

await expect(canvas.getByRole('button', { name: '1' })).toBeVisible();
```

(Same applies to the `RangeCalendar.stories.tsx` copy.)

packages/components/src/Calendar/YearListBox.tsx
aromko: Not intentional — fixed in 7432c0e. The anchor and the filter now share one flag:

```tsx
const anchorAtEraFloor =
  state.focusedDate.calendar.identifier === 'gregory' && !state.minValue;
```

The rule behind it: we supply the missing lower bound, and if the consumer supplied one we keep our hands off. react-aria re-anchors its own window at `minValue`, and a BC `minValue` is an explicit opt-in, so both the anchor and the filter step aside there.

Replayed against `useCalendarYearPicker@3.51.0` across 14 bound combinations:

| case | main | PR as reviewed | now |
| --- | --- | --- | --- |
| unbounded AD 5 | 41 `BC16…25` | 41 `1…41` | 41 `1…41` |
| `max` AD 10, focus AD 5 | 41 `BC31…10` | 10 `1…10` | 10 `1…10` |
| `min` BC 30, focus AD 5 | 41 `BC16…25` | 25 `1…25` | **41 `BC16…25`** |
| `min` 1990 / `max` 2025 / 1900–2100 / 2020–2021 / `min` AD 1 | — | unchanged | unchanged |

Both of your figures reproduce exactly.

One asymmetry worth naming so nobody "fixes" it later: gating the filter on `!maxValue` as well brings the original bug back. `max` AD 10 focused on AD 5 then renders `BC1 … AD10` — the anchor sets `start = 1`, react-aria's max-clamp pulls `minDate` back to 1 BC, and you get two options named `1`. The `maxValue` case needs the filter *because* the anchor is active there.

On the two smaller points:

**Comment** — not restored. Most of the *why* now lives in the changeset, which is the artefact that survives into the release notes. What genuinely has no home any more is the virtualization ceiling: a wide both-bounds range still renders every in-range year, and `1900`–`2100` is 201 DOM options today. That deserves a ticket rather than a comment that only gets read once someone is already in the file — I'll file one unless you'd rather see the block back.

**Naming** — keeping the single constant. `YEARS_AROUND_FOCUS` is a radius and `* 2` is the span, at one call site; a derived `41` is a second name to keep in sync for no reader gain. The 41 is asserted in both story tests, which is where a drift should surface.

Separately, something neither of us raised: `state.minValue.year` is era-relative, so `new CalendarDate('BC', 30, 1, 1).year === 30`, not `-29`. `min = state.minValue.year` therefore makes `focusedYear - min` negative for any BC bound and the `visibleYears` arithmetic is quietly wrong there — on `main` too. Harmless today because the max-side term wins, and out of scope here. Follow-up ticket.

.changeset/dstsup-276_year-list-era-boundary.md
aromko: Rewritten in 7432c0e, in both the changeset and the PR description:

> `minValue` calendars are untouched: react-aria re-anchors its own window there, and a BC `minValue` is an explicit opt-in, so neither the anchor nor the era filter applies. A `maxValue`-only calendar focused below AD 21 does change — the clamp drags the window back across the era boundary and those entries are dropped the same way, so `maxValue` AD 10 focused on AD 5 lists `1 … 10` instead of `31 BC … AD 10`. Every other bounded calendar renders exactly as before: the window still reaches whichever bound is farther and renders every in-range year.

Two calibrations on the report:

- **`maxValue` is not a regression, it is the fix reaching one more case.** 31 of main's 41 entries there were mislabelled BC years — the exact defect this PR removes. You say as much, then file it under "this paragraph isn't accurate" next to the `minValue` case. Only the `minValue` case was a real loss, and that one is fixed in code rather than in wording.
- **"`minValue` calendars don't render every in-range year" holds only for a *BC* `minValue`.** For any AD `minValue` the filter is a no-op: once react-aria re-anchors at the bound, the window cannot cross year 1. Worth not carrying that one forward as a general claim.

packages/components/src/Calendar/Calendar.stories.tsx
aromko: Added in 7432c0e — as a sibling test rather than appended, since CLAUDE.md's AAA rule rules out an act after an assert:

```tsx
Basic.test(
  'Selects year 1 from the anchored list',
  {
    args: { defaultValue: new CalendarDate(5, 6, 15) },
  },
  async ({ canvas, userEvent }) => {
    await userEvent.click(canvas.getByRole('button', { name: '5' }));

    await userEvent.click(canvas.getByRole('option', { name: '1' }));

    await expect(canvas.getByRole('button', { name: '1' })).toBeVisible();
  }
);
```

Mirrored in `RangeCalendar.stories.tsx`. Checked the assertion is unambiguous before relying on it: the year trigger's accessible name is a bare `state.focusedDate.year` (`CalendarListBox.tsx:40`) and RAC labels day cells with the full date, so `button` named `1` can only be the trigger.

Small correction: "every other year-dropdown test in this file asserts the selection round-trip" — 2 of the 5 pre-existing ones don't (`Clamps the year list at minValue and keeps a forward window`, `Reaches both ends of a wide bounded year range`). The substance stands, the count doesn't.

Also closed the coverage gap I flagged last round, since the filter's other arm is the one you got wrong in round 1: `Keeps the centred window in calendars without an era floor` pins the non-Gregorian path at 41 options. It needs `en-US-u-ca-japanese`, not `ja-JP-u-ca-japanese` — the listbox's `aria-label` comes from `Intl.DisplayNames(locale).of('year')`, so a Japanese *language* tag renders it as 年 and `getByRole('listbox', { name: 'year' })` misses. Length-only assertion: the formatted labels are ambiguous across the Heisei/Reiwa boundary, which is this very bug in a calendar we're deliberately not touching.

jim761 (approved):

PR #5767, opened 2026-08-27
refactor(DST-1731): stop cloning children to inject positioning classes in Input

aromko:
# Description

`Input` cloned its `icon` and `action` children to inject `absolute` plus the theme's placement classes. That is the alternative ADR-0005 rejects by name, and it is not one of the record's sanctioned exceptions: this is placement, not a collection, and it needs no runtime measurement.

`Input` now renders the positioned box itself and lets whatever it was handed fill it. `*:size-full` is what makes that work for an arbitrary child: it holds the icon at the box's 16px and stretches the action to the full control height, so the action's touch target stays control-sized whichever element a consumer passes. Without it an unstyled `Button` (`SearchInput`'s clear button, the `IconButton` in `ComboBox`) would collapse to its 16px icon, under the WCAG target-size floor.

**Behaviour is unchanged for every current caller.** What changes for consumers: a Fragment or non-element child is now placed like anything else, and a `className` set on the icon or action is left alone rather than merged into the placement classes.

### Two routes the ticket suggested, and why neither was taken

**CSS-only is not reachable.** `Input` puts `ui-control` on the `<input>` itself, so the visible frame is the input's own border and anything laid out in normal flow lands outside it. Placing the adornments without `absolute` means moving the frame to the wrapper, as `DateField` and `NumberField` already do. That breaks the public `ThemeComponent<'Input'>` contract and rewrites every state currently keyed off the input's own pseudo-classes. Worth doing, but as its own ticket.

**Context is not reachable either.** It only arrives at children that opt in. The icon is a bare `<svg>`, `SearchInput`'s loading action is a `<ProgressCircle>`, and `IconButton` does not consume `ButtonContext`, so it would take three mechanisms to cover what one wrapper covers. `action` is typed `ReactElement<any>`, so anything not opting in would render unpositioned.

### One correction to the ticket

The ticket says cloning "silently overrides a `ref` the consumer set". It does not. `cloneElement` preserves a child's ref, and in React 19 `element.props` carries `ref`, so the old `{...icon.props}` spread re-applied it. I checked both ways: a ref on the icon resolves before and after. The Fragment and `className` problems are real; the ref one is not.

Closes DST-1731
Unblocks DST-1530.

## Screenshots / Preview

Rendered geometry was diffed against the pre-refactor build in Firefox for `Input`, `SearchField` and `ComboBox`. Every measurement is identical, `cursor` included:

| | before | after |
| --- | --- | --- |
| icon svg | 16x16 @ 22,22 | 16x16 @ 22,22 |
| `Input` action button | 36x36 @ 1232,12 | 36x36 @ 1232,12 |
| `SearchField` clear button | 36x36, `cursor: pointer` | 36x36, `cursor: pointer` |
| `ComboBox` trigger button | 36x36, `cursor: pointer` | 36x36, `cursor: pointer` |

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Open Storybook, `Components/Input/WithIcons`. The leading icon sits at the same inset and the trailing action is still a full-height square, not a bare 16px icon.
2. `Components/SearchField/Basic` with text entered. Hover the clear button: the cursor must be a pointer, and the clickable area the full control height, not just the glyph.
3. `Components/ComboBox/Basic`. The chevron trigger is unchanged and still opens the list from anywhere in its 36px box.
4. Set `readOnly` on an `Input` with an `action` and confirm no action renders.
5. `pnpm test:unit` and `pnpm test:sb`.

## Breaking Changes

No. The icon and action each gain a wrapping `<span>`, so a consumer reaching past the public API into `Input`'s DOM with a structural selector could notice. Nothing in this repo does.

## Checklist


packages/components/src/ComboBox/ComboBox.test.tsx
```diff
@@ -28,7 +28,7 @@ test('check classname slots', () => {
   const button = screen.getByRole('button');
 
   expect(button.className).toMatchInlineSnapshot(
-    `"shrink-0 outline-0 absolute cursor-pointer right-0 flex size-control items-center justify-center text-secondary/80"`
+    `"shrink-0 cursor-pointer outline-0 text-secondary/80"`
```
OsamaAbdellateef, line 31: 🔴 **Visual Regression Tests — stale**

The last successful `Visual-Regression-Tests` run was on commit `335055f`, but the branch has since moved to `a5b58249`.

The linked ticket (DST-1731) explicitly calls this change high visual risk and asks for a VRT pass across `Input`, `SearchField`, and `TagField`. Please re-run visual regression tests to cover the latest changes by commenting `/run-chromatic` on this PR.

packages/components/src/Input/Input.tsx
```diff
@@ -76,8 +53,26 @@ const _Input = ({
         ref={ref}
         type={type}
       />
-      {inputIcon}
-      {inputAction}
+      {icon && (
```
OsamaAbdellateef, line 56: 💡 ADR-0005's "Alternatives rejected" section separately rejects "a wrapper `<div>` per region" (invisible-in-JSX, conditional-per-optional-part), and this `<span>`-per-icon/`<span>`-per-action is structurally that pattern, just for absolutely-positioned decorations rather than grid regions. The PR description already argues well why the ADR's two named routes (CSS-only, context) aren't reachable here, and this wrapper avoids all child-cloning/inspection, which is the actual harm the ADR targets. Given ADR-0005 names `Input` as its worked example, it may be worth a short addendum to the ADR noting this third route and why it's exempt, so this tension doesn't get re-flagged later.

packages/components/src/ComboBox/ComboBox.test.tsx
aromko: 👍

packages/components/src/Input/Input.tsx
aromko: Pushed an ADR update rather than an addendum, because I don't think this needs an exemption — and writing it as one is what would invite the re-flagging later.

The record states its own test one paragraph above the `Input` example: *"the test is not 'does this component read children' but 'does it infer what its children are'"*. `icon` and `action` are named props, so the new code reads nothing, sorts nothing, and never touches the element — `Input` is told, the same way `Aside` is told by its tuple type. Against the four reasons the rejection actually gives:

- type-sniffing children into buckets — doesn't apply, named props
- extra elements breaking `grid`/`flex`/`subgrid` inheritance — doesn't apply, the wrappers are `absolute` in a decoration layer, there is no grid relationship to inherit
- a conditional per optional part — applies, but `icon ? … : null` was already there, and a named prop can't be `:has()`-ed away
- load-bearing wrappers invisible in the consumer's JSX — applies, but strictly less invisible than `cloneElement` was

Two of four, and both of those were already true of the code being replaced. Calling it *exempt* implies the rule covers it, which hands the next person a precedent to claim the same exemption for a genuine children-sorting wrapper — the opposite of what the addendum would be for.

The bigger thing your comment pointed at: once this merges, ADR-0005 documents code that no longer exists. It quoted the deleted `cloneElement(icon, …)` block verbatim, said `Input` "fails" the test, and listed DST-1731 as open. So b8887d2 does three things:

- past-tenses the `Input` worked example, and says why the new wrapper looks like the rejected pattern but is not
- scopes the "wrapper `<div>` per region" rejection in one sentence to what it describes: wrappers derived from sorting an opaque `children` list, in normal flow
- drops "quietly overrides a `ref`" from the `cloneElement` rejection as well — same error the PR description corrects, and leaving it there would have contradicted the corrected paragraph two screens up

`pnpm check:memory` and `prettier --check` pass.

OsamaAbdellateef (approved):

PR #5768, opened 2026-08-28
feat(DST-889): mark links that open in a new window

sarahgm:
# Description

A `<Link>` that opens a new window or tab now marks itself: a small external-link icon after the label, plus a hidden, localized "opens in a new window" warning for screen readers ([W3C G200](https://www.w3.org/WAI/WCAG21/Techniques/general/G200), [G201](https://www.w3.org/WAI/WCAG21/Techniques/general/G201)). The API is unchanged, the marker is derived from `target`.

**Automatic, with no opt-out.** The ticket asked for an `external` boolean. Deriving from `target` was chosen instead, because an opt-in prop gets forgotten and a missing icon then reads as "this one stays here". Atlassian's `@atlaskit/link` is the closest precedent and does the same. If a link should not carry the icon, don't force the new window.

**Why "window" and not "tab".** The marker fires on any target that leaves the current window, named windows included, and only the browser decides whether a new browsing context lands as a tab or a window. "Window" is also G200's own phrasing.

**Left alone:** same-window targets (`_self`, `_top`, `_parent`, in any casing), `disabled` links, links with no `href`, and `download` links, where the browser saves the file and ignores `target`.

**`rel` and `aria-label`.** `target="_blank"` defaults `rel` to `noopener`, which your own `rel` overrides. A named window gets no default, since `noopener` makes the browser ignore the name and open a new tab on every click. An `aria-label` replaces a link's content, so the warning is folded into it rather than dropped, and with `aria-labelledby` it is referenced by id. That also fixes the `master`/`admin` access label, which an `aria-label` used to swallow.

**Theme.** `master`/`admin` coloured every descendant `svg`, which would have painted the new glyph too, so `AccessIcon` now carries an `access-icon` class for the theme to select. They also placed their icon with `inline-flex items-center gap-1`, which puts a trailing icon beside a wrapped label instead of after its last word, so both icons are now inline and sized in `em`.

Closes DST-889

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. `pnpm sb` → **Components/Link → New Tab**. The `_blank` links carry an external-link icon after the label, "Same tab" carries none.
2. In the a11y inspector, the first link's accessible name is `Marigold docs opens in a new window`, and the master one is `Support console Master opens in a new window`.
3. Same story: on the `master` link the lock is access-coloured, the external-link glyph is not, and both sit inline after the last word rather than centred beside a wrapped label.
4. `pnpm start` → `/components/actions/link`. Anatomy shows the new-window icon annotation, and **Opening in new windows or tabs** documents the behaviour.
5. `pnpm test:unit` and `pnpm test:sb`.

## Breaking Changes

No API change, but the marker is **retroactive**: every existing `target="_blank"` link gains the icon and a longer accessible name. Consumer tests asserting an exact name need updating, so `getByRole('link', { name: 'Terms' })` becomes `getByRole('link', { name: 'Terms opens in a new window' })`. Shipping as a `minor` with that note in the changeset, flagged here in case a reviewer wants it treated as a major.

## Checklist


packages/components/src/Link/Link.stories.tsx
```diff
@@ -107,3 +108,181 @@ AccessVariants.test(
     expect(admin.querySelector('svg')).toBeInTheDocument();
   }
 );
+
+export const NewTab = meta.story({
```
aromko, line 112: ⚠️ **Visual Regression Tests — stale**

The last successful run was on `597327e3` (2026-08-28 07:40, from your `/run-chromatic`), which was the final feature commit, so the code itself is covered. Since then `22aa881f` merged `main` in, which moves the baseline.

Worth re-running rather than waving through, because this is the most visually retroactive PR in the current queue. Every `target="_blank"` link anywhere in the system gains a glyph, and `Link.styles.ts` drops `inline-flex items-center gap-1` from `master`/`admin` in favour of inline `em` sizing, so those links reflow too. A stale baseline is exactly where that kind of change hides.

`/run-chromatic` when you are ready.

packages/components/src/Link/Link.tsx
```diff
@@ -32,13 +44,70 @@ const _Link = ({
...
+            variant={variant}
+            className={cn('me-[0.25em]', iconPlacement)}
+          />
           {typeof children === 'function' ? children(renderProps) : children}
-          <AccessLabel variant={variant} />
+          {newWindow ? (
+            <ExternalLink
+              size={16}
+              aria-hidden
+              className={cn('ms-[0.25em]', iconPlacement)}
```
aromko: ⚠️ This glyph gets only utility classes, so nothing can select it. That sits awkwardly next to the argument you just made one file over, in `AccessIcon.tsx`:

> The `access-icon` class is the markup contract a theme selects on, like `.selection-indicator`. Selecting on `svg` would catch every other glyph.

Which is right, and it is why `Link.styles.ts` and `Menu.styles.ts` moved off `[&_svg]` in this PR. But the glyph that made `[&_svg]` ambiguous in the first place is the one left without a name. A theme that wants to tint the new-window icon, resize it, or drop it in a dense context has to reach for `[&_svg]`, the selector this PR just removed, and would then hit the access icon too.

The asymmetry also shows inside a single `<a variant="master" target="_blank">`: two glyphs, one addressable, one not.

```suggestion
            <ExternalLink
              size={16}
              aria-hidden
              className={cn('new-window-icon', 'ms-[0.25em]', iconPlacement)}
```

Cheap now, awkward once the first theme needs it and `[&_svg]` is the only way in.

packages/components/src/Link/Link.tsx
```diff
@@ -32,13 +44,70 @@ const _Link = ({
...
+  const download = props.download !== undefined && props.download !== false;
+  const newWindow =
+    !!props.href &&
+    !disabled &&
+    !download &&
+    !!target &&
+    !sameWindowTargets.has(target);
+
+  // `noopener` makes the browser ignore a window name, so limit it to `_blank`.
+  const blank = newWindow && target === '_blank';
```
aromko, line 63: ⚠️ Limiting `noopener` to `_blank` is correct and I would not change the behaviour: `noopener` forces a fresh browsing context, so a named window would stop being reusable, which defeats the point of naming it. There is genuinely no way to have both.

What is missing is that the consequence is never stated anywhere. `target="popup"` ships with no `rel`, so the opened document keeps a live `window.opener` handle on the opener and can navigate it (reverse tabnabbing). Browsers have severed that for `_blank` by default since 2021, but not for named targets, so this is the one case where it still bites.

That matters more now than it did before this PR, because the component has started actively marking named windows as a first-class supported thing. Both the changeset and the docs section explain the `rel` decision purely as a functional trade ("`noopener` makes the browser ignore the window name") without the security half.

One sentence in the docs would cover it: a named window keeps a reference back to your page, so do not point one at an untrusted origin, and use `_blank` when you do not actually need the window reused.

.changeset/dst-889_link-new-tab-indicator.md
```diff
@@ -0,0 +1,19 @@
...
+'@marigold/docs': patch
+---
+
+feat(DST-889): mark links that open in a new window
+
+A `<Link>` that opens a new window or tab now shows an external-link icon after its label plus a hidden, localized "opens in a new window" warning (WCAG G201). Targets that stay in the current window (`_self`, `_top`, `_parent`, in any casing) are untouched, and so are links that open nothing: `disabled` ones, ones with no `href`, and `download` ones, where the browser saves the file and ignores `target`.
+
+`target="_blank"` also defaults `rel` to `noopener`, which your own `rel` still overrides. A named window gets no default `rel`, because `noopener` makes the browser ignore the window name and open a new tab on every click instead of reusing the window.
+
+**This is automatic and retroactive.** Every existing `target="_blank"` link gains the icon and a longer accessible name, so tests asserting an exact name need updating: `getByRole('link', { name: 'Terms' })` becomes `getByRole('link', { name: 'Terms opens in a new window' })`.
```
aromko, line 13: 💡 You asked in the PR description whether a reviewer would want this treated as major rather than minor. My read: **minor is defensible, keep it.**

Nothing about the type surface changes, nothing stops compiling, and every consumer keeps working. What changes is rendered output and accessible names, which is the same class of change as any restyle a design system ships in a minor. The changelog note you wrote already gives the exact before and after for the test breakage, which is the part that actually costs consumers time.

The part I would not treat as routine is the "no opt-out" decision rather than the version number. The reasoning is sound and I agree with it, but it is a cross-cutting product call: a consumer who disagrees has no escape hatch except dropping `target`, and, per my other comment, currently not even a theme-level one. Worth an explicit nod from someone besides the author before it ships, rather than being carried by the PR description alone.

packages/components/src/intl/messages.ts
```diff
@@ -28,6 +28,7 @@ export const intlMessages: Record<string, Record<string, LocalizedMessage>> = {
     loadingMessage: 'Lade...',
     moreInfo: 'Mehr Informationen',
     noResultsFound: 'Kein Ergebnis gefunden',
+    opensInNewWindow: 'öffnet in neuem Fenster',
```
aromko: 💡 Native-speaker call, so take or leave it: "öffnet in neuem Fenster" reads a little clipped with the article dropped. Appended to a link name it comes out as "Marigold Dokumentation öffnet in neuem Fenster", which parses but sounds like a headline rather than a screen-reader announcement.

The idiomatic phrasing for this is passive: "wird in einem neuen Fenster geöffnet". That is longer, but this file already carries full clauses in the same register ("Diese Breadcrumbs sind ausgeblendet", "Aktionen verfügbar."), so it would not be out of house style. "(neues Fenster)" is the other common option if brevity is the priority.

packages/components/src/Link/Link.tsx
sarahgm: Fixed in 03cf0ba5, taking your suggestion as written. The glyph carries `new-window-icon` now, with a one-liner at the call site saying it is this glyph's theme selector, like `access-icon`. No `theme-rui` change on purpose: nothing needs to select it today, the point is that a theme can.

Your argument found a second thing while I was in there. Every one of these tests asserted `querySelector('svg')`, and two counted `toHaveLength(2)`, so they would all still pass with both classes deleted. Exactly the contract you are asking for, and untested. They query `.new-window-icon` and `.access-icon` directly now.

packages/components/src/Link/Link.tsx
sarahgm: Agreed, and the behaviour stays. Documented in 8713e2e0, one paragraph after the existing `rel` explanation in **Opening in new windows or tabs**:

> That trade has a security half. Without `noopener`, the document you open keeps a live `window.opener` reference back to your page and can navigate it somewhere else. Browsers have severed that for `_blank` since 2021, but not for named targets. So don't point a named window at an origin you don't control, and use `_blank` unless you genuinely need the window reused.

You flagged the changeset too, so it got the same clause on its `rel` paragraph rather than only the docs.

packages/components/src/intl/messages.ts
sarahgm: Taken, thank you. `wird in einem neuen Fenster geöffnet` in 23e7fb2c.

Went with the passive clause over `(neues Fenster)` for the reason you gave: the file already carries full clauses in that register, and it composes cleanly with the access label into "Master wird in einem neuen Fenster geöffnet". No test churn either, everything asserts the `en-US` string.

.changeset/dst-889_link-new-tab-indicator.md
sarahgm: Keeping `minor`, agreed on your reasoning.

Worth noting the escape hatch improved since you wrote this: the glyph now carries `new-window-icon`, so a theme can restyle or hide it without touching `target`. That doesn't make it a consumer-level opt-out, which is still the part that needs a nod.

@sebald @OsamaAbdellateef this is the cross-cutting call: the new-window marker is derived from `target` with no `external` prop and no opt-out, and it's retroactive across every existing `target="_blank"` link. Reasoning is in the PR description. Could one of you sign off before this ships?

sarahgm:
Review findings addressed in 03cf0ba5, 8713e2e0 and 23e7fb2c. Re-running against the current head so the baseline covers the `main` merge as well as the fixes.

/run-chromatic

packages/components/src/Link/Link.stories.tsx
aromko: ✅ Resolved. Build [#254](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=254) ran on the current head `23e7fb2c` and came back `831 tests unchanged`, measured against the branch baseline from [#245](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=245) where you accepted the 5 icon changes.

So the `main` merge in `22aa881f` and the three fix commits move nothing visually, which is the thing a stale baseline would have hidden. Nothing left here.

packages/components/src/Link/Link.tsx
aromko: ✅ Resolved in `03cf0ba5`, and the second thing you found while in there is the better half of the fix.

`querySelector('svg')` and `toHaveLength(2)` passed with either marker class deleted, so the contract a theme selects on was the one part untested. Ten assertions now name `.access-icon` and `.new-window-icon` directly and fail if a class goes.

Agreed on leaving `theme-rui` untouched. The point was that a theme *can* select it, not that one does today.

packages/components/src/Link/Link.tsx
aromko: ✅ Resolved in `8713e2e0`. The paragraph names the live `window.opener` handle, the 2021 `_blank` cutoff, and the "use `_blank` unless you genuinely need the window reused" guidance, which is the actionable half.

Putting the same clause on the changeset's `rel` paragraph rather than only in the docs was the right call: the changeset is where a consumer meets this behaviour first.

packages/components/src/intl/messages.ts
aromko: ✅ Resolved in `23e7fb2c`. `wird in einem neuen Fenster geöffnet` is the phrasing I would have picked, and it composes as you say: "Support console Master wird in einem neuen Fenster geöffnet" reads as one clause rather than two fragments.

.changeset/dst-889_link-new-tab-indicator.md
aromko: Agreed, `minor` stands.

Fair point that `new-window-icon` improves the escape hatch, though as you say it is a theme-level one, not a consumer-level one. That remaining call is now correctly parked with @sebald / @OsamaAbdellateef rather than carried by the PR description, which is exactly what I was asking for. Nothing further from me on this thread.

packages/components/src/Link/Link.stories.tsx
aromko: ✅ Resolved. Build [#254](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=254) ran on the current head `23e7fb2c` and came back `831 tests unchanged`, measured against the branch baseline from [#245](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=245) where you accepted the 5 icon changes.

So the `main` merge in `22aa881f` and the three fix commits move nothing visually, which is the thing a stale baseline would have hidden. Nothing left here.

packages/components/src/Link/Link.tsx
aromko: ✅ Resolved in `03cf0ba5`, and the second thing you found while in there is the better half of the fix.

`querySelector('svg')` and `toHaveLength(2)` passed with either marker class deleted, so the contract a theme selects on was the one part untested. Ten assertions now name `.access-icon` and `.new-window-icon` directly and fail if a class goes.

Agreed on leaving `theme-rui` untouched. The point was that a theme *can* select it, not that one does today.

packages/components/src/Link/Link.tsx
aromko: ✅ Resolved in `8713e2e0`. The paragraph names the live `window.opener` handle, the 2021 `_blank` cutoff, and the "use `_blank` unless you genuinely need the window reused" guidance, which is the actionable half.

Putting the same clause on the changeset's `rel` paragraph rather than only in the docs was the right call: the changeset is where a consumer meets this behaviour first.

packages/components/src/intl/messages.ts
aromko: ✅ Resolved in `23e7fb2c`. `wird in einem neuen Fenster geöffnet` is the phrasing I would have picked, and it composes as you say: "Support console Master wird in einem neuen Fenster geöffnet" reads as one clause rather than two fragments.

.changeset/dst-889_link-new-tab-indicator.md
aromko: Agreed, `minor` stands.

Fair point that `new-window-icon` improves the escape hatch, though as you say it is a theme-level one, not a consumer-level one. That remaining call is now correctly parked with @sebald / @OsamaAbdellateef rather than carried by the PR description, which is exactly what I was asking for. Nothing further from me on this thread.

aromko (approved):
## Re-review: all four findings addressed

Went through `03cf0ba5`, `8713e2e0`, `23e7fb2c` and the current head.

| # | Finding | Fix | Verified |
|---|---|---|---|
| 1 | VRT stale, baseline moved by the `main` merge in `22aa881f` | `/run-chromatic` on head | ✅ [#254](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=254) on `23e7fb2c`: `831 tests unchanged`, against the [#245](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=245) baseline where the 5 icon changes were accepted |
| 2 | New-window glyph had no theme selector while `access-icon` did | `03cf0ba5` | ✅ `new-window-icon` on the glyph, and ten assertions moved off `querySelector('svg')` / `toHaveLength(2)` onto the marker classes |
| 3 | `window.opener` consequence of skipping `noopener` on named targets undocumented | `8713e2e0` | ✅ docs paragraph plus the same clause on the changeset's `rel` paragraph |
| 4 | German warning read clipped | `23e7fb2c` | ✅ `wird in einem neuen Fenster geöffnet`, no test churn |

Nothing new in the re-check. The three commits are scoped to exactly what was asked, with no behaviour drift in `Link.tsx` beyond the added class. All 30 checks pass on `23e7fb2c`.

**Approving.** The one thing still open is not code: the "no opt-out" product call you parked with @sebald / @OsamaAbdellateef. Worth ticking the VRT box in the description now that #254 is green.

---

PR #5608, opened 2026-07-03
test(DST-1593): stabilize DatePicker UnavailableDate Chromatic story

sarahgm:
# Description

The `UnavailableDate` story for `DatePicker` produced a non-deterministic Chromatic snapshot, causing intermittent VRT failures on unrelated PRs.

The story renders with `open: true` but set no `value`/`defaultValue` and did not mock the current date. Because no value was set, react-aria opened the calendar on the real current month and auto-focused the real "today". Unavailable dates in react-aria are still keyboard-focusable, so the focus ring landed on today's cell, and the `dateUnavailable` predicate marks every day except the 1st as unavailable. Since the baseline and each run happen on different days, the displayed month and focused cell drifted, so Chromatic saw a pixel diff and failed.

Fix: pin `defaultValue={new CalendarDate(2019, 6, 1)}` (an available day) so the displayed month and focused date are deterministic, while still demonstrating the unavailable-date feature.

Closes DST-1593

## Screenshots / Preview

Story: `Components/DatePicker/UnavailableDate`. The calendar now consistently opens on June 2019 with the 1st focused.

## Test Instructions

1. Open Storybook (`pnpm sb`) and view `Components/DatePicker` → `UnavailableDate`.
2. Confirm the calendar opens on June 2019 with the 1st (the only available day) focused, regardless of the current date.

## Breaking Changes

No

## Checklist

- [x] Stories added/updated (with `component-test` tag where applicable)
- [ ] Unit tests added/updated
- [ ] Component documentation added/updated (if it exists)
- [ ] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) (for new/changed interactive components)
- [x] Visual regression tests updated (for UI changes)
- [ ] Changeset added (`pnpm changeset`) — not needed, story-only change with no published package impact

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5769, opened 2026-08-28
test(DST-1688): pin Home/End focus in cell number fields

sarahgm:
# Description

`Home` and `End` inside a `NumberField` in a `<Table keyboardNavigationBehavior="tab">` used to push focus onto the row instead of staying in the field. It is fixed upstream, so this adds the assertions that would have caught it and drops a stale comment claiming it is still broken.

Verified by rendering the same table against react-aria-components 1.19.0 and 1.20.0 side by side. The escape was never `NumberField`-specific and was never `useSpinButton`: `useGridCell` had no bubble-phase `onKeyDown`, so any key from any focusable child of a cell reached `useSelectableCollection`, which owns `Home` and `End`. A `TextField` in the same table escaped just as readily. adobe/react-spectrum#10159 added the guard that stops keys originating inside a cell, and it shipped in 1.20.0, which we picked up in #5699.

Two steps added:

- `DataEntryGrid` asserts `Home` and `End` keep focus in the `NumberField`.
- `DataEntryGridDefaultNavigation` pins the remaining asymmetry: under arrow navigation a `TextField` loses focus to `Home` but a `NumberField` does not, because `useSpinButton` claims the key first.

No product code changes, so no changeset and no VRT — story-only, matching #5608.

Closes DST-1688

## Test Instructions

1. `pnpm exec vitest run --config vite.config.ts --project=storybook-tests packages/components/src/Table/Table.stories.tsx` — 62 tests pass.
2. Optional sanity check: invert either new assertion to `.not.toBe(balance)` and confirm it fails, so the steps are not vacuous.

## Breaking Changes

No

## Checklist

### DIFF
--- a/packages/components/src/Table/Table.stories.tsx
+++ b/packages/components/src/Table/Table.stories.tsx
@@ -2033,8 +2033,6 @@ DataEntryGrid.test(
       }
     );
 
-    // Asserted for the TextField only: Home inside the NumberField still
-    // reaches the grid, which looks like an upstream useSpinButton issue.
     await step('Home and End keep focus inside a text field', async () => {
       await userEvent.click(first);
       first.setSelectionRange(2, 2);
@@ -2058,6 +2056,14 @@ DataEntryGrid.test(
       await userEvent.keyboard('{ArrowRight}');
       await userEvent.keyboard('{Tab}');
 
+      expect(document.activeElement).toBe(balance);
+    });
+
+    await step('Home and End keep focus inside a number field', async () => {
+      await userEvent.keyboard('{Home}');
+      expect(document.activeElement).toBe(balance);
+
+      await userEvent.keyboard('{End}');
       expect(document.activeElement).toBe(balance);
     });
 
@@ -2114,6 +2120,16 @@ DataEntryGridDefaultNavigation.test(
       await userEvent.keyboard('{Home}');
 
       expect(document.activeElement).not.toBe(first);
+    });
+
+    await step('Home stays put in a number field', async () => {
+      const balance = canvas.getByLabelText(`Balance for ${users[0].name}`);
+
+      await userEvent.click(balance);
+
+      await userEvent.keyboard('{Home}');
+
+      expect(document.activeElement).toBe(balance);
     });
   }
 );
--- a/packages/components/src/Table/Table.stories.tsx
+++ b/packages/components/src/Table/Table.stories.tsx
@@ -2059,6 +2059,9 @@ DataEntryGrid.test(
       expect(document.activeElement).toBe(balance);
     });
 
+    // The field has no minValue or maxValue, so useSpinButton claims both keys
+    // and they no-op. Focus is all there is to assert. Give the field a bound
+    // and they start jumping the value while these assertions still pass.
     await step('Home and End keep focus inside a number field', async () => {
       await userEvent.keyboard('{Home}');
       expect(document.activeElement).toBe(balance);
@@ -2122,6 +2125,8 @@ DataEntryGridDefaultNavigation.test(
       expect(document.activeElement).not.toBe(first);
     });
 
+    // Unlike the TextField above, useSpinButton claims Home before the grid
+    // sees it, so the number field keeps focus under arrow navigation too.
     await step('Home stays put in a number field', async () => {
       const balance = canvas.getByLabelText(`Balance for ${users[0].name}`);
 


packages/components/src/Table/Table.stories.tsx
```diff
@@ -2061,6 +2059,14 @@ DataEntryGrid.test(
       expect(document.activeElement).toBe(balance);
     });
 
+    await step('Home and End keep focus inside a number field', async () => {
```
aromko, line 2065: ⚠️ These two assertions are right, and the focus escape is genuinely fixed. But they pin only half of what DST-1688 asks for, and I think the ticket should not close on this alone.

The ticket's Expected Outcome is: "Home and End either **move the caret within the value or step to min/max**, but in neither case move focus out of the cell." Today neither of the first two happens. Chasing the chain:

1. `useNumberField` always passes `onDecrementToMin` / `onIncrementToMax` (react-aria 3.51.0, `dist/private/numberfield/useNumberField.mjs:98,100`), so they are never undefined.
2. `useSpinButton`'s `Home` shortcut therefore always fires and returns handled, never `return false` (`dist/private/spinbutton/useSpinButton.mjs:85-91`). Handled means the key is consumed, so the caret does not move the way it does in the `TextField` step above.
3. But `decrementToMin` / `incrementToMax` no-op unless `minValue` / `maxValue` are set (react-stately 3.49.0, `dist/private/numberfield/useNumberFieldState.mjs:136-147`).
4. The demo field sets neither (`Table.stories.tsx:1977-1981`).

Net effect: Home and End in that field do nothing at all. Focus stays, which is what you assert, and the value is safe, which is why I am not asking for a value assertion here. What is still true is the sentence from the ticket's Problem section: "A user who cannot reach the start of a number field by keyboard has to arrow through the value character by character."

That is upstream `useSpinButton` behaviour, not something to work around in `Table` (the ticket rules that out explicitly), and the ticket's Included scope already anticipated "if upstream, file with adobe/react-spectrum and link it here". So my suggestion is to either file that second upstream issue and link it, or reword the ticket's Expected Outcome down to the focus half and say so. Merging this PR as the whole of DST-1688 leaves the original usability complaint unrecorded anywhere.

Worth adding one line here either way: if a `minValue` or `maxValue` ever lands on that demo field, Home and End start jumping the value to the bound and these focus-only assertions still pass.

packages/components/src/Table/Table.stories.tsx
```diff
@@ -2115,6 +2121,16 @@ DataEntryGridDefaultNavigation.test(
 
       expect(document.activeElement).not.toBe(first);
     });
+
+    await step('Home stays put in a number field', async () => {
```
aromko, line 2130: 💡 This step asserts `toBe(balance)` nine lines under a step asserting `not.toBe(first)` for the same key on the same grid, and nothing in the file says why. The reasoning is in your PR description ("because `useSpinButton` claims the key first") but it does not survive the merge.

This file documents exactly this kind of asymmetry everywhere else, including the comment right above the step you are following:

```
// The other half of the Home/End row asserted in the DataEntryGrid test:
// on the default, the key belongs to the grid rather than the caret.
```

and the stale one you correctly deleted in this same PR. A matching line here would keep the next reader from filing this as a contradiction.

```suggestion
    // Unlike the TextField above, useSpinButton claims Home before the grid
    // sees it, so the number field keeps focus under arrow navigation too.
    await step('Home stays put in a number field', async () => {
```

packages/components/src/Table/Table.stories.tsx
sarahgm: Agreed, and the chain holds line for line. Re-verified against react-aria 3.51.0 / react-stately 3.49.0, including the step you did not cite that closes it: `createKeyboardShortcutHandler` maps a handler returning `undefined` to `{shouldContinuePropagation: false, shouldPreventDefault: true}`. So the key is swallowed *and* default-prevented, then the value handler no-ops. The keys are inert.

One correction to the framing, and one thing that makes your point stronger than you put it.

**The residual was recorded**, in DST-1688 comment 588186 under "One residual, deliberately not fixed here". But you are right that the conclusion drawn there was wrong, and the reason it was wrong is worse than "we should have filed something". It justified leaving the keys alone on the grounds that this "matches the ARIA spinbutton pattern, where Home and End mean minimum and maximum rather than caret movement". `useNumberField` explicitly strips the spinbutton role from the input:

```js
// override the spinbutton role, we can't focus a spin button with VO
role: null,
'aria-valuemax': null, 'aria-valuemin': null,
'aria-valuenow': null, 'aria-valuetext': null,
```

The input is exposed as a plain textbox, and only the wrapper carries `role="group"`. The APG spinbutton pattern does not govern it. react-aria keeps the spinbutton key bindings while dropping the spinbutton semantics, so a user whose screen reader says "textbox" presses <kbd>Home</kbd> and gets nothing. Defect, not conformance.

**There is already an upstream issue**, which I had missed: [adobe/react-spectrum#2803](https://github.com/adobe/react-spectrum/issues/2803), filed 2022-02-03, reporting exactly this including the `TextField` contrast. `snowystinger` closed it on 2026-06-11 having reached the same cause and the same one-line fix ("if there's no min or no max, we just need to also return false here"), inviting a reopen with a use case. So a new issue would have been a duplicate.

I have not asked for the reopen. The issue drew one comment and zero reactions in four years, nothing here is blocked, and a fix reaches us through a version bump if it ever lands. Happy to be overruled if you think the use case is worth making.

Taken your suggestion on both counts:

- Added the caveat you asked for at this step, in 3e3f7933a.
- DST-1688's Expected Outcome is narrowed to the focus guarantee, with #2803 linked for the other half and the ARIA reasoning corrected on the ticket. It no longer claims something this PR does not deliver.

packages/components/src/Table/Table.stories.tsx
sarahgm: Applied verbatim in 3e3f7933a. You are right that the reasoning did not survive the merge, and the file does document this kind of asymmetry everywhere else.

Added a matching note at the `DataEntryGrid` step too, since that one has the same problem in reverse: it asserts focus only, and it would keep passing unchanged if a `minValue` or `maxValue` ever landed on the demo field and the keys started jumping the value.

packages/components/src/Table/Table.stories.tsx
aromko: Re-checked the chain against the installed versions, and every step holds.

- `createKeyboardShortcutHandler`: `result === undefined && action !== undefined` maps to `{shouldContinuePropagation: false, shouldPreventDefault: true}` (`react-aria/dist/private/interactions/createKeyboardShortcutHandler.mjs:110-113`). Swallowed *and* default-prevented, then the value handler no-ops. That is the step I was missing, and it closes it.
- The `role: null` block with the four `aria-value*` nulls is verbatim at `numberfield/useNumberField.mjs:217-224`, and the wrapper's `role: 'group'` at `:285`. The input is exposed as a textbox, so your correction stands: react-aria keeps the spinbutton key bindings and drops the spinbutton semantics. Defect, not conformance. Agreed.
- `decrementToMin` / `incrementToMax` sit behind `!= null` guards (`react-stately .../useNumberFieldState.mjs:136-147`) while `useNumberField` passes both unconditionally (`useNumberField.mjs:91-103`). Inert, confirmed.
- #2803 is as you describe: filed 2022-02-03, closed 2026-06-11 by `snowystinger`, one comment, zero reactions, same cause and the same one-line fix, reopen invited with a use case. A new issue would have been a duplicate, so good catch before I sent us to file one.

On the reopen I am not overruling you, but I would go the other way if it were mine. What kept it at low priority is the missing use case, and we now hold one sharper than the original report: the role is stripped to `textbox`, so a screen reader user is told they are in a text field where <kbd>Home</kbd> does nothing, and the fix is the line the maintainer already wrote out. The cost is one comment, and ticket comment 588830 is already 90% of it. Not a blocker, and your cost/benefit read is defensible, so your call.

The ticket half I am satisfied with: Expected Outcome narrowed, the other half named out of scope with #2803 in References, ARIA reasoning corrected in place. DST-1688 can close on the focus guarantee.

packages/components/src/Table/Table.stories.tsx
aromko: Confirmed applied at `3e3f7933a`. The caveat you added at the `DataEntryGrid` step is the better half of the change: that one asserts focus only as well, and it is the one that would have looked wrong once a bound landed on the demo field. Nothing further from me here.

aromko (approved):
## Automated Code Review (re-check)

Both round-one comments are addressed at `3e3f7933`, the ticket now claims only what this PR delivers, and every technical claim in the replies verifies against the installed `react-aria@3.51.0` / `react-stately@3.49.0`.

| | |
|---|---|
| Round-one findings | 2, both resolved |
| New findings | none |
| CI | 28 checks green, Storybook Tests 1-4 included |
| Changeset | correctly absent, nothing published changes, precedent #5608 |
| VRT | not applicable, both stories carry `chromatic: { disableSnapshot: true }` and nothing rendered changed |

Since the last review:

- The stale comment is gone and each new assertion carries the reason it can only check focus.
- `DataEntryGridDefaultNavigation` gained the asymmetry note.
- DST-1688's Expected Outcome is narrowed to the focus guarantee, with the inert <kbd>Home</kbd>/<kbd>End</kbd> tracked at adobe/react-spectrum#2803.

One thing left open by choice and not blocking: whether to comment on #2803 with the use case its maintainer invited. Reasoning in the thread.

**Approve.**

---

PR #5770, opened 2026-08-28
fix(DST-1689): clear hydration and a11y warnings in docs demos

sarahgm:
# Description

The Table docs page logged a React hydration mismatch: a `<td>` rendered a different integer on the server than on the client. `table-cell-alignment.demo.tsx` generated its Nights and Guests columns with `Math.random()` at render time, and `ComponentDemo` is a client component, so the demo is server-rendered and then hydrated, running the generator twice. Three consecutive requests to the page returned three different tables. Both columns now come from a fixed fixture merged into the venue data by index, hoisted to module scope since it no longer depends on anything per render.

`table-appearance.demo.tsx` rendered `<Table {...props}>`, and `AppearanceDemo` passes only `variant` and `size`, so that grid had no accessible name. It now carries `aria-label`, matching `listview-appearance.demo.tsx`. That was the sole source of the seven react-aria warnings the page logged.

Two neighbouring pages had the same class of defect and are fixed here rather than left behind:

- `selectlist-selectable-cards.demo.tsx` formatted capacity and price with `toLocaleString()` and no locale argument, so it read the JS runtime default: en-US under Node, the viewer's own locale in the browser (`$1,000` vs `$1.000`). `ComponentDemo` already pins `<I18nProvider locale="en-US">` for exactly this, but `toLocaleString` bypasses the provider. Both values now use `NumericFormat`, which reads the pinned locale through `useNumberFormatter`. Rendered output is unchanged.
- `menu-dialog.demo.tsx` wrapped each controlled dialog in `<Dialog.Trigger>` with the `<Dialog>` as its only child. `Dialog.Trigger` wraps children in a `PressResponder` to turn a child into the trigger, and a `Dialog` never registers a press, so react-aria warned four times. These dialogs are opened by the menu items and need no trigger: `open`/`onOpenChange` move onto `Dialog`, which already renders its own `Modal`. That is what `ConfirmationDialog` and `useConfirmation` do internally.

Console across the three pages goes from 4 errors and 11 warnings to zero.

Docs-only. No package source, exports or theme touched, and no stories changed, so VRT is N/A here.

Closes DST-1689

## Test Instructions

1. `pnpm start`, open `/components/collection/table` with devtools open. Console should be clean apart from the React DevTools tip and `[HMR] connected`.
2. Reload a few times. The Cell alignment table should always show the same Nights/Guests values (2/320, 1/80, 3/145, 1/9) and a `$6,250.00` total.
3. Open `/components/form/selectlist`. Console clean, and the Selectable cards demo shows `$1,000` / `$750` / `$500`.
4. Open `/components/overlay/menu`. Console clean. In the dialog demo, open Actions -> Rename..., close with Cancel; then Actions -> Delete, dismiss with Escape. Both should behave as before.

## Breaking Changes

No

## Checklist


docs/content/components/form/selectlist/selectlist-selectable-cards.demo.tsx
```diff
@@ -43,12 +50,17 @@ export default () => (
...
               </Inline>
               <Text size="sm" variant="muted">
-                Up to {venue.capacity.toLocaleString()} guests
+                Up to <NumericFormat value={venue.capacity} /> guests
               </Text>
             </Stack>
             <Inline space="related" alignY="center">
               <Text size="sm" weight="bold">
-                ${venue.price.from.toLocaleString()}
+                <NumericFormat
```
aromko, line 59: ⚠️ Good catch on this one, but its two closest twins are still out there. `card-grid.demo.tsx:38,45` is a line-for-line copy of the two lines you just fixed here, same `venues` data, same `Up to {capacity} guests` / `${price.from}` shape:

```
Up to {venue.capacity.toLocaleString()} guests
${venue.price.from.toLocaleString()}
```

`card-appearance.demo.tsx:35-36` is the worse of the pair, and worth checking before you decide what to do here. It renders `${venue.price.from.toLocaleString()} – ${venue.price.to.toLocaleString()}` with values of 1000 and 5000, so the grouping separator always fires. And unlike the demos on this page it goes through `AppearanceDemo`, which does *not* wrap its output in `<I18nProvider locale="en-US">` the way `ComponentDemo` does, and the raw text has no `suppressHydrationWarning`. That makes the Card docs page a live, unmasked mismatch on any non-en-US browser, which is a slightly bigger fish than the one this PR started on.

DST-1689 does say "a repo-wide audit of hydration warnings on other docs pages" is **Not included** and should be split into its own ticket, so filing the follow-up is the by-the-book answer and I am happy with that. Flagging it because you already stretched scope to two neighbouring pages on the "same class of defect, fixed here rather than left behind" argument, and these two are nearer neighbours than either of them.

For the ticket, if you file one: seven more bare `toLocaleString()` sites remain under `docs/content/patterns/user-input/table-record-management/` and `docs/app/(examples)/examples/analytics/page.tsx:117`. That is enough repetition that a lint rule banning the no-argument form in `docs/**` would probably outlast fixing them one at a time.

docs/content/components/form/selectlist/selectlist-selectable-cards.demo.tsx
```diff
@@ -43,12 +50,17 @@ export default () => (
                 <Badge variant="info">{venueTypes[venue.type]}</Badge>
               </Inline>
               <Text size="sm" variant="muted">
-                Up to {venue.capacity.toLocaleString()} guests
+                Up to <NumericFormat value={venue.capacity} /> guests
```
aromko: 💡 `NumericFormat` defaults `tabular = true` (`packages/system/src/components/Formatters/NumericFormat.tsx:28`), so this now renders `<span class="tabular-nums">` inside the sentence. Tabular figures exist to line digits up in a column; in running prose they are the wrong tool and read slightly wider than the proportional figures this text had before.

The price below is a fair use of them. This one is doing less than it looks: the first three venues have capacities of 500, 300 and 150, so no grouping separator is ever produced and the locale correctness the change buys is entirely theoretical for the current data.

```suggestion
                Up to <NumericFormat tabular={false} value={venue.capacity} />{' '}
                guests
```

Separately, the changeset says "Output is unchanged (`$1,000`, no cents, via `maximumFractionDigits`)". The *text* is unchanged, and I confirmed `maximumFractionDigits={0}` does not trip the `RangeError` that an explicit max below a defaulted min would, since the spec falls `minimumFractionDigits` back to `min(2, 0)`. But the DOM gains two spans and the digits change typeface metrics, which is worth a half-sentence so the next reader is not surprised by a visual diff.

sarahgm: Applied, thanks. `tabular={false}` on the capacity, price keeps the default.

Changeset amended too. It now says the rendered text is unchanged but the DOM gains a `<span suppressHydrationWarning>` per value, and explains why the capacity opts out of `tabular` while the price does not.

docs/content/components/form/selectlist/selectlist-selectable-cards.demo.tsx
sarahgm: Fixed both here rather than deferring them. You are right that they are the nearer neighbours, and once I looked at them the distinction you drew between the pair does not actually hold, in a way that argues for fixing both:

`I18nProvider` is irrelevant to either of them. It only feeds the react-aria hooks, and `toLocaleString` reads the JS runtime default no matter what provider is above it. That is the sentence already in the changeset. So `ComponentDemo`'s pin was never masking `card-grid.demo.tsx`, and its price line is `venue.price.from` = 1000, which produces a grouping separator in every locale. It was mismatching exactly as reliably as the appearance demo.

`card-appearance.demo.tsx` renders `venues[5]`, not `venues[0]`, so the range is `$3,000 – $10,000`. Four and five digits, so your conclusion holds and is a bit stronger than 1000/5000 would have been.

Both now go through `NumericFormat`. The range is a single tuple value, which `Intl.formatRange` renders as `$3,000 – $10,000`, byte-identical to the hardcoded `&ndash;` it replaces, so the Card pages have no visual diff.

On the follow-up scope, one correction to the count. Four bare sites remain, not seven:

- `docs/app/(examples)/examples/analytics/page.tsx:117`
- `table-record-add-another.demo.tsx:159`
- `table-record-inline-edit.demo.tsx:140`
- `table-record-drawer-crud.demo.tsx:305`

There are seven `toLocaleString()` calls under `table-record-management/`, but four of them already pass `'en-US'` (`table-record-entry-grid.demo.tsx:144` and `table-record-entry-grid-compare.demo.tsx:38,81,248`), so they are not hydration risks. The three bare ones map over the full `venues` array, which includes a capacity of 1000, so each one does mismatch on a row.

The analytics page is worth calling out in the ticket: it is `'use client'`, and the six visible `ticketsSold` values are all four digits (5240, 4480, 4120, 3480, 2670, 2090), so every row of that column mismatches.

Agreed on the lint rule outlasting the one-at-a-time fixes, and the four already-pinned calls show the rule has a clean form to enforce. Follow-up ticket drafted, filing it next.

sarahgm: Filed as [DST-1753](https://reservix.atlassian.net/browse/DST-1753), linked to DST-1689. It covers the four remaining bare sites plus the `no-restricted-syntax` guard scoped to `docs/**`.

docs/content/components/form/selectlist/selectlist-selectable-cards.demo.tsx
aromko: Verified, thanks. `tabular={false}` on the capacity here and on the copy in `card-grid.demo.tsx`, price keeps the default in both. The changeset now says the rendered text is unchanged while the DOM gains a span per value, which was the half-sentence I was after.

docs/content/components/form/selectlist/selectlist-selectable-cards.demo.tsx
aromko: All three corrections hold, and two of them are mine to own.

`venues[5]` is Grand Avenue Ballroom at 3000/10000, not the 1000/5000 I quoted from `venues[0]`, so the range is `$3,000 – $10,000` and your version of the point is the stronger one. And the count: at `c88eb280` there are eight `toLocaleString()` calls left under `docs/`, four bare and four already pinned to `'en-US'`, exactly as you list them. My seven counted the pinned ones.

I did check the byte-identity claim rather than take it, because a currency range is where ICU likes to slip in a narrow no-break space. On Node 24.16:

```
formatRange(3000, 10000)  ->  "$3,000 – $10,000"
codepoints                ->  24 33 2c 30 30 30 20 2013 20 24 31 30 2c 30 30 30
=== "$" + (3000).toLocaleString('en-US') + " – $" + (10000).toLocaleString('en-US')  ->  true
```

The shared literal is `U+0020 U+2013 U+0020`, so it is the plain spaces and en dash the `&ndash;` markup produced, and `maximumFractionDigits: 0` does not trip `formatRange` either.

Two more of your claims I spot-checked and can confirm: `venues[9].capacity` is 1000, so the three `table-record-*` demos that map the full array do mismatch on that row; and the analytics top six are 5240, 4480, 4120, 3480, 2670 and 2090, all four digits, so every visible row of that column mismatches today.

DST-1753 reads well. Scoping the `no-restricted-syntax` selector to `docs/**` while explicitly passing the four `'en-US'` calls is the right shape, and the selector as written only matches `toLocale(String|DateString|TimeString)`, so the `toLocaleLowerCase` caveat you noted is already handled by it.

You were also right to fix these here rather than defer them, and right about `I18nProvider`: that sentence was already in the changeset and I read past it.

docs/content/components/collection/card/card-appearance.demo.tsx
```diff
@@ -32,8 +33,12 @@ export default (props: CardProps) => (
     <Card.Footer>
       <Inline space="regular">
         <Text size="sm" variant="muted">
-          ${venue.price.from.toLocaleString()} &ndash; $
-          {venue.price.to.toLocaleString()}
+          <NumericFormat
```
aromko, line 36: 💡 Non-blocking, and it does not undo the fix. This demo renders through `AppearanceDemo`, which is the one demo wrapper that pins no locale: `ComponentDemo.tsx:83` wraps its output in `<I18nProvider locale="en-US">` with the comment "Pin the locale so demos render identically on server and client", and `AppearanceDemo.tsx` has no equivalent.

There is no mismatch here, and I checked why rather than assume it: react-aria's `useDefaultLocale` returns `en-US` for as long as `useIsSSR()` is true, so the server pass and the hydration pass agree, and `NumericFormat` suppresses the rest. What remains is that a de-DE reader sees `3.000–10.000 $` one render after first paint, where every `ComponentDemo` page on the site shows them `$3,000 – $10,000`.

`table-appearance.demo.tsx` already renders `NumericFormat` under the same wrapper, but its capacities are 150, 600 and 250, so nothing there is locale-dependent. This range is the first number in an appearance demo where the locale actually shows. One line in `AppearanceDemo.tsx` covers both pages, and it belongs to DST-1753 rather than to this PR.

Separate nit, same file, line 28: `Capacity: {venue.capacity}` is still a raw number while `card-grid.demo.tsx` now formats the same field through `NumericFormat`. 800 has no separator in any locale, so there is nothing to fix today, purely a consistency note if you touch the file again.

aromko (approved):
## Recheck of `c88eb280`

Both findings from my earlier pass are resolved, and the three factual corrections in the replies are correct, including the two that were mine. Details in the threads.

Re-verified on this commit:

- `Intl.formatRange` output is byte-identical to the markup it replaces (`U+0020 U+2013 U+0020` between the amounts), and `maximumFractionDigits: 0` does not trip it.
- Four bare `toLocaleString()` calls remain under `docs/`, matching DST-1753's list; the other four already pin `'en-US'`.
- A controlled `Dialog` without `Dialog.Trigger` is the supported path: `Dialog.tsx:142` renders its own `Modal`, and the `InnerDialog` comment documents this exact case.
- The Table fixture keeps every guest count under its venue's capacity, and the total is `$6,250.00`.
- `aria-label="Venues"` matches `listview-appearance.demo.tsx`.
- CI green. Docs-only, no component, theme or story files touched, so VRT stays N/A.

One non-blocking suggestion on `card-appearance.demo.tsx` about `AppearanceDemo` being the only demo wrapper without an `I18nProvider` pin. It is not a mismatch and belongs to DST-1753.

docs/content/components/collection/card/card-appearance.demo.tsx
sarahgm: Agreed on both, and the first one is now in the ticket.

I checked the mechanism rather than take it, and the react-aria source states it outright. `useDefaultLocale` (`react-aria/dist/private/i18n/useDefaultLocale.mjs`) returns `en-US` for as long as `useIsSSR()` is true, carrying the comment "We cannot determine the browser's language on the server, so default to en-US. This will be updated after hydration on the client to the correct value." Server pass and hydration pass therefore agree, and the locale only switches on the render after `useIsSSR()` flips. So there is nothing to suppress here, only the post-paint swap you described, and only for a reader whose `navigator.language` is not en-US.

Your one-line framing holds. `AppearanceDemo.tsx` already imports from `@marigold/components` and nests `OverlayContainerProvider` > `MarigoldProvider` at 181-182, which is the same shape `ComponentDemo.tsx` has at 81-84. The fix is `I18nProvider` added to that existing import and wrapped around `MarigoldProvider` in the matching position, and it covers `card-appearance` and `table-appearance` together.

Added to [DST-1753](https://reservix.atlassian.net/browse/DST-1753) as its own scope bullet, written so the next reader does not file it as a mismatch. It is a consistency fix rather than a hydration one, which is also why it stays out of this PR.

On line 28: `venues[5]` is Grand Avenue Ballroom at a capacity of 800, so no locale produces a separator for it. Leaving it raw.

PR #5771, opened 2026-08-28
fix(DST-1640): give the Sidebar.Rail an overflow affordance and a footer seam

sarahgm:
# Description

When a `Sidebar.Rail`'s top-level item list outgrows the available height it scrolls, but until now nothing except the scrollbar said so, and the pinned footer had no seam against the scrolling list. The two ran together and the scrollbar ended against a hard footer edge.

Two scroll-state-aware cues now sit on top of the scrollbar, so a rail that fits still shows no extra chrome. The list fades its overflowing edges through a new `ui-scroll-mask-y` utility, the block-axis twin of `ui-scroll-mask-x`, with `scroll-padding` keeping a focused tile clear of the fade so its focus ring is never half-erased. The footer takes the `ui-scroll-seam-*` top hairline the single-column sidebar's footer already had, fading out as the list bottoms out.

Unlike the horizontal version, `ui-scroll-mask-y` keeps the native scrollbar rather than hiding it. Down the block axis the scrollbar is the conventional affordance and the one a pointer user looks for, so the fade is additive rather than a replacement. Without scroll-driven animation support (Firefox as of 153) there is no fade, just the scrollbar, and the seam falls back to an always-on hairline. That trade is spelled out in the changeset.

`@marigold/theme-rui` takes a `minor` bump because `ui-scroll-mask-y` is a new public utility.

Closes DST-1640

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. Open Storybook, go to `Components/Sidebar` and pick **Rail Overflow** (nine tiles bounded to a 25rem viewport).
2. In Chromium, confirm the list's bottom edge is faded at rest, that scrolling down reveals a top fade too, and that the footer carries a top hairline which fades out as the list bottoms out.
3. Tab through the rail tiles. A tile scrolled into view should land clear of the faded edge with its focus ring fully drawn.
4. Toggle the rail collapsed via `Navigation umschalten`. The icon-only list should still scroll, with `Hilfe-Center` and `Profil` still pinned below it.
5. Open the plain **Rail** story and confirm a rail that fits shows no fade and no footer divider.
6. In Firefox, expect no fades and an always-on footer hairline. That is the documented `@supports` fallback, not a regression.
7. `pnpm test:sb` covers the structural assertions (8 tests on this story file).

## Breaking Changes

No

## Checklist


packages/components/src/Sidebar/SidebarRail.stories.tsx
```diff
@@ -450,9 +441,128 @@ Rail.test(
...
+      return node;
+    }
+    node = node.parentElement;
+  }
+  return null;
+};
+
+// Overflow: nine tiles in a 25rem viewport. Both cues are scroll-state aware,
+// so a rail that fits shows no extra chrome.
+export const RailOverflow = meta.story({
```
aromko, line 462: ⚠️ **Visual Regression Tests — stale**

The last successful `Visual-Regression-Tests` run was on `c45228e0` (2026-08-28 13:32), but the branch is now at `544cb057`. The two commits in between are comment-only, so the component and theme code is unchanged since that run. What has changed is the baseline: `544cb057` merges `main`.

Worth a re-run anyway, because this PR is unusually dependent on Chromatic. The fade lives behind `@supports (animation-timeline: scroll())`, and the story suite runs in Firefox (`vite.config.ts:64`), which takes the fallback branch. So `RailOverflow` and its mid-scroll snapshot are the *only* coverage the fade has anywhere in CI.

Re-trigger with `/run-chromatic` when you are ready.

themes/theme-rui/src/components/Sidebar.styles.ts
```diff
@@ -244,8 +252,13 @@ export const Sidebar: ThemeComponent<'Sidebar'> = {
...
     ],
   }),
-  // Pinned below the scrolling list; same tiles, same rhythm.
-  railFooter: cva({ base: ['shrink-0 flex flex-col gap-0.5 py-1.5'] }),
+  // Pinned below the list, same tiles and rhythm. Top seam mirrors the single
+  // column's: absent while the list fits, a hairline once it scrolls. Without
+  // scroll-driven animations (Firefox 153) it pins on, kept so that engine
+  // still gets this ticket's seam fixed. Default --seam-color, 0.14 vs 0.13.
+  railFooter: cva({
+    base: ['shrink-0 flex flex-col gap-0.5 py-1.5 ui-scroll-seam-footer'],
```
aromko: ⚠️ The `@supports not` branch of `ui-scroll-seam-footer` pins the hairline on unconditionally, which lands differently here than it does on the single column.

For the single-column sidebar that fallback is usually right: its nav list commonly overflows, so an always-on divider matches what Chromium shows most of the time. The rail is the opposite case. The documented guidance is a small, stable set of sections, so the rail usually does *not* overflow, and today (before this PR) its footer has no divider at all in any engine. After this change, every Firefox rail gains a divider that Chromium shows only on overflow.

That is squarely the one acceptance criterion in DST-1640 that says "when the rail fits, no extra chrome appears (affordance is scroll-state aware)". The changeset argues the trade honestly, so this is a question rather than a blocker: is fixing the Firefox overflow seam worth a permanent divider on the common case? A seam variant without the `@supports not` branch would keep "fits means no chrome" true in every engine, at the cost of leaving Firefox's overflow seam unfixed. I would lean that way, but it is your and design's call.

themes/theme-rui/src/components/Sidebar.styles.ts
```diff
@@ -192,11 +196,15 @@ export const Sidebar: ThemeComponent<'Sidebar'> = {
...
-  // fixed-width labels would otherwise grow a horizontal scrollbar.
+  // The scrolling item list. overflow-x-hidden: fixed-width labels would
+  // otherwise grow a horizontal scrollbar. ui-scroll-mask-y owns the vertical
+  // overflow (hence no overflow-y here) and fades over the scrollbar instead
+  // of replacing it. scroll-py reads its --sc-mask-width so they can't drift.
   rail: cva({
     base: [
-      'flex flex-col gap-0.5 flex-1 min-h-0 overflow-y-auto overflow-x-hidden py-1.5 ui-scrollbar',
+      'flex flex-col gap-0.5 flex-1 min-h-0 overflow-x-hidden py-1.5',
+      'ui-scrollbar ui-scroll-mask-y ui-scroll-seam-timeline',
```
aromko, line 210: 💡 Both existing `ui-scroll-mask-x` consumers pair the mask with two extras this one omits:

- `Tabs.styles.ts:16` — `overscroll-x-contain motion-safe:scroll-smooth`
- `SegmentedControl.styles.ts:53` — same

Without `overscroll-y-contain`, a wheel over the rail chains to the page once the list bottoms out. For a sticky full-height nav rail that may well be what you want, unlike a tab strip, so I am not asking for it blind. `motion-safe:scroll-smooth` is the easier one: focus-driven scrolling currently jumps.

Either way, a one-line note saying the divergence is deliberate would stop the next reader from filing this again.

packages/components/src/Sidebar/SidebarRail.stories.tsx
```diff
@@ -450,9 +441,128 @@ Rail.test(
...
+  tags: ['component-test'],
+  render: () => <RailShell bounded />,
+});
+
+RailOverflow.test(
+  'the item list scrolls while the footer stays pinned outside it',
```
aromko, line 468: 💡 This test covers two independent behaviours: the structural pinning plus seam (assert-only, no act), then a `focus()` act with the scroll-padding assertion. Splitting the focus case into its own `.test()` would keep each one arrange-act-assert and make a failure name the behaviour that broke.

```suggestion
  'the item list scrolls while the footer stays pinned outside it',
```
(no change to this line, just the split, so the second test would be a new `RailOverflow.test('focus lands a below-the-fold tile clear of the faded edge', ...)`)

packages/components/src/Sidebar/SidebarRail.stories.tsx
```diff
@@ -450,9 +441,128 @@ Rail.test(
...
+
+RailOverflow.test(
+  'collapsed the icon-only list still overflows and keeps its footer',
+  { parameters: { chromatic: { disableSnapshot: true } } },
+  async ({ canvas, userEvent }) => {
+    const toggle = canvas.getByRole('button', {
+      name: 'Navigation umschalten',
+    });
+
+    await userEvent.click(toggle);
```
aromko, line 555: 💡 `closestScrollable` runs immediately after the click, while `railItem`'s `grid-template-rows: auto 1fr → auto 0fr` fold is still animating over 200ms. The list is at its *tallest* during that window, so `scrollHeight > clientHeight + 1` passes regardless of whether the settled icon-only list actually overflows. The test's name claims the settled state.

Nine collapsed tiles do overflow 25rem by my arithmetic, so this passes for the right reason today, but a `waitFor` on the settled height would keep it honest if the tile geometry ever changes.

themes/theme-rui/src/components/Sidebar.styles.ts
sarahgm: Agreed, and changed. The asymmetry you name is the whole argument. The single column's nav usually overflows, so the pinned hairline is right most of the time. A rail of three to seven sections usually fits, so it would be wrong most of the time. Measured against `main` it is worse than that: the rail footer has no divider today in any engine, so shipping the fallback would have *regressed* the common Firefox case to fix the rare one.

`ui-scroll-seam-footer` now resolves its fallback through `var(--seam-fallback-color, var(--seam-color, var(--color-border)))`, and `railFooter` sets `[--seam-fallback-color:transparent]`. Leave the property unset and the branch is byte-identical to before, so Dialog's header and the single-column footer are untouched (Dialog's story tests pass unchanged). I preferred that over splitting the fallback into its own utility, which would have meant opting two existing consumers back in and risking a missed class.

Verified in the runner: the rail footer computes `rgba(0, 0, 0, 0) 0px -1px 0px 0px` in Firefox, so nothing is painted. The cost is the one you called out, the footer-meets-scrollbar seam stays unfixed in that engine, but it is left exactly as it is on `main` rather than traded for a regression. I checked whether anything else could make Firefox scroll-state-aware in pure CSS. `scroll-state()` container queries are Chromium-only too, so there is no third option.

This reverses the position in my comment on DST-1640, which I have updated to record the decision.

themes/theme-rui/src/components/Sidebar.styles.ts
sarahgm: Took the easy one and documented the other.

`motion-safe:scroll-smooth` is in, since focus-driven scrolling did jump. Chromatic forces `prefersReducedMotion: 'reduce'` (`.storybook/preview.tsx:39`), so snapshots are unaffected, but the vitest runner does not, which had a knock-on. `focus()` and `scrollTop =` now scroll asynchronously there, so both affected tests wait properly. The mid-scroll snapshot test now waits for its actual target instead of "moved at all", which it needed anyway.

`overscroll-y-contain` stays off, and the slot comment now says why. `SegmentedControl.styles.ts:50` spells out the reason the x-axis consumers contain: it stops horizontal overscroll from triggering the browser's back/forward gesture at the track ends. There is no block-axis twin of that gesture, and the single column's own `nav` doesn't contain either, so containing here would make the rail the odd one out within its own component family rather than bring it in line with Tabs.

packages/components/src/Sidebar/SidebarRail.stories.tsx
sarahgm: Split as suggested. `'the item list scrolls while the footer stays pinned outside it'` is now assert-only, and the focus case is its own `RailOverflow.test('focus lands a below-the-fold tile clear of the faded edge', ...)`.

One thing the split surfaced: the seam assertion in the first test had to change too. `not.toBe('none')` passed vacuously once the footer opted out of the fallback hairline, since a transparent shadow still computes to `rgba(0, 0, 0, 0) 0px -1px 0px 0px`. It now matches `/none|rgba\(0, 0, 0, 0\)/`, so it fails if anything paints that seam back on.

packages/components/src/Sidebar/SidebarRail.stories.tsx
sarahgm: Good catch, and it is not just theoretical. I probed it: immediately after the collapse click the label row is at **13.75px** of its 20px, so the list really was being measured mid-fold, at its tallest.

The test now waits for `grid-template-rows` to settle at `0px` before calling `closestScrollable`, so the overflow claim is about the settled icon-only list, as the name says. Your arithmetic holds either way, nine collapsed tiles do outgrow 25rem, but it now passes for the reason it states.

packages/components/src/Sidebar/SidebarRail.stories.tsx
sarahgm: Confirmed on the staleness, and thanks for spelling out why this PR leans on Chromatic more than most. `RailOverflow` plus its mid-scroll snapshot really are the fade's only coverage anywhere in CI.

Holding the re-run until now would have been pointless, since the seam fallback and the scroll behaviour both changed in `d6906d5`. Triggering it in a top-level comment now that the branch is settled.

themes/theme-rui/src/components/Sidebar.styles.ts
aromko: Agreed with the resolution, and `--seam-fallback-color` is the right shape. Unset resolves byte-identically, so Dialog and the single column really are untouched, and it avoids the "opt two existing consumers back in and risk a missed class" failure you name. Checked both consumers, `Sidebar.styles.ts:75` and `:267`, and only the rail sets the property.

One loose end, in the docs rather than the code. The sentence you added sits in the block comment that covers the whole `ui-scroll-seam-*` family (`ui.css:591-593`), so it reads as a general escape hatch: "set `--seam-fallback-color: transparent` to drop it". But only `ui-scroll-seam-footer` reads it. `ui-scroll-seam-header` still carries the hardcoded `box-shadow: 0 1px 0 0 var(--seam-color, var(--color-border))`, so a future header over a body that usually fits would set the property and get nothing. Either add the same `var()` layer to the header for symmetry (no behaviour change, since unset is identical) or name the footer in the sentence. I would take the second: nothing needs a transparent header fallback today, and the comment is the only thing promising otherwise.

themes/theme-rui/src/components/Sidebar.styles.ts
aromko: Both halves look right. `motion-safe:scroll-smooth` is in, and the `overscroll-y-contain` reasoning holds up: `SegmentedControl.styles.ts:50-51` does pin the containment to the browser's back/forward gesture, and there is no block-axis twin of that gesture.

Small consistency point on the same yardstick, though. The comment argues against containment partly because "the single column's nav doesn't contain either", and that nav (`Sidebar.styles.ts:58`) has no `motion-safe:scroll-smooth` either, so the rail is now the odd one out on that axis by the test the comment itself uses. It is defensible, since the rail is the only one with a mask whose fade a focus jump lands you inside, but that is the reason and the comment does not give it. A clause on the scroll-smooth half would stop the next reader filing the mirror image of this comment.

packages/components/src/Sidebar/SidebarRail.stories.tsx
aromko: Split looks good, and the seam assertion catching its own vacuity is the better half of it.

One notch further is available. The `storybook-tests` project runs Firefox only (`vite.config.ts:116`), so the `none` branch of `/none|rgba\(0, 0, 0, 0\)/` is unreachable, and it is the branch that keeps the assertion from guarding both ways: drop `ui-scroll-seam-footer` from `railFooter` entirely and `box-shadow` computes to `none`, which the regex still accepts. Anchoring on the transparent shadow alone, `/^rgba\(0, 0, 0, 0\)/`, fails if anything paints the seam back on *and* if the seam stops being applied, which is what you describe the test as doing.

packages/components/src/Sidebar/SidebarRail.stories.tsx
aromko: Resolved, and the 13.75px probe is a better answer than my arithmetic was. Waiting for the last grid track to reach `0px` puts the measurement on the settled icon-only list, which is what the name claims. Nothing further from me.

packages/components/src/Sidebar/SidebarRail.stories.tsx
aromko: Confirmed up to date. Build 253 ran on `d6906d5` (the run log records `Commit 'd6906d5' on branch 'DST-1640-rail-overflow-affordance'`), TurboSnap traced the change to `SidebarRail.stories.tsx`, captured 24 snapshots and passed with no changes. The `UI Tests` check is green against the current head, so nothing here blocks.

One thing worth checking before we treat those snapshots as the fade's coverage, though, because it goes to whether they cover it at all. `.storybook/preview.tsx:38` sets `pauseAnimationAtEnd: true`, and Chromatic's documented behaviour is to pause CSS animations at their last frame. `ui-scroll-mask-y` and `ui-scroll-seam-footer` are `animation`s; they just happen to be driven by a scroll timeline rather than by time. If Chromatic's pausing reaches them, every snapshot renders the end keyframe regardless of scroll position: `sc-reveal-top` at `--sc-mask-top: var(--sc-mask-width)` (top fade fully on), `sc-hide-bottom` at `--sc-mask-bottom: 0px` (bottom fade off), `scroll-seam-footer` at `box-shadow: 0 -1px 0 0 transparent` (no seam). That is the inverse of the resting state the description promises, and it would make the resting and mid-scroll snapshots identical.

I could not settle it from outside: Chromatic documents the "last frame" behaviour but not the mechanism, and says nothing about progress-based timelines. The decisive check is two minutes in build 253, opening `RailOverflow` next to the mid-scroll snapshot and seeing whether the fades differ in the expected direction (bottom only at rest, both mid-scroll). If they differ, this thread is done and so is my earlier worry. If they are identical, the affordance has no coverage anywhere in CI and the mid-scroll snapshot is not earning its keep, which would be a `main`-wide problem for `ui-scroll-mask-x` and the Dialog seam too and belongs in its own ticket rather than here.

aromko (approved):
## Re-review of `d6906d5`

All five threads from the previous round are addressed, and addressed at the root rather than patched where they surfaced.

| Round 1 | Status |
| --- | --- |
| VRT stale | Re-run. Build 253 on `d6906d5`, 24 snapshots, passed. One caveat below |
| Firefox seam fallback regressed the common rail case | Fixed via `--seam-fallback-color`, existing consumers byte-identical |
| Divergence from the `ui-scroll-mask-x` consumers undocumented | `motion-safe:scroll-smooth` added, the overscroll divergence documented |
| One test covering two behaviours | Split, and the vacuous seam assertion fixed with it |
| Collapsed overflow measured mid-fold | Waits for the settled grid track |

CI is green on `d6906d5` across the board, including the Firefox story tests and the Chromatic `UI Tests` check.

Nothing blocking. Four things left, three of them one-liners, all in the threads above:

1. **Worth a look before merge:** `pauseAnimationAtEnd: true` may be pinning the scroll-driven fade and seam to their end keyframe in every Chromatic snapshot, in which case build 253 does not cover the scroll-state-aware behaviour at all. Two minutes comparing two snapshots in the build settles it.
2. `ui.css` advertises `--seam-fallback-color` for the whole `ui-scroll-seam-*` family, but only the footer reads it.
3. The `boxShadow` regex's `none` branch is unreachable in a Firefox-only suite, and it is the branch that stops the assertion guarding both ways.
4. `motion-safe:scroll-smooth` diverges from the single column on the same yardstick the adjacent comment uses to justify the overscroll divergence.

**Approve**, on the understanding that #1 gets a look. If Chromatic does turn out to be neutralising the fade, that is a `main`-wide problem for `ui-scroll-mask-x` and the Dialog seam too, and belongs in its own ticket rather than here.

PR #5774, opened 2026-08-31
fix([DST-1684]): ignore virtualized ListBox row measurements taken before the list has a width

OsamaAbdellateef:
# Description

`ListLayout.buildItem` sizes each virtualized row from `virtualizer.size.width`, and `useVirtualizerItem` measures that row's height from the DOM as soon as its layout info is marked estimated. When the width is not a usable number the row wrapper collapses — it is absolutely positioned with `contain: size`, so an absent width leaves it at `0px` — and the measurement describes a row nothing will ever render.

That measurement used to be harmless: `min-width: auto` floored the option at its content's `min-content` width, so the row measured its normal height regardless of the width the layout had assigned. Since `wrap-anywhere` removed that floor (DSTSUP-269), a row measured at `0px` wraps character by character and reports a height in the hundreds of pixels instead of ~32. Those heights push the focused option far down the list, React Aria scrolls it back into view, and that scroll flips `pointer-events: none` on the virtualizer content for React Aria's 300ms scroll cooldown — swallowing a click on an option that was visible the whole time.

`ListBoxLayout` rejects a measurement taken while the list has no usable width, leaving the row at `estimatedRowHeight` — which is what it measured before `wrap-anywhere` existed.

Closes DST-1684

## One correction to the ticket

The ticket names one path to the collapsed state; there are **two**, and it describes the wrong one as the cause of the confirmed failure.

1. **First layout pass after mount.** `ScrollView` starts at `size.width === 0` and only learns its real width a microtask later, from a `ResizeObserver`. Transient — resolves on the next pass.
2. **Our browser-mode test runs.** `vite.config.ts` defines `process.env.NODE_ENV` as `'test'` with no `VIRT_ON`, which is React Aria's switch for reporting an `Infinity × Infinity` viewport so tests can query every row instead of only the visible ones. A non-finite width is dropped as an invalid CSS value, collapsing the wrapper the same way — and this one **never resolves**.

**The user-facing bug does not reproduce.** I drove the virtualized `Select` in a real dev-server browser with a capture-phase `scroll` listener and a `MutationObserver` on the virtualizer content, opening it fresh and reopening it with a non-first option selected. With the fix reverted: zero scroll events, `pointer-events: auto` throughout, the content div created already at its correct settled size (`305px × 26407px`), `scrollTop` never moved. The transient zero-width pass never reaches the DOM, because `ScrollView`'s `queueMicrotask` + `flushSync` resolves before commit.

So what PR #5691's `clickOption` helper was working around is a test-environment failure, not a production one. The ticket's "not a test-environment quirk" line is worth correcting — the confirmation run behind it was `pnpm test:sb`, which is exactly where path 2 applies. The fix covers both paths regardless, so the theoretical first-pass window is closed too.

## Notes for the reviewer

- **The skip path deliberately clears `estimatedSize`.** Leaving it set would re-measure — and force a reflow on — all 800 rows on every render in the runs where the width never becomes finite. The real browser is unaffected: the next `update()` rebuilds layout infos and re-flags them estimated once the width changes, so the row is still measured properly.
- **`.storybook/test-utils.ts` is deleted, along with all five `clickOption` call sites.** That helper was a `waitFor` retry added by #5691 to work around this bug rather than fix it. Keeping it would let a regression hide behind a green suite, so removing it is part of the proof — not incidental cleanup.
- **`fix` rather than `refactor`** because it removes a real failure mode, even though the observable change is confined to the test environment.

## Test Instructions

1. `git revert --no-commit HEAD` (or point `ListBox.tsx` back at RAC's `ListLayout`) and run `pnpm test:sb --run packages/components/src/Select/Select.stories.tsx`. The new `Selects an option clicked immediately after reopening the list` test fails with a 30s timeout — Playwright auto-waits for actionability rather than throwing on `pointer-events: none`.
2. Restore the fix and re-run — it passes.
3. `pnpm test:sb` and `pnpm test:unit` for the full picture.
4. Optional, for the "no real-browser change" claim: `pnpm sb`, open `Components/Select` → `Opens a large option list`, select a non-first option, reopen, and click another one immediately. Behaves identically with and without the fix.

## Breaking Changes

No

## Checklist

- [ ] Unit tests added/updated — N/A, the regression is only reachable through a real browser, so it is covered by a story test
- [ ] Component documentation added/updated (if it exists) — N/A, no API change
- [ ] Visual regression tests updated (for UI changes) — no visual change expected; see the probe results above

## Verification

- New regression test at `Select.stories.tsx`: selects a non-first option, reopens, clicks again with a plain `userEvent.click` and no retry. Confirmed failing before the fix, passing after.
- Full suites green: **1585 unit tests**, **585 story tests**, 0 failures.
- `tsc --noEmit` clean, eslint 0 errors, prettier clean, and `check:rac-imports` / `check:theme-zindex` / `check:memory` all pass.


packages/components/src/ListBox/ListBox.tsx
```diff
@@ -66,7 +67,7 @@ const ListBoxBase = ({
       <div className={classNames.container}>
         {virtualized ? (
           <Virtualizer
-            layout={ListLayout}
+            layout={ListBoxLayout}
```
sarahgm, line 70: 🔴 **Visual Regression Tests — not run**

packages/components/src/ListBox/ListBoxLayout.ts
```diff
@@ -0,0 +1,21 @@
...
+      const layoutInfo = this.layoutNodes.get(key)?.layoutInfo;
+      if (layoutInfo) {
+        layoutInfo.estimatedSize = false;
+      }
+
+      // `false` means "nothing changed", so the virtualizer skips a relayout it does not need.
+      return false;
+    }
+
+    return super.updateItemSize(key, size);
```
sarahgm, line 21: ⚠️ This line never executes in our test suite.

`ScrollView` reports `Infinity` width whenever `NODE_ENV === 'test'` and `VIRT_ON` is unset and `clientWidth` isn't mocked — and `clientWidth` lives on `Element.prototype`, not `HTMLElement.prototype`, so the mock check is false in a real browser too. `vite.config.ts:43-44` defines exactly that pair for both projects. So every `updateItemSize` call in all 585 story tests takes the skip branch above, and `super.updateItemSize` is never reached.

That means the new `Select` story pins the skip path, not that real measurement still works. If someone later inverts the guard, or `virtualizer.size` changes shape upstream, the suite stays green.

A unit test with a stubbed virtualizer would cover both branches cheaply — something like asserting `updateItemSize` returns `false` and clears `estimatedSize` when `size.width` is `0`/`Infinity`, and delegates to the base class when it's a real number.

.changeset/eighty-pugs-repeat.md
```diff
@@ -0,0 +1,13 @@
...
+'@marigold/components': patch
+---
+
+fix(DST-1684): ignore virtualized ListBox row measurements taken before the list has a width
+
+`ListLayout` sizes each virtualized row from the virtualizer's own width, which is `0` on the
+first layout pass and non-finite in browser-mode test runs. Either way the row wrapper collapses,
+and since `wrap-anywhere` removed the `min-width: auto` floor on options (DSTSUP-269), a row
+measured in that state wraps character by character and reports a height in the hundreds of pixels.
+The resulting scroll-into-view put every option behind React Aria's 300ms `pointer-events: none`
```
sarahgm: ⚠️ This contradicts the PR description's own finding.

The changeset tells consumers their swallowed clicks in `Select`, `ComboBox`, `Autocomplete` and `TagField` are fixed. The PR body says in bold that **the user-facing bug does not reproduce**, and that the confirmation run behind the ticket was `pnpm test:sb` — where the test-environment path applies.

This text goes into public release notes. Worth rewording to match what the probe actually showed: the collapsed-width window is closed, the observable failure was in browser-mode test runs, and the transient first-pass window never reached the DOM in a real browser.

packages/components/src/ListBox/ListBoxLayout.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { Key } from '@react-types/shared';
+
+export class ListBoxLayout extends ListLayout<unknown> {
+  updateItemSize(key: Key, size: Size): boolean {
+    const width = this.virtualizer?.size.width;
+
+    if (width == null || !Number.isFinite(width) || width <= 0) {
+      const layoutInfo = this.layoutNodes.get(key)?.layoutInfo;
+      if (layoutInfo) {
+        layoutInfo.estimatedSize = false;
```
sarahgm, line 14: 💡 This line is the half of the fix that stops a re-measure loop, and it's the only one without a comment.

`useVirtualizerItem` measures whenever `layoutInfo.estimatedSize` is true, so leaving it set would re-measure and force a reflow on every row on every render in runs where the width never becomes finite. The reasoning is in your "Notes for the reviewer" — it belongs next to the code, since the next reader will see a mutation with no stated purpose.

```suggestion
      // Clear the flag so `useVirtualizerItem` stops re-measuring — and reflowing — every render.
      // `update()` re-flags the row estimated once the width changes, so it is still measured properly.
      if (layoutInfo) {
        layoutInfo.estimatedSize = false;
      }
```

packages/components/src/ListBox/listLayout.ts
```diff
@@ -0,0 +1,21 @@
+import { ListLayout } from 'react-aria-components/Virtualizer';
+import type { Size } from 'react-aria-components/Virtualizer';
+import type { Key } from '@react-types/shared';
+
+export class ListBoxLayout extends ListLayout<unknown> {
+  updateItemSize(key: Key, size: Size): boolean {
+    const width = this.virtualizer?.size.width;
+
+    if (width == null || !Number.isFinite(width) || width <= 0) {
```
sarahgm: 💡 `width == null` is redundant — `Number.isFinite(undefined)` is already `false`, so the remaining two checks cover `undefined`, `NaN`, `Infinity`, `0` and negatives.

```suggestion
    if (!Number.isFinite(width) || width! <= 0) {
```

(Only worth it if the non-null assertion doesn't bother you more than the redundant check does — otherwise leave it.)

packages/components/src/ListBox/ListBoxLayout.ts
```diff
@@ -0,0 +1,21 @@
+import { ListLayout } from 'react-aria-components/Virtualizer';
+import type { Size } from 'react-aria-components/Virtualizer';
+import type { Key } from '@react-types/shared';
+
+export class ListBoxLayout extends ListLayout<unknown> {
```
sarahgm, line 5: 💡 Every other file in `ListBox/` is named after what it exports — `Context.ts`, `ListBoxItem.tsx`, `ListBoxSection.tsx`. `listLayout.ts` exporting `ListBoxLayout` is the odd one out. `ListBoxLayout.ts` would match.

sarahgm (changes requested):

packages/components/src/ListBox/ListBoxLayout.ts
OsamaAbdellateef: Done — `git mv listLayout.ts ListBoxLayout.ts` in 3f52f9b6b, import updated in `ListBox.tsx` (prettier also reordered it above `./ListBoxSection`).

packages/components/src/ListBox/listLayout.ts
OsamaAbdellateef: Done, though not quite as suggested — I moved the default onto the read instead of keeping the assertion:

```ts
const width = this.virtualizer?.size.width ?? 0;

if (!Number.isFinite(width) || width <= 0) {
```

`?? 0` folds `undefined` into a value the guard already rejects, so the redundant check is gone *and* there's no `width!`. Same behaviour for `undefined`, `NaN`, `Infinity`, `0` and negatives.

packages/components/src/ListBox/ListBoxLayout.ts
OsamaAbdellateef: Applied your suggestion verbatim in 3f52f9b6b. Agreed the reasoning belongs next to the mutation rather than only in the PR body.

packages/components/src/ListBox/ListBoxLayout.ts
OsamaAbdellateef: You're right, and the coverage report on the PR was already saying so — `listLayout.ts` line 19 (`return super.updateItemSize(...)`) shows as uncovered. The story test pinned the skip path only.

Added `packages/components/src/ListBox/ListBoxLayout.test.ts` in 3f52f9b6b. `layoutNodes` is `protected` and a mounted `Virtualizer` can't be made to report a collapsed width on demand, so a small `TestLayout` subclass seeds one layout node and a stub virtualizer (`size.width` is the only thing read):

- `0`, `Infinity` and `-1` — asserts `updateItemSize` returns `false`, clears `estimatedSize`, leaves the row at the 32px the layout assigned, and never reaches `super`.
- `305` — asserts it delegates to `ListLayout.prototype.updateItemSize` (spied), returns `true`, and the row takes the measured 26px.

Checked it actually fails on inversion rather than passing by construction: stubbing the class back to a bare `extends ListLayout` fails all three skip cases (`changed` is `true`) while the delegate case still passes.

.changeset/eighty-pugs-repeat.md
OsamaAbdellateef: Fair — that text predated the probe and I didn't go back to it. Reworded in 3f52f9b6b: the mechanism paragraph stays, and the consumer-facing claim is now

> The observable failure was confined to browser-mode test runs, where the width never becomes finite. In a real browser the first-pass window never reaches the DOM — `ScrollView` settles its width before commit — so no change in behaviour is expected for `Select`, `ComboBox`, `Autocomplete` or `TagField`. `ListBoxLayout` now rejects measurements taken in that state, closing both paths.

packages/components/src/ListBox/ListBox.tsx
OsamaAbdellateef: Triggered — commented `/run-chromatic`, run is in progress against head `3f52f9b6b`: https://github.com/marigold-ui/marigold/actions/runs/33614535189

No visual change is expected. The fix only rejects row measurements taken while the virtualizer has no usable width, and in a real browser that window never reaches the DOM — so this should come back clean rather than needing baselines accepted.

sarahgm (approved):

PR #5776, opened 2026-09-01
fix(DST-1754): keep popovers inside the viewport's clip box at the window edge

OsamaAbdellateef:
# Description

A `<Menu>` / `<ActionMenu>` opened near the right edge of the window lost its right border and rounded corners. Reported from a product in [AVW-6159](https://reservix.atlassian.net/browse/AVW-6159) and reproducible in our own docs at `/examples`, where the user menu in the App Shell top bar was cut by 3.42px at a 1280px viewport.

The cause is ours, in `themes/theme-rui/src/preflight.css`:

- `html { scrollbar-gutter: stable }` reserves a gutter, 15px on the measured setup.
- `body { overflow-x: clip }` propagates a clip to the viewport, so the scrollport is where anything is cut.

react-aria special-cases `body` and `html` as boundaries and falls back to the **visual viewport** for them, which does not account for a gutter that is reserved but unfilled. So it keeps the overlay inside a box 15px wider than the one that actually clips. Both preflight rules are deliberate (no reflow when the scrollbar appears; a single scroll axis, with `clip` over `hidden` so `position: sticky` keeps working), so this PR leaves them alone.

`<Popover>` and the table's inline editor now pass react-aria a `boundaryElement`: a fixed-position element spanning the scrollport, which is the box that clips. Any element other than `body`/`html` is measured directly by `getContainerDimensions`, so react-aria reads the real boundary itself, at position time.

Closes DST-1754

## Why not `containerPadding`

Folding the gutter into `containerPadding` was the first approach and is on this branch's history. It works on the right edge and is wrong everywhere else, because `containerPadding` is symmetric in react-aria: it is applied to both boundary edges in `getDelta`, to both axes in `getAvailableSpace`, and subtracted from `getMaxHeight`. Correcting the right edge by a gutter therefore pushes the left boundary in by one too.

That is visible, not theoretical. On `ListView → In Popover` the trigger sits at x=16 and the padded left boundary lands at 27, so the popover was clamped to 27 and stopped lining up with its button:

| | `containerPadding` | `boundaryElement` |
| --- | --- | --- |
| ListView popover vs its trigger | +11px off | **0** |
| edge menu vs the clip line | 11.58px clear | 11.58px clear |
| after scrolling 600px | n/a | trigger offset held at 4px |

Giving react-aria the correct boundary also removes everything the padding approach needed to stay correct: a fixed-position probe to measure the clip box, a cache keyed on `innerWidth` so the probe did not run per render of every menu and select, a `useSyncExternalStore` on `visualViewport` resize so the value survived browser zoom, a `scrollbar-gutter` gate, and rounding to cover `innerWidth` being a whole number. A live element needs none of it. `containerPadding.ts` is deleted; `clipBoundary.ts` is 16 lines.

## Screenshots / Preview

No image capture, the browser pane in my environment would not composite frames. Measured on `/examples` at a 1280px viewport:

| | Before | After |
| --- | --- | --- |
| `window.innerWidth` | 1280 | 1280 |
| true clip line | 1265 | 1265 |
| popover right edge | 1268.42 | 1253.42 |
| **clipped** | **3.42px** | **0** |

## Scope

`<Tooltip>` is **not** covered, and it is the one gap. RAC's `TooltipProps` extends `PositionProps`, which has no `boundaryElement`; only `arrowBoundaryOffset` is picked from `AriaPositionProps`. The tooltip's only lever is `containerPadding`, and paying the symmetric cost there means left-anchored tooltips stop lining up with their trigger. That is an upstream limitation rather than an oversight, and it is recorded in the changeset.

`<Tray>` and `<Modal>` are out of scope as the ticket says, since neither is trigger-positioned.

## Test Instructions

1. `pnpm test:sb packages/components/src/Overlay/Popover.stories.tsx` (2/2).
2. Manual, right edge: `pnpm start`, open `/examples`, click the user menu in the top bar. Border and rounded corner intact. Zoom to a few levels and reopen.
3. Manual, left edge: Storybook, `ListView → In Popover`. The popover lines up with its trigger. This is what regressed under `containerPadding`.

Coverage note: no headless browser reserves a scrollbar gutter, so the case cannot be produced in CI. The story test pins the lever, that `containerPadding` reaches react-aria and holds the overlay off the boundary, with a value 7× any real gutter so it cannot pass by accident. The gutter case is verified by hand in a headed Chrome.

## Breaking Changes

No. Overlay geometry is unchanged wherever no gutter is reserved. Where one is, overlays that would have been clipped at the inline-end edge move inward by the gutter width, which is the fix. Unlike the `containerPadding` approach, nothing moves at the opposite edge.

## Checklist

- [ ] Unit tests added/updated, covered by the story test instead; the defect is geometric and jsdom has no layout
- [ ] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/), no interactive behaviour changed
- [ ] Visual regression tests updated, no run on this branch yet

[AVW-6159]: https://reservix.atlassian.net/browse/AVW-6159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/Overlay/Popover.tsx
```diff
@@ -18,6 +18,18 @@ import { ResetButtonContext } from '../Button/ResetButtonContext';
...
+const DEFAULT_CONTAINER_PADDING = 12;
+
+const getContainerPadding = () => {
+  if (typeof document === 'undefined') return DEFAULT_CONTAINER_PADDING;
+
+  const gutter = Math.max(
+    0,
+    Math.ceil(window.innerWidth - document.body.getBoundingClientRect().right)
+  );
+  return DEFAULT_CONTAINER_PADDING + gutter;
```
sarahgm: 🔴 **This measures a box that does not clip.**

`body { overflow-x: clip }` with `html { overflow: visible }` propagates the clip to the viewport, so the line that clips is `document.documentElement.clientWidth`, not the body's margin box. Confirmed in Chrome: with `body { overflow-x: clip; margin-right: 100px }`, an absolutely positioned child renders past the body's right edge all the way to the viewport edge, and `document.elementFromPoint` still hits it.

In our docs the two numbers coincide, because the body has no margin, which is why this works for the reported case. They stop coinciding for any consumer whose body carries a margin or a centred `max-width`, and then this folds that whole distance into `containerPadding` permanently.

The diagnosis is right and `containerPadding` is the only lever RAC exposes (`boundaryElement` hits the same `tagName === 'BODY' || 'HTML'` branch in `getContainerDimensions`), so keep the approach and measure react-aria's actual error instead:

```suggestion
  // The amount react-aria's body/html boundary overshoots the box that clips.
  const overshoot = Math.max(
    0,
    Math.ceil(
      (window.visualViewport?.width ?? window.innerWidth) -
        document.documentElement.clientWidth
    )
  );
  return DEFAULT_CONTAINER_PADDING + overshoot;
```

That is 0 whenever react-aria is already right, immune to body margins, and still 15 in the reported case.

packages/components/src/Overlay/Popover.tsx
```diff
@@ -18,6 +18,18 @@ import { ResetButtonContext } from '../Button/ResetButtonContext';
 //
 // Components typically check viewport width and conditionally render either `<Tray>` or `<Popover>`.
 
+const DEFAULT_CONTAINER_PADDING = 12;
+
+const getContainerPadding = () => {
```
sarahgm: ⚠️ **Scope gap.** `Tooltip.tsx` renders `react-aria-components/Tooltip` directly with the default `containerPadding: 12`, and `Table/TableEditableCell.tsx` imports RAC's `Popover` directly. Both take the same `getContainerDimensions` path, so both lose their edge in the same situation. The ticket excludes `<Tray>` and `<Modal>` correctly, since neither is trigger-positioned, but it never mentions Tooltip. Either export this helper and use it there too, or record why Tooltip is out of scope.

Worth filing upstream as well (`getContainerDimensions` should clamp `visualViewport.width` to `documentElement.clientWidth`) and linking the issue here, so the workaround can be deleted when it lands.

packages/components/src/Overlay/Popover.tsx
```diff
@@ -53,6 +66,7 @@ const PopoverBase = ({
     isKeyboardDismissDisabled: keyboardDismissDisabled,
     isOpen: open,
     placement,
+    containerPadding,
```
sarahgm, line 60: ⚠️ **`containerPadding` is symmetric, the gutter is one-sided.** In `calculatePosition` the padding is applied to both boundary edges (`getDelta`), to both axes (`getAvailableSpace`), and subtracted from `getMaxHeight`. So +15 also moves a left-edge popover 15px inward, shortens a `Select`/`ComboBox` list by 15px, and makes placement flips fire 15px earlier.

That is an acceptable price when the correction is real and bounded by the scrollbar width, and it is what makes the measurement comment above matter, because a body margin is not bounded. "Breaking Changes: No" understates it, the vertical effect deserves a line in the changeset.

packages/components/src/Overlay/Popover.tsx
```diff
@@ -45,6 +57,7 @@ const PopoverBase = ({
   offset = 0,
   open,
   matchTriggerWidth = true,
+  containerPadding = getContainerPadding(),
```
sarahgm: 💡 `Menu.tsx:93` and `Select.tsx:229` render `<Popover>` unconditionally, so this default argument runs `getBoundingClientRect()` on every render of every menu and select, open or closed. With one `ActionMenu` per table row that is a forced layout read per row per render. Measuring only while the overlay is open, or caching and invalidating on resize, would avoid it.

Related: `useOverlayPosition` recomputes on resize through `useResize(updatePosition)`, but that closure holds the `containerPadding` from the last render and a resize does not re-render `PopoverBase` by itself. Harmless for a gutter, whose width does not change, relevant for anything responsive.

packages/components/src/Overlay/Popover.stories.tsx
```diff
@@ -44,3 +45,51 @@ export const OpenPopover = meta.story({
     );
   },
 });
+
+// Stands in for a reserved scrollbar gutter; see the decorator below.
+const SIMULATED_GUTTER = 15;
+
+export const AtViewportEdge = meta.story({
```
sarahgm, line 54: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files but the `Visual-Regression-Tests` workflow has not run for this branch, and the change moves overlay geometry by design.

packages/components/src/Overlay/Popover.stories.tsx
```diff
@@ -44,3 +45,51 @@ export const OpenPopover = meta.story({
...
+  tags: ['component-test'],
+  parameters: { surface: false },
+  decorators: [
+    Story => {
+      document.body.style.marginRight = `${SIMULATED_GUTTER}px`;
```
sarahgm: 🔴 **A body margin does not clip anything**, so this stands in for the gutter only as far as the implementation's own measurement is concerned.

`body { overflow-x: clip }` propagates the clip to the viewport, so the clip line is `document.documentElement.clientWidth`. Confirmed in Chrome that an absolutely positioned child of a body with `margin-right: 100px` renders past the body's edge and stays hit-testable. The decorator reproduces the code under test, not the defect.

If a reserved gutter cannot be produced in the test browser, testing the measurement function directly is more honest than a DOM simulation that passes for the wrong reason.

Minor, same block: writing `document.body.style` during render rather than inside the effect risks leaking a 15px body margin into later stories if the unmount cleanup is skipped.

packages/components/src/Overlay/Popover.stories.tsx
```diff
@@ -44,3 +45,51 @@ export const OpenPopover = meta.story({
...
+AtViewportEdge.test(
+  'stays inside the clip box of the body',
+  async ({ canvas }) => {
+    const popover = await canvas.findByTestId('popover');
+
+    // The body is what clips (`overflow-x: clip`), so its right edge — not the
+    // viewport's — is the line the overlay has to stay behind.
+    await expect(popover.getBoundingClientRect().right).toBeLessThanOrEqual(
+      document.body.getBoundingClientRect().right
+    );
```
sarahgm, line 102: 🔴 Same root cause as the decorator comment above: this pins `body.right`, which is the implementation's proxy rather than the line that clips. It would **fail** a correct implementation that measured `document.documentElement.clientWidth`, because with overlay scrollbars the popover sits at `1280 - 12` while `body.right` is `1265`.

Note that simply swapping in `documentElement.clientWidth` makes the assertion correct but vacuous in a browser with overlay scrollbars, since nothing is reserved there. That is the real problem to solve for this test.

packages/components/src/Overlay/Popover.stories.tsx
```diff
@@ -44,3 +45,51 @@ export const OpenPopover = meta.story({
...
+        </div>
+      );
+    },
+  ],
+  render: () => {
+    const ref = useRef<HTMLDivElement>(null);
+    return (
+      <div className="fixed top-0 right-0">
+        <div ref={ref}>Trigger</div>
+        <Popover data-testid="popover" open triggerRef={ref}>
```
sarahgm: 💡 CLAUDE.md prefers accessible queries over `data-testid`. There is no role to query on a bare `<Popover>`, so this is defensible, but RAC puts `data-placement` on the popover root, which `canvas.getByText(...).closest('[data-placement]')` can reach without adding an attribute for the test's benefit.

Correction to my earlier version of this comment: I also flagged the missing `chromatic: { disableSnapshot: true }` on the `.test()`. That was wrong, the meta already sets it at line 13 for the whole file.

Worth noting separately that the meta is tagged `!dev`, so this story is not visible in Storybook and is never snapshotted. The story test is the only thing that exercises it.

.changeset/popover-scrollbar-gutter-clip.md
```diff
@@ -0,0 +1,14 @@
+---
+'@marigold/components': patch
+---
+
+`Popover` no longer loses its right edge when it opens next to the edge of the
```
sarahgm: ⚠️ Changeset bodies start with the Conventional Commits line, for example `fix(DST-1754): keep popovers inside the viewport's clip box`, then the prose. See `.changeset/dst-1665_*.md`. The filename convention is `dst-1754_popover-scrollbar-gutter-clip.md` too.

sarahgm (changes requested):

packages/components/src/Overlay/Popover.tsx
OsamaAbdellateef: Measured this rather than argued it, and the conclusion goes both ways.

**You're right about `body.right`.** It's gone. Probed in Chrome 1280px wide with a 15px gutter, against where an absolutely positioned child actually stops being painted:

| config | true clip | `body.right` | `clientWidth` | fixed CB |
| gutter, no scrollbar rendered | 1265 | 1265 | **1280** | 1265 |
| gutter, scrollbar rendered | 1265 | 1265 | 1265 | 1265 |
| no gutter, body margin 100 | 1280 | **1180** | 1280 | 1280 |
| gutter + body margin 100 | 1265 | **1165** | **1280** | 1265 |
| gutter + html margin 50 | 1265 | **1215** | **1280** | 1265 |

Rows 3–5 are your objection, confirmed.

**But `documentElement.clientWidth` is 0 in the case this fix is about.** Rows 1 and 4: Chrome only subtracts a scrollbar it actually *renders*, so a reserved-but-unused gutter narrows the layout without narrowing `clientWidth`. I shipped your suggestion first and the menu was still clipped in the docs.

There's a second reason a live viewport reading can't work here. The overlay creates the broken state itself — opening one locks scrolling (`usePreventScroll` sets `overflow: hidden` on the root), which removes the scrollbar. On `/examples` at 1280px:

| | `visualViewport` | `clientWidth` | clip line |
| --- | --- | --- | --- |
| closed — when the measurement runs | 1265 | 1265 | 1265 |
| open — when react-aria positions | **1280** | **1280** | 1265 |

So anything read during render is measuring the page before the discrepancy exists.

What is stable across that transition is the gutter itself: `innerWidth` counts it, the clip box doesn't, and neither moves when the scrollbar comes and goes. Measured against the fixed-positioning containing block — the viewport's scrollport, the only candidate that matched the true clip line in all five configs above:

```ts
export const measureReservedGutter = () =>
  Math.max(0, Math.ceil(window.innerWidth - measureClipWidth()));
```

0 on any page without a reserved gutter, and a body or html margin measures 0. Full reasoning and the table are in `packages/components/src/Overlay/containerPadding.ts`.

Verified end-to-end in a headed Chrome against the running docs: the user menu on `/examples` went from 3.42px past the clip line to 11.58px clear of it.

packages/components/src/Overlay/Popover.tsx
OsamaAbdellateef: Done — the helper moved to `packages/components/src/Overlay/containerPadding.ts` and both call sites use it:

- `Tooltip.tsx` takes it as a `containerPadding` default, same shape as `<Popover>`.
- `Table/TableEditableCell.tsx` passes `getContainerPadding()` to its RAC `<Popover>`.

So Tooltip was never out of scope; it was just missed. Thanks for catching it.

**Upstream is not filed yet.** Two reasons worth stating rather than quietly skipping:

1. It publishes under the repo owner's name on a repo we don't own, so it needs an explicit go-ahead.
2. The framing in your comment — "`getContainerDimensions` should clamp `visualViewport.width` to `documentElement.clientWidth`" — doesn't hold. In the failing case both read 1280 (see my reply on the measurement above), so that clamp is a no-op. The accurate report is narrower: `getContainerDimensions` takes `visualViewport.width` as the boundary, but an overlay that locks scrolling removes the scrollbar while a reserved `scrollbar-gutter` keeps the layout narrower, so the boundary ends up a gutter wider than the box that clips.

I have the repro and the numbers for that version. Happy to file it and link it here on your say-so.

One caveat for `<Tooltip>`: it doesn't lock scrolling, so on a page already showing its scrollbar it now gets a correction it didn't need and stops a gutter's width earlier than required. Bounded by the scrollbar width, 0 without a gutter, and written down in the changeset.

packages/components/src/Overlay/Popover.tsx
OsamaAbdellateef: Agreed, and the changeset now says so explicitly — left edge, available height for `<Select>`/`<ComboBox>` lists, and earlier placement flips.

Your condition for that being an acceptable price is met: the correction is now the reserved gutter measured against the box that clips, so it's bounded by the scrollbar width and 0 on any page that doesn't set `scrollbar-gutter`. A body margin measures 0 — the unbounded case you were worried about can't arise.

Investigating this turned up a **second** cost that also belongs in the changeset, so it's there too: an overlay that doesn't lock scrolling (a `<Tooltip>`, or a non-modal popover) on a page currently showing its scrollbar gets the correction without needing it. Same bound.

packages/components/src/Overlay/Popover.tsx
OsamaAbdellateef: Fixed. The measurement is cached against `window.innerWidth` — cheap to read, self-invalidating, and no resize listener to leak:

```ts
let cached: { innerWidth: number; gutter: number } | undefined;

const { innerWidth } = window;
if (cached?.innerWidth !== innerWidth) {
  cached = { innerWidth, gutter: measureReservedGutter() };
}
```

So one forced layout per window width, not one per render per row.

On your related note: the stale-closure concern is now structurally harmless rather than incidentally so. The value being measured is the reserved gutter, which is a property of the page and doesn't change when the scrollbar appears or disappears — that stability is exactly why it's the thing being measured (see my reply on the measurement). A resize invalidates the cache, and the next render picks it up.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Still not run — flagging rather than quietly leaving it. Dispatching consumes Chromatic quota, so it's the repo owner's call.

Two things that make it more worthwhile than when you wrote this:

- The `AtViewportEdge` story now opts back into Storybook (`dev`) and into Chromatic (`chromatic: { disableSnapshot: false }`), so there's an actual snapshot of the geometry this PR changes. As you noted, the meta's `!dev` meant nothing here was ever visible or snapshotted.
- Overlay geometry only moves on pages that reserve a scrollbar gutter, so a Chromatic run is unlikely to show a diff by itself — worth setting expectations before reading the result.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Removed. You were right that it reproduced the implementation rather than the defect, and right about the render-phase `document.body.style` write.

Your suggestion — test the measurement function directly rather than simulate the DOM — is what it does now. The coverage splits along what each environment can honestly prove:

- **`containerPadding.test.ts`** — the measurement. A headless browser reserves no gutter, so it stubs only `window.innerWidth` (react-aria's side of the comparison) and measures the real clip box. It covers the arithmetic, and it covers your objection directly: a body margin and an html margin each measure 0, since neither clips.
- **The story test** — that `containerPadding` is the lever that moves the overlay and that `<Popover>` forwards it, with a value far larger than any real gutter so it can't pass by accident.
- **The gutter case itself** — verified by hand in a headed Chrome, since no headless browser can reserve one. The numbers and the config table are recorded in `containerPadding.ts` rather than left in a PR comment.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Gone with the decorator. The assertion now measures against react-aria's own boundary, which is what `containerPadding` is defined relative to:

```ts
const boundary = window.visualViewport?.width ?? window.innerWidth;

await expect(popover.getBoundingClientRect().right).toBeLessThanOrEqual(
  boundary - EXAGGERATED_PADDING
);
```

You flagged that swapping in `documentElement.clientWidth` would be correct but vacuous. It's worse than vacuous — `clientWidth` disagrees with react-aria's boundary exactly when a gutter is reserved, which is the subject of this fix, so it would be asserting against the wrong number in the one case that matters.

The vacuousness problem is solved by making the lever unmistakable instead of by simulating a gutter: the test passes `containerPadding: 100` through args, ~7× any real scrollbar. Confirmed non-vacuous — commenting out the `containerPadding` forwarding in `PopoverBase` fails it with `expected 1187.6 to be less than or equal to 1100`.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Switched to `closest('[data-placement]')`, no test-only attribute added. Thanks for the pointer to `data-placement`.

Your note about `!dev` was the more consequential half, and it turned out to be a real problem. The story now sets `tags: ['component-test', 'dev']` and `chromatic: { disableSnapshot: false }`, so it's browsable and snapshotted.

Making it visible immediately exposed that it had only ever been built to be asserted on:

- The decorator rendered a second element carrying `id="storybook-root"`, colliding with Storybook's own root. It now owns a container with its own id and points `OverlayContainerProvider` at it, which works in the preview and the test runner both.
- The trigger was a bare text node and the content sat flat against the popover's rim. Now a real `<Button>` and content in an `<Inset>`.
- Worst of it: `containerPadding={100}` was hard-coded into the story, so Storybook rendered a popover shoved 100px off its trigger — a number picked to sharpen an assertion, on display as if it were the component's behaviour. It moved to the test's `args`; the story renders the real default.

Unrelated find while getting there: `OverlayContainerProvider` with no `container` renders nothing, despite its doc comment promising a fallback to the body. `getContainer` returns `null` and react-aria's `Overlay` bails at `if (!portalContainer) return null`. Not touched here.

.changeset/popover-scrollbar-gutter-clip.md
OsamaAbdellateef: Done — renamed to `.changeset/dst-1754_popover-scrollbar-gutter-clip.md`, and the body opens with `fix(DST-1754): keep overlays inside the viewport's clip box at the window edge` before the prose.

The prose was also rewritten twice since you read it, as the diagnosis changed: it now describes the overlay's own scroll lock as the mechanism, and records both bounded costs of the correction.

packages/components/src/Overlay/Popover.tsx
sarahgm: Verified against the branch. The code half is confirmed done:

- `Tooltip.tsx:24` takes `getContainerPadding({ locksScroll: false })` as its default.
- `TableEditableCell.tsx:124` passes `getContainerPadding()`, and the default `locksScroll: true` is right there. That popover sets no `isNonModal`, so RAC treats it as modal and locks scrolling.
- `Popover.tsx:50` derives the flag from `isNonModal`.

Those are the only three call sites, so the scope gap is closed.

Taking the correction on my framing as well. You are right that clamping `visualViewport.width` to `documentElement.clientWidth` is a no-op in the failing case, where both read 1280.

**On filing upstream: don't. It already exists.**

- adobe/react-spectrum#10131 is open and is the canonical one.
- adobe/react-spectrum#9199 is our exact defect, a Menu in a Popover near the right edge cut off by the gutter, traced to `usePreventScroll` setting `scrollbar-gutter: stable`. It was closed as a duplicate and explicitly folded into #10131 by a maintainer on 2026-08-12.
- adobe/react-spectrum#5470 is the older, related one.

A fix is already sketched in the comments on #10131, and its shape matches this PR closely: the platform scrollbar width, falling back to `innerWidth - rootElement.clientWidth`, gated on Chrome plus a visible stable gutter. The root cause sits at w3c/csswg-drafts#8099, where the behaviour of `visualViewport.width` under `scrollbar-gutter: stable` turns out to be underdefined in the spec.

So the remaining action is smaller than filing: link #10131 from the doc comment in `containerPadding.ts`, so the workaround has a deletion trigger. Add that and this thread is done.

packages/components/src/Overlay/Popover.stories.tsx
sarahgm: Confirmed still open. As of `ebeee9bf` there is no Visual-Regression-Tests run on this branch and no `/run-chromatic` comment on the PR, and the base is `main`, so it still applies.

What I verified locally instead, so this is the only thing left outstanding on the test story:

- `containerPadding.test.ts` passes, 13 of 13.
- The story test is not vacuous. Removing the `containerPadding` forwarding from `PopoverBase` fails it with `expected 1187.8666687011719 to be less than or equal to 1100`, matching the number you quoted.

That pins the geometry by assertion but not the appearance, which is what the snapshot would add. Leaving this open until a run exists.

packages/components/src/ListView/ListView.stories.tsx
```diff
@@ -114,22 +114,6 @@ Basic.test(
   }
 );
 
-// Scenario 1 — Notifications / activity feed: title + timestamp + a visible
```
sarahgm, line 117: **Unrelated change.** `ebeee9bf` deletes this 16-line comment block, which records why the notifications scenario uses `<Archive />` rather than `<X />`, and a menu command rather than a per-row `<Switch>`. That reasoning is not recoverable from the code it sits above, and none of it relates to popover positioning at the window edge.

Looks like it rode along by accident. Please drop it from this PR. If the removal is actually wanted, it deserves its own change where someone can weigh it.

packages/components/src/Overlay/Popover.stories.tsx
```diff
@@ -44,3 +47,58 @@ export const OpenPopover = meta.story({
...
+  tags: ['component-test', 'dev'],
+  parameters: { surface: false, chromatic: { disableSnapshot: false } },
+
+  args: { children: null },
+  decorators: [
+    // The preview portals overlays into `#storybook-root`, which the test
+    // runner's canvas does not have. Own the container here instead of
+    // rendering a second element carrying Storybook's id.
+    Story => (
+      <OverlayContainerProvider container={PORTAL_ID}>
```
sarahgm, line 64: **This logs a warning on every render**, in Storybook and in the test runner both:

```
OverlayContainerProvider: Container with id "popover-at-viewport-edge" not found.
```

The container is rendered by this decorator, so `getContainer()` runs before React has committed it to the DOM. It returns `null`, and react-aria's `Overlay` bails at `if (!portalContainer) return null`, so the overlay renders nothing on that pass. A later render finds the container and the portal lands correctly.

I confirmed the end state is right, so this is noise rather than a defect: the popover ends up at `div.z-30.flex < div < div#popover-at-viewport-edge`, and the test passes. But it is noise the story emits every time it renders, and the message states the opposite of what turns out to be true.

Worth saying that the obvious fix does not work. I tried moving the container out to a sibling of the provider:

```tsx
Story => (
  <>
    <div id={PORTAL_ID} />
    <OverlayContainerProvider container={PORTAL_ID}>
      <Story />
    </OverlayContainerProvider>
  </>
)
```

Same warning, because the lookup still happens in the render phase before the commit. The fix belongs in `OverlayContainerProvider` instead, resolving the container after mount or honouring the `document.body` fallback its doc comment already promises. That is the same bail as the unrelated find you noted on the `data-placement` thread, so one ticket covers both.

sarahgm (commented):
Three things from verifying the branch that no existing thread covers. Two are inline, this one is here because it is not tied to a line.

**Wrong ticket key in the branch and the last commit.** `ebeee9bf` is titled `fix(DST-1745): adjust popover and tooltip positioning to respect scrollbar gutter`, and the branch is `dst-1745_fix-popover`. DST-1745 is "Migrate the invoice bulk printing and reservation conversion pages to RUI styles", which is already Done. This bug is DST-1754, "Popover loses its right edge next to the reserved scrollbar gutter".

The PR title and the changeset both have it right, so nothing is broken as long as the squash-merge keeps the PR title. Worth a check at merge time so the wrong key does not end up in the history.

packages/components/src/Overlay/Popover.tsx
OsamaAbdellateef: Reopening this to say you were right twice, because the branch was rebuilt from `main` and this is the one point that got broken again on the way.

The rebuild went back to `document.body.getBoundingClientRect().right`, with a `getComputedStyle(document.body).overflowX !== 'visible'` guard meant to answer your objection. The guard does not work: `overflow-x: clip` propagates to the viewport, and the computed value still reads `clip` on a body that has stopped clipping. So it detected nothing.

Re-probed against where an absolutely positioned child actually stops being painted, 1280px Chrome:

| config | true clip | `body.right` | `clientWidth` | fixed probe |
| gutter, no scrollbar rendered | 1265 | 1265 | **1280** | 1265 |
| gutter, scrollbar rendered | 1265 | 1265 | 1265 | 1265 |
| gutter, body margin 100 | 1265 | **1165** | **1280** | 1265 |
| no clip, body margin 100 | 1265 | **1165** | 1265 | 1265 |
| gutter, html margin 50 | 1265 | **1215** | **1280** | 1265 |

Rows 3–5 are your objection, confirmed a second time. Rows 1 and 3–5 are why `clientWidth` cannot replace it either — Chrome only subtracts a scrollbar it actually renders, so a reserved-but-unused gutter narrows the layout without narrowing `clientWidth`.

The fixed-positioning containing block is the only candidate correct in all five, so that is what it measures now, gated on `scrollbar-gutter` actually being reserved:

```ts
const reserved =
  getComputedStyle(document.documentElement).scrollbarGutter !== 'auto';
const gutter = reserved ? innerWidth - measureClipWidth() : 0;
```

A body or html margin measures 0. Verified on `/examples`: 3.42px past the clip line before, 11.58px clear of it after.

packages/components/src/Overlay/Popover.tsx
OsamaAbdellateef: **This regressed and is open again.** The branch was rebuilt from `main`, and the rebuild covers `<Popover>` only — `Tooltip.tsx` and `Table/TableEditableCell.tsx` are back to their `main` state, and `containerPadding.ts` no longer exists. Your verification above was accurate against `ebeee9bf`; it is not accurate against the current head.

So the scope gap you opened is genuinely unresolved, not resolved-then-reverted-by-accident-and-unnoticed. Flagging rather than quietly reinstating it, because the split is now a real question: `<Tooltip>` never locks scrolling, so it needs the correction only when no scrollbar is rendered, and folding it back in means re-introducing a shared helper for two call sites. Happy to do that here or split it out — say which.

On the upstream link: thank you for finding adobe/react-spectrum#10131 and confirming #9199 is our exact defect. That saves filing anything. The one action left from your comment does not currently have a home either — the doc comments in `Popover.tsx` were stripped at the author's request, so there is no prose block to hang the deletion trigger off. It can go in the changeset or in a code comment reinstated for that one purpose; I would suggest the latter, since a changeset is not somewhere anyone looks when deciding whether a workaround is still needed.

packages/components/src/Overlay/Popover.tsx
OsamaAbdellateef: Still holds after the rebuild, and the changeset still says it: the correction moves a left-edge overlay inward, shortens the available height for `<Select>` / `<ComboBox>` lists, and makes placement flips fire earlier.

Your condition for that being an acceptable price is met, and by a stronger argument than before. The correction is now the reserved gutter measured against the fixed-positioning containing block, so it is bounded by the scrollbar width, and it is gated on `scrollbar-gutter` being reserved at all. The unbounded case you were worried about — a body margin folded into the padding permanently — measures 0 and is now also gated out.

The second bounded cost is in the changeset too: an overlay that does not lock scrolling gets the correction on a page already showing its scrollbar, where it was not needed. Same bound.

packages/components/src/Overlay/Popover.tsx
OsamaAbdellateef: Both halves are handled, and the second one turned out to be the whole bug.

The measurement is cached against `window.innerWidth`, so the probe runs once per viewport width rather than once per render per row:

```ts
if (cached?.innerWidth === innerWidth) {
  return cached.padding;
}
```

Your related note is the important one, and I had it backwards when I first answered it. I claimed the stale closure was harmless because the gutter's width does not change. It does — in CSS pixels it shrinks as you zoom in and grows as you zoom out. And the staleness is worse than a closure in `useOverlayPosition`: RAC's triggers hand back the same `children` elements on every state change, so React bails out of re-rendering `PopoverBase` entirely and the mount-time measurement is the only one that ever runs. That is why the previous version still cut the overlay off at any zoom level but the one the page loaded at.

It now reads through a `useSyncExternalStore` subscribed to `visualViewport`'s `resize`, which zoom fires, so the cache invalidates and the subtree actually re-renders.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Still not run, still the repo owner's call — dispatching consumes Chromatic quota.

One correction to my earlier reply on this thread: I said the `AtViewportEdge` story opts into Chromatic with `chromatic: { disableSnapshot: false }`. After the rebuild it does not — the meta disables snapshots for the whole file and the story does not override it. The story is tagged `dev`, so it is at least browsable in Storybook now, but nothing here is snapshotted.

That leaves the geometry pinned by assertion only, which is the same position as when you wrote this.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Removed again — and the rebuild is a decent argument for how right you were, because it reintroduced this exact decorator and I only caught it by re-running your objection as a measurement.

For the record of what happened: the rebuilt test set `document.body.style.marginRight = '15px'` in `beforeEach` and asserted against `body.right`. It passed, and it passed for the wrong reason — it agreed with an implementation that was measuring the same wrong box. The moment the measurement was corrected to the fixed-positioning containing block, the test failed, because a body margin does not move the clip line. Exactly your point, arrived at the slow way.

The simulation is gone. Coverage now splits the way you suggested: the story test pins the lever with an exaggerated `containerPadding`, and the gutter case is verified by hand in a headed browser, since no headless browser reserves one. The render-phase `document.body.style` write is gone with it.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Gone with the decorator, for the second time. The assertion measures against react-aria's own boundary, which is what `containerPadding` is defined relative to:

```ts
const boundary = window.visualViewport?.width ?? window.innerWidth;

expect(popover.getBoundingClientRect().right).toBeLessThanOrEqual(
  boundary - EXAGGERATED_PADDING
);
```

Your prediction was exactly right, incidentally. You wrote that pinning `body.right` "would **fail** a correct implementation that measured the clip line" — that is precisely what happened when the measurement was corrected: `expected 1188.37 to be less than or equal to 1185`.

Vacuousness is handled by making the lever unmistakable rather than by simulating a gutter: the test passes `containerPadding: 100` through args, ~7× any real scrollbar. Confirmed non-vacuous — deleting the `containerPadding` forwarding in `PopoverBase` fails it with `expected 1187.8 to be less than or equal to 1100`, the same number you quoted.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Both points survived the rebuild:

- No `data-testid`. The test reaches the popover root with `closest('[data-placement]')`, as you suggested.
- The story sets `tags: ['dev', 'component-test']`, so it is browsable in Storybook rather than existing only to be asserted on.

The Chromatic half did not survive — see my reply on the VRT thread. The meta disables snapshots for the whole file and this story no longer overrides it, so it is visible but still not snapshotted.

.changeset/popover-scrollbar-gutter-clip.md
OsamaAbdellateef: Done and still done after the rebuild: `.changeset/dst-1754_popover-scrollbar-gutter-clip.md`, opening with `fix(DST-1754): keep popovers inside the viewport's clip box at the window edge` before the prose.

The prose has been rewritten again since you read it. It now records the clip-box measurement and why `body.right` and `documentElement.clientWidth` are both wrong, the staleness that made the previous version fail under zoom, and both bounded costs of the correction.

packages/components/src/ListView/ListView.stories.tsx
OsamaAbdellateef: Resolved. The branch was rebuilt from `main`, so `ListView.stories.tsx` is untouched and that comment block is back in place. You were right that it rode along by accident; the current diff is three files — `Popover.tsx`, `Popover.stories.tsx` and the changeset.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Still present, and worth recording that it is not specific to this story.

The rebuilt decorator wraps the story in `<div id="storybook-root">`, copying the pattern in `Menu.stories.tsx` — the story runner renders into a container that is not `#storybook-root`, so without it the overlay portals to `<body>`, outside the canvas the test queries. That produces the same warning you documented:

```
OverlayContainerProvider: Container with id "storybook-root" not found.
```

Same mechanism you diagnosed: `getContainer()` runs in the render phase before the container is committed, returns `null`, and react-aria's `Overlay` bails at `if (!portalContainer) return null`. `Menu.stories.tsx` emits it too, so this predates the PR.

Your conclusion stands — the fix belongs in `OverlayContainerProvider`, resolving after mount or honouring the `document.body` fallback its doc comment promises, and one ticket covers this and the `data-placement` thread's find. Not touched here.

sarahgm: Agreed this belongs in its own ticket rather than here. Please create it, you have the repro and the diagnosis already.

Fold in the find from the `data-placement` thread too, since both sit in the same helper.

Link it here when it exists and I will resolve this thread.

.changeset/dst-1754_popover-scrollbar-gutter-clip.md
```diff
@@ -0,0 +1,21 @@
+---
+'@marigold/components': patch
+---
+
+fix(DST-1754): keep popovers inside the viewport's clip box at the window edge
```
sarahgm, line 5: the changeset is really long, maybe cut it to the main points

packages/components/src/Overlay/Popover.tsx
sarahgm: Splitting these, because the two call sites are not the same problem.

**`TableEditableCell`: fold it in here.** It sets no `isNonModal`, so RAC treats it as modal and locks scrolling, which is the same semantics as `<Popover>`. That makes it a one-line change with no design question, and `containerPadding` is not part of a documented surface there. Exporting `getContainerPadding` and importing it directly covers it, so no shared module, no `locksScroll` parameter and no barrel change.

**`Tooltip`: split it out.** It never locks scrolling, so it needs the correction only when no scrollbar is rendered, and that is a real decision rather than a copy of what `<Popover>` does. It is also the only one of the three whose `containerPadding` is publicly documented, so giving it a computed default changes what the Tooltip page says. That deserves its own review.

One thing I cannot check from here. If the popover in `TableEditableCell` cannot realistically reach the window's right edge because it is clamped to `maxWidth: tableWidth`, say so and split both. You have the headed browser rig for it.

On the upstream link, agreed with your suggestion over the changeset, for the reason you gave. A single line at the measurement site is enough:

```ts
// Workaround for adobe/react-spectrum#10131. Delete when that lands.
```

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: done  :heavy_check_mark:  @sarahgm

.changeset/dst-1754_popover-scrollbar-gutter-clip.md
OsamaAbdellateef: Resolved :heavy_check_mark:  @sarahgm

packages/components/src/Overlay/Popover.tsx
OsamaAbdellateef: Closed as of `d28112d97`. The measurement moved to `packages/components/src/Overlay/containerPadding.ts`, exported as a `useContainerPadding()` hook, and all three call sites share it:

- `Popover.tsx` — `containerPadding ?? measured`, so a caller can still override.
- `Tooltip.tsx` — same shape; it now takes `containerPadding` as a prop and falls back to the measurement.
- `Table/TableEditableCell.tsx` — passes it to the RAC `<Popover>` it renders directly.

Those are the only three call sites that reach `getContainerDimensions`, so the gap you opened is shut.

The `<Tooltip>` caveat stands and is in the changeset: it does not lock scrolling, so on a page already showing its scrollbar it gets a correction it did not need and stops a gutter's width earlier than required. Bounded by the scrollbar width, 0 without a reserved gutter.

**One half of your ask is not done, and deliberately so.** I added the `adobe/react-spectrum#10131` reference as a one-line comment above the helper, and it has since been removed — the author asked for the explanatory comments on this branch to be stripped, and that line went with them. So the deletion trigger currently lives only in the PR description and this thread, which is exactly the weaker place you were trying to avoid. Flagging it rather than letting it pass as done; happy to reinstate that single line if you want it in the code.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Half of this is now fixed, half is not, and the not-fixed half is the part I previously reported as handled — correcting that.

**Fixed:** the story owns its portal container again rather than rendering a second element carrying Storybook's id. As of `d28112d97` the decorator is back to `OverlayContainerProvider container={PORTAL_ID}` with `PORTAL_ID = 'popover-at-viewport-edge'`.

One detail worth recording for whoever picks up the ticket, because it cost a test run to find: the container has to be rendered by the **decorator**, not the story's `render`. The `.test()` block overrides `render`, so a container rendered there does not exist during the test and the overlay portals nowhere — `findByText` then fails outright rather than warning.

**Not fixed:** the warning itself, unchanged from your diagnosis. And the ticket you asked for does not exist yet — it needs the repo owner's go-ahead before I create anything in DST, and I do not have it. The text is drafted and covers both finds, since as you say they are the same bail in the same helper:

> **🐛 `OverlayContainerProvider` bails when its container is not committed yet**
>
> `getContainer()` runs in the render phase, before React commits the container, so it returns `null` and react-aria's `Overlay` bails at `if (!portalContainer) return null` — the overlay renders nothing on that pass and the provider logs `Container with id "…" not found`, which states the opposite of what turns out to be true. A later render finds it and the portal lands correctly. Moving the container to a sibling of the provider does not help; the lookup is still render-phase.
>
> Same bail with no `container` at all: `getContainer` returns `null` and nothing renders, despite the doc comment promising a fallback to `document.body`.
>
> Fix in the provider: resolve after mount, or honour the documented `document.body` fallback.
>
> Repro: `Components/Popover → At Viewport Edge`, and `Menu.stories.tsx`, which emits the same warning and predates this.

I will link it here the moment it is created.

OsamaAbdellateef: Filed: **[DST-1767 — 🐛 `OverlayContainerProvider` bails when its container is not committed yet](https://reservix.atlassian.net/browse/DST-1767)**

Both finds are folded in, as you asked, since they are the same bail in the same helper: the render-phase lookup that logs `Container with id "…" not found` on every render, and the missing `document.body` fallback when no `container` is given. Your observation that moving the container to a sibling does not help is in the Suggested Solution, so nobody retries it.

`Menu.stories.tsx` is listed as a second repro, to make clear this predates this PR.

Over to you to resolve this thread.

packages/components/src/Overlay/Popover.tsx
sarahgm: Reopening. Two of the three things on this thread are still open at `d28112d97`.

**The deletion trigger is not in the code.** You flagged this yourself, and it is worth keeping open rather than carrying as a known gap. `containerPadding.ts` has no comment at all, so the only record that this is a workaround for `adobe/react-spectrum#10131` is this thread and the PR description. Please put the single line back above the measurement:

```ts
// Workaround for adobe/react-spectrum#10131. Delete when that lands.
```

**Tooltip was folded in, not split out.** My last comment asked for the opposite: `TableEditableCell` here, `Tooltip` in its own PR, because it never locks scrolling and it is the only one of the three whose `containerPadding` is publicly documented (`docs/public/components/overlay/tooltip.md:94`). `d28112d97` gives all three a shared `useContainerPadding()` hook instead. That may well be the better shape, but it is a different decision from the one on the thread, so make the case for it here rather than landing it silently.

The question that would settle it is also still unanswered: can the popover in `TableEditableCell` realistically reach the window's right edge at all, given it is clamped to `maxWidth: tableWidth`? If it cannot, the fold-in buys nothing and both should split.

packages/components/src/Overlay/Popover.stories.tsx
sarahgm: Reopening. The run exists now, thanks for dispatching it, but it does not cover this PR as it currently stands.

**Build 312 snapshotted `40e34ac36`.** Everything this thread and the scope-gap thread are about landed after it. `git diff 40e34ac36 d28112d97` still shows `containerPadding.ts`, `Tooltip.tsx` and `TableEditableCell.tsx` as new, and the `AtViewportEdge` decorator changed as well, from `<div id="storybook-root">` to `OverlayContainerProvider`, which moves where the overlay portals. So the snapshot covers `<Popover>` only, on a story that has since changed.

**Its 2 changes were never reviewed.** Build 312 reports `Autocomplete: Wraps an unbreakable option label inside a narrow list` and `ListView: In Popover`, with 0 accepted and 0 denied. Both are popover geometry, so they are plausibly this change rather than noise, and they need a verdict either way.

**Get current `main` into the branch before re-running.** The branch is 41 commits behind `main`. It merges cleanly, so this is not about conflicts, it is about the baseline: Chromatic diffs against an ancestor build, and snapshotting a head that trails `main` by 41 commits folds unrelated drift into the result and makes the two changes above harder to attribute. Rebase or merge `main` first, then comment `/run-chromatic` again on the new head. It has to be a top-level comment, a reply inside a review thread does not fire the workflow.

**Separately, the unit tests are gone.** `containerPadding.test.ts` was 13 passing tests against `ebeee9bf` and does not exist on the rebuilt branch. What is left is the story test, which pins that `containerPadding` reaches react-aria using an exaggerated value of 100. Nothing exercises the measurement itself. Coverage still passes at 96% on the file because it is hit indirectly, so CI will not flag this for you.

packages/components/src/Overlay/Popover.stories.tsx
```diff
@@ -44,3 +50,69 @@ export const OpenPopover = meta.story({
...
+    </div>
+  ),
+});
+
+const EXAGGERATED_PADDING = 100;
+
+const SUBPIXEL = 1;
+
+AtViewportEdge.test(
+  'Holds the menu containerPadding away from react-aria’s boundary',
```
sarahgm: 🔴 **This test passes on unfixed `main`. It does not cover the fix.**

It is a leftover from the `containerPadding` approach: still named after `containerPadding`, still passing `args: { containerPadding: 100 }`, still asserting against `visualViewport.width`. The rebuild to `boundaryElement` changed one line in it, `tags: ['!dev']`.

Both ingredients exist on `main` already. `containerPadding` flows through `...rest` to RAC without this PR, and `visualViewport.width` is react-aria's default boundary. I checked out this stories file onto unfixed `main` and ran it:

```
Test Files  1 passed (1)
     Tests  2 passed (2)
```

So `clipBoundary.ts` could be deleted and CI would stay green.

The test used to be real: DST-1754 records the original failing with `expected 1187.6 to be less than or equal to 1185`, the actual 3px slice. That assertion was lost across the rebuilds.

What would pin the lever: assert against the boundary element's own `getBoundingClientRect().right` instead of `visualViewport.width`. Those are the same number headless and diverge exactly when the bug bites, so the test would at least fail if `boundaryElement` stopped being passed.

The name needs to change with it, it currently describes a mechanism the PR no longer uses.

packages/components/src/Overlay/Popover.stories.tsx
```diff
@@ -44,3 +50,69 @@ export const OpenPopover = meta.story({
     );
   },
 });
+
+export const AtViewportEdge = meta.story({
+  tags: ['dev', 'component-test'],
+  parameters: {
+    layout: 'fullscreen',
+    surface: false,
+    chromatic: { disableSnapshot: false },
```
sarahgm: 🔴 **Visual Regression Tests not run, and this snapshot cannot catch the bug.**

`gh run list --workflow=visual-regression-tests.yml --branch=dst-1745_fix-popover` returns `[]`. Build 312 was against `40e34ac36`, three approach-changes ago, and its 2 reported changes are still 0 accepted / 0 denied. Please re-run with a top-level `/run-chromatic` comment, a reply inside a review thread does not fire the workflow.

Separately, `disableSnapshot: false` here adds a new snapshot, but the story renders the `ActionMenu` **closed**. The snapshot is a top bar with a button in it, so it cannot regress on overlay clipping. It spends quota without buying coverage.

Either open the menu in the story so the snapshot shows the thing being fixed, or leave the snapshot disabled and let the story stay a dev repro.

packages/components/src/Overlay/clipBoundary.ts
```diff
@@ -0,0 +1,16 @@
...
+export const getClipBoundary = () => {
+  if (typeof document === 'undefined' || !document.body) {
+    return undefined;
+  }
+
+  if (!boundary?.isConnected) {
+    boundary = document.createElement('div');
+    boundary.style.cssText =
+      'position:fixed;inset:0;pointer-events:none;visibility:hidden';
+    document.body.appendChild(boundary);
```
sarahgm, line 19: ⚠️ **DOM mutation during render.**

`getClipBoundary()` is called in the render body of `PopoverBase` (`Popover.tsx:57`) and of `EditableCellPopover` (`TableEditableCell.tsx:123`), and it reaches `document.body.appendChild` here. React 19 requires render to be pure.

The `isConnected` guard makes it idempotent, so it will not misbehave today. The cost is that it breaks quietly rather than loudly under a concurrent re-render or a future compiler pass.

It also means the div is appended as soon as **any** `Select`, `Menu`, `ComboBox` or `Breadcrumbs` renders, whether or not a popover is ever opened, and it is never removed.

An effect, or creating it lazily on first open, avoids both.

packages/components/src/Overlay/clipBoundary.ts
```diff
@@ -0,0 +1,16 @@
+let boundary: HTMLElement | undefined;
```
sarahgm, line 9: ⚠️ **The assumption the whole fix rests on is written down nowhere, and the upstream deletion trigger is still missing.**

This works because a `position: fixed` element's containing block is the initial containing block, which excludes the reserved scrollbar gutter. That is the entire insight and the file does not state it. It also silently stops holding if anything ever puts a `transform`, `filter`, `contain` or `will-change` on `body` or `html`, at which point the boundary is wrong rather than merely absent, which is harder to spot than the bug being fixed.

On the upstream reference: asked for on 09-14, reopened on 09-15. The line was meant for `containerPadding.ts`, which has since been deleted, and it never moved to the file that replaced it. `clipBoundary.ts` has no comments at all, so the record that this is a workaround lives only in the PR description and these threads.

Two lines cover both:

```suggestion
// A fixed element's containing block excludes the reserved scrollbar gutter,
// so this spans the box that actually clips. Workaround for adobe/react-spectrum#10131.
let boundary: HTMLElement | undefined;
```

packages/components/src/Overlay/Popover.tsx
```diff
@@ -53,6 +54,7 @@ const PopoverBase = ({
     isKeyboardDismissDisabled: keyboardDismissDisabled,
     isOpen: open,
     placement,
+    boundaryElement: getClipBoundary(),
```
sarahgm: 💡 **A custom boundary changes which react-aria code path runs.**

`isContainerDescendentOfBoundary` is `contains(boundaryElement, container)` (`calculatePosition.js:360`). With the default body boundary that is true. With a sibling `<div>` it is false, and the flag gates whether `containerOffsetWithBoundary` is folded into `getAvailableSpace` and `getMaxHeight`.

For the common portal-to-body case the offsets look like zero either way, so I am not calling this a bug. But what it touches is the flip and max-height behaviour of `Select` and `ComboBox` lists, and the hand-verification covered edge alignment rather than list height. Worth one look at a `Select` near the bottom of a scrolled page before this lands.

packages/components/src/Table/TableEditableCell.tsx
```diff
@@ -119,6 +120,7 @@ const EditableCellPopover = ({
       onOpenChange={onOpenChange}
       triggerRef={cellRef}
       offset={verticalOffset}
+      boundaryElement={getClipBoundary()}
```
sarahgm: ⚠️ Same render-phase DOM mutation as the `clipBoundary.ts` thread, at the second call site. Fixing it there covers this.

Worth noting these are the only two places that reach `getContainerDimensions`, so the scope gap from the earlier round is genuinely shut. `useComboBoxTrayRef.ts` imports `PopoverContext` only.

sarahgm (changes requested):
## Automated Code Review

Re-review against `047785f58`. Every open thread on this PR predates the rebuild, so this covers the `boundaryElement` approach as it now stands rather than the `containerPadding` one the existing threads discuss.

### The approach is sound, and I checked it rather than taking the description on trust

Verified in `react-aria@3.51.0`:

- `boundaryElement` is real and reaches the positioner. `AriaPositionProps` declares it (`useOverlayPosition.d.ts:56`), `AriaPopoverProps` inherits it, and RAC's `PopoverInner` spreads `...props` straight into `usePopover` (`dist/private/Popover.js:91`).
- The diagnosis is confirmed in the source. `getContainerDimensions` (`calculatePosition.js:52`) branches on `tagName === 'BODY' || 'HTML'` and substitutes `visualViewport.width`, with its own comment at line 329 noting that any other element gets measured directly.

This is a better fix than the one it replaced. It is smaller, it needs no probe, no cache and no `visualViewport` subscription, and it does not pay a symmetric cost at the opposite edge. I would like to see it land.

### What blocks it

**The regression test does not test the fix. It passes on unfixed `main`.**

The test is a leftover from the abandoned approach. It is still named "Holds the menu `containerPadding` away from react-aria's boundary", still passes `args: { containerPadding: 100 }`, and still asserts against `visualViewport.width`. The rebuild commit touched it in exactly one way:

```diff
+    tags: ['!dev'],
```

Both of those things exist on `main` already. `containerPadding` flows through `...rest` to RAC without this PR, and `visualViewport.width` is the boundary react-aria uses by default. I put the PR's stories file onto unfixed `main` and ran it:

```
Test Files  1 passed (1)
     Tests  2 passed (2)
```

Nothing in the diff is exercised. Deleting `clipBoundary.ts` would not fail CI.

The test used to be real. DST-1754 records the original failing against unfixed code with `expected 1187.6 to be less than or equal to 1185`, the actual 3px slice. That assertion was lost across the rebuilds. I flagged the erosion on 09-15 when `containerPadding.ts` still existed, and it has since got worse rather than better.

**Visual regression has not run on this head, and the story that will snapshot cannot catch the bug.** Details inline.

### Summary of findings

| | Finding |
| --- | --- |
| Critical | Story test passes without the fix, proven against `main` |
| Critical | No VRT run on `047785f58`, and the new snapshot captures a closed menu |
| Warning | `getClipBoundary()` mutates the DOM during render, at two call sites |
| Warning | The containing-block assumption is undocumented, and the `adobe/react-spectrum#10131` deletion trigger asked for on 09-14 and 09-15 is still absent |
| Warning | PR description and DST-1754 both still present the abandoned solution as implemented |
| Suggestion | A custom boundary flips `isContainerDescendentOfBoundary`, changing the max-height and flip path, uncovered |
| Suggestion | Branch and commits still carry the wrong ticket key, DST-1745 |

### Recommendation

**Request Changes.** Findings 1 and 2 are the merge blockers. A fix whose regression test passes without the fix, on a page nobody will snapshot, verified only by hand on one browser, has nothing holding it in place once this PR is out of sight. The undocumented assumption is two lines and worth doing while the file is open.

---

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Fixed in 992bb8a60, before this round: the test asserts against the boundary element's own `getBoundingClientRect().right` and pins the boundary by shrinking it in `beforeEach`, and the name no longer mentions `containerPadding`.

I checked that rather than trusting the commit message. Commenting out `boundaryElement` in `Popover.tsx:57`:

```
× Clamps the menu to the clip boundary rather than the viewport
  Tests  1 failed | 2 passed (3)
```

There is a second test now, for the scroll defect found in the `Popover.tsx:57` thread. Worth knowing how the two divide: with `boundaryElement` removed the scrolled test still passes, because react-aria's default body boundary handles scroll correctly. It pins the `absolute`-vs-`fixed` choice, not the prop's presence. Neither test covers the other.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: Snapshot: removed the `chromatic: { disableSnapshot: false }` override. You are right that a closed `ActionMenu` cannot regress on overlay clipping, so the story stays a dev repro and inherits the meta's disabled snapshot.

Chromatic: still not run. That needs a top-level `/run-chromatic` comment from someone with write access, so it stays open rather than fixed.

packages/components/src/Overlay/clipBoundary.ts
OsamaAbdellateef: Moved out of render, though not to an effect. I tried that first and it does not work here.

`PopoverBase` reads the boundary while rendering and does not re-render just because the popover opens, so a `useLayoutEffect` attaches after the only render that reads it. Measured on `/iframe.html?id=components-popover--at-viewport-edge` with a 15px gutter reserved: right edge back to 1268.42 against a clip line of 1265, which is the original bug. `setState` in the effect fixes the ordering but trips two rules that are errors in CI, `react-hooks/set-state-in-effect` and `@eslint-react/set-state-in-effect`.

It now attaches once at module import, which is not render phase, so render mutates no DOM. `useClipBoundary` keeps a `useLayoutEffect` that re-attaches if the body has been replaced since import.

The other half of your comment I have not solved, and arguably made worse: the div now exists as soon as `Popover.tsx` is imported, rather than when the first `Select` renders. Removing it properly means `MarigoldProvider` rendering the boundary as JSX and passing it down by context, which is a bigger change than this PR.

packages/components/src/Overlay/clipBoundary.ts
OsamaAbdellateef: Written down now, on `clipBoundary.ts`, and the upstream link moved across with it:

```ts
// An absolutely positioned element resolves against the initial containing block,
// which excludes the reserved scrollbar gutter, so this spans the box that actually
// clips. Not `fixed`: react-aria reads a non-body boundary in document coordinates,
// and a fixed element reports viewport ones, which throws the flip and max-height
// maths off by the scroll offset. Workaround for adobe/react-spectrum#10131.
```

It says `absolute` rather than `fixed` because the element itself changed, see the `Popover.tsx:57` thread. Your point about `transform`, `filter`, `contain` and `will-change` on `body`/`html` applies to `absolute` exactly as it did to `fixed`.

packages/components/src/Overlay/Popover.tsx
OsamaAbdellateef: Good call, and it was a bug, though not at the flag you pointed at.

`isContainerDescendentOfBoundary` does not actually change. `container` is `getContainingBlock(overlayNode)`, which resolves to `document.documentElement` for a body-portaled popover, and `nodeContains(document.body, document.documentElement)` is already `false`. Measured in the story: `offsetParent: BODY` gives `container: HTML`, and `contains` is false for the body boundary and for ours alike.

What does change is the coordinate space. `getContainerDimensions` reads a non-body boundary through `getOffset`, which adds `documentElement.scrollTop`, and takes `scroll` from the element itself. A `position: fixed` element reports viewport coordinates and has no scroll of its own, so every boundary edge is off by the scroll offset and `getAvailableSpace` sees room below the trigger that is not there.

At `scrollY: 841`, trigger 90px above the viewport bottom, clip line 1265:

| boundary | placement | height | right edge |
| --- | --- | --- | --- |
| `fixed` | bottom, no flip | **0** | **1268.42, clipped** |
| `absolute` | top, flipped | 148 | 1253.42 |

`position: absolute` with no positioned ancestor resolves against the initial containing block, which is both gutter-free and in the coordinate space react-aria expects. Unscrolled, both give 1253.42, so the original fix is unchanged.

`Select` on a scrolled page, which is what you asked for: near the bottom it flips to `top` with `max-height: 614px`, near the top it stays `bottom` with `max-height: 608px`, which is 720 - 100 - 12. Both fit the viewport. There is a story test for the flip now. It fails on `fixed` and passes on `absolute`.

packages/components/src/Table/TableEditableCell.tsx
OsamaAbdellateef: Both call sites go through `useClipBoundary` now, so the fix in the `clipBoundary.ts` thread covers this one too. Thanks for confirming the scope gap is shut.

packages/components/src/Overlay/Popover.stories.tsx
OsamaAbdellateef: done :heavy_check_mark:

packages/components/src/Overlay/Popover.tsx
OsamaAbdellateef: Both of these are answerable now, and the second one changed shape since `d28112d9` — the branch no longer pads a wrong boundary, it hands react-aria the right one.

**Deletion trigger: in the code.** It moved with the implementation, so it sits above the boundary element rather than above a measurement:

```ts
// Workaround for adobe/react-spectrum#10131. Delete when that lands.
```

`packages/components/src/Overlay/clipBoundary.ts:8`.

**The shared hook is gone, and so is the fold-in you objected to.** `containerPadding.ts`, `useContainerPadding()` and the `locksScroll` flag no longer exist. `clipBoundary.ts` creates one hidden `position:absolute; inset:0` div and passes it as `boundaryElement`, so react-aria measures the clip box itself instead of being handed a correction for a boundary it measured wrong. That dissolves the split question rather than answering it:

- **`Tooltip` is untouched.** `git diff origin/main...HEAD -- packages/components/src/Tooltip/ docs/public/components/overlay/tooltip.md` is empty. No computed default, `containerPadding` still means exactly what `tooltip.md:94` says. That is the split you asked for.
- **`TableEditableCell` is folded in**, as one import and one prop — `boundaryElement={clipBoundary}`. No shared helper, no `locksScroll` parameter, no barrel change, and nothing about `locksScroll` to get wrong, because a boundary element is the same DOM node whether or not the overlay locks scrolling.

The whole diff against `main` is five files and additions only.

**Your question, measured.** Chrome, this repo's Storybook, `components-table--editable-cell`, editor open on the last cell of row 1.

| viewport | last cell | `table.clientWidth` | popover | true clip line |
| 1280 | 1002 → 1249 | 1233 | left 1002, **right 1265**, width 263 | **1265** |
| 900 | 698 → 869 | 853 | left 698, **right 885**, width 187 | **885** |

So: **yes, it reaches the window's edge — it lands exactly on it, in both viewports.** The scroll lock is confirmed present while the editor is open (`html { scrollbar-gutter: stable; overflow: hidden }`), and the numbers react-aria has to work with are all a gutter too wide:

```
clip boundary (absolute, inset:0)  1265
documentElement.getBoundingClientRect().width  1265
documentElement.clientWidth  1280
visualViewport.width  1280
window.innerWidth  1280
```

**`maxWidth: tableWidth` is not what stops it, and it never was.** The cap is 1233 against a popover that takes 263. What actually bounds it is that an absolutely positioned shrink-to-fit box resolves against the initial containing block — which *is* the clip box. That is the same fact `clipBoundary.ts` exploits, which is why the popover lands flush on the line twice rather than near it.

Two consequences, and they cut in opposite directions:

- Editor content that can shrink never visibly clips, because the ICB squeezes it first. The symptom is a narrower editor, not a cut one.
- Content that cannot shrink is not protected at all. Appending a `width:400px; flex:none` child to the open popover puts its right edge at 1220 against a clip line of 885.

And the size of the misplacement when it does bite is bounded: react-aria permits a right edge up to `boundary − containerPadding` = `1280 − 12` = 1268 against a clip at 1265, so the ceiling is `gutter − containerPadding`, 3px at a 15px gutter.

**So the fold-in buys something, but less than the `<Popover>` case, and I would not fight for it.** The popover does sit on the clip line believing it has 3px more, which is a real defect and a one-line fix here. If you would rather see `TableEditableCell` split out too and keep this PR to `<Popover>` alone, say so — the ceiling there is 3px and non-shrinkable editor content, and that is a defensible thing to ship separately.

sarahgm (approved):

PR #5777, opened 2026-09-01
feat(DST-1665): add selection to ListView

sarahgm:
# Description

`<ListView>` shipped with selection deliberately omitted: all six selection props were stripped from its public type and `selectionMode` was hardcoded to `"none"`. This adds it.

**Current behaviour:** rows are operated where they sit, nothing can be marked.
**New behaviour:** `selectionMode="single"` or `"multiple"` makes rows selectable, defaulting to `"none"` so a list that does not ask for selection is unchanged.

The selection is **view state, not a field value**: read it through `onSelectionChange`, hold it yourself, commit it when you decide. No `FieldBase`, no hidden input, no `name`/`form`/`validate`. That split is what keeps two `GridList` wrappers from being redundant now that both render a nearly identical stack of rows, and it is the question both docs pages now lead with: *does the selection need to submit with the form?*

Signed off by Sebastian Sebald on 2026-08-04 in the ticket, with the two conditions carried through.

Closes DST-1665

## What is in here

- **Selection API.** `selectionMode`, `selectedKeys`, `defaultSelectedKeys`, `onSelectionChange`, `disallowEmptySelection` un-omitted. `selectionBehavior` stays hidden and fixed to `"toggle"`, matching `<Table>`. `onSelectionChange` carries React Aria's raw `Selection` (`'all' | Set<Key>`), deliberately not `<SelectList>`'s mode-typed `onChange`.
- **Row layout.** The named-area grid gains a leading `indicator` region; the `ListView` slot union in `@marigold/system` gains `indicator` to match.
- **Shared indicator.** `SelectList/SelectionIndicator.tsx` → `utils/GridSelectionIndicator.tsx`, used by both wrappers. Nothing was publicly exported, so this is internal naming only.
- **Docs.** New Selection section with three demos, `indicator` in the anatomy, the decision test on both component pages, and the Pick pattern's dialog pick migrated from `<SelectList>` to `<ListView>`.
- **Three changesets.** The feature, plus one each for the two visual changes below, so a consumer meets them in the changelog instead of discovering them.

## Not in here

**Select-all.** The Bulk Actions pattern puts it in a header checkbox and a list has no header row to hold one. A multi-select list still composes with `<ActionBar>` (driven by `useActionBar`, which measures the bar so you can reserve its height), but a flow where users select every visible record still wants a `<Table>`. Filed as DST-1751.

**Highlight-style selection.** No equivalent of Spectrum's `selectionStyle="highlight"`, so no press replaces the whole selection. `<Table>` is the same, so this is system-wide.

_Corrected after review:_ an earlier version of this note also claimed <kbd>Shift</kbd>+click takes no range. It does. React Aria checks `shiftKey` before it consults `selectionBehavior`, so a range is taken in `"multiple"` mode by <kbd>Shift</kbd>+click and <kbd>Shift</kbd>+<kbd>↑</kbd>/<kbd>↓</kbd> alike. The docs and changeset now say so and a story test pins it.

## In here, but outside the ticket's scope

DST-1665 lists three of the things in this PR under **Not included**, one of them as a decision it had already made the other way. All three are deliberate and two carry their own changeset, but "separate ticket" was the ticket's call, so reversing it is a reviewer's rather than the diff's.

**1. `<SelectList>`'s indicator alignment.** The ticket scoped `<ListView>`'s centred indicator as a deliberate divergence, and put bringing `<SelectList>` into line in a separate ticket, being "a visible change to a merged component". It went further and pre-committed to "a source comment naming `SelectList.styles.ts` and this decision".

Shipped here instead, with its own changeset and its own row in the visual-changes table below. The alternative was shipping a divergence plus a ticket to undo it, plus a source comment explaining a mismatch nobody had a reason to want. This is the one that reverses a settled decision rather than filling a gap, so it is the one most worth pushing back on.

**2. The Pick pattern's people pick.** The ticket put migrating the Pick demos onto the selectable `<ListView>` in a follow-up depending on this one. The people pick moves here, because a dialog that stages a selection and commits it with its own button is the exact case the field-versus-view split was drawn for, and leaving it on `<SelectList>` ships docs that argue for a rule and a demo that breaks it. It touches `pick-list.demo.tsx` and the "List or table" section, which DST-1651 is also working in.

**3. `<ActionBar>`.** The ticket put bulk-action bar wiring in a separate follow-up. No wiring was added. `<ActionBar>` already takes `selectedItemCount` and `onClearSelection` and only falls back to the context a `<Table>` publishes, so what is here is a story, a demo and a docs section. This one reads as demonstrating a composition that already worked rather than building it.

**4. `<ActionBar>`'s height measurement.** Not in the ticket, and not this PR's bug, but this PR is what exposes it. `useActionBar` returned `actionBarHeight: 0` for the whole lifetime of any bar that was closed when it mounted, which is every bar starting from an empty selection. The reservation the bulk-actions demo is built around never activated, and `<Table>` never reserved room either. Raised by @sebald in review. Fixed in 6c591c817 with its own `@marigold/components: patch` changeset and the first test the height path has ever had. The alternative was shipping a demo and a docs paragraph that both describe something that does not happen.

## Screenshots / Preview

Four visual changes for review, three of them in components this ticket was not about. Please look rather than bulk-accepting.

| Change | Where |
| ------ | ----- |
| Indicator column added to the row | every selectable `<ListView>` |
| Indicator drops ~8px on description-bearing rows | `<SelectList>`, all variants |
| Row focus ring moves from outward outline to inset ring | every focused `<ListView>` row |
| Room is now reserved below the rows while the bar is open | every `<Table>` with an `actionBar`, and the new bulk-actions demo |

Two things I could not judge from source and would like a second opinion on:

1. The people pick at `/patterns/user-input/pick` now uses `<ListView>`, which draws no frame where `<SelectList>` drew a boxed surface. Does it read as intentional inside the dialog, or unmoored?
2. The bulk-actions demo uses a `max-h-64` scroll container so the sticky bar has something to pin to. Does the reserved padding at the bottom look deliberate or like a gap?

## Test Instructions

1. `pnpm sb` → **Components/ListView**. Use the new `selectionMode` dropdown (`none` / `single` / `multiple`) on any story.
2. `SelectionWithRowActivation`: press a row with nothing selected (it opens), tick a checkbox, press another row (it marks instead). Press <kbd>Escape</kbd>, then press a row again.
3. `WithActionBar`: select rows, check the count, clear it.
4. `SelectionWithMixedRowContent`: confirm the indicators of both rows line up despite the badge.
5. `pnpm start` → `/components/collection/listview`, `/components/form/selectlist`, `/patterns/user-input/pick`.
6. `pnpm test:sb` and `pnpm test:unit`.

## Breaking Changes

No. `selectionMode` defaults to `"none"`, so existing lists are unchanged. The two visual changes are patch-level and carry their own changesets.

## Three findings worth a reviewer's attention

**1. React Aria: <kbd>Enter</kbd> is inert while a selection exists.** With `onAction` set alongside `selectionMode`, the gestures split by whether anything is selected. Measured, not inferred:

| Selection | Click a row | <kbd>Enter</kbd> | <kbd>Space</kbd> | <kbd>Escape</kbd> |
| Empty | opens | opens | marks | — |
| Not empty | marks | **nothing at all** | marks | clears |

A click at least marks the row. <kbd>Enter</kbd> fires no action and does not toggle, so a keyboard user gets no response and no hint why. React Aria's selection guide documents only *"clicking or tapping a row toggles the selection"* and says nothing about <kbd>Enter</kbd> going dead, so keyboard and pointer diverging here looks like an oversight rather than a decision. It is arguably a 2.1.1 concern: the open action is reachable by mouse in one state and by keyboard only in the other.

Pinned by a story test (`Enter opens only while the selection is empty`) and named in the ListView docs, which tell you to keep a clear button in reach rather than relying on <kbd>Escape</kbd>. **No ticket filed yet** — worth deciding whether we report it upstream, work around it, or accept it.

**2. The decision test now lives on two pages with nothing checking they agree.** It drifted once within a day of being written, which is why the SelectList page now keeps a one-sentence answer and cross-links to the full comparison rather than repeating it. There is no automated guard.

**3. The `<SelectList>` indicator converged on one axis, not two.** `row-span-2` lines the indicator up with the `action` slot beside it, which already centred against the whole text stack, so vertically the two wrappers now follow one rule. Horizontally they still differ:

| | Horizontal placement in the indicator column |
| --- | --- |
| `<ListView>` | `justify-self-start`, pinned to the column start |
| `<SelectList>` | stretches, `justify-center`, centred in the column |

This is only observable once the indicator column widens, which happens when a child claims no cell and auto-places into it. `<ListView>` documents that case (a stray `<Badge>`) and pins the indicator so the widening cannot drag it. `<SelectList>` has no equivalent documented case, so the divergence is latent rather than live. Left alone deliberately, to keep this PR to one visual change in a merged component rather than two.

The selection changeset says centring is the rule "both wrappers now follow, so the two cannot be told apart on this point". That holds for the vertical axis and overstates the horizontal one. Flagged here rather than edited, so the changelog keeps one sentence instead of a caveat.

**Blast radius of the indicator move**, since the changeset gives only the headline number:

- Options with no `<Description>` do not move. The grid is `grid-rows-[minmax(1.25rem,auto)_auto]`, so the second row collapses to 0 and centring over one row or two lands on the same pixel.
- The bordered "Options as cards" demo does not move either. It authors its content as a single unslotted `<Stack>` that auto-places into row 1 and never uses `<Description>`, so row 2 stays empty. The tallest option shape in the system is untouched.
- What moves is label-plus-description options, by roughly 8px.

One latent risk: a description that wraps to several lines drops the indicator further from the label it belongs to. Nothing in the repo does that today and the docs ask for a short description. If it comes up, the fix is `self-start` past a threshold rather than reverting to `row-start-1`, which would reintroduce the mismatch with `action`.

## Checklist

- [ ] Visual regression tests updated (for UI changes) — needs `/run-chromatic`, see the three diffs above


packages/components/src/ListView/ListView.stories.tsx
```diff
@@ -1,10 +1,13 @@
+import { useState } from 'react';
```
OsamaAbdellateef, line 1: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files (`packages/components/src/**`, `themes/**`, and several `*.stories.tsx` files) but the `Visual-Regression-Tests` workflow has not run for this branch (`feat/DST-1665-listview-selection`).

The PR description itself calls out three visual changes to review (selection indicator column, `<SelectList>` indicator drop, focus-ring style). Please trigger the check by commenting `/run-chromatic` on this PR, or by manually dispatching the workflow.

docs/content/components/collection/listview/index.mdx
```diff
@@ -67,7 +79,47 @@ When no single action stands out, or the available actions differ from row to ro
...
+
+`onSelectionChange` hands you React Aria's `Selection`, which is either a `Set` of keys or the string `'all'`. <kbd>Cmd</kbd>/<kbd>Ctrl</kbd>+<kbd>A</kbd> selects every row of a multi-select list and reports the string, so both shapes reach you even in a list that shows no select-all control. Resolve it to concrete keys in one place before you count or act on it, the way the Bulk Actions pattern [does for a `<Table>`](/patterns/user-input/bulk-actions#resolve-the-selection-to-records-once). Reading `.size` straight off the selection returns `undefined` once it is `'all'`, which surfaces as a wrong number in a confirmation rather than an error.
+
+Rows are marked one at a time. A press adds or removes a single row, and there is no <kbd>Shift</kbd>+click to take a range, in a `<ListView>` or in a [`<Table>`](/components/collection/table). Plan a flow that depends on range selection around that.
+
```
aromko: ⚠️ **Shift+click does take a range here, unless I am misreading react-aria.**

In the installed `react-aria@3.51.0`, `useSelectableItem`'s `onSelect` checks `e.shiftKey` and calls `manager.extendSelection(key)` *before* it looks at `selectionBehavior === 'toggle'`, and `useSelectableCollection` extends on <kbd>Shift</kbd>+<kbd>↑</kbd>/<kbd>↓</kbd> whenever `selectionMode === 'multiple'`. Toggle only decides what a plain press does. So in `selectionMode="multiple"` a plain press toggles one row and a Shift-press extends from the last marked row, in `<Table>` too. What toggle rules out is highlight-style selection, where a plain press replaces the whole selection.

Could you check it in the `MultipleSelection` story (mark one row, Shift+click two rows down)? If it extends, this paragraph, the sentence in `.changeset/dst-1665_listview-selection.md`, and the "Not in here" note in the PR body want rewording, ideally with a story test pinning it next to the Mod+A one.

.changeset/dst-1665_listview-selection.md
```diff
@@ -0,0 +1,46 @@
...
+
+The selection is **view state, not a field value**. Read it through `onSelectionChange`, hold it yourself, and decide when it commits. `<ListView>` has no `FieldBase` wiring, no hidden input, and no `name`, `form` or `validate`, and it will not grow them: a selection that has to submit and validate with a form is what `<SelectList>` is for. That split is what keeps two `GridList` wrappers from being redundant, now that both render a nearly identical stack of rows.
+
+`onSelectionChange` receives React Aria's raw `Selection` (`'all' | Set<Key>`), deliberately not `<SelectList>`'s mode-typed `onChange`, because `'all'` is meaningful for a view and unsubmittable for a field. `disallowEmptySelection` passes React Aria's default through in both modes, unlike `<SelectList>`, which defaults it to `true` in single mode for radio-group semantics. A view's selection has to be abandonable, and `true` would also disable Escape-to-clear.
+
+`selectionBehavior` is fixed to `"toggle"` and stays unexposed, matching `<Table>`. There is no equivalent of React Spectrum's `selectionStyle="highlight"`, so no press replaces the whole selection and <kbd>Shift</kbd>+click does not take a range.
```
aromko: ⚠️ Same as the Selection section in `listview/index.mdx`: react-aria's `extendSelection` branch runs on <kbd>Shift</kbd> regardless of `selectionBehavior`, so "Shift+click does not take a range" is likely wrong in multiple mode. The first half of the sentence, no press replaces the whole selection, holds.

.changeset/dst-1665_listview-focus-ring.md
```diff
@@ -0,0 +1,15 @@
+---
+'@marigold/theme-rui': patch
+---
+
+fix(DST-1665): draw `<ListView>`'s row focus ring inside the row
+
+A focused row used `ui-state-focus`, which paints a 3px outline outside the border box. Rows are siblings painting in tree order, so the next row down covered the focused row's bottom edge with its own background. The divider did it faintly at rest; a selected neighbour did it across the full width. The ring looked cut off along the bottom.
```
aromko: 💡 Heads-up, not blocking today: #5779 (DST-1526) adds a Vale rule at `error` level that rejects semicolons in changeset and docs prose. Three sentences in this PR would fail it once that lands: this line, line 13 below ("...for borderless rows instead; that utility..."), and `.changeset/dst-1665_listview-selection.md` line 24 ("...onto its own line; now it misaligns the list."). The four docs pages are clean.

```suggestion
A focused row used `ui-state-focus`, which paints a 3px outline outside the border box. Rows are siblings painting in tree order, so the next row down covered the focused row's bottom edge with its own background. The divider did it faintly at rest. A selected neighbour did it across the full width. The ring looked cut off along the bottom.
```

packages/components/src/ListView/ListView.stories.tsx
sarahgm: The workflow did run, twice. Chromatic build 255 on `7d8e229` and build 256 on `3dbe62d`, which is this PR's head. Build 256 passed and reports on this PR as `UI Tests: marigold_2026 — pass`, with 10 changes accepted as baselines and 0 unreviewed. The run log confirms the commit it tested (`CHROMATIC_SHA: 3dbe62d62ccebeb8e8e16f44ffcbf6a38c6f0e95`).

The false positive has a specific cause worth knowing for the reviewer. `visual-regression-tests.yml` fires on `issue_comment`, and GitHub attributes an `issue_comment` run to the default branch, not to the PR head. So `gh run list --branch feat/DST-1665-listview-selection` returns nothing even though the run exists and checked out the right commit. Looking for the Chromatic check on the PR instead of a branch-scoped workflow run would avoid this.

Nothing to change here, so I'm leaving the VRT state as is.

docs/content/components/collection/listview/index.mdx
sarahgm: You read it right, and the paragraph was wrong. Fixed in 83919e8.

Confirmed two ways. In `react-aria@3.51.0`, `useSelectableItem.mjs:58` runs `else if (e && e.shiftKey) manager.extendSelection(key)` before the `selectionBehavior === 'toggle'` branch on line 59, and `SelectionManager.extendSelection` in `react-stately@3.49.0` adds the whole `getKeyRange(anchorKey, toKey)`. `useSelectableCollection.mjs:76` does the same on Shift+Arrow whenever `selectionMode === 'multiple'`. Toggle only decides what a plain press does, exactly as you said.

I also ran the check you asked for. A story test on `WithActionBar` clicks Gasometer, Shift+clicks Columbiahalle two rows down, and all three rows come back `aria-selected="true"`. It passes in Firefox. I put it on `WithActionBar` rather than `MultipleSelection` because that story has only two rows, so a range there would prove nothing, and it already carries `chromatic: { disableSnapshot: true }`, so the test costs no snapshot. It sits next to the Mod+A one.

The paragraph now reads:

> A press adds or removes a single row and never replaces the whole selection. <kbd>Shift</kbd>+click extends from the last marked row to the pressed one, and <kbd>Shift</kbd>+<kbd>↑</kbd>/<kbd>↓</kbd> does the same from the keyboard. Both behave this way in a [`<Table>`](/components/collection/table) too.

The changeset sentence and the "Not in here" note in the PR body are corrected too.

.changeset/dst-1665_listview-selection.md
sarahgm: Correct, and fixed in 83919e8. Your reading of the split holds exactly: the first half stands, the range clause does not.

The sentence now reads:

> `selectionBehavior` is fixed to `"toggle"` and stays unexposed, matching `<Table>`. There is no equivalent of React Spectrum's `selectionStyle="highlight"`, so no press replaces the whole selection. Range selection is unaffected: React Aria checks <kbd>Shift</kbd> before it consults `selectionBehavior`, so <kbd>Shift</kbd>+click and <kbd>Shift</kbd>+<kbd>↑</kbd>/<kbd>↓</kbd> extend a range in `"multiple"` mode.

Verification and the new story test are in the thread on `listview/index.mdx`.

.changeset/dst-1665_listview-focus-ring.md
sarahgm: Blocking now rather than a heads-up: #5779 merged this morning at 09:04, so the rule is on `main`. Taken your suggestion verbatim for this line, and fixed the other two the same way in 83919e8.

- `dst-1665_listview-focus-ring.md:13` → "...for borderless rows instead. That utility is the same inset mechanism at full opacity."
- `dst-1665_listview-selection.md:24` → "Before selection this misplaced a badge onto its own line. Now it misaligns the list."

One addition to your count. `docs/content/components/collection/actionbar/index.mdx` also had two, at lines 55 and 104. They were not this PR's doing, and #5779 had already fixed those same two lines on `main`, so merging `main` into the branch cleared them with no conflict. That is why the four docs pages read clean from where you were looking.

I merged `main` first for that reason, then ran Vale 3.19.0 over every changed prose file: 0 errors in 8 files.

docs/content/components/collection/listview/listview-bulk-actions.demo.tsx
```diff
@@ -0,0 +1,78 @@
...
+            </Button>
+          </ActionBar>
+        );
+      },
+    });
+
+  // The bar is sticky, so it pins to the bottom of this scroll container.
+  // Reserving its height in both paddings keeps it off the last rows and stops
+  // keyboard scrolling parking a row underneath it.
+  const room = actionBarHeight
```
sebald, line 39: 🔴 **This demo's height reservation never activates, and the bar covers two rows.**

Measured live on `localhost:3000/components/collection/listview` with one row selected:

| | |
|---|---|
| Bar's real height | **60px** |
| `actionBarHeight` from the hook | **0** |
| `room` | `undefined` |
| Container inline `padding-bottom` | unset |
| Container inline `scroll-padding-bottom` | unset |

With the `max-h-64` container scrolled to the top, the sticky bar overlaps the `floorplan` and `rider` rows. That is exactly the failure the prose above promises this prevents: *"without it the bar covers the last rows and keyboard navigation parks a focused row underneath it."*

The root cause is **not** in this PR. `<Table>` uses the identical expression at `Table.tsx:122-126`, and no `<ActionBar>` demo anywhere on the docs site reserves room either, so `useActionBar`'s height measurement is not reaching consumers at all. The count does arrive (the bar renders "1 selected" from context), so `ActionBarContext` is wired and only `onHeightChange` is not landing.

This PR is what ships a demo built entirely around that measurement, plus prose asserting it works. Worth either fixing the measurement in `useActionBar`/`ActionBar` or softening the claim until it is fixed.

packages/components/src/ListView/ListViewItem.tsx
```diff
@@ -202,24 +173,33 @@ export const ListViewItem = ({
...
-            whichever cell is free. Wrapping it in `<TextValue>` (which
-            defaults to `slot="label"`) gives the shorthand the same cell and
-            styling as the explicit form. */}
-        {typeof children === 'string' ? (
-          <TextValue>{children}</TextValue>
-        ) : (
-          children
-        )}
-      </ItemChildren>
+      {({ selectionMode, isSelected, isDisabled }) => (
```
sebald, line 176: ⚠️ **Moving to a render prop removes React's element-identity bailout for the whole row subtree.**

`<ItemChildren>` used to be a static child, so RAC passed the same element reference straight through and React skipped the subtree on re-render. Inside the render prop, RAC re-invokes this function on every re-render of the row — and a selection change re-renders every row, since they all read the shared `GridListStateContext`.

Measured by patching `Element.prototype.querySelectorAll` in Storybook and counting the `[data-grid-area="actions"]` scans from `useStackedActionsWarning`:

| Interaction | Scans |
|---|---|
| One selection click, 3-row list | **10** |
| Hover, selectable list | 1 |
| Hover, non-selectable list | 0 |

So hover is fine and localised; selection is the cascade, and it scales with row count. The DOM scan itself is dev-only (`NODE_ENV` guard), but re-running `ItemChildren`, its three `useMemo` dependency checks, and the `Provider` element allocations for N rows is not.

Cheap fix: only the indicator needs the render prop's values. Build `<ItemChildren>` outside the closure and pass it in, so the slot machinery keeps its bailout:

```tsx
const itemChildren = (
  <ItemChildren row={rowRef} /* …classNames */>
    {typeof children === 'string' ? <TextValue>{children}</TextValue> : children}
  </ItemChildren>
);

return (
  <RACGridListItem …>
    {({ selectionMode, isSelected, isDisabled }) => (
      <>
        <GridSelectionIndicator … />
        {itemChildren}
      </>
    )}
  </RACGridListItem>
);
```

packages/components/src/ListView/ListView.tsx
```diff
@@ -54,13 +56,22 @@ const ListViewBase = ({
...
   ref,
   ...rest
 }: ListViewProps) => {
   const classNames = useClassNames({ component: 'ListView', variant, size });
 
-  const contextValue = useMemo(() => ({ classNames }), [classNames]);
+  // `useClassNames` returns a fresh object, so memoise on the slot strings.
+  const { list, item, label, description, title, indicator, actions } =
+    classNames;
+  const contextValue = useMemo(
```
sebald, line 69: ⚠️ **Right diagnosis, but the fix belongs one layer down.**

The comment is correct and this is a genuine improvement: `useClassNames` returns a fresh object from `Object.fromEntries` on every call (`packages/system/src/hooks/useClassNames.tsx:81`), so the previous `useMemo(() => ({ classNames }), [classNames])` never once hit cache.

The catch is that the same dead memo exists in **six other components** — `SelectList`, `Table`, `Dialog`, `Drawer`, `Tray`, `RangeCalendar` — and this repairs only `ListView`, by hand-listing all seven slot names in two places. Memoising inside `useClassNames` itself would fix all seven for free and remove the need for any of this.

One thing worth recording so nobody "fixes" it later: I checked, and this **is** compile-safe. Dropping a slot from the literal errors with `TS2322` at line 77, because `ListViewContextProps` requires the full `ComponentClassNames<'ListView'>`. So it is verbose, not fragile.

Suggest a follow-up ticket for the `useClassNames` root fix rather than changing anything here.

packages/components/src/ListView/ListView.stories.tsx
```diff
@@ -621,3 +642,460 @@ export const EmptyState = meta.story({
...
+    expect(canvas.queryByRole('checkbox')).not.toBeInTheDocument();
+
+    await userEvent.click(express);
+
+    expect(express).toHaveAttribute('aria-selected', 'true');
+    expect(standard).toHaveAttribute('aria-selected', 'false');
+  }
+);
+
+export const MultipleSelection = meta.story({
```
sebald, line 692: ⚠️ **The Gasometer/Tempodrom rows are hand-copied verbatim across four stories.**

`MultipleSelection` (here), `SelectionWithRowActions`, `SelectionWithMixedRowContent` and `SelectionWithRowActivation` each inline byte-identical ids, names and descriptions (`Vienna · 1600 seats`, `Berlin · 3800 seats`).

This same file already shows the better pattern: `BulkActionsExample` hoists `const VENUES = [...]` and maps over it. Sharing that array across the other four would leave only the per-story decoration (the `ActionMenu`, the `Badge`, the `ButtonGroup`) inline, and drop the number of places a venue rename has to be applied from five to one.

docs/content/components/collection/listview/listview-row-activation.demo.tsx
```diff
@@ -0,0 +1,57 @@
...
+import type { Selection } from '@marigold/components';
+import {
+  Description,
+  ListView,
+  Stack,
+  Text,
+  TextValue,
+} from '@marigold/components';
+
+const uploads = [
```
sebald: ⚠️ **This `uploads` fixture is byte-for-byte identical to the one in `listview-bulk-actions.demo.tsx`** — same ids, names and detail strings for `report`, `season` and `contract`.

The sibling demo in this very PR shows the established convention for exactly this case:

```tsx
import { venues } from '@/lib/data/venues';
```

A `docs/lib/data/uploads.ts` following the `venues.ts` / `people.ts` pattern would give both demos one source of truth (the bulk-actions demo can extend it with its extra three rows).

packages/components/src/ListView/ListView.stories.tsx
```diff
@@ -621,3 +642,460 @@ export const EmptyState = meta.story({
...
+    await step('Escape clears the selection', async () => {
+      await userEvent.keyboard('{Escape}');
+
+      expect(express).toHaveAttribute('aria-selected', 'false');
+      expect(grid.querySelectorAll('[aria-selected="true"]')).toHaveLength(0);
+    });
+  }
+);
+
+const onOpenVenue = fn();
```
sebald, line 787: 💡 **These module-scope `fn()` mocks are never asserted and never reset.**

Five of them across the new stories: `onOpenVenue`, `onArchiveVenue` (here), `onVenueMenu`, `onVenueArchive`, and `onBulkArchive`. Grepping the file, none appears in an `expect(...)`, and there is no `clearMocks`/`restoreMocks` in `vite.config.ts` or `.storybook`, so they just accumulate call history for the whole test run.

`fn()` from `storybook/test` signals "this is a spy you are about to assert on", which reads as test coverage that does not exist. A plain `() => {}` says the same thing more honestly. (The `fn()`s passed through story `args` and actually asserted — `onSelectionChange`, `onAction` — are correct as they are.)

themes/theme-rui/src/components/ListView.styles.ts
```diff
@@ -24,22 +16,17 @@ export const ListView: ThemeComponent<'ListView'> = {
...
       'max-sm:min-h-touch-target',
       'text-sm text-foreground outline-none',
-      // `background` is deliberately absent: DST-1436 dropped it from hover
-      // transitions theme-wide so hover fills flip instantly.
       'transition-[border,color]',
+      'selected:bg-selected',
       'hover:ui-state-hover',
-      'focus-visible:ui-state-focus',
+      // Inline, not `ui-state-focus-item`: `/50` matches SelectList (DST-1590/1662).
+      'focus-visible:inset-ring-2 focus-visible:inset-ring-ring/50',
```
sebald, line 28: 💡 **Second verbatim copy of this pair, and the comment is the right call for now.**

Identical to `SelectList.styles.ts:68`, and both bypass `ui-state-focus-item` (`themes/theme-rui/src/ui.css:215`), the utility whose stated purpose is *"borderless rows: menu items, list options, table rows"* — precisely this case. The only mismatch is opacity: the utility is full-strength `inset-ring-ring`, both call sites want `/50`.

The inline comment naming DST-1590/DST-1662 makes this defensible, and I verified the ring works: computed `box-shadow` is a 2px inset at 0.5 alpha with `outline: none`, so no sibling can clip it any more.

Raising it only so the pattern is on record: those tickets track the *opacity policy*, not the duplication. A third copy should not land without either parameterising `ui-state-focus-item` or filing something that names both call sites.

docs/content/components/collection/listview/listview-bulk-actions.demo.tsx
```diff
@@ -0,0 +1,78 @@
...
+];
+
+export default () => {
+  // The hook holds the selection and measures the bar, so the count and the
+  // clear button are filled in for you.
+  const { selectedKeys, onSelectionChange, actionBarHeight, actionBarOverlay } =
+    useActionBar({
+      // [!code highlight]
+      actionBar: (keys: Selection) => {
+        const count = keys === 'all' ? uploads.length : keys.size;
```
sebald, line 20: 💡 **`useActionBar` already computes this exact expression, it just does not hand it to you.**

`useActionBar.tsx:71-76` resolves `selectedItemCount: number | 'all'` internally and forwards it into `ActionBarContext` for the bar's own count badge. But the `actionBar` callback receives the raw `Selection`, so anything else that needs the number — like the `alert` text below — has to redo the resolution by hand.

The same one-liner now appears three times in this PR: here, `listview-row-activation.demo.tsx:28`, and `ListView.stories.tsx:915`.

Not a fix for this demo (there is no helper to call today), but it points at a small API gap worth a follow-up: either pass the resolved count as a second argument to the `actionBar` callback, or return `selectedItemCount` from the hook alongside `actionBarHeight`.

sebald (commented):
Automated review pass (code quality + React idiom + live click-testing in Firefox). Eight findings below, none overlapping the existing threads. Skipped a visual-regression finding: `gh run list --branch` reports no runs, but that is the `issue_comment` attribution false positive you already documented on this PR, so it is not a real gap.

One high-severity item (the bulk-actions demo), three medium, four low. Everything else I checked held up, including every keyboard claim in the new docs, the indicator alignment under mixed row content, the inset focus ring, and the `selectionMode="none"` byte-identity claim.

---

packages/components/src/ListView/ListView.stories.tsx
sarahgm: Leaving these as they are, with one correction to the premise.

The five you name are indeed unasserted, and I confirmed all five have zero `expect(...)` hits. But the module-scope `fn()` pattern is not new here and is not uniformly decorative: `main` already carries 15 of them in this file, and four of those **are** asserted.

| Mock | Assertion |
| --- | --- |
| `onArchive` | `ListView.stories.tsx:233` |
| `onMute` | `:249` |
| `onLeaveWorkspace` | `:419` |
| `onRemoveMember` | `:501` |

So in this file `fn()` at module scope already means "a spy, asserted where the story's test needs it". The other 11 pre-existing ones are unasserted for exactly the reason the new five are: the handler is there to make the story interactive, while the test beside it asserts something else, such as that a row control stays reachable or that the indicators line up.

That leaves the change either inconsistent or out of scope. Converting only the five new ones leaves 15 `fn()`s sitting next to 5 plain arrows in one file. Converting all 20 is a sweep on a PR that is already carrying three out-of-scope items.

Your reset point holds, and I checked it rather than taking it on trust: there is no `clearMocks`, `restoreMocks` or `mockReset` in `vite.config.ts` or `.storybook`, so call history does accumulate for the whole run. That is a file-wide property rather than something these five introduce.

Happy to file a follow-up that settles the convention for the whole file in one go, if you think it is worth one.

docs/content/components/collection/listview/listview-bulk-actions.demo.tsx
sarahgm: Fixed in 6c591c817, and you were right that the root cause is not this PR's.

It sits one layer up from where it looks. The measurement ran in the outer `<ActionBar>`, which returns `null` while nothing is selected, and React Aria's `useResizeObserver` wraps `onResize` in `useEffectEvent` and declares its deps as `[ref, box]`:

```js
// react-aria/private/utils/useResizeObserver.mjs
useEffect(() => {
    let element = ref?.current;
    if (!element) return;
    ...
}, [ref, box]);
```

So the effect ran once on mount against an empty ref, bailed out, and never retried, because neither dep ever changes. The observer never attached to an element at all. The fix moves the observation into `ActionBarInner`, which mounts with the bar itself, so the ref is live when its effect runs and a close-then-reopen cycle re-establishes it.

That also explains why nobody caught it: `IntegratedWithTable` starts with two rows already selected, so the bar is open on the first commit and the ref happens to be set. The bug only bites a bar that is closed at mount, which is every bar with an empty starting selection, including this demo.

The height path had no test coverage anywhere, which is how it shipped. There is one now, on `IntegratedWithTable`, deselecting everything and reselecting so the bar goes through a close and reopen. I checked it in both directions rather than only that it passes:

| | Result |
| --- | --- |
| Before the fix | `AssertionError: expected '' to match /^calc\((?!0px)\d/` |
| After the fix | 24 passed |

That empty string is exactly your `room: undefined` row. Carries its own `@marigold/components: patch` changeset, since it silently fixes every `<Table>` with an `actionBar` as well.

packages/components/src/ListView/ListViewItem.tsx
sarahgm: I applied your patch exactly as written and measured it, and it does not move the number. Not applying it, but the cascade you found is real and worth recording properly.

Counting `ItemChildren` renders (what triggers the dev-only `querySelectorAll`) across one selection click on the uncontrolled `MultipleSelection` story:

| Variant | `ItemChildren` renders |
| --- | --- |
| As merged (render prop) | 6 |
| With your hoist applied | 6 |
| Hoist, plus `ItemChildren` stops reading the contexts | 0 |

Instrumenting the three call sites separately gives `{ item: 0, rp: 6, children: 6 }`. `ListViewItem` does not re-render at all on a selection change, so with the hoist applied the element reference already is stable, and the subtree is still re-rendered six times.

The cause is RAC, not the render prop. `GridListItem` wraps the row's children in a `Provider` that supplies `TextContext` with a fresh object literal on every render:

```js
// react-aria-components/dist/private/GridList.mjs:408
[TextContext, { slots: { [DEFAULT_SLOT]: {}, description: descriptionProps } }]
```

`ItemChildren` consumes that through `use(TextContext)`, deliberately, so it can merge RAC's slot config rather than replace it, which is what keeps `descriptionProps`' aria wiring intact. React re-renders a context consumer whenever its provider's value changes identity, regardless of element identity, so no stable element can skip that subtree. The third row of the table is the control: drop the two context reads and the same hoist goes to zero.

So the row re-renders because it is selectable, and `ItemChildren` follows it through a context dependency it needs to do its job. Your own hover numbers point the same way, 1 on a selectable list against 0 on a non-selectable one. The static-child version on `main` would re-render identically if it had selection to react to.

The one cost here that we do control is `useStackedActionsWarning`, whose effect has no dependency array and so runs a `querySelectorAll` on every render. That is dev-only and predates this PR. Happy to file it if you want it gone.

packages/components/src/ListView/ListView.tsx
sarahgm: Agreed on all three points, and I am leaving `ListView` as it is.

Confirmed the other six: `SelectList.tsx:328`, `Table.tsx:88`, `Dialog.tsx:79`, `Drawer.tsx:120`, `Tray.tsx:138` and `RangeCalendar`. I also checked the root fix is feasible rather than only desirable. A single `useMemo` inside `useClassNames`, keyed on theme, component, variant, size and className, works, with the caveat that the two early returns have to be folded into it since hooks cannot be conditional. A caller passing an object `className` would miss the memo, but that is no worse than today.

Drafting a follow-up ticket for it and will link it here.

packages/components/src/ListView/ListView.stories.tsx
sarahgm: Done in 4f20364e5. `VENUES` now sits above `MultipleSelection` with `const [gasometer, tempodrom] = VENUES`, and all four stories source their data from it. Literal venue names in the file drop from roughly 20 to 5: three in the array itself, and two in test assertions that I left literal on purpose, so the tests pin rendered output instead of mirroring the source.

One deliberate difference from what you sketched. I kept the explicit `<ListView.Item>` children rather than mapping over `items={VENUES}`, because `SelectionWithRowActions` and `SelectionWithMixedRowContent` decorate their rows differently per row. Tempodrom's menu has one item against Gasometer's two, and only Gasometer carries the `Badge`, so a uniform map would change what those two stories render. The JSX is also what Storybook's code panel shows a reader.

docs/content/components/collection/listview/listview-row-activation.demo.tsx
sarahgm: Done in 4f20364e5. `docs/lib/data/uploads.ts` now holds all six rows following the `venues.ts` and `people.ts` shape, `as const` with a derived `Upload` type. The bulk-actions demo imports the whole set and this one takes `uploads.slice(0, 3)`, mirroring how `listview-selection.demo.tsx` already does `venues.slice(2, 5)`.

No extra changeset needed: the selection changeset already carries `@marigold/docs: patch`.

themes/theme-rui/src/components/ListView.styles.ts
sarahgm: Taking the record-keeping ask.

Drafting a ticket that names both call sites, `ListView.styles.ts:28` and `SelectList.styles.ts:68`, and proposes parameterising `ui-state-focus-item` for the opacity, so a third copy has somewhere to go instead of another inline pair. I will link it here and extend the inline comment in both files to point at it, so the next reader finds the duplication ticket rather than only DST-1590 and DST-1662, which track the opacity policy.

docs/content/components/collection/listview/listview-bulk-actions.demo.tsx
sarahgm: Agreed, and I am leaving the demo alone.

The resolution appears three times in this PR: here, `listview-row-activation.demo.tsx:28` and `ListView.stories.tsx:915`. The ListView docs already warn about getting it wrong, since `.size` returns `undefined` once the selection is `'all'`, which is a fair sign it should not be hand-rolled per call site.

Drafting a follow-up for returning `selectedItemCount` from `useActionBar` alongside `actionBarHeight`. Deliberately not doing it here even though I was already in `useActionBar` for the height fix, because it is new public API on a PR that is already carrying three out-of-scope items. Will link the ticket here.

packages/components/src/ListView/ListView.tsx
sarahgm: Filed as **DST-1763**, `🧹 Memoise useClassNames so component context values can be stable`, linked to this ticket and to DST-1451.

I checked Jira for an existing owner first and there is none. DST-1451 (`🧹 Rework @marigold/system theme types`, Bereit) edits the same file and even counts the three `@ts-expect-error`s in it, but it is about the type architecture rather than the runtime memo, so the two do not collide. Linked them so whoever picks up either knows about the other, and noted the option of folding the memo into DST-1451 instead.

The ticket carries the six other call sites, the note that the early returns have to be folded into the memo since hooks cannot be conditional, and your finding that ListView's version is compile-safe rather than fragile, so nobody "fixes" it later.

themes/theme-rui/src/components/ListView.styles.ts
sarahgm: This one already has an owner, so no new ticket.

**DST-1662**, `🧹 Migrate borderless rows off the 2.08:1 focus halo onto ui-state-focus-item`, status Bereit. Its scope list already names the sibling verbatim:

> `SelectList.styles.ts:68` — `focus-visible:inset-ring-2 focus-visible:inset-ring-ring/50`, the right shape at the wrong alpha.

What it could not name is `ListView.styles.ts:28`, because that line is new here. Added it in a comment on DST-1662 rather than editing the description, since the ticket is Ready and may already have been sized. So it migrates two rows instead of one.

Worth knowing: DST-1662 carries an accessibility case neither of us mentioned. The `ring/50` halo measures **2.08:1** on white, under the 3:1 that WCAG 2.1 SC 1.4.11 requires, with `ListBox`, `Table` and `Sidebar` in the same migration. So the alpha is not just a styling preference, and "parameterise the utility" is likely the wrong end of it: the ticket's plan is to move these call sites to the full-strength utility and drop the `/50`.

Your read that DST-1590 is the wrong ticket for this holds. It is `💄 Alpha-based state & hierarchy fills`, about the opacity ramp, not the duplication.

Also did the second half of what you asked, in 842c7b939: `SelectList.styles.ts:68` now carries the same inline comment `ListView.styles.ts:28` has, so the duplication is discoverable from both sides rather than one.

docs/content/components/collection/listview/listview-bulk-actions.demo.tsx
sarahgm: Filed as **DST-1764**, `✨ Return the resolved selectedItemCount from useActionBar`, linked to this ticket and to DST-1751.

Checked for an existing owner first and there is none. All the other ActionBar tickets are Done, and DST-1751 (select-all for non-table collections) is the other half of this area rather than this.

The ticket carries your three call sites, and the reason this is more than tidiness: `.size` returns `undefined` once the selection is `'all'`, and Cmd/Ctrl+A produces `'all'` on any multi-select collection, so every consumer meets that case whether or not the UI offers a select-all control. It leaves the shape open between returning it from the hook and passing it as a second callback argument, with a preference for the former, since a consumer that needs the count outside the `actionBar` callback cannot reach a callback argument.

Still not doing it here, for the reason above: new public API on a PR already carrying out-of-scope items.

sebald (approved):

OsamaAbdellateef (approved):

PR #5778, opened 2026-09-02
feat(DST-1522): add the /create-ticket skill and correct CLAUDE.md's Jira section

sarahgm:
# Description

Adds `.claude/skills/create-ticket/SKILL.md`, a nine-step skill that turns a settled scope (normally the summary `/grill` ends with) into one DST-conformant Jira issue: emoji title, issue type, the fields that type actually requires, and the description template Jira carries for that type. It confirms the whole rendered ticket before anything is written.

Closes DST-1522

**This PR deliberately overrules the ticket's Suggested Solution.** DST-1522 asks to *"hardcode the metadata from `CLAUDE.md` so the skill never has to guess"*. It does the opposite: step 4 reads the field metadata from Jira on every run and the skill carries no field table at all. Two reasons, both found while building it:

1. **The copy was already wrong.** `CLAUDE.md` listed the three custom fields under "Required Custom Fields (Task)" as though that settled where they apply. `Appetite` and `Rollout Communication` are on `Task` alone, `Requires UI Kit Update` is on `Task` **and** `Unplanned`, and `Bug` carries none. Sending a field the chosen type does not have fails the whole create.
2. **The copy could never have been right.** Each issue type carries its own description template as that type's `description` field default. A hardcoded skill would have filed every `Bug` on `Task`'s template, losing the reproduce steps a bug report exists for. Nobody had written this down and nobody knew to.

The decision was ratified in-session before this landed.

**It also departs from the ticket's Expected Outcome on sprint placement.** DST-1522 says the skill "then links it to the sprint". It does not. A new ticket lands in the backlog, and reaching the active sprint takes `--sprint` or an explicit choice at the confirm gate. Adding to a running sprint changes that sprint's committed scope, which is a person's decision rather than a skill's default. Flagging it here so the call is the reporter's to overturn.

Also in the diff:

- **`CLAUDE.md`, three corrections to the Jira section.** `Unplanned` is not the ad-hoc exception the issue-type table described: over the last 90 days DST filed 123 Tasks against 121 Unplanned, so it is half the board. The two differ by planning provenance, not by kind of work, which is why a feature and a defect can both be `Unplanned`. The custom-field section now says which types each field is actually on. The description template is marked as `Task`'s, with each type's headings quoted literally rather than slash-separated. That last part fixed a live error: the old wording rendered `Unplanned`'s `Context / Trigger` heading as "Context or Trigger", because a slash-separated list whose middle item contains a slash cannot be read back.
- **`.claude/README.md`, the confirmation rule restated.** It required the gate to be step 1 of the workflow, which was written from `vrt` and only ever fit `vrt`. A skill whose gate exists to *show* what it is about to do cannot meet it, so `create-pr` (step 6 of 8) and `create-ticket` (step 7 of 9) both broke it. The rule is now a hard stop immediately before the first outward call, which is what all three actually do. Alongside it, a note that a gate only holds if the question reaches a human. `AskUserQuestion` is resolved by the permission component, so under `permissions.defaultMode: "auto"` with `skipAutoPermissionPrompt` set it never renders and returns a default indistinguishable from a real answer. Found the slow way in this session.
- **`.claude/skills/grill/SKILL.md`**, one line: its handoff no longer says "once that skill exists".

N/A on most of the checklist below: no UI, no components, no published package. `.claude/` is not published, so no changeset is required. Every skill commit in this repo's history has landed without one.

## Test Instructions

1. Run `/create-ticket` with a thin brief, e.g. `align the checkbox and radio styles`. It should stop at step 2, name which completeness checks failed, and route to `/grill` without drafting.
2. Run `/create-ticket` with a scoped brief (a `/grill` summary, or a path to a file holding one). It should reach the confirm gate having read the issue type's fields from Jira and searched for duplicates.
3. At the gate, switch the issue type to `Unplanned`. It should re-read the metadata and drop `Appetite` from both payload and gate, since that field is not on `Unplanned`'s create screen.
4. Cancel. Nothing should reach Jira.

**Steps 8 and 9 are untested.** No issue has been created through the skill yet, so `createJiraIssue`, `createIssueLink` and the read-back have never run. Everything up to and including the gate has been exercised against live Jira.

## Breaking Changes

No

## Checklist

- [ ] Stories added/updated (with `component-test` tag where applicable) — N/A, no components
- [ ] Unit tests added/updated — N/A, markdown only
- [ ] Component documentation added/updated (if it exists) — N/A
- [ ] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) (for new/changed interactive components) — N/A
- [ ] Visual regression tests updated (for UI changes) — N/A, no file touches `packages/components/src`, `packages/system/src`, `themes/`, `.storybook/`, `*.stories.tsx` or `*.styles.ts`
- [ ] Changeset added (`pnpm changeset`) — N/A, `.claude/` is not a published package


.claude/skills/create-ticket/SKILL.md
```diff
@@ -0,0 +1,196 @@
+---
+name: create-ticket
+description: Marigold repo — Turn a settled scope (normally the summary `/grill` ends with) into one DST-conformant Jira issue, with the emoji title convention, the right issue type, the fields that type actually requires, and the description template Jira carries for that type. Use when the user asks to "create a ticket", "file this in Jira", "make a DST ticket", or types `/create-ticket`. Creating an issue is an outward action, so run this only on an explicit request, never proactively and never as a follow-up to unrelated work. The full rendered ticket is shown for confirmation before anything is written to Jira.
```
aromko: Again here I wouldn't pin it to marigold repo. Should also work in insights and so on.

sarahgm: Renamed. The marker is `DST — ` now, on all six skills, and the rule in `.claude/README.md` says what it names: the team whose workflow the skill encodes, not the checkout it happens to sit in.

The body moved with it, in 44b98ac77. The skill no longer reads `CLAUDE.md` for anything. The coordinates, the issue type ids and the title emoji table travel with it in `references/dst-conventions.md`, and which fields a type requires is still read from Jira at run time. `CLAUDE.md` points at that file now rather than holding a second copy of the two tables. Nothing left in the skill depends on a file from the repo it runs in, which was a shorter list than `/triage-feedback` had, since this one only ever talks to Jira.

Worth saying explicitly, because it is the part that makes the rename honest: DST is not this repository's tracker. `/pick-up` already records that it carries work on the Core app, ClearingAdministration, the Cypress end-to-end suite and the Insights scanner, so a ticket filed from the insights checkout belongs in the same project under the same conventions.

What this does not fix is that a skill in `.claude/skills/` loads only in the repo it is committed to, so typing `/create-ticket` inside insights still will not work. That needs the shared plugin you described on #5781, ours rather than rx-ai-suite. Happy to file it as its own ticket.

aromko (approved):
I think there are still some stuff which isn't handy when to use in another repo but we can look at it again when we want to use it in multiple repos.

PR #5496, opened 2026-06-03
docs: remove dash punctuation and semicolons from documentation prose

sebald:
# Description

Repo-wide documentation style cleanup: removes dash punctuation (em-dashes and en-dashes used as punctuation) and prose semicolons from the docs, rewriting them as plain sentences for easier reading. Split out from the DST-1483 work so it can be reviewed independently.

Scope: authored MDX under `docs/content/**` (component, foundation, getting-started, and pattern pages). Generated changelog pages under `docs/content/releases/**` are excluded.

## What changed

- **Em-dash punctuation** becomes a period or comma (e.g. "the structure of a page. Larger…", "such as hour, minute…").
- **En-dash numeric ranges** read as words (e.g. "1 to 3 actions per row", "3 to 5 items").
- **Prose semicolons** become two sentences (e.g. "…the same. Only the contents change.").
- Code is untouched: `import` statements, inline code, JSX, and CSS keep their semicolons.

## Intentionally kept

Three en-dashes remain because they appear inside example strings that show **literal component output**, where changing them would misrepresent what the component renders:

- `DateFormat` docs: `"Jan 1, 2025 – Dec 31, 2025"`
- `NumericFormat` docs: `"1,000 – 5,000"`
- Filter pattern: an example chip label `"Last 7 days (Aug 3 – Aug 9)"`

## Test Instructions

1. The change is prose only (no JSX, imports, or structure touched), so the docs build and types are unaffected. CI runs the docs build.
2. Spot-check a few pages (Card, Panel, Table, Forms) for readability.

## Breaking Changes

No.

## Checklist

- [ ] Storybook preview and Marigold docs preview are available
- [ ] Stories added/updated (with `component-test` tag where applicable)
- [ ] Unit tests added/updated
- [x] Component documentation added/updated
- [ ] Accessibility reviewed
- [ ] Visual regression tests updated (for UI changes)
- [x] Changeset added (`pnpm changeset`)

PR #5779, opened 2026-09-02
ci(DST-1526): lint docs prose with Vale in pre-commit and CI

aromko:
# Description

Prose style rules for the docs were carried in people's heads and in review comments, so the AI relearned them by being corrected after the fact. This encodes them as Vale checks wired into the pre-commit hook and a CI check, with the writing guidance in `CLAUDE.md`.

Three rules block CI: no em dashes, no semicolons in prose, no en dash asides. Table cells are exempt, because there an em dash is a legitimate "not applicable" marker, and ranges keep the en dash so quoted component output stays accurate. Scope is prose written for a reader (`docs/content/**`, `.changeset/*.md`, published READMEs, top-level markdown); prose written for an agent is deliberately excluded, as are `docs/legal/`, the dated release posts under `docs/content/releases/blog/` and everything generated.

All 126 existing violations across 23 files are rewritten here, so the check starts green rather than accumulating suppressions as files are touched. DST-1526 lists "rewriting existing docs to pass" under **Not included**. That is overruled on purpose: `vale .` lints the whole repo, so fix-forward would leave the check red on day one no matter which files a PR touches. Those edits are punctuation only: four words change beyond punctuation and capitalisation, and line counts and code fences are unchanged in every file.

N/A — no UI changes, so no screenshots and no VRT. No unit tests: the deliverable is a lint configuration, and its behaviour is verified by the check itself passing on this branch.

Closes DST-1526

## Test Instructions

1. `pnpm install`, then `pnpm lint:prose` — expect exit 0. On a platform without a pinned build the script prints a skip notice and exits 0 instead of failing on a missing binary. First run downloads the pinned Vale 3.19.0 binary (~1s, checksum verified); later runs are ~200ms.
2. Add an em dash to any prose line in `docs/content/**/*.mdx`, run `pnpm lint:prose` again, and confirm it fails naming the file, line and rule.
3. Stage that file and try to commit — the pre-commit hook should block it.
4. Add an em dash to `CLAUDE.md` or anything under `.memory/` instead, and confirm it does **not** block: agent-facing prose is out of scope on purpose.
5. Check a table cell still accepts one, e.g. `| — (no class) |`.

## Breaking Changes

No

## Checklist

---

## Notes for the reviewer

Two things worth a closer look:

**`packages/cli/README.md` is 67 of the 126 fixes** — half the prose churn in one file.

**The en dash rule is the one to scrutinise.** It exists because German uses a spaced en dash where English uses an em dash, so it guards a slip this team is likely to make rather than one already made. Its token matches the aside form only (`[a-zA-Z] – [a-z]`), so `4–9`, `Jan 1, 2025 – Dec 31, 2025` and `Reservix – Marigold` stay legal. That matters: `Intl.DateTimeFormat.formatRange` really does emit a spaced U+2013, so the `<DateFormat>` doc lines quoting it are accurate and must not be "fixed". It is the same call #5496 made by hand, now enforced rather than remembered.

**Why the binary is downloaded rather than an npm dependency.** The `@vvago/vale` wrapper derives its download URL from its own package version and is stalled at 3.17.1, which ignores the `table.cell` exclusion and reports only the first match per paragraph (it under-reported this repo by 10 findings). `scripts/install-vale.mjs` pins 3.19.0 and verifies the SHA-256, and `renovate.json` gains a regex custom manager so the pin is visible to Renovate, with those bumps held back from automerge since 3.18 and 3.19 both changed parsing semantics.


.prettierignore
```diff
@@ -13,6 +13,7 @@
 .npmrc
 .nvmrc
 .prettierignore
+.vale.ini
```
sarahgm, line 16: 🔴 **`.vale/bin/` needs to be ignored here too, or `pnpm format` breaks for everyone.**

`.gitignore` hides the downloaded binary from git, but Prettier does not read `.gitignore`, and `prettier --write "**/*"` matches dotted directories. That is exactly why `.vale.ini` had to be added on this line. Reproduced on this branch:

```
$ npx prettier --check "**/*"
[error] No parser could be inferred for file ".vale/bin/vale".
```

The sequence hits every developer: commit once, the pre-commit hook downloads `.vale/bin/vale`, then the next `pnpm format` exits non-zero. CLAUDE.md tells people to run it before committing. CI stays green because the Format job never installs Vale, so nothing catches this before it lands.

```suggestion
.vale.ini
.vale/bin/
```

.lintstagedrc
```diff
@@ -1,3 +1,3 @@
 {
-  "*": "prettier --write"
+  "*": ["prettier --write", "node scripts/install-vale.mjs", ".vale/bin/vale"]
```
sarahgm: ⚠️ **This defeats the installer's own platform fallback.**

`scripts/install-vale.mjs` prints `no pinned build for <key>; skipping` and exits 0 when there is no asset for the platform, but the next command runs `.vale/bin/vale` unconditionally, so the commit fails with ENOENT instead of skipping.

Windows is the realistic case. There is no `win32-x64` entry in `ASSETS`, and adding one would need zip extraction and a `vale.exe` name rather than `tar xzf` plus `vale`. As it stands a Windows contributor cannot commit at all.

Either add the win32 build to the script, or route the vale step through a small wrapper that no-ops when the binary is absent, so the graceful degradation the script already implements actually reaches the caller.

sarahgm: 💡 **Two smaller things about this line.**

**Scope it.** The `*` glob hands every staged file to Vale, including PNGs, SVGs and `pnpm-lock.yaml`. Nothing matches a `.vale.ini` section so nothing is reported, but the hook pays for it on every commit. The installer also receives the same filename list and ignores it, which reads as a lint-staged task but is really a setup step, so `.husky/pre-commit` ahead of `npx lint-staged` is a better home for it.

```suggestion
  "*": "prettier --write",
  "*.{md,mdx}": ".vale/bin/vale"
```

(pairs with moving `node scripts/install-vale.mjs` into the husky hook)

**`--relative` is load-bearing and invisible.** If that flag is what makes the repo-relative `.vale.ini` section globs match, then dropping it later makes Vale silently lint nothing and exit 0 rather than failing loudly. One sentence in the CLAUDE.md Prose Style section would keep that from being rediscovered the hard way.

.changeset/prose-style-as-lint.md
```diff
@@ -0,0 +1,16 @@
+---
+'@marigold/docs': patch
```
sarahgm, line 2: ⚠️ **`@marigold/cli` needs a patch bump too.**

npm publishes `README.md` with the package, and `packages/cli/README.md` is rewritten in 57 lines here (67 of the 135 fixes by your own count). Without a bump the npm page keeps the old prose indefinitely, so the rewrite never reaches the place the README is actually read.

```suggestion
'@marigold/docs': patch
'@marigold/cli': patch
```

The alternative is dropping `{packages,themes,config}/*/README.md` from `.vale.ini` scope, which would also remove half the prose churn from this PR.

.changeset/prose-style-as-lint.md
```diff
@@ -0,0 +1,16 @@
...
+---
+
+Prose style is now enforced rather than remembered. Vale runs over the docs site, the
+changesets and the published READMEs, wired into the pre-commit hook and a CI check, with
+the rules in `.vale/styles/Marigold/` and the writing guidance in `CLAUDE.md`.
+
+Three rules block CI: no em dashes, no semicolons in prose, and no en dash asides. Table
+cells are exempt, because there an em dash is a legitimate "not applicable" marker, and
+ranges keep the en dash so quoted component output stays accurate. All 135 existing
+violations across 30 files are rewritten in this change, so the check starts green.
```
sarahgm: 💡 **The file count disagrees with itself.** This says 30 files, the PR description says 28, and the branch touches 32 prose files (33 changed minus this changeset). Since this text ships to the published changelog, either fix the number or drop it and keep the violation count.

While you are in here: DST-1526 lists "Rewriting existing docs to pass" under **Not included**. I think overruling that is right, because `vale .` lints the whole repo, so fix-forward would leave the check red on day one no matter which files a PR touches. Worth saying so explicitly in the PR description so the reporter sees it was a decision rather than an oversight.

packages/cli/README.md
```diff
@@ -211,16 +211,16 @@ marigold init --skip-install    # apply config edits without running the install
 
 Detects Next.js or Vite, then:
 
-1. Installs `@marigold/components`, `@marigold/system`, `@marigold/theme-rui` — and Tailwind v4 + the framework adapter (`@tailwindcss/postcss` for Next.js, `@tailwindcss/vite` for Vite) if Tailwind isn't already present. The Tailwind install is gated behind a confirmation prompt unless `--yes` is passed.
-2. Patches your global CSS (`app/globals.css`, `src/app/globals.css`, `src/index.css`, or `styles/globals.css`) with the required `@import` and `@source` lines — preserving existing rules.
+1. Installs `@marigold/components`, `@marigold/system`, `@marigold/theme-rui`, and Tailwind v4 + the framework adapter (`@tailwindcss/postcss` for Next.js, `@tailwindcss/vite` for Vite) if Tailwind isn't already present. The Tailwind install is gated behind a confirmation prompt unless `--yes` is passed.
```
sarahgm: ⚠️ **This changed the meaning of the step.**

The em dash separated the unconditional installs from the conditional one. As a flat comma series, `if Tailwind isn't already present` now reads as governing the whole list, so it suggests the Marigold packages themselves are only installed when Tailwind is missing.

```suggestion
1. Installs `@marigold/components`, `@marigold/system` and `@marigold/theme-rui`. Adds Tailwind v4 and the framework adapter (`@tailwindcss/postcss` for Next.js, `@tailwindcss/vite` for Vite) if Tailwind isn't already present. The Tailwind install is gated behind a confirmation prompt unless `--yes` is passed.
```

packages/cli/README.md
```diff
@@ -94,12 +94,12 @@ marigold docs getting-started/installation
...
 Flags:
 
-- `--section <name>` — `props`, `usage`, `examples`, or `all` (default: `all`)
-- `--format <name>` — `markdown` (default), `json`, or `plain`
-- `--fresh` — bypass the local cache
-- `--offline` — use only the local cache; fail if missing
+- `--section <name>`: `props`, `usage`, `examples`, or `all` (default: `all`)
+- `--format <name>`: `markdown` (default), `json`, or `plain`
+- `--fresh`: bypass the local cache
+- `--offline`: use only the local cache. Fail if missing
```
sarahgm: 💡 `Fail if missing` is left as a fragment, and the same line repeats at 130, 163 and 188 in this published README. Keeping it as one clause reads better than two sentences in a flag list:

```suggestion
- `--offline`: use only the local cache, failing if it is missing
```

docs/content/foundations/token-overview/index.mdx
```diff
@@ -424,7 +424,7 @@ Selection tells users which item has already been chosen. Not every selection lo
...
 
-A focus highlight is not a focus indicator on its own. `focus-highlight` composites to around 1.05:1 against the surface it tints, well under the 3:1 that [WCAG 1.4.11](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html) asks of a state indicator, and in a menu the same tint fires on hover because focus follows the mouse. Pair it with a focus ring — `ui-state-focus` where there is a border to recolor, `ui-state-focus-item` on borderless rows — and let the tint track the cursor while the ring proves where keyboard focus is.
+A focus highlight is not a focus indicator on its own. `focus-highlight` composites to around 1.05:1 against the surface it tints, well under the 3:1 that [WCAG 1.4.11](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html) asks of a state indicator, and in a menu the same tint fires on hover because focus follows the mouse. Pair it with a focus ring: `ui-state-focus` where there is a border to recolor, `ui-state-focus-item` on borderless rows, and let the tint track the cursor while the ring proves where keyboard focus is.
```
sarahgm: ⚠️ **The parenthetical became a false list.**

The opening dash turned into a colon and the closing dash into a comma, so `and let the tint track the cursor` now reads as the third item after the two token names rather than as the resumption of the sentence.

Keeping the closing boundary fixes it:

> Pair it with a focus ring, `ui-state-focus` where there is a border to recolor and `ui-state-focus-item` on borderless rows. Let the tint track the cursor while the ring proves where keyboard focus is.

This is the one prose edit in the set where the rewrite costs the reader something. The rest hold up well.

docs/content/foundations/spacing/index.mdx
```diff
@@ -192,9 +192,9 @@ The common case is a wrapper that breaks the container's own rhythm. A `<Panel>`
 
 #### Aligning with a bled container
 
-`--bleed-px` is the one you want when content sits edge-to-edge and needs to line back up with the container's title. Any content area rendered with `bleed` publishes it — `Panel.Content`, `Panel.CollapsibleContent`, `Card.Content`, `Card.Footer` and `Drawer.Content` — always holding that container's own horizontal padding. The same areas clear it again when they are not bled, so it is in effect only inside the nearest bled container and never leaks into a nested one that keeps its padding.
+`--bleed-px` is the one you want when content sits edge-to-edge and needs to line back up with the container's title. Any content area rendered with `bleed` publishes it: `Panel.Content`, `Panel.CollapsibleContent`, `Card.Content`, `Card.Footer` and `Drawer.Content`, always holding that container's own horizontal padding. The same areas clear it again when they are not bled, so it is in effect only inside the nearest bled container and never leaks into a nested one that keeps its padding.
```
sarahgm: 💡 Same shape as the `token-overview` comment, milder. The colon opens a list of components and `always holding that container's own horizontal padding` now dangles off the end of it, so it reads as attaching to `Drawer.Content` rather than to all five. A period after `Drawer.Content` and a fresh clause ("Each holds that container's own horizontal padding.") would keep the scope clear.

docs/content/releases/blog/release-2025-02-14.mdx
```diff
@@ -9,15 +9,15 @@ changed:
   - components/content/xloader
 ---
 
-**Marigold 11 is Here—Just in Time for Valentine's Day!** 💖 Happy Friday and Happy Valentine's Day! Whether you're celebrating with chocolates, flowers, or just a quiet night of coding, we've got something special for you—Marigold 11 is officially here! This release is packed with improvements, thoughtful refinements, and a few features we think you'll love. So, grab a heart-shaped cookie (or two), and let's dive into what's new! 🚀
+**Marigold 11 is Here: Just in Time for Valentine's Day!** 💖 Happy Friday and Happy Valentine's Day! Whether you're celebrating with chocolates, flowers, or just a quiet night of coding, we've got something special for you. Marigold 11 is officially here! This release is packed with improvements, thoughtful refinements, and a few features we think you'll love. So, grab a heart-shaped cookie (or two), and let's dive into what's new! 🚀
```
sarahgm: 💡 **Consider excluding `docs/content/releases/blog/**` from the lint.**

The generated `release.mdx` pages are already out of scope for being generated. These dated announcements are equally historical, and an em dash in "Marigold 11 is Here—Just in Time for Valentine's Day!" is idiomatic marketing typography rather than the slip the rule is aimed at. Restyling eight published posts is the one part of this cleanup that rewrites the past instead of the present, and the exclusion would shrink the diff by eight files.

Not blocking, and it is a judgement call rather than a defect.

.vale/styles/Marigold/EnDash.yml
```diff
@@ -0,0 +1,10 @@
+extends: existence
+message: 'Use a comma, a colon, or a second sentence instead of an en dash aside. German uses a spaced en dash where English uses an em dash. Ranges like `4–9` or `Jan 1 – Dec 31` are fine.'
+level: error
+nonword: true
+scope:
+  - paragraph
+  - list
+  - heading
+tokens:
+  - '[a-zA-Z] – [a-z]'
```
sarahgm, line 10: 💡 Noting the gap so it is on the record rather than rediscovered: the token misses an aside that opens with a capital or a digit, e.g. `the API – React Aria wraps it – handles focus`. Your PR description already explains that this is the price of keeping `Reservix – Marigold` and `Jan 1 – Dec 31` legal, which is the right trade. No change requested.

README.md
```diff
@@ -87,7 +87,7 @@ To start documentation follow these steps:
 
 Releases are published by CI, not locally. `.github/workflows/release.yml` runs
 [`changesets/action`](https://github.com/changesets/action) on every push to `main`: it either opens a
-"release: version packages" PR that consumes the pending changesets, or — once that PR is merged —
+"release: version packages" PR that consumes the pending changesets, or, once that PR is merged,
```
sarahgm, line 91: 💡 Unrelated to this line, but this is the nearest anchor in the diff: the **Command overview** table above (around line 78) gains no `pnpm lint:prose` row, even though CLAUDE.md points people at the script. Worth a row next to `pnpm lint`.

sarahgm (changes requested):

.prettierignore
aromko: Checked this on the branch with the binary present and it does not reproduce. Prettier 3 defaults `--ignore-path` to `[.gitignore, .prettierignore]` (`prettier --help` on 3.9.6 says so), and `.vale/bin/` is gitignored, so `prettier --check "**/*"` never reaches it:

```
$ npx prettier --check "**/*" 2>&1 | grep -c vale/bin
0
$ npx prettier --check "**/*" --ignore-path .prettierignore 2>&1 | grep -c vale/bin
1
```

The error only appears when `.gitignore` is dropped from the ignore path. `.vale.ini` needed its entry because it is tracked, not because Prettier skips `.gitignore`. Leaving `.prettierignore` as is unless you would still like the line for clarity.

.lintstagedrc
aromko: Fixed in 11685a91f. `scripts/install-vale.mjs` is now `scripts/vale.mjs`: it installs on demand, then runs Vale with the given arguments, and exits 0 with the skip notice when there is no pinned build for the platform. lint-staged, `pnpm lint:prose` and the Prose workflow all call that one script, so the fallback reaches every caller.

The win32 build is not added. Windows ships bsdtar, so it would be the zip asset plus `vale.exe`, but I cannot test it here.

aromko: Both in 11685a91f. The glob is `*.{md,mdx}` now, and the installer no longer runs as a lint-staged task.

On `--relative`: confirmed load-bearing. Vale given an absolute path matches no `.vale.ini` section and exits 0. Rather than documenting the flag, the script relativizes its arguments to the repo root itself, and the husky hook is back to plain `npx lint-staged`, so there is no invisible flag left to lose.

.changeset/prose-style-as-lint.md
aromko: Added in b0eab20a2. Kept the READMEs in scope, since npm is where that prose is read.

.changeset/prose-style-as-lint.md
aromko: Dropped both numbers from the changeset in b0eab20a2 and kept the sentence about the check starting green. Measured against main with the final `.vale.ini`, it is 126 violations across 23 files, which is what the PR description now says. The description also states that the ticket's "Not included" was overruled on purpose, with the `vale .` reasoning.

packages/cli/README.md
aromko: Applied as suggested in 45b6c186e.

packages/cli/README.md
aromko: Applied to all four occurrences in 45b6c186e.

docs/content/foundations/token-overview/index.mdx
aromko: Applied your wording in 45b6c186e.

docs/content/foundations/spacing/index.mdx
aromko: Applied in 45b6c186e. The list ends at `Drawer.Content`, and "Each holds that container's own horizontal padding." follows as its own sentence.

docs/content/releases/blog/release-2025-02-14.mdx
aromko: Excluded in b0eab20a2. `docs/content/releases/blog/**` gets an empty `BasedOnStyles` like the generated `release.mdx` pages, the eight posts are reverted to main, and CLAUDE.md names the exclusion so it is not rediscovered. Trade-off noted: future posts go unlinted too.

.vale/styles/Marigold/EnDash.yml
aromko: Agreed, leaving it as is.

README.md
aromko: Added the row after `pnpm lint` in 45b6c186e.

sarahgm (approved):

PR #5780, opened 2026-09-03
feat(DST-1523): add the /pick-up skill

sarahgm:
# Description

Adds `/pick-up`, a skill that turns a DST ticket into a loaded, planned session. It gathers the ticket with its comments, links and attachments, takes precedent from `git log` on the files the ticket names, gates on whether the ticket is startable at all, and writes a spec and ordered plan to `.claude/tasks/` (untracked). Only after you approve that plan does it create the branch, self-assign and move the ticket to In Progress.

The step from "I'll take DST-1234" to writing code is the most repeated cold start in the workflow, and none of it was written down.

Two supporting changes come with it. `.claude/README.md`'s confirmation rule now reads "immediately before the first outward call" rather than "step 1", because a skill that gathers before it writes cannot put its gate in step 1. And `.memory/README.md` drops the `tasks/` row it had reserved for this skill: a per-ticket plan is dead the day the ticket ships, so it fails that store's durable-and-cross-session write rule.

N/A — no UI changes, and nothing published changes, so no changeset.

Closes DST-1523

## Test Instructions

1. Read `.claude/skills/pick-up/SKILL.md`.
2. Run `/pick-up` with no argument. The queue should list startable open-sprint tickets with key, title, status and assignee, and should not auto-take the top row.
3. Run `/pick-up` on any Ready ticket. It should gather, write a plan under `.claude/tasks/`, render a summary at step 6 and a branch-plus-Jira preview at step 7. Choose **Cancel**: no branch should exist and the ticket should be untouched.
4. On a ticket you actually want, choose **Start** and confirm the branch type derives from the title emoji, that you are assigned, and that the status moves to In Progress.

## Breaking Changes

No

## Checklist


aromko:
@sarahgm general question: Did you use the skill-creator from Anthropic for the new skills? I think it's helpful to test the skills and to evaluate the results both qualitatively and quantitatively. Otherwise just from reading I can't really tell if it's working well or not.

sarahgm:
> @sarahgm general question: Did you use the skill-creator from Anthropic for the new skills? I think it's helpful to test the skills and to evaluate the results both qualitatively and quantitatively. Otherwise just from reading I can't really tell if it's working well or not.

I thought they start automatically when i say i want to create a skill, I tested the skill through running the mentioned skill e.g. /pick-up new issue. But I can add evals if you want

aromko:
> > @sarahgm general question: Did you use the skill-creator from Anthropic for the new skills? I think it's helpful to test the skills and to evaluate the results both qualitatively and quantitatively. Otherwise just from reading I can't really tell if it's working well or not.
> 
> I thought they start automatically when i say i want to create a skill, I tested the skill through running the mentioned skill e.g. /pick-up new issue. But I can add evals if you want

I think it would make sense to add evals. I also think we should move them all later to an own marketplace so it is like a preparation for later.

aromko (approved):

PR #5781, opened 2026-09-03
feat(DST-1524): add the /triage-feedback skill

sarahgm:
# Description

Adds `/triage-feedback`, a committed skill that makes one pass over all feedback on a PR from both places it arrives: GitHub review threads and Vercel preview toolbar comments. It triages every item on Validity, Severity and Action, renders the table as the approval gate, then replies and resolves in whichever system each item came from.

Two things shape the skill beyond the ticket's outline:

- **Two gates, not one.** The triage table approves every reply and resolve. The push carries its own separate confirmation, because pushing is governed by a standing rule of its own. Steps 1 to 3 are read-only, nothing before step 5 changes a file, and nothing before step 7 leaves the machine.
- **Two modes.** On your own PR you are the author and the feedback is work incoming. On someone else's you are the reviewer, so there is nothing to fix and the skill skips the apply and push steps entirely. Without that distinction the skill tries to commit fixes to a branch it does not own.

Visual and design comments route straight to `needs-human` instead of being judged from a screenshot, which is the v1 limit the ticket asks for.

Also adds the `gh` body-file convention to `CLAUDE.md`, per the ticket's scope. The trap worth recording is that `gh api -f body=@reply.md` posts the literal string `@reply.md` and **exits 0**, so a nonsense comment lands publicly with no error to catch. `-F` is the flag that reads the file.

Exercised against #5776, the popover clip-box fix, as a real respond-only run rather than only against the checks below. That PR has resolved and outdated threads on it, so it covers the `isOutdated` and already-resolved paths.

No changeset and no visual regression run: this touches no UI files and no published package. The sibling skill PRs #5778 and #5780 carry no changeset either.

Closes DST-1524

## Merge order

Two known conflicts with PRs already in flight, both in files this branch deliberately leaves alone:

1. **`.claude/README.md`** currently requires confirmation to be step 1 of a skill's workflow. This skill gates at steps 4 and 6, which is exactly what #5778 and #5780 each amend that rule to permit, in different wordings. Adding a third edit to the same line would turn one conflict into three, so land #5778 or #5780 first and then rebase. Step 4's reference to `.claude/README.md` depends on the same ordering.
2. **`CLAUDE.md`** gains a `## GitHub CLI` section at the same location #5779 edits. Expect a conflict with whichever of the two merges second.

## Test Instructions

Every acting step sits behind an approval gate, so it is safe to run as far as the table and stop.

1. Run `/triage-feedback 5776` against someone else's PR. Confirm the mode line reads **respond-only**, the table lists every unresolved thread, and nothing is offered to apply or push.
2. Run `/triage-feedback` on a branch of your own that has an open PR. Confirm the mode reads **author** and that the plan below the table names how many fixes and which files.
3. Choose **Stop** at the table both times. Confirm nothing was replied to, resolved, committed or pushed.
4. Spot-check the two API calls the skill leans on:
   - the step 2 GraphQL query returns `isResolved`, `isOutdated` and `pageInfo.hasNextPage`
   - `gh api -H "Accept: application/vnd.github.raw" "repos/marigold-ui/marigold/contents/<path>?ref=<sha>"` returns plain source rather than base64

## Breaking Changes

No

## Checklist


.claude/skills/triage-feedback/SKILL.md
```diff
@@ -0,0 +1,306 @@
+---
+name: triage-feedback
+description: Marigold repo — Make one triage pass over all feedback on a PR, from both GitHub review threads and Vercel preview toolbar comments, then reply and resolve in whichever system each item came from. Use when the user asks to "triage feedback", "go through the review comments", "handle the PR feedback", "address the preview comments", or types `/triage-feedback`. It posts replies, resolves threads and pushes commits, so run it only on an explicit request, never proactively and never as a follow-up to unrelated work.
```
aromko, line 3: Can we remove marigold repo specific naming. Would also use it in insights and maybe also in the core. So it should ask me which repo or which PR I would like to triage.

.claude/skills/triage-feedback/SKILL.md
```diff
@@ -0,0 +1,306 @@
...
+### 7. Reply and resolve
+
+Reply in the origin system. Reply to every row where `Turn` is `yours`, plus any `apply` row you fixed, whichever way its Turn points.
+
+What a reply says depends on the mode:
+
+- **author** — what changed and where. Name the commit when a fix landed **and the branch was pushed**. On the reply-only path the SHA is local and resolves to nothing for the reader, so say the fix is made and pending a push instead.
+- **respond-only** — you are answering as the reviewer. Accept the author's response, or say what still does not hold and why. If they asked you something, answer it. Do not describe fixes: they are not yours to make.
+
+#### GitHub
```
aromko, line 258: would do it also for gitlab or independant of the platform.

.claude/skills/triage-feedback/SKILL.md
sarahgm: Repo is resolved at run time now rather than hardcoded, in f2cacd996: `--repo <owner>/<name>` names it, otherwise `gh repo view` reads it from the checkout's remote, and the owner/name pair is passed to every `gh` call. The mode table gained a cross-repo row, because the worktree being a different repository means no fix can land there. Step 5 no longer assumes `pnpm`.

On being asked which repo: this skill can't use `AskUserQuestion`, since under `defaultMode: auto` it resolves to the first option without a human ever seeing it. So the order is `--repo` first, then the checkout's remote. That covers insights, which is also GitHub.

What it does **not** do is let you type `/triage-feedback` inside the insights or core checkout. A skill in `.claude/skills/` only loads in the repo it's committed to, so real reuse means moving this into the `rx-baseline` plugin. I've kept that as a follow-up rather than doing it here, since it lands in `rx-ai-suite` and wants review from that side.

Leaving this thread open for your read on the split.

.claude/skills/triage-feedback/SKILL.md
sarahgm: Pushing this to the follow-up rather than doing it here. The endpoints map cleanly (`POST .../discussions/:id/notes`, `PUT .../discussions/:id` with `resolved=true`) and `glab api` has the same `-F`-reads-a-file trap as `gh api`, so the reply half ports almost directly. Three things don't:

- GitLab has no `isOutdated`, which is the strongest staleness signal the GitHub half uses. The closest substitute is comparing the note's `position.head_sha` to the MR's `diff_refs.head_sha`, which is a proxy rather than an authority.
- Only `resolvable` discussions can be resolved, so "reply and resolve" splits into "reply, then resolve if resolvable".
- The `glab` wrapper mounts only the repo root into its container, so a reply body written to `/tmp` isn't visible to it.

It also needs `Skill(gitlab)` or a `${CLAUDE_PLUGIN_ROOT}` path, and that variable only resolves inside a plugin, so the GitLab adapter can't live in `.claude/skills/` regardless. Same conclusion as the other thread: plugin move first.

Also worth flagging for that ticket: core has no Vercel previews, so there the skill collapses to the GitLab source alone.

.claude/skills/triage-feedback/SKILL.md
aromko: I wouldn't put this into the rx-ai-suite. I think we should made our own because it is our workflow but shareable in the end.

.claude/skills/triage-feedback/SKILL.md
sarahgm: Correcting myself: the reason I gave for deferring this was wrong, so it is no longer deferred. GitLab MRs work in this branch now, in `bef134b40` and `49cd18f78`.

I said the GitLab adapter could not live in `.claude/skills/` because it needs `${CLAUDE_PLUGIN_ROOT}`. It does not. A repo-local skill can reach a plugin skill, and ours already does: `review-pr` calls `Skill(vercel:react-best-practices)` at `.claude/skills/review-pr/SKILL.md:145`. Step 1 does the same with `Skill(rx-baseline:gitlab)` and uses the absolute wrapper path that returns, so nothing is hardcoded against the version-pinned plugin cache.

My second objection was that nothing in this repo could exercise a GitLab adapter. Also wrong, and for a reason the skill already had written down: the mode table treats a target outside this checkout as a reply pass. An MR is just another such target, so it runs from here against a real MR.

What landed:

* Target and host both come from the URL. Any project on any authenticated host, nested subgroups included. Every call carries `--hostname`, because without it `glab` silently answers from whatever its config names, and a pasted `gitlab.com` URL would have gone to the wrong server.
* Always a reply pass. The worktree is marigold, so steps 5 and 6 never run.
* No Vercel source on an MR, since previews build from the GitHub repo. Passing `--vercel` with `--gitlab` is refused rather than ignored.
* A missing rx-baseline stops the pass with the install command. No degrading: on a GitLab target the wrapper is the only source there is, and a `.claude/skills/` skill has no `dependencies` field to declare that with.

Two things testing corrected that I would have shipped wrong:

1. **Resolvability is a flag, not a note type.** I had written that GitLab only resolves threads anchored to a diff. It does not: a `DiscussionNote` with no diff position is resolvable. Read `resolvable`. A plain comment is the non-resolvable case, and carries `resolved: null` rather than `false`.
2. **The discussions endpoint returns the activity feed.** "added 1 commit", "assigned to", "changed title from", all shaped exactly like comments. Six of nine discussions on `core/main!36307`, twelve of twenty-three on `core/main!36515`. Without the filter every one would have been triaged and replied to. GitHub has no equivalent, which is why the first pass missed it.

Verified read-only against `core/main` and `portal/main`: the MR fetch, discussions, changes, and a raw file read at a commit SHA. The reply and resolve calls are documented but not yet exercised, since that means posting to a real MR.

.claude/skills/triage-feedback/SKILL.md
sarahgm: Understood, and that is the clearer answer than the one I proposed. A Marigold-owned marketplace it is, not rx-ai-suite.

Worth knowing before we write the ticket: we have done this once already. `.claude-plugin/marketplace.json` sat at this repo's root from 26 Mar (DST-1262, #5238) under the name `marigold-plugins`, hosting `create-marigold-app` out of `packages/create-marigold-app/`. It was removed on 7 Apr in #5273, when that skill moved into `rx-frontend` in rx-ai-suite. So this reverses an April decision rather than opening a new question, and the ticket should say why, or the next person undoes it again.

Two findings from looking into the mechanics:

* A consuming repo declares `extraKnownMarketplaces` plus `enabledPlugins` in its `.claude/settings.json`. The marketplace registers itself on folder trust, but since 2.1.195 an external-source plugin still needs one `claude plugin install` per person. Claude Code prints the exact command, so it is a one-time paper cut rather than a blocker.
* Depending on `vercel` and `atlassian` from `claude-plugins-official` works, but only if our `marketplace.json` declares `allowCrossMarketplaceDependenciesOn: ["claude-plugins-official"]`. I validated a prototype of both manifests with `claude plugin validate`.

The duplication you are trying to prevent is already live, incidentally: `marigold-ui/insights` carries its own forked `review-pr`, rewritten for Next.js, plus a `skills-lock.json` vendoring six skills from vercel-labs. And this repo's `.claude/README.md` already records that vendoring "stopped tracking upstream within two months".

One thing I would not move wholesale. `create-pr`, `vrt` and `grill` are bound to changesets, Chromatic and `.memory/`, so they belong here. `triage-feedback`, `pick-up` and `review-pr` are about GitHub, GitLab, Vercel and Jira, and those are the ones worth sharing.

Separately: the GitLab half is no longer waiting on any of this. It landed in this branch, see the other thread.

aromko (approved):

PR #5786, opened 2026-09-04
ci(DST-1527): add claude-code-action workflows for AI review in CI

sarahgm:
# Description

`/review-pr` is powerful but local and opt-in: it only runs when the author remembers, on the author's machine. This adds two GitHub Actions workflows so the same standards have a home in CI.

**`claude-mentions.yml` — ships live.** Anyone with write access can write `@claude` in a PR thread to ask a question or ask for a fix, and Claude answers or pushes a commit to the PR branch.

**`claude-review.yml` — ships live, routed by category.** Posts a baseline review against the checklist `/review-pr` already owns. The `classify` job reads the Conventional Commits type from the PR title and picks one of four profiles: `full`, `docs`, `tests`, or skip, by the mapping tallied below. That routing is the whole gate, so which PRs get reviewed is a property of the title convention the team already follows rather than a setting anyone has to remember. `gh workflow disable "Claude Review"` turns it off without a PR if the signal-to-noise disappoints.

It cannot block a merge. claude-code-action is unable to submit formal GitHub reviews or approve a PR at all, so this is advisory by construction rather than by configuration.

The review prompt points at `.claude/skills/review-pr/references/review-checklist.md` instead of restating it, so the agreed standards stay in one place and keep being reviewed like code.

Closes DST-1527

## Authentication: Amazon Bedrock, no personal tokens

Both workflows assume an IAM role through GitHub OIDC (`aws-actions/configure-aws-credentials`, SHA-pinned) and run against Bedrock. Credentials are short-lived STS sessions, so no long-lived key is stored in this public repo, and usage bills to the AWS account rather than to any individual's subscription. That last point is the answer to the token-usage question on the ticket: the credential is fixed per run and is never drawn from the quota of whoever typed `@claude`, whose identity is used only for the permission check.

The existing `AWS_BEDROCK_ACCESS_KEY_ID` / `AWS_BEDROCK_SECRET_ACCESS_KEY` secrets are deliberately **not** reused. They are static keys scoped to Titan embeddings for `release.yml`, and reusing them would mean widening that policy to model invocation. A separate role with its own narrow `bedrock:InvokeModel*` policy is used instead.

Neither job declares a GitHub environment. The two workflows trigger differently, so their OIDC subjects are already precise enough to scope the trust policy on directly:

| Workflow | Event | OIDC subject |
| --- | --- | --- |
| `claude-review.yml` | `pull_request` | `repo:marigold-ui/marigold:pull_request` |
| `claude-mentions.yml` | `issue_comment` | `repo:marigold-ui/marigold:ref:refs/heads/main` |

That pins mentions to the default branch, which is the only place `issue_comment` ever runs, and review to pull requests. An environment would add a second name to keep in sync with the IAM policy without narrowing anything further, and this repo's existing environments were all created by Vercel and Pages rather than used deliberately.

Fork PRs are not granted `id-token: write`, so they cannot assume the role at all, which backs up the explicit fork guard already in the mentions workflow.

### GitHub identity: a dedicated App for the mentions workflow

The mentions workflow authenticates to GitHub with its own App (`actions/create-github-app-token`), the same pattern `renovate.yml` already uses. This is not cosmetic. GitHub **does not start workflow runs from commits created with `GITHUB_TOKEN`**, so a fix Claude pushed would never be built, typechecked or tested. The earlier version of this branch had that hole: the spec recorded the same mechanism as a benefit ("no bot re-trigger loop") without noticing it cuts both ways.

The review workflow keeps `GITHUB_TOKEN`, since it only ever comments.

The App needs three repository permissions: **Contents**, **Issues** and **Pull requests**, all read and write.

### Write access is verified before the cloud credentials

Anthropic's guidance carries a warning aimed exactly at a public repo like this one: a trigger comment from any user starts the workflow, and the credential steps run before the action checks the commenter's write access, so an unauthorised comment still burns Actions minutes and leaves audit entries.

The `gate` job already keeps the AWS step in a separate job behind `needs:`, but `author_association` is a pre-filter, not a write check: `MEMBER` and `COLLABORATOR` do not imply write access. There is now a real check against the collaborator permission API, ahead of the AWS step.

It is an **allowlist** of `admin` and `write`, deliberately not a denylist of `none`. On a public repository that endpoint answers `read` for someone with no relationship to the repo at all, so testing for `none` would have let everyone through. Verified against this repo's API.

### Region and model, from existing precedent

`core/main` already runs Claude on Bedrock in CI, so the region and model questions are settled rather than open:

- **Region `eu-central-1`.** The data-residency answer, and demonstrably workable rather than a hope.
- **`eu.anthropic.claude-sonnet-5`** for `AI_REVIEW_MODEL_ID`.
- **`eu.anthropic.claude-opus-5`** for `AI_MENTIONS_MODEL_ID`.

Both were then confirmed directly against the design system team's own AWS account: `bedrock list-inference-profiles --region eu-central-1` lists both as `ACTIVE`, alongside an EU Haiku profile. So decision 7 stands as written, Opus for mentions and Sonnet for the automatic review, and no Sonnet fallback is needed.

That also shrinks the ask. Bedrock model access is already enabled in the target account, so this is not a Bedrock onboarding. It is only an identity change: nothing there can currently call `InvokeModel` from GitHub.

We deliberately differ from `core/main` in three ways, all of which predate this precedent: their review is **manual** (`when: manual`, never on a merge request event) where ours is automatic by category, their credentials route through **Vault** where ours assume an IAM role directly, and theirs is **read-only** where our mentions workflow lets Claude commit a fix on request.

## Which PRs get reviewed: category routing

The automatic review runs only on the categories worth reviewing, decided by the Conventional Commits type in the PR title.

**The `type:*` labels cannot do this job.** They are derived from changed paths, and the glob `packages/*` matches a package *manifest*, so every Renovate bump inside a package is labelled `type:feature`. #5827 changed only `packages/cli/package.json` and `pnpm-lock.yaml` and carries `type:feature`; same for #5818, #5820, #5821 and #5822. `labeler.yml` also runs `on: pull_request` in parallel with this workflow, so at `opened` the label may not exist yet.

The mapping was tallied against the **last 400 merged PRs** rather than assumed, which corrected three guesses:

- `perf` has never been used here (0 of 400). Dropped.
- `refa` is used as a synonym for `refactor` (3 of 400). Aliased.
- `style` is **not** formatting in this repo. All four `style:` PRs are substantive design changes to component source carrying changesets, and #5589 `style(DST-1586)` changed AA labels. It gets a full review. Skipping it would have skipped exactly what the checklist exists for.

| Title type | n / 400 | Profile |
| --- | --- | --- |
| `feat` / `fix` | 67 / 88 | full |
| `refactor` + `refa` | 11 + 3 | full |
| `style` | 4 | full |
| `docs` | 45 | docs |
| `test` | 9 | tests |
| `chore` | 151 | skip |
| `ci` / `build` / `release` / `revert` | 6 / 1 / 9 / 1 | skip |
| unparseable | 4 | full |

Titles are parsed after stripping a leading emoji, which rescues `🐛 fix(DSTSUP-256): …` into `fix`. Unparseable titles fail *toward* reviewing, which is not a theoretical branch: the remainder are ticket keys used as the type (`DST-1409:`, `[DST-1396]:`). Nothing lints the title, so a skip-category PR that touches `packages/*/src/**` or `themes/*/src/**` is escalated to a full review. There is no de-escalation in the other direction.

Running the shipped classifier over those 400 titles gives **177 full, 169 skip, 45 docs, 9 tests**. Category routing is also what makes the running cost reasonable: `chore` alone is 151 of 400 and now costs nothing.

### One bug worth flagging, found while testing this

The escalation check originally piped `gh pr diff --name-only` straight into `grep -q`. That is broken under `set -euo pipefail`: `grep -q` exits on the first match, `gh` takes SIGPIPE, and `pipefail` turns the pipeline non-zero, which reads as "no source changed". Verified with a 20000-line all-matching mock, it returned `skip`. It failed in the **unsafe** direction, silently skipping review of a PR that did touch component source. It now captures the list first and uses `grep -c`, which must read all of its input.

## Why this is still a draft

The workflows cannot run until the AWS side exists. That needs someone with AWS admin to:

1. **Submit the Bedrock use-case form once for the account.** Required before the first Anthropic invoke. Under AWS Organizations it can be done once from the management account with `PutUseCaseForModelAccess`, and approval extends to child accounts.
2. ~~Add a GitHub OIDC identity provider.~~ **Already done.** One for `token.actions.githubusercontent.com` with audience `sts.amazonaws.com` was created on 2026-09-15, though in the shared coding-agent account rather than the design system team's own. Which account the role lands in still needs deciding, and Bedrock model access is granted per account, so it needs confirming for whichever one is chosen.
3. Create an IAM role trusting that provider, allowing **both** subjects from the table above, and attach Anthropic's scoped invocation policy verbatim:

   ```json
   "Condition": {
     "StringEquals": { "token.actions.githubusercontent.com:aud": "sts.amazonaws.com" },
     "StringLike": {
       "token.actions.githubusercontent.com:sub": [
         "repo:marigold-ui/marigold:pull_request",
         "repo:marigold-ui/marigold:ref:refs/heads/main"
       ]
     }
   }
   ```

   Both are needed. A policy allowing only `:pull_request` would let the automatic review run and silently refuse every `@claude` mention.

   ```json
   {
     "Version": "2012-10-17",
     "Statement": [
       {
         "Sid": "AllowModelAndInferenceProfileAccess",
         "Effect": "Allow",
         "Action": [
           "bedrock:InvokeModel",
           "bedrock:InvokeModelWithResponseStream",
           "bedrock:ListInferenceProfiles",
           "bedrock:GetInferenceProfile"
         ],
         "Resource": [
           "arn:aws:bedrock:*:*:inference-profile/*",
           "arn:aws:bedrock:*:*:application-inference-profile/*",
           "arn:aws:bedrock:*:*:foundation-model/*"
         ]
       },
       {
         "Sid": "AllowMarketplaceSubscription",
         "Effect": "Allow",
         "Action": ["aws-marketplace:ViewSubscriptions", "aws-marketplace:Subscribe"],
         "Resource": "*",
         "Condition": { "StringEquals": { "aws:CalledViaLast": "bedrock.amazonaws.com" } }
       }
     ]
   }
   ```

   The two inference-profile actions and the marketplace statement are easy to drop as unnecessary and are not. Without `ListInferenceProfiles`, Claude Code cannot check profile availability and applies the region prefix blind, so a missing profile surfaces as an opaque 400 instead of a clear error. Without `GetInferenceProfile`, every new model costs an extra round trip.
Two things that were open here are now settled and need no further work. The models are confirmed `ACTIVE` in `eu-central-1`, as above, and the account is decided, since the design system team already has its own. That matches Anthropic's dedicated-account suggestion without reusing `core-test`.

Worth knowing for whoever picks this up: there are currently **no OIDC identity providers at all** in that account, not even a GitLab one, so this would be the first. The roles reachable there also carry `TerraformStateLockAccess`, which suggests IAM is managed as code. If that holds, a role created by hand in the console would be reverted on the next apply.

Bedrock does not accept API aliases such as `claude-sonnet-5`, which is why the model IDs live in variables rather than in the workflow.

Note that Bedrock introduces a failure mode the earlier federation approach did not have: on-demand throughput is per account and region, so a burst of PRs can throttle.

Three things can only be confirmed on a live run, and the smoke test below is what confirms them:

1. `--append-system-prompt` actually forwarding through `claude_args`.
2. Whether Claude can run `pnpm` in the mentions job. The job puts Node and pnpm on PATH but deliberately does not install, because an install there would resolve the base branch's lockfile. Claude installs on the PR branch itself, and only when changing code. If the action gates Bash by default, that instruction is a no-op and the job needs a narrow `--allowedTools`.
3. That the review posts at all. An explicit `--allowedTools` was removed after code review because it omitted the action's own comment tool and left `gh` without `GH_TOKEN`.

## Test Instructions

Nothing here runs until step 1, so CI on this PR proves only that the YAML is valid. It does pass Prettier, and the classifier was exercised against all 400 real titles plus six real PR file shapes.

1. Complete the AWS provisioning above, then set these at Settings → Secrets and variables → Actions. The role ARN and the App key are secrets, the rest are variables on the same page as `VRT_ENABLED`. No GitHub environment is needed.

   | Name | Kind | Value | Status |
   | --- | --- | --- | --- |
   | `AI_REVIEW_BEDROCK_ROLE_ARN` | secret | the new role's ARN | pending AWS |
   | `AI_REVIEW_BEDROCK_REGION` | variable | `eu-central-1` | |
   | `AI_REVIEW_MODEL_ID` | variable | `eu.anthropic.claude-sonnet-5` | |
   | `AI_MENTIONS_MODEL_ID` | variable | `eu.anthropic.claude-opus-5` | |
   | `CLAUDE_APP_ID` | variable | `4962048` | set 2026-09-16 |
   | `CLAUDE_APP_PRIVATE_KEY` | secret | the App's `.pem` contents | set 2026-09-16 |

   `CLAUDE_APP_PRIVATE_KEY` matches how `RENOVATE_APP_PRIVATE_KEY` is already stored.

   Merge once the role exists. Both workflows are live from the moment this lands on `main`, and having the role in place first is what makes that landing safe. Afterwards, `gh workflow disable "Claude Mentions"` or `"Claude Review"` switches either half off.
2. **Before merging, prove the review from this branch.** A `pull_request` event resolves the workflow from the PR branch, so the full Bedrock path can be exercised while only this branch's scratch PRs are in scope. Open one titled `chore: …` touching only `pnpm-lock.yaml` and confirm the review is **skipped**. Open one titled `feat(DST-…): …` touching `packages/components/src/**` and confirm exactly one sticky comment appears and is rewritten rather than duplicated on the next push.
3. Merge, since `issue_comment` workflows only run from the default branch.
4. On a throwaway PR, comment `@claude explain what this file does`. Expect a reply in the thread.
5. On the same PR, comment `@claude add a missing forwardRef here`. Expect a real commit on the PR branch, and check the log for whether the typecheck actually ran.
6. Ask someone without write access to comment `@claude hello`. Expect nothing to happen.
7. Comment `@claude hello` on an already-merged PR. Expect nothing to happen.

## Breaking Changes

No

## Checklist

Not applicable to a CI-only change, other than the changeset line.

- [x] Storybook preview and Marigold docs preview are available — N/A, no UI change
- [x] Stories added/updated (with `component-test` tag where applicable) — N/A
- [x] Unit tests added/updated — N/A. The classifier was verified against the real 400-title history and six PR file shapes, but there is no test harness in this repo for workflow shell.
- [x] Component documentation added/updated (if it exists) — N/A. `.claude/README.md` gains a section on how the CI layer relates to `/review-pr`.
- [x] Accessibility reviewed against [ARIA APG](https://www.w3.org/WAI/ARIA/apg/) — N/A
- [x] Visual regression tests updated (for UI changes) — N/A, no UI change
- [x] Changeset added (`pnpm changeset`) — N/A. CI-only commits in this repo carry none, cf. `ci(DST-1721)` and `b0120e57f`. No published package changes.

[DST-1396]: https://reservix.atlassian.net/browse/DST-1396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,249 @@
...
+          esac
+
+          # Nothing lints the title, so any narrowed profile touching real
+          # source is promoted. Matching src/** excludes package.json.
+          if [ "$profile" != "full" ]; then
+            # Not `| grep -q`: grep exits on first match, gh takes SIGPIPE, and
+            # pipefail reads that as "no source changed" — unsafe direction.
+            changed=$(gh pr diff "$PR_NUMBER" --repo "$REPOSITORY" --name-only)
+            src_changed=$(printf '%s\n' "$changed" \
+              | grep -cE '^(packages|themes)/[^/]+/src/' || true)
```
aromko: **Warning** — the escalation pattern is `^(packages|themes)/[^/]+/src/`, so it does not cover `docs/`. That contradicts the rest of the file: `paths-ignore` at lines 12-15 was deliberately kept shorter than `visual-regression-tests.yml` so the workflow *does* run on `docs/**` ("a Next.js app with ~600 reviewable .tsx files"), and the prompt at line 207 carries a whole `docs` profile for the `.tsx` demo and app files under `docs/`.

A `chore:` or `ci:` titled PR that touches only docs app source therefore gets skipped rather than escalated. Of the last 150 merged PRs, four non-bot ones land in exactly that hole: #5853 (22 docs `.ts`/`.tsx` files), #5742, #5739 and #5738.

```sh
| grep -cE '^(packages|themes)/[^/]+/src/|^docs/.*\.tsx?$' || true)
```

aromko (commented):
## PR Review: #5786 — ci(DST-1527): add claude-code-action workflows for AI review in CI

### Overview

| Field | Value |
| --- | --- |
| Author | @sarahgm |
| Branch | `feat/DST-1527-ai-review-ci` → `main` |
| Files changed | 4 (+462 / -1) |
| Conventions source | `CLAUDE.md` + `.claude/skills/review-pr/references/review-checklist.md` |
| Visual regression | Not needed (no story, style, template or rendered output in the diff) |

Three warnings and two suggestions are attached as inline comments on the lines they concern.

### Linked ticket

**DST-1527**: 🏗️ AI review in CI via claude-code-action — In Review

The ticket scoped a staged rollout: "Start with `@claude` mentions only; expand to auto-review on `pull_request` once we trust the signal-to-noise." This PR ships both halves in one merge, with category routing standing in for the trust-building phase. That is a reasonable substitution and the description argues it, but it is a deviation from the written scope and worth a sentence on the ticket rather than only in the PR body. The other deviation, Bedrock plus OIDC instead of `/install-github-app` and an API key secret, is strictly better than what the ticket asked for and is documented at length.

The evidence behind the routing table is the strongest part of this PR: 400 real titles tallied rather than assumed, three guesses corrected, and the `SIGPIPE` bug in the escalation check found and fixed with a reproduction. The security posture is also right. Forks cannot reach the role, write access is an allowlist verified against the collaborator API ahead of the AWS steps, `permissions: {}` is the default, and PR content is declared untrusted in both prompts.

### Findings

#### Critical (must fix)

- `.github/workflows/claude-review.yml:196` and `.claude/README.md:114` — the review prompt points at `.claude/skills/review-pr/references/review-checklist.md`, which the unmerged branch `chore/DST-1778-extract-dst-skills-plugin` (commit `a8c7ea9`) deletes when it moves the seven DST skills out to the `dst-toolkit` plugin. Whichever of the two lands second leaves the prompt pointing at a path that is not in the repo, and the prompt has no fallback: the model reads nothing, silently falls back to `CLAUDE.md` alone, and the README line added here, "The checklist has one home", becomes false. Nothing fails, so nobody finds out.

  Two things make this worse than a normal merge race. DST-1778 rewrites `.claude/README.md` heavily but keeps the `## Adding and removing` heading this section is inserted above, so git will most likely auto-merge the new section in with no conflict marker. And the plugin copy is installed per user from the private marketplace, so CI cannot read it: repointing the path is not available as a fix.

  The workable options are to keep the checklist in this repo as a plain file that is not a skill (`.claude/review-checklist.md`, with the plugin skill referencing it) and point the prompt there, or at minimum add a guard step so the degradation is loud:

  ```yaml
  - name: Check the review checklist is present
    run: test -f .claude/skills/review-pr/references/review-checklist.md
  ```

  Worth agreeing the merge order with @aromko before either lands.

### Recommendation

**Comment**

Unusually well evidenced for a CI change: the routing table is measured against real history rather than guessed, the sticky-comment and SIGPIPE problems were both found by testing rather than reasoning, and the Bedrock and identity decisions are each argued from precedent that exists. Nothing in the diff is wrong as it stands today, which is why this is a comment and not a request for changes. The one blocker is a collision the author does not own alone: DST-1778 deletes the checklist this prompt depends on, and the failure mode is silent, so it needs a merge-order decision before either branch lands. The docs escalation gap is a two-token fix to the `grep` pattern and worth taking in this PR, since it is the same class of "which PRs get reviewed" reasoning the rest of the file gets right.

.github/workflows/claude-mentions.yml
```diff
@@ -0,0 +1,198 @@
...
+            echo "proceed=true" >> "$GITHUB_OUTPUT"
+          fi
+
+  claude:
+    needs: gate
+    if: needs.gate.outputs.proceed == 'true'
+
+    name: Respond to @claude
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
```
aromko: **Warning** — `timeout-minutes: 20` against `--max-turns 60` (line 197) and a system prompt that tells Claude to run `pnpm install --frozen-lockfile` then `pnpm typecheck:only` before committing (line 198). Install plus typecheck on this monorepo is a few minutes on its own, so a fix request that uses its turn budget will hit the timeout.

A timeout cancels the job mid-run, which is the half-finished push the concurrency comment at line 16 gives up `cancel-in-progress` to avoid. Either raise the timeout to match a 60-turn budget or lower the turns to match 20 minutes.

.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,249 @@
...
+        with:
+          app-id: ${{ vars.CLAUDE_APP_ID }}
+          private-key: ${{ secrets.CLAUDE_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: ${{ github.event.repository.name }}
+          permission-contents: read
+          permission-pull-requests: write
+
+      - name: Checkout repository
+        uses: actions/checkout@v6
```
aromko: **Warning** — `actions/checkout@v6`, where the repo is on `actions/checkout@v7` in all 20 existing uses across `.github/workflows/`. Renovate will open a bump the week after this merges. Same on `claude-mentions.yml:128`.

.github/workflows/claude-mentions.yml
```diff
@@ -0,0 +1,198 @@
...
+          perm=$(gh api "repos/${REPOSITORY}/collaborators/${ACTOR}/permission" -q .permission)
+          case "$perm" in
+            admin|write) echo "${ACTOR} has ${perm} access." ;;
+            *) echo "::error::${ACTOR} has '${perm}' access, not write. Refusing."; exit 1 ;;
+          esac
+
+      # No `ref:`. The action manages the branch itself, and its security guide
+      # forbids checking out an untrusted ref ahead of it.
+      - name: Checkout repository
+        uses: actions/checkout@v6
```
aromko: **Warning** — `actions/checkout@v6`, same as `claude-review.yml:127`. The repo is on `actions/checkout@v7` everywhere else.

.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,249 @@
...
+               severity guide. Read it. Do not restate it, apply it.
+            2. `CLAUDE.md` — project conventions, already loaded for you.
+
+            The REVIEW PROFILE above narrows what to spend attention on:
+
+            - `full`: the whole checklist. Component and system source is the
+              subject. This includes PRs titled `style:`, which in this
+              repository means a visual or design change to real component code,
+              not formatting.
+            - `docs`: documentation content and the docs app. Prose, MDX, and
```
aromko, line 224: **Suggestion** — the `docs` profile tells the reviewer to look at "Prose, MDX", but `paths-ignore` at lines 13-14 means a prose-only PR never starts the workflow at all. #5728, #5722 and #5834 are all in that shape.

MDX is only ever reviewed when it rides along with a non-md file, which is fine given Vale gates prose separately, but the comment at line 11 is the place to say so before someone reads the profile and tries to "fix" the missing prose review.

.github/workflows/claude-mentions.yml
```diff
@@ -0,0 +1,198 @@
...
+      contents: write # Claude pushes its fix as a commit on the PR branch
+      pull-requests: write
+      issues: write # a PR conversation comment is an issue comment
+      id-token: write # exchanged for AWS credentials below
+      actions: read # lets Claude read CI results when asked why a check failed
+
+    steps:
+      - name: Generate App token
+        id: app-token
+        uses: actions/create-github-app-token@v3
```
aromko, line 104: **Suggestion** — the App token here is minted without `permission-contents` / `permission-pull-requests`, while the identical step in `claude-review.yml` narrows them and the comment there treats narrowing as the point. Mentions genuinely needs all three, so this grants nothing extra today. Stating them anyway pins the grant if the App later gains a permission for another consumer.

.github/workflows/claude-review.yml
sarahgm: Agreed, and the pattern you wrote is the fix. The escalation exists to promote "any narrowed profile touching real source", and this file's own argument is that `docs/**/*.tsx` is real source, so leaving it out contradicts both the shortened `paths-ignore` and the `docs` profile.

One correction to the evidence: #5853 is not in the hole. It touches 75 files under `packages/*/src/`, so it escalates today. The other three hold, all `ci:`/`chore:` titles with 2 to 3 docs `.tsx` files and no package source: #5742, #5739, #5738.

.github/workflows/claude-mentions.yml
sarahgm: Agreed, and the second half is the part that matters: a timeout cancels mid-run, which is the half-finished push `cancel-in-progress: false` gives up cancellation to avoid.

Raising the timeout rather than cutting turns. Sixty turns is the budget for "make the fix and prove it typechecks", and trimming it would make the job fail at the thing it exists for, while `pnpm install --frozen-lockfile` plus `pnpm typecheck:only` spends minutes before Claude has done anything.

.github/workflows/claude-review.yml
sarahgm: Confirmed. These two workflows are the only `@v6` uses in the repo: the other 20 are `@v7` and main is entirely `@v7` since #5877. The branch simply predates it. Bumping both.

.github/workflows/claude-mentions.yml
sarahgm: Same fix as `review.yml:127`, bumping both together.

.github/workflows/claude-review.yml
sarahgm: Fair. A prose-only PR never starts the workflow, so the `docs` profile only ever sees MDX riding along with a non-markdown file.

Putting that on the `paths-ignore` comment rather than changing the profile text, since the profile is right about what to do when MDX does turn up, and the comment is where someone looking for the missing prose review will actually land.

.github/workflows/claude-mentions.yml
sarahgm: Agreed. It grants nothing extra today and the point is pinning it against the App later gaining a permission for another consumer. Stating the four the job already declares: `permission-contents: write`, `permission-pull-requests: write`, `permission-issues: write` and `permission-actions: read`.

sarahgm:
On the mechanism, agreed, and it is worth fixing whatever happens with DST-1778. The prompt points at that path with no fallback, so anything that removes the file degrades the review to `CLAUDE.md` alone with nothing failing. Taking the guard step.

On the collision itself I cannot verify from here: `chore/DST-1778-extract-dst-skills-plugin` is not on origin, `a8c7ea9` returns 422, and no PR references DST-1778, so it reads as unpushed. The checklist is present at this PR's head.

Where does DST-1778 stand? If it is landing soon, moving the checklist to a plain in-repo file the plugin references beats the guard, because the guard only tells us after the fact. If it is further out, the guard covers this PR and the move can happen when DST-1778 is ready.

.github/workflows/claude-mentions.yml
sarahgm: Landed in 65ed3c501: 45 minutes, with the reasoning kept in a comment above the line so the next person does not trim it back.

Leaving this open, since you offered either direction and I picked one. If you would rather cut turns than raise the ceiling, say so and I will swap it.

.github/workflows/claude-review.yml
sarahgm: Landed in 65ed3c501, on the `paths-ignore` comment rather than in the profile text, since the profile is right about what to do when MDX does turn up and the filter is where someone hunting the missing prose review will land.

Leaving open in case you would rather it sat on the profile after all.

sarahgm:
Guard landed in 65ed3c501. It fails the review job with an explicit message, before the AWS credentials are assumed, so the degradation is loud instead of silent.

Still open, and only you can answer it: where DST-1778 stands, and therefore whether the guard is the answer or just the stopgap until the checklist moves to a plain in-repo file the plugin references.

aromko (approved):
No way to test it proper until it finally lands in the main branch 🤞

PR #5787, opened 2026-09-04
feat(DST-1531): add the /review-queue skill

sarahgm:
> Stacked on #5780 (`/pick-up`), which this branch builds on. Review that one first. GitHub will retarget this PR to `main` once it merges.

# Description

Adds `/review-queue`, a skill that answers two questions that currently need several browser tabs: what should I review next, and which of my PRs bounced back. It sweeps every repo in the `marigold-ui` org for open PRs, joins each one to its DST ticket, and prints one capped digest ranked by the Jira board rather than by `updatedAt`. GitHub can list PRs, but it cannot know that a PR attached to a sprint-committed ticket outranks a drive-by change, and that signal lives on the board.

The digest has three sections. **Awaiting your review**, meaning the PRs nobody has reviewed yet plus the ones you already have, ranked into five tiers with a `Ranked by` column so the order can be checked rather than trusted. **Your PRs needing rework**, covering changes requested, unresolved review threads, unresolved Vercel preview comments and failing builds. And a **board tail** listing tickets the board says are in review with no PR found, which is what catches the team's GitLab work that `gh` cannot see at all, and which on the first live run also caught DST-1529, whose PR had merged while the ticket sat in review.

Two things in the ticket did not survive contact with the data, and both are written down in the skill with the measurement behind them. `--review-requested=@me` returns nothing here, because the team assigns reviewers in conversation rather than in GitHub's field, so a queue built on it would be permanently empty and look like a clear board. The sweep is org-wide instead, and "is this mine to review" is decided further down from who has actually reviewed it. Separately, priority is a dead field on this board, with three open issues carrying one at all and none above Medium, so ranking is open-sprint membership plus board rank and makes no claim about priority.

The skill is read-only and deliberately has no confirmation gate, which is where it parts company with `/pick-up` and `/triage-feedback`. Every entry in `allowed-tools` is a read, and there is no `Write`, no `Edit`, no `git`, no `gh pr review`. The narrow list is the guarantee, not the prose.

Two supporting changes come with it. `.claude/skills/references/jira-board.md` is new, holding the JQL traps, field ids and board mechanics that `/pick-up` and `/review-queue` both hit, and `/pick-up`'s own copy is trimmed down to link it instead of duplicating it. `.claude/README.md` gains the convention that says where a shared reference belongs and why it is neither `CLAUDE.md` nor `.memory/`, plus a short section on gates covering the `skipAutoPermissionPrompt` behaviour that leads some skills to render their last gate as text rather than through `AskUserQuestion`.

N/A on both counts below: `.claude/` is not a published package, so no changeset, and nothing here touches UI.

Closes DST-1531

## Test Instructions

1. Read `.claude/skills/review-queue/SKILL.md` and the shared `.claude/skills/references/jira-board.md`.
2. Run `/review-queue`. All three sections should render as markdown tables, each heading line carrying its counts and the scope it gathered ("N repos searched, M archived skipped"). Nothing should be written to disk, and nothing posted, replied to, resolved or transitioned.
3. Check the order is the board's and not GitHub's. A PR on an active-sprint ticket should sit above one whose ticket has no sprint, and a PR where you requested changes and the author has not pushed since should be demoted with `author's turn` in `Ranked by`.
4. Run `/review-queue --review` and `/review-queue --rework`. Each should print only its own section and say plainly that the other was not gathered, so an ungathered section is distinguishable from an empty one. `--all` should lift the per-section cap of 7.
5. Spot-check the two live traps the skill was built around. `#5776`'s branch says `dst-1745` while its title says `DST-1754`, so its row should take the title and carry the warning flag. A red `Vercel Preview Comments` check should read as open feedback pointing at `/triage-feedback`, never as failing CI.

## Breaking Changes

No

## Checklist


OsamaAbdellateef (approved):

.claude/skills/pick-up/SKILL.md
```diff
@@ -36,13 +36,13 @@ project = DST AND sprint in openSprints()
 
 Pass `fields: ["summary", "status", "assignee"]` and `maxResults: 5`. An unfiltered sprint query returns well over a hundred thousand characters and will be truncated before you can read it.
 
-Two things about that query are the result of getting it wrong first:
+Three things about that query are the result of getting it wrong first, and all three are general enough that they now live in [../references/jira-board.md](../references/jira-board.md). What they mean here:
```
aromko, line 39: @marigold-ui/developer  open discussion. Wouldn't it be better to start review from github and picking the right ticket when available? Also the whole explanation in jira to pick the right tickets looks to me like it could get easily wrong when not keep it up to date. E.g. the different names for Bereit, ready and so on

aromko (commented):
Sry wrote it already last week but didn't submit...

.claude/skills/pick-up/SKILL.md
sarahgm: Good question.

`/review-queue` is already GitHub-first. It sweeps `gh search prs --owner=marigold-ui` across the org, pulls the ticket key out of the branch and the title, and only then asks Jira about the keys it found, purely to rank the rows by sprint and board rank. A PR without a key stays in the queue and renders unranked, so the board is enrichment, never the index. `reference-app` has live examples of that today.

The block you commented on is `/pick-up`, which answers a different question: "what should I work on next", where the work has no PR yet, so GitHub has nothing to start from. That one has to begin on the board.

aromko: Ah I think the confusion comes from the PR title - so I just thought is about the /review-queue skill.

aromko: or it comes from the stack thingy because the other one is already merged...

aromko (approved):

PR #5789, opened 2026-09-04
chore(DST-1598): mark internal-only barrel exports with @internal

aromko:
# Description

The Insights scanner treats every runtime export of the package barrels as an official component, so composition internals such as `CloseButton` and `HelpText` show up as "unused" and distort the adoption metrics. This PR tags those exports with a bare `/** @internal */` in the barrel source, which is what the scanner fetches from GitHub. The linked DST-1599 change teaches the classifier to read the tag.

An export is internal when it has no docs page of its own, is not documented as an API on a parent page, and exists to be composed by other Marigold packages. Usage count does not protect an export: `Label` is imported directly 14 times and is still internal, because surfacing that usage is the point.

- `@marigold/components` (19): `FieldBase`, `Label`, `Input`, `SearchInput`, `HelpText`, `Popover`, `Modal`, `NonModal`, `Underlay`, `Tray`, `CloseButton`, `ListBox`, `Collapsible`, `ActionMenuContext`, `ConfirmationProvider`, `ConfirmationContext`, `HeadingContext`, `TextContext`, `Provider`
- `@marigold/system` (25): `useClassNames`, `useStateProps`, the 12 style-prop maps, the 10 CSS-variable helpers, `get`
- `@marigold/icons`: nothing, every runtime export is an icon

Kept public despite missing docs because they were built for consumers: `IconButton`, `useDatePresets`, `useDateRangePresets`, `OverlayContainerProvider`. DST-1758 tracks their documentation.

Tagged exports sit one per statement because the classifier attributes a leading comment to every specifier in the statement. The rule is recorded in `CLAUDE.md` (Code Style) and the term "Internal export" in `.memory/CONTEXT.md`.

No changeset on purpose: tsdown collapses the barrel into a single export statement, so `dist/index.mjs` and `dist/index.d.mts` of both packages are byte-identical before and after this change (verified by hash). No UI change, so no screenshots or VRT.

Closes DST-1598

## Test Instructions

1. `pnpm typecheck:only` and `pnpm --filter @marigold/system --filter @marigold/components build` pass.
2. Compare `shasum packages/{components,system}/dist/index.{mjs,d.mts}` against a build from `main`: the hashes are identical.
3. In `packages/components/src/index.ts` and `packages/system/src/index.ts`, confirm every `/** @internal */` sits directly above a statement exporting exactly one name.
4. `pnpm check:memory` passes for the new glossary term.

## Breaking Changes

No

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -56,6 +56,7 @@ The CLI fetches from the Marigold docs site, caches for 24h, and works offline (
 - Use the `useClassNames` hook from `@marigold/system` for theming
 - Rename react-aria props: `isDisabled` → `disabled`, `isPending` → `loading`
 - Export components with named exports
+- **Barrel exports**: public only if the export has a docs page. Everything else in `packages/*/src/index.ts` gets a bare `/** @internal */` on the line above, one export per tagged statement. The Insights scanner reads the tag from the barrel source, so do not add it to component files (see `.memory/CONTEXT.md` → Internal export)
 - Use React Context for component composition (see `AccordionContext` patterns)
 
 ## Testing
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -56,7 +56,7 @@ The CLI fetches from the Marigold docs site, caches for 24h, and works offline (
 - Use the `useClassNames` hook from `@marigold/system` for theming
 - Rename react-aria props: `isDisabled` → `disabled`, `isPending` → `loading`
 - Export components with named exports
-- **Barrel exports**: public only if the export has a docs page. Everything else in `packages/*/src/index.ts` gets a bare `/** @internal */` on the line above, one export per tagged statement. The Insights scanner reads the tag from the barrel source, so do not add it to component files (see `.memory/CONTEXT.md` → Internal export)
+- **Barrel exports**: an export is internal when it exists for composition by other Marigold packages rather than for consumers. Documented exports (own page, or an API on a parent page) are public; consumer-facing exports that still lack docs stay public with a docs ticket (DST-1758). Internal exports get a bare `/** @internal */` on the line above, one export per tagged statement. The Insights scanner reads the tag from the barrel source, so do not add it to component files (see `.memory/CONTEXT.md` → Internal export)
 - Use React Context for component composition (see `AccordionContext` patterns)
 
 ## Testing
--- a/packages/components/src/index.ts
+++ b/packages/components/src/index.ts
@@ -79,8 +79,10 @@ export type { CheckboxGroupProps } from './Checkbox/CheckboxGroup';
 export { Checkbox } from './Checkbox/Checkbox';
 export type { CheckboxProps } from './Checkbox/Checkbox';
 
+/** @internal */
 export { CloseButton } from './CloseButton/CloseButton';
 
+/** @internal */
 export { Collapsible } from './Collapsible/Collapsible';
 export type { CollapsibleProps } from './Collapsible/Collapsible';
 
@@ -112,11 +114,11 @@ export { Dialog } from './Dialog/Dialog';
 export type { DialogProps } from './Dialog/Dialog';
 export { ConfirmationDialog } from './Dialog/ConfirmationDialog';
 export type { ConfirmationDialogProps } from './Dialog/ConfirmationDialog';
-export {
-  useConfirmation,
-  ConfirmationProvider,
-  ConfirmationContext,
-} from './Dialog/useConfirmation';
+export { useConfirmation } from './Dialog/useConfirmation';
+/** @internal */
+export { ConfirmationProvider } from './Dialog/useConfirmation';
+/** @internal */
+export { ConfirmationContext } from './Dialog/useConfirmation';
 export type {
   ConfirmationResult,
   ConfirmationConfig,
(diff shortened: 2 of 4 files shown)


CLAUDE.md
```diff
@@ -56,6 +56,7 @@ The CLI fetches from the Marigold docs site, caches for 24h, and works offline (
 - Use the `useClassNames` hook from `@marigold/system` for theming
 - Rename react-aria props: `isDisabled` → `disabled`, `isPending` → `loading`
 - Export components with named exports
+- **Barrel exports**: public only if the export has a docs page. Everything else in `packages/*/src/index.ts` gets a bare `/** @internal */` on the line above, one export per tagged statement. The Insights scanner reads the tag from the barrel source, so do not add it to component files (see `.memory/CONTEXT.md` → Internal export)
```
OsamaAbdellateef: ⚠️ **The written rule is stricter than the rule this PR applied.**

"public only if the export has a docs page" — applied literally, this mistags at least eight exports the PR deliberately left public:

- `IconButton`, `useDatePresets`, `useDateRangePresets`, `OverlayContainerProvider` — no docs page; the PR body keeps them public with DST-1758 tracking docs
- `ButtonContext`, `usePageFocus`, `I18nProvider`, `ThemeProvider` — no page of their own, documented as APIs on `button-group`, `page`, `calendar` and `useTheme`

The PR description states the real three-part test. The next person adding a barrel export will read this line, not the PR description.

```suggestion
- **Barrel exports**: an export is internal when all three hold: no docs page of its own, no documentation as an API on a parent page, and it exists to be composed by other Marigold packages. Usage count does not make an export public. Internal exports get a bare `/** @internal */` on the line above, one export per tagged statement. Exports built for consumers but not yet documented stay public (see DST-1758). The Insights scanner reads the tag from the barrel source, so do not add it to component files (see `.memory/CONTEXT.md` → Internal export)
```

packages/components/src/index.ts
```diff
@@ -112,11 +114,11 @@ export { Dialog } from './Dialog/Dialog';
...
 export { ConfirmationDialog } from './Dialog/ConfirmationDialog';
 export type { ConfirmationDialogProps } from './Dialog/ConfirmationDialog';
-export {
-  useConfirmation,
-  ConfirmationProvider,
-  ConfirmationContext,
-} from './Dialog/useConfirmation';
+export { useConfirmation } from './Dialog/useConfirmation';
+/** @internal */
+export { ConfirmationProvider } from './Dialog/useConfirmation';
```
OsamaAbdellateef, line 119: ⚠️ **The one-export-per-statement invariant has no CI guard, and violating it fails silently.**

This is exactly the shape at risk: a three-name statement split into three. If someone merges these back — or an organize-imports pass regroups the 12 style-prop maps in `system` — the classifier attributes the tag to every specifier and the only symptom is wrong numbers in an external dashboard nobody diffs.

The repo already guards comparable conventions in CI (`check:memory`, `check:rac-imports`, `check:theme-zindex`). A `check:internal-exports` asserting that every line after `/** @internal */` in `packages/*/src/index.ts` is a single-specifier runtime `export { X } from …` is ~20 lines and closes the loop.

I verified all 44 tags satisfy the invariant today — this is about keeping it true.

packages/system/src/index.ts
```diff
@@ -8,12 +8,14 @@ export type { NumericFormatProps } from './components/Formatters/NumericFormat';
 export type { NumerFormatterOptions } from './components/Formatters/NumericFormat';
 
 // Hooks
+/** @internal */
```
OsamaAbdellateef, line 11: 💡 **Nothing local explains these tags.** `stripInternal` isn't set in `@marigold/tsconfig/base` and tsdown drops comments, so the tag never reaches `dist/*.d.mts` — the accepted trade-off per `.memory/CONTEXT.md`. But someone opening this file now sees 25 unexplained `/** @internal */` lines. A one-line header near the top of each barrel would carry the context to where it's actually read:

`// /** @internal */ marks composition-only exports for the Insights scanner — see .memory/CONTEXT.md → Internal export`

CLAUDE.md
aromko: Confirmed. All eight are either an API on a parent page or kept public for DST-1758, and the one-liner contradicted the CONTEXT.md definition it links to. Reworded to the audience-based test: composition-only exports are internal, documented or consumer-facing exports are public.

packages/components/src/index.ts
aromko: Checked the exposure. The classifier skips type-only statements and specifiers, so `export { Label, type LabelProps }` cannot mistag. Of the 44 tags, exactly one module exports both tagged and untagged runtime names (`Dialog/useConfirmation`); everywhere else, merging tagged neighbours keeps the classification correct because all runtime names in that module are internal. No formatter or ESLint rule here merges export statements.

The classifier in DST-1599 is the only consumer of the invariant and is not built yet, so the check goes there: fail the scan when a tagged statement has more than one runtime specifier. Added to DST-1599 as a scanner task and acceptance criterion instead of a Marigold CI job.

packages/system/src/index.ts
aromko: Declining. `@internal` is a standard TSDoc tag and says what it means; the non-obvious parts (barrel placement, not reaching dist) are in CLAUDE.md and CONTEXT.md, one grep away from either barrel. The rule already drifted once between the PR body and CLAUDE.md, per your first comment. A header in each barrel would be a third and fourth copy.

OsamaAbdellateef (approved):

PR #5790, opened 2026-09-04
chore(DST-1525): add project pre-flight and post-edit typecheck hooks

sarahgm:
# Description

Two committed hooks in `.claude/settings.json`, so the whole team gets them without opting in.

**`SessionStart` → `.claude/hooks/preflight.mjs`** probes the repo state a session otherwise guesses at: the branch and the trunk it forked from, whether a prerelease channel is open, how stale the local origin refs are, and whether the installed toolchain matches what the repo pins. It probes instead of restating, because written-down environment facts drift. CLAUDE.md claimed "Node.js 22.x required" for months while `.node-version` said 24, and this PR replaces that line with one the hook derives.

**`PostToolUse (Edit|Write)` and `Stop` → `.claude/hooks/typecheck-changed.mjs`** runs the repo's own typecheck after an edit, so a type regression reaches the model while it still has the context to fix it, rather than surfacing in CI. The `Stop` registration is the backstop covering files written through Bash, which never fire `Edit` or `Write`. It is gated on whether the tree moved since the last answer, and blocks only on errors in files that moved, so a branch switch or a hand edit mid-session cannot hand the model a program full of errors it did not cause.

`tsconfig.check.json` also picks up `--noEmit` and `--resolveJsonModule`, which were previously spelled out in three places (`typecheck`, `typecheck:only`, and the hook) with nothing keeping them in sync. `tsc --showConfig` is identical before and after, over the same 2629-file program.

No changeset: nothing under `packages/` or `themes/` changes, and the root `package.json` is private.

Closes DST-1525

## Two deviations from the ticket, both deliberate

**1. It runs the full program, which the ticket excludes.** The ticket lists "a full per-edit monorepo `tsc`" under *Not included*, suggesting the changed package instead. Measuring said otherwise:

| | |
| --- | --- |
| Cold run | ~11s |
| Warm `--incremental` run | ~2.5s |
| End-of-turn gate when the tree has not moved | ~0.1s |

So there is little to gain by narrowing, and nothing obvious to narrow *to*: the only per-package configs are `tsconfig.build.json` emit configs that resolve `@marigold/*` through `node_modules` to `dist`, and `turbo.json` defines no typecheck task. The ticket's own fallback ("run the full check on the `Stop` event instead") is implemented as well, gated on whether anything actually changed. Happy to narrow if you would rather hold the original line.

**2. Scripts are `.mjs`, not `.sh`.** Node built-ins only, matching `scripts/check-*.mjs`. The hook payload arrives as JSON on stdin and `jq` is not guaranteed to be installed.

## Test Instructions

1. Check out the branch and start a session. The pre-flight block appears at the top of the context.
2. Without a session: `echo '{}' | .claude/hooks/preflight.mjs`
3. Use `Edit` or `Write` to introduce a type error in any `.ts`/`.tsx` file under `packages/`. The hook should block with the error and exit 2. Revert it.
4. Write the same error through Bash instead (`>>` or `sed`), which fires no `Edit` hook. The `Stop` hook should catch it at the end of the turn.
5. Confirm the opt-outs are silent: `MARIGOLD_SKIP_PREFLIGHT_HOOK=1` and `MARIGOLD_SKIP_TYPECHECK_HOOK=1`.
6. `pnpm typecheck` and `pnpm lint` should be unchanged.

## Breaking Changes

No, in the API sense. Nothing published changes.

Worth accepting consciously, though: **once this merges, both hooks run on every teammate's machine with no prompt.** Hooks in a settings file are not gated by workspace trust (that gates a project's `permissions.allow` rules), and the file watcher picks up a settings edit mid-session, so a pull can start one running before the next session. `.claude/README.md` gains the rules that stand in for the missing confirmation gate: a hook may write only to a gitignored cache, must reject uninteresting cases in its first few lines, must degrade to silence rather than noise, and must carry a named opt-out.

## Checklist


.claude/hooks/preflight.mjs
```diff
@@ -0,0 +1,203 @@
...
+ * are, and whether the installed toolchain matches what the repo pins.
+ *
+ * Why a hook and not CLAUDE.md: written-down environment facts drift. CLAUDE.md claimed
+ * "Node.js 22.x required" for months while .node-version said 24. This probes instead.
+ *
+ * Contract: SessionStart adds stdout to the model's context on exit 0, and ignores output
+ * entirely on any other exit. Every probe degrades to a missing line, and the process always
+ * exits 0.
+ *
+ * Scope and known limits:
```
aromko: Can we shorten or remove this wall of text?

tsconfig.check.json
```diff
@@ -4,6 +4,9 @@
 {
   "extends": "@marigold/tsconfig/react",
   "compilerOptions": {
+    // Here, not on the command line, so the scripts, CI and the post-edit hook cannot drift.
```
aromko: Remove this comments pls

.claude/hooks/typecheck-changed.mjs
```diff
@@ -0,0 +1,299 @@
+#!/usr/bin/env node
+/**
```
aromko, line 2: Also shorten and a general question? Is this running on each change claude made or wouldn't it be enough after at all. In my mind is just maybe it takes too long and eats more tokens?

.claude/hooks/typecheck-changed.mjs
```diff
@@ -0,0 +1,299 @@
...
+try {
+  result = spawnSync(
+    tscPath,
+    [
+      '--project',
+      'tsconfig.check.json',
+      '--incremental',
+      '--tsBuildInfoFile',
+      buildInfoPath,
+      '--pretty',
```
aromko, line 259: ⚠️ **Guard the tree against an emit**

Rule 1 in `.claude/README.md` says a hook must never dirty the working tree. This is the one place where that can go wrong: if `noEmit` ever leaves `tsconfig.check.json`, `pnpm typecheck` and CI keep passing, they just emit into a throwaway workspace, while this hook writes a `.js`/`.jsx`/`.map` next to every source file on every teammate's machine.

I hit exactly that while reviewing: running this script against a checkout whose `tsconfig.check.json` predates this PR emitted 1720 files. One flag closes it, and it costs nothing while the config also has it.

```suggestion
      '--noEmit',
      '--pretty',
```

.claude/hooks/typecheck-changed.mjs
```diff
@@ -0,0 +1,299 @@
...
+// tsc reads the program up front, so stamping the end would hide an edit that landed mid-run.
+const lastCheckedMs = (() => {
+  try {
+    return statSync(stampPath).mtimeMs;
+  } catch {
+    return null;
+  }
+})();
+
+if (isStop && lastCheckedMs !== null && !changedSince(lastCheckedMs)) process.exit(0);
```
aromko: ⚠️ **The Stop backstop blocks on errors the model did not cause**

On `Stop` the gate is "anything moved since the last check", so a branch switch or a hand edit by the developer mid-session trips it too. The next time Claude finishes a turn, exit 2 hands it every error in the program and tells it to keep going, and it spends a cycle fixing type errors nobody asked about before `stop_hook_active` lets it stop. That is also the token cost I asked about above, and it lands on turns that touched no code.

`changedSince` already walks the files that moved. If it returned them instead of a boolean, the Stop run could exit 2 only when at least one error is in a changed file, and list those first the way the PostToolUse run does with `editedFile`. The trade-off: a changed export in `A.ts` that only breaks the unchanged `B.ts` would pass the Stop gate silently. I would take that over the unrequested fix cycle, but it is a judgment call.

.claude/hooks/typecheck-changed.mjs
```diff
@@ -0,0 +1,299 @@
...
+  // resolve(), not join(): --git-dir is absolute inside a worktree.
+  const gitDir = git(['rev-parse', '--git-dir'])?.trim();
+  if (!gitDir) return true;
+  try {
+    if (statSync(resolve(root, gitDir, 'index')).mtimeMs > stampMs) return true;
+  } catch {
+    // No index to compare against. The file scan below still applies.
+  }
+
+  for (const rel of listed.split('\n')) {
```
aromko: 💡 **The change gate cannot see an untracked file disappear**

`git ls-files` stops listing an untracked file once it is deleted, and `mv` keeps the mtime, so neither a `rm` nor a rename of a file this session created counts as a change when done through Bash. Sequence: Write `Foo.ts` (checked), Write `Bar.ts` importing it (checked), `rm Foo.ts` in Bash. `Bar.ts` is now broken, nothing is newer than the stamp, and Stop says "unchanged". I reproduced the miss locally: after deleting a broken untracked file, the Stop run returned in 66ms without re-checking. Harmless in that direction, not in the one above.

Cheap fix: write the sorted listing into the stamp file and treat a different listing as a change. Tracked deletions are already covered by the stat failure below.

.claude/hooks/typecheck-changed.mjs
```diff
@@ -0,0 +1,299 @@
...
+    ? 'None of these are in the file you just edited.'
+    : null,
+  'Reproduce with `pnpm typecheck:only`.',
+].filter(Boolean);
+
+console.error([header, ...shown.map(l => `  ${l}`), ...footer].join('\n'));
+
+// PostToolUse cannot block, and exit 2 is the documented way to put stderr in front of the
+// model. Exit 2 on Stop prevents the turn from ending on broken types.
+process.exit(2);
```
aromko: 💡 `preflight.mjs` avoids `process.exit()` after writing to the pipe because pipe writes are async on macOS. Same hazard here, only masked by the 25-line cap fitting in the pipe buffer. Same line count, no exception:

```suggestion
process.exitCode = 2;
```

.claude/hooks/preflight.mjs
```diff
@@ -0,0 +1,203 @@
...
+import { execFileSync } from 'node:child_process';
+import { readFileSync, statSync } from 'node:fs';
+import { dirname, join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+if (process.env.MARIGOLD_SKIP_PREFLIGHT_HOOK === '1') process.exit(0);
+
+const root = resolve(dirname(fileURLToPath(import.meta.url)), '..', '..');
+/** Release trunks, most specific first, so the nearest fork point wins a tie. */
+const TRUNKS = ['beta-release', 'main'];
```
aromko: 💡 `origin` has no `beta-release` today: `git ls-remote --heads origin beta-release` is empty, there is no local ref, and no workflow mentions it. The hook tolerates the missing ref, but the header comment above describes that branch's `pre.json` state as a present-tense fact, and it is one more paragraph in the block I asked to shorten. Either keep the name because the next beta cycle recreates the branch, and say that in one line, or drop it, which makes this `['main']` and removes the sort.

.claude/settings.json
```diff
@@ -6,5 +6,41 @@
...
     "code-simplifier@claude-plugins-official": true,
     "atlassian@claude-plugins-official": true
+  },
+  "hooks": {
+    "SessionStart": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "\"$CLAUDE_PROJECT_DIR\"/.claude/hooks/preflight.mjs",
```
aromko: 💡 Prefixing with `node` removes the dependency on the executable bit surviving every clone (`core.fileMode=false`, a zip download, some Windows setups). Same for lines 28 and 39, and the "commit the executable bit" sentence in `.claude/README.md` can then go.

```suggestion
            "command": "node \"$CLAUDE_PROJECT_DIR\"/.claude/hooks/preflight.mjs",
```

.claude/hooks/typecheck-changed.mjs
sarahgm: Header is down to 15 lines. 727c615

On the question: it runs on every `Edit`/`Write` of a `.ts`/`.tsx` file that is in the program, and once more at the end of a turn. Measured on this branch:

| | |
| --- | --- |
| Cold run, no cache | ~11s |
| Warm `--incremental` run | ~2.4s |
| End-of-turn run when the tree has not moved | ~0.11s |

Two things make that cheaper than it looks. A passing check prints nothing and exits 0, so it costs **zero** tokens; tokens are only spent when the check fails. And a burst of edits does not cost one run each, because the second edit finds the lock held and skips, letting the `Stop` run cover it.

Where you are right is the failing case, and it is the same thing you found on line 193: on a tree that already had errors, every single edit re-sent the same 25 lines, and then `Stop` blocked on all of them. That is what was actually eating tokens, and it is fixed there rather than here.

So I have kept both registrations. The `PostToolUse` run is what makes the `Stop` run a 0.11s no-op instead of a 2.4s one, and it is now the only run that reports the whole program, which is what still catches an edit to `A.ts` that breaks the untouched `B.ts`. Happy to drop it and go end-of-turn-only if you would still rather, it is a two-line change to `settings.json`.

.claude/hooks/typecheck-changed.mjs
sarahgm: Agreed, and done the way you sketched it. 727c615

`changedSince` is now `moved`, and it returns the paths instead of a boolean. `Stop` exits 2 only when at least one error line resolves to one of them, and those errors are listed first, the same way `editedFile` already worked. Everything else exits 0 silently.

I took your trade-off rather than trying to engineer around it, but it is narrower than it first looked: the `PostToolUse` run still reports the whole program, so a model edit to `A.ts` that breaks the unchanged `B.ts` is still caught immediately. The blind spot is only for edits that reach `Stop` without an `Edit`/`Write`, i.e. Bash-written ones. That is written down at the gate rather than left implicit.

One interaction worth flagging, because it bit me: this fix on its own swallowed the deletion case from your next comment. A file that vanished breaks its importers, and those did not move, so the attribution filter reported nothing. A disappeared path now widens attribution back to "everything" instead of just flipping the gate.

Checked both directions:

- pre-existing error in an untouched file, unrelated file touched → exit 0, silent
- the file carrying the error touched → exit 2

.claude/hooks/typecheck-changed.mjs
sarahgm: Good catch, and the fix is the one you suggested. 727c615

The stamp file now carries the sorted listing as its content alongside the timestamp as its mtime, and a listing that differs counts as a change.

The part I did not expect: flipping the gate was not enough on its own. The attribution filter from the line-193 thread then dropped the report anyway, because what a deletion breaks is the importers, and those did not move. So a path present in the stored listing and absent now returns "not attributable" rather than a list, which reports everything. It is the honest answer for a deletion.

Your exact repro, replayed against the new version:

```
=== D1. write Foo, check ===        exit=0
=== D2. write Bar importing Foo ===  exit=0
=== D3. rm Foo through Bash, Stop ===
Typecheck failed (1 error(s)):
  packages/components/src/__hookBar.ts(1,21): error TS2307: Cannot find module './__hookFoo' ...
exit=2
```

It also picks up untracked renames for free, which the mtime scan never could.

.claude/hooks/preflight.mjs
sarahgm: Dropped, so `TRUNKS` is now a single `TRUNK = 'main'` constant and the sort is gone. e6fd628

Your reading is right and it is worse than a missing ref: `git ls-remote --heads origin` returns 85 refs and none of them is `beta-release`, but the local `origin/beta-release` I still had points at 2026-08-03 and is **not** an ancestor of `origin/main`. So on a machine that still carries that leftover, the fork-point sort could pick a base that no longer exists on the remote, and report it with a straight face.

What settled the "keep it for the next beta cycle" option is that there will not be one on a branch. The repo README already says prereleases run in changesets pre mode on `main`, with no separate release branch.

That also took out the `VRT is not required for PRs into <trunk>` line, since it only ever fired for a trunk that was not `main`, and pre mode now runs on `main` where VRT *is* required.

.claude/hooks/preflight.mjs
sarahgm: Cut to 11 lines. The two limits that had to survive moved down to the probes they constrain: "never fetches" now sits on the branch probe, "checked-out branch only" on the `pre.json` read, where they can be read next to the code they are about. The rest was PR-body material. e6fd628

tsconfig.check.json
sarahgm: Gone. 2a62071

.claude/hooks/typecheck-changed.mjs
sarahgm: Taken, thanks for chasing it down. 727c615

Confirmed from the other side too: deleted the `noEmit` line from `tsconfig.check.json`, ran the hook, nothing emitted.

.claude/hooks/typecheck-changed.mjs
sarahgm: Taken. 727c615

.claude/settings.json
sarahgm: Taken, all three. e042b6d

The "commit the executable bit" sentence in `.claude/README.md` went with it.

aromko (approved):

PR #5798, opened 2026-09-07
fix(DSTSUP-277): give the DatePicker calendar button a symmetric hit area

OsamaAbdellateef:
# Description

The DatePicker calendar trigger was `h-control pr-3` — **28×36px with all of its padding on the right**, so the icon sat flush against the button's left edge. Every pixel to the left of the 16px glyph belonged to the date input, not to the button. Approaching from above, below or the right worked; approaching from the left meant landing on the glyph itself. A customer reported the picker "hanging" for exactly this reason — they came at it horizontally from the left, missed by a pixel or two, and nothing happened.

Measured before touching anything: `elementFromPoint` 2px left of the icon returned the date input. For comparison, ComboBox's chevron (which goes through `Input`'s action slot) is **36×36 with the icon centred**, 10px of slack on every side — so the date pickers were the odd ones out.

The trigger is now that same control-sized square: 36×36, icon centred, and at the same inset as a ComboBox chevron. Three things follow from enlarging the box:

- **The extra room has to be visible.** An invisible hit area still reads as "aim at the icon", so the whole box now takes a hover wash (`ui-state-hover-ghost`) alongside the existing icon darkening — the same affordance a ghost `Button` uses. Suppressed while disabled.
- **The trigger stays marked while the overlay is open.** The popover takes focus as soon as it opens, so the trigger keeps that surface on `aria-expanded`. Note the RAC plugin's `expanded:` variant does *not* work here — RAC sets `aria-expanded`, not `data-expanded`.
- **Focus is now visible on the button itself.** The field's `has-focus` ring fires for the date segments too, so it could never show *which* part held focus — tabbing to the calendar button looked identical to tabbing into the date. The button now draws its own inset ring (`ui-state-focus-item`; inset because the field is `overflow-hidden` and would clip an outline).

`ui-touch-hitbox` is dropped from the trigger — it only guaranteed a 24×24 minimum, which a 36×36 button clears on its own.

`DateRangePicker` re-exports these styles, so it gets the identical fix.

The date input also drops its right padding when an action is present (`DateField` and the `DateRangePicker` end input). The action box already centres its own icon, so the two insets would otherwise stack and push the icon 22px from the text it belongs to — 12px further than any other field. The gap is now 10px, matching ComboBox exactly, and the fit-width `DateRangePicker` field ends up 4px *narrower* than before rather than 8px wider.

Closes DSTSUP-277

## Screenshots / Preview

| Before | After |
| ------ | ----- |
|        |       |

> **Not filled in by the automation** — I verified this with DOM geometry and computed styles rather than screenshots (the browser pane wasn't compositing frames in this session). Worth grabbing shots of the hover, focus and expanded states from `Components/DatePicker/Basic` before review.

**Expect Chromatic diffs:** the icon shifts 2px left in both DatePicker and DateRangePicker, and the DateRangePicker field is 4px narrower. Nothing else moves.

## Test Instructions

1. `pnpm sb`, open `Components/DatePicker` → `Basic`.
2. Move the pointer at the calendar icon **horizontally from the left**. The button should light up (background wash + icon darkens) well before the pointer reaches the glyph, and clicking anywhere in that area opens the calendar. On `main` the area left of the glyph does nothing.
3. Tab from the year segment to the calendar button. The button should draw its own ring inside the field's ring — on `main` the two states are indistinguishable.
4. Open the calendar. The trigger should stay visibly marked while the popover is open.
5. Repeat 2–4 on `Components/DateRangePicker` → `Basic`.
6. Check the icon still sits at the same inset as the chevron in `Components/ComboBox` → `Basic`, and that the gap between the date text and the icon matches.
7. Automated: `pnpm test:sb` (DatePicker/DateRangePicker) and `pnpm test:unit`.

## Breaking Changes

No — no exports, props or theme keys removed or renamed. Purely a style/layout change plus an internal padding tweak.

## Checklist

### Notes on the unticked items

- **Unit tests** — no `*.test.tsx` changed. The new tests are `.test()` story tests (real browser, real geometry), which is the only place `getBoundingClientRect` and `elementFromPoint` mean anything. `Basic` gained the `component-test` tag on DatePicker so the runner picks it up. **Both new tests were confirmed to fail against the old styles** (`expected 0 to be greater than 0` — precisely the bug).
- **Accessibility** — left for a human to sign off, but for the record: the hit area now clears the WCAG 2.1 AA 24×24 target on its own, and the missing focus indicator on the trigger is fixed. Keyboard focus order and the `aria-expanded` / `aria-haspopup` contract are unchanged. No full APG audit was performed.
- **VRT** — Chromatic will need a review pass; see the expected diffs noted above.

### Spotted in passing, left alone as out of scope

- `disabled:cursor-not-allowed` has been on this button for a while, but I could not get the trigger into a disabled state to confirm it ever fires. The new hover guard is written as `not-disabled:` so it is correct either way, but the disabled path is unverified.
- `DateField.styles.ts` has an `action` slot (including `group-invalid/field:fill-destructive-accent`) that nothing reads — the calendar icon stays secondary-coloured in the error state. Dead code, and possibly a small bug of its own.

### DIFF
--- a/.changeset/dstsup-277_calendar-button-hit-area.md
+++ b/.changeset/dstsup-277_calendar-button-hit-area.md
@@ -1,0 +1,24 @@
+---
+'@marigold/components': patch
+'@marigold/theme-rui': patch
+---
+
+fix(DSTSUP-277): give the DatePicker calendar button a symmetric hit area
+
+The calendar trigger was `h-control pr-3` — 28×36px with all of its padding on the right. The icon sat **flush against the button's left edge**, so every pixel to the left of the 16px glyph belonged to the date input, not to the button. Approaching from above, below or the right worked; approaching from the left meant landing on the glyph itself. A customer reported the picker "hanging" for exactly this reason: they came at it horizontally from the left, missed by a pixel or two, and nothing happened.
+
+The trigger is now the same control-sized square that `Input` already hands to `ComboBox` and `SearchField` — 36×36 with the icon centred, so there is 10px of slack on **every** side and the icon lands at the same inset as a ComboBox chevron.
+
+Three things follow from making the box bigger:
+
+- **The extra room has to be visible.** An invisible hit area is still "aim at the icon", so the whole box now takes a hover wash (`ui-state-hover-ghost`) alongside the existing icon darkening, the same affordance a ghost `Button` uses. It is suppressed while the trigger is disabled.
+- **The trigger stays marked while the overlay is open.** The popover takes focus as soon as it opens, so the trigger keeps the hover surface on `aria-expanded`.
+- **Focus is now visible on the button itself.** The field's `has-focus` ring fires for the date segments too, so it could never show *which* part held focus — tabbing to the calendar button looked identical to tabbing into the date. The button now draws its own inset ring (`ui-state-focus-item`, inset because the field is `overflow-hidden` and would clip an outline).
+
+`ui-touch-hitbox` is gone from the trigger: it only guaranteed a 24×24 minimum, which a 36×36 button clears on its own.
+
+`DateRangePicker` re-exports these styles, so it gets the identical fix.
+
+The date input drops its right padding when an action is present (`DateField` and the `DateRangePicker` end input). The action box already centres its own icon, so the two insets would otherwise stack and push the icon 22px away from the text it belongs to — 12px further than every other field. With this, the text-to-icon gap is 10px, matching `ComboBox` exactly, and the fit-width `DateRangePicker` field is 4px narrower than before rather than 8px wider.
+
+Visually the icon shifts 2px left in both components; nothing else moves.
--- a/packages/components/src/DateField/DateInput.tsx
+++ b/packages/components/src/DateField/DateInput.tsx
@@ -120,9 +120,6 @@ const _DateInput = ({
         className={cn(
           'flex flex-1 items-center',
           classNames.input,
-          // The action is a control-sized box that already centers its icon,
-          // so the input's own right padding would stack on top of that inset
-          // and push the icon away from the text it belongs to.
           action && 'pr-0'
         )}
         {...props}
--- a/packages/components/src/DateField/DateInput.tsx
+++ b/packages/components/src/DateField/DateInput.tsx
@@ -117,7 +117,14 @@ const _DateInput = ({
       onPaste={handlePaste}
     >
       <DateInput
-        className={cn('flex flex-1 items-center', classNames.input)}
+        className={cn(
+          'flex flex-1 items-center',
+          classNames.input,
+          // The action is a control-sized box that already centers its icon,
+          // so the input's own right padding would stack on top of that inset
+          // and push the icon away from the text it belongs to.
+          action && 'pr-0'
+        )}
         {...props}
       >
         {segment => (
--- a/packages/components/src/DatePicker/DatePicker.stories.tsx
+++ b/packages/components/src/DatePicker/DatePicker.stories.tsx
@@ -123,6 +123,7 @@ const meta = preview.meta({
 });
 
 export const Basic = meta.story({
+  tags: ['component-test'],
   render: args => {
     return (
       <I18nProvider locale="de-DE">
@@ -136,6 +137,31 @@ export const Basic = meta.story({
     );
   },
 });
+
+Basic.test(
+  'gives the calendar trigger slack on every side of the icon',
+  { parameters: { chromatic: { disableSnapshot: true } } },
+  async ({ canvas }) => {
+    const trigger = canvas.getByRole('button');
+    const icon = trigger.querySelector('svg')!;
+
+    const button = trigger.getBoundingClientRect();
+    const glyph = icon.getBoundingClientRect();
+
+    const left = Math.round(glyph.left - button.left);
+    const right = Math.round(button.right - glyph.right);
+
+    // The icon used to sit flush against the button's left edge, so reaching
+    // the trigger from the left meant landing on the 16px glyph itself.
+    expect(left).toBeGreaterThan(0);
+    expect(left).toBe(right);
+
+    // The slack is part of the trigger, not of the date input next to it.
+    expect(
+      document.elementFromPoint(glyph.left - 4, button.top + button.height / 2)
+    ).toBe(trigger);
+  }
+);
 
 export const Controlled = meta.story({
   parameters: { chromatic: { disableSnapshot: true } },
--- a/packages/components/src/DateRangePicker/DateRangePicker.stories.tsx
+++ b/packages/components/src/DateRangePicker/DateRangePicker.stories.tsx
@@ -231,6 +231,31 @@ export const Mobile: any = meta.story({
 Basic.tags = ['component-test'];
 
 Basic.test(
+  'gives the calendar trigger slack on every side of the icon',
+  { parameters: { chromatic: { disableSnapshot: true } } },
+  async ({ canvas }: any) => {
+    const trigger = canvas.getByRole('button');
+    const icon = trigger.querySelector('svg')!;
+
+    const button = trigger.getBoundingClientRect();
+    const glyph = icon.getBoundingClientRect();
+
+    const left = Math.round(glyph.left - button.left);
+    const right = Math.round(button.right - glyph.right);
+
+    // The icon used to sit flush against the button's left edge, so reaching
+    // the trigger from the left meant landing on the 16px glyph itself.
+    expect(left).toBeGreaterThan(0);
+    expect(left).toBe(right);
+
+    // The slack is part of the trigger, not of the end date input next to it.
+    expect(
+      document.elementFromPoint(glyph.left - 4, button.top + button.height / 2)
+    ).toBe(trigger);
+  }
+);
+
+Basic.test(
   'opens the calendar popover and selects a range',
   async ({ canvas, step, userEvent }: any) => {
     const trigger = canvas.getByRole('button');
--- a/packages/components/src/DateRangePicker/DateRangePicker.tsx
+++ b/packages/components/src/DateRangePicker/DateRangePicker.tsx
@@ -256,7 +256,13 @@ const DateRangeInput = ({ action }: DateRangeInputProps) => {
         <div className="min-w-0" onPaste={handlePaste('end')}>
           <AriaDateInput
             slot="end"
-            className={cn('flex items-center pl-0', classNames.input)}
+            className={cn(
+              'flex items-center pl-0',
+              classNames.input,
+              // See `DateInput`: the action centers its own icon, so the end
+              // input must not add padding on that side as well.
+              action && 'pr-0'
+            )}
           >
             {segment => (
               <DateSegment className={classNames.segment} segment={segment} />
--- a/themes/theme-rui/src/components/DatePicker.styles.ts
+++ b/themes/theme-rui/src/components/DatePicker.styles.ts
@@ -2,34 +2,10 @@ import { type ThemeComponent, cva } from '@marigold/system';
 
 export const DatePicker: ThemeComponent<'DatePicker'> = cva({
   base: [
-    /**
-     * A control-sized square, the same trailing action box `Input` hands to
-     * ComboBox and SearchField. The icon used to sit flush against the
-     * button's left edge (`h-control pr-3`), so approaching the trigger from
-     * the left meant landing on the 16px icon itself -- everything to its
-     * left belonged to the date input. Centering it in a 36x36 box leaves
-     * equal slack on every side and puts the icon at the same inset as a
-     * ComboBox chevron.
-     */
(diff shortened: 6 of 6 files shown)


.changeset/dstsup-277_calendar-button-hit-area.md
```diff
@@ -0,0 +1,24 @@
+---
+'@marigold/components': patch
+'@marigold/theme-rui': patch
+---
+
+fix(DSTSUP-277): give the DatePicker calendar button a symmetric hit area
+
+The calendar trigger was `h-control pr-3` — 28×36px with all of its padding on the right. The icon sat **flush against the button's left edge**, so every pixel to the left of the 16px glyph belonged to the date input, not to the button. Approaching from above, below or the right worked; approaching from the left meant landing on the glyph itself. A customer reported the picker "hanging" for exactly this reason: they came at it horizontally from the left, missed by a pixel or two, and nothing happened.
```
aromko, line 8: 🔴 **Prose lint will fail on `main`**

`node scripts/vale.mjs` on this file returns 6 errors:

    8:43    error  Rephrase with a comma, a colon, or a second sentence instead of an em dash.  Marigold.EmDash
    8:287   error  Use a period or a comma instead of a semicolon in prose.                     Marigold.Semicolon
    10:109  error  Rephrase with a comma, a colon, or a second sentence instead of an em dash.  Marigold.EmDash
    16:159  error  Rephrase with a comma, a colon, or a second sentence instead of an em dash.  Marigold.EmDash
    22:258  error  Rephrase with a comma, a colon, or a second sentence instead of an em dash.  Marigold.EmDash
    24:53   error  Use a period or a comma instead of a semicolon in prose.                     Marigold.Semicolon

`.github/workflows/prose.yml` lints `.changeset/*.md` with `Marigold.EmDash` and `Marigold.Semicolon` at `level: error`. It landed on `main` in `39072e8df`; this branch is 16 commits behind, so the check never ran here. Once the file is on `main` it does run, and every other changeset in `.changeset/` is currently clean.

Rephrase the em dashes with a comma, a colon or a second sentence, and replace the two semicolons with a period. Verify with `node scripts/vale.mjs .changeset`.

packages/components/src/DateField/DateInput.tsx
```diff
@@ -117,7 +117,11 @@ const _DateInput = ({
       onPaste={handlePaste}
     >
       <DateInput
-        className={cn('flex flex-1 items-center', classNames.input)}
+        className={cn(
```
aromko, line 120: 🔴 **Visual Regression Tests - not run**

This PR changes UI-affecting files but the `Visual-Regression-Tests` workflow has not run for this branch: `gh run list --workflow=visual-regression-tests.yml --branch=dstsup-277_fix-calendar-button-padding` returns nothing, and there is no `UI Tests` check-run or Chromatic status on `97755a4`.

The description already predicts the diffs (icon shifts 2px left in both pickers, DateRangePicker field 4px narrower) and the screenshot table is empty, so Chromatic is the only record of what actually changed.

packages/components/src/DateField/DateInput.tsx
```diff
@@ -117,7 +117,11 @@ const _DateInput = ({
...
+          'flex flex-1 items-center',
+          classNames.input,
+          action && 'pr-0'
```
aromko, line 123: ⚠️ `action` is a public prop on `DateField` (`DateField.tsx:34`), so this drops the input's right padding for consumer-supplied actions too, not just for the calendar `IconButton`. `DateField.test.tsx:257` already renders `action={<div>huhu</div>}`; that element now sits flush against the date text with no gap.

The premise ("the action box already centres its own icon") only holds for Marigold's own 36x36 box. `Input` solves this by wrapping the action itself (`Input.tsx:66-74` plus the `action` slot in `Input.styles.ts`), which is also why its `pr-control` reservation is always safe. `DateField.styles.ts` already defines an `action` slot for the same purpose, but nothing consumes it. Wrapping the action here would make this unconditional and correct for any child.

packages/components/src/DateRangePicker/DateRangePicker.tsx
```diff
@@ -256,7 +256,13 @@ const DateRangeInput = ({ action }: DateRangeInputProps) => {
         <div className="min-w-0" onPaste={handlePaste('end')}>
           <AriaDateInput
             slot="end"
-            className={cn('flex items-center pl-0', classNames.input)}
+            className={cn(
+              'flex items-center pl-0',
+              classNames.input,
+              // See `DateInput`: the action centers its own icon, so the end
```
aromko, line 262: 💡 Two small things:

- The pointer dead-ends: `DateInput` carries no comment explaining the rule, so "See `DateInput`" sends the reader somewhere that does not have the answer. Either put the why once at the origin, or drop the pointer.
- `action &&` can never be false here. `DateRangeInput` is module-local and its single call site (`DateRangePicker.tsx:169`) always passes `action`, so this is `'pr-0'` with extra steps. Making the prop required and dropping the condition would be honest about that.

packages/components/src/DatePicker/DatePicker.stories.tsx
```diff
@@ -137,6 +138,31 @@ export const Basic = meta.story({
...
+    const trigger = canvas.getByRole('button');
+    const icon = trigger.querySelector('svg')!;
+
+    const button = trigger.getBoundingClientRect();
+    const glyph = icon.getBoundingClientRect();
+
+    const left = Math.round(glyph.left - button.left);
+    const right = Math.round(button.right - glyph.right);
+
+    // The icon used to sit flush against the button's left edge, so reaching
```
aromko, line 154: 💡 This comment restates the test name ("gives the calendar trigger slack on every side of the icon") and is duplicated verbatim in `DateRangePicker.stories.tsx:246`. The changeset already tells the story in full.

The `elementFromPoint` comment below it is worth keeping: that one explains a non-obvious assertion rather than the history.

packages/components/src/DateRangePicker/DateRangePicker.stories.tsx
```diff
@@ -230,6 +230,31 @@ export const Mobile: any = meta.story({
 
 Basic.tags = ['component-test'];
 
+Basic.test(
```
aromko, line 233: 💡 This test is byte-for-byte identical to the one in `DatePicker.stories.tsx:141` apart from one word in a comment. A shared assertion helper would follow the precedent `../firePaste` already sets in this folder, and would keep the two from drifting apart.

packages/components/src/DateRangePicker/DateRangePicker.stories.tsx
```diff
@@ -230,6 +230,31 @@ export const Mobile: any = meta.story({
...
+  'gives the calendar trigger slack on every side of the icon',
+  { parameters: { chromatic: { disableSnapshot: true } } },
+  async ({ canvas }: any) => {
```
aromko, line 236: 💡 The identical test in `DatePicker.stories.tsx:144` destructures `{ canvas }` with no annotation and typechecks fine, so the `: any` is not needed. It matches the rest of this file, which is already full of them, but CLAUDE.md asks for no new ones.

```suggestion
  async ({ canvas }) => {
```

themes/theme-rui/src/components/DatePicker.styles.ts
```diff
@@ -2,10 +2,11 @@ import { type ThemeComponent, cva } from '@marigold/system';
 
 export const DatePicker: ThemeComponent<'DatePicker'> = cva({
   base: [
-    'h-control pr-3',
+    'flex size-control items-center justify-center rounded-surface',
```
aromko, line 5: ⚠️ This re-creates the control-sized centered action box that already exists as `Input.styles.ts`'s `action` slot (`right-0 flex size-control items-center justify-center`, plus the `*:size-full *:items-center *:justify-center` wrapper in `Input.tsx:66-74`). `DateField.styles.ts` even defines its own `action` slot for this, but `classNames.action` is never read anywhere in the DateField tree, which is presumably why the box had to be rebuilt here.

Related: the hover wash, `aria-expanded` surface and focus ring on lines 7-9 are a real improvement, but they now apply only to the date pickers. The ComboBox chevron and SearchField clear button, which the description cites as the consistency target, still have neither a hover affordance nor their own focus ring. If the reasoning holds here (an invisible hit area still reads as "aim at the icon"), it holds there too. Worth lifting into the shared action box, or getting a design sign-off on the divergence.

.changeset/dstsup-277_calendar-button-hit-area.md
OsamaAbdellateef: Fixed in 38e98147f: rephrased the four em dashes with a colon or period, and replaced the two semicolons with periods. Verified there are no remaining em dashes or semicolons in the file.

packages/components/src/DateField/DateInput.tsx
OsamaAbdellateef: Not run yet — I didn't trigger it automatically since posting on the PR is a side-effecting action. Comment `/run-chromatic` when you'd like the Chromatic diff for review, or let me know and I'll post it.

packages/components/src/DateField/DateInput.tsx
OsamaAbdellateef: Fixed in 38e98147f. `DateField.styles.ts`'s `action` slot now carries the layout (`flex size-control items-center justify-center`) alongside its existing fill/color classes, and `DateInput.tsx` wraps `action` in a span using that slot with `*:flex *:size-full *:items-center *:justify-center`, mirroring `Input.tsx`'s pattern for its own `action` slot. Any child — the built-in `IconButton` or an arbitrary consumer element like the test's `<div>huhu</div>` — now gets centered and sized to the control box regardless of what it renders itself, so `pr-0` is correct unconditionally. Confirmed `DateField.test.tsx`'s `action={<div>huhu</div>}` test still passes.

packages/components/src/DateRangePicker/DateRangePicker.tsx
OsamaAbdellateef: Fixed in 38e98147f. Made `DateRangeInputProps.action` required (dropping the `action &&` check, since it's module-local and always passed) and dropped the "See `DateInput`" pointer instead of relocating the why — the input side of this had already collapsed to `pr-0 pl-0` unconditionally, so there was nothing left to explain there.

packages/components/src/DatePicker/DatePicker.stories.tsx
OsamaAbdellateef: Fixed in 38e98147f. Extracted the assertion into a shared `assertCalendarTriggerHitArea` helper in `packages/components/src/calendarTrigger.utils.ts` (same pattern as the existing `firePaste.ts`), used by both this test and the identical one in `DateRangePicker.stories.tsx`. The redundant comment is gone along with the inline body; the `elementFromPoint` check now lives once, in the helper.

packages/components/src/DateRangePicker/DateRangePicker.stories.tsx
OsamaAbdellateef: Fixed in 38e98147f. This test now calls the shared `assertCalendarTriggerHitArea` helper (`packages/components/src/calendarTrigger.utils.ts`) instead of duplicating the assertion body, so DatePicker and DateRangePicker can no longer drift apart on it.

packages/components/src/DateRangePicker/DateRangePicker.stories.tsx
OsamaAbdellateef: Fixed in 38e98147f, though not with the suggested one-liner: `Basic` in this file is exported as `Basic: any`, so `Basic.test`'s callback loses inference entirely and a bare `async ({ canvas }) =>` fails to typecheck here (confirmed with `tsc`) even though it works in `DatePicker.stories.tsx` where `Basic` isn't widened. Typed the param explicitly as `{ canvas: BoundFunctions<typeof queries> }` from `@testing-library/dom` instead, which avoids `any` without breaking the build.

themes/theme-rui/src/components/DatePicker.styles.ts
OsamaAbdellateef: Fixed in 38e98147f: the box's own sizing is gone from this file. `size-control` was removed from `DatePicker.styles.ts` (now just `flex items-center justify-center rounded-surface` plus the interactive states), and the trigger's sizing now comes solely from `DateField.styles.ts`'s `action` slot via the wrapping span, consumed by both `DateInput.tsx` (DatePicker) and `DateRangeInput` in `DateRangePicker.tsx` (which I also updated to wrap its `action` the same way, since it was rendering the button unwrapped). `classNames.action` is no longer dead code, and there's a single source of truth for the box now instead of two.

Left the hover/focus-ring parity with ComboBox's chevron and SearchField's clear button alone, since you framed that as needing a separate lift-or-sign-off decision rather than a required fix here, and it reaches into components this PR doesn't otherwise touch.

aromko (approved):

PR #5833, opened 2026-09-15
fix(DST-1769): give CloseButton an accessible name

sebald:
# Description

`<Dialog closeButton>` rendered its close button with no accessible name. `CloseButton` renders an icon only, `Dialog` passed no `aria-label`, and the `X` glyph contributes nothing to the name computation, so the computed name was the empty string. VoiceOver announced a bare "button". This is a WCAG 2.1 SC 4.1.2 (Name, Role, Value) failure for every product embedding a Marigold dialog. Not a total loss of function, since React Aria's `Modal` still renders its own visually hidden dismiss button and Escape still closes the dialog, but the visible affordance was the unlabelled one.

`Dialog` now passes the localized `close` label, matching the five other call sites (`Toast`, `SectionMessage`, `Drawer`, `SidebarModal`, `FileFieldItem`). `CloseButton` gained a fallback to that same string, because it is public API and consumer code cannot be linted from here.

The fallback stands down when a `slot` is set. That guard is load-bearing: React Aria's `TagGroup` names its `remove` slot through context, and a locally set `aria-label` wins over a context one, so an unconditional fallback relabelled every removable tag's button from "Remove News" to "Close". A test in `TagGroup.test.tsx` pins that.

The fallback also warns in development, following the pattern `Dialog` already uses when it renders without an accessible name. A silent fallback alone would trade a loud bug for a quiet one: `CloseButton` is labelled "Close navigation", "Dismiss drawer" and "Remove file" at its existing call sites, so a generic "Close" is often the wrong word. An unnamed button fails an automated a11y check. A confidently mislabelled one passes every check and misinforms the user.

`Dialog`'s close button test now queries by accessible name instead of DOM position. The positional lookup is what let this ship.

No visual change, since an `aria-label` is invisible. The Screenshots section is omitted and VRT is left unticked for that reason.

Closes DST-1769

## Test Instructions

1. Run the affected unit tests: `pnpm test:unit packages/components/src/CloseButton packages/components/src/Dialog packages/components/src/TagGroup`
2. In Storybook, open `Components/Dialog/Basic`, open the dialog, and inspect the X button. It should carry `aria-label="Close"`.
3. With VoiceOver on, tab to the X and confirm it announces "Close, button" rather than "button".
4. Open `Components/TagGroup/RemovableTags` and confirm each tag's remove button still announces "Remove <tag>", not "Close".
5. Render a bare `<CloseButton />` and confirm the dev console warns and the button still gets a name.

## Breaking Changes

No. `CloseButtonProps` gains an optional `aria-labelledby`, which is additive.

## Checklist


sebald (commented):
## Automated Code Review

Reviewed against DST-1769 and the Marigold conventions in `CLAUDE.md`. Claims about React Aria behaviour below were verified against `react-aria-components@1.20.0` in `node_modules`, not from memory.

### Overview

| Field | Value |
|-------|-------|
| Author | @sebald |
| Branch | `fix/DST-1769-closebutton-accessible-name` -> `main` |
| Files Changed | 6 (+119 / -7) |
| CI | Mostly pending. Typecheck, Format, Size Limit, Prose and the theme/memory guards pass |
| Visual Regression | Not started (no runs on this branch) |

### Linked Ticket

**DST-1769** - Dialog's close button has no accessible name. Status: In Review.

The implementation matches the ticket scope item for item: the `Dialog` label, the slot-conditional fallback, the dev warning, the rewritten positional query, the four test cases, the changeset. Nothing from the ticket's "Not included" list leaked in.

### Checklist

| Area | Result |
|---|---|
| TypeScript (no `any`, `import type`) | Pass |
| Component patterns (`useClassNames`, no new `className`/`style` surface) | Pass |
| React (functional, unconditional hooks, early returns) | Pass |
| React performance | Not applicable. No memoization, data fetching or bundle surface is touched, and `@react-aria/i18n` is already a declared dependency of `@marigold/components` |
| Testing (Vitest, `userEvent`, accessible queries, no `getByTestId`) | Pass, except the fixture/story point below |
| Accessibility | See Critical 1 |

`Dialog.test.tsx` earns its `eslint-disable testing-library/no-node-access` removal: no node-access patterns remain in the file.

### Issues Found

#### Critical

**1. The `slot` guard reopens DST-1769 for `slot="close"`.** Details inline on `CloseButton.tsx`.

**2. Visual regression tests have not run on this branch.** The PR touches `packages/components/src/**`, which the review checklist treats as UI-affecting, and `gh run list --workflow=visual-regression-tests.yml` returns nothing for this branch.

Stating it plainly: this one looks dismissible. The diff adds an `aria-label` and nothing else that reaches the DOM, so the rendered pixels cannot change, and the PR description already says as much. It is raised because the rule is unconditional, not because a diff is expected. No workflow has been dispatched.

#### Warnings

**3. The slot test pins the hole described in Critical 1.** Inline on `CloseButton.test.tsx`.

**4. The new test builds a fixture instead of importing a story.** Inline on `CloseButton.test.tsx`.

#### Suggestions

**5. `Dialog`'s explicit label is now byte-identical to the fallback.** Inline on `Dialog.tsx`.

**6. `'aria-labelledby'` is hand-declared rather than picked from `ButtonProps`.** Inline on `CloseButton.tsx`.

**7. `CloseButton.test.tsx:2-3` uses two separate `vitest` imports** where `Dialog.test.tsx` uses a single `import { MockInstance, vi } from 'vitest'`.

Not raised: the warning fires on every render. `Dialog.tsx:55-66` does the same, so that is the established precedent in this package, as the PR description notes.

### Recommendation

**Request Changes**, on Critical 1 alone. The rest of the change is well argued and tracks the ticket closely, but the `slot` guard reintroduces the same WCAG 4.1.2 failure for `slot="close"`, which is the most likely way a consumer reaches for this component.

---

packages/components/src/CloseButton/CloseButton.tsx
```diff
@@ -21,14 +24,36 @@ export const CloseButton = ({
...
+  const stringFormatter = useLocalizedStringFormatter(intlMessages);
   const classNames = useClassNames({
     component: 'CloseButton',
     className,
     size,
     variant,
   });
+
+  const hasAccessibleName =
+    !!props['aria-label'] || !!props['aria-labelledby'] || !!props.slot;
```
sebald: :red_circle: **`!!props.slot` is treated as proof of an accessible name, but React Aria slots do not all supply one.**

The guard stands down both the fallback and the warning for any `slot`. That is right for `remove`, because RAC's `TagGroup` labels it. It is not right in general.

Verified in `react-aria-components@1.20.0`, `dist/private/Dialog.cjs:157-166`:

```js
[ButtonContext, { slots: { [DEFAULT_SLOT]: {}, close: { onPress: () => state?.close() } } }]
```

RAC's `Dialog` registers a `close` slot that provides **only `onPress`**. No `aria-label`, no `aria-labelledby`.

So `<CloseButton slot="close" />` inside a Dialog, which is the documented RAC idiom for a custom close button and the most obvious slot name anyone would reach for on this component, gets no fallback, no warning, and no accessible name. That is DST-1769 reproduced, silently, through the fix for DST-1769.

The guard generalises from the one slot that happens to carry a label to every slot. Narrowing it to what is actually resolved would hold:

```tsx
const slottedContext = useSlottedContext(ButtonContext, props.slot);
const hasAccessibleName =
  !!props['aria-label'] ||
  !!props['aria-labelledby'] ||
  !!slottedContext?.['aria-label'] ||
  !!slottedContext?.['aria-labelledby'];
```

A cheaper variant keeps the fallback off whenever a `slot` is set, preserving today's `TagGroup` behaviour exactly, but still warns when the resolved context supplies no label.

packages/components/src/CloseButton/CloseButton.tsx
```diff
@@ -12,6 +14,7 @@ interface CloseButtonProps extends Pick<
   className?: string;
   style?: CSSProperties;
   ref?: React.Ref<HTMLButtonElement>;
+  'aria-labelledby'?: string;
```
sebald: :bulb: `'aria-labelledby'` is hand-declared while every other prop comes from `Pick<ButtonProps, ...>`. Adding it to the `Pick` list on line 12 keeps a single source of truth and inherits RAC's own type:

```tsx
interface CloseButtonProps extends Pick<
  ButtonProps,
  'onPress' | 'size' | 'variant' | 'slot' | 'aria-label' | 'aria-labelledby'
> {
```

packages/components/src/CloseButton/CloseButton.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+});
+
+test('does not apply the fallback when a slot supplies the name', () => {
+  // A `slot` means a parent labels the button through context. Setting
+  // `aria-label` here would win over it, so the fallback must stand down.
+  renderCloseButton(<CloseButton slot="remove" />);
+
+  expect(screen.getByRole('button')).not.toHaveAttribute('aria-label');
+  expect(warnMock).not.toHaveBeenCalled();
+});
```
sebald: :warning: **This test pins the hole described on `CloseButton.tsx`.**

It renders `<CloseButton slot="remove" />` with no provider above it, so nothing supplies a name, and then asserts that this is fine: no `aria-label`, no warning. The test is green while describing precisely the failure mode the PR exists to eliminate, a nameless close button with nothing signalling it.

The guarantee the comment describes ("a parent labels the button through context") is already covered properly by the new `TagGroup.test.tsx` case, which exercises the slot with its real provider. If the guard is narrowed, this case should either move inside a real context or go away.

packages/components/src/CloseButton/CloseButton.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+beforeEach(() => {
+  warnMock = vi.spyOn(console, 'warn').mockImplementation(() => null);
+});
+
+afterEach(() => {
+  warnMock.mockRestore();
+});
+
+const renderCloseButton = (ui: React.ReactElement) =>
+  render(<MarigoldProvider theme={theme}>{ui}</MarigoldProvider>);
```
sebald: :warning: `renderCloseButton` wraps an inline `MarigoldProvider theme={theme}`, which is the pattern `CLAUDE.md` asks you to avoid: *"Import stories instead of creating fixtures"*.

The underlying cause is that `packages/components/src/CloseButton/` has no `CloseButton.stories.tsx` at all, and the "Stories added/updated" box is unticked. For a barrel-exported public component (`index.ts:83`, untagged) whose labelling behaviour just changed, a story tagged `component-test` is the idiomatic home for the "falls back to Close" and "explicit label wins" cases.

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -105,6 +108,7 @@ const InnerDialog = ({
     >
       {closeButton && (
         <CloseButton
+          aria-label={stringFormatter.format('close')}
```
sebald, line 112: :bulb: After the `CloseButton` change this is byte-identical to what the fallback produces, so the `Dialog` diff is functionally inert.

It is defensible as written: it keeps `Dialog` out of the dev-warning path and states intent at the call site. But a one-line comment saying that would stop the next reader from deleting it as redundant.

.changeset/dst-1769_closebutton-accessible-name.md
```diff
@@ -0,0 +1,19 @@
...
+
+`<Dialog closeButton>` rendered its close button with no accessible name at all. `CloseButton` renders an icon only, `Dialog` passed no `aria-label`, and the `X` glyph contributes nothing to the name computation, so the computed name was the empty string. VoiceOver announced it as a bare "button". Every other call site in the package already passed a localized label, so `Dialog` was the only component affected, but it is one of the most used, and `closeButton` is on in most of our examples.
+
+This was a WCAG 2.1 SC 4.1.2 (Name, Role, Value) failure for every product embedding a Marigold dialog. It was not a total loss of function: React Aria's `Modal` still renders its own visually hidden dismiss button, and Escape still closes the dialog. The visible affordance was the unlabelled one.
+
+**`Dialog` now passes a localized `close` label**, matching `Toast`, `SectionMessage`, `Drawer`, `SidebarModal` and `FileFieldItem`.
+
+**`CloseButton` gained a fallback**, because it is public API and consumer code cannot be linted from here. When nothing else names the button, it falls back to the localized `close` string, so an unnamed close button can no longer reach production.
```
aromko: ⚠️ `packages/components/src/index.ts:82` carries `/** @internal */` on the `CloseButton` export, merged in #5789 (DST-1598) and present on this PR's own head commit. The ticket was written against a pre-#5789 view of the barrel and the changeset inherited the claim.

This text ships to release notes, so it publishes an incorrect statement about the package's API surface. The fallback is still defensible as a floor for future in-repo call sites, the rationale just needs restating.

aromko (commented):
Review notes inline.

One thing that cannot be an inline comment: `Check prose style` fails on `.changeset/dst-1625-mcp-telemetry.md`, which this PR does not touch. It was fixed on `main` by #5834 (`05cd8e021`) after this branch was cut, so merging or rebasing on `main` turns it green.

packages/components/src/CloseButton/CloseButton.tsx
```diff
@@ -21,14 +24,36 @@ export const CloseButton = ({
...
-    <Button ref={ref} className={classNames} {...props}>
+    <Button
+      ref={ref}
+      className={classNames}
+      aria-label={
+        // Set undefined, so we use the accessible label from props
+        hasAccessibleName ? undefined : stringFormatter.format('close')
+      }
+      {...props}
+    >
```
aromko, line 66: ⚠️ `aria-label` is set before `{...props}`, so a caller forwarding an optional label (`aria-label={props.label}`, a common shape) re-introduces the key as `undefined` and overwrites the fallback. The result is the exact bug this PR fixes: an unnamed button. The warning still fires, but the floor the changeset promises has a hole in it.

```suggestion
      {...props}
      aria-label={
        hasAccessibleName ? props['aria-label'] : stringFormatter.format('close')
      }
    >
```

packages/components/src/CloseButton/CloseButton.tsx
aromko: 💡 `ButtonProps` extends `Omit<RAC.ButtonProps, …>`, which already carries `aria-labelledby` via `AriaLabelingProps`. Folding it into the `Pick` on line 12 next to `'aria-label'` keeps one source of truth instead of a hand-written duplicate that can drift from RAC's type.

packages/components/src/CloseButton/CloseButton.tsx
aromko: 💡 `!!props.slot` is broader than "has a name". Only RAC's `remove` slot actually supplies a label through context, so `<CloseButton slot="close" />` satisfies `hasAccessibleName`, skips both the fallback and the warning, and renders unnamed.

The guard is still the right shape (an unconditional fallback demonstrably breaks TagGroup), so this is a known ceiling rather than a defect. Worth a line here or in the ticket so the next reader does not have to re-derive it.

sebald: Fixed locally, not yet pushed. `CloseButton` now reads the resolved slot context with `useSlottedContext(ButtonContext, props.slot)` and falls back only when that context supplies neither `aria-label` nor `aria-labelledby`. `<CloseButton slot="close" />` inside a Dialog therefore gets the fallback and the warning, and the TagGroup `remove` slot keeps its context label. The commit lands with the next push.

packages/components/src/CloseButton/CloseButton.tsx
sebald: Fixed locally, not yet pushed. `'aria-labelledby'` is now part of the `Pick<ButtonProps, …>` list and the hand-declared field is gone. Lands with the next push.

packages/components/src/CloseButton/CloseButton.test.tsx
sebald: Fixed locally, not yet pushed. The test is now "still falls back when a slot resolves no name from context": it renders `<CloseButton slot="remove" />` without a provider and asserts the `Close` fallback and the warning. The labelled slot stays covered by `TagGroup.test.tsx` against its real provider. Lands with the next push.

packages/components/src/Dialog/Dialog.tsx
sebald: Fixed locally, not yet pushed. The explicit label now carries a comment saying it is intentional: it keeps `Dialog` out of the dev-warning path and names the intent at the call site. Lands with the next push.

packages/components/src/CloseButton/CloseButton.tsx
sebald: Agreed, fixed locally and not yet pushed. `'aria-labelledby'` is now picked from `ButtonProps` next to `'aria-label'`. Lands with the next push.

packages/components/src/CloseButton/CloseButton.tsx
sebald: Went one step further than a comment, in a local commit that is not yet pushed. The guard no longer trusts `props.slot` at all. `CloseButton` asks `useSlottedContext(ButtonContext, props.slot)` what the slot resolved to and falls back only when that context carries no `aria-label` or `aria-labelledby`. So `slot="close"` inside a Dialog now gets the fallback and the warning, and `slot="remove"` in a TagGroup keeps RAC's label. The comment explaining why is in the code and the changeset. Lands with the next push.

packages/components/src/CloseButton/CloseButton.test.tsx
sebald: Fixed in 9a81b281e. `CloseButton.stories.tsx` is new, with a `Basic` story tagged `component-test` and two `.test()` cases for the fallback name and the explicit label. `CloseButton.test.tsx` now imports `Basic.Component` instead of wrapping an inline `MarigoldProvider`, and keeps the warning and slot cases that need a `console.warn` spy.

.changeset/dst-1769_closebutton-accessible-name.md
sebald: Right, `index.ts` has carried `@internal` on this export since #5789. Fixed in 9a81b281e: the changeset no longer calls `CloseButton` public API and restates the fallback as a floor for in-repo call sites, since `Dialog` shipped unnamed precisely because nothing forced a name.

packages/components/src/CloseButton/CloseButton.tsx
sebald: Confirmed and fixed in 9a81b281e. `aria-label` is now set after `{...props}`, so a forwarded `undefined` can no longer wipe out the fallback. I checked the TagGroup side too: React Aria's `mergeProps` keeps the slotted context label when the local value is `undefined`, so `remove` buttons stay "Remove News". A unit test pins the forwarded-`undefined` shape.

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -104,7 +107,11 @@ const InnerDialog = ({
       )}
     >
       {closeButton && (
+        // The label is what `CloseButton`'s fallback would produce anyway, but
```
aromko: Maybe in general remove/reduce the made comments.

sebald: Trimmed in d48741be4. Every comment this PR added is now one or two lines, including this one, which just says the label is explicit to keep `Dialog` out of `CloseButton`'s dev warning. The longer rationale stays in the changeset.

aromko (approved):

PR #5852, opened 2026-09-17
ci(DST-1748): stop Renovate cancelling its own runs, lift the 2-PR-per-week ceiling

sarahgm:
# Description

Renovate has been delivering exactly 2 dependency updates per week against a backlog of 43 rate-limited items. Two independent config faults caused it, and this fixes both.

**It cancelled its own runs.** `issues: [edited]` combined with `cancel-in-progress: true` meant Renovate's own Dependency Dashboard write triggered a run that killed the run that wrote it. 15 of the 23 scheduled runs in this workflow's history died this way, and the split on the issue-triggered ones names the victim. Of the last 100, going back to late May, the 61 that a human dashboard tick started were cancelled without exception, while 33 of the 39 started by the bot's own write succeeded. The dashboard was not being neglected, it was being used constantly against a pipeline that ate the result. The job now ignores edits made by a bot, in-progress runs are never cancelled, and the runs that get skipped are kept out of the shared concurrency group, so a tick made during a run queues behind it instead of replacing it or being replaced.

**The ceiling was arithmetic.** `prHourlyLimit` was unset and falls back to Renovate's default of `2`, and the cron fired once a week. Two per run times one run per week is exactly the observed rate. It is now set explicitly to `5`, `branchConcurrentLimit` goes to `10`, and routine updates run Monday and Thursday. That is 10 PRs a week against today's 2, with at most 10 branches standing open at once.

Two smaller fixes ride along. The token action moves from the deprecated `app-id` to `client-id`, and `vulnerabilityAlerts` gets `branchConcurrentLimit: 0`. Renovate counts vulnerability branches separately from the rest, so other dependency PRs were never competing with a security fix for a slot. What did bind is the top level: the `vulnerabilityAlerts` default carries `prConcurrentLimit: 0`, and `branchConcurrentLimit` falls back to it only when it is not a number, so the top-level value capped concurrent security branches instead of leaving them unlimited.

Closes DST-1748

## Where this differs from the ticket

Five deliberate departures, all decided during implementation:

- **The workflow runs daily, the routine cadence is Monday and Thursday.** Twice a week at 5 per run is a steadier trickle than the ticket's daily sketch, but the two are separable. The cron is `0 0 * * *` and `schedule: ["* * * * 1,4"]` in `renovate.json` is what holds routine updates to twice a week. `vulnerabilityAlerts` carries `schedule: []` by default, so security work ignores that window and lands within a day of the alert instead of waiting up to four days for the next Monday. `updateNotScheduled: false` keeps the five extra runs from rebasing standing branches, so routine behaviour and CI cost match a Monday and Thursday cron. A dashboard tick still bypasses the schedule, not just the limits (`!config.isScheduledNow && !dependencyDashboardCheck` in `workers/repository/update/branch/index.ts`).
- **`prHourlyLimit` is 5 and `branchConcurrentLimit` is 10.** They bound different things rather than matching. The hourly limit is the per-run rate, since a run finishes inside one UTC hour, and the branch limit is the standing total. Majors are not gated, never automerge and hold their branch while they wait for review, so a limit of 5 would have let five of them stop even the updates that automerge themselves away.

- **Skipped runs get their own concurrency group.** The sender check is at job level, so Renovate's dashboard write still creates a run and only the job is skipped. That run would otherwise enter the shared group and evict a tick queued behind a run in flight, which is the failure this ticket exists to end. 19 of the last 60 issue-triggered runs came from the bot.
- **Majors are not gated behind `dependencyDashboardApproval`.** Dropped by request. Majors keep opening PRs and keep competing for slots. They never automerge, since the automerge rule matches minor and patch only.
- **Autodiscovery is removed rather than pinned.** The ticket proposed keeping `RENOVATE_AUTODISCOVER: true` and narrowing it with `RENOVATE_REPOSITORIES`. That does not work. In `workers/global/autodiscover.js` the discovered list *replaces* `repositories` wholesale, and a configured entry only supplies per-repo settings for a repo already discovered. Dropping the flag is what makes the pin take effect.

## Done outside this diff

Unattended automerge for minor and patch updates is now live, and it needed two branch-protection changes that are settings rather than code:

1. **17 CI checks promoted to required on `main`**, bringing the total to 20 with the three Vercel contexts. Previously only the Vercel checks were required, which meant GitHub's native auto-merge would merge as soon as those went green, ignoring Typecheck, Lint, Format, Unit Tests, Storybook Tests and the repo guards. This also closed a gap unrelated to Renovate: until now a human PR with a red test suite could be merged into `main`.
2. **`marigold-renovate` added to `bypass_pull_request_allowances.apps`.** Renovate already enabled native auto-merge and already performed the merge. The single unmet condition was the one approving review.

Order mattered: granting the bypass first would have opened a window where a red suite merged itself.

### Still to do, after merge

**Delete the `RENOVATE_APP_ID` repo variable.** Nothing reads it once this lands, and deleting it any earlier breaks the workflow on `main`, which still does. Its value is `3315507`, recorded here so the step is reversible. A GitHub App ID is not a secret, the private key stays in `secrets`.

**Watch whether majors ratchet into every branch slot.** Majors never automerge under the minor and patch rule and nothing gates them, so each one holds a branch slot until a human deals with it, and the top-level `branchConcurrentLimit: 10` is the only ceiling. The symptom is patch and minor PRs stopping while the open Renovate list is all majors, visible from the PR list with nothing to instrument.

It is not observable yet, which is why no ticket is filed. Non-majors sort first and are created out of the same running count, so with 43 items rate-limited they saturate all 5 slots on every run and majors are created out of leftover headroom only. The real steady state appears once that backlog drains, in roughly four to five weeks. If the symptom shows up then, the fix is a `packageRules` entry giving majors their own `branchConcurrentLimit: 5`, which is `branchConcurrentLimit` minus `prHourlyLimit` and so leaves a full run budget for the updates that automerge themselves away. The reasoning, the alternative of gating majors behind `dependencyDashboardApproval`, and the residual worth measuring first are all in [this thread](https://github.com/marigold-ui/marigold/pull/5852#discussion_r4036416099).

Two consequences worth knowing. A flaky check now blocks `main` for everyone until the job is re-run, and `enforce_admins: true` means nobody can override. And minor and patch bumps of non-`0.x` packages now land in `main` without a human, with CI as the only safety net.

## Test Instructions

Nothing here can be exercised before merge, because `on: schedule` and `on: issues` only fire from the workflow on `main`. Verified statically so far: `renovate-config-validator` passes against the repo schema (not just the looser global one), the workflow YAML parses and every changed value reads back as intended, and Prettier is clean.

After merge:

1. Tick a rate-limited checkbox on the [Dependency Dashboard](https://github.com/marigold-ui/marigold/issues/5292) and confirm a Renovate run starts within ~30s and is **not** cancelled.
2. While that run is in flight, tick two more boxes. Confirm the running job survives, one run queues, and the queued run picks up both ticks.
3. Confirm Renovate's own dashboard write creates a run whose job is skipped, that it lands in its own `renovate-<run_id>` group, and that a tick queued at the same time still runs afterwards.
4. Confirm the run creates up to 5 PRs rather than 2, and that it keeps going up to 10 open branches across runs.
5. Confirm a ticked box still produces its PR even beyond the limit and outside the Monday and Thursday window, since a dashboard tick bypasses both the limits and the schedule by design.
6. Confirm the run log shows one repository, not every repo the app is installed on.
7. Confirm no `app-id` deprecation warning from `actions/create-github-app-token`.
8. Wait for a minor or patch PR and confirm it merges itself only after all 20 required checks are green.
9. On a run outside Monday and Thursday, confirm the log reports `not-scheduled` for routine updates, that no new dependency PR is created, and that standing branches are left alone rather than rebased.
10. Confirm a `[SECURITY]` PR appears within a day of an alert rather than waiting for the next scheduled day.

## Breaking Changes

No, for the published packages. The branch-protection changes above do change the merge workflow for everyone.

## Checklist

- [ ] ~~Stories added/updated~~ N/A, CI configuration only
- [ ] ~~Unit tests added/updated~~ N/A, CI configuration only
- [ ] ~~Component documentation added/updated~~ N/A, CI configuration only
- [ ] ~~Accessibility reviewed~~ N/A, no UI
- [ ] ~~Visual regression tests updated~~ N/A, no UI
- [ ] ~~Changeset added~~ N/A, `.github/` is not a published package


OsamaAbdellateef (commented):
## Automated Code Review

### Overview
| Field | Value |
|-------|-------|
| Author | @sarahgm |
| Branch | `chore/DST-1748-unblock-renovate-self-cancelling-runs` → `main` |
| Files Changed | 2 (`.github/renovate.json`, `.github/workflows/renovate.yml`) |
| Additions | +16 |
| Deletions | -9 |
| CI | All 21 checks passing |
| Visual Regression | Not needed — no `packages/`, `themes/`, `.storybook/`, `*.stories.tsx`, or `*.styles.ts` files touched |

### Linked Ticket
**[DST-1748](https://reservix.atlassian.net/browse/DST-1748)**: 🏗️ Unblock Renovate: stop it cancelling its own runs and lift the 2-PR-per-week ceiling
- Status: In Review
- Expected Outcome: Renovate runs to completion instead of cancelling itself; ticking a dashboard checkbox reliably produces the PR; backlog drains at a chosen rate.
- The four documented departures from the ticket (biweekly vs. daily cron, `prHourlyLimit: 5` not `10`, dropping `dependencyDashboardApproval` for majors, removing autodiscovery instead of pinning it) are all explained with concrete reasoning and check out against the ticket's own text.

### Code Review Checklist
This is a CI-configuration-only change (no components, no React, no TypeScript, no tests), so the usual component/TypeScript/testing/accessibility categories don't apply. Verified instead:

- [x] `prHourlyLimit: 5` set explicitly, matches unchanged `branchConcurrentLimit: 5` (the root cause: an unset `prHourlyLimit` defaulted to 2 and was the binding limit)
- [x] `cron: '0 0 * * 1,4'` correctly targets Monday + Thursday
- [x] `cancel-in-progress: false` + `github.event.sender.type != 'Bot'` — verified against GitHub Actions concurrency semantics; matches the failure sequence in the ticket (bot's own dashboard write → `issues.edited` → self-cancel)
- [x] `client-id` input — confirmed `actions/create-github-app-token@v3` supports it, and confirmed via the GitHub API that `RENOVATE_APP_CLIENT_ID` already exists as a repo variable, so the switch won't break the workflow at runtime
- [x] `RENOVATE_REPOSITORIES: ${{ github.repository }}` replacing `RENOVATE_AUTODISCOVER: true` — correct per the PR's stated reasoning (autodiscovery replaces rather than merges with `repositories`)
- [x] `.github/renovate.json` is valid JSON

### Issues Found

No critical or warning-level issues.

- `.github/renovate.json` — `vulnerabilityAlerts.branchConcurrentLimit: 0` is likely a no-op. Per Renovate's docs and [renovatebot/renovate#45686](https://github.com/renovatebot/renovate/pull/45686), vulnerability-alert PRs already bypass `branchConcurrentLimit`/`prConcurrentLimit` by default unless explicitly tightened via `vulnerabilityAlerts.*`. Setting it to `0` (the "no limit" value) just makes explicit what Renovate already does by default — harmless, and arguably good as self-documenting intent, but worth knowing it's not adding new behavior.
- The old `RENOVATE_APP_ID` repo variable is now unreferenced by any workflow after this change. Fine to leave for now; worth a cleanup pass next time someone's in these settings.

### Recommendation
**Approve**

Small, well-scoped config fix with unusually thorough root-cause documentation in the PR body. Independently verified the two claims that mattered most for correctness — that `client-id` is supported by the action, and that `RENOVATE_APP_CLIENT_ID` actually exists as a repo variable — and both check out. CI is green, and nothing here touches UI code so visual regression testing doesn't apply.

---

sarahgm:
Thanks for the thorough pass. Addressing both suggestions.

**1. `vulnerabilityAlerts.branchConcurrentLimit: 0`** is not a no-op, and it is load-bearing in this config specifically. The default exemption is real for `prConcurrentLimit` and not for `branchConcurrentLimit`:

- The `vulnerabilityAlerts` default in `lib/config/options/index.ts` is `{ groupName, schedule, dependencyDashboardApproval, minimumReleaseAge, rangeStrategy, commitMessageSuffix, branchTopic, prCreation, vulnerabilityFixStrategy, prConcurrentLimit: 0 }`. It carries `prConcurrentLimit: 0` and no `branchConcurrentLimit`.
- `calcLimit` inherits one from the other only when the branch value is not a number:

  ```ts
  // inherit prConcurrentLimit value incase branchConcurrentLimit is null
  if (!isNumber(limit) && limitName === 'branchConcurrentLimit') {
    limit = upgrade.prConcurrentLimit;
  }
  ```

  (`lib/workers/global/limits.ts`)
- This config sets `branchConcurrentLimit: 10` at the top level. That survives the merge into a vulnerability upgrade, so the value is a number, the fallback never fires, and the security branch budget is capped at 10 instead of unlimited.

On renovatebot/renovate#45686: that one is a `feat(onboarding)` change to the onboarding PR's wording. Its text covers `prConcurrentLimit` and `prHourlyLimit` and never mentions `branchConcurrentLimit`, which I suspect is where the conflation comes from.

It would be a genuine no-op if the top level left `branchConcurrentLimit` unset. Since it does not, I have added a `description` to the object recording this, because the line evidently reads as dead on inspection: f529b847f

**2. `RENOVATE_APP_ID`** is confirmed unreferenced in the repo, and agreed it stays for now, though the reason is sharper than optional: `main` still has `app-id: ${{ vars.RENOVATE_APP_ID }}` at `.github/workflows/renovate.yml:28`, so deleting the variable before this merges would break the token step on the default branch. It is a post-merge cleanup rather than a discretionary one.

.github/renovate.json
```diff
@@ -6,7 +6,8 @@
   ],
   "timezone": "Europe/Berlin",
   "ignoreScripts": true,
-  "branchConcurrentLimit": 5,
+  "prHourlyLimit": 5,
+  "branchConcurrentLimit": 10,
```
aromko, line 10: ⚠️ The description rejects `branchConcurrentLimit: 5` because "a limit of 5 would have let five of them stop even the updates that automerge themselves away." That argument does not stop at 5: DST-1748 counts 19 majors in the backlog, and majors are not gated, never automerge, and hold their branch while waiting for review.

`getConcurrentBranchesCount` counts standing branches, so once 10 major branches exist they occupy every slot and keep it, and no minor or patch gets through regardless of cron frequency. Gating majors was dropped by request, so this is a flag rather than a re-litigation.

Symptom to watch for after merge: patch PRs stop appearing while the open list is all majors.

.github/workflows/renovate.yml
```diff
@@ -2,30 +2,34 @@ name: Renovate
...
   workflow_dispatch:
   issues:
     types: [edited]
 
 concurrency:
-  group: renovate
-  cancel-in-progress: true
+  # A run the `if` will skip gets its own group, so it cannot take the single
+  # pending slot from a tick queued behind a run in flight.
+  group: renovate-${{ github.event.sender.type == 'Bot' && github.run_id || 'shared' }}
```
aromko: 💡 The comment above says "A run the `if` will skip gets its own group", but the group keys on `sender.type == 'Bot'` while the `if` also skips issue edits on any non-`Dependency Dashboard` issue. Those runs still land in `shared`, where they can evict a tick sitting pending behind a run in flight, which is the failure this PR exists to end.

Latent today: all 57 of the last 57 issue-triggered runs were on "Dependency Dashboard". Worth aligning so a later widening of the `if` does not silently reopen it.

```suggestion
  group: >-
    renovate-${{ (github.event_name == 'issues' && (github.event.issue.title != 'Dependency Dashboard' || github.event.sender.type == 'Bot')) && github.run_id || 'shared' }}
```

sarahgm: Good catch, applied in 0898abc07 with your expression as suggested.

Confirming the mechanism for the record: `concurrency` is evaluated at the workflow level, so a run enters the group before the job's `if` is read. Being skipped afterwards does not hand the slot back, which is exactly what lets a skipped run evict a pending tick.

I also rewrote the comment above it, since it asserted the property the expression did not actually have.

One correction to the latency figure, in your favour: all **87** issue-triggered runs in the last 100 workflow runs were on the Dependency Dashboard, not 57 of 57. Same conclusion, larger sample.

.github/renovate.json
sarahgm: Confirmed, and left open pending a decision.

The conclusion holds. `getConcurrentBranchesCount` counts standing branches, majors never automerge under the `minor`/`patch`-only rule, and nothing gates them. The dashboard currently lists 15 majors rate-limited, so a full backlog does exceed a limit of 10 and would hold every slot.

The path there is not the one described, though, and it changes what the symptom looks like. `sortBranches` orders `pin, digest, patch, minor, major`, so patches and minors claim slots first on every run and majors are only created out of leftover headroom. It is not an immediate takeover once 10 major branches exist. It is a slow ratchet: minors and patches automerge away and free their slots, majors accumulate and never leave, and the standing set drifts toward all-majors over weeks rather than within a run.

That makes the symptom you named later-arriving but no less real, and you are right that cron frequency cannot help either way.

The call for now is to ship this as-is and watch for your tell, holding a `packageRules` entry that gives majors their own `branchConcurrentLimit` as the follow-up. Two things pushed it that way.

`handleConcurrentLimits` compares a per-branch limit against the total `Branches` count rather than a per-type count. Combined with patches and minors sorting first, that means the reservation has to be set *high* to work at all: `matchUpdateTypes: ["major"]` with `branchConcurrentLimit: 7` blocks majors once 7 branches stand in total, leaving slots 8 to 10 for the automerging updates. Setting it low, at 3, would throttle majors to roughly zero instead, because the total is already past 3 by the time majors are evaluated, and the backlog would never drain.

And the ratio that ought to set that number is not measurable yet. Renovate has been self-cancelling and capped at 2 PRs per week for as long as the run history goes back, so a few weeks on the fixed config gives the real steady-state ratio of automerging updates to standing majors rather than a guess.

Leaving this thread open so it does not get lost.

aromko: Agreed on shipping as-is. Current state backs it: only 3 Renovate branches stand open right now, so nothing is starving yet.

Your mechanism correction is right and I had it wrong. Verified both halves: `sortBranches` is `pin, digest, patch, minor, major, lockFileMaintenance` with `prPriority` sorting ahead of all of it (`process/sort.ts`), and `Branches` is seeded once by `getConcurrentBranchesCount` then incremented per created branch inside the loop (`process/write.ts:157` and `:219`). So the count majors are measured against includes the non-majors created earlier in the same run. Ratchet, not takeover.

On the number, I would go to 5 rather than 7. Non-majors sort first, so in the worst case they get `min(prHourlyLimit, 10 - M)` slots per run:

- M = 7 leaves 3, below the run budget of 5. Once 7 majors stand, the automerging stream drops from 10 PRs a week to 6. That pays 40% of the throughput this PR exists to create, to hold 2 more unreviewed majors open.
- M = 5 leaves exactly 5, so the automerging stream is never throttled.

Your "throttled to roughly zero" objection is real, but I think it lands softer at 5 than at 3. Majors are blocked once the running total reaches 5, and since non-majors sort first and automerge away between runs, the standing set at run start is mostly majors. That settles at roughly 3 to 5 concurrent majors, not zero. It is a WIP limit rather than a throttle, and 3 to 5 open majors is already more than gets reviewed in a week, so review capacity stays the binding constraint either way. The remaining 11 or so wait on the dashboard, which costs nothing, rather than as open PRs that rebase and re-run 20 required checks on every push to `main`. A standing major PR is not free, and that is the part I would weigh most.

```json
{
  "description": "Majors never automerge and are review-bound, so cap the branch slots they can hold. The limit is compared against the total branch count rather than a per-type one, so 5 against a branchConcurrentLimit of 10 leaves a full prHourlyLimit of headroom for the updates that automerge themselves away.",
  "matchUpdateTypes": ["major"],
  "branchConcurrentLimit": 5
}
```

Separately, and once only since it was dropped by request: `dependencyDashboardApproval: true` on majors dissolves the arithmetic instead of tuning it, and is the only option where 16 unreviewed majors cost no CI. If the objection was that majors should keep surfacing as PRs, that is precisely the property creating the contention, so it is worth having the reason in the thread before the follow-up ticket picks a number.

.github/workflows/renovate.yml
```diff
@@ -2,30 +2,38 @@ name: Renovate
 
 on:
   schedule:
-    # Monday at 2am Europe/Berlin (0am UTC)
-    - cron: '0 0 * * 1'
+    # Monday and Thursday at 00:00 UTC
+    - cron: '0 0 * * 1,4'
```
sebald: ⚠️ **Security-fix latency is now the binding constraint, and nothing says so.**

The limit side of the ticket's "security patches inherit the same ceiling" is genuinely fixed here, and I checked both halves against Renovate's source: vulnerability branches are counted separately (`vulnerabilityCountName` → `VulnerabilityBranches`) and they skip the hourly PR limit outright (`handleConcurrentLimits`: `if (!config.isVulnerabilityAlert)`), so neither `prHourlyLimit: 5` nor the top-level `branchConcurrentLimit` binds them any more.

What still binds is cadence. `schedule: ["at any time"]` and `prCreation: immediate` only take effect **inside a run**, and the only cadence is `0 0 * * 1,4`. An alert appearing Thursday at 01:00 UTC waits until Monday: a worst case of ~3.5 days.

This is better than the weekly cron it replaces, so it is not a regression. But "Where this differs from the ticket" argues Mon/Thu purely on routine review volume and never weighs this, and security latency is one of the ticket's three stated reasons for the work. Worth either a separate daily trigger for vulnerability work or writing the tradeoff down, so the next person does not have to re-derive it. A plain daily cron is not the answer: at `prHourlyLimit: 5` that is ~35 PRs a week, which is the opposite of what the cadence was chosen for.

.github/workflows/renovate.yml
```diff
@@ -2,30 +2,38 @@ name: Renovate
...
   workflow_dispatch:
   issues:
     types: [edited]
 
 concurrency:
-  group: renovate
-  cancel-in-progress: true
+  # A run the `if` will skip gets its own group, so it cannot take the single
+  # pending slot from a tick queued behind a run in flight. The condition has to
+  # mirror the job's `if` exactly, or a skipped run still lands in `shared`.
```
sebald: 💡 The comment says the condition "has to mirror the job's `if` exactly", but it mirrors its **negation**: the group keys on the runs the `if` will skip. The expression itself is correct, and I verified the `github.event_name == 'issues'` guard short-circuits before the null `issue.title` comparison on `schedule` and `workflow_dispatch`, so those still land in `shared`. One word would stop the next reader checking it twice.

```suggestion
  # mirror the job's `if` inverted, or a skipped run still lands in `shared`.
```

.github/workflows/renovate.yml
```diff
@@ -2,30 +2,38 @@ name: Renovate
...
-    # On issue edits, only react to the Renovate Dependency Dashboard.
+    # Renovate's own dashboard write is an issue edit, so a run would trigger another.
     if: >-
-      github.event_name != 'issues' ||
-      github.event.issue.title == 'Dependency Dashboard'
+      github.event_name != 'issues' || (
+        github.event.issue.title == 'Dependency Dashboard' &&
+        github.event.sender.type != 'Bot'
+      )
     runs-on: ubuntu-latest
```
sebald, line 31: 💡 No `permissions:` block on the job. Renovate runs on the app token from the step below, so the default `GITHUB_TOKEN` grants are unused by the step that actually matters. `permissions: {}` would shrink the blast radius of a workflow reachable from `issues: [edited]`.

Pre-existing rather than introduced here, so purely optional, but this PR is already the one touching the trigger surface.

.github/workflows/renovate.yml
```diff
@@ -2,30 +2,38 @@ name: Renovate
...
     steps:
       - uses: actions/checkout@v6
 
       - uses: actions/create-github-app-token@v3
         id: app-token
         with:
-          app-id: ${{ vars.RENOVATE_APP_ID }}
+          client-id: ${{ vars.RENOVATE_APP_CLIENT_ID }}
```
sebald, line 42: 💡 `vars.RENOVATE_APP_ID` is now unused. Both it and `RENOVATE_APP_CLIENT_ID` exist as repo variables today (confirmed via `gh api repos/marigold-ui/marigold/actions/variables`), so deleting the stale one stops a future editor updating the variable the workflow no longer reads.

.github/renovate.json
sarahgm: Taking your 5, and your arithmetic is the reason. `10 - 5 = 5` is exactly `prHourlyLimit`, so the automerging stream is never throttled, and the 40% throughput cost at 7 is not worth two more unreviewed majors. It also gives a better rule than the number I had picked: the reservation is `branchConcurrentLimit - prHourlyLimit`, not an arbitrary high value. That generalises if either limit moves later, which "7" did not.

One residual worth writing into the follow-up rather than leaving implicit. Because `Branches` is incremented inside the loop and non-majors sort first, a run where non-majors fill all five slots evaluates majors at a running total of 5 and creates none. So the steady state is not "3 to 5 concurrent majors" as a property of the limit, it is whatever majors slipped in on runs where non-major volume happened to be under 5. That is still a WIP limit rather than a throttle, and your point that review capacity is the binding constraint either way holds, but the drain rate depends on non-major volume in a way the number itself does not show.

Since the follow-up will now be picking a number against measured data, that residual is the thing to measure: how often a run has fewer than 5 non-majors available. If that is rare, 5 drains majors very slowly and the case for `dependencyDashboardApproval` gets stronger on its own.

On which, your reason is recorded here as you asked. The reason it was dropped is planning provenance rather than arithmetic: it was dropped by request, so re-adding it is a product call about whether majors keep surfacing as PRs, not a config tuning decision. Your observation that the surfacing is itself what creates the contention is the sharpest argument against that call that has been made, and it should go in the follow-up ticket verbatim rather than being re-derived.

Note that one thing changed under this thread since you wrote it. The cron is now daily with `schedule: ["* * * * 1,4"]` holding routine updates to Monday and Thursday, so majors are still only created twice a week and the accumulation rate you reasoned about is unchanged. See the security-latency thread for why.

Leaving this open until the follow-up ticket exists, which is not filed yet.

.github/workflows/renovate.yml
sarahgm: Fixed in e98549a5, and you were right that the cadence was the last thing binding. Neither of your two options quite, so here is what landed and why.

The cadence moves out of the cron and into Renovate. The workflow cron is now `0 0 * * *`, and `schedule: ["* * * * 1,4"]` in `renovate.json` is what holds routine updates to Monday and Thursday. Your objection to a plain daily cron is exactly what that prevents: the schedule blocks routine branch creation on the other five days, so the run happens without producing the ~35 PRs a week you warned about. `vulnerabilityAlerts` carries `schedule: []` in its documented default, and `isScheduledNow` short-circuits to `true` on an empty schedule, so security work ignores the window and lands within a day of the alert.

Three things I checked rather than assumed, since a silently wrong schedule string would have turned this into precisely the failure you named:

- A dashboard tick still bypasses the schedule and not just the limits. `update/branch/index.ts:406` is `if (!config.isScheduledNow && !dependencyDashboardCheck)`, so a tick skips the whole block.
- `updateNotScheduled: false` is in there so the five extra runs do not rebase standing branches. Without it, up to 10 Renovate branches would rebase on days that previously had no run at all, each one re-running 20 required checks. `main` has `strict: false`, so a branch being behind `main` never blocks its automerge, which is what makes deferring those rebases to Monday and Thursday free rather than slow.
- `renovate-config-validator --no-global --strict` passes in repo mode, which is the schema that validates schedule strings. I ran a negative control to confirm that check is live: `"0 0 * * 1,4"` is rejected with "has cron syntax, but doesn't have * as minutes" and `"utter nonsense here"` with "Failed to parse".

One correction to your figure, and it is slightly worse than you had it. Thursday 00:00 UTC to Monday 00:00 UTC is 4 days, so an alert at Thursday 01:00 waited 3 days and 23 hours rather than ~3.5. Same conclusion, and it is now under 24 hours.

The tradeoff is written down either way. "Where this differs from the ticket" now leads with the cron and schedule being separate things, and the test instructions gained a check that an off-day run reports `not-scheduled` and creates nothing, plus one that a `[SECURITY]` PR appears within a day.

.github/workflows/renovate.yml
sarahgm: Fixed in 8526a9c7, with your wording. It reads "mirror the job's `if` inverted" now.

You are right that this is the second reader in a row checking that expression twice, since my own previous comment on it asserted the same wrong property. The expression has not changed, only the sentence above it.

.github/workflows/renovate.yml
sarahgm: Applied in bfec39d6, though as `permissions: { contents: read }` rather than `permissions: {}`, and the difference matters.

`actions/checkout@v6` authenticates with `github.token` by default, so a token with no scopes at all puts the clone at the mercy of whether GitHub falls back to anonymous read for a public repo. It probably would, but "probably" is not what you want holding up the only step that has to succeed before Renovate starts. One read scope is the blast radius you were asking for, since everything that actually matters authenticates with the app token from the step below.

Placement follows the convention in the rest of the repo, after `runs-on`, matching `test.yml`, `codeql-analysis.yml` and `dependency-track.yml`, all of which use `contents: read` for the same reason.

Agreed on pre-existing rather than introduced, and it was worth doing here for the reason you gave: this is the PR touching the trigger surface.

.github/workflows/renovate.yml
sarahgm: Confirmed unused, and I checked the same thing from the other direction: after this merges, `RENOVATE_APP_ID` is referenced by nothing in the repo. It is also still referenced by exactly one thing right now, which is `main`'s copy of this workflow, so deleting it before this lands breaks Renovate on `main`. That makes it a post-merge step rather than something this PR can close.

It is now written into the PR body under "Still to do, after merge", together with its value so the deletion is reversible. Leaving this thread open until the variable is actually gone, so it does not quietly become the stale variable the next person updates.

sarahgm:
@OsamaAbdellateef one correction on the review above, since the claim is the kind that gets copied into a later decision.

`vulnerabilityAlerts.branchConcurrentLimit: 0` is not a no-op. Two separate mechanisms are being conflated:

**Counting** is indeed separate. `lib/workers/global/limits.ts` keeps `VulnerabilityBranches` and `VulnerabilityConcurrentPRs` apart from the general counters, and `handleConcurrentLimits` skips the hourly PR limit outright with `if (!config.isVulnerabilityAlert)`. So other dependency PRs were never competing with a security fix for a slot, which is the part your sources are about.

**The limit value applied to that separate counter** still comes from config, and that is where it bit. The documented `vulnerabilityAlerts` default is `{ groupName: null, schedule: [], dependencyDashboardApproval: false, minimumReleaseAge: null, rangeStrategy: 'update-lockfile', commitMessageSuffix: '[SECURITY]', branchTopic: ..., prCreation: 'immediate', vulnerabilityFixStrategy: 'lowest', prConcurrentLimit: 0 }`. Note what is absent: there is no `branchConcurrentLimit` in it. And `calcLimit` inherits one only when the other is not a number:

```ts
// inherit prConcurrentLimit value incase branchConcurrentLimit is null
if (!isNumber(limit) && limitName === 'branchConcurrentLimit') {
  limit = upgrade.prConcurrentLimit;
}
```

Because `vulnerabilityAlerts` is `mergeable: true`, the top-level `branchConcurrentLimit: 10` merges in, so `limit` *is* a number, the inheritance never fires, and concurrent security branches would be capped at 10 rather than unlimited. The explicit `0` is what makes `calcLimit` return 0, which is the "no limit" value. Remove the line and the cap comes back.

The config carries that reasoning as a `description` now, added in f529b847 about 20 minutes after your review, so the diff you read did not have it. Your other point about `RENOVATE_APP_ID` is confirmed and is now a post-merge step in the PR body, tracked in [the thread @sebald opened](https://github.com/marigold-ui/marigold/pull/5852#discussion_r4037381644).

One stale detail for the record: the review says `prHourlyLimit: 5` "matches unchanged `branchConcurrentLimit: 5`". It was already 10 by then, raised in 08d762fb about 14 minutes before the review. They bound different things and are not meant to match, which is the second bullet under "Where this differs from the ticket".

.github/renovate.json
sarahgm: Following up on the `dependencyDashboardApproval` half, since "dropped by request" is the provenance rather than the reason and you asked for the reason.

The first reason offered was operational: approval checkboxes get forgotten, so gating majors there would mean they never happen. I went to check that and it does not hold, so I am not going to rest the case on it. Over the last 100 issue-triggered runs of this workflow, back to 2026-05-29:

```
61  sarahgm                 cancelled
33  marigold-renovate[bot]  success
 6  marigold-renovate[bot]  cancelled
```

I ticked boxes on that dashboard 61 times and every one of those runs was killed, while the bot's own writes are the ones that survived. So the dashboard was not being neglected, it was being used constantly against a pipeline that ate the result. That is also a cleaner statement of this PR's root cause than the "23 of 44" figure in the description, and it cuts against the forgetfulness argument rather than for it.

Two reasons I think do hold.

**A major cannot be triaged from a checkbox.** A dashboard row is a package name and a version. Deciding a major here needs the diff, the CI signal, the changelog, and for anything in the react-aria family a `scripts/check-react-aria-dedupe.mjs` run against an actual install. Gating majors there means ticking blind to find out what the PR does, which is the same review round trip with an extra hop in front of it. The dashboard is a good place to hold work you have already decided about, which is what the rate-limited section does today, and a poor place to decide from.

**Your cost objection is now mostly answered by the other half of this thread.** You priced a standing major PR at a rebase and 20 required checks on every push to `main`, and that was the part you said you would weigh most. Since you wrote that, two things changed. The reservation caps standing majors at 5 rather than letting them reach 10, and `updateNotScheduled: false` plus the schedule change means standing branches are only rebased on Monday and Thursday instead of on every run. So the churn you were pricing is roughly a third of what it was, without hiding anything from review.

Which cuts the other way for the reservation itself: if majors keep surfacing as PRs and review capacity stays the binding constraint, that entry is the only thing between 15 open majors and every slot. It got more load-bearing, not less, which is worth carrying into the ticket.

Still leaving this open until that ticket exists.

sarahgm: Changing the plan on the follow-up, and this supersedes the "leaving this open until the follow-up ticket exists" line in both of my comments above.

There is no ticket to file yet, because the number we agreed on cannot be chosen against anything until this config has run for a while, and the first few weeks of data will point the wrong way.

Here is the part I had not thought through when I said I would file it. Non-majors sort first and `Branches` is incremented inside the write loop, so majors are only created out of leftover headroom. With 43 items rate-limited on the dashboard, non-majors will saturate all 5 slots on every run for as long as that queue lasts, which is roughly four to five weeks at 10 PRs a week. During that window majors are created roughly never and the standing set looks healthy whether or not the reservation is needed. The ratchet you flagged only becomes observable after the backlog drains. So a ticket filed today sits untouched for exactly the period in which its own input is unmeasurable, and then still needs the observation doing.

**One correction to my comment above, since it argued from a change that is not in the diff.** I wrote that "the reservation caps standing majors at 5 rather than letting them reach 10" while pricing your rebase-churn objection. That was wrong in tense: the reservation is not shipping in this PR, so standing majors can still reach 10. What did land is `updateNotScheduled: false` plus the schedule move, which is the half that cuts rebase churn to Monday and Thursday. The churn argument survives on that half alone, the WIP-limit half does not.

What carries the work instead of a ticket:

- Your watch symptom is now in the PR body under "Still to do, after merge", next to the `RENOVATE_APP_ID` deletion, so it outlives this thread. Patch and minor PRs stopping while the open Renovate list is all majors, visible from the PR list with nothing to instrument.
- Revisit in about four weeks, once the backlog is drained and the steady state is the real one. File then, with the measurement, or not at all if the symptom never appears.

What stays settled here so nobody re-derives it:

- The entry, as you wrote it: `matchUpdateTypes: ["major"]` with `branchConcurrentLimit: 5`.
- The rule behind the number, which is the thing that generalises: the reservation is `branchConcurrentLimit - prHourlyLimit`, so it moves if either limit does.
- The residual to actually measure: how often a run has fewer than 5 non-majors available. If that turns out to be rare, 5 drains majors very slowly and your `dependencyDashboardApproval` argument gets stronger on its own rather than as a re-litigation.
- Your sharpest point against the current call, which I want on the record in your words rather than my summary: surfacing majors as PRs is itself what creates the contention.

Leaving the thread open, now until we have the four weeks of runs rather than until a ticket exists.

aromko: Agreed on the plan change, and the correction is the right one: the reservation is not in this diff, so standing majors can still reach 10 and only the `updateNotScheduled: false` half of the churn argument survives. Verified against `e98549a5`: no `matchUpdateTypes: ["major"]` entry in `packageRules`, `schedule: ["* * * * 1,4"]` and `updateNotScheduled: false` both present.

Your reason for not filing is better than the ticket would have been. A number chosen against four weeks of unmeasurable input is a worse artefact than a note, and the record survives where it needs to: the entry, the `branchConcurrentLimit - prHourlyLimit` rule and the watch symptom are all in the PR body, which outlives this thread.

Which is why I want to close it rather than hold it.

`main` has `required_conversation_resolution` enabled, so an unresolved thread is a merge gate rather than just an open conversation. Holding this one for four weeks deadlocks the thing the plan depends on: the backlog cannot drain until this config is live on `main`, the config cannot go live while the thread is open, so the four weeks of runs never start accruing. Same treatment you already gave the `RENOVATE_APP_ID` thread, which is resolved with its post-merge step living in the PR body.

Resolving it on that basis. Nothing here is unsettled, and the watch is in the one place that survives the merge.

sebald (approved):
Approving. All four of my findings from `0898abc` are integrated, and I re-verified each against `e98549a5` rather than reading the replies.

The cadence fix is better than either option I offered. Moving the routine window out of the cron and into `schedule: ["* * * * 1,4"]` is what lets the workflow run daily without producing the ~35 PRs a week a plain daily cron would, and `updateNotScheduled: false` keeps the five extra runs from rebasing standing branches. I checked the schedule string is the form Renovate's validator accepts, that `vulnerabilityAlerts` carries its own `schedule: ["at any time"]` so security work ignores the window, and that the concurrency group expression is now the exact boolean negation of the job's `if`. Your correction to my latency figure is right: Thursday 00:00 to Monday 00:00 is 4 days, so the worst case was 3d23h rather than ~3.5d.

`permissions: { contents: read }` over `permissions: {}` is the right call for the reason you gave. `actions/checkout@v6` authenticates with `github.token`, so an empty block leaves the clone resting on anonymous-read fallback, and one read scope is the blast radius I was actually asking for.

**One thing to fix before this merges, in the body rather than the diff.** "Done outside this diff" item 2 states that `marigold-renovate` was added to `bypass_pull_request_allowances.apps` and that unattended automerge is "now live". Neither is true of the current state:

```
bypass_pull_request_allowances: { apps: [], teams: [], users: [aromko] }
```

Item 1 of that section checks out exactly as written: 20 required contexts, the 3 Vercel ones plus 17 promoted, with `enforce_admins: true`.

#5859's body already names this and even names the correction owed here, so nothing is unknown. It is worth making anyway, because this section is the durable record of a branch-protection change that is hard to reconstruct later, and as written it claims a security-relevant allowance exists when it does not. Leaving the diff alone.

Worth stating the consequence out loud for whoever merges first: with the bypass gone, this PR raises the routine rate to 10 PRs a week and nothing can merge them unattended until #5859 lands. That is safe, not broken, but the queue grows in the meantime.

The `RENOVATE_APP_ID` deletion is still outstanding, as expected. Both it and `RENOVATE_APP_CLIENT_ID` are live repo variables right now, and the workflow reads only the client ID after this lands, so the post-merge step in the body is the correct home for it.

PR #5853, opened 2026-09-17
chore(DST-1779): remove redundant comments and normalize duplicated prop docs

sarahgm:
# Description

Acts on the repo-wide comment audit in DST-1779 (1,640 files, 5,962 logical comments). Nine commits, ordered by risk.

**Fixes one published defect.** 45 `children` prop descriptions in `packages/components` carried three phrasings of the same sentence. They render in the docs prop tables through `AutoTypeTable`, so marigold-ui.io was showing three descriptions for one prop. They now read `Children of the component.` House style carries no leading article, and `AppShell` and `BaseLoader` keep their own wording because it says more.

**Removes comments that restate the code.** 19 bare `// Arrange` / `// Act` / `// Assert` labels, where the convention is the blank line that the other ~230 test files use. 4 ASCII banner boxes trimmed to the house `// Label` plus rule form. Verbose blocks in `packages/components` condensed by a third to a half. A final pass over `themes/`, `docs/` and `packages/cli` removed 15 more line comments and JSDoc blocks whose every clause was recoverable from the line beneath.

**Condenses the theme style comments.** Commits 5 to 9, from `99070fb24` onward. The densest `*.styles.ts` blocks in `theme-rui` ran to 15 lines or more, and are now capped at four and reworded to lead with the constraint rather than the narrative. This is the part of the diff that reaches past the ticket's scoped list, and the part worth reading closely. The rule applied was to keep every measurement, ratio, specificity figure and ticket reference, and to drop only restatement and re-explanation. Files touched: Accordion, Calendar, Checkbox, DateRangePicker, Dialog, Input, Label, Menu, Modal, Radio, SectionMessage, SegmentedControl, Select, SelectList, Sidebar, Slider, Table, Tabs, ToggleButton, Tooltip.

**Corrects four comments that were wrong rather than redundant.** `Aspect.test.tsx` blamed Jest for not supporting `aspect-ratio`, but unit tests load no stylesheet at all, so no utility resolves. `react.test.tsx` said its test existed to satisfy Jest, and claimed `<Button as={Link} />` passes `isDisabled`, which it does not. `packages/types` is excluded in `vitest.config.shared.ts`, so that file is verified by `tsc`. `simplify-jsx.ts` had a bare `//` in an empty catch, now stating that an unparseable expression falls through to the node being dropped.

**Normalizes 36 `Connector Dots` / `Connector Lines` anatomy labels** in `docs/content` onto the lowercase form the majority already used. These are JSX comments, so no rendered SVG changes.

Every comment carrying a reason the code cannot state was kept, including the DSTSUP-261 cause, the detached-fragment reason for the hidden-tree probe, the StrictMode reasoning in `usePageFocus` with both DST-1492 TODOs, the `ariaHideOutside` failure chain, and the RAC focus-proxy and roving-tabindex notes.

Two sweeps were deliberately left alone, because a partial removal is worse than none: the 228 `// Props` / `// Component` section dividers across 105 files in `packages/*/src`, and the 7 barrel section headers in the two `index.ts` files. The 8 lowercase variants inside `packages/components/src/Tabs/` are normalized to the capitalized form the other 150 already use, so the dividers are consistent without being removed.

The inert `@internal` tag on `HiddenSelection` that the ticket lists for removal is kept on purpose. The tag documents non-public API for anyone reading the file, not only for the Insights scanner, so its being absent from the barrel is not a reason to strip it. The ticket's scope line is stale on that point, and so was the `CLAUDE.md` bullet that read "do not add it to component files": raised in review, and amended in `a1c22ea13` so the rule states the narrower constraint it actually encodes (the scanner reads the barrel source, so the barrel line is the one that has to carry the tag) instead of a prohibition the code has always contradicted.

No behavior changes. No component renders differently, so there is nothing for Chromatic to catch. Across the 840 changed lines, the only non-comment edits are the 36 anatomy label cases, one `mdxOptions: {}` removal in `docs/source.config.ts` (the key only ever held the comment, dropped on review), and one deleted JSX comment in `packages/types/src/react.test.tsx`.

Closes DST-1779

## Screenshots / Preview

The only user-visible delta is the docs prop tables, which now show one `children` description instead of three. The anatomy label change is a source comment and never reaches the rendered page.

| Before | After |
| ------ | ----- |
|        |       |

## Test Instructions

1. `pnpm install && pnpm typecheck:only && pnpm lint` (all pass locally).
2. `pnpm test` to confirm the four corrected test comments did not come with behavior changes. The assertions are untouched.
3. Open the docs preview and check a component prop table, for example Button or Card, to confirm `children` reads `Children of the component.`
4. To confirm the diff is comment-only, run
   `git diff origin/main...HEAD -U0 | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]\s*(//|/\*|\*|\*/|$)'`
   The only output is the 36 anatomy label pairs, the `mdxOptions: {}` collapse and the one removed JSX comment.

## Breaking Changes

No

## Checklist


themes/theme-rui/src/components/Checkbox.styles.ts
```diff
@@ -13,16 +13,10 @@ export const Checkbox: ThemeComponent<'Checkbox'> = {
...
-      // `focus-visible` is in the chain for that reason and not for symmetry:
-      // this rule lands at (0,5,0) against (0,2,0) for the focus-visible border
-      // above, so without it hover wins on a focused checkbox and the border
-      // never flips to --color-ring. That leaves the halo as the only thing
-      // separating focused from hovered, and the halo alone is 2.08:1.
-      // `disabled` is absent by contrast: its rule carries `!important`.
+      // Matches ToggleButton's hover step, not its value: this slot idles at +0.06.
+      // Each `:not()` outranks a state rule above, so removing one lets hover
+      // repaint it. `focus-visible` is in the chain because this lands at (0,5,0)
+      // against the focus border's (0,2,0); `disabled` isn't, its rule is `!`.
```
sebald: ⚠️ Three figures leave this block, and the rewrite breaks its own sentence on the way out.

- **`+0.12` is gone**, so "not its value" now has no value to contrast with. The comparison no longer parses.
- **`2.08:1` is gone**, along with the consequence it belonged to: without `focus-visible` in the chain, hover wins on a focused checkbox, the border never flips to `--color-ring`, and the halo alone is all that separates focused from hovered. That clause is the reason `:not(:focus-visible)` can never be removed, and the specificity figures alone don't say it.
- **`` `!` `` should be `` `!important` ``** — nine characters for a token that otherwise reads as a typo.

```suggestion
      // Matches ToggleButton's hover step, not its value: this slot idles at
      // +0.06, so its +0.12 would be half the step. Each `:not()` outranks a
      // state rule above, so removing one lets hover repaint it.
      // `focus-visible` is in the chain because this lands at (0,5,0) against
      // the focus border's (0,2,0): without it hover wins on a focused checkbox,
      // the border never flips to --color-ring, and the halo alone (2.08:1) is
      // all that separates focused from hovered. `disabled` isn't, its rule
      // carries `!important`.
```

themes/theme-rui/src/components/SegmentedControl.styles.ts
```diff
@@ -105,44 +84,18 @@ export const SegmentedControl: ThemeComponent<'SegmentedControl'> = {
...
-        // current tokens. Derived from `--control-alpha` rather than hard-coded, so
-        // retuning the track can't silently detune the rim (it did: the old
-        // hand-tuned -0.08 was measured against the opaque charcoal-300 track and
-        // left the rim at an effective 0.31 against 0.26 on a field). Verified on
-        // rendered pixels: #c0bfbf on `surface`, matching an Input to within a
-        // rounding unit on all four grounds. ui-state-focus overrides it on focus.
+        // Flat `ui-control` thumb, inset 3px (not 4px: 30px tall, still clears the
+        // focus ring, which is drawn here rather than on the cell). Rim alpha is
+        // solved from `--control-alpha`: an outset rim composites over the track,
+        // so a hard-coded step detunes when the track is retuned (it did, at -0.08).
```
sebald: ⚠️ This block drops seven figures, and two of them are the ones nobody can reconstruct.

- **The composite formula and `0.119`.** The `calc()` on the line below is the *inverse* of `a + (1 - a) * x`. It is the single least self-explanatory expression in this file, and the comment now says only "solved from `--control-alpha`" — the derivation and its result are both gone.
- **The verification record.** `#c0bfbf` on `surface`, matching an Input to within a rounding unit on all four grounds. That line is what tells the next person the maths was checked against rendered pixels rather than reasoned on paper, and it is the most expensive thing here to re-establish.
- The "why shared `ui-state-focus` and not a hand-rolled outline" note also goes, which existed to stop exactly that regression (the outline alone reads noticeably lighter than every other control).

Same rule as the rim applies to the radius block at line 15, which keeps `R_outer = R_inner + d` but loses the `11px` result and the outset-rim parenthetical (9px arc, 2px gap) that pre-empts the obvious objection. Worth restoring that parenthetical too.

```suggestion
        // Flat `ui-control` thumb, inset 3px (not 4px: 30px tall, still clears the
        // focus ring, which is drawn here rather than on the cell, as the shared
        // `ui-state-focus` so it matches a focused Input). Rim alpha is solved from
        // `--control-alpha`: an outset rim composites over the track, and two layers
        // of one colour stack to `a + (1 - a) * x`, so inverting that gives 0.119 at
        // the current tokens. Derived, not hard-coded, because a hard-coded step
        // detunes when the track is retuned (it did, at -0.08). Verified on rendered
        // pixels: #c0bfbf on `surface`, matching an Input on all four grounds.
```

sebald (changes requested):
## Automated Code Review

I checked the body's claims against the diff rather than taking them on trust, and they hold: the diff really is comment-only (3 non-comment lines, all the `mdxOptions: {}` collapse), `Children of the component.` lands on exactly 45, all four ticket references and both DST-1492 TODOs survive, and no `eslint-disable` / `@ts-expect-error` / `@internal` directive was touched. CI is green across 28 checks. Nice catch moving the `hyphens-auto needs a document lang` note down beside the class it constrains.

Two changes go against the PR's own rule of keeping every measurement and ratio, and both drop a figure that can't be recovered from the code. Details inline, each with a suggested restoration that still comes in at well under half the original length.

---

themes/theme-rui/src/components/Checkbox.styles.ts
```diff
@@ -13,16 +13,10 @@ export const Checkbox: ThemeComponent<'Checkbox'> = {
...
-      // +0.06, so its +0.12 would be half the step. Each `:not()` outranks a
-      // state rule above; removing one lets hover repaint that state.
-      //
-      // `focus-visible` is in the chain for that reason and not for symmetry:
-      // this rule lands at (0,5,0) against (0,2,0) for the focus-visible border
-      // above, so without it hover wins on a focused checkbox and the border
-      // never flips to --color-ring. That leaves the halo as the only thing
-      // separating focused from hovered, and the halo alone is 2.08:1.
-      // `disabled` is absent by contrast: its rule carries `!important`.
+      // Matches ToggleButton's hover step, not its value: this slot idles at +0.06.
```
aromko: ⚠️ This kept the specificity figures but dropped the consequence. The old block explained *what breaks*: without `focus-visible` in the chain, hover wins on a focused checkbox, the border never flips to `--color-ring`, and the halo alone is `2.08:1`.

That contrast ratio is the reason the token is load-bearing. The new text tells a maintainer the specificity fact but not what they'd be breaking, which is the one thing this comment exists to prevent. Worth restoring the failure mode and the ratio.

themes/theme-rui/src/components/SegmentedControl.styles.ts
```diff
@@ -105,44 +84,18 @@ export const SegmentedControl: ThemeComponent<'SegmentedControl'> = {
...
-        // `control-alpha + (1 - control-alpha) * a`, so solving that for
-        // `control-border`'s alpha gives the thinned value below — 0.119 at the
-        // current tokens. Derived from `--control-alpha` rather than hard-coded, so
-        // retuning the track can't silently detune the rim (it did: the old
-        // hand-tuned -0.08 was measured against the opaque charcoal-300 track and
-        // left the rim at an effective 0.31 against 0.26 on a field). Verified on
-        // rendered pixels: #c0bfbf on `surface`, matching an Input to within a
-        // rounding unit on all four grounds. ui-state-focus overrides it on focus.
+        // Flat `ui-control` thumb, inset 3px (not 4px: 30px tall, still clears the
+        // focus ring, which is drawn here rather than on the cell). Rim alpha is
```
aromko: ⚠️ Same pattern as `Checkbox.styles.ts`, and the more costly instance. The rim alpha is a derived value, and the numbers that let anyone re-verify the `oklch(…calc(…))` expression are gone: `0.119 at the current tokens`, `#c0bfbf on surface`, `0.31 against 0.26 on a field`.

The `-0.08` history survived, but that's the anecdote. The measurements are what make the formula checkable after the next token retune. The PR description sets the rule "keep every measurement, ratio, specificity figure" — this block and Checkbox are where it wasn't held.

docs/source.config.ts
```diff
@@ -62,8 +60,6 @@ export const docs = defineDocs({
 });
 
 export default defineConfig({
-  mdxOptions: {
-    // MDX options
-  },
+  mdxOptions: {},
```
aromko: 💡 `mdxOptions: {}` only ever existed to hold the comment, and `mdxOptions` appears nowhere else in `docs/`. Drop the key.

```suggestion
```

themes/theme-rui/src/components/Sidebar.styles.ts
```diff
@@ -187,38 +173,30 @@ export const Sidebar: ThemeComponent<'Sidebar'> = {
...
-  // otherwise grow a horizontal scrollbar. ui-scroll-mask-y owns the vertical
-  // overflow (hence no overflow-y here) and fades over the scrollbar instead
-  // of replacing it. scroll-py reads its --sc-mask-width so they can't drift,
-  // and scroll-smooth keeps focus from jumping to a tile below the fold.
-  // No overscroll containment, unlike the ui-scroll-mask-x consumers: theirs
-  // guards the browser's back/forward swipe, which has no block-axis twin, and
-  // the single column's nav doesn't contain either.
+  // overflow-x-hidden: fixed-width labels would otherwise grow a horizontal
+  // scrollbar. ui-scroll-mask-y owns the vertical overflow (hence no overflow-y)
+  // and scroll-py reads its --sc-mask-width so the two can't drift. No overscroll
```
aromko: 💡 `motion-safe:scroll-smooth` is still in the class list below but lost its reason ("keeps focus from jumping to a tile below the fold"). The *why* for a class that's still there is the kind that's worth the line.

themes/theme-rui/src/components/SelectList.styles.ts
```diff
@@ -96,8 +88,6 @@ export const SelectList: ThemeComponent<'SelectList'> = {
           'group-orientation-horizontal/list:@max-[40rem]/selectlist:not-last:border-b',
         ],
         bordered: [
-          // Each item is a control: a dense control-border ring, flat (the list
-          // pads itself so the ring isn't clipped).
           'ui-control min-h-14',
```
aromko, line 93: 💡 "the list pads itself so the ring isn't clipped" was a cross-slot constraint, not recoverable from `ui-control min-h-14`. Someone tightening the list padding later has no signal that this slot depends on it.

packages/components/src/Page/usePageFocus.ts
```diff
@@ -33,47 +26,35 @@ import { usePageContext } from './Context';
...
  *   return null;
  * };
- *
- * <RouterProvider navigate={navigate}>
- *   <Page>
- *     <PageFocus pathname={location.pathname} />
- *     <Outlet />
- *   </Page>
- * </RouterProvider>
+ * // <Page><PageFocus pathname={location.pathname} /><Outlet /></Page>
```
aromko: 💡 The usage example is now a commented-out line inside the ```tsx fence, so it renders greyed out on IDE hover and reads as "disabled" rather than "this is how you call it". Either make it real JSX or drop the line.

aromko (commented):
One finding has no line in this diff to attach to, so it lands here:

⚠️ **`packages/components/src/HiddenSelection/HiddenSelection.tsx:27`** — the `@internal` tag is kept against both DST-1779's scope line and CLAUDE.md ("The Insights scanner reads the tag from the barrel source, so do not add it to component files"). The reasoning in the PR description is fair, but as it stands the file contradicts a written rule with the justification living only in a PR body. Either drop the tag, or amend that CLAUDE.md bullet in this PR so the rule and the code agree. A third reader hitting this file shouldn't have to find PR #5853 to know it's deliberate.

---

themes/theme-rui/src/components/Checkbox.styles.ts
sarahgm: Applied your suggestion verbatim in `3d96e3b57`. `+0.12`, the failure mode and `2.08:1` are back, and `!` is spelled out as `!important`.

You were right that the comparison stopped parsing. "not its value" with no value left in the sentence was a rewrite that broke on the way out, not a trim.

themes/theme-rui/src/components/SegmentedControl.styles.ts
sarahgm: Applied in `3d96e3b57`, with two additions on top of your suggestion.

The rim block takes your text plus the effective figure from the `-0.08` era, which @aromko flagged separately: `(it did: -0.08 left the rim at an effective 0.31 against 0.26 on a field)`. That is the number that makes the anecdote checkable rather than just historical.

The radius block regains `so 11px` and the outset-rim parenthetical (9px arc, 2px gap), as you asked.

On the general point: the rim is the block where "keep every measurement" was stated and then not held. The composite formula and `#c0bfbf` were the two nobody could re-derive, and dropping the verification record was the expensive half of that.

themes/theme-rui/src/components/Checkbox.styles.ts
sarahgm: Fixed in `3d96e3b57`. The failure mode and `2.08:1` are back, so the block now says what breaks and not only where it ranks.

Same fix as @sebald's thread on this block, whose suggested wording is what landed.

themes/theme-rui/src/components/SegmentedControl.styles.ts
sarahgm: Fixed in `3d96e3b57`. All three are back: `0.119 at the current tokens` with the `a + (1 - a) * x` derivation it comes from, `#c0bfbf on surface` as the rendered-pixel check, and `0.31 against 0.26 on a field` attached to the `-0.08` history so that clause carries a number instead of an anecdote.

Agreed on which half mattered. The history is the cheap part to reconstruct, the measurements are not.

docs/source.config.ts
sarahgm: Dropped in `3d96e3b57`. `defineConfig` takes `mdxOptions` as optional, and `pnpm typecheck:only` is clean without it.

themes/theme-rui/src/components/Sidebar.styles.ts
sarahgm: Restored in `3d96e3b57`, and the sentence around it repunctuated so the clause sits with the class rather than trailing the mask note.

themes/theme-rui/src/components/SelectList.styles.ts
sarahgm: Restored in `3d96e3b57`, worded so the constraint points at the thing that could break it:

```
// Dense control-border ring, flat. The list pads itself so this ring
// isn't clipped: tightening that padding has to keep the room.
```

packages/components/src/Page/usePageFocus.ts
sarahgm: Made it real JSX in `3d96e3b57`:

```tsx
<Page>
  <PageFocus pathname={location.pathname} />
  <Outlet />
</Page>
```

Still shorter than the original, which wrapped the whole thing in a `RouterProvider` that the example does not need.

sarahgm:
> Either drop the tag, or amend that CLAUDE.md bullet in this PR so the rule and the code agree.

Took the second option, in `a1c22ea13`. The tag stays and the rule now says why it is allowed to.

The constraint the old wording encoded is real, but narrower than the prohibition it grew into: the Insights scanner reads the barrel source, so the barrel line is the one that has to carry the tag, and tagging a component file instead does not register the export. That is a reason the component-file tag cannot *substitute* for the barrel line. It was never a reason it cannot exist. The bullet now draws that distinction and names `HiddenSelection` as the case, so a third reader hits the answer in `CLAUDE.md` rather than in this PR.

`.memory/CONTEXT.md` → Internal export is left alone. It defines the barrel-line term and is accurate as written.

sebald (approved):
Verified all eight threads against `a1c22ea`, resolving them.

- **Checkbox**: `+0.12`, `(2.08:1)` and `!important` are back, applied verbatim.
- **SegmentedControl**: `a + (1 - a) * x`, `0.119` and the `#c0bfbf on surface` verification record restored, and the radius block regained `so 11px` plus the (9px arc, 2px gap) parenthetical. The `-0.08` clause now carries `0.31 against 0.26 on a field`, which is better than what I suggested.
- **source.config.ts**, **Sidebar**, **SelectList**, **usePageFocus**: all confirmed in the file.

One note for the record, not a blocker: `a1c22ea` rewrites the barrel-export bullet in `CLAUDE.md` to allow a documenting `@internal` in a component file's own JSDoc. No thread asked for that, but the example it cites checks out, `HiddenSelection` carries the tag in its JSDoc and is not in the barrel.

PR #5858, opened 2026-09-18
fix(DST-1719): apply a string `thumbLabels` to the whole thumb

sarahgm:
# Description

`SliderProps.thumbLabels` is typed `string | [string, string]`, but the component indexed it positionally without normalising the string case. For a string, `thumbLabels?.[0]` returns the **character** at index 0, so a single-thumb slider written the way the type invites named its thumb `aria-label="A"`:

```tsx
<Slider label="Gebühren-Splitting" thumbLabels="Anteil Reservix" />
```

`thumbLabels` is now normalised to a list the same way the sibling `name` prop already was twelve lines earlier. The tuple form is unchanged. This matters because for a single-thumb slider the thumb label is the only lever for naming what the value refers to: `aria-valuetext` cannot be overridden from outside, so a wrong thumb name has no fallback.

The docs only described `thumbLabels` under "Ranges", so the single-thumb string form was undocumented. A new "Labelling the thumbs" section covers both forms.

No visual change: the diff is `aria-label` wiring, tests and prose only.

Closes DST-1719

## Test Instructions

1. `pnpm test:unit packages/components/src/Slider/Slider.test.tsx` — 8 tests, three of them new.
2. In Storybook, render `<Slider label="Gebühren-Splitting" thumbLabels="Anteil Reservix" />` and inspect the thumb in the browser's accessibility tree. The accessible name should read `Anteil Reservix Gebühren-Splitting`, not `A Gebühren-Splitting`.
3. Confirm `Components/Slider/MultipleThumbs` still names its thumbs `start` and `end`.
4. `pnpm start` and open `/components/form/slider` to read the new "Labelling the thumbs" section.

## Breaking Changes

No. The prop type is unchanged and the tuple form behaves exactly as before. Only the previously broken string form changes, from a one-character name to the intended one.

## Checklist

---


packages/components/src/Slider/Slider.tsx
```diff
@@ -72,6 +74,9 @@ const _Slider = <T extends number | number[]>({
   });
 
   const names = Array.isArray(name) ? name : [name];
+  const thumbLabelsList = Array.isArray(thumbLabels)
```
aromko, line 77: 🔴 **Visual Regression Tests — not run**

This PR changes UI-affecting files but the `Visual-Regression-Tests` workflow has not run for this branch, and there is no `UI Tests` status on `74f6d53`.

Worth saying out loud: this diff is `aria-label` and `name` wiring only, so a pixel delta would be surprising. The run is here to prove that, not because anything looks risky.

docs/content/components/form/slider/index.mdx
```diff
@@ -36,10 +36,16 @@ A slider is useful when selecting a value or range within a continuous or evenly
...
 Use the range feature of a slider when users need to select both a minimum and maximum value within the same scale, such as defining a price range, date span, or acceptable measurement limits. This approach works best when the range boundaries are equally important and the relationship between them is immediately clear in a single, continuous control.
 
-When multiple thumbs are rendered, each thumb should have an `aria-label`, which is provided via the `thumbLabels` prop.
+Name both thumbs so users can tell the two ends of the range apart.
 
 <ComponentDemo name="slider-range" />
 
+### Labelling the thumbs
+
+Give a thumb its own name whenever the slider's label does not already say what that thumb controls. Screen readers announce the thumb name ahead of the slider label, so a slider labelled "Fee split" whose thumb is named "Organiser share" tells the user exactly which side of the split they are about to move.
```
aromko, line 45: ⚠️ Two shipped examples break this rule as of this PR, and the fix changes what they announce:

- `docs/app/(examples)/examples/filter/FilterBar.tsx:175`
- `docs/content/patterns/user-input/filter/filter-group-sections.demo.tsx:37`

Both pass `thumbLabels="price"` to a single-thumb slider labelled `"Max. Price"`. Before this fix the thumb was named `"p"`, so the accessible name read `"p Max. Price"`. After it, `"price Max. Price"`, which is exactly the redundancy the sentence above tells people to avoid: the slider label already says what the thumb controls.

Dropping `thumbLabels` from both keeps the new guidance and the examples in agreement, and it means the first two places a reader meets the prop are not counter-examples.

docs/content/components/form/slider/index.mdx
```diff
@@ -36,10 +36,16 @@ A slider is useful when selecting a value or range within a continuous or evenly
...
+
+This matters most when one slider stands for two related quantities, or when a page holds several sliders adjusting the same kind of value. Set the names with the `thumbLabels` prop: one name for a single-thumb slider, one name per thumb for a range.
```
aromko, line 47: ⚠️ Every other `###` section on this page closes with a `<ComponentDemo>`: `slider-range`, `slider-granularity`, `slider-format`, `slider-min-max`, `slider-form`. "Labelling the thumbs" is the only one without, and it is the section whose whole reason for existing is that the single-thumb string form was undocumented. A reader still has no copyable example of it anywhere in the docs.

A `slider-thumb-labels.demo.tsx` showing both forms would close this, and the single-thumb half of it doubles as the story the fix currently lacks: the ticket names "every story and test in the repo uses the tuple form" as the reason this went unnoticed, and stories are still tuple-only, so the browser test run and Chromatic never exercise the string path.

aromko (commented):
Automated review via the `review-pr` skill. Findings are inline.

packages/components/src/Slider/Slider.tsx
```diff
@@ -135,8 +140,8 @@ const _Slider = <T extends number | number[]>({
                 className={cn('top-1/2 cursor-pointer', classNames.thumb)}
                 key={i}
                 index={i}
-                aria-label={thumbLabels?.[i]}
-                name={names?.[i]}
+                aria-label={thumbLabelsList[i]}
+                name={names[i]}
```
sebald, line 144: 💡 Out of scope for DST-1719, but this line is the reason to mention it: `name` has the same shape as `thumbLabels` and the same gap, and normalising `thumbLabels` to match it inherits that rather than closing it.

`names = Array.isArray(name) ? name : [name]` means a string `name` on a **range** slider lands on thumb 0 and leaves thumb 1 as `name={undefined}`:

```tsx
<Slider defaultValue={[20, 40]} label="Fee" name="fee" />
// thumb 0 -> name="fee", thumb 1 -> no name attribute
```

That one is worse than the labelling case this PR fixes. A thumb with no `aria-label` still falls back to the slider's own label, so the user gets a usable if redundant name. An input with no `name` is simply absent from the submitted form data, silently, with nothing in the UI to show for it.

Both props could be made safe by the type rather than at runtime, since `_Slider` is already generic over `T extends number | number[]`:

```ts
name?: T extends number[] ? [string, string] : string;
thumbLabels?: T extends number[] ? [string, string] : string;
```

Worth checking inference before committing to it: `T` is inferred from `value` / `defaultValue` inside `rest`, so a slider given neither would resolve the conditional to the `string` branch. That is the right default for a single-thumb slider, but it is an assumption to verify rather than assume.

Happy for this to become its own ticket instead, since the form-data half is a separate bug from the one you are fixing here.

sebald (commented):
One observation on the sibling prop, out of scope for this ticket. Not a blocker, and it doesn't overlap anything @aromko has already raised.

packages/components/src/Slider/Slider.tsx
sarahgm: Triggered: [run 35336774512](https://github.com/marigold-ui/marigold/actions/runs/35336774512), against `d2e7b578e`.

Two things moved since you flagged this. The branch has three more commits, and the new `ThumbLabels` story opts out of Chromatic at story level, so this should come back with no visual changes and no new baselines to accept rather than one of each.

Leaving the thread open until the run reports.

docs/content/components/form/slider/index.mdx
sarahgm: Fixed in `0d10b8b51`. Dropped `thumbLabels` from both.

One thing I kept: `name="price"` on the `FilterBar` slider. Your objection was about the announced name, and `name` is the form field key, so removing it would have changed what the example submits rather than what it announces. `filter-group-sections.demo.tsx` had no `name` to keep.

docs/content/components/form/slider/index.mdx
sarahgm: Fixed in `8389d92eb`. Added `slider-thumb-labels.demo.tsx` showing both forms and referenced it from the section, so it is no longer the only `###` on the page without a demo.

On the story half, you were right that nothing exercised the string path. There is now a `ThumbLabels` story with a `.test()` asserting the accessible name of the single thumb and of each range thumb, so it reaches the browser run.

It deliberately does not reach Chromatic, though. `thumbLabels` only sets `aria-label`, so a pixel diff cannot catch a regression in the thing the story exists to prove, and the shapes it renders are already snapshotted by `ValueFormatting` and `MultipleThumbs`. That is the same reasoning behind `MultiThumbsControlled` and `MultiThumbsDisabled` opting out. Done in `d2e7b578e`.

packages/components/src/Slider/Slider.tsx
sarahgm: Agreed, and I will take the separate ticket rather than fold it in here.

Confirmed the gap is real: `names = Array.isArray(name) ? name : [name]` leaves thumb 1 with `name={undefined}` on a range slider, and unlike the labelling case there is no fallback to soften it, so the value is simply missing from the submitted form data.

Keeping it out of this PR because DST-1719 is scoped to the `aria-label` path, and your type-level fix changes the public type of both props, which wants its own review and its own changeset. Your inference caveat is worth carrying into that ticket: a slider given neither `value` nor `defaultValue` resolves `T` to the `string` branch.

Leaving this thread open until the follow-up ticket exists.

packages/components/src/Slider/Slider.tsx
sarahgm: Run [35336774512](https://github.com/marigold-ui/marigold/actions/runs/35336774512) came back green against `d2e7b578e`: no visual changes and no new baselines to accept, which is what an `aria-label` and `name` wiring diff should produce. `UI Tests: marigold_2026` is green on [build #345](https://www.chromatic.com/build?appId=6a0c09c122d3a0a4a6da5d41&number=345), and the Slack notification step was skipped, which is what a run with nothing to report looks like.

Resolving.

packages/components/src/Slider/Slider.tsx
sebald: Agreed on the split, and resolving this rather than holding the PR for the ticket.

To be clear about what is being resolved: the gap itself is still open, and there is no follow-up ticket yet. `names = Array.isArray(name) ? name : [name]` with `name={names[i]}` is unchanged on `d2e7b578e`, so a string `name` on a range slider still leaves thumb 1 without a `name` attribute and silently out of the submitted form data.

The asymmetry this PR introduces is worth carrying into that ticket verbatim, because it is now pinned by a test on one side and nothing on the other. `Slider.test.tsx` asserts the label fallback:

```tsx
expect(end).toHaveAccessibleName('Range');
```

There is no equivalent on the form-data side, because there is no fallback to assert. That is the whole difference between the two props: an unnamed thumb degrades to something usable, an unnamed input just disappears.

Your scoping reason holds, though. The type-level fix changes the public type of both props, so it wants its own review and its own changeset rather than riding along with an `aria-label` fix.

Filing the ticket is on me, not on you. Nothing here blocks this PR.

sebald (approved):
Approving. Every point raised on this PR is either fixed or deliberately and correctly scoped out, and I verified each against `d2e7b578e` rather than taking the replies at face value.

**The fix itself.** Normalising `thumbLabels` the same way `name` already was is the right shape, and the three new unit tests pin all three cases that matter: the string on a single thumb, the tuple across a range, and the string that must *not* spread across a range. That last one is the test I would have asked for if it were missing, and pinning the fallback (`toHaveAccessibleName('Range')`) rather than asserting an absent attribute is the more durable assertion.

**On the two judgement calls that went against the review, both hold up.**

Keeping `name="price"` on the `FilterBar` slider while dropping `thumbLabels="price"` is correct. `name` is the form field key and is not an accessible-name source, so removing it would have changed what the example submits without changing a word of what it announces.

Opting the `ThumbLabels` story out of Chromatic is also correct, and the supporting claims check out: `MultiThumbsControlled` and `MultiThumbsDisabled` opt out at story level the same way, and the two shapes the story renders are already snapshotted by `ValueFormatting` (single thumb, percent) and `MultipleThumbs` (range), neither of which opts out. Since `thumbLabels` only writes `aria-label`, a pixel diff could not catch a regression in the thing the story exists to prove. The browser test is the right guard, and it is there.

**Non-blocking, tracked separately.** The sibling `name` prop has the same shape and the same gap, and this PR inherits it rather than closing it: a string `name` on a range slider leaves thumb 1 with no `name` attribute and silently out of the submitted form data. That is a worse failure than the labelling bug fixed here, because an unnamed thumb still falls back to the slider's label while an unnamed input just disappears. Correctly kept out of scope, since the type-level fix changes the public type of both props. Filing that ticket is on me.

CI is green across all 30 checks, including `UI Tests: marigold_2026` on build #345 with 915 tests unchanged, which is what an `aria-label` wiring diff should produce.

PR #5859, opened 2026-09-18
ci(DST-1748): approve automergeable Renovate PRs

sarahgm:
# Description

Renovate PRs that are green and marked automergeable still wait for a human. `main` requires one approving review, and GitHub's native auto-merge does not use a bypass allowance. The bypass only applies when the app performs the merge itself, which happens only while a Renovate run is in flight.

#5854 is the evidence. Its last required check went green at 06:50:19 and it merged at 08:16:21, 86 minutes later, inside the window of a `workflow_dispatch` run. Auto-merge had been queued on it the whole time and never fired. Of the last 60 merged Renovate PRs it is the only one without a human approval, and 51 of the other 59 were approved by the same person.

This adds a workflow that approves as `marigold-approver`, a GitHub App created for this and installed on this repository alone, holding `Pull requests: Read and write` and nothing else. Its credentials are the `APPROVER_APP_CLIENT_ID` variable and the `APPROVER_APP_PRIVATE_KEY` secret, both already in place (DST-1782). Renovate cannot approve its own PR, since GitHub refuses a review from the PR's author, so the approval has to come from a second identity.

`marigold-renovate` has since been removed from `bypass_pull_request_allowances.apps`, which closed a real hole: while it stood, Renovate could merge its own PR with no approving review at all whenever a run was in flight. It also means unattended merging is currently zero, and this PR is what restores it.

**What decides which PRs get approved.** The job gates on `autoMergeRequest`, which is non-null exactly when Renovate queued auto-merge, and Renovate queues it only for what `renovate.json` allows: `minor` and `patch` on packages whose current range is not `0.x`, minus the `vale-cli/vale` carve-out. Policy stays in one file, and this workflow needs no edit when it changes. An earlier revision read the `🚦 **Automerge**: Enabled.` line out of the PR body instead. That was wrong in both directions: Renovate pastes a dependency's release notes into the body, so a changelog could forge the marker, and a future Renovate template change would silently skip the job with no failure to notice.

Three further gates: the PR author has to be `marigold-renovate[bot]`, the head branch has to live in this repo rather than a fork, and it has to be named `renovate/*`.

**Not stacking approvals.** `dismiss_stale_reviews` is off, so an approval survives the rebases Renovate does. The job counts reviews by the app in **any** state, dismissed included, and exits if it finds one. Dismissing its approval is how a maintainer stops a queued merge, and counting only live approvals would have re-approved on the next rebase and undone that.

The job takes a per-PR concurrency group so overlapping `opened` and `synchronize` runs cannot both read zero reviews and both approve, and the step runs under `set -euo pipefail` so a failed `gh` call aborts rather than reading as zero.

No changeset: nothing published changes, the diff is one CI workflow. No UI, so screenshots and VRT are not applicable.

Part of DST-1748. Related: DST-1782 for the app itself, and #5852, whose "Done outside this diff" section records unattended automerge as already working on the strength of the bypass allowance. That section needs the correction this PR's first paragraph makes.

## Why not `platformAutomerge: false`

The one-line alternative. Renovate would stop delegating to GitHub's auto-merge and merge each PR itself, under the bypass allowance it used to have, with no second app and no synthetic approval in the review timeline.

It was rejected for two reasons. Merges would only happen while a run is in flight, so they batch to the cron rather than landing minutes after checks pass, weekly today and daily once #5852 is in. And it depends on the bypass allowance, which has now been removed deliberately: a PR merging with no approving review on it is the thing that made a red suite mergeable before #5852 promoted the 17 checks to required.

## Test Instructions

1. On this PR, confirm the `Renovate auto-approve` workflow appears and its job is **skipped**. The author gate is false here, which verifies the YAML parses and the `if` holds.
2. After merge, tick a rate-limited box on the [Dependency Dashboard](https://github.com/marigold-ui/marigold/issues/5292) and confirm the minor or patch PR it opens is approved by `marigold-approver` within about a minute, then merges by itself once all 20 required checks are green.
3. Confirm a major, or the next `vale-cli/vale` bump, gets no approval. The job should run, log `Renovate did not queue auto-merge here, leaving it for a human.` and exit 0.
4. Force a rebase of an already approved Renovate PR and confirm no second approval appears.
5. Dismiss the app's approval on a queued PR, let Renovate rebase it, and confirm it is **not** re-approved.
6. The four PRs open as of writing (#5846, #5855, #5856, #5857) predate the workflow and need one manual approval each, or a Renovate rebase to reach `synchronize`.

## Breaking Changes

No

## Checklist


sebald (commented):
Reviewed `f2a6a70`. The design is right and most of this is solid. One finding I would fix before merge, plus four smaller things.

## What I checked rather than took on trust

Every verifiable claim in the description holds. `main` carries `dismiss_stale_reviews: false`, `require_last_push_approval: false`, `required_approving_review_count: 1` and no code-owner requirement, so an approval genuinely does survive Renovate's rebases. `marigold-renovate[bot]` is the correct `pull_request.user.login` (confirmed on #5857). `APPROVER_APP_CLIENT_ID` is in place. #5854 really does carry zero reviews while #5846, #5855, #5856 and #5857 are all approved by the same person, exactly as the evidence paragraph says.

The two subtle decisions are both right. Gating on `autoMergeRequest` rather than parsing `🚦 **Automerge**: Enabled.` out of the body is correct for both reasons you give. Counting reviews in **any** state so a dismissal is durable is the kind of thing that gets "simplified" into a bug later, and the comment above it earns its place. `cancel-in-progress: false` on a per-PR group does serialize rather than drop, so the race you describe is genuinely closed. `permissions: {}` with no checkout step is the right shape.

## ⚠️ `autoMergeRequest != null` is not proof that Renovate queued it

This is the one worth acting on. The description says policy "stays in one file" and that Renovate queues auto-merge only for what `renovate.json` allows. Auto-merge is not Renovate-exclusive though: anyone with write access can press **Enable auto-merge**, or run `gh pr merge --auto`, on any PR. That includes a Renovate **major** bump, or the `vale-cli/vale` bump the config deliberately carves out.

When that happens this workflow supplies the approving review, and because the approver is a separate identity GitHub's self-approval prohibition never fires. So one person with write access can land a major dependency bump in `main` with no second pair of eyes, which is the thing `required_approving_review_count: 1` exists to prevent. It is the same shape as the hole you closed by removing the bypass allowance, reached by a different route.

The fix is one line and the field already exists. Confirmed against a live PR:

```console
$ gh pr view 5857 --repo marigold-ui/marigold --json autoMergeRequest \
    --jq '.autoMergeRequest.enabledBy.login'
app/marigold-renovate
```

```bash
enabled_by=$(gh pr view "$PR" --repo "$REPO" --json autoMergeRequest \
  --jq '.autoMergeRequest.enabledBy.login // ""')

if [ "$enabled_by" != "app/marigold-renovate" ]; then
  echo "Auto-merge here was not queued by Renovate, leaving it for a human."
  exit 0
fi
```

Worth noting the identifier differs by API: `github.event.pull_request.user.login` gives `marigold-renovate[bot]` while `gh`'s JSON renders the same actor as `app/marigold-renovate`. Both forms are already load-bearing in this file, so it is worth a comment wherever the second one lands.

## ⚠️ A missed queue window fails silently, and after #5852 the recovery is days away

The poll gives Renovate 30 seconds. Miss it and the job logs `Renovate did not queue auto-merge here, leaving it for a human.` and exits 0, which is indistinguishable in the Actions list from the intended major-bump case.

The only recovery is the next `synchronize`. Once #5852 lands with `schedule: ["* * * * 1,4"]` and `updateNotScheduled: false`, standing branches are rebased Monday and Thursday only, so a single slow queue can leave a PR unapproved for up to four days with nothing surfacing it. It fails safe, so this is not a blocker, but the two exit paths are worth distinguishing in the log given that unattended merging is the entire point.

## 💡 Smaller things

**The `set -euo pipefail` claim is half true.** The description says a failed `gh` call "aborts rather than reading as zero". That holds for `reviewed=$(...)`, where the non-zero exit propagates. It does not hold for the poll loop: `set -e` is suppressed inside an `if` condition, so a failing `gh pr view` there yields an empty string, reads as "not queued", and exits down the human path. Same silent outcome as above, different cause.

**`test(env.APP)` is a substring regex, and that is doing real work.** `gh` renders bot logins as `app/<slug>` while this file's own `if` matches `<slug>[bot]`, so the loose match absorbs the disagreement. I could not confirm which form `reviews[].author.login` takes, since no app-authored review exists in this repo yet. A one-line comment would stop a later reader tidying it into `==`: if that match silently stopped hitting, the job would re-approve on every rebase and undo exactly the maintainer dismissal the comment above it protects.

**No `branches:` filter on the trigger.** The repo has a `Beta Release branch` ruleset, so a Renovate PR targeting a non-`main` base would also be auto-approved. Nothing does that today.

**Test step 6 is already stale.** #5846, #5855, #5856 and #5857 have since been approved by @aromko, so they no longer need the manual approval the step describes.

## Sequencing

Merge #5852 first, then this. #5852 raises the routine rate to 10 PRs a week while unattended merging sits at zero, so the gap between the two is a growing queue rather than a problem, but it is worth closing quickly. I have approved #5852, with one note: its "Done outside this diff" section still records the bypass allowance as live, and the correction your first paragraph describes has not been made yet. Verified just now:

```
bypass_pull_request_allowances: { apps: [], teams: [], users: [aromko] }
```

sarahgm:
Thanks, all six are in `40cc841`. Description updated to match.

## ⚠️ `enabledBy`

Taken as written. The job now compares `autoMergeRequest.enabledBy.login` against `app/marigold-renovate` and walks away from anything else, naming who queued it in the log. Your reading of the hole is right: with a second identity supplying the review, the self-approval prohibition never fires, so a single person could have landed a major.

Confirmed the spelling on the five Renovate PRs open right now, all `app/marigold-renovate`, and both forms carry a comment saying which API they came from.

## ⚠️ Missed queue window

Partly as asked, and I want to flag where I went a different way.

The two exits can't actually be told apart. A major bump never gets an `autoMergeRequest` at all, so "held back by `renovate.json`" and "queued after the poll gave up" are the same observation from inside the job, and any message that picks one would be guessing. So the annotation names both and says both need a human:

```
::warning::No auto-merge request after 60s. Either renovate.json holds this update
back, or Renovate queued it too late for this run. Either way it needs a human: the
only automatic retry is the next rebase, and those are scheduled twice a week.
```

A `::warning::` surfaces it in the Actions list, which was the point. It fires on every held-back PR, which I think is right rather than noisy: those genuinely are the ones waiting on a person. Poll widened 30s → 60s to shrink the residual. Happy to reconsider if you'd rather it stayed quiet for the expected case.

The cleanly separable case is the new third one, auto-merge queued by someone who isn't Renovate, and that does get its own message.

## 💡 `set -euo pipefail`

You were right and the fix is the assignment moving out of the condition:

```bash
enabled_by=$(gh pr view "$PR" --repo "$REPO" --json autoMergeRequest \
  --jq '.autoMergeRequest.enabledBy.login // ""')
if [ -n "$enabled_by" ]; then
  break
fi
```

A failed `gh` now aborts the job red instead of reading as "not queued". Worth saying that my first attempt to verify this reported a pass because the stub `gh` had `exit 1` inside a pipeline subshell and so returned 0. Once the stub failed properly, the four paths are: queued by Renovate approves, queued by a human exits 0 with a name, nothing queued exits 0 with the annotation, API failure exits 1. Plus a late-queue case that picks it up on poll 4 of 12.

## 💡 `test(env.APP)`

Comment added, spelling out that `==` against the wrong form would match nothing, re-approve on every rebase, and undo the dismissal the line above it exists to protect. Still no app-authored review in the repo to settle which form the field uses, so the loose match stays load-bearing until there is one.

## 💡 `branches:`

Added `branches: [main]`. One correction while I was in there: `main` isn't on a ruleset at all, it's classic branch protection. The only ruleset in the repo is `Beta Release`, and it targets `refs/heads/beta-release`, which no longer exists on origin. So the filter is defensive rather than closing something live.

## 💡 Stale test step

Replaced. #5846, #5855, #5856 and #5857 are all merged and approved by @aromko now. The step lists today's backlog instead, #5862 through #5866.

## Sequencing

#5852 merged on the 18th, so that's clear. I've added a dated correction block to its "Done outside this diff" section: point 2 is marked reverted, with why the bypass was the wrong instrument and the current `{apps: [], teams: [], users: [aromko]}` recorded. Point 1 stands.

sarahgm:
Correcting myself on the second ⚠️. `711ea74` changes it.

**I said the two exits could not be told apart. That was wrong.** Renovate states the reason in the PR body, and it splits cleanly:

```
5866..5850:  🚦 **Automerge**: Enabled.
5849, 5839, 5813, 5812, 5811, 5806:  🚦 **Automerge**: Disabled by config.
```

I had ruled the body out on forgery grounds, but that objection was load-bearing only where the body would have *granted the approval*. Using it to pick the wording of a log line carries none of that risk: `enabledBy` still gates the approval and cannot be forged from a changelog, so a release note that happened to carry the marker costs a misleading message and nothing else. I conflated the two uses and talked myself out of a distinction that was available the whole time.

**The blanket warning was also too noisy, which is what sent me looking again.** 6 of the last 60 Renovate PRs have no `autoMergeRequest`, so roughly one in ten would have raised an annotation, and again on every rebase. At that rate people learn to scroll past it, which costs more than the silence it was meant to fix.

So there are now three quiet exits:

| Condition | Log | Annotation |
|---|---|---|
| Queued by someone other than Renovate | names who queued it | no |
| `renovate.json` holds it back | `renovate.json holds this update back, leaving it for a human.` | no |
| Renovate promised automerge, queued nothing in 60s | the missed window | **yes** |

The match is on `Automerge**: Disabled by config` rather than on `Enabled`, so anything unrecognised falls through to the annotation and the failure direction is toward visibility. Verified against all six held-back PRs (#5849, #5839, #5813, #5812, #5811, #5806), all classified correctly, and the five exit paths including the API-failure one still behave as described in my previous comment.

Net effect on your original point: the annotation now fires only on the case you actually wanted surfaced, and the major-bump case it was drowning in is a plain log line again.

sebald (approved):
Approving `711ea74`. All six findings are in, and I checked the load-bearing ones against live data rather than taking them:

- `enabledBy.login` is `app/marigold-renovate` on #5862 and #5863, so the exact match is correct today.
- `Automerge**: Disabled by config` appears verbatim in #5849, #5839 and #5813, so the held-back classification holds.
- #5852's "Done outside this diff" now carries the dated correction, with `bypass_pull_request_allowances` recorded as `{apps: [], teams: [], users: [aromko]}`. That closes the sequencing note.
- Every check is green and the `approve` job reports **skipping**, which is test step 1 passing.

The self-correction on the second finding was the right call. A `::warning::` on roughly one PR in ten, repeating on every rebase, would have taught people to scroll past it, which costs more than the silence it replaced. Reserving it for "Renovate promised automerge and queued nothing" keeps the annotation meaning something, and matching on `Disabled by config` rather than on `Enabled` points the unrecognised case at visibility.

One residual, not blocking and not worth holding the merge for. The `==` against `app/marigold-renovate` is now the only fail-quiet path left in the file: if that field ever renders `marigold-renovate[bot]`, every PR takes the "queued by someone other than Renovate" branch, logs a plain line, and the workflow silently stops approving anything. That is the same spelling-drift argument the comment on `test(env.APP)` makes fifteen lines below, pointing the other way. If you want it covered later, the cheap version is to raise an annotation instead of a plain log when the rejected `enabled_by` contains `renovate`, so the drift surfaces the first time it happens rather than as an unexplained queue.

Merging.

PR #5860, opened 2026-09-18
feat(DST-1527): document the Badge variant prop

sarahgm:
**Throwaway. Do not merge. Do not review by hand.**

Smoke test for the Bedrock path on #5786, per step 2 of that PR's test plan. Based on
`feat/DST-1527-ai-review-ci` so the diff stays at one file while the workflows still
resolve from the head branch.

Title type is `feat`, so `classify` should pick the `full` profile and `Baseline review`
should actually attempt the AWS assume rather than skip.

What this is checking, in order:

1. `Check the Bedrock configuration is present` passes, meaning the role ARN, region and
   model id are all set.
2. `Assume the Bedrock role` succeeds, meaning the `marigold-ci` role exists and its trust
   policy accepts `repo:marigold-ui/marigold:pull_request`.
3. Exactly one sticky review comment appears.

Closing this as soon as the run reports.


sarahgm:
Test complete. `Assume the Bedrock role` failed with `The web identity token provided could not be validated`, so the OIDC trust anchor is not in place yet. Closing.

